PR #8, opened 2016-12-18
Add rename functionality

alanz:
This introduces the function `lsp--apply-text-edits` which should perhaps be used in one or two other places too, for the formatting response.

I have not done this. as I do not have a means to test.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -259,6 +259,29 @@ interface Range {
   (lsp--range (lsp--point-to-position start)
 	      (lsp--point-to-position end)))
 
+(defun lsp--apply-workspace-edits (workspace-edits)
+  ;; (message ":workspace-edit: %s" workspace-edits)
+  (maphash (lambda (k v)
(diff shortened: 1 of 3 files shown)


lsp-methods.el
```diff
@@ -259,6 +259,26 @@ interface Range {
   (lsp--range (lsp--point-to-position start)
 	      (lsp--point-to-position end)))
 
+(defun lsp--apply-workspace-edits (workspace-edits)
+  (maphash (lambda (key value) (lsp--apply-workspace-edit key value)) (gethash "changes" edits))
```
vibhavp: What is edits supposed to be here? Shouldn't it be workspace-edits?

alanz: See https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#workspaceedit

It is accessed via the "changes" key.  As presented in emacs-lsp when the reply is sent, and working in vscode.

vibhavp:
Pushed with a few changes, thanks!

Issue #14, opened 2017-01-07
Process lifecycle not managed


alanz:
Prior to exiting, each process should receive a [ShutdownRequest](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#shutdown) followed by an [Exit Notification](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#exit-notification).

This does not currently happen, and I gather the subprocess has to exit on a SIGHUP from emacs.

alanz:
@vibhavp , perhaps we should consider doing this: https://github.com/jsierant/lsp-mode/commit/b4c125dbb02804add5ea67c32aa04b831d3ccc43

yyoncho:
@alanz at this point lsp-mode sends the required notifications but it kills the server right after that. Can you advice whether this is the correct behaviour or lsp should wait for the server to tear down by itself and kill it only if it takes too long?

alanz:
The protocol says that the client should send a shutdown request, wait for the ack then send the notification.

I can imagine an LSP server that receives the request, does an orderly flush of it's cache, saves values that need to be persisted, then sends the ack.

For servers not using this notification to persist state, I don't think it really matters.  But my preference would be to do it according to the spec, with a timeout.

yyoncho:
It seems like we are ok after https://github.com/emacs-lsp/lsp-mode/pull/689 .

PR #15, opened 2017-01-07
Add global hook to shutdown the language servers on exit

alanz:
This is incomplete, it should first send a shutdown message and wait for the
response before sending the exit notification.

Partly completes #14

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -157,3 +157,18 @@ Optional arguments:
     (switch-to-buffer "lsp-capabilities")))
 
 (provide 'lsp-mode)
+
+;; Clean up the entire state of lsp mode when Emacs is killed, to get rid of any
+;; pending language servers.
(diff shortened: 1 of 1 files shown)


vibhavp:
The rust language server doesn't follow the spec here, and doesn't send a response to the `shutdown` request. I plan to add a way to send requests without waiting for their responses, after which I can add functionality for clean server shutdowns.

alanz:
Well, the reason there is a wait for the response is if the server takes a while to e.g. save its current state. 

Maybe add a setting for the rust server to ignore the wait.

alanz:
Perhaps file a bug report against the rust server, it won't work with vscode, which does honour the shutdown lifecycle.

alanz:
rust is now compliant, see https://github.com/rust-lang-nursery/rls/issues/258#issuecomment-294191309

PR #19, opened 2017-01-19
Add missing initial value

syohex:
Initial value must be necessary if `:read-only` attribute is specified. I got following error at loading library on Emacs 26.0.50.1

```
error Invalid options for slot range in lsp-diagnostic
```


vibhavp:
Thanks

PR #27, opened 2017-02-11
Add MELPA badge

Wilfred:
I didn't realise that this package was available on MELPA initially, so I thought I'd add a badge to clarify.


vibhavp:
Thanks

PR #28, opened 2017-03-01
Add flycheck support

fmdkdd:
I've added flycheck support, by hooking into `lsp--on-diagnostics` and adding a generic checker that transforms the `lsp-diagnostic` struct into `flycheck-error`.

I've not activated this suppor by default.  One should run:
```
 (require 'lsp-flycheck)
 (lsp-flycheck-setup)
```
beforehand.

After that, whenever `lsp-mode` gets diagnostics back from the language server, `lsp-flycheck` start `flycheck-buffer` with the collected diagnostics.

I've only tested it with the latest RLS.

### DIFF
--- a/lsp-flycheck.el
+++ b/lsp-flycheck.el
@@ -1,0 +1,78 @@
+;;; lsp-flycheck.el --- Flycheck support for lsp-mode -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2017  fmdkdd
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


vibhavp:
looks good. Thanks!

fmdkdd:
Great!  Feel free to tinker with it.

vibhavp:
btw, wouldn't disable automatic syntax check for flycheck _everwhere_? That probably isn't desirable, so is there a way to the next timed check do nothing?

fmdkdd:
Hmm right, I thought it was buffer-local without checking it.  You should use `setq-local` rather to set it only in the current buffer.

vibhavp:
Alright, thanks.

PR #30, opened 2017-03-01
Fix lsp-capabilities: mismatch between selector and data structure

ruimams:
Fixes the bug described in vibhavp/emacs-lsp#29.


vibhavp:
Thanks!

PR #31, opened 2017-03-11
Fix package descriptor

purcell:
- Dependency on flycheck was missing, so byte compilation would fail
- Remove redundant "Emacs" from package description: all elisp packages are for Emacs


purcell:
Replaced by #32

PR #32, opened 2017-03-11
Misc fixes

purcell:
I've been getting interested in LanguageServer support for Emacs and thought I'd send you a few little fixes. Hope these are helpful.

You should also provide a minor-mode and global minor-mode pair, rather than only a global minor mode: it should be possible for users to enable/disable lsp support on a per-buffer or per-project basis using the regular Emacs support for mode hooks and/or directory local variables. That might look something like this:

```el
;;;###autoload
(define-minor-mode lsp-mode
  "Minor mode for interaction with a Language Server."
  nil " LSP" nil
  (if lsp-mode
      (progn
        (add-hook 'find-file-hook #'lsp-on-open nil t)
        (add-hook 'after-save-hook #'lsp-on-save nil t)
        (add-hook 'kill-buffer-hook #'lsp-on-close nil t)
        (lsp-on-open))
    (remove-hook 'find-file-hook #'lsp-on-open t)
    (remove-hook 'after-save-hook #'lsp-on-save t)
    (remove-hook 'kill-buffer-hook #'lsp-on-close t)
    (lsp-on-close)))

(defun lsp--enable-maybe ()
  (when (and (not (minibufferp))
             (buffer-file-name)
             (file-exists-p (buffer-file-name)))
    (lsp-mode 1)))

;;;###autoload
(define-globalized-minor-mode global-lsp-mode lsp-mode lsp--enable-maybe)
```


vibhavp:
I've been wanting to replace the global minor mode with a LSP minor mode that hooks in to other language mode, would that be better?

purcell:
Yes. Separating the minor and global minor modes as above would allow users to add the (non-global) minor mode to their language mode hooks as desired. (That's not something you should do on their behalf: just make it easy, and perhaps provide examples.)

purcell:
P.S. Have you seen https://github.com/sourcegraph/emacs-lsp ? You guys should join forces!

vibhavp:
Indeed. I'd like an eventual merge at someone point. Thanks!

PR #35, opened 2017-04-01
Make flycheck support rust-agnostic

### LINKED ISSUE #33: lsp-flycheck is rust specific
I'm trying out lsp-mode with [PyLS](https://github.com/palantir/python-language-server) and activated `lsp-flycheck` with `lsp-flycheck-setup` but it failed since it wanted to run with the Rust checker. 

See [these lines](https://github.com/vibhavp/emacs-lsp/blob/master/lsp-flycheck.el#L62-L63) for why.

fmdkdd:
Flycheck still requires an explicit list of modes, so we can't leave that undefined.

For the moment I've added `python-mode` and `go-mode`, but did not test them.

Should fix #33.

### DIFF
--- a/lsp-flycheck.el
+++ b/lsp-flycheck.el
@@ -25,7 +25,6 @@
 
 ;;; Code:
 
-(require 'lsp-mode)
 (require 'lsp-notifications)
 (require 'flycheck)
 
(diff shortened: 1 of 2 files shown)


fmdkdd:
Got the impression that @mattiasb was trying `lsp-mode` with PyLS, and the `:modes` of the checker was an impediment.

mattiasb:
Yeah, but just adding python-mode if that isn't also defined in the mode doesn't seem like a solid solution. :/

I hope that language-servers became ubiquitous enough that the lsp definitions would be added in the various major-modes instead of in here.

Meanwhile, would it be possible to make the list of modes to add to flycheck a defcustom?

lsp-flycheck.el
```diff
@@ -58,19 +58,19 @@ CALLBACK is the status callback passed by Flycheck."
   "Setup Flycheck for use with lsp-mode."
   ;; Disable automatic syntax checks, since lsp-mode will call `flycheck-buffer'
   ;; directly.
-  (add-to-list 'flycheck-checkers 'rust-lsp)
   (setq-local flycheck-check-syntax-automatically nil
               flycheck-checker 'rust-lsp)
```
mattiasb: This needs to be 'lsp

mattiasb: Or rather, it needs to be the same as the checker defined below. But that checker should be namespaced properly, so probably `lsp-flycheck-checker`.

mattiasb: Hm, unless that gets namespaced by flycheck in some weird way. *confused*

fmdkdd: This indeed needs to be `'lsp`.  There is namespacing done through lisp symbols, which can act as dictionaries.

fmdkdd:
I agree this is a crutch.  The ideal situation would be to have the generic checker be enabled whenever `lsp-mode` is also enabled.  But, the `:modes` property is for major modes only, so we cannot put `lsp-mode` in there.  And it cannot be empty.

What about adding the corresponding major modes in `lsp-define-client`?

fmdkdd:
With the new patch, the `'lsp` checker now has `go-mode haskell-mode rust-mode` set by `lsp-define-client`.  Should you add Python support through `define-client`, the checker will follow.

lsp-flycheck.el
mattiasb: Ok cool!

mattiasb:
Awesomesauce!

PR #41, opened 2017-04-06
eclipse.jdt.ls for java support

tutysara:
Has methods to initialize a java language server based on rust server.
Optionally downloads and installs the server.


vibhavp:
Thanks. Could you remove `jls--download-server`, and use the `lsp--`namespace? Downloading the server is something I'd like to be left to the user.

tutysara:
Sure, let me change and send a pull request

PR #48, opened 2017-04-08
Remove confusing startup java message

rski:
When lsp-global-mode is enabled, a confusing "getting java root"
message appeared regardless of what kind of buffer is currently
open.


tutysara:
yeah, this is a test message that should be removed, ty.

PR #49, opened 2017-04-08
Don't depend on a hardcoded java ls jar version

rski:
The java language server has no stable release. However, there is no
reason to depend on an arbitrary snapshot of the java language
server. The version is a leftover of the initial pull request adding
support for the java language server, which installed the server as
well, so we can remove it.

### DIFF
--- a/lsp-jls.el
+++ b/lsp-jls.el
@@ -11,8 +11,6 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
-(defconst lsp-jls-jar-version "1.4.0.v20161219-1356")
-
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


tutysara:
I kept the jar version because we can baseline on a version of eclipse.jdt to support. I see there are 3 versions of the protocol at present and it is still evolving. So, idea is to support the version of jdt jar that speaks the same protocol version as emacs-lsp.

rski:
imo, it's not moving that fast to justify pinning it right now, it's mostly bugfixes and such. The pinning only makes it a pain if you want to use a newer version.

vibhavp:
Is it possible to add a `defcustom` that the user could change to decide which server version needs to be used?

rski:
The user can select which version of the server to use by downloading that version, I don't see a point in also asking them to put that version in their emacs config.

vibhavp:
I see. Thanks.

rski:
thanks for merging :-) I'll play around a bit more and try to contribute more fixed if needed

PR #64, opened 2017-04-21
Bring in some LSP 3.0 features

alanz:
- Send a `codeAction` request based on current point
- Start on UI elements to choose which code actions to apply, based on the intero code
- Send expanded client capabilities with `initialize`
- Introduce `executeCommand` and expose it for language specific use
- manage request messages from the server, starting with `workspace/applyEdit`


vibhavp:
Looks good, thanks.

PR #68, opened 2017-04-26
Request and apply codeActions

alanz:
As discussed with @vibhavp , this needs some polishing.

However it is useful, and I would like to see it in master, so I can continue making other changes based on it.


vibhavp:
I'll be occupied with exams for the next 10 days. Will take a look at this
ASAP.

alanz:
Thanks. You can clean it up at your leisure, when you have more time.

vibhavp:
Yeah, I'll see how much this can be shifted to easymenu and friends.
Also, could you see when and how frequently do other editors request code actions? I'm not sure if doing it on every hover is the right way to go.

PR #74, opened 2017-05-03
Make `lsp-mode` a non-global mode, invoked by hooking into a major-mode.

vibhavp:
Other changes:
* Indentation is enabled for servers that support it.
* Symbols under point are highlighted.


vibhavp:
@alanz could you see if this breaks anything with Haskell? `lsp-mode` is enabled with `(add-hook 'haskell-mode-hook #'lsp-mode)`.

alanz:
checking now ...

alanz:
I am getting something in messages about `global-lsp-mode` here: https://github.com/emacs-lsp/lsp-mode/blob/lsp-start/lsp-flycheck.el#L65

alanz:
But it seems to startup ok. I am now receiving `textDocument/documentHighlight` requests, which is not in the capabilities my server sends back.

PR #76, opened 2017-05-04
Shutdown hook

alanz:
Add global shutdown handler 

Otherwise the lsp server is just killed, with no chance to clean up, if required

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -280,6 +280,23 @@ interface TextDocumentItem {
   (lsp--send-notification (lsp--make-notification "exit" nil))
   (lsp--uninitialize-workspace))
 
+
+;; Clean up the entire state of lsp mode when Emacs is killed, to get rid of any
+;; pending language servers.
+(add-hook 'kill-emacs-hook #'lsp--global-teardown)
+
+(defun lsp--global-teardown ()
+  (maphash (lambda (key value) (lsp--teardown-client value)) lsp--workspaces)
+  )
+
+(defun lsp--teardown-client (client)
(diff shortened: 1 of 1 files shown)


lsp-methods.el
```diff
@@ -1115,6 +1146,8 @@ command COMMAND and optionsl ARGS"
     (list :command cmd)))
 
 (defalias 'lsp-point-to-position #'lsp--point-to-position)
+(defalias 'lsp-get-start-position #'lsp--get-start-position)
```
vibhavp: Why do we need to make this a public alias?

lsp-methods.el
```diff
@@ -1115,6 +1146,8 @@ command COMMAND and optionsl ARGS"
...
+(defalias 'lsp-get-end-position #'lsp--get-end-position)
```
vibhavp: Same here

vibhavp:
Added two review comments.

alanz:
@vibhavp you have to submit the review comments when done, otherwise I do not see them.

vibhavp:
Should've known that, sorry.

alanz:
I discovered it this morning, doing the same thing. :)

lsp-methods.el
alanz: I use them in lsp-haskell. Is there another way of doing it?

lsp-methods.el
alanz: ditto.

lsp-methods.el
vibhavp: Ah, that's okay then.

Issue #77, opened 2017-05-04
Toggling the lsp-mode does not actually toggle the server


rski:
If lsp-mode was active, doing M-x lsp-mode (toggle off) M-x lsp-mode (toggle on again) doesn't try to stop and then start the server again.

As a result of this, if the server for some reason dies, it is not possible to restart it in the current project unless you restart emacs or muck with variables.

vibhavp:
Ah, never thought of that. I'll try adding toggle functionality, thanks

alanz:
Also, see the approach in https://github.com/commercialhaskell/intero/issues/364

There is both a blacklist and a whitelist

alanz:
Perhaps toggling `lsp-mode` in the presence of the `lsp-project-blacklist` should offer to add it to the `lsp-project-whitelist`.

Alexander-Miller:
> it is not possible to restart it in the current project unless you restart emacs

Not even closing emacs works. It tries to shut down the lsp process, sees that's it's not live, and throws an error (this happens in lsp--stdio-send-(a)sync). I've had to use xkill to exit emacs when that happened.

astahlman:
I believe the error that @Alexander-Miller referenced is triggered in the process sentinel, here:

https://github.com/emacs-lsp/lsp-mode/blob/bb936fdfdcc0bbf33d02b66ad015cf437cc2fcd3/lsp-methods.el#L369

The function references `lsp--cur-workspace`, but by the time the sentinel runs, the workspace has already been set to nil, here:

https://github.com/emacs-lsp/lsp-mode/blob/bb936fdfdcc0bbf33d02b66ad015cf437cc2fcd3/lsp-methods.el#L313

Regarding the original proposal: correct me if I'm wrong, but wouldn't toggling the minor mode in a single buffer affect **all** buffers associated with the project workspace? In other words, if buffer A and buffer B are both visiting files under the same project root directory, and I toggle lsp-mode in buffer A, then the language server would "disappear" from the perspective of buffer B.

Is that what's intended here? As a user, I think this would be surprising and not very intuitive. It seems cleaner (albeit more difficult) to decouple the lifecycle of the language server from the status of the minor-mode in each buffer.

astahlman:
@alanz I see that you were working on this in https://github.com/emacs-lsp/lsp-mode/pull/82. Any comment on the relationship between the the language server process and the individual documents in a workspace?

IMHO, the relationship between:

- the language server process(es)
[...]

PR #81, opened 2017-05-06
Bring in lsp-project-whitelist

alanz:
As an alternative to 'lsp-project-whitelist', this allows listing the projects
for which to use LSP mode.

The current logic is that if the whitelist is not nil, then the blacklist is
ignored.  This can be changed if a different policy makes sense, the decision
process is broken out into 'lsp--should-start-p'

cc #77

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -389,7 +389,6 @@ disappearing, unset all the variables related to it."
   (let* ((client (lsp--get-client t))
           (root (funcall (lsp--client-get-root client)))
           (workspace (gethash root lsp--workspaces))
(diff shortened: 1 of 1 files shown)


vibhavp:
Looks good, do you think we should warn the user if both the whitelist and blacklist are set?

alanz:
We could do that.

I was thinking that maybe if it is in the whitelist we should always start, and maybe warn if it is in the blacklist too.

Otherwise they should be able to coexist, if we change the logic in the checking function.

vibhavp:
Yeah, a warning would do it.

PR #82, opened 2017-05-06
Toggle a workspace

alanz:
I am stuck on this, when I turn off a workspace, after a few seconds I get

```
Shutting down workspace for /home/alanz/tmp/haskell-hie-test-project/
set lsp--cur-workspace to nil
workspace uninitialized: /home/alanz/tmp/haskell-hie-test-project/
error in process sentinel: or: Wrong type argument: lsp--workspace, nil
error in process sentinel: Wrong type argument: lsp--workspace, nil
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Clone this repository to a suitable path, and add
 (with-eval-after-load 'lsp-mode
     (require 'lsp-flycheck))
 (require 'lsp-mode)
-(add-hook 'prog-major-mode #'lsp-mode)
+(add-hook 'prog-major-mode #'lsp-mode-hook)
 ```
 to your .emacs, where `prog-major-mode` is the hook variable for a supported
 programming language major mode.
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -303,14 +303,13 @@ interface TextDocumentItem {
   )
 
 
-
-
 (defun lsp--uninitialize-workspace ()
   "When a workspace is shut down, by request or from just
 disappearing, unset all the variables related to it."
   (remhash (lsp--workspace-root lsp--cur-workspace) lsp--workspaces)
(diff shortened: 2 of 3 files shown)


vibhavp:
Looks like the sentinel also runs when `lsp--uninitialize-workspace` kills the language server process (which in return also calls lsp-uninitialize-workspace, causing the error). You can try separating the `kill-process` code from `lsp--uninitialize-workspace`.

A way to fix this would be to make `shutdown-workspace` not call `uninitialize-workspace`, and adding the `kill-process` form to it.

alanz:
Thanks, that has sorted it out.

alanz:
I think this is ready for review/merge

vibhavp:
Hm, toggle functionality should be exposed to the user by the mode entry point function (`lsp-mode`, in our case). Could you make `lsp--toggle` non-interactive, and call `lsp--toggle` instead of `lsp--start` in the `define-minor-mode` form in `lsp-mode.el`?

alanz:
I knew that had to happen, but I was not sure exactly how to do it, so I focussed on the actual switching process initially.

Will update.

alanz:
And that actually changes the dynamics, because now, if the whitelist is not `nil`, it starts a session and asks if it should be added to the whitelist.

What I would expect to happen is

1. If a haskell file is opened, and it is on the whitelist, then an LSP server is started. If not, then not, silently.

2. If a file is opened without an LSP server (through not being on the whitelist), then the `lsp-mode` command should start a session, and ask to add it to the whitelist.

In other words, the initial call to start the mode needs to be conditional, and differentiate between being called interactively vs being called as a byproduct of a file being opened somehow.

alanz:
I suspect this is not honouring the general minor mode protocol, in terms of prefix args.

I guess that will be the next thing to patch. I will only be able to look at this again in a few days time though.

alanz:
Superseded by #141

PR #88, opened 2017-05-10
Fix rootUri in initialize

smarter:
- local URIs should start with file://
- Send rootUri from `lsp--start` too


vibhavp:
Thanks.

PR #92, opened 2017-05-31
README.md: replace hook name (prog-mode-hook)

gicmo:
The `prog-major-mode` didnt work for me and a quick google left me questioning its existence, so I assume `prog-mode-hook` was meant instead.


vibhavp:
It's meant to be placeholder for the language major mode's hook (`python-mode-hook`, for instance).

gicmo:
Ah, now I understand. Sorry.

PR #94, opened 2017-06-23
Add support for request "completionItem/resolve"

sham1:
This commit might be useful for gathering things like documentation when you're completing a function, so you can compare different functions before deciding which one to use (This functionality is especially useful with company-mode). This can also be useful for different kinds of completion packages that support YASnippet-like snippet insertion and editing where the snippet can be provided by `CompletionItem::insertText`, assuming obviously that `CompletionItem::insertTextFormat` is set to be `InsertTextFormat::Snippet`.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -879,7 +879,19 @@ to a text document."
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (comp)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (unless (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (user-error "This language server doesn't support resolving completion items"))
+  (when (null comp)
+    (user-error "Completion item must not be nil"))
+  (let* ((item (cond
+		((hash-table-p comp) comp)
+		((stringp comp) (plist-get (text-properties-at 0 comp) 'lsp-completion-item)))))
+    (cl-reduce #'list (lsp--send-request (lsp--make-request
+					  "completionItem/resolve"
+					  item)))))
 
 (defun lsp--location-to-xref (location)
   "Convert Location object LOCATION to an xref-item.


vibhavp:
Have you tried testing this with a completion backend that supports resolving completion candidates? AFAIK, the native system in emacs doesn't.

lsp-methods.el
```diff
@@ -879,7 +879,19 @@ to a text document."
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (comp)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (unless (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (user-error "This language server doesn't support resolving completion items"))
```
tigersoldier: Can you simply return the completion item? Callers shouldn't worry about it.

lsp-methods.el
```diff
@@ -879,7 +879,19 @@ to a text document."
...
+  (when (null comp)
+    (user-error "Completion item must not be nil"))
+  (let* ((item (cond
+		((hash-table-p comp) comp)
+		((stringp comp) (plist-get (text-properties-at 0 comp) 'lsp-completion-item)))))
```
tigersoldier: I think it's better to just accept a hashtable for the CompletionItem message. The callers should be responsible to get it from the completion candidate string.

lsp-methods.el
```diff
@@ -879,7 +879,19 @@ to a text document."
...
+    (cl-reduce #'list (lsp--send-request (lsp--make-request
```
tigersoldier: The return value is a hashtable of the updated completion item. Why reducing it to a list?

tigersoldier:
I'm planning to support resolving completion items in company-lsp. Would love to see this feature added to upstream and use it.

lsp-methods.el
sham1: Like the current one? Yes, that is indeed one possibility.

lsp-methods.el
sham1: Yes, I agree. I really don't know what I was thinking here.

sham1:
Welp, I somehow managed to screw up my repository beyound belief when I updated it...

So, I'll close this one and open another one.

sham1:
Welp, I somehow managed to screw up my repository beyound belief when I updated it...

So, I'll close this one and open another one.

lsp-methods.el
tigersoldier: Consider writing a function `lsp--get-completion-doc`. One may write:

```elisp
(defun lsp--get-completion-doc (item)
  (or (gethash "documentation" item)
      (gethash "documentation" (lsp--resolve-completion item)))
```

This will throw an error if the server doesn't support resolving completion items. To write an error-free implementation, it has to either catch the error, or check `resolveProvider` before calling.

Simply return `item` when `resolveProvider` is nil can avoid that.

sham1: Ah yes, I did do that on that new PR of mine (because this one was being quite silly indeed and wouldn't update after this particular change).

Do you recon that we should also return a nil completion item if someone passes us nil?

tigersoldier: Will comment on the new PR.

PR #99, opened 2017-08-14
Recognise sync method for version 3 protocol too

alanz:


lsp-methods.el
```diff
@@ -337,11 +337,18 @@ disappearing, unset all the variables related to it."
   (lsp--workspace-server-capabilities lsp--cur-workspace))
 
 (defun lsp--set-sync-method ()
-  (setq lsp--server-sync-method (or lsp-document-sync-method
-                                  (alist-get
-                                    (gethash "textDocumentSync"
-                                      (lsp--server-capabilities))
-                                    lsp--sync-methods))))
+  ;; (message "lsp--set-sync-method called") ;; AZ-DEBUG
```
vibhavp: Also, could you remove these? I would like to add some logging functionality later, so we can correctly do this.

PR #100, opened 2017-08-15
Fix xref calling textDocument/definition

alanz:


vibhavp:
What does this fix? IIRC, the xref docs recommend `xref-backend-definitions` to fetch positional parameters from the symbol itself.

alanz:
For me the `params` end up as `null` in the message, so the backend returns an error message.

In fact, the `properties` are `nil` too, iirc.

The identifier has a value, but `textDocument/definition` needs a position.

vibhavp:
How are you invoking xref? Did you use `xref-find-definitions`?



-- 
Vibhav Pant
vibhavp@gmail.com

alanz:
yes.

And spacemacs `gd` which invokes `spacemacs/jump-to-definition`.

vibhavp:
This doesn't seem to be the case in `lsp-python`, though. Does this only happen with Haskell?

alanz:
I have only tried it with haskell.

It sounds like it needs to be conditional then, if the params end up `nil` then use the current position.

Do the params end up as a properly formatted message for `textDocument/definition` when using python?

alanz:
I just ran it against python, using the following

```elisp
(cl-defmethod xref-backend-definitions ((_backend (eql xref-lsp)) identifier)
  (let ((properties (text-properties-at 0 identifier))
        (params (plist-get properties 'def-params))
        (def (lsp--send-request (lsp--make-request
                                 "textDocument/definition"
                                 (lsp--text-document-position-params)))))
    (message "xref-backend-definitions: %s %s" properties params)
    (if (consp def)
        (mapcar 'lsp--location-to-xref def)
      (and def `(,(lsp--location-to-xref def))))))
```

What I get in `*Messages*` when I call `xref-find-definitions` is

```
Symbol’s value as variable is void: properties
```

In other words the same problem I get in haskell.

This is using spacemacs, I do not know how to invoke the language server using `emacs -Q`

alanz:
And I see in the `xref` documentation that 

> IDENTIFIER can be any string returned by
> ‘xref-backend-identifier-at-point’, or from the table returned by
> ‘xref-backend-identifier-completion-table’.

and tha `xref-backend-identifier-at-point` is simply a string, as I am seeing.

So it appears it can compliantly be either a string or something more structured.

So, is the current POINT something we can assume, or only fall back to it in the unstructured IDENTIFIER case?

vibhavp:
It's actually a propertize string: https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-methods.el#L1080

That's where we get the position and reference params from.

alanz:
I don't really know this stuff, and it is marked as experimental in `xref.el`.   I did find [this](https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/xref.el#L266)

What I do know, is that when I run it against both the python and haskell backends there are no properties.

And I am using vanilla spacemacs.  So the non-property version is present in the wild, I believe we need to explicitly deal with it.

alanz:
And now that I look more closely at your code I share your puzzlement. Let me investigate further.

vibhavp:
Also, you seem to be using an old version of `lsp-mode`. Try updating it.

alanz:
Ok, I did a hard reset to master on my repo, and it now behaves as expected.

I was running against my (rebased) toggle branch, which is obviously messing something up.

alanz:
But I have learned a whole lot :)

alanz:
And I see what is happening.

In haskell-mode, under spacemacs, `xref-find-definitions` works as expected.  But `gd` which is bound to a function to go to a definition also ends up calling `xref-backend-definitions`, and the properties are not set in this case.

The normal definition search is into a tags file, which is also loaded. So I suspect it is going through all of them to find a match.

I am going to rework this to fall back for this case.

alanz:
It gets called directly here: https://github.com/haskell/haskell-mode/blob/master/haskell.el#L350

PR #101, opened 2017-08-15
Make xref-backend-definitions more robust

alanz:
In some modes, such as haskell-mode, xref-find-definitions gets called without
setting the expected property values.

Check for this, and use the current point location instead if necessary.

This is take 2 on https://github.com/emacs-lsp/lsp-mode/pull/100

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1092,11 +1092,17 @@ interface DocumentRangeFormattingParams {
 ;;   nil)
 
 (cl-defmethod xref-backend-definitions ((_backend (eql xref-lsp)) identifier)
-  (let* ((properties (text-properties-at 0 identifier))
-          (params (plist-get properties 'def-params))
-          (def (lsp--send-request (lsp--make-request
-                                    "textDocument/definition"
-                                    params))))
+  (let* ((maybeparams (get-text-property 0 'def-params identifier))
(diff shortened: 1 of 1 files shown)


lsp-methods.el
```diff
@@ -1092,11 +1092,17 @@ interface DocumentRangeFormattingParams {
...
-          (params (plist-get properties 'def-params))
-          (def (lsp--send-request (lsp--make-request
-                                    "textDocument/definition"
-                                    params))))
+  (let* ((maybeparams (get-text-property 0 'def-params identifier))
+         ;; In some modes (such as haskell-mode), xref-find-definitions gets
+         ;; called directly without applying the properties expected here. So we
+         ;; must test if the properties are present, and if not use the current
+         ;; point location.
+         (params (if (symbolp maybeparams) ;; if it is a symbol it is not an assoc
```
vibhavp: `get-text-property` returns `nil` when the prop value doesn't exist. Wouldn't `(null maybeparams)` be a clearer check instead?

alanz: I will try it. My elisp knowledge is weak, I tried testing `eq nil maybeparams` which did not work, and also just `(if maybeparams ..`) which also did not work. Trial and error gave me that code, and I was hoping to be advised on a more appropriate way to do it.

vibhavp:
LGTM

PR #103, opened 2017-08-16
Changes command to be a function pointer

bkchr:
This patch changes the command to be a function pointer. In my setup, cargo, rls etc are not in global path, so I need to setup the path before lsp can find the programs. By making the command a function pointer, the function is called when the lsp-mode is activated and not when registered. This gives me the chance to setup the PATH properly.

For lsp-rust I added a pull request that reflects the required changes: https://github.com/emacs-lsp/lsp-rust/pull/7

If you would like to merge this, I could also adapt the other repos.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -30,9 +30,10 @@
 (require 'cl-lib)
 (require 'network-stream)
 
-(defun lsp--make-stdio-connection (name command)
+(defun lsp--make-stdio-connection (name command command-fn)
   (lambda (filter sentinel)
-    (let ((final-command (if (consp command) command (list command))))
+    (let ((command (if command-fn (funcall command-fn) (command))))
+      (let ((final-command (if (consp command) command (list command))))
       (unless (executable-find (nth 0 final-command))
         (error (format "Couldn't find executable %s" (nth 0 final-command))))
       (make-process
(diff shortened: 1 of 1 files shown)


vibhavp:
Thanks. However, this would probably break someone else's setup. Instead, you could add a new parameter `:command-fn` which returns the appropriate command to run.

bkchr:
And how would I implement that for rust for example? 

You really think that finding rls when the client is created, would break other users setups?

vibhavp:
> You really think that finding rls when the client is created, would break other users setups?

Perhaps. Some users might have their own defined clients, and changing the intended type of `:command` would break that. A way to implement this would be to add an optional `command-fn` param to both `lsp--make-*-connection` and `lsp-define-*-connection` variants and use that to obtain the command if it's non-nil.

bkchr:
Okay, I added the command-fn as extra argument, as you proposed :)

vibhavp:
Looks good, could you send the corresponding PR to `lsp-rust` too?

bkchr:
I already opened the pull request for lsp-rust :) https://github.com/emacs-lsp/lsp-rust/pull/7

bkchr:
@vibhavp will you merge these changes and the changes for lsp-rust? :)

vibhavp:
Crap, forgot this one. Sorry!

PR #107, opened 2017-08-27
Speed up response parsing

gpittarelli:
Currently, lsp-mode freezes everything up for a while when trying to process huge responses from the language server (some can send back multi-megabyte responses when, eg, trying to autocomplete in the global scope). Profiling + logging eventually revealed that practically all the time and memory usage is by `lsp--parser-read`. I suspected that doing all the parsing a single character at a time was the culprit (tons of allocations, etc.). So I rewrote the parser to handle the data in large chunks instead, storing `leftovers` between parsing each chunk. This makes reads go way faster.

The same use case, completing global scope with https://github.com/sourcegraph/javascript-typescript-langserver now takes ~1s. Previously it froze up for multiple minutes.

### DIFF
--- a/lsp-receive.el
+++ b/lsp-receive.el
@@ -24,13 +24,10 @@
 (cl-defstruct lsp--parser
   (waiting-for-response nil)
   (response-result nil)
-  (cur-token nil) ;; the current token being parsed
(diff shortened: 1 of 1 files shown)


vibhavp:
SeemsGood

PR #108, opened 2017-08-29
Make parsing even faster, and use less memory

gpittarelli:
Further refinement to https://github.com/emacs-lsp/lsp-mode/pull/107. `cpu+mem` profiler in emacs still showed a global scope completion with javascript-typescript-langserver chewing through 119MB and GC was taking up 28% of the runtime. Profiling still pointed towards `lsp--parser-read`. I suspected constantly concat'ing chunks was making a ton of extra allocs; eg we'd get a 4k chunk initially, then we'd alloc a new 8k chunk when (to hold leftover 4k + new chunk), then 12k, 16, ... compounding until we reached the full megabyte response.

This change preallocates a single big string of the total Content-Length size, then fills it as we go. This cuts memory usage to around 29M and GC cpu time to ~23%.

Note: trusting the server to not set a malicious Content-Length...

As an aside, `json-read-from-string` is now the the single biggest CPU bottleneck, followed by GC and then various internal emacs completion stuff. So I guess we're good for now?

Number are from running a test script with `emacs -q --script`. Profiles are in the zip (open with `profiler-find-profile`)
[profiles.zip](https://github.com/emacs-lsp/lsp-mode/files/1259295/profiles.zip)

### DIFF
--- a/lsp-receive.el
+++ b/lsp-receive.el
@@ -27,7 +27,9 @@
   (headers '()) ;; alist of headers
   (body nil) ;; message body
   (reading-body nil) ;; If non-nil, reading body
-  (leftovers "") ;; Leftover data from previous chunk; to be processed
+  (body-length nil) ;; length of current message body
+  (body-received 0) ;; amount of current message body currently stored in 'body'
+  (leftovers nil) ;; Leftover data from previous chunk; to be processed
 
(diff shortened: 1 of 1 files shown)


vibhavp:
Thanks!

PR #110, opened 2017-08-30
Handle bytes better

gpittarelli:
Okay, hopefully the last touches needed on the parser for a while.

1. Fix a freeze and/or error happening due to multibyte characters in responses. This is done by changing the `:coding` of the server processes to be `no-conversion`, which basically causes the strings to act like byte buffers. We then decode the message as utf-8 only once we have the whole thing.
2. Change `lsp--parser-read` to return a list of parsed messages (there can be multiple in a single chunk), instead of calling straight into `lsp--parser-on-message`. This change makes testing easier since we can just check the return value instead of having to mock `lsp--parser-on-message`. 
3. Add `ert` test cases. Run with `emacs -batch -l ert -L . -l test/lsp-receive-tests.el -f ert-run-tests-batch-and-exit`

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -39,6 +39,7 @@
       (make-process
         :name name
         :connection-type 'pipe
+        :coding 'no-conversion
         :command final-command
         :filter filter
         :sentinel sentinel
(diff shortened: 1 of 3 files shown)


vibhavp:
LGTM, thanks.

PR #113, opened 2017-08-30
Send changes immediately instead of queueing them

### LINKED ISSUE #112: textDocument/didChange interpretation
At the moment we accumulate changes and send them once a timer triggers.

Each emacs onChange event is wrt to the document at the point when the change is made.

This means the order of application is important, and needs to be preserved.

However

According to https://github.com/Microsoft/language-server-protocol/pull/279, the changes in a single `textDocument/didChange` message should all be wrt the same document, and hence they need to be applied in reverse order.

So, to fix this we can either

- Send one `textDocument/didChange` per emacs change, as it is reported; or
- Adjust the edit locations so that they all refer to the original document, so they can be applied according to the specification.

The first option is simpler, and seems to be what happens in vscode.  But, it means we potentially run diagnostics generation more frequently than required.

alanz:
For incremental change mode, all the changes in a `textDocument/didChange`
message are interpreted as applying to the same version of a particular
document.

But each emacs change corresponds to the state of the document when the change
happens. So harmonise the expectations by sending each change in its own
message.

closes #112

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -733,7 +733,14 @@ to a text document."
     (setq lsp--has-changes t)
     (lsp--rem-idle-timer)
     (when (eq lsp--server-sync-method 'incremental)
-      (lsp--push-change (lsp--text-document-content-change-event start end length)))
+      ;; (lsp--push-change (lsp--text-document-content-change-event start end length)))
+
+      ;; Each change needs to be wrt to the current doc, so send immediately.
+      ;; Otherwise we need to adjust the coordinates of the new change according
+      ;; to the cumulative changes already queued.
+      (progn
+        (lsp--push-change (lsp--text-document-content-change-event start end length))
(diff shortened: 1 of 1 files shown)


vibhavp:
Text changes strike again, argh. I'll need to see how much does this end up hurting performance, because sending a message to the language server on every change to the buffer might slow up emacs considerably :/

alanz:
It should be possible to come up with a function that can go over a list of changes and re-align them to the original state, as expected by LSP.   But it would have to be tested pretty comprehensively.

And I am not sure that my elisp is up to it.

alanz:
However, without this patch the incremental state in the server is guaranteed to be wrong.

vibhavp:
Yeah, that's a big problem. I'll merge this for now, though we need to find a way to coalesce changes to a buffer.

Issue #114, opened 2017-09-14
Calculate wrong end position after kill a line


kongds:
The function lsp--point-to-position just go to point and calculate pos of this point.
But when go to end point after kill a line,  this point is the new pos to old end point.
For example one buffer has the three lines like this
```
this is the line will be killed
this is second line
this is third line
```
use kill-line at point 0
so the start pos is (0,0)
but the end pos will be (2,7) should be (0, length of the first line)

astahlman:
I've run into this problem, as well. Unfortunately, I think it's actually not possible for Emacs to correctly implement the protocol - at least not using text change hooks. I'm reposting some analysis from https://github.com/Microsoft/language-server-protocol/issues/9#issuecomment-331211372 inline below.

Consider the case of a single-line deletion. Here are the buffer contents at time 1.

1234
6789

And here are the buffer contents at time 2, after line 1 is deleted. (Note that a newline character is at position 5).

6789

From the Emacs documentation on Change Hooks:

Three arguments are passed to each function: the positions of
the beginning and end of the range of changed text,
and the length in chars of the pre-change text replaced by that range.
(For an insertion, the pre-change length is zero;
for a deletion, that length is the number of chars deleted,
and the post-change beginning and end are at the same place.)

So after the line deletion, our after-change-function is invoked with the following arguments:

    start = 1
    end = 1
    length = 5

The desired payload of the TextDocumentContentChangeEvent for this edit looks like this:

```
{
    'range': {
        'start': {'line': 0, 'character': 0},
        'end': {'line': 1, 'character': 0}
    },
  'rangeLength': 5,
  'text': ''
}
```

range.start can be calculated given the value of start and the current buffer contents. rangeLength is equal to the value of length. The problem is in calculating range.end; there is no way to reconstruct the pre-edit state of the buffer given the information available.

As proof, instead of

1234
6789

consider the following, alternate, prior state of the buffer:

1
34
6789

Deleting the first 2 lines would produce the same buffer state as before:

6789

And the arguments to our after-change-function would be the same:

    start: 1
    end: 1
    length: 5

But the correct payload for the TextDocumentContentChangeEvent corresponding to this change is:

```
{
    'range': {
        'start': {'line': 0, 'character': 0},
        'end': {'line': 2, 'character': 0}
    },
  'rangeLength': 5,
  'text': ''
}
```

(Note the difference in the `range.end` field - `line: 2` instead of `line: 1`.)

[...]

PR #120, opened 2017-10-02
Use buffer-undo-history to recover range for change events

astahlman:
This is a proof-of-concept of an idea suggested by @kongds in https://github.com/emacs-lsp/lsp-mode/issues/114. In order to correctly calculate the end position of the `Range` for a change event corresponding to the deletion or substitution of text, we examine the most recent change in the buffer's undo history to recover the affected region. 

I've tested this approach manually and haven't found any issues. I've also added unit tests, which may be useful regardless of whether this approach is ultimately adopted.

The downside of this approach is that it breaks if the user has `undo-tree-mode` enabled in the same buffer as `lsp-mode`. I think it would be possible to tailor this logic to work with `undo-tree-mode` by advising the `undo` and `redo` functions to track the current position in the undo tree. For now, I've just added a warning that LSP is incompatible with undo-tree.

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -69,5 +69,14 @@ If no such directory could be found, log a warning and return `default-directory
               "Couldn't find project root, using the current directory as the root.")
             default-directory)))))
 
+(defconst DEBUG 0)
+(defconst INFO 1)
+
+(defvar lsp--log-level INFO)
+
+(defmacro lsp--log (lvl fmt &rest args)
+  `(when (>= ,lvl lsp--log-level)
+     (message ,(format "[%s]: %s" lvl fmt) ,@args)))
+
 (provide 'lsp-common)
 ;;; lsp-common.el ends here
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -599,6 +599,73 @@ interface Range {
   "Get the value of capability CAP.  If CAPABILITIES is non-nil, use them instead."
   (gethash cap (or capabilities (lsp--server-capabilities))))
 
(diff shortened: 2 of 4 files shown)


vibhavp:
Thanks for writing this. I'll study this and submit a review in a few days.

vibhavp:
This wouldn't only break `buffer-undo-list`, but any elisp program that uses `buffer-undo-list` to achieve undo/redo functionality. IMO, we will end up sacrificing way too much for this.

astahlman:
It doesn't modify the `buffer-undo-list`, so `lsp-mode` wouldn't break `buffer-undo-tree`. It's the other way around, `buffer-undo-tree` modifies `buffer-undo-list`, which would break `lsp-mode`. I think we could fix `buffer-undo-tree`, but it would be annoying to debug if other packages are modifying `buffer-undo-list`.

I agree though that this probably introduces more complexity than it's worth, especially since @alanz now has a proposed implementation. I still think the best way to handle this would be server-side, but unfortunately I haven't seen any comments from the maintainers on https://github.com/Microsoft/language-server-protocol/issues/9

kongds:
`undo-tree-mode` use `undo-list-transfer-to-tree` change `buffer-undo-list` only before redo, undo, switch-branch, save the undo  tree history and others.
Every `undo` and `redo` changing will modify `buffer-undo-list` but the first of `buffer-undo-list` is the change what `undo` and `redo` caused.
So I think `undo-tree-mode` won't break `lsp-mode`.

PR #121, opened 2017-10-04
Delete unused function lsp--initialize

astahlman:
I was trying to understand the difference between `lsp--start` and `lsp--initialize` and noticed that the latter doesn't seem to be referenced anywhere, either in this repository or in any repository on Github that depend on lsp-mode.


vibhavp:
Wow, no idea how this never came up earlier. Thanks!

PR #122, opened 2017-10-05
Use symbol to get completion prefix bounds.

tigersoldier:
word fails to get the completion prefix with underscores. For example, if
try to complete at

    FOO_B
         ^

The prefix optained from the bound is B, not FOO_B. Using symbol fixes this
issue.


vibhavp:
LGTM, thanks.

PR #123, opened 2017-10-05
Provide trigger character hint to company-mode

tigersoldier:
If the server provides trigger characters and the point is after one of them,
set :company-prefix-length to t so that company-capf always shows completion
candidates regardless the actual length of the prefix (which is likely to be 0).

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -831,9 +831,19 @@ to a text document."
                                 (lsp--sort-string c2))
                           t))))
 
+(defun lsp--get-trigger-character-at-point ()
+  (when-let (completionProvider (lsp--capability "completionProvider"))
+    (seq-find (lambda (trigger-character)
+                (let ((len (length trigger-character)))
+                  (when (> (point) len)
+                    (string-equal (buffer-substring-no-properties (- (point) len) (point))
+                      trigger-character))))
+      (gethash "triggerCharacters" completionProvider))))
+
 (defun lsp--get-completions ()
   (lsp--send-changes lsp--cur-workspace)
-  (let ((bounds (bounds-of-thing-at-point 'word)))
+  (let ((completion-trigger-character (lsp--get-trigger-character-at-point))
+        (bounds (bounds-of-thing-at-point 'symbol)))
     (list
       (if bounds (car bounds) (point))
       (if bounds (cdr bounds) (point))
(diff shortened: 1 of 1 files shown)


bkchr:
Works for me :) 

@vibhavp anything against merging? :)

vibhavp:
No issues as such, but this ends up adding a soft dependency on company-mode, right?

bkchr:
You mean that it only works with company? Yeah you are probably right. Maybe the annotation could be made configurable

tigersoldier:
@vibhavp It does provide company-only feature, since trigger character doesn't make sense to completion-at-point.

@bkchr I'm on the fence whether to add a custom variable to toggle it. It doesn't take much time to get the trigger characters. IMO it's not a good experience to have a long list of customizable settings for every tiny details.

bkchr:
@tigersoldier I came across your https://github.com/tigersoldier/company-lsp project. I think, if that supports the trigger characters for company, it is okay to not merge the feature into lsp-mode directly.
While looking over your code, I saw that some functionality like:
`(advice-add 'lsp--parser-on-message :after 'company-lsp--on-message)`
Could be integrated into lsp-mode by providing a hook instead of hacking into the process.

sebastiencs:
> While looking over your code, I saw that some functionality like:
(advice-add 'lsp--parser-on-message :after 'company-lsp--on-message)
Could be integrated into lsp-mode by providing a hook instead of hacking into the process.

My fault, I've wrote that functionality. I didn't know the code of lsp-mode enough to make a PR to support async.
But it seems that there is now a PR for this purpose: https://github.com/emacs-lsp/lsp-mode/pull/138

bkchr:
@sebastiencs yeah no problem! :) Just thought loud about how the projects could be working together :)

tigersoldier:
@bkchr I agree both projects can work better together and would love to push features upstream. Just wanted to get the feature landed quickly so I can use it :)

Back to the topic of this PR. My original plan was to implement all company-related feature here and just deprecate company-lsp in favor of `company-capf`. Now company-lsp is implementing some features that are hard to port to completion-at-point-function (e.g async completion, post-completion actions), the original goal is no longer feasible.

Let's close this PR.

vibhavp:
@tigersoldier I've added support for async requests in b56a06e5ca4cf01b126521183d2820745b06f63f. If that helps, `company-lsp` should use it instead of advising internal functions.

tigersoldier:
Sure, I'll update it.

PR #124, opened 2017-10-06
[Discussion] Fix incremental

alanz:
This uses the `before-change` hook to store the original start and end positions, and then takes the end position from that if a delete change must be generated.

The emacs documentation warns that the before and after events may not bracket, and may not be the same.

We do a sanity check on them before using, and if there is a mismatch send a full sync instead.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -89,7 +89,7 @@ for a new workspace."
   :group 'lsp-mode)
 
 ;;;###autoload
-(defcustom lsp-document-sync-method 'full
+(defcustom lsp-document-sync-method nil
   "How to sync the document with the language server."
   :type '(choice (const :tag "Documents should not be synced at all." 'none)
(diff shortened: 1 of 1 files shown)


alanz:
Closes #114

vibhavp:
How would we do a full sync? I don't think we can change the sync method once it has been set during initialization.

alanz:
When I say full sync I mean this: https://github.com/emacs-lsp/lsp-mode/blob/fix-incremental/lsp-methods.el#L627

Send an incremental change message, but with the entire contents of the buffer.  Hopefully this will only be needed rarely, and makes sure the server state matches.

vibhavp:
Ah, I forgot that `range` and `rangeLength` can be omitted. Is there anything left to be done? Otherwise, it looks good.

alanz:
I have been using it myself, and it seems to work fine. So I am happy to go with it.

vibhavp:
Squashed commits and pushed manually, thanks!

PR #127, opened 2017-10-12
Deal with falsy values in completion response properly

### LINKED ISSUE #119: completion-at-point not working in prototype php-language-server integration
As an Emacser who's wished for good PHP integration many times in the past decade, and who just discovered LSP and lsp-mode, I've tried hooking [php-language-server](https://github.com/felixfbecker/php-language-server) up to lsp-mode [in my PHP config](https://github.com/NateEag/.emacs.d/blob/8261d90ca053af74632c57983620df0ed66fe7ce/site-lisp/mode-configs/php-mode-init.el#L96).

Jump to definition works nicely, so I know the basics are working, but `completion-at-point` always fails, either returning "[No matches]" (in cases where that's true) or failing with the error "Wrong type argument: stringp, nil".

I've written one or two elisp packages, so I'm not afraid of getting my hands dirty, but the few times I've glanced at this so far, I haven't figured out a clear angle of attack on the problem (largely because I don't really know either codebase, and am new to LSP in general).

Any advice on how I might approach debugging this?

I realize this isn't a great bug report - I can assemble a minimal reproduction if someone else wants to take a look at it.

NateEag:
Fixes #119.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -806,8 +806,24 @@ to a text document."
      (17 . "File")
      (18 . "Reference")))
 
+(defun lsp--gethash (key table &optional dflt)
+  "Look up KEY in TABLE and return its associated value,
+unless KEY not found or its value is falsy, when it returns DFLT.
+DFLT defaults to nil.
+
+Needed to implement completion request fallback behavior for the
+fields 'sortText', 'filterText', and 'insertText' as described
+here:
+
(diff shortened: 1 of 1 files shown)


vibhavp:
Thanks.

NateEag:
You're wecome. Thanks for merging!

PR #128, opened 2017-10-13
Make sure we send the whole buffer for a lsp--full-change-event

alanz:
By removing any narrowing in effect.


vibhavp:
LGTM, thanks.

PR #131, opened 2017-10-18
Remove major-mode<->client association

vibhavp:
This PR turns all `lsp-define-{tcp,stdio}-client` functions into macros, which create a new function `<name>-enable`. Defining a client no longer needs an associated major mode symbol, as clients are now enabled by their respective functions instead, which removes the dependency on the major mode's package for an LSP client. For instance, here's how a client for python would be defined:
```lisp
(lsp-define-stdio-client lsp-python "python"
			 (lsp-make-traverser #'(lambda (dir)
						 (directory-files
						  dir
						  nil
						  "\\(__init__\\|setup\\)\\.py")))
			 '("pyls"))
```

This defines the function `lsp-python-enable`, which is hooked to `python-mode` with `(add-hook 'python-mode #'lsp-python-enable)`. Also, the package's version has been bumped to `3.0` to accommodate for the
 API change. 

TODO:
- [ ] Document changes to API in `API.org`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -101,10 +101,13 @@ Optional arguments:
          (unless lsp-mode
            ,(when (plist-get args :initialize)
               `(funcall ,(plist-get args :initialize) client))
-           (lsp-mode 1)
-           (lsp--start client))))))
+           (if (lsp--should-start-p (funcall (lsp--client-get-root client)))
+             (progn
+               (lsp-mode 1)
+               (lsp--start client))
+             (message "Not initializing project %s" root)))))))
 
-(defun lsp-define-tcp-client (mode language-id get-root command host port &rest args)
+(defmacro lsp-define-tcp-client (mode language-id get-root command host port &rest args)
   "Define a LSP client using TCP.
 NAME is the symbol to use for the name of the client.
 MODE is the major mode for which this client will be invoked.
@@ -131,8 +134,11 @@ Optional arguments:
          (unless lsp-mode
            ,(when (plist-get args :initialize)
               `(funcall ,(plist-get args :initialize) client))
-           (lsp-mode 1)
-           (lsp--start client))))))
+           (if (lsp--should-start-p (funcall (lsp--client-get-root client)))
+             (progn
+               (lsp-mode 1)
+               (lsp--start client))
+             (message "Not initializing project %s" root)))))))
 
 ;;;###autoload
 (define-minor-mode lsp-mode ""
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -32,8 +32,8 @@
 
 (defun lsp--make-stdio-connection (name command command-fn)
   (lambda (filter sentinel)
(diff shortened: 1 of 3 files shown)


alanz:
Basics look good in a quick look. I will give it a go for haskell and comment further. Likely to only be tomorrow though.

alanz:
It seems to work with my simple change for haskell mode.

What about selective enable-disable based on a whitelist / blacklist?

vibhavp:
Whitelist and blacklists are used in `lsp-should-start-p`, which is used in `lsp--start`

alanz:
Great, I did not realise that had come in.

I am happy for a merge then.

lsp-mode.el
```diff
@@ -74,64 +74,77 @@
         "lsp-define-client: :ignore-regexps element %s is not a string"
         e))))
 
-(defun lsp-define-stdio-client (mode language-id type get-root name command &rest args)
-  "Define a LSP client using stdio.
+(defmacro lsp-define-stdio-client (name language-id get-root command &rest args)
+   "Define a LSP client using stdio.
+NAME is the symbol to use for the name of the client.
 MODE is the major mode for which this client will be invoked.
```
phst: You should remove this line as MODE doesn't exist any more.

lsp-mode.el
```diff
@@ -74,64 +74,77 @@
...
-                                                     args
-                                                     :ignore-regexps))))
-      (puthash mode client lsp--defined-clients))))
-
-(defun lsp-define-tcp-client (mode language-id type get-root name command host port &rest args)
+`:command-fn' is a function that returns the command string/list to be used to launch the language server. If non-nil, COMMAND is ignored.
+`:initialize' is a function called when the client is intiailized. It takes a single argument, the newly created client.
+"
+  (let ((enable (intern (format "%s-enable" name)))
+         (disable (intern (format "%s-disable" name))))
```
phst: `disable` is unused and can be removed.

lsp-mode.el
```diff
@@ -74,64 +74,77 @@
...
+                       :get-root ,get-root
+                       :ignore-regexps ,(plist-get args :ignore-regexps))))
+         (unless lsp-mode
+           ,(when (plist-get args :initialize)
+              `(funcall ,(plist-get args :initialize) client))
+           (if (lsp--should-start-p (funcall (lsp--client-get-root client)))
+             (progn
+               (lsp-mode 1)
+               (lsp--start client))
+             (message "Not initializing project %s" root)))))))
```
phst: `root` isn't defined here.

lsp-mode.el
```diff
@@ -74,64 +74,77 @@
...
+
+(defmacro lsp-define-tcp-client (mode language-id get-root command host port &rest args)
```
phst: Same comments as above apply to this function.

phst: You should rename `mode` to `name`.

PR #133, opened 2017-10-19
Avoid deleting overlays created by Flycheck (fix #87)

### LINKED ISSUE #87: `lsp-symbol-highlight` should retrigger flycheck if necessary
`lsp-symbol-highlight` calls `lsp--remove-cur-overlays` which may remove in-buffer highlighting done by flycheck. To counter that, `lsp-symbol-highlight` should call flycheck if it's enabled in the buffer after adding its own highlighting.

fmdkdd:
Previously we saved only the positions of the overlays created for highlighting
symbols, and called `remove-overlays` with this position.  This had the
side-effect of deleting *any* overlay within this range, not just the ones
created by lsp-mode for highlighting.

This commit saves the overlays created when highlighting directly in a list, and
deletes them using `delete-overlay`.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -51,7 +51,7 @@
   (proc nil) ;; the process we communicate with
   (cmd-proc nil) ;; the process we launch initially
   (buffers nil) ;; a list of buffers associated with this workspace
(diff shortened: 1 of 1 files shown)


vibhavp:
Looks good, thanks.

PR #134, opened 2017-10-19
Allow the mode-line to change dynamically

fmdkdd:
This allows us to give some feedback to the user when the language server is
busy.

E.g., the RLS sends a `rustDocument/beginBuild` message when it starts building
the workspace, which can take quite some time on larger projects.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -18,7 +18,7 @@
 ;; Author: Vibhav Pant <vibhavp@gmail.com>
 ;; URL: https://github.com/emacs-lsp/lsp-mode
 ;; Package-Requires: ((emacs "25.1") (flycheck "30"))
-;; Version: 3.3
+;; Version: 3.4
 
 ;;; Commentary:
 
@@ -185,11 +185,19 @@ Optional arguments:
                  (lsp--start client ,extra-init-params))
                (message "Not initializing project %s" root))))))))
 
+(defvar-local lsp-status nil
+  "The current status of the LSP server.")
+
+(defun lsp-mode-line ()
+  "Construct the mode line text."
+  (concat " LSP" lsp-status))
+
 ;;;###autoload
 (define-minor-mode lsp-mode ""
   nil nil nil
-  :lighter " LSP"
-  :group 'lsp-mode)
+  :lighter (:eval (lsp-mode-line))
+  :group 'lsp-mode
+  (lsp--start))
 
 (defconst lsp--sync-type
   `((0 . "None")


sebastiencs:
I think it would be better to set the variable `mode-line-process` rather than changing the minor mode name.
`mode-line-process` is often displayed in a separate place than all the minors mode names and so it would be easier to see the lsp status in the mode-line.
And it's probably the purpose of `mode-line-process`

fmdkdd:
Good point, I'll look into it and update the PR if it works out.

fmdkdd:
So there are two issues with `mode-line-process`.  The first is that we don't control its format in the mode-line.  By default, it appears after the major mode, but user who customize the mode line format could have put it anywhere (and even disabled it!).

Here is what it looks like on a vanilla Emacs:
```
(Rust (building) LSP FlyC:1/8)
```

And here is by customizing the minor mode lighter:
```
(Rust LSP(building) FlyC:1/8)
```

The second issue is that we don't actually know if the major mode does not set `mode-line-process` at some point, erasing our value.  I guess `mode-line-process` is really useful for signaling the status of the process managed by the *major mode* (like `M-x shell`), and here we have a minor mode.

FWIW, I've went with customizing the minor mode light because that's how Flycheck does it.

sebastiencs:
I'm not sure to understand your point with the first issue.
I would prefer to see the lsp status in another place than with minor modes lighters, that's why I've commented. IMHO this issue is more like an advantage, for me
`mode-line-process` could be disabled, like the lsp lighter (lot of people are using diminish for example)

I agree that the variable being overwritten by another mode is an issue.
The difference with flycheck is that the lsp status are informations that should be displayed for a limited time (a few seconds ?).

So yes it's probably better to not use `mode-line-process` and let the major mode use it

fmdkdd:
> The difference with flycheck is that the lsp status are informations that should be displayed for a limited time (a few seconds ?).

You are right, the process information is temporary, at least in the case of the RLS.  Although we could also report whether we are connected to a language server, and in this case the information would be here as long as the minor mode is active.  (But maybe we should always be connected to a language server?)

> (lot of people are using diminish for example)

That's a good point.  If we put that information in the lighter, then you cannot diminish the lighter anymore without losing that info.

So I agree that `mode-line-process` has advantages in customization.  Ultimately, it feels to me that `mode-line-process` should be left to the major mode, but I don't have a strong opinion either way.

sebastiencs:
Can it be merged ? Let the major mode handle `mode-line-process`

fmdkdd:
> Can it be merged ?

It could, but in the meantime the notifications from the RLS have been handled with messages:

https://github.com/emacs-lsp/lsp-rust/blob/a281f3daa4953be8f1ae2c35d3fb9059ae58ef0c/lsp-rust.el#L61-L70

So do we want to use the modeline for that after all?  cc @vibhavp

sebastiencs:
Handle the server status with messages isn't a good solution IMO.
The message get delete when the cursor is moved, with eldoc, from other messages...

The modeline seems a better place for that information.
And with the method you provide (the variable `lsp-status`), it's easy to customize where and how to show the information (with [spinner](https://github.com/Malabarba/spinner.el) for example)

vibhavp:
@fmdkdd Sure, that was meant to be a temporary solution. Showing build status on the modeline would be more appropriate, IMO.

fmdkdd:
Ok, I've rebased and this should be good.  I'll update emacs-lsp/lsp-rust#11 to use this new variable.

vibhavp:
As this introduces a new API feature, could you increment the version string? Thanks.

fmdkdd:
@vibhavp Done.

vibhavp:
@fmdkdd The version is 3.4, 3.3 was when string rendering was added.

vibhavp:
Thanks!

Issue #132, opened 2017-10-19
"No Content-Length header" error


MageSlayer:
Hi

Looks like RLS (Rust language server) does something unexpected for lsp-mode.
Emacs emits following error:
```
error in process filter: Error parsing language server output: (error No Content-Length header)
```
... can something be done about it? 
Or maybe RLS devs must be informed?

vibhavp:
What action were you performing when you encountered this?

MageSlayer:
Just ordinary edits in buffer.
It seems it appears when RLS re-parses sources.

bkchr:
Yeah, I already observed this error. I think when rls sends big error messages, there are `\n\n` in these messages. The parsing algorithm in lsp-mode does think that after each `\n\n` the content starts. So, the algorithm gets confused while parsing.

sebastiencs:
I also have this error quite often with rls

vibhavp:
RLS shouldn't be sending error messages on stdout in the first place. Could you paste the output of the *stderr* buffer?

MageSlayer:
Here it is. It happened while I was editing macro in Rust source file.
```
{"message":"method is never used: `is_cardinal`","code":null,"level":"warning","spans":[{"file_name":"ast.rs","byte_start":29821,"byte_end":298\
53,"line_start":825,"line_end":825,"column_start":9,"column_end":41,"is_primary":true,"text":[{"text":"        fn is_cardinal($($arg)*) -> bool\
 {","highlight_start":9,"highlight_end":41}],"label":null,"suggested_replacement":null,"expansion":{"span":{"file_name":"ast.rs","byte_start":3\
4798,"byte_end":34827,"line_start":954,"line_end":954,"column_start":5,"column_end":34,"is_primary":false,"text":[{"text":"    is_pred!([&self]\
, self.tp());","highlight_start":5,"highlight_end":34}],"label":null,"suggested_replacement":null,"expansion":null},"macro_decl_name":"is_pred!\
","def_site_span":{"file_name":"ast.rs","byte_start":28748,"byte_end":30316,"line_start":799,"line_end":840,"column_start":1,"column_end":2,"is\
_primary":false,"text":[{"text":"macro_rules! is_pred {","highlight_start":1,"highlight_end":23},{"text":"    ([$($arg:tt)*], $tp:expr) => {","\
[...]

PR #135, opened 2017-10-20
Don't reset parser in lsp--parser-on-message

bkchr:
We don't need to reset the parser in lsp--parser-on-message, because the parser
is already reset when a full message was read. Resetting the parser in the
lsp--parser-on-message function leads to errors when the parser is currently
parsing a message, but the chunk did not contain all the required data.

Fixes: https://github.com/emacs-lsp/lsp-mode/issues/132


vibhavp:
Looks good, thanks.

PR #138, opened 2017-10-24
[WIP] Run document highlights request / response async

alanz:
This is a proof of concept of providing a response handler when sending the message, so that the entire UI does not hang while waiting for a response.

@vibhavp please let me know if you believe this approach is feasible.

I am concerned about making sure the callback function runs against the correct buffer.


vibhavp:
I had code for async requests stashed away for a while now. I'll take a look a look at this functionality first.

alanz:
Closed in favour of https://github.com/emacs-lsp/lsp-mode/pull/155

PR #141, opened 2017-10-25
[WIP] Whitelist features

alanz:
Split the lsp-XXXX-enable into non-interactive and interactive

The non-interactive one gets called in the mode hook, and is quiet

The interactive one will ask to add a project to the whitelist if invoked to
start LSP for it and it would not otherwise start.

For some reason I can't get both macros to be defined at once, so currently you
need to call something like

    (lsp-define-stdio-client-noninteractive
                     lsp-haskell "haskell"
                     #'lsp-haskell--get-root
                     '("hie" "--lsp" "-d" "-l" "/tmp/hie.log") )
    (lsp-define-interactive-client lsp-haskell)

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -348,12 +348,18 @@ disappearing, unset all the variables related to it."
               exit-str)
             (lsp--uninitialize-workspace)))))))
 
+(defvar lsp--project-whitelist-temp nil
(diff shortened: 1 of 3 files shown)


alanz:
Superseded by https://github.com/emacs-lsp/lsp-mode/pull/156

PR #146, opened 2017-10-27
Add support for request "completionItem/resolve"

sham1:
This request was originally to be added with #94 however I screwed up my local branch so I decided to scrap that one and create this new one (the issue was that the PR wouldn't update after my latest change...)

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -940,7 +940,17 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (comp)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (if (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (progn
+      (when (null comp)
+	(error "Completion item must not be nil"))
+      (lsp--send-request (lsp--make-request
+			  "completionItem/resolve"
+			  comp)))
+    comp))
 
 (defun lsp--location-to-xref (location)
   "Convert Location object LOCATION to an xref-item.
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -943,14 +943,12 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
 (defun lsp--resolve-completion (comp)
   (lsp--cur-workspace-check)
   (lsp--send-changes lsp--cur-workspace)
-  (if (gethash "resolveProvider" (lsp--capability "completionProvider"))
-    (progn
-      (when (null comp)
-	(error "Completion item must not be nil"))
-      (lsp--send-request (lsp--make-request
-			  "completionItem/resolve"
-			  comp)))
-    comp))
+  (unless (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (error "ResolveProvider not found"))
+  (when comp
+    (lsp--send-request (lsp--make-request
+			"completionItem/resolve"
+			comp))))
 
 (defun lsp--location-to-xref (location)
   "Convert Location object LOCATION to an xref-item.
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -940,15 +940,16 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-(defun lsp--resolve-completion (comp)
+(defun lsp--resolve-completion (item)
   (lsp--cur-workspace-check)
   (lsp--send-changes lsp--cur-workspace)
-  (unless (gethash "resolveProvider" (lsp--capability "completionProvider"))
-    (error "ResolveProvider not found"))
-  (when comp
-    (lsp--send-request (lsp--make-request
-			"completionItem/resolve"
-			comp))))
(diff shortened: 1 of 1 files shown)


sham1:
Well this is an oopsie, I didn't at all remember to make this a remote branch.

lsp-methods.el
```diff
@@ -940,7 +940,17 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
...
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (comp)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (if (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (progn
+      (when (null comp)
+	(error "Completion item must not be nil"))
```
tigersoldier: nit: indent?

lsp-methods.el
```diff
@@ -940,7 +940,17 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (comp)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (if (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (progn
+      (when (null comp)
```
tigersoldier: I would make it fail if `resolveProvider` is nil. Reasons:

 * It's more consistent between the 2 cases
 * It's an error the caller should take care of. Early throwing is better than silently swallowing it.

lsp-methods.el
sham1: Yeah, I'll fix that indentation along with that other point.

lsp-methods.el
```diff
@@ -940,7 +940,15 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (comp)
```
tigersoldier: rename comp to item to be consistent

lsp-methods.el
```diff
@@ -940,7 +940,15 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
...
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (unless (gethash "resolveProvider" (lsp--capability "completionProvider"))
+    (error "ResolveProvider not found"))
```
tigersoldier: hmmm this is not what i meant. The previous version is ok. Just need to move the nullness check of `comp` out of the `resolveProvider` check. So when `comp` is nil, it fails no matter `resolveProvider` is nil or not.

vibhavp: It seems unnecessary to error out when the server does not provide `resolve` support. Returning `nil` would suffice.

tigersoldier: This is already changed in 27ba5409dc86949d9d0ce751ab15f26a6c7f5ae0.

I don't think retuning nil is a good idea. This makes callers worry about the capability. Always returning a completion item is a more consistent behavior.

vibhavp: Indeed, returning the item is much better.

lsp-methods.el
```diff
@@ -940,7 +940,16 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
       :annotation-function #'lsp--annotate
       :display-sort-function #'lsp--sort-completions)))
 
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (item)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (unless item
```
vibhavp: A cleaner way of doing this would be `(cl-assert item nil "completion item is nil")` (see [`lsp--send-request-async`](https://github.com/emacs-lsp/lsp-mode/blob/364dd567d3bd3246d541314de03ec4b6b9f051e0/lsp-methods.el#L240))

vibhavp:
@tigersoldier any more changes you want?

lsp-methods.el
```diff
@@ -940,7 +940,15 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
...
-;;; TODO: implement completionItem/resolve
+(defun lsp--resolve-completion (item)
+  (lsp--cur-workspace-check)
+  (lsp--send-changes lsp--cur-workspace)
+  (cl-assert item nil "Completion item must not be nil")
+  (if (gethash "resolveProvider" (lsp--capability "completionProvider"))
+      (lsp--send-request (lsp--make-request
+			  "completionItem/resolve"
+			  item))
+    item)
```
kat-co: I think you need another paren here to close out the `defun` (CI failed because of this).

tigersoldier:
@vibhavp Sorry forgot to submit my pending review comments. Other than the missing paren issue @kat-co pointed out, the PR looks good to me.

lsp-methods.el
sham1: Ah... sorry 'bout that. Sometimes you just forget those parenthesis :P

kat-co: I understand! I highly recommend a mode like `paredit` for editing Lisps.

sham1: I usually use smartparens, but it just didn't happen for whatever reason that time.

vibhavp:
Looks good, thanks.

PR #148, opened 2017-10-27
Document the Version of the LSP Supported

### LINKED ISSUE #144: Consider Documenting Which Version of the LSP lsp-mode Targets
It appears `lsp-mode` might be targeting the evolving [3.0 spec](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md).

We might also want to consider adopting [semantic versioning](http://semver.org/) and making statements about what changing the major.minor.patch portions mean in relation to this and to `lsp-mode` itself.

kat-co:
This adds two pieces of documentation:

1. A programatic value for code to match against
2. A small update to the README for humans

fixes #144


vibhavp:
Looks good, thanks.

PR #155, opened 2017-11-02
Use async for codeAction and documentHighlight

alanz:
And add the jsonrpc header to lsp--make-response


vibhavp:
Looks good, we might need to later adjust for out of order highlights later.

Issue #157, opened 2017-11-03
Allow other packages to register new client capabilities


tigersoldier:
As of tigersoldier/company-lsp#5, company-lsp is planning to support expanding completion snippets. A proper way to support it is to set `TextDocumentClientCapabilities.completion.completionItem.snippetSupport` to be true on initialization, or allow company-lsp to dynamically register it.

However, we shouldn't just set snippetSupport to true unconditionally. lsp-mode itself doesn't support post-completion actions, so cannot support snippet expansion. It should only be turn on when company-lsp is in use.

That said, lsp-mode needs to provide an API to allow other packages to register and unregister client capabilities that is not supported by the core. Ideally it should extend the initial client capability if possible, because not all language servers support dynamic registration.

bkchr:
@vibhavp This ticket can be closed :)

vibhavp:
yup, thanks.

PR #161, opened 2017-11-06
RFC: #157 Allow other packages to register client capabilities.

tigersoldier:
Adds two functions for registering/unregistering client capabilities:
`lsp-register-client-capabilities` and `lsp-unregister-client-capabilities`.

Registered capabilities are stored in the `lsp--extra-client-capabilities`
alist in the format of `(PACKAGE-NAME . CAPS)` cons cell. CAPS can be either
a plist, nil, or a function.

All registerd client capabilities are merged into the default client
capabilities. The `lsp--merge-two-plists` function implements a deep merging
algorithm for plists. Existing leaf capability entry cannot be overwritten.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -54,7 +54,13 @@
   (cmd-proc nil) ;; the process we launch initially
   (buffers nil) ;; a list of buffers associated with this workspace
   (highlight-overlays nil) ;; a list of overlays used for highlighting the symbol under point
-  )
+
+  ;; Extra client capabilities provided by third-party packages using
+  ;; `lsp-register-client-capabilities'. It's value is an alist of (PACKAGE-NAME
+  ;; . CAPS), where PACKAGE-NAME is a symbol of the third-party package name,
+  ;; and CAPS is either a plist of the client capabilities, or a function that
+  ;; takes no argument and returns a plist of the client capabilities or nil.")
+  (extra-client-capabilities nil))
 
 (defvar-local lsp--cur-workspace nil)
 (defvar lsp--workspaces (make-hash-table :test #'equal)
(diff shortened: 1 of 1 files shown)


vibhavp:
How would this be used by `company-lsp`? AFAICT, `lsp-register-client-capabilities` would have to be called before the document/workspace is initialized, right?

tigersoldier:
My current solution is to call it when `company-lsp` is required. See tigersoldier/company-lsp@be7fa992a0d3f2d42f5081a144164693d0e93e21

The idea of accepting a function is to allow packages to detect whether they are used an not, and return nil if they are not used. I haven't tested if the order of major mode hooks can affect the results. Ideally `initialize` request is delayed until all the major mode hooks are run so that all the variables are ready.

bkchr:
Couldn't we introduce a hook that is called, before a workspace is initialized? This hook then could be used to register further client capabilities.

vibhavp:
@bkchr Added in ed0c4523665c44d8a836d854dbdd60d9ac19b416

@tigersoldier Instead of associating with package names, why not use a documentation string instead?

bkchr:
@vibhavp ty for the fast implementation :)

What do you mean by documentation string? Just an arbitrary string like "company-lsp-caps"?

tigersoldier:
@vibhavp @bkchr before initialization hook is great. Let me update the implementation to limit the registration scope in workspace.

Why strings are more preferred than symbols?

tigersoldier:
Update the PR to register client capabilities into workspaces and run the pre initialize hook before getting client capabilities.

lsp-methods.el
```diff
@@ -344,6 +409,48 @@ disappearing, unset all the variables related to it."
   "Client Text document capabilities according to LSP."
   `(:synchronization (:didSave t)))
 
+;;;###autoload
```
bkchr: If this function is called in the workspace hook, the autoload can be removed? As lsp-mode should already be loaded, when the hook is executed.

bkchr:
Looks good so far, maybe next time use `git rebase` to get the changes from upstream master. But that is only a personal preference :)

tigersoldier:
@bkchr removed autoload

I use git rebase on my own projects, but I am not familiar with the pull request system of github so not sure what the best practice is. Rebased the PR and squashed the commits.

bkchr:
If you rebase and force push, github automatically filters out all commits that are already in the branch you are want to merge against :) 
But squashing is also nice. :)

@vibhavp any further concerns?

vibhavp:
Yeah, looks good. Thanks!

PR #172, opened 2017-11-16
Using if and let instead of if-let*

rattit0:
`if-let*` is only supported since Emacs 26.1


sebastiencs:
`if-let` is depracted since emacs 26.1 so it would be better to use `let` and `if` separately

vibhavp:
`if-let` is deprecated, so I switched to plain ol' `setq`.

PR #173, opened 2017-11-16
whitelist features

alanz:
If a whitelist exists, and the current project is not on it, lsp-mode cannot be started for it, without this patch.

An alternative would be some simple way to add the current project root to the whitelist, which would require calling the relevant function to determine the root.

Supersedes https://github.com/emacs-lsp/lsp-mode/pull/156

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -481,12 +481,18 @@ registered client capabilities by calling
                      exit-str)
             (lsp--uninitialize-workspace)))))))
 
+(defvar lsp--project-whitelist-temp nil
+  "If a project is started manually and not added to the
+  whitelist, the root is stored here so other files belonging to
+  it will also be added to the session.")
+
 (defun lsp--should-start-p (root)
   "Consult `lsp-project-blacklist' and `lsp-project-whitelist' to
 determine if a server should be started for the given ROOT
 directory."
   (if lsp-project-whitelist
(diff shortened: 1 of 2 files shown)


bkchr:
Do you know that you don't need to open a new pull request for each rebase?

If I understand correctly, with this feature I need to add every project I want to use lsp-mode for, put onto this whitelist? That does sound complicated...

alanz:
There is already support for a whitelist and a blacklist.

I have some projects that I do not want to use LSP on (such as GHC).   And when testing, I want to be able to enable/disable it.

So it is more about managing the whitelist than anything else.

What this patch does, is make it so that when you open a file in a project not on the whitelist, you need to do `lsp-XXXX-enable` to start up a LSP server, and it offers to add it to the whitelist for future automatic starts.

bkchr:
Why not enable by default and disable if wanted by the user?

alanz:
Thats the thing. enabling/disabling is not currently provided, and is
complicated because you need the project root directory as returned by the
relevant function.

And then need to customise.

It should be a simple process, which can be done via M-x

bkchr:
I would prefer that you introduce a global variable that determines if the user should be asked or not.

alanz:
Ok, that should be simple enough.

alanz:
Superseded by https://github.com/emacs-lsp/lsp-mode/pull/190

Issue #175, opened 2017-11-16
Don't prompt about killing server processes


phst:
Killing LSP processes shouldn't cause data loss, so prompting the user when quitting them isn't necessary. I think this can be prevented by adding `:noquery t` at https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L49.

bkchr:
Duplicate of: https://github.com/emacs-lsp/lsp-mode/issues/174

phst:
This is not a duplicate. This bug is about Emacs prompting when killing a process. #174 is about `lsp-mode` prompting when closing a workspace. The issues are somewhat related, but these are two different prompts.

bkchr:
Ahh, sorry! I also tried to solve this problem by using [set-process-query-on-exit-flag](https://www.gnu.org/software/emacs/manual/html_node/elisp/Query-Before-Exit.html), but I could not get this working. Do you have any ideas?

phst:
That should work. I think adding `:noquery t` to `make-process` would be a bit cleaner, but in the end it's the same thing.
Your PR would still make sense to fix the `lsp-mode` prompt, though maybe some users want the prompt (thus I suggested adding a user option).

bkchr:
I also tried `:noquery t` and that did not change the result...
I could close my pr and you open a new one with user prompt and maybe a fix for this issue?

phst:
You PR is fine. I'd let the maintainer decide whether the behavior should be controllable by a user option.

Adding `:noquery t` does not prevent the prompt issued by `lsp-mode`. It prevents different prompts when exiting Emacs while there are still running LSP processes that haven't been closed otherwise.

bkchr:
Yeah I know what you wanted to achieve with `:noquery t`and I tested that, but for me it did not work. I will maybe give it another try!

vibhavp:
Indeed, the user doesn't interact with the server directly, asking the user isn't of any use.

phst:
As explained above, we still should have `:noquery t` so that users aren't prompted about lingering processes when quitting Emacs.

Issue #174, opened 2017-11-16
lsp--shut-down-p: add user option to suppress prompt


phst:
Some users don't like to be prompted when closing a workspace. Therefore I'd propose to add a user option to customize the behavior of `lsp--shut-down-p`, like:

```elisp
(defcustom lsp-close-server-on-kill-buffer :prompt
  :type '(choice (const :never) (const :always) (const :prompt))

(defun lsp--shut-down-p ()
  (cl-ecase sp-close-server-on-kill-buffer
    (:never nil)
     (:always t)
    (:prompt (y-or-n-p "...")))
```

bkchr:
Why did you create two issues for your problem?

phst:
Fixed by https://github.com/emacs-lsp/lsp-mode/pull/176

PR #176, opened 2017-11-17
Don't ask when closing the language server

bkchr:
Fixes: https://github.com/emacs-lsp/lsp-mode/issues/175


phst:
Note, this fixes #174, not #175.

PR #177, opened 2017-11-17
Fixes bug with default value of `language-fn-id`

bkchr:
cl-defmacro automatically evaluates default values. To prevent evaluation, the
default value needs to be escaped, but when escaped `language-id` is not
replaced by the macro system. To circumvent these problems, the value of
`language-fn-id` is checked at execution and if the value is `nil` a lambda
returning the `language-id` is used.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -79,7 +79,7 @@
 
 (cl-defmacro lsp-define-stdio-client (name language-id get-root command
                                        &key docstring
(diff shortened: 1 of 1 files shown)


bkchr:
@vibhavp do you want to merge your fix or my? :)

vibhavp:
Yeah, your implementation is more succinct. Thanks.

PR #178, opened 2017-11-17
don’t prompt when killing LSP processes

### LINKED ISSUE #175: Don't prompt about killing server processes

phst:
Fixes #175

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -46,7 +46,8 @@
        :command final-command
        :filter filter
        :sentinel sentinel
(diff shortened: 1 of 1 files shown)


bkchr:
Hmm, my emacs still asks me when I want to close him^^

phst:
This might be due to Emacs bug https://debbugs.gnu.org/cgi/bugreport.cgi?bug=30031

PR #179, opened 2017-11-19
Scan unparsed leftovers when first entering lsp--parser-read

jaelsasser:
The existing `(lsp--parser-read)` logic doesn't properly handle the corner-case where the headers are split across multiple function calls. Was noticing it a lot with [cquery](https://github.com/jacobdufault/cquery)'s startup splat.

(Don't know if it matters right now, but I've filed the necessary papers for Emacs copyright assignment.)


vibhavp:
Looks good, thanks.

PR #181, opened 2017-11-20
Support initializeOptions when starting an LSP server

### LINKED ISSUE #180: Support for optional per-workspace initializeOptions
The [Initialize Request InitializeParams](https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#initialize) support an `initializeOptions` blob. [cquery](https://github.com/jacobdufault/cquery) uses these to set both a per-project cacheDirectory and a global resourceDirectory.

I'm going to take a pass at this and (probably) extend `lsp-define-*-client` with a parameter that either takes a plist or a function to generate the `initializeOptions` plist from the workspace.

jaelsasser:
LSP v3.0 spec defines an optional member of initialize:InitializeParams,
initializeOptions, that can be used to pass user-provided initialization
options to the server.

Implements an (optional) extra arg to lsp--start, extra-init-params.
This can be either a fixed plist of options or a function that takes
an LSP workspace and generates the initializeOptions params.

Existing clients should not be affected. Fixes #180

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -489,7 +489,7 @@ directory."
       (member root lsp-project-whitelist)
     (not (member root lsp-project-blacklist))))
 
-(defun lsp--start (client)
+(defun lsp--start (client &optional extra-init-params)
   (when lsp--cur-workspace
     (user-error "LSP mode is already enabled for this buffer"))
   (let* ((root (funcall (lsp--client-get-root client)))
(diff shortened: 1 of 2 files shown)


jaelsasser:
(If there's any pieces of the API I've added that you'd refined / renamed / changed, let me know and I can make the adjustments.)

lsp-methods.el
```diff
@@ -521,10 +521,21 @@ directory."
...
-                           :rootUri ,(concat "file://" root)
-                           :capabilities ,(lsp--client-capabilities)))
-      (setf response (lsp--send-request (lsp--make-request "initialize" init-params)))
+      (setq init-params
+            (lsp--merge-plists
+             `(:processId ,(emacs-pid)
+                :rootPath ,root
+                :rootUri ,(concat "file://" root)
+                :capabilities ,(lsp--client-capabilities))
+             (if extra-init-params
```
vibhavp: What about this:
```emacs-lisp
`(:processId ,(emacs-pid)
 :rootpath ,root
 :rootUri ,(concat "file://" root)
 :capabilities ,(lsp--client-capabilities)
 :initializationOptions ,(if (functionp extra-init-params) (funcall extra-init-params lsp--cur-workspace) extra-init-params))
```
If `extra-init-params` is nil, `json-encode` will marshal it to `null`.

jaelsasser: Yeah, this is a lot cleaner. I'll fixup the commit and push.

(I wanted to avoid even including the `initializationOptions` JSON blob if it wasn't specified in the client, but I guess if it's in the protocol clients should at least be able to ignore the unexpected field.)

jaelsasser:
(Went through and set up lsp-python to ensure that it wouldn't choke on a nil initializeOptions. Seems to be fine.)

vibhavp:
Looks good, thanks.

PR #183, opened 2017-11-20
Use forward-chars instead of move-to-column in lsp--position-to-point

jaelsasser:
On Emacs 26.0.90, I was getting weird highlights on files with `indent-tabs-mode` set. Root cause seemed to be in `lsp--position-to-point`.

Before:
screen shot 2017-11-19 at 20 37 38

After:
screen shot 2017-11-19 at 20 39 59


vibhavp:
Ah, so that was causing those erroneous highlights. Thanks.

PR #185, opened 2017-11-20
Fix 'end of buffer' error

sebastiencs:
When forward-char reaches the end of buffer it raises an error.
This commit silent the errors.

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -40,14 +40,18 @@
 
 (defvar lsp--no-response)
 
-;; from http://emacs.stackexchange.com/questions/8082/how-to-get-buffer-position-given-line-number-and-column-number
 (defun lsp--position-to-point (params)
   "Convert Position object in PARAMS to a point."
   (save-excursion
-    (goto-char (point-min))
-    (forward-line (gethash "line" params))
-    (forward-char (gethash "character" params))
-    (point)))
+    (save-restriction
+      (widen)
+      (goto-char 1)
+      ;; The next line calculs the point from the LSP position.
+      ;; We use `goto-char' to ensure that we return a point inside the buffer
+      ;; to avoid out of range error
+      (goto-char (+ (line-beginning-position (1+ (gethash "line" params)))
+                    (gethash "character" params)))
+      (point))))
 
 (defun lsp-make-traverser (name)
   "Return a closure that walks up the current directory until NAME is found.


lsp-common.el
```diff
@@ -45,8 +45,11 @@
   "Convert Position object in PARAMS to a point."
   (save-excursion
     (goto-char (point-min))
-    (forward-line (gethash "line" params))
-    (forward-char (gethash "character" params))
+    (let ((line (gethash "line" params))
+          (char (gethash "character" params)))
+      (ignore-errors
+        (forward-line line)
```
phst: `forward-line` doesn't signal errors, so you don't need to include it here.

lsp-common.el
```diff
@@ -45,8 +45,11 @@
...
+        (forward-char char)))
```
phst: I'd avoid `ignore-errors` if possible, it masks unrelated bugs such as type errors.
How about converting the entire function to

```elisp
  (save-restriction
    (widen)
    (+ (line-beginning-position line) column))
```

sebastiencs: I agree, it's a better idea

sebastiencs:
@phst Like that ?

vibhavp:
When does this error happen, though? An out of range error would mean that a language server is sending an incorrect `position` object, right?

sebastiencs:
Yes probably, the rls sometimes send incorrect position, I can see that with symbols highlighting

vibhavp:
Is that a known bug in rls? I suspect this might be a bug in lsp-mode itself.

sebastiencs:
I just checked the positions returned by the server with `lsp-print-io` and they are incorrect. I will open an issue on `rls` if the bug isn't known

vibhavp:
Alright, LGTM. @phst any more feedback?

lsp-common.el
```diff
@@ -40,14 +40,15 @@
...
-    (goto-char (point-min))
-    (forward-line (gethash "line" params))
-    (forward-char (gethash "character" params))
-    (point)))
+    (save-restriction
+      (widen)
+      (goto-char (point-min))
+      (goto-char (+ (line-beginning-position (1+ (gethash "line" params)))
+                    (gethash "character" params)))
+      (point))))
```
phst: Please add a comment why we use `goto-char` and `point` instead of just returning the value.

lsp-common.el
```diff
@@ -40,14 +40,15 @@
...
 (defun lsp--position-to-point (params)
   "Convert Position object in PARAMS to a point."
   (save-excursion
-    (goto-char (point-min))
-    (forward-line (gethash "line" params))
-    (forward-char (gethash "character" params))
-    (point)))
+    (save-restriction
+      (widen)
+      (goto-char (point-min))
```
phst: `(goto-char (point-min))` is the same as `(goto-char 1)`

sebastiencs:
Done. I have rebased the commits.
The bug with rls is [fixed](https://github.com/rust-lang-nursery/rls/pull/584)

jaelsasser:
I pulled ed551a94f2974e33b7cf345b348616d313a31485 into my local tree and have been running it today. It's done the trick for me.

vibhavp:
Thinking about this, if a server sends an incorrect `Position` in a `TextEdit` object to be applied to a document, and the new `lsp--position-to-point` wraps it to a point so as to avoid errors, wouldn't we be applying a wrong text edit? IMO, a wrong `Position` is a problem with the language server, which can't be fixed by silently ignoring it and converting it to a wrong location.

lsp-common.el
MaskRay: Looks good. `xref.el` might should be updated in the same manner. It uses the following snippet:

```elisp
(save-restriction
        (widen)
        (save-excursion
          (goto-char (point-min))
          (beginning-of-line line)
          (forward-char column)
          (point-marker)))
```

yyoncho:
Cleanup - this PR is stalling for about a year. Closing as per https://github.com/emacs-lsp/lsp-mode/pull/185#issuecomment-346941486 please reopen if you disagree.

PR #186, opened 2017-11-20
Asynchronous "hover" (eldoc)

topisani:
This is definately not perfect, but its a lot better than the synchronous aproach which would often freeze.

Instead of actually returning the eldoc string from the eldoc backend, it returns nil, and sends an async hover request. in the callback it then displays the string using `eldoc-message`

It might be interesting to do something like [this](https://github.com/commercialhaskell/intero/commit/1bf8470f27beb2e2b2aec775acfbd22364963b24) instead, but for me at least the simple solution seems fine

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1170,26 +1170,26 @@ Returns xref-item(s)."
     contents: MarkedString | MarkedString[];
     range?: Range;
 }
-
 type MarkedString = string | { language: string; value: string };"
   (lsp--cur-workspace-check)
   (lsp--send-changes lsp--cur-workspace)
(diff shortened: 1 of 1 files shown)


vibhavp:
I've improved this and pushed the changes to https://github.com/emacs-lsp/lsp-mode/tree/eldoc-async.
Could you test this branch with cquery? Thanks.

topisani:
Thats looking good, and cquery appears to like it!

vibhavp:
Alright, will merge this: https://github.com/emacs-lsp/lsp-mode/pull/187. Thanks!

PR #188, opened 2017-11-22
Highlight only visible references

sebastiencs:
This is intended to limit the number of overlays in the buffer and
should probably improve performances with big files and lots of references.


vibhavp:
Looks good, thanks.

PR #189, opened 2017-11-22
Add macros to add and remove a project root from the whitelist

alanz:
As an alternative to putting this in the enable function, as to my previous (ignored) PRs

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -133,10 +133,15 @@ Optional arguments:
 
 `:initialize' is a function called when the client is intiailized. It takes a
  single argument, the newly created client."
-  (let ((enable (intern (format "%s-enable" name))))
-    `(defun ,enable ()
+   (let ((enable (intern (format "%s-enable" name))))
+    `(progn
+    (lsp-define-whitelist-enable ,name ,get-root)
+    (lsp-define-whitelist-disable ,name ,get-root)
+
+    (defun ,enable ()
        ,docstring
        (interactive)
+       (lsp-define-whitelist-disable ,name ,get-root)
        (let ((client (make-lsp--client
                        :language-id (if ,language-id-fn ,language-id-fn #'(lambda (_) ,language-id))
                        :send-sync #'lsp--stdio-send-sync
@@ -155,7 +160,7 @@ Optional arguments:
                (progn
                  (lsp-mode 1)
                  (lsp--start client ,extra-init-params))
-               (message "Not initializing project %s" root))))))))
+               (message "Not initializing project %s" root)))))))))
 
 (cl-defmacro lsp-define-tcp-client (name language-id get-root command host port
                                      &key docstring
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -79,6 +79,32 @@
                 "lsp-define-client: :ignore-regexps element %s is not a string"
                 e))))
 
+(cl-defmacro lsp-define-whitelist-enable (name get-root
+                                               &key docstring)
+  "Define a function to add the project root for the current buffer to the whitleist.
+NAME is the base name for the command.
+GET-ROOT is the language-specific function to determint the project root for the current buffer."
+  (let ((enable (intern (format "%s-whitelist-add" name))))
+    `(defun ,enable ()
+       ,docstring
+       (interactive)
+       (let ((root (funcall ,get-root)))
+         (customize-save-variable 'lsp-project-whitelist
+                                    (add-to-list 'lsp-project-whitelist root))))))
+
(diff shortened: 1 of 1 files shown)


bkchr:
Where would you call this macro?
`lsp-project-whitelist` is nil by default, where do you initialize the variable?

alanz:
It is a `defcustom` variable, see https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-methods.el#L131

So you mean the code needs to cater for the initial nil case?

bkchr:
I would think so, but I'm no elisp expert. If you tried and it works for you, then it should be okay :)

What I wanted to ask with my first question, this `cl-defmacro` means that you would need to instantiate the macro. I wanted to know where you would need to instantiate the macro? 

Also I think, that this function don't need to be instantiated per supported language. Something like `lsp-whitelist-project`should be okay? `lsp-whitelist-project-remove` for the other case?

alanz:
It is defined per language, haskell version is https://github.com/alanz/lsp-haskell/blob/manage-whitelist/lsp-haskell.el#L258

bkchr:
Do you think it is important to have it per language? I think one for all should be okay.

alanz:
Well, the key thing is to find the project root, which is specific to the language

Editing a whitelist by hand otherwise is too much of a pain. In my opinion.

bkchr:
Yeah that would be a pain, you are right. Then I would recommend to instantiate your macros by `lsp-define-stdio-client` and `lsp-define-tcp-client`. So, the functions are automatically defined for each supported language :)

alanz:
If you can show me how to define more than one top-level macro at a time, I
will gladly do it.

alanz:
And I just deleted the whitelist from my custom settings and added a project. It worked fine. So `nil` initial value seems not to be a problem.

bkchr:
Maybe @vibhavp has an opinion or idea on how to call a macro inside of a macro :D

vibhavp:
Nah, I'm clueless about this :/
@alanz I thought your whitelist PR was still WIP, so I chose to comment once it was completed. Sorry about that.

alanz:
@vibhavp Ok, I will rebase my original then, and request a review

bkchr:
Okay, after fighting with lisp, I finally got something working:

```diff --git a/lsp-mode.el b/lsp-mode.el
index a806d1c..5ff4767 100644
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -133,10 +133,15 @@ Optional arguments:
 
 `:initialize' is a function called when the client is intiailized. It takes a
  single argument, the newly created client."
-  (let ((enable (intern (format "%s-enable" name))))
-    `(defun ,enable ()
+   (let ((enable (intern (format "%s-enable" name))))
+    `(progn
+    (lsp-define-whitelist-enable ,name ,get-root)
+    (lsp-define-whitelist-disable ,name ,get-root)
+
+    (defun ,enable ()
        ,docstring
        (interactive)
+       (lsp-define-whitelist-disable ,name ,get-root)
        (let ((client (make-lsp--client
                        :language-id (if ,language-id-fn ,language-id-fn #'(lambda (_) ,language-id))
                        :send-sync #'lsp--stdio-send-sync
@@ -155,7 +160,7 @@ Optional arguments:
                (progn
                  (lsp-mode 1)
                  (lsp--start client ,extra-init-params))
-               (message "Not initializing project %s" root))))))))
+               (message "Not initializing project %s" root)))))))))
 
 (cl-defmacro lsp-define-tcp-client (name language-id get-root command host port
                                      &key docstring
```

Could apply this patch and try? This generates the add/remove function while evaluating the main macro.

alanz:
I can try, but do not hold out much hope. An earlier version I tried looked like that, and did not work.

I cam to the conclusion that you can only define one macro per top level call

bkchr:
No, the functions are declared for me and I can call them :) But the whitelisting does not work for me :D

alanz:
ok, will give it a try. A bit later.

alanz:
@bkchr your patch works, great.

But I am going to go back to my original approach insteead

bkchr:
I would like this new approach. Your old approach introduces a new client and that needs to be implemented for each new/old language. I would propose to use this new approach, as no changes to the language plugins is required.

alanz:
Yes, I intend using your example to make the necessary functions for my approach in https://github.com/emacs-lsp/lsp-mode/pull/173, since @vibhavp  has stated that he is OK with that concept

alanz:
Superseded by https://github.com/emacs-lsp/lsp-mode/pull/190

PR #190, opened 2017-11-22
Split the lsp-XXXX-enable into non-interactive and interactive

alanz:
The non-interactive one gets called in the mode hook, and is quiet

The interactive one will ask to add a project to the whitelist if invoked to
start LSP for it and it would not otherwise start, if `lsp-enable-whitelist-ask'
is set.

Also add
  lsp-XXXX-whitelist-add
  lsp-XXXX-whitelist-remove

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ Clone this repository to a suitable path, and add
 ;;
 ;; This function will turn lsp-mode on and call the command given to
 ;; start the LSP server.
-(add-hook 'prog-major-mode #'lsp-prog-major-mode-enable)
+(add-hook 'prog-major-mode #'lsp-prog-major-mode-enable-hook)
 ```
 to your .emacs, where `prog-major-mode` is the hook variable for a supported
 programming language major mode.
@@ -92,3 +92,20 @@ to find the references to the symbol under point.
 
 ### Flycheck
 ![flycheck](./examples/flycheck.gif)
+
+## Finer Control of Starting lsp-mode
+
+In order to more finely control the `lsp-mode` startup, there are a number of
+customizable variables.
+
+`lsp-project-whitelist` : Defaults to `nil`. If set, `lsp-mode` will only be
+started if the given project root appears in the whitelist.
+
+`lsp-project-blacklist` : Defaults to `nil`. If set, all projects will be
+started except those in this list. It is ignored if `lsp-project-whitelist` is
+set.
+
+There are also the functions `lsp-MAJOR-MODE-whitelist-add` and
+`lsp-MAJOR-MODE-whitelist-remove` to adjust the current buffer project root
+entry on the whitelist.
+
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -79,6 +79,48 @@
                 "lsp-define-client: :ignore-regexps element %s is not a string"
                 e))))
 
+;; --------------------------------------------------------
+
+(cl-defmacro lsp-define-whitelist-enable (name get-root
+                                               &key docstring)
+  "Define a function to add the project root for the current buffer to the whitleist.
+NAME is the base name for the command.
+GET-ROOT is the language-specific function to determint the project root for the current buffer."
+  (let ((whitelist-add      (intern (format "%s-whitelist-add" name)))
+        (enable-interactive (intern (format "%s-enable" name))))
+    `(defun ,whitelist-add ()
+       ,docstring
+       (interactive)
+       (let ((root (funcall ,get-root)))
+         (customize-save-variable 'lsp-project-whitelist
+                                  (add-to-list 'lsp-project-whitelist root))
+         (,enable-interactive)
+         ))))
+
+(cl-defmacro lsp-define-whitelist-disable (name get-root
+                                               &key docstring)
+  "Define a function to remove the project root for the current buffer from the whitleist.
+NAME is the base name for the command.
+GET-ROOT is the language-specific function to determint the project root for the current buffer."
+  (let ((whitelist-remove (intern (format "%s-whitelist-remove" name))))
+    `(defun ,whitelist-remove ()
+       ,docstring
+       (interactive)
+       (let ((root (funcall ,get-root)))
+         (customize-save-variable 'lsp-project-whitelist
+                                  (remove root 'lsp-project-whitelist ))))))
+
+(cl-defmacro lsp-define-interactive-client (name)
+  "Define a function of the form `lsp-NAME-enable' that calls the
+`lsp-NAME-enable-hook' with its FORCE flag set to `t', to enable
+LSP mode regardless of whitelist/blacklist settings."
+  (let ((enable-noninteractive (intern (format "%s-enable-hook" name)))
+        (enable-interactive    (intern (format "%s-enable" name)))
+        )
+    `(defun ,enable-interactive (&optional force)
+       (interactive)
+       (,enable-noninteractive t))))
+
 (cl-defmacro lsp-define-stdio-client (name language-id get-root command
                                        &key docstring
                                        language-id-fn
@@ -109,30 +151,27 @@ Optional arguments:
  a single argument (LSP workspace) and returning a plist is also accepted.
 
 `:initialize' is a function called when the client is intiailized. It takes a
- single argument, the newly created client."
-  (let ((enable (intern (format "%s-enable" name))))
-    `(defun ,enable ()
-       ,docstring
-       (interactive)
-       (let ((client (make-lsp--client
-                       :language-id (if ,language-id-fn ,language-id-fn #'(lambda (_) ,language-id))
-                       :send-sync #'lsp--stdio-send-sync
-                       :send-async #'lsp--stdio-send-async
-                       :new-connection (lsp--make-stdio-connection
+  single argument, the newly created client."
+  (let ((enable (intern (format "%s-enable-hook" name))))
+    `(progn
+       (lsp-define-whitelist-enable ,name ,get-root)
+       (lsp-define-whitelist-disable ,name ,get-root)
+       (lsp-define-interactive-client ,name)
+
+       (defun ,enable (&optional force)
+         ,docstring
+         (interactive)
+         (let ((client (make-lsp--client
+                        :language-id (if ,language-id-fn ,language-id-fn #'(lambda (_) ,language-id))
+                        :send-sync #'lsp--stdio-send-sync
+                        :send-async #'lsp--stdio-send-async
+                        :new-connection (lsp--make-stdio-connection
                                          ,(symbol-name name)
                                          ,command
                                          ,command-fn)
-                       :get-root ,get-root
-                       :ignore-regexps ,ignore-regexps)))
-         (unless lsp-mode
-           ,(when initialize
-              `(funcall ,initialize client))
-           (let ((root (funcall (lsp--client-get-root client))))
-             (if (lsp--should-start-p root)
-               (progn
-                 (lsp-mode 1)
-                 (lsp--start client ,extra-init-params))
-               (message "Not initializing project %s" root))))))))
+                        :get-root ,get-root
+                        :ignore-regexps ,ignore-regexps)))
+           (lsp--start-client-maybe client force :extra-init-params ,extra-init-params :initialize ,initialize))))))
 
 (cl-defmacro lsp-define-tcp-client (name language-id get-root command host port
                                      &key docstring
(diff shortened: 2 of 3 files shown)


alanz:
/cc @bkchr 

@vibhavp This one is for real :)

bkchr:
You also need to add support for lsp TCP client.

(tip: you don't need to create a new pull request for each new version )

alanz:
True.

The other PR was taking a simpler approach to just do the whitelist add/remove, that is folded in here.

And for the prior ones, I don't like force pushing a rebase, it leaves people stranded who may have been using the original one.

So I end up making a new PR on every rebase.

alanz:
@bkchr 

I am splitting out the initialise logic that is common, and end up with

```elisp
(defun lsp--start-client-maybe (client
                                &key extra-init-params initialize)
"Start a client if lsp-mode is currently inactive, and the whitelist/blacklist allows it."
           (unless lsp-mode
             ,(when initialize
                `(funcall ,initialize client))
             (let ((root (funcall (lsp--client-get-root client))))
               (if (lsp--should-start-p root)
                   (progn
                     (lsp-mode 1)
                     (lsp--start client ,extra-init-params))
                 (if (and ask lsp-enable-whitelist-ask)
                     (let ((question (format-message "Add %s to lsp-project-whitelist? " root)))
                       (if (y-or-n-p question) ;; cannot use when , side effects problem
                           (progn (message "About to customize lsp-project-whitelist")
                                  (customize-save-variable 'lsp-project-whitelist (add-to-list 'lsp-project-whitelist root)))
                         (setq lsp--project-whitelist-temp (add-to-list 'lsp--project-whitelist-temp root)))
                       ;; start regardless, this is called interactively
                       (lsp-mode 1)
                       (lsp--start client))
                   (message "Not initializing project %s" root))))))
```

How would I call it to pass on the two optional keyword params from the original param list?

bkchr:
Your question is, how you would call this function with optional parameters?

alanz:
I want to enable this line : https://github.com/emacs-lsp/lsp-mode/pull/190/commits/a0238defb9f010e9a3e34fc3cc4223d57e49ab02#diff-b1e9bc82c3b6bd6e95470e18ee41a19bR165

bkchr:
@alanz I fixed your problem. But your implementation has some further logic problems. I also don't understand what the variable `lsp--project-whitelist-temp` is for. Further, you also need to define `lsp-enable-whitelist-ask` as default to `nil`. While looking over the code, I thought that this question should probably asked in the `lsp--should-start-p` function, as this function already is doing the logic of checking the whitelist.


```diff --git a/lsp-mode.el b/lsp-mode.el
index 7e22540..3ff1af7 100644
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -162,49 +162,30 @@ Optional arguments:
                                          ,command-fn)
                         :get-root ,get-root
                         :ignore-regexps ,ignore-regexps)))
-           ;; (apply #'lsp--start-client-maybe client &rest)
-           (unless lsp-mode
-             ,(when initialize
-                `(funcall ,initialize client))
-             (let ((root (funcall (lsp--client-get-root client))))
-               (if (lsp--should-start-p root)
-                   (progn
-                     (lsp-mode 1)
-                     (lsp--start client ,extra-init-params))
-                 (if (and ask lsp-enable-whitelist-ask)
-                     (let ((question (format-message "Add %s to lsp-project-whitelist? " root)))
-                       (if (y-or-n-p question) ;; cannot use when , side effects problem
-                           (progn (message "About to customize lsp-project-whitelist")
-                                  (customize-save-variable 'lsp-project-whitelist (add-to-list 'lsp-project-whitelist root)))
-                         (setq lsp--project-whitelist-temp (add-to-list 'lsp--project-whitelist-temp root)))
-                       ;; start regardless, this is called interactively
-                       (lsp-mode 1)
-                       (lsp--start client))
-                   (message "Not initializing project %s" root)))))
-           ))))
-  )
-
-;; (defun lsp--start-client-maybe (client
-;;                                 &key extra-init-params initialize)
-;; "Start a client if lsp-mode is currently inactive, and the whitelist/blacklist allows it."
-;;            (unless lsp-mode
-;;              ,(when initialize
-;;                 `(funcall ,initialize client))
-;;              (let ((root (funcall (lsp--client-get-root client))))
-;;                (if (lsp--should-start-p root)
-;;                    (progn
-;;                      (lsp-mode 1)
-;;                      (lsp--start client ,extra-init-params))
-;;                  (if (and ask lsp-enable-whitelist-ask)
-;;                      (let ((question (format-message "Add %s to lsp-project-whitelist? " root)))
-;;                        (if (y-or-n-p question) ;; cannot use when , side effects problem
-;;                            (progn (message "About to customize lsp-project-whitelist")
-;;                                   (customize-save-variable 'lsp-project-whitelist (add-to-list 'lsp-project-whitelist root)))
-;;                          (setq lsp--project-whitelist-temp (add-to-list 'lsp--project-whitelist-temp root)))
-;;                        ;; start regardless, this is called interactively
-;;                        (lsp-mode 1)
-;;                        (lsp--start client))
-;;                    (message "Not initializing project %s" root))))))
+           (lsp--start-client-maybe client :extra-init-params ,extra-init-params :initialize ,initialize)
+  )))))
+
+(cl-defun lsp--start-client-maybe (client
+                                 &key extra-init-params initialize)
+ "Start a client if lsp-mode is currently inactive, and the whitelist/blacklist allows it."
+            (unless lsp-mode
+              (when initialize
+                 (funcall initialize client))
+              (let ((root (funcall (lsp--client-get-root client))))
+                (if (lsp--should-start-p root)
+                    (progn
+                      (lsp-mode 1)
+                      (lsp--start client extra-init-params))
+                  (if (and ask lsp-enable-whitelist-ask)
+                      (let ((question (format-message "Add %s to lsp-project-whitelist? " root)))
+                        (if (y-or-n-p question) ;; cannot use when , side effects problem
+                            (progn (message "About to customize lsp-project-whitelist")
+                                   (customize-save-variable 'lsp-project-whitelist (add-to-list 'lsp-project-whitelist root)))
+                          (setq lsp--project-whitelist-temp (add-to-list 'lsp--project-whitelist-temp root)))
+                        ;; start regardless, this is called interactively
+                        (lsp-mode 1)
+                        (lsp--start client extra-init-params))
+                    (message "Not initializing project %s" root))))))
 
 (cl-defmacro lsp-define-tcp-client (name language-id get-root command host port
                                      &key docstring
```

alanz:
Thanks, I appreciate your help. I am very much a part-time elisper.

lsp-mode.el
```diff
@@ -161,29 +194,48 @@ Optional arguments:
...
+                (if (lsp--should-start-p root)
+                    (progn
+                      (lsp-mode 1)
+                      (lsp--start client extra-init-params))
+                  (if (and ask lsp-enable-whitelist-ask)
+                      (let ((question (format-message "Add %s to lsp-project-whitelist? " root)))
+                        (if (y-or-n-p question) ;; cannot use when , side effects problem
+                            (progn (message "About to customize lsp-project-whitelist")
+                                   (customize-save-variable 'lsp-project-whitelist (add-to-list 'lsp-project-whitelist root)))
+                          (setq lsp--project-whitelist-temp (add-to-list 'lsp--project-whitelist-temp root)))
```
bkchr: I still don't understand the purpose of the `lsp--project-whitelist-temp` variable? The `lsp--project-whitelist` already stores all relevant information.

lsp-mode.el
```diff
@@ -161,29 +194,48 @@ Optional arguments:
...
+                        ;; start regardless, this is called interactively
+                        (lsp-mode 1)
+                        (lsp--start client extra-init-params))
```
bkchr: The client is started whether the user selects `y` or `n` in the question. I would suspect that the language server is not started if the I select `n` to white-listing.

lsp-mode.el
alanz: If the user is asked to add the project to the whitelist, and says no, the whitelist is not updated.

If they now open another file, that shares the same project root, the fact that it is temporarily allowed to start is stored in `lsp--project-whitelist-temp`.  Otherwise it needs to be enabled manually for the second file, and may create an additional session.

Basically it allows the whitelist override to operate at the project root level, rather than at the individual buffer level.

lsp-mode.el
alanz: No. The key feature of this PR is that we end up with `lsp-MAJOR-MODE-enable -hook` and `lsp-MAJOR-MODE-enable`.

The first one gets added to the `MAJOR-MODE` hook, and sets the `ask` parameter to nil.

The second one is called via `M-x` (i.e. interactively) and sets the `ask` parameter to `t`.

The logic is that `lsp-MAJOR-MODE-enable` will *always* start a session.  The only difference is that it may prompt to *also* add the project root to the whitelist, for future sessions.

bkchr: But who would call the second interactively, if he already has enabled the hook? 
And if you always want to call this interactively, you don't need the whitelist, as you are the one who decides when to start lsp.

lsp-mode.el
bkchr: I understand, but that is *REALLY* hacky. You would be asked everytime if you want to add this to the whitelist, that is a real shitty experience. 

I would really like the approach that you just call `lsp-whitelist-project-add` and if the project is added, you could automatically start the lsp client.

alanz: Ok, that could work.

I guess as a server-developer I am out of the normal use-case on this stuff.

As I want to have fine control of whether the server starts or not, based on the state of development of the server.

lsp-mode.el
alanz: I have a mixed case. Some projects I always want to start, others to only start by request.

Especially as the haskell lsp-server uses quite a bit of resources, so if you just want to look at code for a project while working on another, you do not necessarily want to start a server.

lsp-mode.el
bkchr: But why wouldn't your use case work with the `lsp-whitelist-project-add` function? As I understand, you want control, when you want a project to be started with the language server?

By call the `lsp-whitelist-project-add` you would enable the language server and by removing, you would stop it. Sorry if I'm not understanding something right :/ 
You could also stop the lsp-client if you remove the project from a whitelist.

alanz: I agree. I originally had the complex approach, then introduced the add/remove stuff, to do only that.

But making it control the server too is a good idea.

Thanks.

lsp-mode.el
bkchr: Hmm, I understand. Then I would propose. You remove the asking stuff of the interactive function, but preserve the interactive function. If you have a project that you always want to start by request, you use the interactive version. If you decide, for this project the language server should always be started you call `lsp-whitelist-project-add`.

alanz:
Ok, it definitely feels better/more logical now.

README.md
```diff
@@ -92,3 +92,22 @@ to find the references to the symbol under point.
...
+
+## Finer Control of Starting lsp-mode
+
+In order to more finely control the `lsp-mode` startup, there are a number of
+customizable variables.
+
+`lsp-project-whitelist` : Defaults to `nil`. If set, `lsp-mode` will only be
+started if the given project root appears in the whitelist.
+
+`lsp-project-blacklist` : Defaults to `nil`. If set, all projects will be
```
bkchr: Maybe add, that if `lsp-project-whitelist` is set, `lsp-project-blacklist` is ignored.

README.md
```diff
@@ -92,3 +92,22 @@ to find the references to the symbol under point.
...
+started except those in this list.
+
+If a project root appears in both lists, the `lsp-project-whitelist` overrides
```
bkchr: That is not right, as I have written above. If the whitelist is used, the blacklist is ignored.

lsp-mode.el
```diff
@@ -79,6 +79,45 @@
...
+  "Define a function to add the project root for the current buffer to the whitleist.
+NAME is the base name for the command.
+GET-ROOT is the language-specific function to determint the project root for the current buffer."
+  (let ((enable (intern (format "%s-whitelist-add" name))))
+    `(defun ,enable ()
+       ,docstring
+       (interactive)
+       (let ((root (funcall ,get-root)))
+         (customize-save-variable 'lsp-project-whitelist
+                                    (add-to-list 'lsp-project-whitelist root))))))
```
bkchr: I would start the lsp client here, if the project was added to the whitelist.

lsp-mode.el
```diff
@@ -79,6 +79,45 @@
...
+  "Define a function to remove the project root for the current buffer from the whitleist.
+NAME is the base name for the command.
+GET-ROOT is the language-specific function to determint the project root for the current buffer."
+  (let ((enable (intern (format "%s-whitelist-remove" name))))
+    `(defun ,enable ()
+       ,docstring
+       (interactive)
+       (let ((root (funcall ,get-root)))
+         (customize-save-variable 'lsp-project-whitelist
+                                  (remove root 'lsp-project-whitelist ))))))
```
bkchr: Here I would stop the client if the project was removed.

lsp-mode.el
```diff
@@ -160,30 +196,50 @@ Optional arguments:
...
+
+(cl-defun lsp--start-client-maybe (client force
+                                 &key extra-init-params initialize)
+ "Start a CLIENT if lsp-mode is currently inactive, and the whitelist/blacklist allows it.
+If FORCE is `t', start regardless off whitelist/blacklist setting."
+            (unless lsp-mode
+              (when initialize
+                 (funcall initialize client))
+              (let ((root (funcall (lsp--client-get-root client))))
+                (if (lsp--should-start-p root)
```
bkchr: Nitpick: could be `(or force (lsp--should-start-p root)` and 
```
(lsp-mode 1)
(lsp--start client extra-init-params)
```
could be removed from the lower if

alanz: Thanks, nothing like getting lessons in idiomatic elisp

lsp-mode.el
alanz: I don' think there is an `lsp-MAJOR-MODE-disable` at the moment.

lsp-mode.el
bkchr: Can't say if that is idiomatic elisp, just my general programming skills that want to remove duplicate code :D

lsp-mode.el
bkchr: You could call `lsp--shutdown-cur-workspace`.

alanz: I just tried that, and got a mountain of problems. I suspect that the shutdown is not all that solid, tbh.

I would rather put this in without that feature, and it can be added later as part of sorting out the shutdown.

This thing is complex enough already.

bkchr: Yeah :) You are right.

bkchr:
@vibhavp I think we are ready to merge :)

lsp-mode.el
```diff
@@ -106,30 +148,27 @@ Optional arguments:
...
-    `(defun ,enable ()
-       ,docstring
-       (interactive)
-       (let ((client (make-lsp--client
-                       :language-id (if ,language-id-fn ,language-id-fn #'(lambda (_) ,language-id))
-                       :send-sync #'lsp--stdio-send-sync
-                       :send-async #'lsp--stdio-send-async
-                       :new-connection (lsp--make-stdio-connection
+  single argument, the newly created client."
+  (let ((enable (intern (format "%s-enable-hook" name))))
```
vibhavp: By renaming the entrypoint to `%s-enable-hook`, we'd be break `lsp-mode` for all people that currently use it in their configs. Is there a better way to achieve this?

README.md
bkchr: Does this questions belongs to this code section? I don't understand it :D

lsp-mode.el
bkchr: Hmm good question. We could increase the version number or maybe find some better trick :/

alanz:
I am getting tired of reworking this and not having it land.

@vibhavp if I revert to just adding whitelist manipulation functions, will you merge it?

vibhavp:
The main issue for me is breaking already existing entrypoints for lsp-mode. #204 is a good first step, I'll merge that for now.

alanz:
I think this has been resolved by the whitelist-add and whitelist-remove commands.

PR #191, opened 2017-11-23
lsp-mode complains about "(wrong-number-of-arguments (1 . 2) 0)"

jaelsasser:
A recent commit attached a call to `(lsp--start)` to the lsp-mode's define-minor-mode call. Causes the following error while toggling the minor mode:

```
Debugger entered--Lisp error: (wrong-number-of-arguments (1 . 2) 0)
  lsp--start()
  lsp-mode(toggle)
  funcall-interactively(lsp-mode toggle)
  call-interactively(lsp-mode record nil)
  command-execute(lsp-mode record)
  execute-extended-command(nil "lsp-mode" "lsp-mode")
  funcall-interactively(execute-extended-command nil "lsp-mode" "lsp-mode")
```

Also drops an unneeded call to `(lsp--merge-plists)` that I introduced with #181. It's not related, but it's really been bugging me (but doesn't seem worth an entire PR on its own)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -196,8 +196,7 @@ Optional arguments:
 (define-minor-mode lsp-mode ""
   nil nil nil
   :lighter (:eval (lsp-mode-line))
-  :group 'lsp-mode
-  (lsp--start))
+  :group 'lsp-mode)
 
 (defconst lsp--sync-type
   `((0 . "None")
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -196,8 +196,7 @@ Optional arguments:
 (define-minor-mode lsp-mode ""
   nil nil nil
   :lighter (:eval (lsp-mode-line))
-  :group 'lsp-mode
-  (lsp--start))
+  :group 'lsp-mode)
```
jaelsasser: At some point, it'd probably be good to have lsp-mode do more than just mange the lsp-mode variable, but for now this is causing issues.

vibhavp:
Thanks.

PR #196, opened 2017-11-25
Show reference'line with xref-find-references

sebastiencs:
Extract the line from the buffer visiting its file and show it in the
list of references with xref-find-references.

xref

Old behavior:
xref_old

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1149,7 +1149,6 @@ The function returns a list of `xref-item'."
          (fn (lambda (loc) (lsp--xref-make-item filename loc))))
     (if visiting
         (with-current-buffer visiting
-          (font-lock-ensure)
           (mapcar fn (cdr file)))
       (when (file-readable-p filename)
         (with-temp-buffer
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1113,19 +1113,48 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
 	item))
     item))
 
-(defun lsp--location-to-xref (location)
-  "Convert Location object LOCATION to an ‘xref-item’.
-interface Location {
-    uri: string;
-    range: Range;
-}"
-  (lsp--send-changes lsp--cur-workspace)
-  (let ((uri (string-remove-prefix "file://" (gethash "uri" location)))
-        (ref-pos (gethash "start" (gethash "range" location))))
-    (xref-make uri
-               (xref-make-file-location uri
-                                        (1+ (gethash "line" ref-pos))
-                                        (gethash "character" ref-pos)))))
+(defun lsp--extract-line-from-buffer (pos)
+  "Return a line from the current buffer.
+POS is a LSP position on the line."
+  (let* ((point (lsp--position-to-point pos))
+         (inhibit-field-text-motion t))
+    (save-excursion
+      (goto-char point)
+      (buffer-substring (line-beginning-position) (line-end-position)))))
+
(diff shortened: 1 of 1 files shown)


MaskRay:
Thanks for making this PR. This is awesome! I think for future improvement, we may use a file cache (as Python's `linecache` module used by @DeltaEvo in https://github.com/autozimu/LanguageClient-neovim/blob/master/rplugin/python3/LanguageClient/LanguageClient.py) for files that are not opened. For a definitions/references response, if the file is being visited, fetch the line from the buffer; otherwise `stat(2)` the file and check whether the file has been updated (`linecache.checkcache()`), if not retrieve the line from the file cache (`linecache.getline()`).

DeltaEvo:
It's not me, it's @autozimu ;)

lsp-methods.el
```diff
@@ -1113,16 +1113,44 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
...
+             (line-range (save-excursion (goto-char start)
+                                         (cons (line-beginning-position)
+                                               (line-end-position))))
+             (line (buffer-substring (car line-range) (cdr line-range)))
+             (len (length line))
+             (start-on-line (- start (car line-range)))
+             (end-on-line (+ start-on-line (- end start))))
+        (add-face-text-property (max (min start-on-line len) 0)
+                                (max (min end-on-line len) 0)
+                                '(:box (:line-width -1 :color "white")) t line)
```
jaelsasser: Nit: maybe use `lsp-face-highlight-textual` here so that the lines match the in-buffer symbol highlighting.

sebastiencs:
@MaskRay Thanks for your advice, That's really interesting.
It's definitely something we should think about

lsp-methods.el
sebastiencs: I don't think it's a good idea. I use a discret background for `lsp-face-highlight-textual`, not something really visible. In the contrary the reference should be very visible on the xref list.
Instead of box I use the face `highlight` now

vibhavp:
Looks good. Have you tried this with slightly larger projects?

lsp-methods.el
jaelsasser: Ah, fair, I’ve got a pretty bold symbol highlight face.

Maybe a defcustom / new lsp face? This is the kinda thing I can see users wanting to tweak.

sebastiencs:
I have tested it to find 1153 references in 83 files. The average is 0.27 second.
The test doesn't include the request to the server and I use a ssd.

lsp-methods.el
sebastiencs: Yes why not, but the face `highlight` is already customizable. Do you think it worth it to create a new face ?

vibhavp:
Looks good, thanks. We can add a customization option to limit the number of line lookups later.

vibhavp:
Sorry, I had to remove this as it removed `lsp--location-to-xref`, which broke all other functions that were using it.

jaelsasser:
I have this [up in a branch](https://github.com/jaelsasser/lsp-mode/tree/show-references-compat) with a small compat patch because I'm looking forward to using this.

`xref-backend-apropros` needs to be fixed for this, too, but I don't think that's as pressing, as it seems to be broken right now on master anyways. (Can't get it to work with/without this loaded up. I think we should be calling some `lsp--location-to-xref-location` instead of what's effectively  `(xref-make description (xref-make ...)`.)

sebastiencs:
Oops.. That's embarassing
I will make a new PR

PR #199, opened 2017-11-28
[WIP] Fix xref

sebastiencs:
\+ Change `xref-backend-definitions`, `lsp--get-references` and `lsp--get-definitions`

Following #196

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1122,15 +1122,6 @@ POS is a LSP position on the line."
       (goto-char point)
       (buffer-substring (line-beginning-position) (line-end-position)))))
 
-(defun lsp--location-to-xref (location)
-  "Convert Location object LOCATION to an `xref-item'.
- interface Location {
- 	uri: string;
- 	range: Range;
- }"
-  (let ((uri (string-remove-prefix "file://" (gethash "uri" location))))
-    (lsp--xref-make-item uri location)))
-
 (defun lsp--xref-make-item (filename location)
   "Return a xref-item from a LOCATION in FILENAME."
   (let* ((range (gethash "range" location))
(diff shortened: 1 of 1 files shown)


sebastiencs:
I will refactor the code and fix apropos

sebastiencs:
It should be ok now

lsp-methods.el
```diff
@@ -1113,30 +1113,70 @@ https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#co
...
+                            (max (min end len) 0)
+                            'highlight t line)
+    (xref-make (or line filename)
+               (xref-make-file-location filename
+                                        (1+ (gethash "line" pos-start))
+                                        (gethash "character" pos-start)))))
+
+(defun lsp--get-xrefs-in-file (file)
+  "Return all references that contain a file.
+FILE is a cons where its car is the filename and the cdr is a list of locations.
```
vibhavp: Why is the CDR a list of locations? If we're only removing the "file://" prefix from the filename string, shouldn't that be just a single element? Eg: `(("file:///etc/foo" "/etc/foo") ("file:///etc/bar" "/etc/bar"))`

sebastiencs: The param `file` is an element returned by `seq-group-by`:
```elisp
(seq-group-by (lambda (n) (if (evenp n) "even" "odd")) '(1 2 3 4 5 6))
=> (("odd" 1 3 5) ("even" 2 4 6))
```
The car of each element is the result of the function given to `seq-group-by` and their cdr are unchanged element from the param `SEQUENCE`.
In our case elements are locations:
```elisp
("/path/file" ("file:///path/file" range) ("file:///path/file" another_range))
```

vibhavp: Ah, got it. Thanks

PR #202, opened 2017-11-30
Fix diagnostics ranges

sebastiencs:
Prior to this commit, the range was calculated using
`lsp--position-to-point`. But this function is supposed to be called
from the buffer where it has to calculate the point.
As diagnostics can be reported for differents files, it could lead to
wrong points. And errors were thrown from `lsp--position-to-point`


vibhavp:
LGTM, thanks.

PR #203, opened 2017-12-02
Try sending another newline when there is no response before failing

fejfighter:
This seems to be required for clangd.

### DIFF
--- a/lsp-send.el
+++ b/lsp-send.el
@@ -38,7 +38,9 @@
   (with-local-quit
     (accept-process-output proc lsp-response-timeout))
   (when lsp--no-response
-    (signal 'lsp-timed-out-error nil)))
+    (process-send-string proc "\r\n") ; sometimes the server just needs a little coaxing
+    (when lsp--no-response
+    (signal 'lsp-timed-out-error nil))))
 
 (defun lsp--stdio-send-async (message proc)
   (when lsp-print-io


fejfighter:
lsp-clangd located here for the curious https://github.com/fejfighter/lsp-clangd

vibhavp:
Wouldn't this break behavior with other language servers? Would sending an empty message (`Content-Length:0\r\n\r\n`) work?

fejfighter:
I'm not sure, and in my rush before pushing up, I didn't check.
The intention was to only give the server a 'nudge' if it wasn't responding (I dropped the timeout to 2s when testing it)

Having thought about it, and looked through the spec a little there is no requirement for a closing `\r\n` so it's possible that this is a workaround for clangd, not needed elsewhere

Using clangd directly and sending the empty message had issues, (text after message content.

fejfighter:
https://github.com/llvm-mirror/clang-tools-extra/blob/6f250f705d384f21677d1efcbfad8993c797424c/clangd/JSONRPCDispatcher.cpp#L133

Is the assumption that the clangd guys have made, but I can't find a reference to a blank link for termination in the LSP spec

vibhavp:
Have you tried taking this up with clangd?

fejfighter:
I have started digging through the clangd source.

I think it may relate to differences in the handling of line endings, but that is pure speculation 

I'll leave this here as a WIP until I can understand it better

fejfighter:
@vibhavp, It appears to related to the length calculation/newline issues
emacs-lsp produces this:

`Content-Length: 336

{"jsonrpc":"2.0","method":"initialize","params":{"processId":9774,"rootPath":"/Users/fejfighter/dev/mu/","rootUri":"file:///Users/fejfighter/dev/mu/","capabilities":{"workspace":{"executeCommand":{"dynamicRegistration":true}},"textDocument":{"synchronization":{"willSave":true,"didSave":true}}},"initializationOptions":null},"id":3}`

dropping the length from 336 down to 333 and it works immediately. which makes sense, there are 3 newlines in there.

I am not sure where the blame lies, but this is not the correct fix, so I will close this MR

PR #205, opened 2017-12-02
Add lsp-imenu.el

vibhavp:
Enable with
```emacs-lisp
(require 'lsp-imenu)
(add-hook 'lsp-after-open-hook 'lsp-enable-imenu)
```

Examples:

Imenu categorizing symbol types:

Browsing through functions:

### DIFF
--- a/lsp-imenu.el
+++ b/lsp-imenu.el
@@ -1,0 +1,54 @@
+;; Copyright (C) 2016-2017  Vibhav Pant <vibhavp@gmail.com>  -*- lexical-binding: t -*-
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


vibhavp:
Also, here's an example of the same code being used with `helm`/`helm-imenu`:

MaskRay:
Nice! I'll abandon https://github.com/emacs-lsp/lsp-ui/pull/5

Can you also add an imenu entry for `textDocument/workspaceSymbol`? It looks up all symbols in a workspace, while `documentSymbol` is for symbols in current document. But I'm unclear how to have imenu with different purposes.

blahgeek:
> Can you also add an imenu entry for `textDocument/workspaceSymbol`? It looks up all symbols in a workspace, while `documentSymbol` is for symbols in current document. But I'm unclear how to have imenu with different purposes.

Is there any progress on this feature?

Issue #207, opened 2017-12-05
lsp--text-document-did-close fails for buffers with no file associated


Fuco1:
# how to reproduce

Invoke magit blame and try to kill the blame buffer.

I fixed it by wrapping the function inside `when buffer-file-name`, not sure what it does really so I don't know if this is appropriate.

There's also related error in `lsp--send-changes` where it tries to `with-current-buffer` on a buffer that was alerady killed... will error with "selecting deleted buffer".  I fixed that by wrapping it with `(buffer-live-p buffer)` check. 

I'm not really sure what is going on but I need to work now so I just hacked it in this way, seems to work fine. Let me know if you need some more info.

jaelsasser:
I've run into the same problem, but `(when buffer-file-name lsp-MODE-enable)` didn't do the trick for me. I'm using this instead:

```elisp
(defun jae--setup-cquery ()
  (when (and buffer-file-name (not magit-buffer-revision))
    (lsp-cquery-enable))
```

vibhavp:
I've pushed a fix at https://github.com/emacs-lsp/lsp-mode/pull/208. Could you try testing it?

jaelsasser:
Went through and tested this without the hack I mentioned above. (I forgot that I was still using it after all these weeks…). Looks like there's still some rough edges (using the latest `cquery`):

* navigate to a version-controlled project that's set up with a `compile_commands.json`
* `lsp-mode` starts up
* check out a revision using `magit-find-file`
* calling `xref-find-defintions` results in the following error message: `url-hexify-string: Wrong type argument: arrayp, nil` (see the attached backtrace)
* quitting the buffer results in the following error message: `Error on ‘lsp--text-document-did-close’: (wrong-type-argument arrayp nil)`

Backtrace for the `xref-find-definitions` error:

```emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument arrayp nil)
  file-truename(nil)
  (url-hexify-string (file-truename buffer-file-name) url-path-allowed-chars)
  (concat lsp--uri-file-prefix (url-hexify-string (file-truename buffer-file-name) url-path-allowed-chars))
  (list :uri (concat lsp--uri-file-prefix (url-hexify-string (file-truename buffer-file-name) url-path-allowed-chars)))
  (or nil (list :uri (concat lsp--uri-file-prefix (url-hexify-string (file-truename buffer-file-name) url-path-allowed-chars))))
[...]

PR #208, opened 2017-12-10
Cleanup how changes are sent.

vibhavp:
Removed all dead changes code.

Also, added a fix for #207

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -785,38 +785,32 @@ interface Range {
   ;;   lsp-on-change:(start,end,length)=(19,19,8)
 
   (if (eq length 0)
-      ;; Adding something only, work from start only
-      `(:range ,(lsp--range (lsp--point-to-position start)
-                            (lsp--point-to-position start))
-               :rangeLength 0
-               :text ,(buffer-substring-no-properties start end))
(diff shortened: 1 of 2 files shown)


lsp-methods.el
```diff
@@ -785,38 +785,32 @@ interface Range {
...
-          `(:range ,(lsp--range (lsp--point-to-position start)
-                                (plist-get lsp--before-change-vals :end-pos))
-                   :rangeLength ,length
-                   :text ,(buffer-substring-no-properties start end))
-        (lsp--change-for-mismatch start end length)))))
+        ;; The before-change value is bracketed, use it
+        `(:range ,(lsp--range (lsp--point-to-position start)
+                    (plist-get lsp--before-change-vals :end-pos))
+           :rangeLength ,length
+           :text "")
```
alanz: Are you sure this will always be the empty string? I guess we are in the "deleting only" zone, so should be.

alanz:
Ok, I looked in ediff where I can see more clearly what is going on, and it makes sense.  A lot simpler, since we are not able to batch up the changes anyway.

alanz:
Works as expected locally, so I am happy.

PR #209, opened 2017-12-12
Kill the stderr buffer when the process has exited

sebastiencs:


lsp-methods.el
```diff
@@ -493,17 +494,23 @@ registered client capabilities by calling
...
+            (with-current-buffer buf
+              (message "%s: %s has exited (%s)"
+                       (lsp--workspace-root lsp--cur-workspace)
+                       (process-name (lsp--workspace-proc lsp--cur-workspace))
+                       exit-str)
+              (lsp--uninitialize-workspace))))
+      (let ((status (process-status process))
+            (buffer-stderr (get-buffer stderr)))
+        (and (buffer-live-p buffer-stderr)
+             (or (equal status 'exit)
```
vibhavp: Since we're comparing symbols, `eq` would be better suited here.

sebastiencs:
Done

vibhavp:
LGTM, thanks.

PR #215, opened 2017-12-19
Add support for windows path uris

akermu:
https://en.wikipedia.org/wiki/File_URI_scheme#Windows_2

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -28,6 +28,11 @@
   :group 'lsp-mode
   :type 'boolean)
 
+(defvar lsp--uri-file-prefix (pcase system-type
+                               (`windows-nt "file:///")
+                               (_ "file://"))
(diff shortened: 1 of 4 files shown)


lsp-common.el
```diff
@@ -28,6 +28,11 @@
   :group 'lsp-mode
   :type 'boolean)
 
+(defvar lsp--uri-file-prefix (pcase system-type
```
vibhavp: Wouldn't using `eval-when-compile` be a better idea here?

akermu: Compiled elisp files should be portable across operating systems and a `eval-when-compile` in this case would violate this condition.

vibhavp: ah, okay

vibhavp:
Looks good, thanks.

PR #216, opened 2017-12-19
lsp--make-hover-callback: Allow `hover` being null

MaskRay:
null is allowed as a response of textDocument/hover, and cquery may emit messages like {"jsonrpc":"2.0","id":93,"result":null}


vibhavp:
Thanks.

PR #217, opened 2017-12-21
Inhibit the window-show-message echo if lsp-inhibit-message is non-nil

pgmcgee:
I'm adding lsp-inhibit-message to inhibit LSP messages from getting echo'd if lsp-inhibit-message is non-nil.

Some language servers, like PHP, are really chatty with messages. The PHP language server sends a message every time a file is parsed. This can really add up on a big project that takes minutes to parse. With this flag, the messages are still sent to the *Messages* buffer, but they aren't echoed to the mode line, which is more sane.

### DIFF
--- a/lsp-notifications.el
+++ b/lsp-notifications.el
@@ -27,13 +27,9 @@
 (defun lsp--window-show-message (params)
   "Send the server's messages to message, inhibit if `lsp-inhibit-message'
 is set."
-  (let* ((prev-inhibit-message inhibit-message))
-    (when lsp-inhibit-message
-      (setq inhibit-message t))
+  (let* ((inhibit-message (or inhibit-message lsp-inhibit-message)))
     (message "%s" (lsp--propertize (gethash "message" params)
-                    (gethash "type" params)))
-    (when lsp-inhibit-message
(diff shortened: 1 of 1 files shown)


lsp-notifications.el
```diff
@@ -19,9 +19,21 @@
...
+  "If non-nil, inhibit the message echo via `inhibit-message'."
+  :type 'boolean
+  :group 'lsp-mode)
+
 (defun lsp--window-show-message (params)
-  (message "%s" (lsp--propertize (gethash "message" params)
-                                 (gethash "type" params))))
+  "Send the server's messages to message, inhibit if `lsp-inhibit-message'
+is set."
+  (let* ((prev-inhibit-message inhibit-message))
```
vibhavp: Why not simply `let` bind `inhibit-message`?

```emacs-lisp
(let ((inhibit-message lsp-inhibit-message))
  (message "%s" (lsp--propertize (gethash "message" params)
				 (gethash "type" params))))
```

pgmcgee: Because I've been doing too much PHP and not enough Lisp and I totally forgot that was possible. Fixed now.

vibhavp:
Looks good, thanks.

PR #218, opened 2017-12-21
lsp--apply-workspace-edits: Add support for TextDocumentEdit

MaskRay:
cquery emits
```json
{"jsonrpc":"2.0","id":160,"result":{"documentChanges":[{"textDocument":{"uri":"file:///tmp/c/a.cc","version":14},"edits":[{"range":{"start":{"line":3,"character":6},"end":{"line":3,"
character":7}},"newText":"b"},{"range":{"start":{"line":4,"character":2},"end":{"line":4,"character":3}},"newText":"b"}]}]}}
```

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -711,10 +711,16 @@ interface Range {
   (inline-quote (plist-get (plist-get ,range :end) :line)))
 
 (defun lsp--apply-workspace-edits (edits)
-  (cl-letf (((lsp--workspace-change-timer-disabled lsp--cur-workspace) t))
-    (maphash (lambda (key value)
-               (lsp--apply-workspace-edit key value))
-             (gethash "changes" edits))))
+  (cl-letf (((lsp--workspace-change-timer-disabled lsp--cur-workspace) t)
+             (fn (lambda (key value)
+                   (lsp--apply-workspace-edit key value))))
+    (if-let (changes (gethash "documentChanges" edits))
+      (seq-do (lambda (td-edit)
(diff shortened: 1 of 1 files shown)


lsp-methods.el
```diff
@@ -711,10 +711,16 @@ interface Range {
   (inline-quote (plist-get (plist-get ,range :end) :line)))
 
 (defun lsp--apply-workspace-edits (edits)
-  (cl-letf (((lsp--workspace-change-timer-disabled lsp--cur-workspace) t))
-    (maphash (lambda (key value)
-               (lsp--apply-workspace-edit key value))
-             (gethash "changes" edits))))
+  (cl-letf (((lsp--workspace-change-timer-disabled lsp--cur-workspace) t)
+             (fn (lambda (key value)
+                   (lsp--apply-workspace-edit key value))))
```
vibhavp: `documentChanges` is actually an array of `TextDocumentEdit` objects, which use `TextEdit` instead of `WorkspaceEdit`, so you would need to use `lsp--apply-text-edits` instead.

vibhavp:
I'll try making a few changes to this to make sure that the file version is correct.

vibhavp:
I'll try making a few changes to this to make sure that the file version is correct.

Plus, `TextDocumentEdit` is applied using `lsp--apply-text-edit` instead of `lsp--apply-workspace-edit`.

MaskRay:
It'd be great if you could come up with a better commit. I can abandon this one at any time.

vibhavp:
https://github.com/emacs-lsp/lsp-mode/pull/222

PR #221, opened 2017-12-23
Use save-match-data to avoid interference with query-replace

amol-mandhane:
Every time I do a "query-replace" and answer "y", that first replacement happens and then the command aborts with:

replace-match-maybe-edit: Match data clobbered by buffer modification hooks.

Wrapping lsp-on-change with save-match-data should fix this issue.


vibhavp:
Looks good, thanks!

PR #222, opened 2017-12-24
Rewrite edits

vibhavp:
Inspired by https://github.com/emacs-lsp/lsp-mode/pull/218

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -720,12 +720,13 @@ interface WorkspaceEdit {
     (if document-changes
       (mapc #'lsp--apply-text-document-edit document-changes)
 
-      (maphash
-        (lambda (uri text-edits)
-          (let ((filename (string-remove-prefix lsp--uri-file-prefix uri)))
-            (with-current-buffer (find-file-noselect filename)
-              (lsp--apply-text-edits text-edits))))
-        changes))))
+      (when (hash-table-p changes)
+        (maphash
+          (lambda (uri text-edits)
+            (let ((filename (string-remove-prefix lsp--uri-file-prefix uri)))
+              (with-current-buffer (find-file-noselect filename)
+                (lsp--apply-text-edits text-edits))))
+          changes)))))
 
 (define-inline lsp--apply-workspace-edits (edits)
   (inline-quote (mapc #'lsp--apply-workspace-edit ,edits)))
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -489,7 +489,7 @@ registered client capabilities by calling
                                     method))))
 
 (defun lsp--workspace-apply-edit-handler (_workspace params)
-  (lsp--apply-workspace-edits (gethash "edit" params)))
+  (lsp--apply-workspace-edit (gethash "edit" params)))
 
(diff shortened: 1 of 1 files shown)


lsp-methods.el
```diff
@@ -707,25 +708,50 @@ interface Range {
...
+interface WorkspaceEdit {
+	changes?: { [uri: string]: TextEdit[]; };
+	documentChanges?: TextDocumentEdit[];
+}"
+  (let ((changes (gethash "changes" edit))
+         (document-changes (gethash "documentChanges" edit)))
+    (if document-changes
+      (mapc #'lsp--apply-text-document-edit document-changes)
+
+      (maphash
```
MaskRay: When documentChanges does not exist and changes is an empty list, `maphash` will error on `nil` input.

MaskRay:
And in line 1483, remove `s`

`    (lsp--apply-workspace-edits edits)))` to `    (lsp--apply-workspace-edit edits)))`

Then it works for cquery

MaskRay:
Thanks!

We should also sort the edits before applying them.

https://github.com/autozimu/LanguageClient-neovim/blob/master/rplugin/python3/LanguageClient/LanguageClient.py#L168 sorts the edits:
```python
    edits = sorted(reversed(edits), key=lambda edit: (
        -1 * edit["range"]["start"]["line"],
        -1 * edit["range"]["start"]["character"],
    )) 
```

Otherwise when there are edits on the same line, the result will be messed up.

lsp-methods.el
```diff
@@ -1481,7 +1481,10 @@ interface RenameParams {
   (let ((edits (lsp--send-request (lsp--make-request
                                    "textDocument/rename"
                                    (lsp--make-document-rename-params newname)))))
-    (lsp--apply-workspace-edits edits)))
+    (if (sequencep edits)
+      (lsp--apply-workspace-edits edits)
```
MaskRay: If I read it correctly, the `sequencep` case is not needed.

https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-1-x.md#workspaceedit
https://github.com/Microsoft/language-server-protocol/blob/master/versions/protocol-2-x.md#workspaceedit
https://microsoft.github.io/language-server-protocol/specification

Thanks! It works perfectly with cquery.

lsp-methods.el
```diff
@@ -751,9 +748,14 @@ interface TextDocumentEdit {
         (lsp--apply-text-edits (gethash "edits" edit))))))
 
 (defun lsp--text-edit-sort-predicate (e1 e2)
```
MaskRay: Comparing both end point and start point addresses one corner case: when one range is empty and the other shares the same end point with it, i.e.  [3, 3) and [1, 3).  We want [3, 3) come first.

Since text edits are guaranteed to be non-overlapping. This can be simplified to compare with their middle point, i.e.
`e1.start + e1.end > e2.start + e2.end` (division by 2 on both sides are omitted).

vibhavp: Since 3 > 1, wouldn't `[3,3)` be sorted before `[1,3)` anyway?

PR #224, opened 2017-12-27
[hover] Display the last MarkedString when there are multiple.

MaskRay:
;; We may receive multiple MarkedString's in the Hover response,
        ;; render the last one because others are likely comments which will
        ;; make the eldoc message span multiple lines.

For comments, @sebastiencs 's lsp-ui-doc.el should be preferred.

See https://github.com/emacs-lsp/lsp-ui/issues/19 for some demos.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1284,7 +1284,7 @@ export interface MarkupContent {
                                  (funcall renderer (gethash "value" e)))
                                (gethash "value" e))
                              e)))
-              (if (listp contents) contents (list contents)) "\n")))))))
+              (if (listp contents) (last contents) (list contents)) "\n")))))))
 
 (defun lsp-provide-marked-string-renderer (client language renderer)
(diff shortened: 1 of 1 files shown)


vibhavp:
Why discard other lines? I think this could be better achieved by let-binding.

MaskRay:
I discard other lines because only the last line is about function/variable/type signature, others are usually documentation. Displaying multiple lines in eldoc is less appealing. Too be more precise, we could filter out all languages other than ones with the language-id the server tells us.

rls
```json
{"jsonrpc":"2.0","id":190,"result":{"contents":[" meow\n",{"language":"rust","value":"fn () -> i32"}],"range":null}}
```

[cquery](https://github.com/jacobdufault/cquery/blob/master/src/messages/text_document_hover.cc#L100)
`"contents":[{"language":"text",...}, {"language":"cpp",...}]`

Issue #214, opened 2017-12-19
Show method signature when writing the arguments of the method


amol-mandhane:
When writing the arguments of the method, textDocument/hover method doesn't apply and there's no documentation available. However, textDocument/signatureHelp is meant exactly for this use-case. LSP client should show the signature when writing the arguments.

signatureHelp method: https://github.com/Microsoft/language-server-protocol/blob/master/protocol.md#textDocument_signatureHelp

amol-mandhane:
I have a local change for this. If the idea sounds okay, I can send a pull request.

vibhavp:
Sure, thanks. I'll add a few `customize` options to your PR and make it work with `SignatureHelpRegistrationOptions.triggerCharacters`, if it isn't already using it.

PR #226, opened 2018-01-02
Create method for textDocument/signatureHelp.

amol-mandhane:
This isn't integrated to lsp--on-hover yet as it can cause conflict between
textDocument/hover and this method. A clean integration has to be designed for
this.

This is the implementation for issue #214.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1289,6 +1289,61 @@ export interface MarkupContent {
   (interactive)
   (lsp--text-document-hover-string))
 
+(defvar-local lsp--current-signature-help-request-id nil)
+
+(defun lsp--text-document-signature-help ()
+  "interface SignatureHelp {
+signatures: SignatureInformation[];
+activeSignature?: number;
+activeParameter?: number;
+};
+
+interface SignatureInformation {
+label: string;
+documentation?: string | MarkupContent;
+parameters?: ParameterInformation[];
+};
+
+interface ParameterInformation {
+label: string;
+documentation?: string | MarkupContent;
+};
+
+interface MarkupContent {
+kind: MarkupKind;
+value: string;
+};
+
+type MarkupKind = 'plaintext' | 'markdown';"
+  (lsp--cur-workspace-check)
+  (when lsp--current-signature-help-request-id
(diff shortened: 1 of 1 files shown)


vibhavp:
Thanks.

Yevgnen:
How can I have this feature when using `lsp-python`? Thanks!

amol-mandhane:
The usage of this is dependent on how #235 goes.

Fuco1:
I'm trying this with the php lsp server but it always returns nil, is there any simple way to debug what is being sent back and forth?

amol-mandhane:
Apart from whatever debug mechanism lsp-mode has (if any), the way I debug is to just add (message ...) calls wherever required.

vyorkin:
I guess the usage is not just as simple as this:
```
    (when lsp-enable-eldoc
      (lsp--text-document-signature-help))
    ; (when lsp-enable-eldoc
    ;   (lsp--text-document-hover-string))
```
Is there any workaround to use this function instead of the `lsp--text-document-hover-string`?
Sorry, I'm new to Emacs Lisp :)

lsp-methods.el
```diff
@@ -1289,6 +1289,61 @@ export interface MarkupContent {
...
+kind: MarkupKind;
+value: string;
+};
+
+type MarkupKind = 'plaintext' | 'markdown';"
+  (lsp--cur-workspace-check)
+  (when lsp--current-signature-help-request-id
+    (lsp--cancel-request lsp--current-signature-help-request-id))
+  (let (bounds body)
+    (when (symbol-at-point)
```
yyoncho: I am testing the method with JDT and it works only if it is invoked within a function invocation. @vibhavp do we need to fix this function or lsp-java should implement it's own version?

vibhavp: Isn't that the intended functionality for signatureHelp?

yyoncho: @vibhavp
Assuming that you have the following code and the cursor is at |

> foo.bar(param1, |);

In this case, the LSP should display signature help for the method bar but since (symbol-at-point) will return false the signature help method will not run at all.

PR #230, opened 2018-01-08
Surround ‘lsp--on-hover’ with ‘with-demoted-errors’.

phst:
Otherwise very common actions such as hovering over a symbol lead to hard
errors if the client is somehow misbehaving.


vibhavp:
Thanks.

PR #236, opened 2018-01-10
truncate response id

topisani:
The [ResponseMessage](https://github.com/Microsoft/language-server-protocol/blob/gh-pages/specification.md#response-message) `id` is either a `number`, a `string` or `null`, but since lsp-mode only sends `numbers`, we can assume it's a number. However, some servers ([cquery](https://github.com/jacobdufault/cquery)) respond with a floating point, even when they got an int on the request. I think this is conforming behaviour, as `number` can be a float, and `10.0 == 10`, but of course emacs hashing doesn't agree.

TL;DR: truncate ids recieved in `ResponseMessage`s before using them as indicies in the hashmap

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -211,6 +211,7 @@ Else it is queued (unless DONT-QUEUE is non-nil)"
     (pcase (lsp--get-message-type json-data)
       ('response
         (cl-assert id)
+        (setq id (truncate id))
         (setq callback (gethash id (lsp--client-response-handlers client) nil))
         (if callback
           (progn (funcall callback (gethash "result" json-data nil))


vibhavp:
What if the fractional part of the request id != 0?

topisani:
Dont we control the ids? i dont believe we ever send any fractional ones

phst:
The server might be misbehaving, and we should then react accordingly without responding to an incorrect request.

How about

```
(let ((int-id (cond ((integerp id) id)
  ((and (floatp id) (eql (ftruncate id) id)) (truncate id))
  (t (signal ...)))
```

vibhavp:
We do, but the server can also send requests to the client, for which it
uses its own IDs.

topisani:
@vibhavp of course, i forgot that part
@phst Yeah, that, and then maybe trying the non-truncated if the truncated didnt give anything. I think that order makes the most sense, since i doubt a lot of servers give floating ids

yyoncho:
As a side note, the Visual Studio Code is sending Strings as IDs - https://github.com/Microsoft/vscode/blob/master/src/vs/workbench/services/extensions/node/rpcProtocol.ts#L182

yyoncho:
Cleanup: closing this PR since it is not linked to issue with a concrete server - please reopen and resolve conflicts if the issue still exists.

PR #237, opened 2018-01-12
Separate symbol highlighting from eldoc.

sebastiencs:
Symbol highlight was depending on eldoc function.
If the user or another package change
eldoc-documentation-function (which lsp-ui-doc do), the symbol
highlighting was lost.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -329,6 +329,12 @@ before saving a document."
   :group 'lsp-mode)
 
 ;;;###autoload
+(defcustom lsp-symbol-highlight-delay eldoc-idle-delay
+  "Seconds of idle time to wait before showing symbol highlight."
+  :type 'number
+  :group 'lsp-mode)
+
+;;;###autoload
 (defface lsp-face-highlight-textual
   '((((background dark))  :background "saddle brown")
     (((background light)) :background "yellow"))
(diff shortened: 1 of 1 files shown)


vibhavp:
Adding `lsp-highlight` to `post-command-hook` is inefficient, as that would call the function even at points when we don't want it to.

IMO, a better way to do this would be to create a new idle timer with `secs` = `eldoc-idle-delay`, so as to sync with eldoc behavior, and make the timer function process symbol highlights.

sebastiencs:
Eldoc itself does the same thing (adding `eldoc-schedule-timer` to `post-command-hook`).
> that would call the function even at points when we don't want it to.

It would call the function as many times as `lsp--on-hover`.
It isn't more inefficient than eldoc is.

> IMO, a better way to do this would be to create a new idle timer with secs = eldoc-idle-delay, so as to sync with eldoc behavior, and make the timer function process symbol highlights.

I'm not sure to understand your point, if you mean create a timer with a non-nil `repeat`, that would call the function even more than with `post-command-hook` (when the point hasn't changed its position)

vibhavp:
> Eldoc itself does the same thing (adding eldoc-schedule-timer to post-command-hook).

Ah, I was under the impression that eldoc used an idle timer to show documentation (which is what flycheck does). Nevermind.

sebastiencs:
I rebased my branch. It should be good

sebastiencs:
@vibhavp Any concern about this PR ?

yyoncho:
@sebastiencs do you still want to merge the PR, if so, can you resolve the conflicts?

MaskRay:
I have an eldoc refactor https://github.com/emacs-lsp/lsp-mode/pull/448

yyoncho:
Closing in favour of #448

Issue #227, opened 2018-01-03
rootUri should be URI escaped


MaskRay:
`lsp-methods.el`:560
```elisp
                         :rootUri ,(concat lsp--uri-file-prefix root)
```

If the filename contains `%`, the server may un-escape it and get the wrong filename. `+` is not rare and  should be escaped (I don't know if some server may treat it as a space).

Found by @jiegec in https://github.com/jacobdufault/cquery/issues/215

PR #238, opened 2018-01-14
Add methods to convert from a URI to a path and vice versa

akermu:
Solves #227

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ install:
   - cask install
 script:
   - cask build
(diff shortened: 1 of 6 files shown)


vibhavp:
Thanks for the PR, could you add a few tests for these functions to `tests/lsp-common-tests.el`?

akermu:
I added tests. Please wait before you merge this, as I am not sure anymore if `url-hexify-string` and `url-unhex-string` are doing the right thing. I'll have to read the RFC.

akermu:
Okay I'm confident that those functions do the right thing. Can you restart the CI please, as it failed to build emacs!

akermu:
Okay travis fails to compile emacs, probably because of some Meltdown/Spectre mitigation..

vibhavp:
Thanks.

PR #239, opened 2018-01-15
Idle timer for lsp-on-change

topisani:
Accumulate the change-event parameters in `lsp-on-change` and limit them with an idle timeout. Also make sure these events are then sent before sending any other requests/notifications. `lsp--send-did-change` does the actual sending if any changes are lined up.

This was sparked over at https://github.com/emacs-lsp/lsp-ui/issues/45, and seems to be the best solution overall, since the didChange events are sent extremely frequently by default. This is a nice place to do performance tuning.

We are still testing it a bit with cquery, but it would be great if someone could try it out with some other servers!

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -215,6 +215,11 @@ initialized. When set this turns off use of
   "Face used for highlighting symbols being written to."
   :group 'lsp-faces)
 
+(defcustom lsp-change-idle-timeout 0.1
+  "Wait till emacs has been idle for this number of seconds before sending `textDocument/didChange'"
+  :type 'number
+  :group 'lsp-mode)
+
 (define-error 'lsp-error "Unkown LSP error")
 (define-error 'lsp-empty-response-error
   "Recived empty response from language server" 'lsp-error)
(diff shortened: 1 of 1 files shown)


alanz:
We used to have this, but the problem is that there is no way to combine multiple changes into a single change message that is meaningful to the LSP server.

alanz:
See https://github.com/emacs-lsp/lsp-mode/issues/112

yyoncho:
@alanz 
> We used to have this, but the problem is that there is no way to combine multiple changes into a single 
change message that is meaningful to the LSP server.

Why do we need to merge them in single document change message? Can we collect the N changes and send N document change notifications without the need to merge them into single one.

alanz:
> Can we collect the N changes and send N document change notifications without the need to merge them into single one.

@yyoncho what is the benefit of this running on a timer? Vs sending them as they happen, as now.

yyoncho:
@alanz Hm, I believe that you are right - it won't benefit a lot(theoretically if the server has delay it will increase the chance to skip rebuilt). I think that the changes can be reavaluated to work as they are performed on  the original document as you have mentioned in #112

yyoncho:
Denying the PR as it is as per https://github.com/emacs-lsp/lsp-mode/pull/239#issuecomment-358957751 .

PR #241, opened 2018-01-16
Correctly queue and dequeue notifications.

### LINKED ISSUE #240: Diagnostics notifications getting queued
I am not sure how it is happening, but by putting some message commands into `lsp--on-notification` it shows that after a while notifications get queued.

```elisp
(defun lsp--on-notification (p notification &optional dont-queue)
  "If response queue is empty, call the appropriate handler for NOTIFICATION.
Else it is queued (unless DONT-QUEUE is non-nil)"
  (let ((params (gethash "params" notification))
        (client (lsp--workspace-client (lsp--parser-workspace p)))
        handler)
    (message "lsp--on-notification called for %s %s" (gethash "method" notification) dont-queue) ;;; AZ-DEBUG
    ;; If we've been explicitly told to queue
    (if (and (not dont-queue) (lsp--parser-response-result p))
        (progn
          (message "lsp--on-notification: queuing notification") ;; AZ-DEBUG
          (push (lsp--parser-queued-notifications p) notification))
      ;; else, call the appropriate handler
      (pcase (gethash "method" notification)
```

Example log

```
lsp--on-notification called for textDocument/publishDiagnostics nil
lsp--on-diagnostics called
lsp--on-notification called for textDocument/publishDiagnostics nil
lsp--on-diagnostics called
Mark saved where search started
lsp--on-notification called for textDocument/publishDiagnostics nil
lsp--on-notification: queuing notification
lsp--on-notification called for textDocument/publishDiagnostics nil
lsp--on-notification: queuing notification
```

Once it starts all subsequent notifications get queued.  This behaviour only seems to happen in files with a lot of notifications.

So

 - Why does it happen? My reading of the source seems to indicate that it pretty much should never happen
 - Why is `lsp--flush-notifications` never called anywhere?

And if queuing should happen, where should the flush call be placed? At the end of `lsp--send-request `?

alanz:
Closes #240

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -102,7 +102,7 @@ Else it is queued (unless DONT-QUEUE is non-nil)"
         handler)
     ;; If we've been explicitly told to queue
     (if (and (not dont-queue) (lsp--parser-response-result p))
-        (push (lsp--parser-queued-notifications p) notification)
+        (push notification (lsp--parser-queued-notifications p))
       ;; else, call the appropriate handler
       (pcase (gethash "method" notification)
         ("window/showMessage" (lsp--window-show-message params))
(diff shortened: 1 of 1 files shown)


alanz:
This works, but it is possible that `lsp--flush-notifications` can be in a more optimal place.

But it seems to make sense to me to be where it is.

vibhavp:
Alternatively, why queue notifications in the first place? If the server sends a notification before responding to a request, it probably means that it wants the notification to be handled before.

alanz:
I do not mind either way, so long as they are not lost.

And without this, once things start queuing they are always queued, until the queue is flushed.

vibhavp:
Replaced by https://github.com/emacs-lsp/lsp-mode/commit/9de9d6e316e56b578ef2f5b427337bb0fef97c88

PR #244, opened 2018-01-19
Notification `params` are `any` and not `null`

bkchr:


vibhavp:
Thanks, but that isn't how message arguments work. If you take a look at https://microsoft.github.io/language-server-protocol/specification#initialized, the final JSON message should look like this:
```json
{  
   "jsonrpc":"2.0",
   "method":"initialized",
   "params":null
}
```

With your PR, it would be
```json
{  
   "jsonrpc":"2.0",
   "method":"initialized",
   "params":{  
      "InitializeParams":null
   }
}
```

bkchr:
The current implementation does not work anymore with the rust language server :( With my pull request it works. I don't know if the error is on the side of `lsp-mode` or `rls`..

sebastiencs:
I think the message should be:
```json
{  
   "jsonrpc":"2.0",
   "method":"initialized",
   "params": {}
}
```

bkchr:
Perfect @sebastiencs, Json is always so complicated ^^ Tomorrow, I can update the pull request with your proposed solution.

bkchr:
Fixed.

lsp-methods.el
```diff
@@ -246,7 +246,7 @@ initialized. When set this turns off use of
   "Create notification body for method METHOD and parameters PARAMS."
   (inline-quote
     (progn (cl-check-type ,method string)
-      (list :jsonrpc "2.0" :method ,method :params ,params))))
+      (list :jsonrpc "2.0" :method ,method :params ,(or params (make-hash-table))))))
```
vibhavp: `any` can also mean `null`, right?

lsp-methods.el
```diff
@@ -571,7 +571,7 @@ directory."
             (gethash "capabilities" response))
       ;; Version 3.0 now sends an "initialized" notification to allow registration
       ;; of server capabilities
-      (lsp--send-notification (lsp--make-notification "initialized" nil))
+      (lsp--send-notification (lsp--make-notification "initialized"))
```
vibhavp: Why not just have a single line change of the form
```
(lsp--send-notification (lsp--make-notification "initialized" (make-hash-table)))
```

lsp-methods.el
vibhavp: parameters to a notification _can_ be null in some cases. See https://microsoft.github.io/language-server-protocol/specification#exit

lsp-methods.el
vibhavp: IMO, having a single line change as 
```
(lsp--send-notification (lsp--make-notification "initialized" (make-hash-table)))
```
would be better.

sebastiencs:
@vibhavp from the protocol, void is not null:
void: `{}`
null: `null`

sebastiencs:
Well, after reading the typescript [documentation](https://www.typescriptlang.org/docs/handbook/basic-types.html#void) for void, it seems that void can be null.
Not sure why they'are using 2 differents terms (`void` and `null`) in the LSP. I don't find it pretty clear

vibhavp:
`any`, `null` and `void` are type specifiers in typescript: https://www.typescriptlang.org/docs/handbook/basic-types.html'

A variable with type `any` can be set to null: https://www.typescriptlang.org/play/#src=let%20a%3A%20any%20%3D%20null%3B%0D%0Aconsole.log(JSON.stringify(a))%3B

sebastiencs:
Yep, I was wrong.
Can you update the PR @bkchr?

vibhavp:
Should I do this myself? It's a single-line change.

bkchr:
@vibhavp yeah do it, I close the pr.

PR #248, opened 2018-01-21
Disable query-on-exit for LSP processes

gpittarelli:
When closing emacs with multiple lsp-mode buffers open, I get prompted to close the active LSP servers.

image

I can't imagine when a user would ever want to answer no.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -38,15 +38,18 @@
            (final-command (if (consp command) command (list command))))
       (unless (executable-find (nth 0 final-command))
         (error (format "Couldn't find executable %s" (nth 0 final-command))))
-      (make-process
-       :name name
-       :connection-type 'pipe
-       :coding 'no-conversion
-       :command final-command
-       :filter filter
-       :sentinel sentinel
-       :stderr stderr
-       :noquery t))))
+      (let ((proc (make-process
+                    :name name
+                    :connection-type 'pipe
+                    :coding 'no-conversion
+                    :command final-command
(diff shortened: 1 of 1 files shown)


vibhavp:
Looks good, thanks.

lsp-mode.el
```diff
@@ -38,15 +38,18 @@
...
+      (let ((proc (make-process
+                    :name name
+                    :connection-type 'pipe
+                    :coding 'no-conversion
+                    :command final-command
+                    :filter filter
+                    :sentinel sentinel
+                    :stderr stderr
+                    :noquery t)))
+        (set-process-query-on-exit-flag proc nil)
```
phst: This line is not necessary because `:noquery t` above does the same.

lsp-mode.el
```diff
@@ -38,15 +38,18 @@
...
+        (set-process-query-on-exit-flag (get-buffer-process (get-buffer stderr)) nil)
```
phst: This line is only necessary because of https://debbugs.gnu.org/cgi/bugreport.cgi?bug=30031 (fixed in Emacs 26). It's fine to add it as workaround, but please add a TODO to remove it once `lsp-mode` has stopped supporting Emacs 25 and below.

gpittarelli: Ah, that explains it, thanks. I made another PR to fix these comments https://github.com/emacs-lsp/lsp-mode/pull/249

phst: Thanks!

PR #250, opened 2018-01-21
Don't message on request cancelled

gpittarelli:
Some LSP servers (for example, https://github.com/sourcegraph/javascript-typescript-langserver) send back errors when the client cancels a previous request (`'$/cancelRequest'`). Currently these get treated like all other errors and are `message`d to the user, even though they aren't really an error. This PR silences that message.

image

The LSP spec has a standard error code for this: https://github.com/Microsoft/language-server-protocol/commit/f3ef9bf65e5545467366f3568e2b1e7f8b400cee
which matches what I'm actually seeing: `{"jsonrpc":"2.0","id":9,"error":{"message":"Request cancelled","code":-32800}}`

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -156,7 +156,14 @@
     (-32602 "Invalid Parameters")
     (-32603 "Internal Error")
     (-32099 "Server Start Error")
-    (-32000 "Server End Error")))
+    (-32000 "Server End Error")
+    (-32002 "Server Not Initialized")
+    (-32001 "Unknown Error Code")
+    (-32800 "Request Cancelled"))
+  "alist of error codes to user friendly strings.")
+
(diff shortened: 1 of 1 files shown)


lsp-io.el
```diff
@@ -156,7 +156,14 @@
...
     (-32603 "Internal Error")
     (-32099 "Server Start Error")
-    (-32000 "Server End Error")))
+    (-32000 "Server End Error")
+    (-32002 "Server Not Initialized")
+    (-32001 "Unknown Error Code")
+    (-32800 "Request Cancelled"))
+  "alist of error codes to user friendly strings.")
+
+(defconst lsp--silent-errors '(-32800)
```
vibhavp: This could later be changed to a `defcustom`, to allow easier customization.

PR #251, opened 2018-01-22
Apply ignore regexps after unchunking server messages

gpittarelli:
LSP servers occassionaly split up a single message into multiple chunks: (I've observed both of these situations happening with https://github.com/sourcegraph/javascript-typescript-langserver)

 1. The `Content-Length` header and the actual body of the message are sent in separate `write`s
 2. Very long message bodies get split across multiple `write`s

Both of these occasionally break when a chunk matches one of the `ignore-regexps`: each chunk is tested individually, so when 1 chunk is ignored, the overall message gets corrupted. (We get left with a `Content-Length` but no body in case 1, or a corrupt body that doesn't fill the stated `Content-Length` in case 2). We could also fail to ignore a message if the text matching an ignore-regexp gets split across chunk boundaries. 

(What is great is that `lsp-mode` eventually recovered well after these errors, aside from functionality temporarily breaking and/or showing error messages to the user)

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -281,23 +281,22 @@
 (defun lsp--parser-make-filter (p ignore-regexps)
   #'(lambda (proc output)
       (setq lsp--no-response nil)
-      (when (cl-loop for r in ignore-regexps
-                     ;; check if the output is to be ignored or not
-                     ;; TODO: Would this ever result in false positives?
-                     when (string-match r output) return nil
-                     finally return t)
-        (let ((messages
-               (condition-case err
-                   (lsp--parser-read p output)
-                 (error
-                  (progn
-                    (lsp--parser-reset p)
-                    (setf (lsp--parser-response-result p) nil
-                          (lsp--parser-waiting-for-response p) nil)
-                    (error "Error parsing language server output: %s" err))))))
-
-          (dolist (m messages)
-            (when lsp-print-io (message "Output from language server: %s" m))
+      (let ((messages
+             (condition-case err
+                 (lsp--parser-read p output)
+               (error
+                (progn
+                  (lsp--parser-reset p)
+                  (setf (lsp--parser-response-result p) nil
+                        (lsp--parser-waiting-for-response p) nil)
+                  (error "Error parsing language server output: %s" err))))))
+        (dolist (m messages)
+          (when lsp-print-io (message "Output from language server: %s" m))
+          (when (cl-loop for r in ignore-regexps
+                         ;; check if the output is to be ignored or not
+                         ;; TODO: Would this ever result in false positives?
+                         when (string-match r m) return nil
+                         finally return t)
             (lsp--parser-on-message p m))))
       (when (lsp--parser-waiting-for-response p)
         (with-local-quit (accept-process-output proc)))))
--- a/test/lsp-io-tests.el
+++ b/test/lsp-io-tests.el
@@ -58,18 +58,34 @@
 		(should (equal (lsp--parser-read p "\"somedata\":2}")
 									 '("{\"somedata\":2}")))))
 
+(defun random-string-chunks (str)
+  "Randomly split a string 'abcdefghi' into chunks of random lengths:
+(\"ab\" \"cdefgh\" \"i\"). Slightly biased towards having longer
+chunks earlier in the list."
+  (if (string= str "") '()
+    (let* ((n (length str))
+           (i (1+ (random n))))
+      (cons (substring str 0 i)
+            (random-string-chunks (substring str i n))))))
+
+(defconst lsp--example-ignore-msg
+  (concat
+   "Content-Length: 2\r\n\r\n{}"
+   "Content-Length: 15\r\n\r\n{\"IGNORE_ME\":1}"
+   "Content-Length: 2\r\n\r\n{}"))
+
 (ert-deftest lsp--parser-read--multiple-chunks-ignored ()
-  (let* ((log '())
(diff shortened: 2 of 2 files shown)


test/lsp-io-tests.el
```diff
@@ -58,5 +58,18 @@
 		(should (equal (lsp--parser-read p "\"somedata\":2}")
 									 '("{\"somedata\":2}")))))
 
+(ert-deftest lsp--parser-read--multiple-chunks-ignored ()
```
phst: Thanks for adding this test! Would you mind adding more tests, e.g. with the `Content-Length` header itself split in multiple chunks, or the `\r\n` in two chunks?

test/lsp-io-tests.el
```diff
@@ -58,5 +58,18 @@
...
+  (let* ((log '())
+         (p (make-lsp--parser :workspace lsp--test-workspace))
+         (lsp--parser-reset p)
+         (filter (lsp--parser-make-filter p '("IGNORE_ME"))))
+    (cl-letf (((symbol-function 'lsp--parser-on-message)
```
phst: `cl-letf*` is a generalization of `let*`, so you can use a single `cl-letf*` form here.

test/lsp-io-tests.el
```diff
@@ -58,5 +58,18 @@
...
+               (lambda (p msg) (push msg log))))
```
phst: This is totally fine, but you could also use a mock framework such as `el-mock`:

```elisp
(with-mock
  (mock (lsp--parser-on-message * "{}") :times 2)
   ...)
```

lsp-io.el
```diff
@@ -281,23 +281,22 @@
...
-            (when lsp-print-io (message "Output from language server: %s" m))
+      (let ((messages
+             (condition-case err
+                 (lsp--parser-read p output)
+               (error
+                (progn
+                  (lsp--parser-reset p)
+                  (setf (lsp--parser-response-result p) nil
+                        (lsp--parser-waiting-for-response p) nil)
+                  (error "Error parsing language server output: %s" err))))))
```
phst: Somewhat unrelated: catching and re-raising errors destroys the backtrace. It's better to write:

```elisp
(let ((sucess nil))
  (unwind-protect
     (progn
       ...
       (setq success t))
    (unless success
      ...)))
```

vibhavp:
The intent behind adding `ignore-regexps` was to work with language servers that would incorrectly print initial log messages to stdout, which would mix up with LSP data and confuse the parser.

The main parser loop is in `lsp--parser-read`, which only un-chunks text if it detects a `\r\n\r\n` separator, after which it will treat the gathered output as a list of headers. From what I can observe in this PR, non LSP data will never be unchunked, but be concatenated to `(lsp--parser-leftovers p)`, which will result in an error while parsing the header string.

Ignoring LSP messages should be handled in `lsp--parser-on-message`, or an appropriate request/response/notification handler callback. [lsp-go](https://github.com/emacs-lsp/lsp-go/blob/d270b7c0bd893fdc73a97763e07706830756cad2/lsp-go.el#L17) is an example of how `ignore-regexps` is used.

gpittarelli:
Okay, I understand we should leave `ignore-regexps` as-is. Unfortunately setting a handler won't quite work yet for this case, since we're dealing with `window/logMessage` messages, and [`lsp--on-notification`](https://github.com/gpittarelli/lsp-mode/blob/apply-ignore-regexps-on-full-messages/lsp-io.el#L102) has a special case for that before delegating to any language-specific handlers.

I'll make 2 PRs: (if you approve of this direction)
 1. Change `lsp--on-notification` to check the per-language `lsp--client-notification-handlers` hash _before_ the default impls
 2. Switch lsp-javascript to set a handler instead of using `ignore-regexp`

---
Details:

I ran into this when using lsp-javascript, which has a similar `ignore-regexp` https://github.com/emacs-lsp/lsp-javascript/blob/456854bdee8729c21331f93ee9054f2d980fe2ab/lsp-javascript-typescript.el#L38

However, that `ignore-regexp` is trying to match full messages looking like:
```
{
  "jsonrpc": "2.0",
  "method": "window/logMessage",
  "params": {
    "type": 2,
    "message": "readFile /some/path requested by TypeScript but content not available"
  }
}
```
But the language server is, for reasons unknown, sending the headers and that body in separate `write` syscalls (verified via strace). With `lsp-print-io=t` I saw a bunch of lines like this (with nothing in between, the bodies having been blocked by `ignore-regexp`)
```
Output from language server: Content-Length: 123^M

Output from language server: Content-Length: 123^M

Output from language server: Content-Length: 123^M

```

This was from the strace of the language server:
```
write(1, "Content-Length: 240\r\n\r\n", 23) = 23
write(1, "{\"jsonrpc\":\"2.0\",\"method\":\"window/logMessage\",\"params\":{\"type\":2,\"message\":\"readFile /some/path requested by TypeScript but content not available\"}}", 240) = 240
```

vibhavp:
> Change lsp--on-notification to check the per-language lsp--client-notification-handlers hash before the default impls

LGTM. Although a quicker way to accomplish this would be to add a `defcustom` variable that stores regexp(s) of `logMessage` messages that can be ignored/silenced.

> Switch lsp-javascript to set a handler instead of using ignore-regexp

Yes, this is not how `ignore-regexp` was supposed to be used.

gpittarelli:
Closing in favor of the issues raised being addressed in https://github.com/emacs-lsp/lsp-mode/pull/267 + https://github.com/emacs-lsp/lsp-mode/pull/259

Issue #255, opened 2018-01-23
Alternate lsp-project-whitelist/lsp-project-blacklist logic


MaskRay:
First, `lsp-project-whitelist lsp-project-blacklist` should be a list of regular expressions / glob patterns.

The logic to test whether a project should be initialized:

1. If in `lsp-project-whitelist`, pass
2. Else if in `lsp-project-blacklist`, fail
3. Else pass

Moreover, in `lsp-define-stdio-client`, 

```elisp
             ,(when initialize
                `(funcall ,initialize client))
             (let ((root (funcall (lsp--client-get-root client))))
               (if (lsp--should-start-p root)
                   (progn
                     (lsp-mode 1)
                     (lsp--start client ,extra-init-params))
                 (message "Not initializing project %s" root)))))))))
```

The `initialize` function should be able to tell if it is able to initialize the project. It may have some other settings.

MaskRay:
Fixed by #256

PR #256, opened 2018-01-24
Change project whitelist/blacklist checking logic

MaskRay:
`lsp-mode` will only be started if the given project root matches one pattern
in the whitelist, or does not match any pattern in the blacklist.

This is more flexible than the original "either whitelist or blacklist".

With this, people can have blanket `^/tmp/.*$` blacklist with a few exemptions

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,12 +104,11 @@ to your init file to enable imenu integration.
 In order to more finely control the `lsp-mode` startup, there are a number of
 customizable variables.
 
-`lsp-project-whitelist` : Defaults to `nil`. If set, `lsp-mode` will only be
-started if the given project root appears in the whitelist.
+`lsp-project-whitelist` : Defaults to `nil`.
+`lsp-project-blacklist` : Defaults to `nil`.
 
-`lsp-project-blacklist` : Defaults to `nil`. If set, all projects will be
-started except those in this list. It is ignored if `lsp-project-whitelist` is
-set.
+`lsp-mode` will only be started if the given project root matches one pattern
+in the whitelist, or does not match any pattern in the blacklist.
 
 There are also the functions `lsp-MAJOR-MODE-whitelist-add` and
 `lsp-MAJOR-MODE-whitelist-remove` to adjust the current buffer project root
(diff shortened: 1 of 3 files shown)


MaskRay:
This should fix https://github.com/emacs-lsp/lsp-mode/issues/255 

@alanz

alanz:
So this is basically just generalising what was already there to use regexps.  Sounds fine.

MaskRay:
There is also  stop for all except (WL + BL), which cannot be captured by prioritizing BL over WL.

alanz:
@MaskRay please give me an example of that, as I do not understand the case.

Stop for all except WL makes sense, what does the (+BL) mean in this case? That if it is in both the WL and BL it should not start?

MaskRay:
e.g. blacklist all projects in `/tmp` with an exception `/tmp/a`.

This cannot be caught if the blacklist takes precedence over the whitelist.

A second whitelist that has lower precedence seems viable.

Actually I don't think there should be two variables whitelist and blacklist, one like `auto-mode-alist` is sufficient. The position in the list decides the precedence.

alanz:
To work with your example, we would have WL = `/tmp/a`, and BL = `/tmp/*`.

What happens if I try to open `~/foo`?  In my scheme it would need
  WL = `/tmp/a, ~/foo`, and BL = `/tmp/*`

And effectively it means anything you need to open needs to be in WL, so the value in the BL becomes irrelevant.

And I do not understand the reference to `auto-mode-alist`, which gives a list of file extensions and the mode to apply to them.

MaskRay:
> Now we have the weird situation where the only start for case cannot be captured.

It can be captured. whitelist = ["^/tmp/a"], blacklist = [".*"]

Here is a mapping:

Old scheme: whitelist = [], blacklist = as
New scheme: whitelist = [], blacklist = map regexQuote as

Old scheme: whitelist = as, blacklist = []  where not (null as); the blacklist is irrelevant (ignored) here
New scheme: whitelist = as, blacklist = [".*"]

For some logic like:
whitelist = ["^/tmp/"], blacklist = ["^/tmp/a"]

cannot be captured in the old scheme

alanz:
> It can be captured. whitelist = ["^/tmp/a"], blacklist = [".*"]

I forgot that these are now regexes. And that now makes sense to me.  Sorry for being obtuse.

But perhaps we should put this in the docstring?

PR #257, opened 2018-01-24
Set projectile-project-root

MaskRay:
This is useful for projects extracted from an archive or others that do not have .git .hg .svn project root indicators.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -529,6 +529,7 @@ directory."
          (workspace (gethash root lsp--workspaces))
          new-conn response init-params
          parser proc cmd-proc)
+    (setq-local projectile-project-root root)
     (if workspace
         (setq lsp--cur-workspace workspace)
 


vibhavp:
Wouldn't this add a soft dependency on projectile? IMO, this should be done in a seperate package (`lsp-ui`, perhaps) or be a feature in projectile itself.

MaskRay:
It doesn't add a dependency on projectile. This just sets a buffer-local variable which will be picked by projectile. The variable `projectile-project-root` is just used by projectile to detect the root.

vibhavp:
Merging for now. We should probably move this to `lsp-ui` later. Ideally, this should be a feature in `projectile` itself.

jaelsasser:
Could we at least gate this behind a `custom-set-variable` in the interim? I've got one git repo with multiple independent (but related) C projects under it, each with their own `compile_commands.json` under it, and this PR caught me completely off guard.

MaskRay:
@jaelsasser Do you mean: you want to

* run your C++ language server for a sub projects
* and in the meanwhile, on all the projects as a whole

? 

I git clone llvm projects and read the code with cquery. I have only a compile_commands.json in the project root, not in sub projects.

jaelsasser:
@MaskRay The mono-repo I use at work looks like this:

```
project/
       cpp/
          src/
              libA/compile_commands.json
              projectA/compile_commands.json
              projectB/compile_commands.json
       c/
         src/
             libB/compile_commands.json
             projectC/compile_commands.json
             projectD/compile_commands.json
```

Each of those `compile_commands.json` runs their own language server. It'd be nice to cut that down, but I haven't gotten around to integrating overall compile_command generation with the (rather sprawling and fragile) build system. In the meantime, though, this PR means I can't do a one-step `projectile-find-file` to jump between files in `libB` and `projectD` anymore.

MaskRay:
you can always change the projectile in an after hook. i don't see a need
to have another variable to customize the one line change

vibhavp:
A `lsp-set-projectile-root` customize variable wouldn't hurt, though.

sebastiencs:
I don't think lsp-mode or lsp-ui should set a projectile variable, it should be managed by projectile, or maybe projectile-lsp

MaskRay:
I should have thought this again and this seems unnecessary. Feel free to revert it.

sudeepdino008:
Hey. What's the status of this? Is the conclusion that projectile should allow some way to set custom roots, and so we should raise the request with projectile if this is needed?

yyoncho:
@sudeepdino008 projectile support setting custom project root via `projectile-add-known-project`. The recommended flow is to use `I` from the import project dialog to manually pick lsp-mode project root when the default suggestion does not match the real one.

Issue #253, opened 2018-01-22
lsp-format-buffer doesn't preserve cursor position


casouri:
I'm using ```lsp-mode``` and ```lsp-python```. If ```lsp-format-buffer``` modifies anything the cursor jumps to the beginning of the buffer. 
I don't know whether the problem is due to ```python-language-server``` or ```lsp-mode```. But probably ```lsp-mode```
If  ```lsp-mode``` can be confirmed as the cause, is it expected to happened and requires customization or there can be a fix?

bkchr:
I have the same problem with `lsp-rust`.

sebastiencs:
#258 should fix that

casouri:
Great! Thanks.

PR #258, opened 2018-01-26
Fix format-buffer #253

sebastiencs:
While applying text changes to the buffer, the cursor position can
be lost with delete-region. See:
https://emacs.stackexchange.com/a/7577/15635

So we apply the changes in a temporary buffer and replace the content
of the current buffer with it using the function `replace-buffer-contents` which handle very well the modifications (cursor position, undo, ...).

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1422,16 +1422,22 @@ If title is nil, return the name for the command handler."
   (interactive)
   (let ((edits (lsp--send-request (lsp--make-request
                                    "textDocument/formatting"
-                                   (lsp--make-document-formatting-params))))
-        (content (buffer-string))
-        (buffer (get-buffer-create " *lsp-formatting*")))
-    (replace-buffer-contents
-     (with-current-buffer buffer
-       (erase-buffer)
-       (insert content)
-       (lsp--apply-text-edits edits)
-       (current-buffer)))
-    (kill-buffer buffer)))
+                                   (lsp--make-document-formatting-params)))))
+    (if (fboundp 'replace-buffer-contents)
+        (let ((content (buffer-string))
+              (buffer (get-buffer-create " *lsp-formatting*")))
+          (unwind-protect
+              (replace-buffer-contents
+               (with-current-buffer buffer
+                 (erase-buffer)
+                 (insert content)
+                 (lsp--apply-text-edits edits)
+                 (current-buffer)))
+            (kill-buffer buffer)))
+      (let ((point (point)))
+        (lsp--apply-text-edits edits)
+        (goto-char point)
+        (goto-char (line-beginning-position))))))
 
 (defun lsp--make-document-range-formatting-params (start end)
   "Make DocumentRangeFormattingParams for selected region.
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1421,9 +1421,17 @@ If title is nil, return the name for the command handler."
   "Ask the server to format this document."
   (interactive)
   (let ((edits (lsp--send-request (lsp--make-request
-                                    "textDocument/formatting"
-                                    (lsp--make-document-formatting-params)))))
-    (save-excursion (lsp--apply-text-edits edits))))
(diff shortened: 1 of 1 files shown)


bkchr:
@sebastiencs I get the following error: `let: Symbol’s function definition is void: replace-buffer-contents [2 times]` :(

sebastiencs:
@bkchr Oh it seems to be a new function in emacs 26.. :(
I will provide an alternative

sebastiencs:
@bkchr I didn't find a way to keep the cursor at the same position without `replace-buffer-contents` which seems to have been introduced in emacs 26 for this purpose:

From the elisp doc
> One potential case where this behavior is useful is external code formatting programs: they typically write the reformatted text into a temporary buffer or file, and using ‘delete-region’ and ‘insert-buffer-substring’ would destroy these properties.

It will try to move to the same line, but it's not exact

bkchr:
Your implementation is almost perfect. I made one small addition, I save the `window-start`. Now, it feels really nice :) Could you apply my changes to your commit?

```
diff --git a/lsp-methods.el b/lsp-methods.el
index 2f0b6b3..41a65e5 100644
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1434,10 +1434,11 @@ If title is nil, return the name for the command handler."
                  (lsp--apply-text-edits edits)
                  (current-buffer)))
             (kill-buffer buffer)))
-      (let ((point (point)))
+      (let ((point (point)) (wstart (window-start)))
         (lsp--apply-text-edits edits)
         (goto-char point)
-        (goto-char (line-beginning-position))))))
+        (goto-char (line-beginning-position))
+       (set-window-start (selected-window) wstart)))))
 
 (defun lsp--make-document-range-formatting-params (start end)
   "Make DocumentRangeFormattingParams for selected region.
```

sebastiencs:
Alright, done

lsp-methods.el
```diff
@@ -1421,9 +1421,25 @@ If title is nil, return the name for the command handler."
...
   (interactive)
   (let ((edits (lsp--send-request (lsp--make-request
-                                    "textDocument/formatting"
-                                    (lsp--make-document-formatting-params)))))
-    (save-excursion (lsp--apply-text-edits edits))))
+                                   "textDocument/formatting"
+                                   (lsp--make-document-formatting-params)))))
+    (if (fboundp 'replace-buffer-contents)
+        (let ((content (buffer-string))
+              (buffer (get-buffer-create " *lsp-formatting*")))
```
vibhavp: Why not use `generate-new-buffer` to get a unique buffer name?

phst: Or just `with-temp-buffer`, then you can also remove the `unwind-protect`.

lsp-methods.el
```diff
@@ -1421,9 +1421,25 @@ If title is nil, return the name for the command handler."
   "Ask the server to format this document."
   (interactive)
   (let ((edits (lsp--send-request (lsp--make-request
-                                    "textDocument/formatting"
-                                    (lsp--make-document-formatting-params)))))
-    (save-excursion (lsp--apply-text-edits edits))))
+                                   "textDocument/formatting"
+                                   (lsp--make-document-formatting-params)))))
+    (if (fboundp 'replace-buffer-contents)
+        (let ((content (buffer-string))
```
phst: Prefer `buffer-substring-no-properties` to avoid copying the properties around.

Even better, use `insert-buffer-substring-no-properties` below to avoid creating a temporary string.

lsp-methods.el
sebastiencs: @phst `with-temp-buffer` cannot be used. It would be deleted in the `replace-buffer-contents` call

lsp-methods.el
sebastiencs: @phst Yes `insert-buffer-substring-no-properties` seems better. And thanks for your work on `replace-buffer-contents` :)

vibhavp:
Looks good, thanks.

PR #259, opened 2018-01-26
Allow overriding any notification handlers

gpittarelli:
Follow up from https://github.com/emacs-lsp/lsp-mode/pull/251. `lsp--on-notification` is rewritten to check per-client notification handlers before the builtin handlers, so that language clients can override `"window/logMessage"`, etc.

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -99,23 +99,27 @@
       'notification
       (signal 'lsp-unknown-message-type (list json-data)))))
 
+(defconst lsp--default-notification-handlers
+  #s(hash-table
+     test eq
+     data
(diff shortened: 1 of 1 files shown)


lsp-io.el
```diff
@@ -99,23 +99,27 @@
       'notification
       (signal 'lsp-unknown-message-type (list json-data)))))
 
+(defconst lsp--default-notification-handlers
+  #s(hash-table
+     test eq
```
vibhavp: Should the test here be `equal`? `eq` only checks for reference equality, not object/structural equality:

```emacs-lisp
(setq h (make-hash-table :test 'eq))
(puthash "foo" 'bar h)
(gethash "foo" h)
```

evaluates to `nil`.

PR #260, opened 2018-01-28
Add include-declaration to lsp--make-reference-params

MaskRay:


lsp-methods.el
```diff
@@ -1179,12 +1179,12 @@ Returns xref-item(s)."
     ;; textDocument/definition returns Location | Location[]
     (lsp--locations-to-xref-items (if (listp defs) defs (list defs)))))
 
-(defun lsp--make-reference-params (&optional td-position)
+(defun lsp--make-reference-params (&optional td-position include-declaration)
```
vibhavp: Could you the docstring excerpt for the new parameter?

MaskRay: Done

Issue #261, opened 2018-01-28
lsp--uri-to-path is not permissive enough (minor bug)


esbena:
`lsp-ui-peek-find-defintions` (and similar functions, I guess) do not work correctly when the LSP server emits URIs of the form `"file:/home/user/foo.java"`, an error about a non-existing file is raised.

## Core problem/solution

`lsp--uri-to-path` is not permissive enough, it should remove other more prefixes than just `"file://"`.
In particular it should also support removing the `"file:"` prefix of `"file:/home/user/foo.java"`.

(Ideally the implementation should use a URI parsing library.)

## Techicalities

According to section ["3. Syntax Components" of RFC3986](https://tools.ietf.org/html/rfc3986#section-3), the part of the URI prior to the authority does not have to be of the form `"scheme://"`, it could for instance also be of the form `"scheme:"`.

In particular `"file:///foo"` and `"file:/foo"` resolve to the same file (namely `"/foo"`), as evidenced by the [vscode-uri](https://github.com/Microsoft/vscode-uri) module, mentioned as the reference URI parser in the [LSP Specification](https://microsoft.github.io/language-server-protocol/specification#uri).

## Workaround

Use `(setq lsp--uri-file-prefix "file:")`when interfacing with an LSP server that emits path-URIs of the form `"file:/foo"`.

(Thank you for your amazing work. I hope to contribute back some day)

PR #265, opened 2018-02-01
Use url-parse to extract filename #261

sebastiencs:
Issue #261

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -75,10 +75,16 @@ If no such directory could be found, log a warning and return `default-directory
 (defun lsp--uri-to-path (uri)
   "Convert URI to a file path."
   (let* ((url (url-generic-parse-url (url-unhex-string uri)))
-         (type (url-type url)))
+         (type (url-type url))
+         (file (url-filename url)))
     (when (and type (not (string= type "file")))
       (error "Unsupported file scheme: %s" uri))
-    (url-filename url)))
+    ;; `url-generic-parse-url' is buggy on windows:
+    ;; https://github.com/emacs-lsp/lsp-mode/pull/265
+    (or (and (eq system-type 'window-nt)
+             (eq (elt file 0) ?\/)
+             (substring file 1))
+        file)))
 
 (defun lsp--path-to-uri (path)
   "Convert PATH to a uri."
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -17,6 +17,7 @@
 
 (require 'compile)
 (require 'url-util)
+(require 'url-parse)
 (require 'subr-x)
 
 (defconst lsp--message-type-face
@@ -73,7 +74,11 @@ If no such directory could be found, log a warning and return `default-directory
 
 (defun lsp--uri-to-path (uri)
   "Convert URI to a file path."
-  (string-remove-prefix lsp--uri-file-prefix (url-unhex-string uri)))
+  (let* ((url (url-generic-parse-url (url-unhex-string uri)))
+         (type (url-type url)))
+    (when (and type (not (string= type "file")))
+      (error "Unsupported file scheme: %s" uri))
+    (url-filename url)))
 
 (defun lsp--path-to-uri (path)
   "Convert PATH to a uri."
--- a/test/lsp-common-test.el
+++ b/test/lsp-common-test.el
@@ -27,7 +27,8 @@
     (should (equal (lsp--path-to-uri "/root/file/hallo welt") "file:///root/file/hallo%20welt"))))
 
 (ert-deftest lsp--path-to-uri ()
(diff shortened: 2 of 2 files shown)


vibhavp:
Could you look into why this fails the tests?

sebastiencs:
I think the tests are wrong:
```el
(should (equal (lsp--path-to-uri "c:/Users/?/") "file:///c:/Users/%3F/"))
(should (equal (lsp--uri-to-path "file:///c:/Users/%7B%7D/") "c:/Users/{}/"))
```
[link](https://github.com/emacs-lsp/lsp-mode/blob/master/test/lsp-common-test.el#L25)

Is `file:///c:/path` a valid uri ? @esbena @akermu

akermu:
Yes it is, see https://blogs.msdn.microsoft.com/ie/2006/12/06/file-uris-in-windows/.

I think `url-generic-parse-url` is not a *URI*-parser.

akermu:
Okay, reading the info nodes, the function is described as an URI-parser, so I would consider this an emacs bug in `url-generic-parse-url`.

esbena:
For reference, here are the URI objects from the two tests, when the strings are parsed by the LSP reference URI converter: https://github.com/Microsoft/vscode-uri.

`(should (equal (lsp--path-to-uri "c:/Users/?/") "file:///c:/Users/%3F/"))`:
```
> fromPath('c:/Users/?/')
URI {
  _scheme: 'file',
  _authority: '',
  _path: '/c:/Users/?/',
  _query: '',
  _fragment: '',
  _formatted: null,
  _fsPath: null }
> fromURI('file:///c:/Users/%3F/')
URI {
  _scheme: 'file',
  _authority: '',
  _path: '/c:/Users/?/',
  _query: '',
  _fragment: '',
  _formatted: null,
  _fsPath: null }
```

`(should (equal (lsp--uri-to-path "file:///c:/Users/%7B%7D/") "c:/Users/{}/"))`: 
```
> fromURI('file:///c:/Users/%7B%7D/')
URI {
  _scheme: 'file',
  _authority: '',
  _path: '/c:/Users/{}/',
  _query: '',
  _fragment: '',
  _formatted: null,
  _fsPath: null }
> fromPath('c:/Users/{}/')
URI {
  _scheme: 'file',
  _authority: '',
  _path: '/c:/Users/{}/',
  _query: '',
  _fragment: '',
  _formatted: null,
  _fsPath: null }
```
As you can see, the resulting URI objects are equivalent, so the tests are correct as is.

sebastiencs:
Sorry, I'm a bit lost with your comment @esbena.
As I understand it, `vscode-uri` parse the same way that does `url-generic-parse-url`.

From your results (with vscode-uri):
`file:///c:/Users/%7B%7D/` -> `/c:/Users/{}/`  
with url-generic-parse-url:
`file:///c:/Users/%7B%7D/` -> `/c:/Users/{}/`   (same result)
But [our tests](https://github.com/emacs-lsp/lsp-mode/blob/master/test/lsp-common-test.el#L31) expect:
`file:///c:/Users/%7B%7D/` -> `c:/Users/{}/`

So whether there is a bug in both `url-generic-parse-url` and `vscode-uri`, or `emacs` and `vscode` can open files with such path: `/c:/path/file`.

That's how I interpret it, am I missing something ?

akermu:
As `emacs` on windows does not handle a path like `/c:/pathfile` correctly, I would propose to use `url-generic-parse-url` and remove the slash on windows, if it's the first character.

sebastiencs:
@akermu Done

esbena:
@sebastiencs you are right, and I support @akermus suggestion and your implementation. Thank you for the work.

(
I somehow missed that there is no leading slash in the test expectation. My big bad.

> But our tests expect:
> `file:///c:/Users/%7B%7D/ -> c:/Users/{}/`

)

vibhavp:
Merged, thanks.

kimim:
Hello, I got this error:
```
Debugger entered--Lisp error: (args-out-of-range "" 0)
  lsp--uri-to-path-1("file://D:\\workspace\\vision\\src\\clj\\vorstellung\\routes\\home.clj")
  lsp--uri-to-path("file://D:\\workspace\\vision\\src\\clj\\vorstellung\\routes\\home.clj")
```

Is this considered in the test? I am running Emacs from MingW64:
GNU Emacs 27.1 (build 1, x86_64-w64-mingw32) of 2020-11-20

Thanks.

kiennq:
The URI is wrong https://en.wikipedia.org/wiki/File_URI_scheme
It should be `file:///D:\\workspace\\vision\\src\\clj\\vorstellung\\routes\\home.clj` (3 slashes) since the `hostname` is empty

PR #267, opened 2018-02-01
Add :ignore-messages client option

gpittarelli:
Adds an :ignore-messages option (akin to :ignore-regexps, but applied specifically to the text of messages to be shown to the user) as discussed in https://github.com/emacs-lsp/lsp-mode/pull/251


vibhavp:
Thanks.

PR #268, opened 2018-02-02
Handle some nil results better

gpittarelli:
Fixes 2 places nils are being handled poorly:

 1. In the on-hover callback a nil result (representing an empty results list) triggers eldoc unnecessarily with an empty string. (Hard to say if this is actually user facing, but eldoc still seems to behave correctly with this fix so at the very least its cutting out a redundant function call)
 2. When using `lsp-rename`, errors aren't handled: They get sent back as nil values from `lsp--send-request` which then triggers an unrelated elisp error. Handling the nil allows the user a chance to see the actual error message.

Demonstrated on `emacs -Q` with current master:

Before:
rename-bad

After:
rename-good

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1618,9 +1618,9 @@ interface RenameParams {
   (lsp--cur-workspace-check)
   (unless (lsp--capability "renameProvider")
     (signal 'lsp-capability-not-supported (list "renameProvider")))
-  (let ((edits (lsp--send-request (lsp--make-request
-                                   "textDocument/rename"
-                                   (lsp--make-document-rename-params newname)))))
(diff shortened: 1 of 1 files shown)


vibhavp:
Isn't when-let only in Emacs 26? Using `when-let*` would be better.

phst:
Vibhav Pant <notifications@github.com> schrieb am Fr., 2. Feb. 2018 um
11:28 Uhr:

> Isn't when-let only in Emacs 26? Using when-let* would be better.
>
>
It's the other way round, when-let* is 26-only, when-let is deprecated in
26.
To be compatible with both 25 and 26 please don't use either of when-let or
when-let*. Just use a plain (let ... (when ...)), or -when-let from dash.el.

>

PR #275, opened 2018-02-04
Add a trivial test server.

phst:
This serves as a starting point for integration tests that verify how lsp-mode
interacts with various server behaviors.


phst:
I'll fix the CI once the other PRs are merged.

vibhavp:
@gpittarelli, any thoughts? He has been working on something similar here: https://github.com/gpittarelli/emacs-lsp-acceptance-testing/

gpittarelli:
They serve different testing situations: I'm trying to ensure good interop with actual LSP servers (or so far, just trying to track down existing bugs); this PRs seems to be building a mock server of sorts to emulate particular response patterns. Both are useful.

My bigger concern is adding a `go` build dependency to this repo... I could see this sort of fake LSP server being useful for other LSP clients as well, so perhaps it would live better as its own project, which `lsp-mode` can then make use of.

phst:
> My bigger concern is adding a go build dependency to this repo...

Would that be a problem in practice? Travis supports Go just fine, and installing it locally should also be no problem.

> I could see this sort of fake LSP server being useful for other LSP clients as well, so perhaps it would live better as its own project, which lsp-mode can then make use of.

Maybe, but for now I'd like to use it for one LSP client :) Moving it into its own repository would mean I'd have to deal with compatibility issues, which reduces development speed. It's definitely something to keep in mind as a future project, but I think for now we should keep it here.

yyoncho:
@phst are you still interested in merging this PR? As a side note, lsp-java has tests against real sever which I believe is doable for the other servers as well.

yyoncho:
closing pr for now, please reopen if you still have the interest to merge it.

PR #278, opened 2018-02-05
Implement window/showMessageRequest

gpittarelli:
Implements this [window/showMessageRequest](https://microsoft.github.io/language-server-protocol/specification#window_showMessageRequest) message type.

Currently I've only seen it actually used by [flow-language-server](https://github.com/flowtype/flow-language-server) (and even then, they seem to be abusing it by sending back 0 options)

Demo: (using a [dummy script](https://github.com/gpittarelli/emacs-lsp-acceptance-testing/blob/master/servers/dummy/showMessage.sh) as the server)

showmessagerequest

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -140,6 +140,11 @@
           (dolist (reg (gethash "registrations" params))
             (lsp--server-register-capability reg))
           empty-response)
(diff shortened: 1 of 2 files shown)


vibhavp:
LGTM, thanks

PR #281, opened 2018-02-07
Allow clients to customize how to get prefix bounds.

tigersoldier:
Add a new `:prefix-function` parameter to the define functions.

Currently lsp-mode graps the symbol before point as the completion prefix.
The prefix is used for matching completion candidates and will be replaced by
the selected completion candidate. Some servers use different methods to
calculate prefixes. If the prefixes between the server and the client don't
match possible results are:

 * Candidates are filtered out unexpecedly, because the client side prefix has
   extra leading characters.
 * Completed candidate has duplicated leading characters, because the server
   side prefix has extra leading characters and all the candidates also have
   the leading characters.

Only the clients knows which server they are talking to, so it's the clients'
responsibility to implement custom prefix funtions if necessary.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -126,7 +126,13 @@
 
   ;; Function to enable the client for the current buffer, called without
   ;; arguments.
-  (enable-function nil :read-only t))
+  (enable-function nil :read-only t)
+
+  ;; ‘prefix-function’ is called for getting the prefix for completion.
+  ;; The function takes no parameter and returns a cons (start . end) representing
+  ;; the start and end bounds of the prefix. If it's not set, the client uses a
+  ;; default prefix function."
+  (prefix-function nil :read-only t))
 
 (cl-defstruct lsp--registered-capability
   (id "" :type string)
@@ -1273,9 +1279,15 @@ https://microsoft.github.io/language-server-protocol/specification#textDocument_
                         (lsp--sort-string c1)
                         (lsp--sort-string c2)))))
 
+(defun lsp--default-prefix-function ()
+  (bounds-of-thing-at-point 'symbol))
+
 (defun lsp--get-completions ()
   (with-demoted-errors "Error in ‘lsp--get-completions’: %S"
-    (let ((bounds (bounds-of-thing-at-point 'symbol)))
+    (let* ((prefix-function (or (lsp--client-prefix-function
+                                  (lsp--workspace-client lsp--cur-workspace))
+                             #'lsp--default-prefix-function))
+            (bounds (funcall prefix-function)))
       (list
        (if bounds (car bounds) (point))
        (if bounds (cdr bounds) (point))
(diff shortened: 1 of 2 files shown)


tigersoldier:
@vibhavp @gpittarelli mind reviewing this PR? I have pending PRs in company-lsp and lsp-javacomp showcasing how this will be used:
tigersoldier/company-lsp#26
tigersoldier/lsp-javacomp#1

This would provide the foundation for fixing a cquery-related bug: tigersoldier/company-lsp#21

vibhavp:
Isn't getting the bounds of the identifier under point a feature provided by the language major mode? Dealing with semantics of language syntax should not be lsp-mode's job, IMO.

tigersoldier:
The definition of the bound under the point is different between emacs major mode and the language servers.

----

For example, consider the the C++ code below:

```C++
#include|
```

`|` denotes the cursor position. cc-mode considers the bounds to be around `include`, while cquery considers the bounds to be around `#include`. The candidates cquery provides all start with `#`. When a candidate is chosen, say `#include <stdio>`, `include` is replaced with the candidate, resulting the content to be

```C++
##include <stdio>|
```

with unexpected redundant `#`.

----

A second case is Java. Consider the following code:

```java
@Over|
```

java-mode considers the bounds to be `@Over`. However JavaComp considers the bounds to be `Over`. All candidates don't have the `@` prefix and are filtered out because none of them match `@Over`.

----

It doesn't make sense for the language servers to match the bounds algorithm in emacs because the servers are editor-independent.

It doesn't make sense for lsp-mode to match the bounds algorithm in language servers because lsp-mode is language-agnostic.

It doesn't make sense for major modes to match the bounds algorithm in language servers because a language can have multiple servers and they may have different algorithms.

So it should be clients' responsibility to provide the bounds algorithm for the server they talk to. It's optional. If a client doesn't want to bother providing bounds, lsp-mode still provide a default one. But lsp-mode should allow clients to customize it if they want to.

tigersoldier:
@vibhavp Do you have any other concerns on this PR?

vibhavp:
I haven't been able to find time for reviewing this, I'll do so the coming week. Sorry for the delay!

MaskRay:
LG. In the cquery case, triggering `completion-at-point` at `#|` results in the following response from the server:

```json
"textEdit":{"range":{"start":{"line":2,"character":0},"end":{"line":2,"character":1}},"newText"
:"#include <aio.h>"}}
```

The first character `#` is supposed to be replaced by `#include <aio.h>`

vibhavp:
@tigersoldier Looks good. I'll merge this if you can fix the conflicts.

lsp-methods.el
```diff
@@ -39,6 +39,7 @@
   (get-root nil :read-only t)
   (ignore-regexps nil :read-only t)
   (ignore-messages nil :read-only t)
+  (prefix-function nil :read-only t)
```
vibhavp: Could you make `prefix-function` the last struct member? Adding a new member in the middle of the struct breaks binary compatibilities with byte-compiled client packages.

tigersoldier: Done

tigersoldier:
Please take a look again. I rebased to HEAD of master. Addressed comments.

vibhavp:
Thanks.

PR #283, opened 2018-02-09
Update link to spec in lsp--gethash's docstring

NateEag:
The current link now points to a shut down GitHub repo, so I've updated it to
 point to the site the old repo recommended.


gpittarelli:
Thanks!

NateEag:
You're welcome!

Thanks for the refined anchor - I thought the name I grabbed was ugly but assumed that was just what they'd used.

That's what I get for browsing with narrow windows (the TOC isn't shown below some breakpoint, it seems).

PR #286, opened 2018-02-14
Clean up process sentinel.

phst:
- Only uninitialize workspace if the process has actually exited.  Emacs calls
  the sentinel after other events as well, e.g. starting, stopping, or
  continuing the process, and uninitializing the workspace then would be
  incorrect.

- Leave the standard error buffer intact if the server exited abnormally, for
  debugging purposes.

- Only print exit message once, not once per buffer.

- Instead of passing the (arbitrary) buffer that triggered creating the
  workspace, pass the workspace directly.  This means we don’t have to check
  any more whether the buffer is live.


vibhavp:
Thanks.

PR #289, opened 2018-02-15
Merge workspace and client structures.

phst:
There doesn’t seem to be any conceptual difference between workspaces and
clients, so let’s get rid of one of the concepts.

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -69,7 +69,7 @@
   (queued-notifications nil) ;; Unused field
   (queued-requests nil)
 
-  (workspace nil) ;; the workspace
+  (client nil) ;; the client
   )
 
 (define-error 'lsp-parse-error
@@ -102,8 +102,8 @@
       'notification
       (signal 'lsp-unknown-message-type (list json-data)))))
 
-(defun lsp--default-message-handler (workspace params)
-  (lsp--window-show-message params workspace))
+(defun lsp--default-message-handler (client params)
+  (lsp--window-show-message params client))
 
 (defconst lsp--default-notification-handlers
   #s(hash-table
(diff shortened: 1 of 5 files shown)


tigersoldier:
This is a backwards incompatible change and is likely breaking company-lsp. Please consider providing alias functions and bump the version.

phst:
> This is a backwards incompatible change and is likely breaking company-lsp.

This refactoring only affects internal data structures, which other packages shouldn't rely on. If `company-lsp` is broken by this, it should be changed to only rely on the public API (extending the public API if necessary).

tigersoldier:
There is no proper public APIs for tools like company-lsp to work with lsp-mode. This change not only breaks company-lsp, but also lsp-ui hosted in this very same organization. These two packages are well-received by the lsp-mode users. Breaking them causes undesirable pains to the users.

Refactoring is good and I don't want to block it. But please consider doing it in a less harmful way. Either a) define true public APIs that company-lsp and lsp-ui can rely, wait until packages migrate to using them, or b) provide backward-compatible function aliases so that they are not broken.

If you are taking option a), I'm happy to list all private APIs company-lsp is using.

cc @sebastiencs

phst:
I think we should definitely move toward option (a). Not distinguishing between public APIs and private implementation details makes refactorings too onerous.

MaskRay:
I agree that with this PR, the name "client" captures more of the intuitive meaning. The specification does use both "workspace" and "client", but in a different way.

`lsp--cur-workspace` `lsp--workspace-root`  `lsp--workspace-client` have been used many places in lsp-ui

yyoncho:
Closing in favour of #469 . `lsp--client` will be used as a registration spec.

PR #290, opened 2018-02-18
Lsp restart

### LINKED ISSUE #24: add `lsp-restart`
It would kill the current lsp if any is running, and restart a new one (without query).

This is useful for debugging, etc.

alanz:
Closes #24


lsp-methods.el
```diff
@@ -329,37 +330,68 @@ interface TextDocumentItem {
...
+      ;; Shut down the LSP mode for each buffer in the workspace
+      (dolist (buffer1 old-buffers)
+        (with-current-buffer buffer1
+          (lsp--text-document-did-close)
+          (setq lsp--cur-workspace nil)
+          (lsp-mode -1)))
+
+      ;; Need to let the sentinel finish off. It would be better to make this
+      ;; deterministic, not sure how to.
+      (sleep-for 2)
```
phst: How about:

```elisp
(while (process-live-p proc)
  (accept-process-output proc))
```

lsp-methods.el
```diff
@@ -45,7 +45,8 @@
   (response-handlers (make-hash-table :test 'eq) :read-only t)
 
   (string-renderers '())
-  (last-id 0))
+  (last-id 0)
+  (restart-command nil :read-only t))
```
phst: Please add a comment what `restart-command` is (i.e. its type and expected behavior).

lsp-methods.el
```diff
@@ -1722,6 +1754,7 @@ command COMMAND and optionsl ARGS"
 ;; (defalias 'lsp-on-change #'lsp--text-document-did-change)
 (defalias 'lsp-completion-at-point #'lsp--get-completions)
 (defalias 'lsp-error-explainer #'lsp--error-explainer)
+(defalias 'lsp-restart-workspace #'lsp--restart-workspace)
```
phst: Why this alias? If `lsp--restart-workspace` is meant to be called by users, just make it public.

lsp-mode.el
```diff
@@ -176,7 +176,8 @@ Optional arguments:
                            :stderr stderr
                            :get-root ,get-root
                            :ignore-regexps ,ignore-regexps
-                           :ignore-messages ,ignore-messages)))
+                           :ignore-messages ,ignore-messages
+                           :restart-command (function ,enable))))
```
phst: Not sure whether the field should be called `:restart-command`, it doesn't really restart the server, but can only start it.

lsp-methods.el
alanz: I am just cargo culting the code, elisp is not my primary language :)

lsp-mode.el
alanz: It is only used to restart it, it is actually a reference to the function being defined in this macro.

phst: Yes, I see that :) But for future readers of the code, it should describe what it does, not how it's used or implemented.
How about renaming it to `:enable-function`? It's not required to be a command, and `enable` expresses better what it does.

lsp-methods.el
```diff
@@ -329,37 +334,69 @@ interface TextDocumentItem {
...
-    (maphash (lambda (_k value) (lsp--teardown-client value)) lsp--workspaces)))
+    (maphash (lambda (_k value) (lsp--teardown-workspace value)) lsp--workspaces)))
 
-(defun lsp--teardown-client (client)
-  (setq lsp--cur-workspace client)
+(defun lsp--teardown-workspace (workspace)
+  (setq lsp--cur-workspace workspace)
   (lsp--shutdown-cur-workspace))
 
+(defun lsp--shutdown-cur-workspace ()
```
phst: Nit: I'd prefer not moving this function around, as it only generates a needless diff.

lsp-methods.el
```diff
@@ -329,37 +334,69 @@ interface TextDocumentItem {
...
+    (lsp--send-request (lsp--make-request "shutdown" (make-hash-table)) t)
+    (lsp--send-notification (lsp--make-notification "exit" nil)))
+  (lsp--uninitialize-workspace))
+
 (defun lsp--uninitialize-workspace ()
   "When a workspace is shut down, by request or from just
 disappearing, unset all the variables related to it."
-  (remhash (lsp--workspace-root lsp--cur-workspace) lsp--workspaces)
-  (let (proc)
+  (let (proc
```
phst: Since you're touching this code anyway, you can refactor it by assigning `proc` while binding it (`(let ((proc (lsp-workspace-proc ...)) ...)` or something more fancy like `(pcase-exhaustive lsp--cur-workspace ((cl-struct lsp--workspace proc root) ...))`).
Also you can remove the `(with-current-buffer (current-buffer))`, it doesn't do anything.

lsp-methods.el
```diff
@@ -329,37 +334,69 @@ interface TextDocumentItem {
...
+the workspace, closing the process managing communication with
+the client, and then starting up again."
+  (interactive)
+  (when (and (lsp-mode) (buffer-file-name))
+    (let ((old-buffers (lsp--workspace-buffers lsp--cur-workspace))
+          (restart (lsp--client-restart-command (lsp--workspace-client lsp--cur-workspace)))
+          (proc (lsp--workspace-proc lsp--cur-workspace)))
+
+      ;; Shut down the LSP mode for each buffer in the workspace
+      (dolist (buffer1 old-buffers)
```
phst: Why calling these variables `buffer1` and `buffer2`? `buffer` should work just fine.

lsp-methods.el
```diff
@@ -329,37 +334,69 @@ interface TextDocumentItem {
...
+        (with-current-buffer buffer1
+          (lsp--text-document-did-close)
+          (setq lsp--cur-workspace nil)
+          (lsp-mode -1)))
```
phst: Could you modify `lsp-mode` itself to do this? That way you'd turn `lsp-mode` into a real mode that can be turned on and off.

lsp-methods.el
```diff
@@ -329,37 +334,69 @@ interface TextDocumentItem {
...
+
+      ; Let the process actually shut down
+      (while (process-live-p proc)
+        (accept-process-output proc))
+
+      ; Re-enable LSP mode for each buffer
+      (dolist (buffer2 old-buffers)
+        (with-current-buffer buffer2
+          (funcall restart)
```
phst: Nit: it's more common in Lisp to not insert linebreaks before closing parentheses.

lsp-methods.el
```diff
@@ -329,37 +334,69 @@ interface TextDocumentItem {
...
       (unless (eq (process-status proc) 'exit)
         (kill-process (lsp--workspace-proc lsp--cur-workspace)))
       (setq lsp--cur-workspace nil)
       (lsp--unset-variables)
-      (kill-local-variable 'lsp--cur-workspace))))
+      (kill-local-variable 'lsp--cur-workspace))
+    (remhash root lsp--workspaces))
+  )
+
+(defun lsp-restart-workspace ()
```
phst: Could you add a unit test for this function?

lsp-methods.el
alanz: I was getting errors, wanted to know from the backtrace which one.

lsp-methods.el
alanz: No, it is outside the scope of this PR. And I don't know how to do it, and got all sorts of crazy feedback the last time I did something similar.

I will leave it as an enhancement for a future contributor.

lsp-methods.el
alanz: Do we have a sample server we are using for the tests?

phst: Not yet, https://github.com/emacs-lsp/lsp-mode/pull/275 should add one.

alanz:
@vibhavp I have rebased this, please take a look and merge if you are happy

alanz:
Ping @vibhavp @phst 

Is there a reason this cannot go in?

alanz:
I generally prefer someone else to merge my PRs. Thanks for merging.

PR #301, opened 2018-03-03
Don't set projectile-project-root in lsp-start

### LINKED ISSUE #300: lsp-mode breaks projectile-switch-project
It appears that lsp-mode breaks projectile projecti switching.
I have several project that use lsp-mode.
When I use `projectil-switch-project` for the first time it works fine.
by default projectile show me a list of files for new project after switch. And when I'm trying to switch another project, it doesn't work as expected. Intead of showing files from new project it shows files from previous one.

Seems like that issue it cause by this code: 
https://github.com/emacs-lsp/lsp-mode/blob/13addc30bbaafe19bd203ba0be9a19b6f50c9c20/lsp-methods.el#L732
```
    (when (boundp 'projectile-project-root)
(setq-local projectile-project-root root))
```

Simple renaming of fix the problem. I suspect because in `projectile` it's a function name
https://github.com/bbatsov/projectile/blob/master/projectile.el#L898

jaelsasser:
Partially reverts #257 by moving the projectile-specific configuration into a hook script. Should also fix #300 (or at least shove the issue off into `init.el` rather than `lsp-mode`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,3 +110,19 @@ There are also the functions `lsp-MAJOR-MODE-whitelist-add` and
 `lsp-MAJOR-MODE-whitelist-remove` to adjust the current buffer project root
 entry on the whitelist.
 
+
+### Hooks
+
+`lsp-mode` provides a handful of hooks that can be used to extend and configure
+the behaviour of language servers. A full list of hooks is available in the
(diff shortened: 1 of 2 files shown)


vibhavp:
LGTM, thanks.

PR #307, opened 2018-03-10
Make symbol highlighting looks good on dark themes.

AmaiKinono:
The highlighting faces' background is decided by the theme's background, thus makes it looks good on both light and dark themes.

lsp-face-dark

lsp-face-light


vibhavp:
Thanks!

PR #308, opened 2018-03-10
Optimize symbol highlighting

sebastiencs:
With hundreds of references, symbol highlighting was very slow.
This PR eliminates totally the slow down.

On my machine, I now start to feel a slow down with +2000 references, which is due to `json-read`


vibhavp:
Thanks.

Issue #309, opened 2018-03-10
Issues using java (JDT)


yyoncho:
I did some testing using lsp-mode, lsp-ui and lsp-java and I encountered several issues and it is not clear to me whether they are caused by my setup or they require additional work on lsp-mode side.

1. The JDT server is returning the ids as a string and as a result the callbacks are not matched. 
2. The JDT server sends java.apply.workspaceEdit actions which contain the changes but they are not applied but the action is sent to the server. 
3. The logic related to applying changes fails due to the sorting of the edits e. g. it is common for the java server to send edits for the same position and they end up being applied in wrong order. I changed the code to apply the updates in reverse order it works fine.

yyoncho:
1. Fixed with #313 
2. Will be fixed in lsp-java

PR #324, opened 2018-03-28
Fix handling of overlapping edits

yyoncho:
* see https://github.com/Microsoft/language-server-protocol/commit/005d12dc021b9ea087e4d1fe4915c43fe7860be8

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -982,7 +982,7 @@ interface TextDocumentEdit {
 (define-inline lsp--apply-text-edits (edits)
   "Apply the edits described in the TextEdit[] object in EDITS."
   (inline-quote
-    (mapc #'lsp--apply-text-edit (sort ,edits #'lsp--text-edit-sort-predicate))))
+    (mapc #'lsp--apply-text-edit (sort (reverse ,edits) #'lsp--text-edit-sort-predicate))))
 
 (defun lsp--apply-text-edit (text-edit)
   "Apply the edits described in the TextEdit object in TEXT-EDIT."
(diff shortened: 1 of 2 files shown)


yyoncho:
Fixes point 3 from https://github.com/emacs-lsp/lsp-mode/issues/309

vibhavp:
Not sure how reversing the text edit order would conform to the standard. Quoting from the spec:

> If multiple inserts have the same position, the order in the array defines which edit to apply first.

By reversing the edits array, `TextEdit`s that need to be applied later would be applied _first_.

yyoncho:
@vibhavp 

You are right, the operations must be applied in the same order they are received. 

Assuming lsp-mode has received the following updates

> [0, 0, "Hello"]
> [0, 0, " world!"] 
> [1, 0, "Second line..."]

Before the fix after  the updates are sorted they will be applied in the following order:
3, 1, 2, after the fix they will be applied in the correct order 3, 2, 1.

Here it is the same code in vscode. 
https://github.com/Microsoft/vscode/blob/master/src/vs/editor/common/model/pieceTreeTextBuffer/pieceTreeTextBuffer.ts#L207
It is applying stable sort which reverses the elements in the map reversing the original order before applying the edits. In the emacs use case the sort function is already performing stable sort so we need to ensure that the overlapping edits are reversed before applying. 

I have performed few debug session of the vscode and if you are still not convinced I might attach few screenshots of the edits when they are received and before they are applied.

vibhavp:
Ah, I completely forgot about why we sorted text edits in the first place. I'll add some additional documentation to this, thanks.

PR #326, opened 2018-03-31
Provide option different modes to plugin uri<->path conversion

yyoncho:
https://github.com/emacs-lsp/lsp-java/issues/18

When the user looks for Definition/implementation/reference that is not part of the
current project source code the JDT server returns URI prefixed with jdt: and
then it has a separate call which retrieves the source. After this change, different modes will be able to register a URI resolver.

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -19,6 +19,7 @@
 (require 'url-util)
 (require 'url-parse)
 (require 'subr-x)
+(require 'cl)
 
 (defconst lsp--message-type-face
   `((1 . ,compilation-error-face)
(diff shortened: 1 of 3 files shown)


yyoncho:
@vibhavp just to make sure we are on the same page, I will create a method for registering uri handlers similar to the notifications/response.

Are you ok with this approach?

Thanks

vibhavp:
Yeah, sound good.

yyoncho:
@vibhavp, I have addressed the comments, do you have more concerns?

lsp-common.el
```diff
@@ -36,6 +36,9 @@
                                (_ "file://"))
   "Prefix for a file-uri.")
 
+(defvar-local lsp-buffer-uri nil
```
vibhavp: Alternatively, is it possible to use a programmer-provided function for computing the current buffer uri too?

yyoncho: It is possible, I choose this approach since it allows inlining and the original method was inlined.

PR #328, opened 2018-04-01
Extend lsp-mode to provide option to register action handlers

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/309


vibhavp:
IMO, url handlers can be handled/registered better by a LSP client in the `lsp-define-client` macros.

yyoncho:
Hi @vibhavp ,

Thanks for the feedback!

I am a bit confused, I tried to follow the patters established by the lsp-mode, e. g. both notification-handlers have corresponding register method to make it as consistent as possible, are you planning to move both to the define client macros as well?

yyoncho:
@vibhavp Now I get it -  I believe your comment is not for this request but for https://github.com/emacs-lsp/lsp-mode/pull/326 ?  I was considering moving the url handlers in the lsp--workspace but I wasn't sure what was the lifetime of the client. I will try to move it.

vibhavp:
Yeah, this was on the incorrect issue. Regardless, does the specification allow code actions to be executed by the client?

yyoncho:
From the spec:

"The protocol currently doesn’t specify a set of well-known commands. So executing a command requires some tool extension code."

lsp-methods.el
```diff
@@ -113,6 +113,12 @@
   ;; single argument, the deserialized response parameters.
   (response-handlers (make-hash-table :test 'eql) :read-only t)
 
+  ;; ‘action-handlers’ is a hash table mapping action to a handler function. It
+  ;; can be used in `lsp-execute-code-action' to determine whether the action
+  ;; current client is interested in executing the action instead of sending it
+  ;; to the server.
+  (action-handlers (make-hash-table :test 'equal) :read-only t)
```
vibhavp: Could you make this the last field of the struct? Otherwise, we'd be unnecessarily breaking binary compatibility.

yyoncho: Sure, I didn't know that this will have an effect on binary compatibility, thanks!

vibhavp:
Merging for now, as all documentation for code actions I could find is still ambiguous.

PR #338, opened 2018-04-10
Added missing capabilities to lsp-capabilities

cormacc:
List of reported server capabilities didn't check for a number of items,
including typeDefinitionProvider and implementationProvider

Also added capability check / signal to lsp-format-buffer if capability not supported (for consistency with other existing methods -goto-type-definition etc)


vibhavp:
Looks good, thanks.

PR #356, opened 2018-05-11
Offer a option for user to choose project root interactively

samrayleung:
When no file could be found to identify project root, allowing `lsp-mode` to prompt user to choose project root interactively, or log a warning and return `default-directory`. This feature is toggled by `lsp-prompt-project-root` variable

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -31,6 +31,12 @@
   :group 'lsp-mode
   :type 'boolean)
 
+
+(defcustom lsp-prompt-project-root nil
+  "If non-nil, choose project root interactively when could not find specified project root."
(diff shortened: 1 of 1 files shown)


yyoncho:
Thanks  for the PR - @MaskRay have reimplemented it at https://github.com/emacs-lsp/lsp-mode/commit/f65aab2f0eed74f3d92cefbe1cd172f4fb541b8d . Marking as closed.

MaskRay:
You can advice `lsp--suggest-project-root` to prompt for the project root. But IMHO prompting by default does not provide a good experience. There are `/usr/include/*` files, single-file projects where the user does not bother starting a language server.

Issue #235, opened 2018-01-10
Eldoc should show function signature only


amol-mandhane:
Currently, the eldoc of lsp-mode shows entire documentation string of the function/class/variable. This can show a lot of text in minibuffer which limits visibility (for example #96 #224). This get seriously messy with functions with long documentations. Showing full documentation on hover is more suitable for UI-oriented editors like VSCode which can show a popup with scroll-bar to show documentation in a more compact fashion.

As per [Eldoc Homepage](https://www.emacswiki.org/emacs/ElDoc), it is meant only for function signatures. Quite a lot of Emacs tooling follows this convention. LSP protocol has a method to get signature, "textDocument/signatureHelp". This should be used for eldoc instead of textDocument/hover. The hover method can be linked to a "help-at-point" kind of keybinding instead.

#226 adds signatureHelp support to LSP mode. This can be a drop in replacement for `lsp--on-hover`.

tadfisher:
Yep, I'd suggest adding an option to display hover info in a separate window, like C-h f.

wberrier:
Is this still an issue?

I see the `lsp-hover-text-function` customization, but when I set it to `'lsp--text-document-signature-help`, I don't get any eldoc functionality at all.

?

yyoncho:
@wberrier you may try to `lsp-eldoc-render-all` to `nil`.

yyoncho:
I am closing this issue since there already is support signature help in lsp-mode.

wberrier:
@yyoncho setting `lsp-eldoc-render-all` to `nil` produces only the function documentation, but not the signature.

I guess I should open a new issue then that `'lsp--text-document-signature-help` doesn't seem to be working correctly?

wberrier:
Opened a new issue: #443

PR #357, opened 2018-05-11
Show function signature on hover instead of full help.

izahn:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/235. I guess this is premature since we are waiting for discussion/consensus, but IMO Eldoc is clearly the wrong tool for displaying full help pages, and the current behavior is just not useful. (At least it is not useful in Python or R, the two languages I use most often.) I'm submitting this PR in an attempt to move this issue toward a solution, even if we're not ready to merge it yet.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1505,7 +1505,7 @@ Returns xref-item(s)."
     (when (and (lsp--capability "codeActionProvider") lsp-enable-codeaction)
       (lsp--text-document-code-action))
     (when lsp-enable-eldoc
-      (lsp--text-document-hover-string))))
+      (lsp--text-document-signature-help))))
 
 (defvar-local lsp--cur-hover-request-id nil)
 


yyoncho:
I believe that there must be an option to switch between both. At this point there is an issue in the signature calculation - the method does not work in brackets but some of the servers (e. g. JDT) return function signature only if the method is invoked in brackets.

gpittarelli:
This was fixed by https://github.com/emacs-lsp/lsp-mode/commit/499cf809fdc46282d625ad82c79a812a426b0046

To yyoncho's point, maybe we should provide an option to run both functions and use signature if available, else the hover text.

yyoncho:
@gpittarelli IMO the following processing order will handle all of the cases:

If there is `symbol-at-point` - use `onHover`. I have added an option to render only symbol information 
https://github.com/emacs-lsp/lsp-mode/pull/365 so `eldoc-mode` won't render all symbol information.
If there is no symbol at point - use signature.

PR #365, opened 2018-05-20
document/onHover improvements

yyoncho:
1. Added option(lsp-eldoc-render-all) to display only the first message from the list retrieved from
the server.
2. Added option to specify default client renderer via
lsp-default-renderer-language.
3. Added function to display help on demand(lsp-describe-thing-at-point)

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -29,8 +29,6 @@
     (changed . 2)
     (deleted . 3)))
 
-(defconst lsp-default-renderer-language "__default__renderer__language")
-
 ;; A ‘lsp--client’ object describes the client-side behavior of a language
 ;; server.  It is used to start individual server processes, each of which is
 ;; represented by a ‘lsp--workspace’ object.  Client objects are normally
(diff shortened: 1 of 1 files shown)


vibhavp:
Why have a default renderer, though? Each language has its own syntax highlighting rules, and a default renderer wouldn't do anything. If anything, a *fallback* renderer might be better.

yyoncho:
Just to elaborate a bit more - from what I am seeing the JDT server is returning a plain string object. So I want to have a workspace specific handler in order to turn it to markup. There is no option to register handler for `nil` so I decided to introduce `default` renderer for the workspace. The alternative was to set `lsp-render-markdown-markup-content` but this will affect all workspaces.

yyoncho:
@vibhavp do you have more concerns?

yyoncho:
@MaskRay can you take a look? I guess @vibhavp is pretty busy lately. I want to have that fix in before creating `Spacemacs` java layer.

vibhavp:
Sorry for the late replies, as I have been busy with GSoC. Wouldn't these changes break renderers for languages with the id "DEFAULT"?

yyoncho:
Yes. I could change the constant to "__default__renderer__language" .

vibhavp:
Isn't this using a string nonetheless? Could it be possible to use a sentinel value which isn't a string type?

yyoncho:
It has to be string, unless we change the way we store renderers:
```emacs-lisp
(defun lsp-provide-marked-string-renderer (client language renderer)
  (cl-check-type language string)
```

yyoncho:
@vibhavp I have removed the string constant, do you have more concerns?

yyoncho:
@vibhavp any more concerns? I have removed the usage of the string constant.

vibhavp:
Looks good, thanks. Sorry about the late replies!

PR #367, opened 2018-05-23
Check CAPS argument type in lsp-register-client-capabilities

iquiw:
whether it is either list or function.

Relevant commit: https://github.com/emacs-lsp/lsp-mode/commit/d9e31de8a8e43fe39d2e14b96b90bd809a26ab8a


vibhavp:
Thanks.

Issue #342, opened 2018-04-22
Warning (lsp-mode): Couldn’t find project root, using the current directory as the root.


casouri:
Apparently this warning appears when lsp cannot find project root. It used to be in echo area so I can bear it. Now this warning appears in a warning buffer which pops up every time and interrupts my work. 

Is there an option to alter this behavior? Can I choose to not show any warning or make warnings appear in echo area?

innerout:
if you are using lsp-python you can create a file in the root of your project with the name setup.py and the error stops showing up.

casouri:
As a scripting language, python is wrote in a single file in many cases. Creating another file is a bit redundant.

yyoncho:
You could change the way lsp-python looks for its root folder here it is the code fragment:
`(lsp-define-stdio-client lsp-python "python"
			 (lsp-make-traverser #'(lambda (dir)
						 (directory-files
						  dir
						  nil
						  "setup.py")))
			 '("pyls"))`
if you change "setup.py" to whatever file you have in the root(e. g. git) it will work or you can return projectile root, etc.

casouri:
Thanks.

But still, what if the user wants to edit a simple python script? Does lsp-mode HAVE to warn about root dir not found in a poped-up warning buffer?

I'll close this issue if you think lsp-mode does.

nicholas-miklaucic:
Giving this warning in a buffer that doesn't appear automatically and block work would be, IMHO, more akin to the philosophy of the rest of Emacs of rarely if ever giving modal dialogs or blocking other windows.

benjaminor:
How about if you defined a variable, let's say 'show-project-root-warning' (set by default to true), which one could customize at emacs start-up time?
And then deactivate the warning if show-project-root-warning is set to nil?

I could create a pull request if you think that's a good idea.

xedrac:
This has been driving me crazy too.  I'm fine with the warning getting added to a *Warnings* or *Messages* buffer, but we need to be able to prevent the automatic display of the warnings buffer every time something is written to it.

vibhavp:
`lsp-message-project-root-warning` should fix this.

PR #370, opened 2018-06-04
Show project root warning optionally in *Messages* buffer

benjaminor:
Create option to show project root warning as a message and not in *Warnings* buffer.
Possible solution for Issue  #342.


vibhavp:
Looks good, thanks.

PR #373, opened 2018-06-10
Make sure lsp--capability never errors

wyuenho:
There are lots of code that call (lsp--capability) assuming the server is ready, thus not providing a default capabilities hash table, but that's not always the case for slow servers like JS/TS/Flow. If those (lsp-langserver-enable) are used as major mode hooks, and if lsp-ui is enabled, lots of code like [lsp-ui-sideline](https://github.com/emacs-lsp/lsp-ui/blob/master/lsp-ui-sideline.el#L400) will make a request to the language server as soon as the major mode is on but before the server is ready, in which case (get-hash) will be given a nil instead of a hash table, and errors out. This PR fixes it.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1064,7 +1064,7 @@ interface TextDocumentEdit {
 
 (define-inline lsp--capability (cap &optional capabilities)
   "Get the value of capability CAP.  If CAPABILITIES is non-nil, use them instead."
-  (inline-quote (gethash ,cap (or ,capabilities (lsp--server-capabilities)))))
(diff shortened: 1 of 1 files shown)


gpittarelli:
Thanks for the PR. The code looks fine. The only way I could repro the bug is by interrupting (with C-g) the language startup (`lsp--start`) before it receives the server initialize message. In that situation the lsp-mode experience will be reasonably poor anyways... but I guess it's worthwhile to suppress some more errors so at least its not a poor experience and throwing errors on every cursor move.

PR #374, opened 2018-06-10
Fix indices for sync-methods

robertoaloi:
Maybe I'm missing something, but shouldn't the indices match with the ones defined as constant?


gpittarelli:
Doh, good catch! Thanks. (And thankfully that only affected the view when using `lsp-capabilities` instead of breaking the actual text syncing :sweat_smile: )

PR #375, opened 2018-06-10
Don't request hover when server cannot handle textDocument/hover

wyuenho:


gpittarelli:
Tested with a language server that doesn't support hover and this resolved an error. Thanks!

PR #380, opened 2018-06-15
Enable lsp-mode after setting the current workspace.

vibhavp:


gpittarelli:
Tested it out on all the modes I have setup and it seemed to work fine :+1: 

(I was confused a bit, so got this from vib:)
> lsp-mode-hook will be invoked after `lsp--cur-workspace` is set
> So any hooks will be able to send requests and notifications to the server
> https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-methods.el#L482
> lsp--send-request needs lsp--cur-workspace to be set

PR #383, opened 2018-06-19
Add port-fn for TCP clients

gpittarelli:
This is an important missing piece for getting TCP servers to work smoothly:
  - support dynamic port number (eg, when a server calls `listen` with port=0 to get a random OS-assigned port)
  - support waiting a bit between starting the language server process and attempting a TCP connection (currently, we don't wait at all, so the language server doesn't get a chance to start listening in time) 

The best Ruby language server currently only runs in TCP mode; so see https://github.com/emacs-lsp/lsp-ruby/pull/1 as an example of how this could be used to achieve reliable startup.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -59,11 +59,11 @@
         (set-process-query-on-exit-flag (get-buffer-process (get-buffer stderr)) nil)
         proc))))
 
-(defun lsp--make-tcp-connection (name command command-fn host port stderr)
+(defun lsp--make-tcp-connection (name command command-fn host port stderr &key port-fn)
   (lambda (filter sentinel)
     (let* ((command (if command-fn (funcall command-fn) command))
-            (final-command (if (consp command) command (list command)))
-            proc tcp-proc)
+           (final-command (if (consp command) command (list command)))
+           proc tcp-proc)
       (unless (executable-find (nth 0 final-command))
         (error (format "Couldn't find executable %s" (nth 0 final-command))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -262,7 +264,12 @@ Optional arguments:
 `:prefix-function' is a function called for getting the prefix for completion.
  The function takes no parameter and returns a cons (start . end) representing
  the start and end bounds of the prefix. If it's not set, the client uses a
- default prefix function."
+ default prefix function.
+
+`:port-fn' is a function called after starting the language server process, but
+ before connecting to it. The function must return a number, which will be used
+ as the port to connect to. This is useful for servers which use random port
+ numbers, and also for delaying until the server is ready to accept connections."
```
vibhavp: Isn't this too hacky? Instead, we can replace this with a way to defer requests, or wait for when the server starts accepting connections/input.

gpittarelli: Agreed using this to delay the connection is super hacky and not the ideal goal. I can change the docs to not mention that and work on a better asynchronous solution. We do need this function anyways to handle dynamic port numbers.

waymondo:
I ran into this as well trying to get `lsp-ruby` to work with `solargraph`. I also agree that synchronously waiting within a `:port-fn` function is not the ideal solution, but the general issue does exist. A dynamic port function seems reasonable to add, as well as improving the asynchronous initialization of TCP clients.

PR #387, opened 2018-06-23
Guard against position being out of bounds of the buffer

### LINKED ISSUE #386: Formatting fails silently if the end position in the returned TextEdit is incorrect
When using `lsp-format-buffer` if a language server responds with an end position that is invalid for the file (for example: scala language server returns with end position as (Int.MaxValue, Int.MaxValue)) it fails silently.

This is due to `forward-char` in `lsp--position-to-point`, giving a `N` greater than characters available in the buffer, seems to terminate with message `End of buffer` without progressing further.

Even though the language server is at fault here it may be wise to have some checks on `lsp--position-to-point`, possibly using `point-max`.

yyadavalli:
Fix #386

### DIFF
--- a/lsp-common.el
+++ b/lsp-common.el
@@ -59,14 +59,21 @@
 
 ;; from http://emacs.stackexchange.com/questions/8082/how-to-get-buffer-position-given-line-number-and-column-number
 (defun lsp--position-to-point (params)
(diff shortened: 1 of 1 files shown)


yyadavalli:
Closed prefer #388

PR #406, opened 2018-08-17
Temporary comment the check for executeCommandProvider

yyoncho:
JDT LS does not declare executeCommandProvider so lsp-java is broken. Comment
this check until the JDT fix is in.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -2135,8 +2135,9 @@ interface RenameParams {
   "Create and send a 'workspace/executeCommand' message having
 command COMMAND and optionsl ARGS"
   (lsp--cur-workspace-check)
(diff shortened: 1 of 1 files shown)


vibhavp:
This would end up affecting all lsp client packages. Could you use `lsp-after-initialize-hook` for this instead, by modifying `server-capabilities` in `lsp--workspace`?

yyoncho:
Will try it, thanks.

yyoncho:
I am closing the pull request. As a side note in client initialized the workspace is not set and there is no way to get it. I guess this is something that has to be fixed on `lsp-mode` side (`lsp-java` is using advices in order to accomplish that functionality.

Issue #345, opened 2018-04-27
Support workspace folders


yyoncho:
When the server provides support for workspace folders lsp-mode should expose methods for manipulating it and it should expose hooks for added/removed events. 

https://microsoft.github.io/language-server-protocol/specification#workspace_workspaceFolders

yyoncho:
@vibhavp I am planning to address this issue, can you assign it to me?

My plan is to create a file in the workspace folder which will be used to persist the workspace folders. Each time the user adds/removes a folder the file will be updated. It will be read and populated on workspace startup.

PR #409, opened 2018-08-23
Implemented multiple folders support

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/345

* Added methods for adding/removing folders
* Tested with JDT server which seems to be the only server supporting the
feature.
* Verified that clients which does not support workspace folders work as
  expected(verfied with lsp-python)

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -13,6 +13,8 @@
 ;; You should have received a copy of the GNU General Public License
 ;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
+;;; Code:
+
 (require 'cl-lib)
 (require 'json)
 (require 'xref)
@@ -24,7 +26,7 @@
 (require 'inline)
 (require 'em-glob)
 
-(defcustom lsp-workspace-folders-changed nil
+(defcustom lsp-workspace-folders-change nil
   "Hooks to run after the folders has changed.
 The hook will receive two parameters list of added and removed folders."
   :type 'hook
@@ -891,7 +893,7 @@ FILE-NAME the file name."
       (puthash dir lsp--cur-workspace lsp--workspaces))
 
     (lsp--update-folders current-folders)
-    (run-hook-with-args 'lsp-workspace-folders-changed directories nil)))
+    (run-hook-with-args 'lsp-workspace-folders-change directories nil)))
 
 (defun lsp-workspace-folders-remove (directories)
   "Remove DIRECTORIES to the list of workspace folders.
@@ -924,7 +926,7 @@ remove."
       (setq current-folders (delete dir current-folders))
       (remhash dir lsp--workspaces))
     (lsp--update-folders current-folders))
-  (run-hook-with-args 'lsp-workspace-folders-changed nil directories))
+  (run-hook-with-args 'lsp-workspace-folders-change nil directories))
 
 (defun lsp-workspace-folders-switch()
   "Switch to another workspace folder from the current workspace."
@@ -996,7 +998,7 @@ remove."
                          :rootPath ,root
                          :rootUri ,(lsp--path-to-uri root)
                          :capabilities ,(lsp--client-capabilities)
-                                :initializationOptions ,extra-init-params-resolved)
+                         :initializationOptions ,extra-init-params-resolved)
        response (lsp--send-request
                  (lsp--make-request "initialize" init-params))
 
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -24,6 +24,12 @@
 (require 'inline)
 (require 'em-glob)
 
+(defcustom lsp-workspace-folders-changed nil
+  "Hooks to run after the folders has changed.
+The hook will receive two parameters list of added and removed folders."
+  :type 'hook
+  :group 'lsp-mode)
+
 (defconst lsp--file-change-type
   `((created . 1)
     (changed . 2)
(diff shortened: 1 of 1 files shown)


vibhavp:
Thanks for the PR! Unfortunately, I've been a little busy with real life, so it might take some time before I get to review this PR.

yyoncho:
@vibhavp sure, I will extend the PR description with more information on how it is going to work to help you with the reviewing.

yyoncho:
Updated the fix - added some more explanation in the commit description. Fixed corner cases related to deletion of the last element. I have also renamed the function `lsp--get-project-root` to `lsp--suggest-project-root` because the name should not imply that the method returns the proper project root.

yyoncho:
@vibhavp did you had a chance to take a look?

yyoncho:
Rebased onto master and removed formatting changes to avoid constant merging.

yyoncho:
@MaskRay - can you take a look, it looks like @vibhavp is pretty busy these days?

lsp-methods.el
```diff
@@ -855,16 +973,39 @@ directory."
...
+       init-params `(:processId ,(emacs-pid)
                          :rootPath ,root
                          :rootUri ,(lsp--path-to-uri root)
                          :capabilities ,(lsp--client-capabilities)
-                         :initializationOptions ,(if (functionp extra-init-params)
-                                                     (funcall extra-init-params lsp--cur-workspace)
-                                                   extra-init-params)))
-      (setf response (lsp--send-request
-                      (lsp--make-request "initialize" init-params)))
+                                :initializationOptions ,extra-init-params-resolved)
```
MaskRay: indent this line

lsp-methods.el
```diff
@@ -855,16 +973,39 @@ directory."
...
+
+      (setf
+       extra-init-params-resolved (if (functionp extra-init-params)
+                                      (funcall extra-init-params lsp--cur-workspace)
+                                    extra-init-params)
+       ;; join the saved workspace folders + the folders comming from the
+       ;; language extension to keep backward compatibility with `lsp-java'
+       workspace-folders (append
+                          (lsp--read-from-file
+                           (concat (file-name-as-directory root) ".folders"))
```
MaskRay: Is it conventional to use `$root/.folders`?

lsp-methods.el
```diff
@@ -819,19 +829,127 @@ directory."
     (cl-notany (lambda (p) (string-match-p p root))
       lsp-project-blacklist)))
 
+(defun lsp--workspace-folders-capability-p ()
+  "Return whether current server supports workspace folders."
+  (when-let (workspace (gethash "workspace" (lsp--server-capabilities)))
+    (when-let (capability (gethash "workspaceFolders" workspace))
```
MaskRay: `when-let*` ?

lsp-methods.el
```diff
@@ -24,6 +24,12 @@
 (require 'inline)
 (require 'em-glob)
 
+(defcustom lsp-workspace-folders-changed nil
```
MaskRay: How about `on-workspace-folders-change`? `workspace-folders-changed`

MaskRay:
Is https://github.com/emacs-lsp/lsp-mode/pull/356 superseded by this PR? You may also ask @samrayleung to review

I don't know/write much elisp...

lsp-methods.el
yyoncho: I am open to alternative suggestions.

lsp-methods.el
yyoncho: Sure, will update the PR.

lsp-methods.el
yyoncho: Just checked out when-let* is not available in emacs 25(unless I am mistaken).

lsp-methods.el
MaskRay: @yyoncho You might think I can merge PRs in this repo but I can't.. The most I can give you is an approval..

stardiviner:
I have an question, does this PR add a feature that user can get lsp-mode works without `lsp--cur-workspace`? For example, I want to get code completion with `company-lsp` in Org Mode source code opened dedicated buffer.

yyoncho:
@stardiviner no. Can you share which languages do you want to use in org-mode buffers?

stardiviner:
@yyoncho lsp-python and lsp-php and lsp-java.

yyoncho:
@stardiviner do you know that should be done on `org-mode` side to do the integration? I am looking at https://stackoverflow.com/questions/14704077/how-to-enable-auto-complete-in-emacs-org-babel is this how it is supposed to work?

stardiviner:
Org Mode seems can't implement this. The answers does not work in this case. The first selected answer use `narrow` to create a new indirect buffer that can change `major-mode`. But will mess up original buffer syntax highlighting (font-locking?). The second answer need the Org source block language support session. Also this session support are usually in REPL buffer.

If `lsp-mode` can't support completion without workspace. I think my feature request is not able to implemented.

Sorry for distribution.

yyoncho:
@stardiviner I believe that it can work by providing session properties, but I need working example to start with and I can provide a sample which works with LSP mode.

MaskRay:
For C/C++ lsp-mode users who want to try out this PR, I have implemented workspace folders in ccls in the weekend.

I'm down for this PR as long as there is still some way to enable automatic root detection which is convenient for browsing random new projects.

yyoncho:
@MaskRay can you elaborate?  This CL does not break the current behaviour(ability to browse random new projects), it only adds an option to call add/remove folders for a particular workspace.

inspell:
This issue is PITA for `lsp-java`. Could someone from `lsp-mode` take a look?

vibhavp:
LGTM, Apologies for the late review. Thanks!

PR #410, opened 2018-08-24
Add lsp-find-custom and deduplicate textDocument/{implementation,typeDefinition}

MaskRay:
This is similar to `lsp-ui-peek-find-custom` in lsp-ui and provides xref interface of custom cross references (e.g. `$ccls/callers` `$ccls/vars` ...)


vibhavp:
Looks good, thanks.

PR #411, opened 2018-08-25
Fixed #400 .

### LINKED ISSUE #400: lsp-format-buffer error caused by lsp-methods.el - lsp-format-buffer: replace-buffer-contents
`emacs -Q`, create a new `.py` file, insert:

```python
def hello():
    pass
```

`alt + shift + :` and eval:

```emacs-lisp
(if (fboundp 'replace-buffer-contents)
    (let ((current-buffer (current-buffer)))
      (with-temp-buffer
        (insert "'''你好啊你好啊'''\ndef hello():\n    pass")
        (let ((temp-buffer (current-buffer)))
          (with-current-buffer current-buffer
            (replace-buffer-contents temp-buffer))))))
```

I got

```python
'''你好啊你好啊'''
def hello():def hello
```

With a clean `python-mode` buffer I could get the current content:

```python
'''你好啊你好啊'''
def hello():
    pass
```

nasyxx:
This fixed #400 .

In #400 

```elisp
(if (fboundp 'replace-buffer-contents)
    (let ((current-buffer (current-buffer)))
      (with-temp-buffer
        (insert "'''你好啊你好啊'''\ndef hello():\n    pass")
        (let ((temp-buffer (current-buffer)))
          (with-current-buffer current-buffer
            (replace-buffer-contents temp-buffer))))))
```

get 

```
'''你好啊你好啊'''
def hello():def hello
```

which isn't a correct answer.

In my fixing, 

```elisp
(if (fboundp 'replace-buffer-contents)
    (let* ((current-buffer (current-buffer))
           (temp-buffer (get-buffer-create "*lsp-format*")))
      (with-current-buffer temp-buffer
        (insert "'''你好啊你好啊'''\ndef hello():\n    pass")
        (copy-to-buffer current-buffer (point-min) (point-max))
        (kill-buffer temp-buffer))))
```

will get the answer

```
'''你好啊你好啊'''
def hello():
    pass
```

which is the excepted one.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1878,13 +1878,13 @@ Optionally, CALLBACK is a function that accepts a single argument, the code lens
                                    "textDocument/formatting"
                                    (lsp--make-document-formatting-params)))))
     (if (fboundp 'replace-buffer-contents)
-        (let ((current-buffer (current-buffer)))
-          (with-temp-buffer
+        (let* ((current-buffer (current-buffer))
+               (temp-buffer (get-buffer-create "*lsp-format*")))
+          (with-current-buffer temp-buffer
             (insert-buffer-substring-no-properties current-buffer)
             (lsp--apply-text-edits edits)
-            (let ((temp-buffer (current-buffer)))
-              (with-current-buffer current-buffer
-                (replace-buffer-contents temp-buffer)))))
+            (copy-to-buffer current-buffer (point-min) (point-max))
+            (kill-buffer temp-buffer)))
       (let ((point (point))
             (w-start (window-start)))
         (lsp--apply-text-edits edits)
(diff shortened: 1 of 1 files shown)


nasyxx:
Now, I am not sure if my change is a good idea.

vibhavp:
IIRC, not using `replace-buffer-contents` had its own problems.

nasyxx:
It seems that this problem has been solved.

yssource:
The issue still happens to me with the latest source.

In short, it still doesn't work with non-ascii text.

For instance.

``` python
def main():
    print('你好, 世界！')
    

if __name__ == '__main__':
    main()

```

After `lsp-format-buffer`,  it mess up the whole codes.

``` python
def main():
    print('你好, 世界！')
    

if _e__ == '__main__': # <--, here, changed to be strange code, _e__
    main()

```

PR #416, opened 2018-09-07
Add missing CompletionItemKind constants and use vector for performance

MaskRay:
With this change, in [ccls](https://github.com/MaskRay/ccls),


vibhavp:
Looks good, thanks.

Issue #390, opened 2018-07-03
lsp--diagnostics should be per workspace (was: shouldn't the lsp--diagnostics variable be defvar-local?)


Havner:
As in title. I've found this variable to gain weight pretty significantly on large projects (measured even in GBs). Closing the projects doesn't reclaim this memory. And its description suggests it's per file.
I've marked it local for the test and it seems to work fine.
I'm not sure that is correct, but after I switched to using LSP (with cquery) I've found my Emacs to consume lots of memory at times. And the culprit is this particular variable.

yyoncho:
`lsp--diagnostigs` is used by `lsp-ui-flycheck` when rendering errors it should not be buffer local. But there is an issue indeed - `lsp--diagnostigs` must be part of `lsp--client` or `lsp--workspace` structures and it should have the same lifetime as the workspace.

Havner:
Yeah, I know it's used for that but I still don't get why it can't be buffer local. Flycheck errors are per file/buffer.

yyoncho:
It stores the errors for all of the files in the workspace. Actually in any workspace.

Havner:
And why does flycheck need errors from other buffers to display them? :-)

yyoncho:
Sorry for being unclear - the errors are reported for all of the buffers by LSP server. So when you switch to a buffer it displays the errors from that buffer(the already reported one). As a side note, flycheck is capable of displaying errors from multiple buffers.

Havner:
Ok, got it, so for the first point it goes asynchronously. Thanks for explanation.
Anyway, the issue for high memory usage still remains.
This also is especially prominent for big repos where there is no compile_commands.json and presumably there is lots of errors due to missing headers.
I've opened kernel source once this way and emacs memory usage went to 6GB before I had to kill it. And it was still growing. Making this variable per workspace won't fix that.

D4N:
> Anyway, the issue for high memory usage still remains.

I can confirm this, as I have recently tried opening the Kernel version 2.6.32 sources (which wouldn't compile with gcc 8 and produced a bazillion of warnings). Emacs with lsp-mode was consuming about 5-10 times more memory then cquery was, before I had to kill emacs. I was only able to use cquery on the kernel with eglot (but uses flymake instead of flycheck).

vibhavp:
Making it buffer local now would probably break API compatibility with packages using previous package version. Introducing a new variable `lsp--buffer-diagnostics` would be a better idea, IMO.

[...]

PR #417, opened 2018-09-07
Add option to keep the diagnostics if there is no open buffer

yyoncho:
- Whithout this CL you won't be able to check the errors on project level.

### DIFF
--- a/lsp-notifications.el
+++ b/lsp-notifications.el
@@ -24,6 +24,11 @@
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-report-if-no-buffer t
+  "If non nil the errors will be reported even when the file is not open."
+  :type 'boolean
+  :group 'lsp-mode)
+
 (defun lsp--window-show-message (params &optional workspace)
   "Send the server's messages to message, inhibit if `lsp-inhibit-message' is set,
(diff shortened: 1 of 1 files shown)


lsp-notifications.el
```diff
@@ -24,6 +24,11 @@
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-report-if-no-buffer t
```
vibhavp: Shouldn't the default value be `nil`, since that's what the current behaviour is?

yyoncho: The original behaviour was to preserve the errors but it was changed in https://github.com/emacs-lsp/lsp-mode/commit/d17c55448c26c8475a5e38354ca42d7875b239b4 and this CL is actually redoing this commit. IMO we should keep it as it is since https://github.com/emacs-lsp/lsp-mode/commit/d17c55448c26c8475a5e38354ca42d7875b239b4 was introduced as a potential workaround of https://github.com/emacs-lsp/lsp-mode/issues/390 and it should not be the default behaviour. See also https://github.com/emacs-lsp/lsp-mode/issues/390#issuecomment-416924475

vibhavp: Ah, I didn't realize this was breaking `lsp-ui`. I'll merge this, then.

Issue #210, opened 2017-12-13
Use native JSON functionality in Emacs 27


vibhavp:
Emacs 27.1 recently introduced a native JSON parser using the Jansson library. The entrypoints functions are `json-serialize` and `json-parse-string`, but they only work with hash tables and vectors. 

Profiler data shows that maximum percentage of CPU time is spent in `json-read-from-string` and `json-encode` provided by `json.el`.

We can introduce a "polyfill" library that uses the native JSON implementation when available, and reverts to `json.el` otherwise.

sebastiencs:
I've seen those features too and I have started to working on it.
Have you already write some code ? 
As you say, the major issue is that it uses hash tables for object and vectors for arrays

phst:
Note: we can't use lists for both objects and arrays because they are ambiguous (every alist is also a normal list). My plan is to allow alists for objects, but continue requiring vectors for arrays.

luxbock:
I'd like to 👍 this effort, as the LSP provided auto-complete is at times unusable for me due to input lag, and profiling seems to point at JSON parsing being the culprit.

Here is the relevant part from profiler output:
```
- company-post-command                                              19  45%
 - company--perform                                                 19  45%
  - company--continue                                               18  42%
   - company-calculate-candidates                                   18  42%
    - company--fetch-candidates                                     18  42%
     - company-call-backend-raw                                     18  42%
      - apply                                                       18  42%
       - company-capf                                               18  42%
        - completion-all-completions                                18  42%
         - completion--nth-completion                               18  42%
          - completion--some                                        18  42%
           - #<compiled 0xdae825>                                   18  42%
            - completion-pcm-all-completions                        13  30%
             - completion-pcm--find-all-completions                 13  30%
              - completion-pcm--all-completions                     13  30%
               - all-completions                                    13  30%
                - #<compiled 0x2867681>                             13  30%
                 - #<compiled 0x4305719>                            13  30%
                  - lsp--send-request                               13  30%
                   - lsp--make-message                              8  19%
                    - json-encode                                   8  19%
                     - json-encode-list                             8  19%
                      - json-encode-plist                           8  19%
                       - json-encode                                8  19%
                        - json-encode-list                          8  19%
                         - json-encode-plist                        8  19%
                          - json-encode                             8  19%
                           - json-encode-array                      7  16%
                            - mapconcat                             7  16%
                             - json-encode                          7  16%
                              - json-encode-list                    7  16%
[...]

PR #418, opened 2018-09-09
Use generic sequence functions to accommodate future json-array-type-> 'vector change

MaskRay:
This also helps https://github.com/emacs-lsp/lsp-mode/issues/210


vibhavp:
Looks good, thanks!

PR #422, opened 2018-09-11
Highlight active parameter with face eldoc-highlight-function-argument

MaskRay:
```cpp
int foo(int a, int b, int c = 7) {
  return a+ b + c;
}

int main() {
  foo(1, 2<point>, 3);
}
```

In ccls, this was rendered as `foo(int a, int b, int c = 7) -> int` before, and now `foo(int a, <strong>int b</strong>, int c = 7) -> int`.

Similar approaches are used by

LanguageClient-neovim src/languageclient.rs

eglot (but it trims the signature label `  (propertize (replace-regexp-in-string "(.*$" "(" label)` which will remove `-> return_type`)

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -1757,13 +1757,15 @@ type MarkupKind = 'plaintext' | 'markdown';"
     (when (and signature-help
                (lsp--point-is-within-bounds-p start end)
                (eq (current-buffer) buffer) (eldoc-display-message-p))
-      (let* ((active-signature-number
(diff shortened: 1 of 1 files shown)


MaskRay:
`(defun lsp--text-document-signature-help () ...) only takes effect when `(symbol-at-point)` is not `nil`. This does not work well for `foo(arg0, <point>` (when there is no symbol in the argument position)

vibhavp:
Thanks.

PR #425, opened 2018-09-15
Add lsp--json-array-use-vector. If set, use default json-array-type vector and use json-parse-string if exists

MaskRay:
https://github.com/emacs-lsp/lsp-mode/issues/210

It works fine with [lsp-ui](https://github.com/emacs-lsp/lsp-ui) + [emacs-ccls](https://github.com/MaskRay/emacs-ccls)

### DIFF
--- a/lsp-io.el
+++ b/lsp-io.el
@@ -28,6 +28,16 @@
   :type 'number
   :group 'lsp-mode)
 
+(defcustom lsp--json-array-use-vector nil
+  "Use vectors to represent JSON arrays.
+
+If set, native JSON serialization will be used if available.
+
+This option exists to port existing plugins that rely on JSON arrays serialized as lists.
+It will be removed and vectors will be used by default in the future."
+  :type 'boolean
+  :group 'lsp-mode)
+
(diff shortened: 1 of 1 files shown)


vibhavp:
While this would definitely move us to use the native JSON API, it would also potentially break other packages that use lsp-mode (other than `lsp-ui` and `ccls`). Adding a client local variable `lsp--client-use-native-json` will help maintain compatibility.

MaskRay:
Added
```
(defcustom lsp--json-array-use-vector nil
  "Use vectors to represent JSON arrays.

If set, native JSON serialization will be used if available.

This option exists to port existing plugins that rely on JSON arrays serialized as lists.
It will be removed and vectors will be used by default in the future."
  :type 'boolean
  :group 'lsp-mode)
```

MaskRay:
Ping?

vibhavp:
Sorry for the late reply, but this would make it a global setting, which would still break some clients. I was suggesting a client-local option for using native JSON, something I tried in this PR: https://github.com/emacs-lsp/lsp-mode/pull/435

PR #427, opened 2018-09-16
Added helper macros for executing code in workspace context

yyoncho:


vibhavp:
Thanks, these definitely look pretty useful.

PR #429, opened 2018-09-24
Handle set-buffer-file-name

### LINKED ISSUE #413: `write-file` is not handled correctly.
How to reproduce:

1. Set `lsp-print-io` to t.
2. Open a file that will be handled by lsp-mode.
3. `C-x C-w`  to write file with a different name.

Expected: lsp-mode should send `didClose` to close the previous file, and send `didOpen` to open the new file.

Actual: lsp-mode sends `didSave` for the new file only.

tigersoldier:
Add an advice around it. Send didClose for the old file and didOpen for the new
file. This fixes #413


vibhavp:
Thanks!

PR #431, opened 2018-09-29
Fix performance issue related to using lsp--equal-files

yyoncho:
- I did performance testing I noticed that 25%(!) of the time is spent in
lsp--equal-files.

``` emacs-lisp
(with-current-buffer "PoolSubpage.java"
  (let ((file (lsp--uri-to-path (gethash "uri" params))))
    (message "Perf = %s" (benchmark-run 100 (find-buffer-visiting file)))
    (message "Perf = %s"(benchmark-run 100
       (cl-loop for buffer in (lsp--workspace-buffers lsp--cur-workspace)
                when (lsp--equal-files (buffer-file-name buffer) file)
                return buffer
                finally return nil)))))
```
Results in:
```
Perf = (0.073634171 0 0.0)
Perf = (4.673262101 1 0.3096860720000052)
```

### DIFF
--- a/lsp-imenu.el
+++ b/lsp-imenu.el
@@ -111,14 +111,13 @@ SYM can be either DocumentSymbol or SymbolInformation."
 
 (defun lsp--symbol-filter (sym)
   "Determine if SYM is for the current document."
-  (if-let ((location (gethash "location" sym)))
-      ;; It's a SymbolInformation
-      (not
-       (lsp--equal-files
-        (lsp--uri-to-path (gethash "uri" (gethash "location" sym)))
-        (buffer-file-name)))
-    ;; It's a DocumentSymbol, which is always in the current buffer file.
-    nil))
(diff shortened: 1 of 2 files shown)


lsp-imenu.el
```diff
@@ -111,14 +111,13 @@ SYM can be either DocumentSymbol or SymbolInformation."
...
-       (lsp--equal-files
-        (lsp--uri-to-path (gethash "uri" (gethash "location" sym)))
-        (buffer-file-name)))
-    ;; It's a DocumentSymbol, which is always in the current buffer file.
-    nil))
+  ;; It's a SymbolInformation or DocumentSymbol, which is always in the current
+  ;; buffer file.
+  (when-let (location (gethash "location" sym))
+    (not
+     (equal
```
MaskRay: LG. I think `eq` can be used here...

yyoncho:
Applied comments.

PR #437, opened 2018-10-07
Clean diagnostics map when there is no errors for the buffer

yyoncho:


lsp-notifications.el
```diff
@@ -112,8 +112,10 @@ interface PublishDiagnosticsParams {
                           return buffer
                           finally return nil))
 
-    (when (or lsp-report-if-no-buffer buffer)
-      (puthash file (mapcar #'lsp--make-diag diagnostics) lsp--diagnostics))
+    (if diagnostics
```
MaskRay: WoW. Do you have idea why I can see lingering diagnostics if I `(setq lsp-ui-flycheck-live-reporting nil)`?

yyoncho: I have never used that option, I will check it at let you know.

vibhavp:
Looks good, thanks.

PR #441, opened 2018-10-15
lsp--apply-text-edits: support vectors and fix comment

MaskRay:


lsp-methods.el
```diff
@@ -1222,17 +1222,16 @@ interface TextDocumentEdit {
...
-
-    (seq-do #'lsp--apply-text-edit (sort (nreverse ,edits) #'lsp--text-edit-sort-predicate))))
+  ;; We sort text edits so as to apply edits that modify latter parts of the
+  ;; document first. Furthermore, because the LSP spec dictates that:
+  ;; "If multiple inserts have the same position, the order in the array
+  ;; defines which edit to apply first."
+  ;; We reverse the initial list and sort stably to make sure the order among
+  ;; edits with the same position is preserved.
+  (mapc #'lsp--apply-text-edit
+        (sort (nreverse (seq-into edits 'list)) #'lsp--text-edit-sort-predicate)))
```
tigersoldier: Do you need to call `seq-into` here? IMO the functions (`mapc`, `sort`, `nreverse`) support sequences.

lsp-methods.el
```diff
@@ -1585,10 +1584,10 @@ https://microsoft.github.io/language-server-protocol/specification#textDocument_
   (lsp--gethash "sortText" c (gethash "label" c "")))
 
 (defun lsp--sort-completions (completions)
-  (sort completions (lambda (c1 c2)
-                      (string-lessp
-                        (lsp--sort-string c1)
-                        (lsp--sort-string c2)))))
+  (sort (seq-into completions 'list) (lambda (c1 c2)
```
tigersoldier: Is it better to call `seq-into` later (i.e. `(seq-into (sort ...))`).

Not sure about the performance difference though.

lsp-methods.el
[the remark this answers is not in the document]
MaskRay: Changed to `seq-reverse`.

I didn't know `sort` can be used on vectors. `sort` on vectors is faster as the merge sort can be done in-place (`fns.cn`)

MaskRay: Thanks.

Changed to `  (mapc #'lsp--apply-text-edit
        (sort (nreverse edits) #'lsp--text-edit-sort-predicate))
`

Issue #443, opened 2018-10-16
Setting lsp-eldoc-render-all to nil can still show multi-line text


wberrier:
If I customize `lsp-hover-text-function` to use `lsp--text-document-signature-help` I don't get any help documentation.

I've verified this with lsp-python, cquery, and lsp-java

Anyone having `lsp--text-document-signature-help` working?

This is how I have it configured:

```
(use-package lsp-mode
  :custom
  (lsp-hover-text-function (quote lsp--text-document-signature-help))
  )
```

It also doesn't work if I use M-x customize-group and set it through the menus.

yyoncho:
I could speak from `lsp-java` perspective - the JDT server returns the signature only if the code is in () but the `lsp--text-document-signature-help` sends requests only if there is symbol at point. My current workaround is to move the cursor over the function name. `vscode-java` does not show the signature too. It would be good if someone does some research what is the proper way to implement that feature on lsp-mode side since having both signature help and hover does not feel correct.

wberrier:
In checking the protocol interaction between language servers for c++, java, and python, the signatureHelp returns empty results when hovering.

I think the answer I'm looking for is to set `'lsp-eldoc-render-all` to `nil`.  This works great for the java language server, but not for cquery.  The reason is, for the java language server, textDocument/hover returns:

```
Output from language server: {
  "jsonrpc": "2.0",
  "id": 8,
  "result": {
    "contents": [
      {
        "language": "java",
        "value": "void java.io.PrintStream.println(String x)"
      },
      " *  #### println ####\n    \n        public void println(String x)\n    \n    Prints a String and then terminate the line. This method behaves as though it invokes `print(String)` and then `println()`.\n    \n    Parameters:\n    \n    `x` \\- The `String` to be printed."
    ]
  }
}
```

And cquery/c++ language server returns:

```
Output from language server: {
  "jsonrpc": "2.0",
  "id": 84,
  "result": {
    "contents": [
      "\nmatch condition to make sure buffer has at least specified bytes\n\nIntended to be used with async_read_until\n\nConvenient if a buffer is used to read unknow size (ie: http\nheaders) and then known size (ie: content-length body)\n\nOtherwise, the first buffer read may have some data in it, and\nbefore doing the next read, you have to check if the buffer has the\ndata in it already.  Otherwise, you risk doing another read when\nactually all the data has been transferred already.\n\nIt's surprising that asio doesn't provide this match condition already",
      {
        "language": "cpp",
[...]

PR #445, opened 2018-10-18
Handle pyls output when lsp-eldoc-render-all is nil

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/443


wberrier:
Nice!

yyoncho:
@MaskRay check https://github.com/palantir/python-language-server/issues/439

PR #447, opened 2018-10-21
Refactor lsp--start

MaskRay:
* Deduplicate lsp--enable-tcp-client and lsp--enable-stdio-client
* Allow :get-root to be nil
* Root is now: (or (when root-fn (funcall root-fn))
                   (lsp--suggest-project-root))
  That is, :get-root in lsp-* adapters is used to override projectile/project roots.
  But most adapters can leave :get-root unspecified to get a sensible value from projectile/project.
  If root is nil (you probably need (setq projectile-require-root t)), the adapter will not initialize the workspace.

* Don't suggest default-directory in lsp--suggest-project-root.
  Use (or (lsp--suggest-project-root) default-directory) instead in workspaceFolders

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -25,6 +25,8 @@
 (require 'pcase)
 (require 'inline)
 (require 'em-glob)
+(unless (version< emacs-version "26")
+  (require 'project))
 
 (defcustom lsp-workspace-folders-change nil
   "Hooks to run after the folders has changed.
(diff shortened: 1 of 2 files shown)


MaskRay:
Most `lsp-*` should delete their brittle `:get-root` functions as projectile/project does a much better job to detect the root.

lsp-methods.el
```diff
@@ -858,11 +859,11 @@ directory."
...
 (defun lsp--suggest-project-root ()
-  "Caculates project root or fallbacks to the current directory."
-  (cond
-   ((and (featurep 'projectile) (projectile-project-p)) (projectile-project-root))
-   ((vc-backend default-directory) (expand-file-name (vc-root-dir)))
-   (t default-directory)))
+  "Get project root."
+  (or
+   (and (featurep 'projectile) (projectile-project-root))
+   (when-let ((project (project-current)))
```
yyoncho: project.el is emacs 26 specific, you should use it conditionally.

PR #449, opened 2018-10-23
Implement textDocument/onTypeFormatting

### LINKED ISSUE #430: Support textDocument/onTypeFormatting
[ccls](https://github.com/MaskRay/ccls) and [clangd](https://clang.llvm.org/extra/clangd.html) support `textDocument/onTypeFormatting`

```javascript
{
  "documentFormattingProvider":true,
  "documentRangeFormattingProvider":true,
  "documentOnTypeFormattingProvider":{"firstTriggerCharacter":"}","moreTriggerCharacter":[]}
}
```

When `}` is typed, they will find the closest `{` and format the enclosed region.

Regarding `lsp-format-region`, it'll also be nice to have Evil visual-state integration of `textDocument/rangeFormatting`.

MaskRay:
Close #430

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -382,6 +382,11 @@ the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-enable-on-type-formatting t
+  "Enable `textDocument/onTypeFormatting' integration."
+  :type 'boolean
+  :group 'lsp-mode)
+
 ;;;###autoload
 (defcustom lsp-before-save-edits t
   "If non-nil, `lsp-mode' will apply edits suggested by the language server
@@ -1078,6 +1083,7 @@ remove."
   ;; Make sure the hook is local (last param) otherwise we see all changes for all buffers
   (add-hook 'before-change-functions #'lsp-before-change nil t)
   (add-hook 'after-change-functions #'lsp-on-change nil t)
+  (add-hook 'post-self-insert-hook #'lsp--on-self-insert nil t)
   (add-hook 'after-revert-hook #'lsp-on-revert nil t)
   (add-hook 'before-save-hook #'lsp--before-save nil t)
   (add-hook 'auto-save-hook #'lsp--on-auto-save nil t)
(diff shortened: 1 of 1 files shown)


lsp-methods.el
```diff
@@ -1428,6 +1435,17 @@ Added to `after-change-functions'."
                                         start end length)))
                  ('full (vector (lsp--full-change-event))))))))))))
 
+(defun lsp--on-self-insert ()
+  (when-let* ((provider (and lsp-enable-on-type-formatting
+                             (lsp--capability "documentOnTypeFormattingProvider")))
+              (ch (gethash "firstTriggerCharacter" provider))
+              (_ (eq (string-to-char ch) last-command-event)))
+    (lsp--apply-text-edits
```
yyoncho: this must be async otherwise the typing will be blocked if the server is busy and it cannot provide the formatting.

yyoncho: To elaborate a bit more - the same problem exists in onDidSave - emacs blocks when you save a file and server is busy.

vibhavp: Is async formatting feasible? If new text has been inserted into the document between the time `lsp-mode` requests `textDocument/onTypeFormatting` and gets a response, applying an edit might mess up the newly inserted contents.

yyoncho: The text edits have a version, so if the text document version has changed the edit won't be applied. We can also display a spinner when there is async call to the server. VScode has only async calls so it should be possible.

yyoncho: For the record, here it is vscode implementation:
```js
	private trigger(ch: string): void {

		if (this.editor.getSelections().length > 1) {
			return;
		}

		var model = this.editor.getModel(),
			position = this.editor.getPosition(),
			canceled = false;

		// install a listener that checks if edits happens before the
		// position on which we format right now. If so, we won't
		// apply the format edits
		var unbind = this.editor.onDidChangeModelContent((e) => {
			if (e.isFlush) {
				// a model.setValue() was called
				// cancel only once
				canceled = true;
				unbind.dispose();
				return;
			}

			for (let i = 0, len = e.changes.length; i < len; i++) {
				const change = e.changes[i];
				if (change.range.endLineNumber <= position.lineNumber) {
					// cancel only once
					canceled = true;
					unbind.dispose();
					return;
				}
			}

		});

		let modelOpts = model.getOptions();

		getOnTypeFormattingEdits(model, position, ch, {
			tabSize: modelOpts.tabSize,
			insertSpaces: modelOpts.insertSpaces
		}).then(edits => {
			return this.workerService.computeMoreMinimalEdits(model.uri, edits);
		}).then(edits => {

			unbind.dispose();

			if (canceled || isFalsyOrEmpty(edits)) {
				return;
			}

			EditOperationsCommand.executeAsCommand(this.editor, edits);
			alertFormattingEdits(edits);

		}, (err) => {
			unbind.dispose();
			throw err;
		});
	}
```

MaskRay: I don't know how to do this asynchronously and I think formatting is an operation that is fast enough. In reality, very few language servers (only ccls and clangd) support the feature

yyoncho: Thanks!

PR #452, opened 2018-10-24
Add mode as an optional parameter to lsp--send-request-async

MaskRay:


lsp-methods.el
```diff
@@ -1882,18 +1883,15 @@ RENDER-ALL if set to nil render only the first element from CONTENTS."
...
 ;; start and end are the bounds of the symbol at point
-(defun lsp--make-hover-callback (client start end buffer)
+(defun lsp--make-hover-callback (client start end)
   (lambda (hover)
-    (with-current-buffer buffer
-      (setq lsp--cur-hover-request-id nil))
+    (setq lsp--cur-hover-request-id nil)
     (when (and hover
-               (lsp--point-is-within-bounds-p start end)
-               (eq (current-buffer) buffer) (eldoc-display-message-p))
```
MaskRay: I don't think `(eq (current-buffer) buffer) (eldoc-display-message-p)` are valuable.

lsp-methods.el
```diff
@@ -545,9 +545,13 @@ If NO-WAIT is non-nil, don't synchronously wait for a response."
   "Send BODY as a request to the language server, and call CALLBACK with
 the response recevied from the server asynchronously."
   (let ((client (lsp--workspace-client lsp--cur-workspace))
+        (buf (current-buffer))
         (id (plist-get body :id)))
     (cl-assert id nil "body missing id field")
-    (puthash id callback (lsp--client-response-handlers client))
+    (puthash id (lambda (result)
+                  (when (buffer-live-p buf)
```
yyoncho: clients might still be interested in the result even if the buffer is closed. E. g. if we do "did save handling" async the original buffer would be deleted.

yyoncho: I propose explicit optional parameter(s) which turn on/turn off this feature.

lsp-methods.el
yyoncho: what happens if I have switched to another buffer without killing the buffer? IMO eldoc-display-message-p is needed in case you are in helm session or completing-read the eldoc message should not appear.

MaskRay: Restored `(eldoc-display-message-p)`

yyoncho: I think that we should check whether the buffer is the current one. Since that there are 3 different behaviours:

1. Call callback only if the buffer was not changed and the original buffer is live(e. g. hovers)
2. Call callback if the original buffer is live(e. g. code actions)
3. Call callback no matter whether the original buffer is alive(e. g. save actions)

we can extend the lsp-request-async with a param specifying which behaviour do we want with default behaviour 3) since this is the current one. This will simplify `lsp-ui` as well.

yyoncho: As a side note if we do so we will fix `lsp-ui` bugs since it currently displays overlays in the wrong buffer in case you have switched the buffer before the server has returned.

MaskRay: Done

PR #448, opened 2018-10-22
Refactor eldoc

MaskRay:
* Rename lsp-symbol-highlight to lsp-document-highlight
  We should keep consistency with LSP names
* Delete lsp-hover-text-function, lsp-enable-eldoc, lsp-highlight-symbol-at-point in favor of lsp-eldoc-hooks
  lsp-ui-doc users are not affected.
* Add defcustom lsp-eldoc-hooks to customize requests to send (default: documentHighlight, hover)
* Use add-function instead of setq to set eldoc-documentation-function

Change on lsp-ui side: https://github.com/MaskRay/lsp-ui/tree/eldoc

PR #453, opened 2018-10-25
Extract initialize/uninitialize logic to lsp--managed-mode

MaskRay:


MaskRay:
After this is merged, I would have to rebase https://github.com/emacs-lsp/lsp-mode/pull/448 which is painful

MaskRay:
Ping?

PR #455, opened 2018-10-25
Set default-directory before make-process

felixSchl:
Shadow the global default-directory with a local let binding before calling `make-process` in order to start the subprocess in the commands `get-root` directory.

Motivation: the change was required in order to get lsp-haskell to work

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -38,10 +38,11 @@
   :lighter (:eval (lsp-mode-line))
   :group 'lsp-mode)
 
-(defun lsp--make-stdio-connection (name command command-fn stderr)
+(defun lsp--make-stdio-connection (name command command-fn get-root-fn stderr)
   (lambda (filter sentinel)
     (let* ((command (if command-fn (funcall command-fn) command))
-           (final-command (if (consp command) command (list command))))
+           (final-command (if (consp command) command (list command)))
+           (default-directory (if get-root-fn (funcall get-root-fn) default-directory)))
       (unless (executable-find (nth 0 final-command))
         (error (format "Couldn't find executable %s" (nth 0 final-command))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -38,10 +38,11 @@
   :lighter (:eval (lsp-mode-line))
   :group 'lsp-mode)
 
-(defun lsp--make-stdio-connection (name command command-fn stderr)
+(defun lsp--make-stdio-connection (name command command-fn get-root-fn stderr)
   (lambda (filter sentinel)
     (let* ((command (if command-fn (funcall command-fn) command))
-           (final-command (if (consp command) command (list command))))
+           (final-command (if (consp command) command (list command)))
+           (default-directory (funcall get-root-fn)))
```
yyoncho: get-root-fn might be null

lsp-mode.el
```diff
@@ -38,10 +38,11 @@
   :lighter (:eval (lsp-mode-line))
   :group 'lsp-mode)
 
-(defun lsp--make-stdio-connection (name command command-fn stderr)
+(defun lsp--make-stdio-connection (name command command-fn get-root-fn stderr)
   (lambda (filter sentinel)
     (let* ((command (if command-fn (funcall command-fn) command))
-           (final-command (if (consp command) command (list command))))
+           (final-command (if (consp command) command (list command)))
```
MaskRay: The root is decided in `lsp--start`. This will not work as `get-root-fn` may be nil (I have PRs to change some clients, https://github.com/emacs-lsp/lsp-python/pull/26) and I think it is recommended to use `nil`.

yyoncho:
Closing in favour of https://github.com/emacs-lsp/lsp-mode/pull/486

PR #468, opened 2018-11-03
remove executable bit from lsp-common.el

thkoch2001:


damon-kwok:
lsp-clangd:
(file-missing "Cannot open load file" "No such file or directory" "lsp-common")

damon-kwok:
lsp-python:
(file-missing "Cannot open load file" "No such file or directory" "lsp-common")

Issue #424, opened 2018-09-14
Support multiple language servers for single file/workspace


yyoncho:
LSP mode must support multiple LSP servers for single file/workspace similar to what the other implementations(Che/VSCode) already do. 

See https://github.com/eclipse/che/issues/4609

TysonAndre:
Possible duplicate of https://github.com/emacs-lsp/lsp-mode/issues/105

yyoncho:
@TysonAndre no, #105 is to support different servers for the same mode per workspace. This bug is to support multiple simultaneous servers running in the same file and providing data for the file simultaneously, e.g. running HTML and JS LSP servers in one file.

yyoncho:
@vibhavp can you share your thoughts? If there are no volunteers to implement this feature I will work on it, but this is not going to happen anytime soon. Since this would require some bit restructuring on `lsp-mode` side I want to discuss it upfront. I think that it would be good to address some of the `lsp-mode` flaws related to process management.

yyoncho:
@vibhavp @MaskRay 
I am investigating whether `sesman` could be a good fit for our usecase which is used by cider for a similar use case. I believe that it could improve the user experience related to visibility what are the currently running servers(check `sesman-browser`), it will cleanup a lot of code related to session management which complexity is going to increase. Also, it will make the writing of extension against LSP mode much easier compared to current ad-hoc session management with `lsp--workspaces` hash table. 

(cc `sesman` owner @vspinu)

TysonAndre:
Is the FAQ up to date? 

Issue 105 (Multiple clients per major mode?) was completed, **so I'm not sure why this FAQ is still there or what it now means:**
(I assumed that all of them would be run, excluding blacklisted ones)

https://github.com/emacs-lsp/lsp-mode#i-have-multiple-language-servers-registered-for-language-foo-which-one-will-be-used-when-opening-a-project

> I have multiple language servers registered for language FOO. Which one will be used when opening a project?
> 
> The one with highest priority wins. lsp-clients.el predefined servers have priority -1, lower than external packages (priority 0 if unspecified).


---

The readme says this ticket was completed, which contradicts that:

> lsp-mode now supports multiple language servers per single file and workspace.

yyoncho:
@TysonAndre 
The FAQ is not complete. When server is marked with flag `:add-on?` it will be started in parallel to the other server(s) handling that mode. Do you have particular usecase?

danilevy1212:
[...]

Issue #392, opened 2018-07-05
Be more graceful when the lsp process dies or does not exist etc.


Fuco1:
`lsp-mode` hooks into a bunch of stuff including `imenu` (maybe it comes from a different package?).  Not important though.

Every time I idle for half a second I get this error.  Why should I care as the end user.  If the server does not exist just fall back to something that works, maybe print a message in the minibuffer.  I feel that in general the error handling in this package is very suboptimal.  Lots of assertions or simply assumptions that things work, and when they don't Emacs becomes completely unusable (for example you can't even kill a buffer because lsp is trying to send the "document closed" message).

```
Debugger entered--Lisp error: (cl-assertion-failed (lsp--cur-workspace "No language server is associated with this buffer."))
  cl--assertion-failed(lsp--cur-workspace "No language server is associated with this buffer." nil nil)
  lsp--get-document-symbols()
  lsp--imenu-create-index()
  imenu--make-index-alist(t)
  which-function()
  which-func-update-1(#<window 1835 on abstract-order-method.php>)
  which-func-update()
  apply(which-func-update nil)
  timer-event-handler([t 0 0 500000 t which-func-update nil idle 0])
```

I feel I am quite harsh in my words but I don't know how to express it much better.  I am critical but doing so with love :heart: Lots of great effort has gone into this package.

As a start using `user-error` over `error` would help a great deal where it makes sense.  As an elisp developer I usually have `toggle-debug-on-error` on and so this is really really getting in the way.

vibhavp:
Yes, lsp-mode needs to be more graceful with process crashes. Automatically restarting servers (what eglot does) is a good start.

Issue #376, opened 2018-06-11
[Improvement] make single entry point for lsp-mode


yyoncho:
Currently, every extension has its own entry point. We could change the code to start the proper LS server based on the major mode via generic method like `lsp-start`.

gpittarelli:
Any thoughts on how best to define the mapping from major mode <-> lsp client definition? Maybe the client's language id.

More generally, I'm already working on something that'll work with the current many-packages ecosystem and basically just aim to automate the "typical case" setup (potentially even automating installing the appropriate lsp-(language) package and downloading the necessary language server executables). I hope to get it usable and published over this weekend. It may need to be yet another separate mode (in order not to pollute the main lsp-mode package with URLs of external, potentially non-GPL binaries).

yyoncho:
I was thinking about extending `lsp-define-*-client` to include major mode(vscode is using file pattern + language id) or a function like `lsp-are-you-interested-in-that-file` but I haven't really researched the topic/options. Eglot is doing that via major-mode in define client function.
Snippet from vscode configuration:
```json
    "languages": [
      {
        "id": "java",
        "extensions": [
          ".class"
        ],
        "configuration": "./language-configuration.json"
      }
    ],
```
Few more things to consider(not sure whether it is related to your effort):

1. Multi root workspaces - I have implemented adhoc multiroot workspace handling in `lsp-java` but it is really ugly - https://github.com/emacs-lsp/lsp-java/blob/master/lsp-java.el#L489 . I am not sure how many of the lsp servers are multiroot ready but at some point it will become a problem.
2. Root detection - IMO LSP mode should never automatically "assume" the root folder but only provide a suggestion due to multiroot support and multi language workspaces. Also, the decision whether to fire up new LSP server must be in the language extension, not in `lsp-mode`(I have ugly workarounds for this in `lsp-java` too).

seagle0128:
Vote as well.
[eglot](https://github.com/joaotavora/eglot) and [flycheck](https://github.com/flycheck/flycheck) do the similar things in the package, so that they are out of box to use.

alyssackwan:
Please consider #393 and #397. This should be project specific.

yyoncho:
See also: https://github.com/emacs-lsp/lsp-mode/issues/424

yyoncho:
[...]

PR #469, opened 2018-11-10
Multi LSP Server support

### LINKED ISSUE #225: Reuse LSP session of "caller" file after jumping to system header file (or files outside of the project)
Currently `lsp--start` calls `(lsp--client-get-root client)` to associate a buffer with a workspace. This is fine with files with a common path (usually the project root). However, this get-root model is insufficient for files outside of the current project (e.g. system headers).

In cquery, `xref-find-definitions` on `#include <assert.h>` visits `assert.h`, then I lose the LSP session in the new buffer and I can no longer use LSP features. We should have some "dynamic scoping" sense here as a file can inherit the sessions from its "dynamic chain" (caller file).

VSCode supports [Multi-root Workspaces](https://code.visualstudio.com/docs/editor/multi-root-workspaces) which I wonder if more or less similar.

yyoncho:
Fixes #225 #424 #376 #360 #280 #285 #264 #229 #117 #111 #335 #392 #390 #359 #340 #225

Higher level goals:

* Allow multiple servers to work in one project and in one file
* Make finding of project roots easier and more explicit for the users.
* Removed all synchronous calls from the server startup.
* Improved hadling of status messages.
* Implemented workspace folders support as a first class citisen instead of
  being a patch in the existing configuration.
* Simplified the code as much as possible

The following things are implemented:

1. Moved `lsp-mode` related code into single file. Now clients should either
require `lsp-mode` or `lsp`. After this CL `lsp-mode.el` and `lsp.el` extesions
like `ccls` and `lsp-ui` will be able to work against either of these.
`lsp-mode` will be used by default and if you want to use the new `lsp.el` you
should do `(require 'lsp)` before loading lsp-mode.

2. Introduced `lsp-session` which will be resposible for handling project
session configuration like:
* What currently imported projects.
* What server are running and the folders that are associated with the session.
* What folders are ignored and wont be used to initialize a project in it. The
session is persisted and automatically loaded the first time user calls `lsp`

3. Introduced `lsp-language-id-configuration` which will hold the configuration
"language" -> emacs major mode. This map is used to render eldoc information and
also to determine the language id when sending various notifications to the
server.

4. Restructored `lsp--client` by removing the runtime related
stuff (like last-id) so it can be used as a template for registration. This will
simplify the way lsp client registration and it will make easier to add new
functionality

5. lsp-ui and company-lsp should be adapted as well. I will prepare separate CLs.

6. Implemented frontend for reviewing what servers are running, which buffers
are managed, and so on the method have replaced the existing `lsp-capabilities`
method.

7. Reworked the modeline so it can provide info for the servers currently
handling current file.

8. Moved snippet configuration into `lsp-mode` from `company-lsp` since snippet
support will be needed as part of implementing snippet support in lsp-mode #350.
Aa a result it was possible to clean registration related methods since they
where used only by company-lsp.

9. Added the option each client to define the location of its libraries and thus
avoid starting LSP servers for library folders.

10. Method selection based on the configuration in `lsp-method->capabilities`.

Sample client implementation:

``` emacs-lisp
(lsp-register-client
 (make-lsp-client
  :new-connection (lsp-stdio-connection "pyls")
  :major-modes '(python-mode)
  :server-id 'pyls
  :library-folders-fn (lambda (_workspace)
                        (list "/usr/"))))
```

### DIFF
--- a/lsp.el
+++ b/lsp.el
@@ -25,14 +25,104 @@
 
 ;;; Code:
 
+(unless (version< emacs-version "26")
+  (require 'project))
+
+(eval-when-compile
+  (require 'cl))
+
+(require 'cl-lib)
+(require 'compile)
+(require 'dash)
+(require 'em-glob)
+(require 'f)
+(require 'filenotify)
+(require 'files)
+(require 'imenu)
+(require 'inline)
+(require 'json)
+(require 'network-stream)
+(require 'pcase)
+(require 'seq)
+(require 'spinner)
+(require 'subr-x)
+(require 'url-parse)
+(require 'url-util)
+(require 'widget)
+(require 'xref)
+
 (defconst lsp--message-type-face
   `((1 . ,compilation-error-face)
     (2 . ,compilation-warning-face)
     (3 . ,compilation-message-face)
     (4 . ,compilation-info-face)))
 
+(defconst lsp--errors
+  '((-32700 "Parse Error")
+    (-32600 "Invalid Request")
+    (-32601 "Method not Found")
+    (-32602 "Invalid Parameters")
+    (-32603 "Internal Error")
+    (-32099 "Server Start Error")
+    (-32000 "Server End Error")
+    (-32002 "Server Not Initialized")
+    (-32001 "Unknown Error Code")
+    (-32800 "Request Cancelled"))
+  "Alist of error codes to user friendly strings.")
+
+(defconst lsp--silent-errors '(-32800)
+  "Error codes that are okay to not notify the user about.")
+
+(defconst lsp--completion-item-kind
+  [nil
+   "Text"
+   "Method"
+   "Function"
+   "Constructor"
+   "Field"
+   "Variable"
+   "Class"
+   "Interface"
+   "Module"
+   "Property"
+   "Unit"
+   "Value"
+   "Enum"
+   "Keyword"
+   "Snippet"
+   "Color"
+   "File"
+   "Reference"
+   "Folder"
+   "EnumMember"
+   "Constant"
+   "Struct"
+   "Event"
+   "Operator"
+   "TypeParameter"])
+
 (defcustom lsp-print-io nil
   "If non-nil, print all messages to and from the language server to *Messages*."
+  :group 'lsp-mode
+  :type 'boolean)
+
+(defcustom lsp-inhibit-message t
+  "If non-nil, inhibit the message echo via `inhibit-message'."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-report-if-no-buffer t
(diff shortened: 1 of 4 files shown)


lsp.el
```diff
@@ -0,0 +1,511 @@
...
+
+(defun lsp--auto-configure ()
+  "Autoconfigure `lsp-ui', `company-lsp' if they are installed."
+  (when (featurep 'lsp-ui)
+    (flycheck-mode)
+    (lsp-ui-mode))
+
+  (when (featurep 'company-lsp)
+    (company-mode-on)
+    (setq company-backends '(comapny-lsp))))
```
MaskRay: Typo.

Though personally I prefer a single entry `company-lsp`, some may prefer others.

From my doom-emacs:

```elisp
(company-lsp
 (:separate company-capf company-yasnippet)
 company-capf)
```

yyoncho: I am fine with both approaches. This will act like default configuration and if you want something custom you can always turn off lsp-auto-configure and do the configuration by yourself.

MaskRay: SG. I don't change the doom-emacs default just because I haven't found other backends get in the way. If they do I'll ruthlessly delete them :)

I think `company-backends` is a defcustom and not buffer-local by default.

FirstLoveLife:
`lsp-ui-peek-jump-backward/forward` will mess up when multiples projects open in different workspaces(persp-mode + ccls). For example, 
1. projectA:`lsp-ui-peek-jump-forward`
2. switch to projectB
3. projectB: `lsp-ui-peek-jump-forward`
4. return to projectA
5. `lsp-ui-peek-jump-backward` => goto projectB's buffers, which is not what I want. Can `lsp-session` fix this problem?

Thanks in advance.

vibhavp:
I still haven't examined all of this, unfortunately. Thanks for submitting this, I'll be sure to have a complete review ready by next week.

yyoncho:
@vibhavp thanks, it still not 100% ready, some parts are missing, I have posted it only for reviewing the general idea. I do a PR for `lsp-ui` and `company-lsp` which do work with this CL. I will ping you when it it ready to be tried. As a side note, should be reviewed from the bottom up.

MaskRay:
Since merging `lsp-*.el` is a separate step, can you create a separate PR/commit for it? I'm totally fine with it. That will make the diff smaller and easier to review.

yyoncho:
I will try to do that. Not sure whether this is going to result in better diff.

tsoernes:
How do I make use of this?
I want to use both `pyls` and `pyright` LSP servers for Python. They do different functions.
I've tried to do 
```
       (python            ; beautiful is better than ugly
        +lsp
        +pyright
        )
```
In my `init.el` but pyright is not enabled.

PR #470, opened 2018-11-10
Optimize project root detection. 

seagle0128:
Logic: detecting via projectile or project; If nil, use the current directory.

Fix https://github.com/emacs-lsp/lsp-python/issues/28.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -852,8 +852,7 @@ directory."
 (defun lsp--suggest-project-root ()
   "Get project root."
   (or
-   (when (featurep 'projectile)
-     (projectile-ensure-project (projectile-project-root)))
+   (when (featurep 'projectile) (projectile-project-root))
    (when (featurep 'project)
      (when-let ((project (project-current)))
        (car (project-roots project))))
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -558,11 +558,11 @@ interface TextDocumentItem {
     text: string;
 }"
   (inline-quote
-   (let ((language-id-fn (lsp--client-language-id (lsp--workspace-client lsp--cur-workspace))))
-     (list :uri (lsp--buffer-uri)
-           :languageId (funcall language-id-fn (current-buffer))
-           :version (lsp--cur-file-version)
-           :text (buffer-substring-no-properties (point-min) (point-max))))))
+    (let ((language-id-fn (lsp--client-language-id (lsp--workspace-client lsp--cur-workspace))))
+      (list :uri (lsp--buffer-uri)
+	      :languageId (funcall language-id-fn (current-buffer))
+	      :version (lsp--cur-file-version)
+	      :text (buffer-substring-no-properties (point-min) (point-max))))))
 
 ;; Clean up the entire state of lsp mode when Emacs is killed, to get rid of any
 ;; pending language servers.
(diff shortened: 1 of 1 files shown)


MaskRay:
I don't like the change. If you do want default-directory as a fallback, please set projectile-require-project-root to nil.

seagle0128:
I think the fallback is necessary. How to handle `project` then?

MaskRay:
The user should then install projectile.

MaskRay:
I don't know much about Org but if org src code block is not backed by file, `projectile-project-root` returning `nil` is a reasonable, as I think otherwise many components of lsp-mode will fail.

I have noticed a related change in eglot https://github.com/joaotavora/eglot/pull/150 Do you have any idea what it does?

> If so, should add to REAME or FAQ, I think. I am fine to close this PR with updating docs.

You can probably repurpose this PR to update the document or let @vibhavp make wiki publicly editable.

seagle0128:
In the screenshot above, not org src code, just a single python script file in the junk directory. `projectile-project-root` doesn't work well and I don't know why, even after checking the codes.

> I have noticed a related change in eglot joaotavora/eglot#150 Do you have any idea what it does?

I think for non-file-visiting buffer, `buffer-file-name` may return nil, then use `(expand-file-name (buffer-name)` instead to construct uri for eglot.  The details are described in https://github.com/joaotavora/eglot/issues/116#issuecomment-436586878.

seagle0128:
@MaskRay I checked `projectile` history today, and found this clue https://github.com/bbatsov/projectile/issues/1304#issuecomment-425894184 and https://github.com/bbatsov/projectile/issues/1304#issuecomment-425894344. 

> Basically to get the old behaviour you need to use (projectile-ensure-project (projectile-project-root)).

So the current codes are not desired as you described. I update the codes a little.

lsp-methods.el
```diff
@@ -852,10 +852,13 @@ directory."
...
-   (when (featurep 'projectile) (projectile-project-root))
+   (when (featurep 'projectile)
+     (projectile-ensure-project (projectile-project-root)))
    (when (featurep 'project)
-     (when-let ((project (project-current)))
-       (car (project-roots project))))))
+     (let ((project (project-current)))
+       (if project
+           (car (project-roots project))
+         default-directory)))))
```
yyoncho: That means users will get nil when they do not have `project` installed? Is this the intended behavior?

seagle0128: Yes. I don't like it, but @MaskRay wants this behavior. The first edition is my purpose, and I use it in my personal config.

seagle0128: How about this?

```elisp
(defun lsp--suggest-project-root ()
  "Get project root."
  (or
   (when (featurep 'projectile)
     (projectile-ensure-project (projectile-project-root)))
   (when (featurep 'project)
     (when-let ((project (project-current)))
       (car (project-roots project))))
   default-directory))
```

yyoncho: This one makes more sense to me. Not sure what are the advantages of the previous version.

seagle0128: I am not very sure the background, but I think this version is safer and solid for all scenarios.
I updated and re-submitted the codes.

lsp-methods.el
```diff
@@ -855,7 +855,8 @@ directory."
    (when (featurep 'projectile) (projectile-project-root))
```
MaskRay: If I open `/usr/include/stdio.h` directly and expect `projectile-project-root` to return nil (not to start a language server), what can I do to prevent it from returning `default-directory`

lsp-methods.el
MaskRay: I hope this function can return `nil` when the user does not want to start a language server.

MaskRay: This is different from an outright blacklist:

*  jumping from a project file to `/usr/include/stdio.h` adds the new buffer to the session
* opening `/usr/include/stdio.h` directly does not start a language server

MaskRay:
There are several commits changing the same piece of code back and forth. It may be easier for review by squashing them.

seagle0128:
IMO you should disable lsp-mode temporarily.

It's a requirement issue. I think it's fine to start a lsp server even just open a non-project file. Let's get an ultimate solution.

seagle0128:
> There are several commits changing the same piece of code back and forth. It may be easier for review by squashing them.

The latest commit is fine to review. The behaviors  between `projectile` and `project` in the old codes are different. At least should make them same.

lsp-methods.el
yyoncho: @MaskRay now I understood what are your concerns with defaulting root and they look valid to me. I think at this point it is trade-off which "feature" we want to sacrifice. I will let @vibhavp decide whether to merge it. As a side note, the library handling is covered in #469 as allowing registration of library folder paths which are not going to be used for starting a lsp server.

seagle0128: Can't agree any more. It's up to the requirement and needs a trade-off.
Anyway I use the workaround in my config for now.

vibhavp:
Looks good, thanks.

MaskRay:
@vihavp Why did it get merged? These commits should at least be squashed (there is a dropdown menu where you can choose squash)

MaskRay:
And the anthor has never addressed my comments. Some files (/usr/include/stdio.h) are in an indeterminate state and don't belong to any projects (default-directory /usr/include/ will be bad). The reasonable approach is to return nil and let these buffers be stolen by other project buffers (through a jump-to relation)

PR #475, opened 2018-11-18
Cleanup lsp-mode.el public interface

yyoncho:
The packages like `lsp-ui` and `company-lsp` do use some internal `lsp-mode`
functions. This CL defines them so then can be implmeneted by `lsp.el` (see #469)

This will allow `lsp-ui` and `company-lsp` to work aganst both `lsp.el` and
`lsp-mode.el`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -404,5 +404,27 @@ If WORKSPACE is not specified defaults to lsp--cur-workspace."
       (view-mode 1))
     (switch-to-buffer buffer-name)))
 
+(defun lsp-get-renderer (language)
+  "Return a function to fontify a string in LANGUAGE."
+  (thread-last lsp--cur-workspace
+    lsp--workspace-client
+    lsp--client-string-renderers
+    (assoc-string language)
+    cdr))
+
+(defun lsp-workspace-root (&optional _path)
+  "Find the workspace root for the current file."
+  (when lsp--cur-workspace
+    (lsp--workspace-root lsp--cur-workspace)))
+
+(defun lsp-buffer-language ()
+  "Get language for the current buffer."
+  (when-let (language-id-fn (lsp--client-language-id (lsp--workspace-client lsp--cur-workspace)))
+    (funcall language-id-fn (current-buffer))))
+
+(defun lsp-diagnostics ()
+  "Return diagnostics."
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -404,5 +404,27 @@ If WORKSPACE is not specified defaults to lsp--cur-workspace."
...
+  "Find the workspace root for the current file."
+  (when lsp--cur-workspace
+    (lsp--workspace-root lsp--cur-workspace)))
+
+(defun lsp-buffer-language ()
+  "Get language for the current buffer."
+  (when-let (language-id-fn (lsp--client-language-id (lsp--workspace-client lsp--cur-workspace)))
+    (funcall language-id-fn (current-buffer))))
+
+(defun lsp-diagnostics ()
```
MaskRay: Is this function really necessary? Is it idiomatic to expose the variable `lsp-diagnostics` (after renaming from the double-dash)

yyoncho: As per https://github.com/emacs-lsp/lsp-mode/issues/390 the diagnostics will be moved as part of the workspace and there could be multiple servers providing diagnostics for the current file so having a variable wont work.

lsp-mode.el
```diff
@@ -404,5 +404,27 @@ If WORKSPACE is not specified defaults to lsp--cur-workspace."
...
+  (thread-last lsp--cur-workspace
+    lsp--workspace-client
+    lsp--client-string-renderers
+    (assoc-string language)
+    cdr))
+
+(defun lsp-workspace-root (&optional _path)
+  "Find the workspace root for the current file."
+  (when lsp--cur-workspace
+    (lsp--workspace-root lsp--cur-workspace)))
```
MaskRay: The pattern ` (lsp--workspace-root lsp--cur-workspace)` has been used 4 times in lsp-mode and twice in lsp-ui. Making a dedicated function sounds good but I'm not sure if the `when` check is necessary

yyoncho: I wanted to preserve the original behaviour - there is one place in which the presence of `lsp--cur-workspace` is performed before getting the root - refer to https://github.com/emacs-lsp/lsp-ui/pull/192/commits/43ab6fceeea0eb241d3a80e8da6d248ab9a87097

PR #487, opened 2018-11-30
Refactor lsp-hover

MaskRay:
The original code cancelled pending hover request, but this behavior is not used in any other language client I know about. textDocumenthover is fast so this is not very necessary.

To reduce hover requests, we cache "contents" and "range", and reuse "contents" if (point) is still within "range".

lsp--hover-callback is better using eldoc-message but after a movement operation, the echo area is cleared and eldoc-message does not redisplay the message. So use message for now.

### DIFF
--- a/lsp-methods.el
+++ b/lsp-methods.el
@@ -16,6 +16,7 @@
 ;;; Code:
 
 (require 'cl-lib)
+(require 'dash)
 (require 'json)
 (require 'xref)
 (require 'subr-x)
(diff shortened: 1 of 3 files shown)


lsp-methods.el
```diff
@@ -1894,14 +1859,35 @@ It will be used when no language has been specified in document/onHover result."
...
   "Show relevant documentation for the thing under point."
   (interactive)
-  (lsp--text-document-hover-string))
+  (if (and lsp--hover-saved-bounds
+           (lsp--point-in-bounds-p lsp--hover-saved-bounds))
+      (lsp--hover-callback t)
+    (lsp--send-request-async
+     (lsp--make-request "textDocument/hover"
+                        (lsp--text-document-position-params))
+     (lambda (hover)
```
yyoncho: you may write it like this:
``` emacs-lisp
(-lambda ((&hash "contents" "range")) ...)
```

MaskRay: Changed to (-lambda ((&hash "contents" "range")) ...)

I have to update both `lsp.el` and `lsp-methods.el` because I can update `lsp-ui-doc--hover` to advice `lsp--hover-callback` as a follow-up without breaking both interfaces.

PR #492, opened 2018-12-02
Lsp notification

MaskRay:


lsp.el
```diff
@@ -965,6 +965,13 @@ If WORKSPACE is not provided current workspace will be used."
 
 (defalias 'lsp-send-notification 'lsp--send-notification)
 
+(cl-defun lsp-notify (method params &key single)
```
yyoncho: Instead of single at this point the pattern is 

(with-current-workspace workspace
  (lsp--send-notification)
...
)

PR #509, opened 2018-12-06
Fix wrong javascript-typescript-langserver in lsp-clients.

seagle0128:
The new langserver has no `--stdio` while  providing `javascript-typescript-stdio`.
And the name is `javascript-typescript-langserver` while not `typescript-language-server`.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -89,7 +89,7 @@ This directory shoud contain a file matching groovy-language-server-*.jar"
                   :server-id 'html-ls))
 
 ;;; Typescript
-(defcustom lsp-clients-typescript-server "typescript-language-server"
+(defcustom lsp-clients-typescript-server "javascript-typescript-stdio"
   "The typescript-language-server executable to use.
 Leave as just the executable name to use the default behavior of
 finding the executable with `exec-path'."
@@ -106,11 +106,10 @@ finding the executable with `exec-path'."
 (defun lsp-typescript--ls-command ()
   "Generate the language server startup command."
   `(,lsp-clients-typescript-server
-    "--stdio"
     ,@lsp-clients-typescript-server-args))
 
 (lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection '("typescript-language-server" "--stdio"))
+ (make-lsp-client :new-connection (lsp-stdio-connection (lsp-typescript--ls-command))
                   :major-modes '(typescript-mode js-mode js2-mode rjsx-mode)
                   :server-id 'ts-ls))
 


yyoncho:
Thanks!

yyoncho:
Actually, I was testing with https://github.com/theia-ide/typescript-language-server . From what I can see, it works better than javascript-typescript-langserver . Better completion, code actions, and so on.

yyoncho:
@shackra do you have some experience with testing any of javascript-typescript-langserver and https://github.com/theia-ide/typescript-language-server ?

seagle0128:
Hi, I just realized you had wanted to use `typescript-language-server` instead of `javascript-typescript-langserver`. I submitted this PR according to the https://github.com/emacs-lsp/lsp-mode/blob/master/README-NEXT.md#supported-languages. There is only   `javascript-typescript-langserver`. Well, which ever lang server, the implementation was not perfect. `lsp-clients-typescript-server` and `lsp-clients-typescript-server-args` weren't used in old codes.

I used `tide-mode` before, which leverage `tsserver` from Microsoft Typescript team. `typescript-language-server` is a wrapper of `tsserver`. `tsserver` is good and powerful, but via `RPC` while not `JSON` and `REST`, so it also has some disadvantages. While using `typescript-language-server`, 2 processes will be served. 

I found some discussions for Eclipse plugin: https://github.com/eclipse/wildwebdeveloper/issues/22. It happend one year ago. They decided to stay with `javascript-typescript-langserver` until now. Some comments are below.

> At the moment, there is no value for BlueSky to adopt another LS for JS/TS. So let's close this issue.
BlueSky can consider switching to another LS when there is a clear added-value for most users.

I tried both `typescript-language-server` and `javascript-typescript-langserver` today. Honestly I couldn't find obvious differences. Since I am not the Javascript expert, I leave the final call to you guys.

PS: @yyoncho, one question, after customizing the server and args for langserver, I have to restart Emacs to enable it. `lsp-restart-workspace`, reverting/reopening the files doesn't work. Is it as designed?

yyoncho:
> I tried both typescript-language-server and javascript-typescript-langserver today. Honestly I couldn't > find obvious differences. Since I am not the Javascript expert, I leave the final call to you guys.

I do not use JS/TS either.

@cmatzenbach @shackra what do you think? Based on my short trial typescript-language-server was better since it was navigating properly to definitions in `.node_modules` , it was displaying code actions.

> PS: @yyoncho, one question, after customizing the server and args for langserver, I have to restart Emacs to enable it. lsp-restart-workspace, reverting/reopening the files doesn't work. Is it as designed?

It was typo - 1954390

seagle0128:
Thank you for https://github.com/emacs-lsp/lsp-mode/commit/1954390ed4bfa3f1d9534e896af6256af725ef50.

shackra:
> @shackra do you have some experience with testing any of javascript-typescript-langserver and https://github.com/theia-ide/typescript-language-server ?

I, uh, used to work with the former (for two months straight) but after the update I just use the latter and indeed it provides better completion

yyoncho:
```
(setq lsp-clients-typescript-server "typescript-language-server"
      lsp-clients-typescript-server-args '("--stdio"))
```
+ 
npm install -g typescript-language-server

shackra:
> @shackra or @seagle0128 Interesting - I want to try out that other server. How do you switch out the default servers with the new configuration? Would one of you mind posting your elisp snippet? I, for whatever reason, can't get it pointed to the new one

@yyoncho reply should be enough @seagle0128 , but if you are interested my configuration for Javascript is here https://github.com/shackra/emacs#paquetes-para-editar-javascript

seagle0128:
@cmatzenbach @yyoncho provided the configs. My config is [Centaur Emacs](https://github.com/seagle0128/.emacs.d).

seagle0128:
Hi folks,

Did you have the final decision? I tried both recently, and seems `typescript-language-server` is more intelligent. It's used in VSCode, so I am fine to use it as the default server for `javascript`.

yyoncho:
@seagle0128  We may have 2 LS configurations and set the one with typescript-language-server with higher priority so it will be preferred if both are present.

seagle0128:
Cool! Sounds great. Thank you!

cmatzenbach:
Sorry for the late reply.  I've extensively used both at work - for now, typescript-language-server seems to be a bit less buggy and more intelligent.  They both do some things the other doesn't, and each have their share of bugs, but I do slightly prefer that one.  As of now, @seagle0128 is right and typescript-language-server is a bit more polished and should be the default.

PR #516, opened 2018-12-08
Handle buffers with no buffer-file-name

yyoncho:
- This was reported in gitter by @dsyzling

### DIFF
--- a/lsp.el
+++ b/lsp.el
@@ -3386,8 +3386,9 @@ current language. When IGNORE-MULTI-FOLDER is nil current file
 will be openned in multi folder language server if there is
 such."
   (interactive)
-  (when (setq-local lsp--buffer-workspaces (or (lsp--try-open-in-library-workspace)
-                                               (lsp--try-project-root-workspaces ignore-multi-folder)))
+  (when (and (buffer-file-name)
+             (setq-local lsp--buffer-workspaces (or (lsp--try-open-in-library-workspace)
+                                                    (lsp--try-project-root-workspaces ignore-multi-folder))))
     (lsp-mode 1)
     (when lsp-auto-configure (lsp--auto-configure))))
 


lsp.el
```diff
@@ -3290,7 +3292,7 @@ current language. When IGNORE-MULTI-FOLDER is nil current file
 will be openned in multi folder language server if there is
 such."
   (-if-let* ((session (lsp-session))
-             (project-root (lsp--calculate-root session (buffer-file-name)))
+             (project-root (lsp--calculate-root session (or (buffer-file-name) default-directory)))
```
MaskRay: Why is `default-directory` used here?

yyoncho: When you do find-file and create a new file it won't have buffer-file-name but effectively will be part of the current project.

lsp.el
```diff
@@ -3308,17 +3309,18 @@ Returns nil if the project should not be added to the current SESSION."
...
+    (when-let (workspace (->> (lsp-session)
+                              (lsp--session-workspaces)
+                              (--first
+                               (when-let (library-folders-fn
+                                          (-> it lsp--workspace-client lsp--client-library-folders-fn))
+                                 (-first (lambda (library-folder)
+                                           (f-ancestor-of? library-folder (buffer-file-name)))
+                                         (funcall library-folders-fn it))))))
+      (lsp--open-in-workspace workspace)
+      (view-mode t)
```
MaskRay: `(view-mode 1)`

lsp.el
```diff
@@ -637,7 +637,8 @@ This is equivalent to `display-warning', using `lsp-mode' as the type and
   "Return URI of the current buffer."
   (or lsp-buffer-uri
       (lsp--path-to-uri
-       (or buffer-file-name (ignore-errors (buffer-file-name (buffer-base-buffer)))))))
+       (or buffer-file-name (ignore-errors (buffer-file-name (buffer-base-buffer)))
+           (f-join default-directory (buffer-name))))))
```
MaskRay: `(expand-file-name (buffer-name))`

PR #519, opened 2018-12-08
Default lsp-prefer-flymake to t and fix enabling of flymake/flycheck

MaskRay:
See companion change https://github.com/emacs-lsp/lsp-ui/pull/202


yyoncho:
Looks good. Can you update the readme as well? It says that flyckeck is the default.

MaskRay:
Updated.

I changed `As you type reporting of parsing and compilation errors` to simplify `diagnostics` as `textDocument/publishDiagnostics` is not necessarily triggered by a `textDocument/didChange` on the server side.

PR #520, opened 2018-12-08
Delete legacy lsp-*.el other than lsp-mode.el

MaskRay:
Since legacy interface has gone, lsp-*-enable no longer works and there
is no need to keep these lsp-*.el


yyoncho:
Sure, but I suspect that we need lsp-mode.el for melpa . At least it contains also out package requirements.

MaskRay:
I've kept both `lsp.el` and `lsp-mode.el`. My usage of wildcard does not include `lsp-mode.el` if it is not clear..

MaskRay:
lsp-ui has been updated

PR #525, opened 2018-12-09
Polish flymake

MaskRay:
* When range.start = range.end, e.g. error: expected ';' at end of declaration
      The overlay is empty and thus invisible. Use flymake-diag-region to
      make the overlay visible.
* Create error types lsp-{error,warning,note} (:success did not exist)
* Add lsp--range-to-region to simplify code

### DIFF
--- a/lsp.el
+++ b/lsp.el
@@ -599,6 +599,10 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
               (point))
           (error (point)))))))
 
+(defun lsp--range-to-region (range)
+  (cons (lsp--position-to-point (gethash "start" range))
+        (lsp--position-to-point (gethash "end" range))))
+
 (defmacro lsp-define-stdio-client (&rest _rest)
   "No op - only for backward compatibility.")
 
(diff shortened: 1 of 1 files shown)


lsp.el
```diff
@@ -807,34 +811,38 @@ WORKSPACE is the workspace that contains the diagnostics."
...
+      (lsp--flymake-backend lsp--flymake-report-fn)
+      (remove-hook 'lsp-after-diagnostics-hook 'lsp--flymake-after-diagnostics t)))
+
+  (defun lsp--flymake-backend (report-fn &rest _args)
+    "Flymake backend."
+    (funcall report-fn
+             (-some->> (lsp-diagnostics)
+                       (gethash buffer-file-name)
+                       (--map (-let [(&hash "message" "severity" "range") (lsp-diagnostic-original it)]
+                                (-let [(start . end) (lsp--range-to-region range)]
```
yyoncho: is lsp--range-to-region really needed? Can we destruct and then assign to start and end?

lsp.el
```diff
@@ -807,34 +811,38 @@ WORKSPACE is the workspace that contains the diagnostics."
...
+                                  (when (= start end)
+                                    (-let* (((&hash "line" "character") (gethash "start" range))
+                                            (region (flymake-diag-region (current-buffer) (1+ line) character)))
+                                      (setq start (car region) end (cdr region))))
```
yyoncho: could be destructed

lsp.el
```diff
@@ -807,34 +811,38 @@ WORKSPACE is the workspace that contains the diagnostics."
...
+    (when lsp--flymake-report-fn
+      (lsp--flymake-backend lsp--flymake-report-fn)
+      (remove-hook 'lsp-after-diagnostics-hook 'lsp--flymake-after-diagnostics t)))
+
+  (defun lsp--flymake-backend (report-fn &rest _args)
+    "Flymake backend."
+    (funcall report-fn
+             (-some->> (lsp-diagnostics)
+                       (gethash buffer-file-name)
+                       (--map (-let [(&hash "message" "severity" "range") (lsp-diagnostic-original it)]
```
yyoncho: join both let?

lsp.el
MaskRay: It can be used to simply other similar patterns.

lsp.el
MaskRay: ?

lsp.el
MaskRay: Joined

PR #529, opened 2018-12-10
Fix typos in user question

drrlvn:


yyoncho:
Thanks, I am not a native speaker so wording review is definitely needed.

PR #536, opened 2018-12-13
Initial support for Dart in lsp-client

nykma:
[natebosch/dart_language_server](https://github.com/natebosch/dart_language_server) haven't implement debugger of Dart, so here goes no debugger support.

### DIFF
--- a/README.org
+++ b/README.org
@@ -70,6 +70,7 @@
   | C++                   | [[https://clang.llvm.org/extra/clangd.html][clangd]]                      | Yes          | [[https://clang.llvm.org/extra/clangd.html][clangd]]                                         |          |
   | C++                   | [[https://github.com/cquery-project/cquery][cquery]]                      | [[https://github.com/cquery-project/emacs-cquery][emacs-cquery]] | [[https://github.com/cquery-project/cquery][cquery]]                                         |          |
   | CSS                   | [[https://github.com/vscode-langservers/vscode-css-languageserver-bin][css]]                         | Yes          | npm install -g vscode-css-languageserver-bin   |          |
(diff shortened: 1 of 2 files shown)


yyoncho:
Thanks. AFAIK there is a debugger implementation at https://github.com/Dart-Code/Dart-Code/ . Since there was a request to integrate it in dap-mode, do you think this could play well with your language server implementation? Or do you have plans for implementing support for debug adapter protocol?

yyoncho:
@DanTup thanks for the update. Can you provide more details on the LSP support that you are working on so we can include it as part of lsp-mode configuration? 

Also, if you decouple Dart debugger from vscode give us a shout so we can add support for it.

DanTup:
@yyoncho It's not ready for general consumption (so you shouldn't ship anything on it) but the current status is tracked at https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md. It'd be good to get some feedback from editors though to ensure it does everything required to support them. If you're interested I can send you some notes on how to get it set up to try out.

> Also, if you decouple Dart debugger from vscode give us a shout so we can add support for it.

If it happens, it's likely a little way off, but I will do :-)

yyoncho:
@DanTup lsp-mode is designed in a way that it uses a particular configuration only if it is available and if there is no configuration with higher priority so it is fine to add non-prod ready configurations. I am personally not a dart user but if we have it listed on our readme page you may receive feedback from the interested parties.

DanTup:
@yyoncho It's definitely not ready to be advertised to end users - it's incomplete and hasn't had much testing/validation. I was really looking for an individual that might want to set it up for themselves and run it locally to test and provide feedback. No worries if you're not using Dart, I'll see if I can find someone that is (or if you know of anyone, feel free to send them in my direction). Thanks!

Issue #405, opened 2018-08-15
Can't configure lsp servers with file/directory local variables


duog:
Hi,

I've been trying to configure the haskell lsp server with directory local variables and have run into problems.

It seems that directory local variables are [after major-mode hooks are run](https://stackoverflow.com/questions/19280851/how-to-keep-dir-local-variables-when-switching-major-modes). This means that if i use the natural
```
(add-hook 'haskell-mode 'lsp-haskell-enable)
```
then the server will be started before the path to its executable is configured by my .dir-locals.el.

There are [workarounds](https://stackoverflow.com/questions/5147060/how-can-i-access-directory-local-variables-in-my-major-mode-hooks), however I find these unsatisfying since everyone (who wants to use .dir-locals.el like this) has to remember to do it every time they enable an lsp-server.

Flycheck seems to solve this same problem by -- instead of running a check -- [setting a deferred flag on initialisation](https://github.com/flycheck/flycheck/blob/f6f091949cdaf89f228af46f1562c1438e2b1890/flycheck.el#L2404) and [checking that flag](https://github.com/flycheck/flycheck/blob/f6f091949cdaf89f228af46f1562c1438e2b1890/flycheck.el#L2354) on `post-command-hook` and `window-configuration-change-hook`. Note that I have not been able to find anything to say that they do this to support local variables.

Would you consider using a similar strategy in the implementation of `lsp--enable-stdio-client`?

yyoncho:
I might be missing something, but wouldn't adding ...

``` emacs-lisp
(run-hooks (intern (concat (symbol-name major-mode) "-local-vars-hook")))
```
... to the *-enable function do the job as per the link you have provided?

yyoncho:
Will be solved by #544. It will allow the following configuration:

```
(add-hook 'hack-local-variables-hook (lambda () (when (derived-mode-p 'prog-mode) (lsp))))
```

yyoncho:
Fixed with in latest lsp-mode, sample configuration:

```
(add-hook 'hack-local-variables-hook
          (lambda () (when (derived-mode-p 'XXX-mode) (lsp))))
```

S-D-:
Hi, this doesn't work for me for some reason...
In my init.el I have:
```
(add-hook 'hack-local-variables-hook
          (lambda () (when (derived-mode-p 'prog-mode) (print ccls-initialization-options) (lsp))))
```
and in .dir-locals.el:
```
((prog-mode
[...]

PR #544, opened 2018-12-14
Improve session folder handling and added events for handling them

### LINKED ISSUE #512: [Enhancement] Automatic client detection
Hi, I am using new interface `lsp`.  Is it any possible to add automatic client detection for `lsp-clients`? Like `eglot`, then only one config to setup.

```elisp
(add-hook 'prog-mode-hook 'lsp)
```

Users don't need to care about `lsp` client should be used in which mode. `lsp` knows which client should be started in current programming mode. That's more intelligent.

yyoncho:
Fixes #512

- added `lsp-auto-require-clients` to remove the need to require lsp-clients.
  When set to t(default) it will autorequire lsp-clients.

- `lsp` command ignores the clients which binary is not present.

- `lsp` no longer throws an error when there is no client installed

- `lsp` will ask which client to power on if there are multiple clients for a
  single mode present(also if their binary is available). This will happen each
  time you open a new file. In order to disable that you will have to remove the
  server that you do not want to run from `lsp-clients`. Later, when we solve
  #405 we will be able to specify per project preferences like: ProjectA uses
  ccls, projectB uses clangd.

- If you want your client to run in parallel with other server you will have to
  specify `:add-on?` = t when registering the client.

- added 3 different message `lsp--info`, `lsp--warn` and `lsp--error`

- cleanup
- fixed support for tcp LANGUAGE server which apparently works only on Linux.

``` example registration
(lsp-register-client
 (make-lsp-client :new-connection (lsp-tcp-connection
                                   (lambda (port)
                                     `("php"
                                       ,(expand-file-name "~/.composer/vendor/felixfbecker/language-server/bin/php-language-server.php")
                                       ,(format "--tcp-server=localhost:%s" port)
                                       "--memory-limit=4095M")))
                  :major-modes '(php-mode)
                  :server-id 'php-ls))
```

### DIFF
--- a/README.org
+++ b/README.org
@@ -47,16 +47,13 @@
 
 ** Spacemacs
    [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] is included in spacemacs. Add ~lsp~ to ~dotspacemacs-configuration-layers~ and configure the language that you want to use to be backed by ~lsp~ backend.
-
 * Configuration
-  Add the following line in your configuration file where ~programming-mode-hook~ stands for the language specific hook like ~python-mode-hook~.  In addition, ~lsp-mode~ will automatically detect and configure [[https://github.com/emacs-lsp/lsp-ui][lsp-ui]] and [[https://github.com/tigersoldier/company-lsp][company-lsp]]. To turn off that behavior you could set ~lsp-auto-configure~ to ~nil~.
+  Add the following line in your configuration file:
   #+BEGIN_SRC emacs-lisp
     (require 'lsp)
-    ;; in case you are using client which is available as part of lsp refer to the
-    ;; table bellow for the clients that are distributed as part of lsp-mode.el
-    (require 'lsp-clients)
-    (add-hook 'programming-mode-hook 'lsp)
+    (add-hook 'XXX-mode-hook 'lsp)
   #+END_SRC
+  where ~XXX~ could be major mode like ~python~, ~java~, ~c++~. Alternatively, if you want to minimize you configuration you may use ~prog-mode-hook~. In case you do that, `lsp` will try to start for each programming mode and echo a message when there is no client registered for the current mode or if the corresponding server is not present. In addition, ~lsp-mode~ will automatically detect and configure [[https://github.com/emacs-lsp/lsp-ui][lsp-ui]] and [[https://github.com/tigersoldier/company-lsp][company-lsp]]. To turn off that behavior you could set ~lsp-auto-configure~ to ~nil~.
 
 * Supported languages
   Some of the servers are directly supported by ~lsp-mode~ by requiring
@@ -141,7 +138,13 @@
                       :major-modes '(python-mode)
                       :server-id 'pyls))
   #+END_SRC
-
+** FAQ
+*** How to configure a server with local variables?
+    Add ~lsp~ server call to ~hack-local-variables-hook~ which runs right after the local variables are loaded.
+    #+BEGIN_SRC emacs-lisp
+      (add-hook 'hack-local-variables-hook
+                (lambda () (when (derived-mode-p 'XXX-mode) (lsp))))
+    #+END_SRC
 ** See also
    - [[https://github.com/yyoncho/dap-mode][dap-mode]] - Debugger integration for ~lsp-mode~.
    - [[https://github.com/joaotavora/eglot][eglot]] - An alternative minimal LSP implementation.
--- a/lsp.el
+++ b/lsp.el
@@ -129,6 +129,11 @@
   :group 'lsp-mode
   :type 'boolean)
 
+(defcustom lsp-auto-require-clients t
+  "Auto require lsp-clients."
+  :group 'lsp-mode
+  :type 'boolean)
+
 (defvar-local lsp--cur-workspace nil)
 
 (defvar lsp--uri-file-prefix (pcase system-type
@@ -268,9 +273,15 @@ the symbol information."
     (1 . "Full Document")
     (2 . "Incremental Changes")))
 
-(defcustom lsp-workspace-folders-change nil
+(defcustom lsp-workspace-folders-changed-hook nil
   "Hooks to run after the folders has changed.
 The hook will receive two parameters list of added and removed folders."
+  :type 'hook
+  :group 'lsp-mode)
+
+(defcustom lsp-on-hover-hook nil
+  "The hooks that run after on hover and signature information has been loaded.
+The hook is called with two params: the signature information and hover data."
   :type 'hook
   :group 'lsp-mode)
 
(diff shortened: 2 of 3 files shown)


MaskRay:
I don't recommend advertising `(add-hook 'prog-mode-hook 'lsp)`. It can be a supported configuration though. People don't use LSP for every prog-mode derivative.

lsp.el
```diff
@@ -2867,40 +2826,82 @@ Return a nested alist keyed by symbol names. e.g.
...
+
+(defun lsp--find-available-port (host starting-port)
+  "Find available port on HOST starting from STARTING-PORT."
+  (let ((success nil)
+        (port starting-port))
+    (while (and (not success))
+      (condition-case _err
+          (progn
+            (delete-process (open-network-stream "*connection-test*" nil host port :type 'plain))
+            (setq port (1+ port)))
```
MaskRay: `(cl-incf port)`

yyoncho: thanks, will be fixed.

yyoncho:
> I don't recommend advertising `(add-hook 'prog-mode-hook 'lsp)`. It can be a supported configuration though. People don't use LSP for every prog-mode derivative.

After this PR when you open a file which does not have corresponding language server it will print the following message(no errors):

selection_068

I will extend the readme section with:
"You could either attach `lsp` to `prog-mode-hook` which will try to start a language server for all programming languages and log a message when there is no configuration for the current `major-mode` or you could try to start `lsp` only for specific modes(+example code). " 

What do you think?

README.org
```diff
@@ -47,15 +47,20 @@
 
 ** Spacemacs
    [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] is included in spacemacs. Add ~lsp~ to ~dotspacemacs-configuration-layers~ and configure the language that you want to use to be backed by ~lsp~ backend.
-
 * Configuration
-  Add the following line in your configuration file where ~programming-mode-hook~ stands for the language specific hook like ~python-mode-hook~.  In addition, ~lsp-mode~ will automatically detect and configure [[https://github.com/emacs-lsp/lsp-ui][lsp-ui]] and [[https://github.com/tigersoldier/company-lsp][company-lsp]]. To turn off that behavior you could set ~lsp-auto-configure~ to ~nil~.
+  Add the following line in your configuration file.
+  #+BEGIN_SRC emacs-lisp
+    (require 'lsp)
+    (add-hook 'prog-mode-hook 'lsp)
```
MaskRay: Even a warning `language server ... is not available for some mode` may be annoying for some people.

README.org
```diff
@@ -141,7 +146,13 @@
                       :major-modes '(python-mode)
                       :server-id 'pyls))
   #+END_SRC
-
+** FAQ
+*** How to configure a server with local variables?
+    Add ~lsp~ server call to ~hack-local-variables-hook~ which runs right after the local variables are loaded.
+    #+BEGIN_SRC emacs-lisp
+      (add-hook 'hack-local-variables-hook
+                (lambda () (when (derived-mode-p 'prog-mode) (lsp))))
```
MaskRay: The `prog-mode` example should be specific as I have mentioned that `prog-mode` might not be a good example...

lsp.el
```diff
@@ -396,18 +407,33 @@ must be used for handling a particular message.")
...
+  (message "%s :: %s" (propertize "LSP" 'face 'error) (apply #'format format args)))
+
+(defun lsp--eldoc-message (msg)
+  "Show MSG in eldoc."
+  (run-at-time 0 nil (lambda () (eldoc-message msg))))
+
+(defun lsp-message (format &rest args)
+  "Wrapper over `message' which preserves the `eldoc-message'.
+FORMAT with ARGS are the original message formats."
+  (-let ((inhibit-message lsp-inhibit-message)
```
MaskRay: Does plain `let` work?

lsp.el
```diff
@@ -488,7 +514,9 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
   ;; identifiers it supports or expects.
   (language-id nil :read-only t)
 
-  (DEPRECATED_send-sync nil :read-only t)
+  ;; ‘add-on?’ when set to t this will indicate that the server can be started as
+  ;; a additional to another server. In case there are several
+  (add-on? nil :read-only t)
```
MaskRay: The parameter name and its association comments are still obscure to me.

lsp.el
```diff
@@ -815,9 +836,9 @@ WORKSPACE is the workspace that contains the diagnostics."
         (run-hooks 'lsp-after-diagnostics-hook)))))
 
 (with-eval-after-load 'flymake
-  (put 'lsp-note 'flymake-category 'flymake-note)
-  (put 'lsp-warning 'flymake-category 'flymake-warning)
-  (put 'lsp-error 'flymake-category 'flymake-error)
+  ;; (put 'lsp-note 'flymake-category 'flymake-note)
```
MaskRay: https://github.com/emacs-mirror/emacs/blob/master/lisp/progmodes/flymake.el#L81

A different symbol `lsp-note` gives more flexibility.

lsp.el
```diff
@@ -1071,6 +1092,10 @@ If NO-WAIT is non-nil, don't synchronously wait for a response."
 (cl-defun lsp-request (method params &key no-wait)
   (lsp--send-request `(:jsonrpc "2.0" :method ,method :params ,params) no-wait))
 
+(defun lsp-request-async (method params callback &optional mode)
```
MaskRay: `mode` may be a good candidate for a keyword parameter.

README.org
yyoncho: This option will be for initial configuration and targeted at beginners which are looking for zero configuration. Once you find the message irritating you can change it to work as you want(e. g. attach to the proper mode hook). Of course we may restructure the README based on the user feedback.

lsp.el
yyoncho: Yes, I will change it.

README.org
yyoncho: see above.

lsp.el
yyoncho: Oh, the comment is not finished. When the server has add-on? = true this means that it will be started along the other servers for that mode. For example, spring-boot-language-server could run in parallel with jdt-ls. I am open for suggestions for better naming.

lsp.el
```diff
@@ -2041,72 +2025,85 @@ RENDER-ALL - nil if only the first element should be rendered."
...
-      (cl-assert (integerp lsp--current-signature-help-request-id)))))
+  "Display signature or hover info based on the current position."
+  (if (and lsp--hover-saved-bounds
+           (lsp--point-in-bounds-p lsp--hover-saved-bounds))
+      lsp--eldoc-saved-hover-message
+    (setq lsp--hover-saved-bounds nil
+          lsp--eldoc-saved-hover-message nil)
+
+    (let (signature-response hover-response)
+      (if (lsp--capability "signatureHelpProvider")
```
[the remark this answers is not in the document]
yyoncho: The problem with the current solution is that you should select one of signature or hover, but in the general case you want to see both which does not work without lsp-ui-doc. At this point you could disable whatever method you want by altering  `lsp-method-requirements` which will be generic way to disable certain feature, I havent yet added this to the docs VSCode works in similar way. We could also add generic method toggle feature off to make it more user friendly.

README.org
MaskRay: Beginners may not know what `prog-mode-hook` is and this suggestion does not provide any convenience.

MaskRay: People choose language servers very conservatively, as the status quo tools are not necessarily worse than LSP counterparts.

lsp.el
MaskRay: It works here.

lsp.el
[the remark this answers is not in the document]
MaskRay: But they can also provide their own callback, e.g. 

```elisp
(defun +my/hover-or-signature-help ()
  (if (evil-insert-state-p)
      (lsp-signature-help)
    (lsp-hover)))
```

lsp.el
yyoncho: Ok, I will recompile latest emacs and I will test on emacs 26.1

lsp.el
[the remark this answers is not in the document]
yyoncho: I will change the code to support this configuration.

yyoncho:
@MaskRay  Pushed version based on offline discussion, please take look whether I have missed something.

PR #546, opened 2018-12-15
Readd support for Sass and Less

robbert-vdh:
Sass (SCSS) and Less support seems to have accidentally been dropped after [lsp-css](https://github.com/emacs-lsp/lsp-css) got ported to `lsp-clients.el`. I've simply added the modes back in to the CSS language server definition.


yyoncho:
Thanks!

PR #548, opened 2018-12-16
Fix lsp--find-workspaces-for

flodiebold:
`-if-let` doesn't seem to be the right thing here: We want the if branch if the
entry in lsp-method-requirements exists, but `-if-let` apparently only goes into
the if branch if both `capability` and `registered-capability` are set:

```
ELISP> (-if-let ((&plist :a :b) '(:a "a" :b "b")) (message ":a %s :b %s" a b) (message "else")) ;; fine
":a a :b b"
ELISP> (-if-let ((&plist :a :b) '(:a "a")) (message ":a %s :b %s" a b) (message "else")) ;; not fine
"else"
```

Also add missing method requirements for `textDocument/documentHighlight`.

### DIFF
--- a/lsp.el
+++ b/lsp.el
@@ -365,6 +365,7 @@ than the second parameter.")
      :registered-capability "workspace/executeCommand")
     ("textDocument/hover" :capability "hoverProvider")
     ("textDocument/documentSymbol" :capability "documentSymbolProvider")
+    ("textDocument/documentHighlight" :capability "documentHighlightProvider")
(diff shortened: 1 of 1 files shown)


yyoncho:
Thanks for fixing this! Just curious, which server did this cause a problem for?

flodiebold:
[rust-analyzer](https://github.com/rust-analyzer/rust-analyzer), it doesn't provide hover, so it got really annoying :smile:

PR #550, opened 2018-12-16
Add lsp-find-{definition,references} and rename lsp-goto-{,type-}implementation

MaskRay:
Also:
* Add keyword argument `:display-action`, which can be used as: `(lsp-find-definition :display-action 'window)`
* set `evil-set-command-property :jump t` for these commands


yyoncho:
Can you add `defalias` for all of the renamed public functions so we can delete them when starting the new release?  For example `lsp-find-custom` is used in the spacemacs layer.

PR #572, opened 2018-12-22
Make lsp-clients defined clients priority -1 and pick the client with highest client

MaskRay:


yyoncho:
@cormacc After this CL (remhash 'clangd) won't be needed.

PR #575, opened 2018-12-22
Delete client-specific use-native-json and add lsp-use-native-json

MaskRay:
The intention is to catch hash table cases that should be migrated to plists.

It is used by no client so we can repurpose its usage.


yyoncho:
Rejecting in favour of https://github.com/emacs-lsp/lsp-mode/pull/691 (the main difference is that it switches from list to vector by default + adjusting the code to work with vectors).

PR #578, opened 2018-12-25
Implemented support for server running over TRAMP

### LINKED ISSUE #197: Run lsp-mode in Tramp buffers
I've been poking around a bit at `lsp-mode` to see what it would take to [integrate it with Tramp](https://www.gnu.org/software/emacs/manual/html_node/tramp/Remote-processes.html). Blocker right now is that the call to `(make-process)` doesn't have any Tramp hooks.

Other packages use `start-file-process` to pick up Tramp support, but that function doesn't support stderr-separation. Might need to get more creative for lsp-mode.

yyoncho:
Fixes #197

- refer to readme for implementation details and how to use
- more idomatic emacs lisp - (replacing if+let with if-let idiomatic clojure)
  destructoring

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -428,6 +428,8 @@ must be used for handling a particular message.")
 (defvar lsp--session nil
   "Contain the `lsp-session' for the current Emacs instance.")
 
+(defvar lsp--tcp-port 10000)
+
 (defun lsp--info (format &rest args)
   "Display lsp info message with FORMAT with ARGS."
   (message "%s :: %s" (propertize "LSP" 'face 'success) (apply #'format format args)))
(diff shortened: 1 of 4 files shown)


yyoncho:
Merging the PR based on the informal review/testing in: 
https://github.com/yyoncho/lsp-mode/commit/c586861de17f5dc3b374458ffeca02fd8e24c1c1#commitcomment-31872770 by @Kypert

lsp-mode.el
```diff
@@ -2967,35 +2949,54 @@ SLEEP-INTERVAL is the sleep interval between each retry."
         (file-error (setq success t))))
     port))
 
-(defun lsp-tcp-connection (command-fn)
+(defun lsp-tcp-connection (command)
```
kenkangxgwe: Sorry, but why was this renamed to `command`

yyoncho: It looks like a leftover change, will be fixed.

yyoncho: Fixed with 03e727718e6da93a2344392545534102de715c5c

PR #579, opened 2018-12-26
Overhaul eldoc

MaskRay:
We request both hover and signatureHelp in the eldoc callback and we
face a choice when both are available.

Add a variable lsp-eldoc-prefer-signature-help to customize the
preference (defaults to t). This variable can be extended in the future
to take triggerCharacters and evil-insert-state into account.

There are some cons to send the signatureHelp request unconditionally because:

1. signatureHelp is a feature tightly connected with completion,
they are expensive in some language servers, and no good caching
mechanism. We may consider integrating triggerCharacters in the future.
2. signatureHelp may be state-aware: it is mostly useful in insert state.
In normal state (read-only browsing), its information can largely be
compensated by requesting hover on its function identifier.  There is a
very little loss by preferring hover.  On the other hand, the loss of
hover is hardly remedied by other methods.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -232,6 +232,21 @@ It contains all of the clients that are currently regitered.")
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-eldoc-enable-hover t
+  "If non-nil, eldoc will display hover info when it is present."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-eldoc-enable-signature-help t
+  "If non-nil, eldoc will display signature help when it is present."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-eldoc-prefer-signature-help t
+  "If non-nil, eldoc will display signature help when both hover and signature help are present."
+  :type 'boolean
+  :group 'lsp-mode)
+
 (defcustom lsp-eldoc-render-all nil
   "Define whether all of the returned by document/onHover will be displayed.
 If `lsp-markup-display-all' is set to nil `eldoc' will show only
@@ -338,18 +353,6 @@ than the second parameter.")
   :type 'boolean
   :group 'lsp-mode)
 
-(defcustom lsp-signature-enabled t
-  "If non-nil, eldoc will display signature info when it is present."
-  :type 'boolean
-  :group 'lsp-mode)
-
-(defcustom lsp-hover-enabled t
-  "If non-nil, eldoc will display hover info when it is present.
-In case both signature and hover info are present and
-`lsp-signature-enabled' is t eldoc will display signature info."
-  :type 'boolean
-  :group 'lsp-mode)
-
 (defvar-local lsp--flymake-report-fn nil)
 (defvar-local lsp--flymake-report-pending nil)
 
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -232,6 +232,21 @@ It contains all of the clients that are currently regitered.")
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-eldoc-enable-hover t
+  "If non-nil, eldoc will display hover info when it is present."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-eldoc-enable-signature-help nil
```
yyoncho: IMO it would be better to have that t and have let server which has expensive signature help should implement a setting for enabling signature capability. This is how it works in VSCode - by default you do not get signature help for JDT server while you get signature help for JS and Python. Disabling the signature globally will lead to poor feature discoverability.

lsp-mode.el
```diff
@@ -232,6 +232,21 @@ It contains all of the clients that are currently regitered.")
...
+  "If non-nil, eldoc will display signature help when it is present."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-eldoc-prefer-signature-help nil
```
yyoncho: Similar comment here: once we have requested the signature why do we still show the hover considering that lsp-ui will display the hover too?

lsp-mode.el
```diff
@@ -2053,68 +2056,48 @@ RENDER-ALL - nil if only the signature should be rendered."
...
+             (run-hook-with-args 'lsp-on-hover-hook hover)))))
+      (when (and lsp-eldoc-enable-signature-help (lsp--capability "signatureHelpProvider"))
+        (lsp-request-async
+         "textDocument/signatureHelp"
+         (lsp--text-document-position-params)
+         (lambda (result)
+           (when (eq request-id lsp-hover-request-id)
+             (setq signature result)
+             (when (or lsp-eldoc-prefer-signature-help (not hover))
+               (if signature
```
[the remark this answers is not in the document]
MaskRay: Done.

lsp-mode.el
```diff
@@ -2053,68 +2056,48 @@ RENDER-ALL - nil if only the signature should be rendered."
...
+         "textDocument/hover"
+         (lsp--text-document-position-params)
+         (lambda (result)
+           (when (eq request-id lsp-hover-request-id)
+             (when (and (setq hover result))
+               (when-let (range (gethash "range" hover))
+                 (setq lsp--hover-saved-bounds (lsp--range-to-region range))))
+             (when (or (not lsp-eldoc-prefer-signature-help) (not signature))
+               (if hover
+                   (-let (((&hash "contents") hover))
```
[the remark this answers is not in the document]
MaskRay: Cannot think of a scenario where this happens. I don't want to handle this case until there is some language server telling me that.

lsp-mode.el
MaskRay: I disagree. In vscode signatureHelp is used in combination with completion. It is not triggered so frequently as lsp-mode currently does.

MaskRay: We can add evil integration in the future

lsp-mode.el
MaskRay: lsp-ui-doc does not display signature information by default

lsp-mode.el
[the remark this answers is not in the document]
yyoncho: JDT LS does that check https://github.com/emacs-lsp/lsp-mode/commit/aa97218669bcbe3311c455f4c6a953853bc39170

lsp-mode.el
yyoncho: If you are on a parameter you won't see the signature but the parameter hover info.

lsp-mode.el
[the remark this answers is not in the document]
MaskRay: done

MaskRay:
Reinstated the `signatureHelp` over `hover` preference but we should rethink it. Simplified the logic a bit.

PR #584, opened 2018-12-28
missing dash?

massemanet:
using `GNU Emacs 25.2.1 (x86_64-pc-linux-gnu, X toolkit, Xaw3d scroll bars) of 2018-08-27, modified by Debian`

I had this problem `Symbol’s value as variable is void: session` which seems to be a missing `-` (possibly a emacs 25 peculiarity re `when-let*`).


yyoncho:
Thanks

PR #585, opened 2018-12-29
Use bingo as go LSP server

iNecas:
As
https://github.com/sourcegraph/go-langserver/commit/3c8059f14a3aa6cfe9d9388cd7fbb5cbb4726675
notes, the go-langserver is not longer maintained and additional fixes are not
added to it, so the new features (such as go mod) are not (and will not) be
supported.

The authors recommend https://github.com/saibing/bingo as alternative and my
tests show it fixes the issues I was hitting with the go-langserver for me (such
as missing flycheck).


yyoncho:
Instead of changing the configuration, can you add another one with higher priority so it won't break any existing configurations(changing the current one to -2, new one -1)? `lsp-mode` first checks whether the server exists and if it exists it choses the one with higher priority.

lsp-clients.el
```diff
@@ -151,17 +151,12 @@ finding the executable with `exec-path'."
...
 ;;; GO language
 
-(defcustom lsp-clients-go-server "go-langserver"
+(defcustom lsp-clients-go-server "bingo"
   "The go-langageserver executable to use."
   :group 'lsp-go
   :risky t
   :type 'file)
 
-(defcustom lsp-clients-go-executable-path (executable-find "go-langserver")
```
iNecas: Doesn't look like this has been used anywhere

iNecas:
@yyoncho you guys really have thought the thing though! Nice suggestion. Updated and works as expected.

yyoncho:
Thanks

seagle0128:
I am so disappointed `go-langserver` is deprecated, but glad to see `bingo` is added, although I encountered errors while installing it. I also noticed this: https://github.com/saibing/bingo/issues/13. It seems a new official package `golsp` is being developing.

iNecas:
Yes: I guess we will switch to the golsp once it's out, we will probably switch to it

seagle0128:
Thank you for the update, @iNecas .

shuxiao9058:
The `golsp` was released. @iNecas 

https://github.com/golang/tools/tree/master/cmd/golsp

PR #586, opened 2018-12-29
Don't steal focus when showing helping at point

iNecas:
When using go-mode, I found it pretty helpful to not loose focus from the
current buffer when showing doc at point. I've inspired by
https://github.com/dominikh/go-mode.el/blob/v1.5.0/go-mode.el#L292
and used `display-buffer` in the corresponding function to achieve this
behavior.

I believe this is more user-friendly than the current behaviour, therefore
suggesting this as the default behavior for this function.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1963,7 +1963,7 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
              (insert (lsp--render-on-hover-content contents t))
              (goto-char (point-min))
              (view-mode t)
-             (current-buffer))))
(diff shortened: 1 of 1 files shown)


yyoncho:
IMO it is more common to change the focus when performing lookup(e. g. `elisp-slime-nav-describe-elisp-thing-at-point`, `cider-doc`) so I would prefer either to pass a parameter or introduce a configuration option to turn on that behaviour. Also, this is the expected behaviour in spacemacs layer and fixing it will take a lot of time...

yyoncho:
Closing stalled PR.

PR #587, opened 2018-12-31
Make `lsp-workspace-folders-add` work non-interactively

danielmartin:
This interactive function asks for the project root to add, but it also needs to work
programmatically.

(This is a first step to fix the integration tests in `dap-mode`.)


yyoncho:
Thanks. I wonder whether moving DAP tests to use mock debugger (https://github.com/Microsoft/vscode-mock-debug) will improve the tests.

Issue #576, opened 2018-12-23
lsp-mode for dart does not guess correct project root


bradyt:
The prompts created by `lsp--find-root-interactively` and `lsp--suggest-project-root` will quite likely try to pass a directory having a `.git` directory, or similar. This may frequently be incorrect, as I believe the `dart_analysis_server` should be passed a directory at or above a directory with `pubspec.yaml` (or `BUILD`). I suspect we can in fact pass any directory _above_ such, in particular, the directory of current file.

This way, the prompts will be more helpful for user, and even, user can better use `lsp-auto-guess-root`.

The following diff helps ensure the right directory is passed to `dart_language_server`. (I'm making a lot of assumptions about how `dart_language_server`, LSP, and `lsp-mode` all work together.)

``` diff
@@ -1266,6 +1266,11 @@ disappearing, unset all the variables related to it."
 (defun lsp--suggest-project-root ()
   "Get project root."
   (or
+   (when (eq major-mode 'dart-mode)
+     (let ((file (buffer-file-name (current-buffer))))
+       (or (locate-dominating-file file "pubspec.yaml")
+           (locate-dominating-file file "BUILD")
+           (file-name-directory file))))
    (when (featurep 'projectile) (condition-case nil
                                     (projectile-project-root)
                                   (error nil)))
```

My assumptions are based on the following lines from https://github.com/natebosch/dart_language_server/blob/master/lib/src/utils/package_dir_detection.dart#L19-L21, and some notes about `rootUri` in the changelog at same project.

``` dart
bool _isPackageRoot(Directory dir) => dir.listSync().any((f) =>
    f.path.endsWith('${Platform.pathSeparator}pubspec.yaml') ||
    f.path.endsWith('${Platform.pathSeparator}BUILD'));
```

bradyt:
Does lsp-mode set project root based on a response from server? If so, can we simply write the following?

``` diff
@@ -1266,6 +1266,8 @@ disappearing, unset all the variables related to it."
 (defun lsp--suggest-project-root ()
   "Get project root."
   (or
+   (when (eq major-mode 'dart-mode)
+     (file-name-directory (buffer-file-name (current-buffer))))
    (when (featurep 'projectile) (condition-case nil
                                     (projectile-project-root)
                                   (error nil)))
```

yyoncho:
> Does lsp-mode set project root based on a response from server?

[...]

PR #590, opened 2019-01-01
Support for the Fortran-language-server

MagB93:
Added support for the [fortls](https://github.com/hansec/fortran-language-server) via 
a configuration inside ```lsp-clients.el```.


yyoncho:
Looks good. Can you remove the guess function, here it is a discussion on how the same functionality can be achieved using project.el/projectile - see also https://github.com/emacs-lsp/lsp-mode/issues/576 . I want to keep the root suggestion logic outside of lsp-mode codebase since it is fairly complex and already implemented/provided by project.el/projectile.

MagB93:
I've pushed a commit that deleted that part. The server starts just as fine without it!
The reason I put that there in the first place was that I saw something similar in [emacs-ccls](https://github.com/MaskRay/emacs-ccls/blob/master/ccls.el#L129).

yyoncho:
Thanks!

Issue #592, opened 2019-01-10
[Feature Request]: Add recipe to start a minimal environment for convenience of bug-reporting and testing


cireu:
English is not my mother-tongue, plz ignore my typo or grammarly issue ;)

I think it's useful to have a recipe (or Makefile target) like [this](https://github.com/abo-abo/swiper/blob/master/Makefile#L18) to help user start a minimal `lsp-mode` environment. It can help user locate the bug or issue more conveniently.

Since `lsp-mode` has more heavier dependencies than `ivy`. But I think we can let user to specific their package directory which includes all dependencies of `lsp-mode`. So we can using following code to add all the packages to `load-path`

```Elisp
(let ((default-directory (file-name-as-directory package-dir)))
  (push default-directory load-path)
  (normal-top-level-add-subdirs-to-load-path))
```

And now `lsp-mode` can require its dependencies and start a minimal environment.

yyoncho:
I have a similar thing in `lsp-java` - there is quick start setup section and which then could be loaded `emacs -q -l quick-start.el` which will fire-up clean environment. I believe that this will be more suitable for `lsp-mode` than having make target since in `lsp-mode` the issue might be related to `lsp-ui`/`company-lsp` and the user might want to include/exclute components and it won't require cloning the repo. What do you think?

cireu:
That's a good idea.

But I think it will be better for version control to put the recipe into a single file and tell user to run it like this
```Bash
emacs -q --eval '(with-current-buffer (url-retrieve-synchronously "https://raw.githubusercontent.com/emacs-lsp/lsp-mode/master/quick-start.el") (eval-buffer))'
```

cireu:
And I also think it will be better to load these file with `(load "lsp-mode.el")` or `(require 'lsp-mode "lsp-mode.el")` rather than using `use-package`. Emacs prefer byte-compiled `.elc` file than `.el` file. But the backtrace from byte-compiled files are wired and not human-readable.

cireu:
And I also think it will be better to load these file with `(load "lsp-mode.el")` or `(require 'lsp-mode "lsp-mode.el")` rather than using `use-package`. Emacs prefer byte-compiled `.elc` file than `.el` file. But the backtrace from byte-compiled files are wired and not human-readable.

yyoncho:
I believe that we may force loading of `.el` files by modifying `load-suffixes`.

[...]

PR #594, opened 2019-01-12
Add lsp-start-plain.el to provide a quick starter for minimal env.

cireu:
See https://github.com/emacs-lsp/lsp-mode/issues/592

How to use this:
1. Download the file and run it with `emacs -q`
2. using `emacs -q --eval '(with-current-buffer (url-retrieve-synchronously "http://github.com/cireu/lsp-mode/raw/master/lsp-start-plain.el") (goto-char (point-max)) (eval-print-last-sexp))'

I've test it using [emacs-china](http://elpa.emacs-china.org/) elpa mirror. But my network is too slow to connect to melpa. Can someone help testing it?


cireu:
I still can't find a way to avoid recursive loading issue. But I find it's possible to let `package.el` don't compile packages.

So I rewrote it. Can someone help test it on Windows and MacOS platform? Thx a lot :smile:

scripts/lsp-start-plain.el
```diff
@@ -0,0 +1,54 @@
...
+;; for more readable backtrace.
+
+;;; Code:
+
+
+(let* ((lexical-binding t)
+       (package-archives '(("melpa" . "http://melpa.org/packages/")
+                           ("gnu" . "http://elpa.gnu.org/packages/")))
+       (no-byte-compile t)
+       (package-user-dir (expand-file-name "tmpelpa" user-emacs-directory))
```
yyoncho: maybe use make-directory/make-temp-name to generate new temp directory each time.

scripts/lsp-start-plain.el
```diff
@@ -0,0 +1,54 @@
...
+;; It will install  `lsp-mode', `lsp-ui' and `company-lsp'
+;; with their depedencies to start a minimal lsp environment.
+
+;; And it forces Emacs to load `.el' files rather than `.elc' files
+;; for more readable backtrace.
+
+;;; Code:
+
+
+(let* ((lexical-binding t)
```
yyoncho: not needed when headers starts with:
```
;;; lsp-start-plain.el --- LSP mode quick starter      -*- lexical-binding: t; -*-
```

cireu:
I am considering to put this util function into a part of `lsp-mode` like magit did.

https://github.com/magit/magit/blob/cc435005b07bd7ba17962ffa4e210b441e8a1a52/lisp/magit-utils.el#L725

What's your opinion? Keep it a single file or merge it into `lsp-mode`?

cireu:
Er.. Can you describe your idea with more details? In my understanding, Cask has done what you said.

IMO, we can get this PR done first, so users can use it and benefit from it. And then is time to think about what to do next. We may need more people's help and idea.

yyoncho:
My idea which is not directly related to that issue is to create one meta package, let's call it `lsp-ide` which will have a dependency to all lsp related - lsp-mode/lsp-ui/company-lsp/helm-lsp + few more that I am planning in the future. What I am saying is that a method like `lsp-ide-emacs-Q` looks like a good fit for that package.

cireu:
Ah.. You mean a package(`lsp-ide`) to intergrate and associate these packages to create a `lsp-mode` based workflow in Emacs (like [aweshell](https://github.com/manateelazycat/aweshell), lazycat packs some package to provide an out-of-box and fancy eshell)? That sounds good! And such `lsp-ide-emacs-Q` functions will be more suitable in `lsp-ide` than `lsp-mode`

cireu:
My opinion is to keep this file single now and let it be a part of `lsp-ide` in near future :)

PR #600, opened 2019-01-19
Change tramp implementation to not rely on TCP connections.

karstenpatzwaldt:
Opening TCP ports is not very secure, and often not possible. Instead,
tunnel client invocations directly through tramp.

For this to work, 3 things needed to be changed:

- open the connection with start-file-process instead of make-process
- append an extra newline at the end of the JSON payload to work
  around line buffering issues
- add an option to disable CRLF line breaks, since some language
  servers seem to not like them

This was tested with tramp to docker with ccls and pyls.

Note that in theory, since this unbreaks the auto-detection on remote clients, this could be unified into the regular stdio-client, which would make tramp "just work" without extra configuration. However, since start-file-process does not support stderr redirection, this may break compatibility with some LSP implementations that spew onto stderr.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -784,7 +784,7 @@ DELETE when `lsp-mode.el' is deleted.")
 (defun lsp--path-to-uri (path)
   "Convert PATH to a uri."
   (concat lsp--uri-file-prefix
-          (url-hexify-string (file-truename (or (file-remote-p path 'localname t) path))
+          (url-hexify-string (or (file-remote-p path 'localname t) path)
                              url-path-allowed-chars)))
 
 (defun lsp--string-match-any (regex-list str)
@@ -1389,8 +1389,10 @@ If WORKSPACE is not provided current workspace will be used."
                         (fboundp 'json-serialize))
                    (json-serialize params :null-object nil
                                    :false-object json-false)
-                 (json-encode params))))
-    (format "Content-Length: %d\r\n\r\n%s" (string-bytes body) body)))
+                 (json-encode params)))
+         (body-with-newline (concat body "\n")))
+    (concat (format "Content-Length: %d\r\n\r\n" (string-bytes body-with-newline))
+                    body-with-newline)))
 
 (defun lsp--send-notification (body)
   "Send BODY as a notification to the language server."
@@ -3240,7 +3242,13 @@ Return a nested alist keyed by symbol names. e.g.
 
 (defun lsp-server-present? (final-command)
   "Check whether FINAL-COMMAND is present."
-  (executable-find (nth 0 final-command)))
+  ;; executable-find only gained support for remote checks after 26.1 release
+  (cond
+   ((not (file-remote-p default-directory))
+    (executable-find (nth 0 final-command)))
+   ((not (version<= emacs-version "26.1"))
+    (executable-find (nth 0 final-command) (file-remote-p default-directory)))
+   (t)))
 
 (defun lsp-stdio-connection (command)
   "Create LSP stdio connection named name.
@@ -3319,32 +3327,23 @@ COMMAND-FN will be called to generate Language Server command."
                 (cons tcp-proc proc)))
    :test? (lambda () (-> command lsp-resolve-final-function lsp-server-present?))))
 
-(defun lsp-tramp-connection (command-fn)
-  "Create LSP TRAMP connection named.
-COMMAND-FN will be called to generate Language Server command."
-  (list
-   :connect (lambda (filter sentinel name)
-              (let* ((host (file-remote-p (buffer-file-name) 'host t))
-                     (port (lsp--find-available-port host (cl-incf lsp--tcp-port)))
-                     (command (funcall command-fn name port))
-                     (proc (progn
-                             (lsp--info "Starting %s" command)
-                             (async-shell-command command
-                                                  (generate-new-buffer-name (format "%s::stdout" name))
-                                                  (generate-new-buffer-name (format "%s::stderr" name)))))
-                     (tcp-proc (lsp--open-network-stream host port (concat name "::tramp::tcp"))))
-
-                (set-process-sentinel proc sentinel)
-                (set-process-query-on-exit-flag proc nil)
-                (set-process-query-on-exit-flag tcp-proc nil)
-                (set-process-filter tcp-proc filter)
-                (cons tcp-proc proc)))
-   :test? (-const t)))
-
-(defun lsp-make-nc-tramp-command (program)
-  "Return a function which will convert PROGRAM from STDIO command to TCP one."
-  (lambda (_name port)
-    (format "nc -l -p %s -c '%s'" port program)))
+(defun lsp-tramp-connection (local-command)
+  "Create LSP stdio connection named name.
+COMMAND is either list of strings, string or function which
+returns the command to execute."
+  (list :connect (lambda (filter sentinel name)
+                   (let* ((final-command (lsp-resolve-final-function local-command))
+                          ;; wrap with stty to disable converting \r to \n
+                          (wrapped-command (append '("stty" "-icrnl" ";") final-command))
+                          (process-name (generate-new-buffer-name name)))
+                     (let ((proc (apply 'start-file-process-shell-command process-name
+                                        (format "*%s*" process-name) wrapped-command)))
+                       (set-process-sentinel proc sentinel)
+                       (set-process-filter proc filter)
+                       (set-process-query-on-exit-flag proc nil)
+                       (set-process-coding-system proc 'binary 'binary)
+                       (cons proc proc))))
+        :test? (lambda () (-> local-command lsp-resolve-final-function lsp-server-present?))))
 
 (defun lsp--auto-configure ()
   "Autoconfigure `lsp-ui', `company-lsp' if they are installed."
@@ -3473,7 +3472,7 @@ SESSION is the active session."
       (run-hooks 'lsp-before-initialize-hook)
       (lsp-request-async "initialize"
                          (list :processId (emacs-pid)
-                               :rootPath (expand-file-name root)
+                               :rootPath (file-local-name (expand-file-name root))
                                :rootUri (lsp--path-to-uri root)
                                :capabilities (lsp--client-capabilities)
                                :initializationOptions initialization-options)
--- a/README.org
+++ b/README.org
@@ -175,23 +175,16 @@
     LSP mode has support for tramp buffers with the following requirements.
     - The language server have to be present on the remote server since it is running on the remote machine.
     - Having multi folder language server (like [[https://github.com/eclipse/eclipse.jdt.ls][Eclipse JDT LS]]) cannot have local and remote servers.
-    - The server must be using TCP connection or it should be converted to TCP via ~netcat~ or similar tool.
-**** How it works?
-     ~lsp-mode~ detects that a particular file is located on remote machine and looks for a clients which can handle the current major mode and it is marked as with ~:remote?~ t. Then ~lsp-mode~ starts the client and it expects to use TCP connection. Since not all of the server have TCP support ~netcat~ is used to convert STDIO to TCP using ~lsp-make-nc-tramp-command~.
-     #+BEGIN_SRC bash
-       nc -l -p {port} -c '{command}'
-     #+END_SRC
-     In case the server already supports TCP connection it could be used directly without ~netcat~ adapter.
+**** How does it work?
+    ~lsp-mode~ detects whether a particular file is located on remote machine and looks for a client which can handle the current major mode and it is marked as ~:remote?~ t. Then ~lsp-mode~ starts the client through tramp.
 **** Sample configuration
-     Here it is example how you can configure python language server to work when using ~TRAMP~.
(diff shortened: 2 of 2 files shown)


yyoncho:
Thanks for the PR - definitely an improvement over the current implementation!


> append an extra newline at the end of the JSON payload to work around line buffering issues

Can you elaborate on why this is needed since it is a bit hacky so it is good to have it documented?

> Note that in theory, since this unbreaks the auto-detection on remote clients, this could be unified into the regular stdio-client, which would make tramp "just work" without extra configuration. 

Yes, at some point I realized it too but decided not to go that far as part of the first version.

> However, since start-file-process does not support stderr redirection, this may break compatibility with some LSP implementations that spew onto stderr.

That is why I decided to go with TCP - we may fix this issue by changing the way parser works, e. g. we may search forward for presense of JSON RPC headers instead of relying on messages to be one after another. I was planning to offer ssh port forwarding to solve the secirity issue but it we can make it work over stdio that would be better.

yyoncho:
Ok, thanks. Can you resolve the conflicts so I can merge it?

karstenpatzwaldt:
Did some more digging, and found out that both the CRLF issue, and the line buffering, are due to the terminal settings that tramp is using. I changed the code to wrap the commands with proper stty settings, so the extra newline is not required anymore.

lsp-mode.el
```diff
@@ -3240,7 +3240,7 @@ Return a nested alist keyed by symbol names. e.g.
 
 (defun lsp-server-present? (final-command)
   "Check whether FINAL-COMMAND is present."
-  (executable-find (nth 0 final-command)))
+  (executable-find (nth 0 final-command) (file-remote-p default-directory)))
```
Kypert: Here I get:

```
(wrong-number-of-arguments (1 . 1) 2)
  executable-find("/home/mj/cquery/build/release/bin/cquery" "/ssh:someserver.se:")
  lsp-server-present\?(("/home/mj/cquery/build/release/bin/cquery"))
```
Maybe missing something obvious?

#### System Info :computer:
- OS: gnu/linux
- Emacs: 26.1
- Spacemacs: 0.300.0
- Spacemacs branch: 20181231 (rev. 2c28c7d21)
- Graphic display: t
- Distribution: spacemacs
- Editing style: vim
- Completion: helm
- Layers:
```elisp
(helm auto-completion emacs-lisp git markdown multiple-cursors org
      (shell :variables shell-default-height 30 shell-default-position 'bottom)
      spell-checking syntax-checking
      (version-control :variables version-control-diff-tool 'diff-hl version-control-global-margin t version-control-diff-side 'left)
      (c-c++ :variables c-c++-backend 'lsp-cquery c-c++-lsp-executable "/home/mj/cquery/build/release/bin/cquery" c-c++-default-mode-for-headers 'c++-mode)
      themes-megapack go
      (plantuml :variables plantuml-jar-path "/usr/share/plantuml/plantuml.jar")
      yaml)
```
- System configuration features: XPM JPEG TIFF GIF PNG RSVG IMAGEMAGICK SOUND GPM DBUS GSETTINGS NOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE M17N_FLT LIBOTF XFT ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 MODULES THREADS XWIDGETS LIBSYSTEMD LCMS2

lsp-mode.el
```diff
@@ -3319,32 +3319,23 @@ COMMAND-FN will be called to generate Language Server command."
...
-  (lambda (_name port)
-    (format "nc -l -p %s -c '%s'" port program)))
+(defun lsp-tramp-connection (local-command)
+  "Create LSP stdio connection named name.
+COMMAND is either list of strings, string or function which
+returns the command to execute."
+  (list :connect (lambda (filter sentinel name)
+                   (let* ((final-command (lsp-resolve-final-function local-command))
+                          ;; wrap with stty to disable converting \r to \n and line buffering
+                          (wrapped-command (append '("stty" "-icrnl" "extproc" ";") final-command))
```
Kypert: My stty (8.22) is probably too old, is there any other way to workaround this?

```
Warning (lsp-mode): Failed to parse the following chunk:
’’’
stty: invalid argument ‘extproc’
Try 'stty --help' for more information.
Content-Length: 152

{"jsonrpc":"2.0","method":"$cquery/progress","params":{"indexRequestCount":0,"doIdMapCount":0,"onIdMappedCount":0,"onIndexedCount":0,"activeThreads":0}}
’’’
with message (error No Content-Length header)
```

karstenpatzwaldt: This is indeed a recent stty feature, which is unfortunate. I changed the fix for this back to adding whitespace to the payload, which works independently of the stty version.

Thanks for testing early and catching this!

lsp-mode.el
karstenpatzwaldt: This is indeed a post-26.1 feature. Added a fallback implementation for current releases.

yyoncho:
Thank you! :heart: 
And congratulations on your first `lsp-mode` PR :tada:

This is a huge improvement over what we currently have!

Kypert:
@karstenpatzwaldt : Many thanks for the PR! Would you mind mapping out how you tested it with `ccls` remotely? Did you start it with any different settings than the stdio-variant?

My setup seems to trigger the init msg and the server responds with the capabilities, but after loading the cache it goes silent. Async msg's are sent to the server, but no reaction recorded (checked by starting `ccls` with `-log-file`). I am not sure how to take it further. Any advice much appreciated :)

yyoncho:
@Kypert can you check the content of also \*lsp-log\* after setting `lsp-print-io` to t? That would be helpfull. I am assuming that you are using emacs 26.1, right?

Kypert:
@yyoncho yes, I am using 26.1, spacemacs and the ccls layer. 
https://gist.github.com/Kypert/df879fff4e1316d81939f52414e5dc91

My current client is registered like this (had to suppress info logging via -v):

```
(lsp-register-client
 (make-lsp-client
  :new-connection (lsp-tramp-connection "/proj/tools/ccls/Release/ccls -v=-1")
  :major-modes '(c-mode c++-mode cuda-mode objc-mode)
  :remote? t
  :server-id 'ccls-tramp
  :multi-root nil 
  :notification-handlers
  (lsp-ht ("$ccls/publishSkippedRanges" #'ccls--publish-skipped-ranges)
          ("$ccls/publishSemanticHighlight" #'ccls--publish-semantic-highlight))
  :initialization-options (lambda () ccls-initialization-options)
  :library-folders-fn nil))
```

ccls is working just fine when using the same setup, working locally on the remote host.

yyoncho:
I doubt that this is causing the issue but "/proj/tools/ccls/Release/ccls -v=-1" should be (list "/proj/tools/ccls/Release/ccls" "-v=-1"). I will test in my setup whether it works with CCLS.

Kypert:
Unfortunately, making a list of the command did not help, still same behaviour. Thanks for investigating! :)

yyoncho:
Just to confirm - the ccls process is alive and there is nothing in the err buffer?

Kypert:
The ccls process is alive on the remote host.

If you are referring to the *ccls-tramp* buffer, then it is empty. I am not able to spot any other errors in any other buffers either.

Kypert:
I am especially puzzled since a smaller project is able to work over the tramp-connection.

Using `tramp-current-method ssh`, but noticed that `tramp-default-method scp`.
Local Ubuntu 16.04.5 LTS: `Linux mjbox 4.15.0-43-generic #46~16.04.1-Ubuntu SMP Fri Dec 7 13:31:08 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`
Remote Red Hat Enterprise Linux Server release 7.5 (Maipo): `Linux mjhost 3.10.0-862.14.4.el7.x86_64 #1 SMP Fri Sep 21 09:07:21 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux`

Kypert:
Truly appreciate your insight :) During today's investigation I found that it actually worked a few (very rare) times out of many. So it feels like a timing issue somewhere, or do we dare to blame packet loss? Please let me know if I should take this discussion elsewhere (new issue or gitter?).

yyoncho:
I guess you may open a new issue and keep @MaskRay  in the loop too to help to diagnose the CCLS side(I think that there was an option to dump the whole communication). My wild guesses are that it might be buffering issue that @karstenpatzwaldt solved by additional "\n", tramp bug or wrong setting.

Kypert:
Thank you for your help and suggestions. I got it to work finally. Here are my findings/struggles:

When I use the remote tramp connection to open a file larger than 4080 chars, the text content in the method `textDocument/didOpen` is truncated. Interestingly, the closing brackets in the json req are presents. lsp-log shows the full content, but when ccls is receiving on stdin the content is truncated.

By running the diagnostics suggested by the documentation of `tramp-chunksize`, it seems like my env requires me to set the chunksize. This could explain the truncated results described above.

```
Bytes sent: 4100    Bytes received: 4095
You should set ‘tramp-chunksize’ to a maximum of 4050
```

However, the issue persisted even after lowering the chunk to 500.

After some more investigation, adding `-icanon` to stty makes it possible to load the larger files, even without setting `tramp-chunksize`. Another way is to set `raw`. In essence we are trying to handle ^D, I think. I am not sure which setting is more appropriate, or if one could slice or omit the file content of didOpen, I will let you consider the best solution here :)

Sources:
* https://unix.stackexchange.com/questions/204815/terminal-does-not-accept-pasted-or-typed-lines-of-more-than-1024-characters/204820
* https://unix.stackexchange.com/questions/131105/how-to-read-over-4k-input-without-new-lines-on-a-terminal

karstenpatzwaldt:
That's very good information.

'raw' also contains '-icrnl', so if 'raw' works for you, then swapping '-icrnl' for 'raw' sounds like the correct thing to do.

Can you try this PR: https://github.com/emacs-lsp/lsp-mode/pull/635

wangfenjin:
> The ccls process is alive on the remote host.

@Kypert Could you please share your configuration in client and server side? I also want to try this, but seems from your client config, it always try to use find the ccls binary in local host, but not remote.

Thanks!

yyoncho:
Did you add the ccls folder to `tramp-remote-path`?

wangfenjin:
> tramp-remote-path

No, I'm using spacemacs, kind of new to all these configs... I will try to search more info

wangfenjin:
No lucky. This is my current config:

```
    (with-eval-after-load 'lsp-mode
        (require 'tramp)
        (add-to-list 'tramp-remote-path "/home/wangfenjin/bin/ccls")
        (add-to-list 'tramp-remote-path 'tramp-own-remote-path)
        (lsp-register-client
        (make-lsp-client
        :new-connection (lsp-tramp-connection "/home/wangfenjin/bin/ccls" "-v=-1")
        :major-modes '(c-mode c++-mode cuda-mode objc-mode)
        :remote? t
        :server-id 'ccls-tramp
        :multi-root nil
        :notification-handlers
        (lsp-ht ("$ccls/publishSkippedRanges" #'ccls--publish-skipped-ranges)
                ("$ccls/publishSemanticHighlight" #'ccls--publish-semantic-highlight))
        :initialization-options (lambda () ccls-initialization-options)
        :library-folders-fn nil))
    )
```

yyoncho:
>  (add-to-list 'tramp-remote-path "/home/wangfenjin/bin/ccls")

withtout ccls 

> :new-connection (lsp-tramp-connection "/home/wangfenjin/bin/ccls" "-v=-1")

only "ccls" without the dir.

yyoncho:
> > :new-connection (lsp-tramp-connection "/home/wangfenjin/bin/ccls" "-v=-1")
> 
> only "ccls" without the dir.

also, the second param is not the args, you should pass a list if you want to pass args.

wangfenjin:
Yes, thank you @yyoncho . I change the config to:
```

  (with-eval-after-load 'lsp-mode
    (require 'tramp)
    (add-to-list 'tramp-remote-path "/home/wangfenjin/bin")
    (add-to-list 'tramp-remote-path 'tramp-own-remote-path)
    (setq max-lisp-eval-depth 100000)
    (setq max-specpdl-size 100000)
    (lsp-register-client
     (make-lsp-client
      :new-connection (lsp-tramp-connection "ccls")
      :major-modes '(c-mode c++-mode cuda-mode objc-mode)
      :remote? t
      :server-id 'sg-devbox
      :multi-root nil
      :notification-handlers
      (lsp-ht ("$ccls/publishSkippedRanges" #'ccls--publish-skipped-ranges)
              ("$ccls/publishSemanticHighlight" #'ccls--publish-semantic-highlight))
      :initialization-options (lambda () ccls-initialization-options)
      :library-folders-fn nil))
    )
```

I think it works now, but may because my repo is really large, so after enlarge these too args my emacs just crashed...
```
    (setq max-lisp-eval-depth 100000)
    (setq max-specpdl-size 100000)
```

wangfenjin:
I think there is no way to fix this... Thanks you anyway

yyoncho:
you may try to stop file watches. Generally, emacs should work no matter what the size of the repo is.

wangfenjin:
I tried to disable file watches, still no lucky. 

It's weird as all works fine if I use emacs in remote server.

PR #603, opened 2019-01-19
Add Flow Language Server support

danielmartin:
This language server is considered an "add-on", that is, it will be
enabled simultaneously with another JavaScript/TypeScript server if
both are present and the buffer meets the conditions of a Flow source
file. The server used is the official one from https://flow.org, which
is supposedly the one that will be better supported now and in the
future.

In order to fine grain the decision of when to enable this server, a
new `activation-fn` has been added to the `lsp-client` structure. If
present, this function overrides `major-modes` and `remote?`, in case
that any of them are also present when registering a new language
client.

In particular, the new Flow client considers that the file is a Flow
file if the following conditions are met:

- The major mode is one of js-mode, js2-mode, or flow-js2-mode.
- The file resides in a Flow project (there is a .flowconfig file).
- The file has a `// @flow` or `/* @flow */` tag.


yyoncho:
Thanks!

PR #604, opened 2019-01-20
Log to buffer rather than messages

### LINKED ISSUE #599: Server logging messing with the message area
**Describe the bug**
When loading large projects, php-language-server produces a lot of logging about what it parses, which seems to make messages in the echo area disappear and eldoc tips flicker.  

Apart from it makes emacs seem flaky while it's scanning, it makes the UX suffer as messages disappear to quickly to be read, and helpful prompts, like the `C-x-`one gets after pressing a prefix key, gets lost.

**To Reproduce**
You'll need a large PHP codebase:
```sh
cd /tmp
wget https://www.drupal.org/download-latest/tar.gz
tar -zxf tar.gz
```
Open `drupal-8.6.7/core/lib/Drupal.php` and check the message buffer to see the logging. Notice that pressing `C-x` looses the usual prompt and eldoc messages flicker.

**Expected behavior**
No flickering and that messages stick around as they used to.

lsp-php

**OS**
Ubuntu 18.10

**Error callstack**
No errors.

**Attempted fixes**
I've tried a few hacks to try and fix it, without success: I've tried making `lsp-message` use `(current-message)` instead of `eldoc-last-message` and I've tried making it just call `(message)`, neither which seemed to do much difference. Tried copying `:ignore-regexps '("\"message\":\"Parsing file:" "\"message\":\"Restored .*from cache")` from https://github.com/emacs-lsp/lsp-php/, which suppressed the messages in the message buffer, but also seemed to kill everything that lsp-ui added (sideline, docs, etc).

xendk:
Closes #599

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -478,7 +478,7 @@ must be used for handling a particular message.")
 (defvar lsp--tcp-port 10000)
 
 ;; Buffer local variable for storing number of lines.
-(defvar lsp-log-lines)
+(defvar lsp--log-lines)
 
 (cl-defgeneric lsp-execute-command (server command arguments)
   "Ask SERVER to execute COMMAND with ARGUMENTS.")
@@ -510,7 +510,7 @@ FORMAT and ARGS i the same as for `message'."
         (setq log-buffer (get-buffer-create "*lsp-log*"))
         (with-current-buffer log-buffer
           (view-mode 1)
-          (set (make-local-variable 'lsp-log-lines) 0)))
+          (set (make-local-variable 'lsp--log-lines) 0)))
       (with-current-buffer log-buffer
         (let* ((current-point (point))
                ;; Count newlines in message.
@@ -522,14 +522,14 @@ FORMAT and ARGS i the same as for `message'."
                (at-bottom (eq current-point (point-max))))
           (goto-char (point-max))
           (insert (concat (apply 'format format args) "\n"))
-          (setq lsp-log-lines (+ lsp-log-lines newlines 1))
-          (when (and (integerp lsp-log-max) (> lsp-log-lines lsp-log-max))
-            (let ((to-delete (- lsp-log-lines lsp-log-max)))
+          (setq lsp--log-lines (+ lsp--log-lines newlines 1))
+          (when (and (integerp lsp-log-max) (> lsp--log-lines lsp-log-max))
+            (let ((to-delete (- lsp--log-lines lsp-log-max)))
               (save-excursion
                 (goto-char (point-min))
                 (forward-line to-delete)
                 (delete-region (point-min) (point))
-                (setq lsp-log-lines lsp-log-max))))
+                (setq lsp--log-lines lsp-log-max))))
           (unless at-bottom
             (goto-char current-point)))))))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -477,6 +477,9 @@ must be used for handling a particular message.")
 
 (defvar lsp--tcp-port 10000)
 
+;; Buffer local variable for storing number of lines.
+(defvar lsp-log-lines)
+
 (cl-defgeneric lsp-execute-command (server command arguments)
   "Ask SERVER to execute COMMAND with ARGUMENTS.")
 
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -2555,8 +2581,8 @@ PARSER is the workspace parser used for handling the message."
       (signal 'lsp-unknown-message-type (list json-data)))))
 
 (defconst lsp--default-notification-handlers
-  (lsp-ht ("window/showMessage" 'lsp--window-show-message)
-          ("window/logMessage" 'lsp--window-show-message)
+  (lsp-ht ("window/logMessage" 'lsp--window-log-message)
```
yyoncho: changing showMessage to logMessage seems to be incorrect or not relevant to the PR

lsp-mode.el
```diff
@@ -446,15 +450,42 @@ must be used for handling a particular message.")
...
-(defun lsp-message (format &rest args)
-  "Wrapper over `message' which preserves the `eldoc-message'.
-FORMAT with ARGS are the original message formats."
-  (let ((inhibit-message lsp-inhibit-message)
-        (last eldoc-last-message))
-    (apply #'message format args)
-
-    (when lsp-inhibit-message
-      (lsp--eldoc-message last))))
+(defun lsp-log (format &rest args)
```
yyoncho: can you defalias `lsp-message`? -  it is used from the other packages like lsp-java (it was public)?

lsp-mode.el
xendk: Well, I thought it would be confusing to have a function called `lsp--window-show-message` not actually show any message.

At first I'd implemented `lsp--window-log-message` alongside `lsp--window-show-message` and had `window/logMessage` and `window/showMessage` route to the appropriate, but then all the `Parsing ..` lines of php-language-server ended up in the message buffer again, so apparently it uses showMessage for those.

lsp-mode.el
```diff
@@ -110,10 +110,14 @@
   :group 'lsp-mode
   :type 'boolean)
 
-(defcustom lsp-inhibit-message t
-  "If non-nil, inhibit the message echo via `inhibit-message'."
-  :type 'boolean
-  :group 'lsp-mode)
+(defcustom lsp-log-max 1000
```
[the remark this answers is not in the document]
xendk: Took 1000 from the `message-log-max`. Seems to me to be a decent default. I don't think most end-users wont be using the log buffer, so keeping it low to just catch the latest messages for when one needs to catch something.

People regularly looking at the logs can bump it.

How about using `message-log-max` as the default value?

xendk:
I'll give it another pass when I got time, @arnested mentioned `view-mode` to me which would probably also be nice.

lsp-mode.el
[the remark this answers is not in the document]
yyoncho: Make sense.

lsp-mode.el
yyoncho: Are we talking about the same thing? I mean that "window/showMessage" should not be changed to "window/logMessage".

lsp-mode.el
arnested: Maybe use `make-obsolete-variable` instead of `defalias`?

yyoncho:
I did some testing, it works very well and it is definitely easier to track the logs. We may even font-lock some of the messages for better readability.

1. You have to calculate the size of the message after replacing the format params, consider the following message (the format is `<<<< jdtls:19137\n%s`):
```
<<<< jdtls:19137
{
  "jsonrpc": "2.0",
  "method": "window/logMessage",
  "params": {
    "type": 3,
    "message": "Jan 22, 2019 8:47:29 AM >> document/signatureHelp"
  }
}
```
2. Consider adding a function `lsp-show-log` (might not be part of this PR).
3. Change rename `lsp-log-lines` to `lsp--log-lines` since it is private.

xendk:
@yyoncho have an example for a log showing command? `(pop-to-buffer "*lsp-log*")` doesn't seem of much.

yyoncho:
Thank you :heart:!
And congratulations on your first lsp-mode PR :tada:!

I guess the other option is using display-buffer - not sure which one is better fit.

PR #607, opened 2019-01-21
Fix action support for servers that return a CodeAction instead of a Command

danielmartin:
The new TypeScript server can return a list of CodeAction structures
instead of a list of Commands. This means that, with the current code,
the subsequent call to 'workspace/executeCommand' was performed with
incorrect arguments, causing the refactor to fail.

With this patch, the response of a 'textDocument/codeAction' is
correctly parsed either as a CodeAction or a Command, and tsserver,
and potentially other language servers, now perform the refactor
operation correctly.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2423,18 +2423,40 @@ RENDER-ALL - nil if only the signature should be rendered."
 
 (defalias 'lsp-get-or-calculate-code-actions 'lsp-code-actions-at-point)
 
+(defun lsp--execute-code-action (action)
+  "Parses a code action represented as a CodeAction LSP type."
+  ;; If we have edits, we can apply them directly without incurring in
+  ;; another roundtrip.
+  (when-let ((edit (gethash "edit" action)))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -2423,18 +2423,40 @@ RENDER-ALL - nil if only the signature should be rendered."
...
+                             (gethash "arguments" action)))
+
+(defun lsp--execute-command-or-code-action (action)
+  "Parses and calls 'workspace/executeCommand' on the result of a
+'textDocument/codeAction' call, which can be a Command or a
+CodeAction type."
+  (when-let ((command (gethash "command" action)))
+    ;; If we have a "command" and it's of string type, we received a
+    ;; Command; otherwise, a CodeAction.
+    (if-let ((is-command? (stringp command)))
```
yyoncho: seems like is-command? is not used and if-let could be changed to if

yyoncho:
Thanks!

PR #608, opened 2019-01-21
Add Swift LLDB compatibility to the supported languages table

danielmartin:
Swift was tested with success when LLDB was integrated into
dap-mode. You only need to compile the fork version of LLDB, as the
manual in dap-mode explains.


yyoncho:
Thanks.

PR #609, opened 2019-01-22
adds rjsx-mode as a possible flow mode

benbot:
flow is very popular in react development land, so using flow with rjsx-mode won't be an uncommon case.


yyoncho:
Thanks!

PR #611, opened 2019-01-22
[#610] Create custom activation function for typescript servers

### LINKED ISSUE #610: Create custom activation function for typescript servers (was typescript/tsx assumes spacemacs usage)
**Describe the bug**
The current version activates the typescript lang server for typescript-tsx-mode which only exists for spacemacs users. Would be good to add a faq entry with a suggested typescript/tsx configuration like what I have below. This is just a copy of the spacemacs approach, but it is useful to demonstrate how to target a language server with something like web-mode that supports multiple languages.;

Here's a faq entry you could drop in:

* Web-mode, by default is set up for an HTML language server. If you use web-mode for TSX/React, you can define a derived mode as follows that will trigger the use of the typescript language server. This is not necessary if you are using spacemacs since it already handles the creation of the derived mode.

```
(use-package web-mode
  :config
  (define-derived-mode typescript-tsx-mode web-mode "TypeScript-tsx")
  (add-to-list 'auto-mode-alist '("\\.tsx\\'" . typescript-tsx-mode)))
```

cwhatley:
Fixes #610 

Typescript + TSX and sometimes Javascript + JSX are edited with web-mode which defaults to wanting an HTML language server in the current code.

This switches typescript-javascript over to using an activation function which checks for the known modes for javascript/typescript as well as checking for a file using web-mode to edit JSX or TSX.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -108,11 +108,19 @@ finding the executable with variable `exec-path'."
   :risky t
   :type '(repeat string))
 
+(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename major-mode)
+  "Checks if the javascript-typescript language server should be enabled
+based on FILE-NAME and MAJOR-MODE"
+  (or (member major-mode '(typescript-mode typescript-tsx-mode js-mode js2-mode rjsx-mode))
+      (and (eq major-mode 'web-mode)
+           (or (string-suffix-p ".tsx" filename t)
+               (string-suffix-p ".jsx" filename t)))))
+
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection
                                    (-const `(,lsp-clients-javascript-typescript-server
                                              ,@lsp-clients-typescript-javascript-server-args)))
-                  :major-modes '(typescript-mode typescript-tsx-mode js-mode js2-mode rjsx-mode)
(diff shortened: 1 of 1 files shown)


cwhatley:
You mean js under web mode? I suppose it would, but I hadn’t considered anyone using web-mode for basic javascript or JSX in files with a .js extension.

cwhatley:
I only put it there to make the change as specific as possible, but I'm fine with taking it out if you prefer.

danielmartin:
Looks good. Visual Studio Code also supports activation events based on filename extensions. Sometimes the major mode is not enough to make a decision.

yyoncho:
@cwhatley 
> but I'm fine with taking it out if you prefer.

The decisions in `lsp-mode` are based on the user feedback/preferences and real scenarios.

I believe the same function might be used for the other JS server which is the recommended one for JS/TS.

Thank you :heart:!
And congratulations on your first `lsp-mode` PR :tada:!

twlz0ne:
Where can I find the `typescript-tsx-mode`?

yyoncho:
@twlz0ne you may get the definition from here https://github.com/emacs-lsp/lsp-mode/issues/610 but I think that do not need it at all since after this PR all files in web-mode  with the proper extension(e. g. tsx) will automatically be assigned to this mode.

PR #621, opened 2019-01-23
Ignore telemetry/event

### LINKED ISSUE #619: HTML Language server: "Unknown method: telemetry/event"
**Describe the bug**
When using the [HTML language server](https://github.com/vscode-langservers/vscode-html-languageserver) in an embeded `<script>` tag, lsp-mode spams `Warning (lsp-mode): Unknown method: telemetry/event` in the `*Warnings*` buffer.

**To Reproduce**
1. Install HTML language server
2. Run `lsp` in a html buffer with a script tag
3. Type something in the script tag

**Expected behavior**
Not spamming the `*Warnings*` buffer with useless messages.

HTML language server

**OS**
Arch Linux

danielmartin:
Fixes #619 

Some servers send this event to track usage, etc., polluting the
warnings log. Nothing we want to do for now, so ignore them.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2901,7 +2901,8 @@ PARSER is the workspace parser used for handling the message."
           ("window/logMessage" 'lsp--window-log-message)
           ("textDocument/publishDiagnostics" 'lsp--on-diagnostics)
           ("textDocument/diagnosticsEnd" 'ignore)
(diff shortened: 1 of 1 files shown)


yyoncho:
I think that each of the servers has a setting which disables these messages, at least I have seen something like that in vscode. We have to investigate at some point.

PR #626, opened 2019-01-25
Add alias for file-local-name to allow compatibiliy with emacs 25.*

bergmannf:
The `file-local-name` function was only introduced in emacs 26.* and lsp-mode
breaks on 25.* installations (e.g. Ubuntu 18.04 / 18.10).


yyoncho:
Thanks!

Issue #612, opened 2019-01-22
5 failing tests while building on Debian


thkoch2001:
I'm packaging version 6.0 now for Debian. 5 tests fail now:

1.   FAILED  lsp--parser-read--ignored-messages
2.   FAILED  lsp--test-overlapping-updates
3.   FAILED  lsp-file-watch--notifications
4.   FAILED  lsp-file-watch--relative-path
 5.  FAILED  lsp-file-watch--relative-path-glob-patterns

I reported the lsp-file-watch--* failures already in https://github.com/emacs-lsp/lsp-mode/issues/467.

This is the full build log with elisp backtraces and full dependencies:
[lsp-mode_6.0-1_amd64-2019-01-21T21:35:31Z.build.txt](https://github.com/emacs-lsp/lsp-mode/files/2784245/lsp-mode_6.0-1_amd64-2019-01-21T21.35.31Z.build.txt)

Tests 1. and 2. fail with `(void-variable method)` or `(void-variable range)` respectively.

maybe relevant dependencies:

elpa-dash (2.14.1+dfsg-1)
elpa-f (0.20.0-1)
elpa-ht (2.2-1)
elpa-s (1.12.0-2)
elpa-spinner (1.7.3-1)
emacs-nox (1:26.1+1-3)

I've no experience with elisp, so I'd be thankful for help here.

yyoncho:
> Tests 1. and 2 

This is caused by dash being old. We are using smart destructuring which is available only in the latest dash.
@fuco1 do you have plans for releasing new dash version?

thkoch2001:
We only have a few days left for updates to get into the next Debian stable. Do you think there is a way you could change your code to work with dash 2.14? I doubt we will be able to have an upstream release and a debian release of dash 2.20 and handle all possible problems that the dash upgrade might cause in other Debian packages.

yyoncho:
I think that I could do that change if this is the only option. What is the deadline?

Fuco1:
Out of curiosity why are we packaging Emacs packages for Debian?

Yes, new version of dash is quite necessary.  I'm going to release 2.15 right now.  It's been stable and we are just lazy to click the button.

thkoch2001:
@yyoncho Thanks a lot!

The final deadline is March 1st. If an error is found afterwards there is no option to upload a fix anymore.
https://lists.debian.org/debian-devel-announce/2019/01/msg00008.html

[...]

PR #628, opened 2019-01-25
Change lsp-mode to use dash 2.14.1

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/612


thkoch2001:
I applied this patch to version 6.0 and the tests (beside the excluded file-notify tests) went fine this time.
I can just take this single commit and distribute it as a patch in the package. There is no need to release a new version just for this. Thank you!

yyoncho:
Can you wait till the end of the day, I want to test it additionally since the test suite is not full?

thkoch2001:
No worries, I'm not that fast :-)

PR #638, opened 2019-02-05
Consider multiline flow comments to enable the flow language server

danielmartin:
Flow files can have the following header marker:

```
/**
 * @flow
 */
```

This commit enhances the regular expressions we use to match that
pattern.


yyoncho:
Looks good.

PR #640, opened 2019-02-06
Check derived mode when decided to activate Flow

d4ncer:
This change enhances the eligible mode pool by introducing `web-mode`, and uses
`derived-mode-p` to check if the current major mode is either one of the modes
passed or derived from them.

This ensures the Flow LSP is correctly activated for custom modes derived from
one of the popular choices.


yyoncho:
Thank you! :heart: 
And congratulations on your first `lsp-mode` PR :tada:

PR #642, opened 2019-02-08
add support for `less-css-mode`

waymondo:
The most popular package for editing Less CSS files is called `less-css-mode`, not `less-mode`. This registers that mode with the CSS lsp client. I also updated the readme add these css language's support.


yyoncho:
Thank you!  :heart: 
And congratulations for your first `lsp-mode` PR :tada:

PR #645, opened 2019-02-10
Exposed pyls configuration settings

yyoncho:


yyoncho:
@noctuid

noctuid:
Thanks! I believe the type should instead be something like `'(plist)` (defaults for `:key-type` and `:value-type` are symbol and sexp, which is correct here).

yyoncho:
@noctuid fixed and force-pushed.

cslux:
pyls also takes some extra arguments, maybe this should be supported too.

``` emacs-lisp
(defcustom lsp-clients-python-extra-arguments '()
  "Lsp client extra arguments."
  :group 'lsp-python
  :risky t
  :type 'list)

(lsp-register-client
 (make-lsp-client :new-connection (lsp-stdio-connection 
                                   (push "pyls" lsp-clients-python-extra-arguments))
                  :major-modes '(python-mode)
                  :priority -1
                  :server-id 'pyls
                                    (with-lsp-workspace workspace
                                      (lsp--set-configuration lsp-clients-python-settings)))
                  :library-folders-fn (lambda (_workspace)
                                        lsp-clients-python-library-directories)))
```

yyoncho:
@cslux thanks for the review - can you take a look at the latest version?

cslux:
@yyoncho Perfect, works! :+1:

Issue #646, opened 2019-02-13
lsp--uri-to-path not decoding utf-8 path back to multibyte string


wenxin-wang:
If an URI is encoded from a path containing utf-8 characters, `lsp--uri-to-path` does not decode it back to a string containing the original multi-byte characters. This makes `lsp-ui-flycheck` unable to display underlines for diagnostics returned from the server.

**To Reproduce**

Create a file "~/你好/try.py" with content:

```python
import sys
class Test:
    pass
```

See that flycheck does not mark any warning on the buffer, but `lsp-ui-flycheck-list` shows 2 warnings:

```
\344\275\240\345\245\275/try.py
1: pyflakes: 'sys' imported but unused
2: pycodestyle: E302 expected 2 blank lines, found 0
```

`(setq lsp-print-io t)` also confirms that `pyls` is providing the correct diagnostics.

If this file is moved so that its path contains ascii-only characters, the warnings are marked correctly with underlines by flycheck.

**Expected behavior**

If `lsp--uri-to-path` is redefined as shown below, open "try.py" and the two errors will be marked. `lsp-ui-flycheck-list` again shows 2 warnings:

```
你好/try.py
1: pyflakes: 'sys' imported but unused
2: pycodestyle: E302 expected 2 blank lines, found 0
```

The only thing changed in `lsp--uri-to-path` is that `decode-coding-string` is used to decode the unhexed file name. The new definition of `lsp--uri-to-path` is:

```elisp
(defun lsp--uri-to-path (uri)
  "Convert URI to a file path."
  (let* ((url (url-generic-parse-url (url-unhex-string uri)))
	   (type (url-type url))
	   (file (decode-coding-string (url-filename url) locale-coding-system))
	   (file-name (if (and type (not (string= type "file")))
			  (if-let ((handler (lsp--get-uri-handler type)))
			      (funcall handler uri)
			    (error "Unsupported file scheme: %s" uri))
			;; `url-generic-parse-url' is buggy on windows:
			;; https://github.com/emacs-lsp/lsp-mode/pull/265
			(or (and (eq system-type 'windows-nt)
				 (eq (elt file 0) ?\/)
				 (substring file 1))
			    file))))

    (lsp--fix-path-casing
     (concat (-some 'lsp--workspace-host-root (lsp-workspaces)) file-name))))
```

According to [elisp url library's doc](https://www.gnu.org/software/emacs/manual/html_mono/url.html#URI-Encoding):

[...]

PR #647, opened 2019-02-13
lsp--uri-to-path: handle utf-8 path

wenxin-wang:
See https://github.com/emacs-lsp/lsp-mode/issues/646

I had read somewhere that this kind of encoding problems are very tricky to handle. So this current solution may break things.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -768,7 +768,7 @@ On other systems, returns path without change."
   "Convert URI to a file path."
   (let* ((url (url-generic-parse-url (url-unhex-string uri)))
          (type (url-type url))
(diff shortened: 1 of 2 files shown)


yyoncho:
Can you add your case to https://github.com/emacs-lsp/lsp-mode/blob/master/test/lsp-common-test.el ? My main concern is whether it will work against different emacs versions and travis is covering emacs 25.* I will test locally with emacs 27+ .

wenxin-wang:
I added a simple test, is that enough?

yyoncho:
Yes.

Thank you! :heart: 
And congratulations for your first commit! :tada:

PR #648, opened 2019-02-14
lsp-save-logs: New function (Fixes #644).

### LINKED ISSUE #644: Supporting logging to files
The official LSP website comes with an [Inspector app ](https://microsoft.github.io/language-server-protocol/inspector/)that allows you to examine a detailed log of all requests/responses/notifications between a language server and a client. `lsp-mode` should support logging communication to a JSON file that can be read by the app, which would make debugging easier than it is now, for both developers and users.

vibhavp:


lsp-mode.el
```diff
@@ -1416,9 +1423,72 @@ If WORKSPACE is not provided current workspace will be used."
...
+   :process-time process-time
+   :method method
+   :id id
+   :type type
+   :body (or arg (plist-get body :params))))
+
+(defun lsp-save-logs (file)
+  (interactive "F")
+  (let ((trace (reverse (lsp--client-message-trace (lsp--workspace-client
+                                                    (nth 0 (lsp-workspaces))))))
```
yyoncho: we may interactively ask which workspace to save if there are more than one workspaces active

vibhavp: Is there a way to map workspaces to a user-selectable string?

yyoncho: yeah - lsp--workspace-print

vibhavp: Done, thanks.

PR #649, opened 2019-02-15
Add enh-ruby-mode as alternative to ruby-mode

richarddewit:
[`enh-ruby-mode`](https://github.com/zenspider/enhanced-ruby-mode) is an alternative `ruby-mode`. This PR is to add that mode to the Ruby LSP client.


yyoncho:
Thank you and congrats for your first `lsp-mode` PR :tada:

PR #653, opened 2019-02-17
Add `textDocument/documentLink` support.

vibhavp:
* `lsp-enable-links`, `lsp-links-check-internal`: New variables for
  customizing `documentLink` behavior (enabled by default).

* `lsp--update-document-links`: Main entrypoint, used by
  `lsp--set-document-link-timer`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2282,7 +2282,7 @@ Applies on type formatting."
                 (find-file file)))
       ((or "http" "https") (browse-url url))
       (type (if-let ((handler (lsp--get-uri-handler type)))
-                (funcall handler uri)
+                (funcall handler url)
               (signal 'lsp-file-scheme-not-supported (list url)))))))
 
 (defun lsp--document-link-keymap (link)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -163,6 +163,8 @@ the buffer when it becomes large."
   "Timed out while waiting for a response from the language server" 'lsp-error)
 (define-error 'lsp-capability-not-supported
   "Capability not supported by the language server" 'lsp-error)
+(define-error 'lsp-file-scheme-not-supported
+  "Unsupported file scheme" 'lsp-error)
 
 (defcustom lsp-auto-guess-root nil
   "Automatically guess the project root using projectile/project."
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -2207,6 +2226,72 @@ Applies on type formatting."
...
+                           (seq-do
+                            (lambda (link)
+                              (with-current-buffer buffer
+                                (-let* (((&hash "range") link)
+                                        (start (lsp--position-to-point (gethash "start" range)))
+                                        (end (lsp--position-to-point (gethash "end" range)))
+                                        (button (make-button start end 'action
+                                                             (lsp--document-link-keymap link))))
+                                  (push button lsp--link-overlays))))
+                            links))))
```
yyoncho: you may pass :mode 'alive and the callback wont run if the buffer was killed.

lsp-mode.el
```diff
@@ -2207,6 +2226,72 @@ Applies on type formatting."
...
+    (cancel-timer lsp--links-idle-timer)
+    (setq-local lsp--links-idle-timer nil)))
+
+(defun lsp--document-link-handle-target (url)
+  (let* ((parsed-url (url-generic-parse-url (url-unhex-string url)))
+         (type (url-type parsed-url))
+         (file (decode-coding-string (url-filename parsed-url)
+                                     locale-coding-system)))
+    (if (string= type "file")
```
yyoncho: We support multiple protocol schema in addition to `file:`. For example, jdt ls supports `jdt:` and `chelib:` - take a look at `lsp--uri-to-path`. I suspect that if you check for http/https and then use browse-url and fallback to lsp--uri-to-path in case it is not `http`/`https` will cover more cases.

vibhavp:
> Looks good. I believe that the link updating mechanism must be consolidated with the lenses(and eventually with documentColorProvider?) to avoid additional traffic when there are no changes to the document. You may take a look at `lsp--lens-idle-function` - we may handle the links in similar way - request them only when document has changed. What do you think?

@yyoncho That's why `lsp--set-document-link-timer` is called in `lsp-on-change`

yyoncho:
> @yyoncho That's why `lsp--set-document-link-timer` is called in `lsp-on-change`

Yeah, my bad - didn't pay attention to the fact that the timer is scheduled one time. Can add one more line to the `lsp--set-document-link-timer` to cancel the old timer if it is present?

PR #654, opened 2019-02-18
Fix custom settings definitions

ilysym:
Change `'bool` (which is not supported) to `'boolean` in `defcustom` forms. Add missing custom group definitions referenced in `defcustom` forms so they could be discovered in customization UI. Additionally, change some tag and docstrings to make them more consistent with other Emacs settings.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -30,6 +30,12 @@
 
 
 ;; Python
+
+(defgroup lsp-python nil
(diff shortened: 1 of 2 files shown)


yyoncho:
Thank you for improving the customization section! :heart: 
And congrats for your first `lsp-mode` commit! :tada:

PR #658, opened 2019-02-20
lsp-clients: fix disabling snippet completion for go

fjl:
This adds the disableFuncSnippet initialization option which is recognized by the bingo
language server. See https://github.com/saibing/bingo/blob/master/langserver/initialization.go for supported initialization options.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -434,12 +434,13 @@ defaults to half of your CPU cores."
 (defun lsp-clients-go--make-init-options ()
   "Init options for golang."
   `(:funcSnippetEnabled ,(lsp-clients-go--bool-to-json lsp-clients-go-func-snippet-enabled)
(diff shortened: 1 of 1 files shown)


fjl:
BTW, I don't really understand why there are two ways to disable snippet completion for
Go. There is `lsp-enable-snippet` to announce snippet capability and `lsp-client-go-func-snippet-enabled` to configure the language server via initialization options.

yyoncho:
> lsp-client-go-func-snippet-enabled

This is go language server specific setting while lsp-enable-snippet is the standard setting in lsp protocol - refer to the spec https://microsoft.github.io/language-server-protocol/specification

yyoncho:
Thank you for the PR and congrats for your first `lsp-mode` PR! :tada:

fjl:
I'll try to also fix the bingo server so it doesn't send snippets if they're not announced in client capabilities.

PR #659, opened 2019-02-20
lsp-mode: fix boolan conversion in client capabilities

fjl:
This change ensures that "snippetSupport" is sent as JSON true or false
instead of true or null.


yyoncho:
Just curious - why do you want to disable snippets in go?

fjl:
It's just personal preference. All I need is simple completion. I've been using the `company-go` wrapper for many years now and it worked great until recently. Trying to switch to LSP-based setup today, let's see what other issues I'll find.

yyoncho:
@fjl make sense - if you have interest you may try to add support for debugger for https://github.com/yyoncho/dap-mode via https://github.com/Microsoft/vscode-go.git .

PR #662, opened 2019-02-20
lsp-clients: add lsp-clients-go-server-args

fjl:
There are two client definitions for go: one hard-codes go-langserver, the
other one uses lsp-clients-go-server which is usually "bingo". The flags variable
isn't used at all. Remove it and add lsp-clients-go-server-args to match all other
languages.


yyoncho:
Thank you!

PR #672, opened 2019-02-27
lsp--client-capabilities: specify content format of hover

Khady:
This is necessary so that servers know lsp-ui is able to render markdown.

One thing that I would like to add is a way for the user to customize this value. Someone who doesn't want to use lsp-ui but only display the eldoc message in the minibuffer will probably deactivate markdown for example. I am not sure where this configuration value should be. In lsp-mode or in lsp-ui?

https://microsoft.github.io/language-server-protocol/specification#initialize

```typescript
/**
 * The client supports the follow content formats for the content
 * property. The order describes the preferred format of the client.
 */
contentFormat?: MarkupKind[];
```

https://github.com/emacs-lsp/lsp-ui/issues/246

I tested only with ocamlmerlin-lsp (a lsp server for ocaml).


yyoncho:
Thank you for investigating and fixing this! :heart: 
And congrats for your first `lsp-mode` PR! :tada:

PR #675, opened 2019-02-28
Use HTTPS for package-archive URLs in lsp-start-plain.el

hpdeifel:
Fetching and installing packages over plain HTTP might not be a good idea.


yyoncho:
Thank you!

PR #680, opened 2019-03-01
lsp--flymake-backend: Fix typo in case expression

hpdeifel:
This leads to flymake diagnostics with `nil` as type, which in turn leads to
`nil` in the overlay property of the diagnostic. On flymake-1.0.5 (The one
shipped with emacs 26.1 is not affected), this triggers an error in the report
handler, which disables the lsp backend, which leads to stale error indicators
in the buffer.


yyoncho:
Thanks for tracking and fixing this!

Issue #674, opened 2019-02-27
Flymake diagnostics don't update


hpdeifel:
**Describe the bug**
Flymake diagnostics are reported correctly for a few (just one?) edit operations and are not updated any more after further edits. Disabling and re-enabling `flymake-mode` fixes the problem for another edit and then it gets stuck again. This happens consistently for me in `rls` projects (I haven't tried other language servers). 

Flycheck integration works fine. `eglot`'s flymake integration also works fine.

**To Reproduce**
1. `emacs -Q`
2. load `lsp`
3. open a file in a rust project
4. `M-x lsp`
5. Type some code into the buffer

**Expected behavior**
Flymake diagnostics are correctly displayed and updated while typing.

**Which Language Server did you use**
`rls-preview 1.31.6 (bfa1371 2018-12-07)`

**OS**
Emacs :slightly_smiling_face: (actually Arch Linux)

yyoncho:
I am unable to reproduce it with emacs27 . Which emacs version do you use?

Can you repro with using https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el (you will have to manually install rust-mode or add it to the list of packages to install)?

hpdeifel:
I'm on Emacs 26.

I can't reproduce the exact behavior with `lsp-start-plain.el`; no idea what's different between `emacs -Q` with manually loaded packages and this temporary elpa directory...

Anyway, it's still wonky. On small projects, it seems to work alright but on larger ones it exhibits similar but not quite the same behavior: Typing very fast sometimes leads to stale flymake diagnostics that obviously refer to previous buffer content. In this case, typing a few more letters (or killing and yanking something) fixes it.

I have tried and reproduced this with the built-in flymake from Emacs 26 and the latest version from elpa. I'll try a few public rust repos so that you can try to repro with the exact same code as I have.

Since this seems to be some kind of racy heisenbug, what should I do to debug it myself?

yyoncho:
I think that I might have found what might cause this behavour. Can you retest with this change, seems like it might fix your issue and report back to extend the solution with the proper cleanup:

```
  (defun lsp--flymake-after-diagnostics ()
    "Handler for `lsp-after-diagnostics-hook'"
    (when lsp--flymake-report-fn
      (lsp--flymake-backend lsp--flymake-report-fn)
      ;; (remove-hook 'lsp-after-diagnostics-hook 'lsp--flymake-after-diagnostics t)
[...]

PR #681, opened 2019-03-01
lsp--flymake-backend: Force flymake to delete old diagnostics

hpdeifel:
This fixes a bug where flymake would not delete old diagnostics if the buffer
hasn't changed since the previous call to the report function. This can happen
e.g. if a change in another file removes the errors.

See https://github.com/joaotavora/eglot/issues/159 for details.

Relates to #674


yyoncho:
Thank you!

PR #684, opened 2019-03-02
Show flymake diagnostics immediately when they arrive

### LINKED ISSUE #674: Flymake diagnostics don't update

hpdeifel:
This changes the flymake backend to show diagnostics as soon as they come in from the language server instead of only when flymake decides to call our backend function. This fixes some occurrences of stale diagnostics and is also how eglot works.

Fixes #674

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1047,14 +1047,16 @@ WORKSPACE is the workspace that contains the diagnostics."
 
   (defun lsp--flymake-after-diagnostics ()
     "Handler for `lsp-after-diagnostics-hook'"
-    (when lsp--flymake-report-fn
-      (lsp--flymake-backend lsp--flymake-report-fn)
-      (remove-hook 'lsp-after-diagnostics-hook 'lsp--flymake-after-diagnostics t)))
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for your contribution! I did the reporting in the other way around since we got some complaints for flycheck that the updates are pushed by lsp-mode without being requested by flycheck so I did the flymake in this way which apparently does not work.

PR #687, opened 2019-03-03
Byte compilation test

yyoncho:
- added byte compilation test and fixed all the warnings.

I have used with-no-warnings a few times since I was unable to make eval-when(and)-compile work. We may revisit this code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,3 +46,4 @@ flycheck_*.el
 
 /*-autoloads.el
 /*-pkg.el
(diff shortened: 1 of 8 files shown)


yyoncho:
@vibhavp I am proceeding with this PR due to https://github.com/emacs-lsp/lsp-mode/commit/24d421dc7e0b0e4d96b468e870e8161656a3142e#commitcomment-32570721 

(cc @rompy )

PR #689, opened 2019-03-03
lsp--shutdown-workspace: Send 'shutdown' as jsonrpc request

### LINKED ISSUE #685: shutdown is send as notification instead of as request
**Describe the bug**
Trying to call `lsp-shutdown-workspace` on `rls` projects results in an error with `lsp-unknown-message-type`. 

This is what happens:

1. lsp-mode sends `shutdown` as a notification
2. rls is strict about the spec and doesn't accept this. It responds with an error.
3. This error is not the reply to a request (because there is no request to reply to) but also a notification.
4. lsp-mode in turn doesn't understand this in `lsp--get-message-type`

The spec says `shutdown` should be a request, not a notification.

rls-preview 1.31.6 (bfa1371 2018-12-07)

**OS**
GNU Emacs 27.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.5) of 2019-03-01

**Error callstack**

```
Debugger entered--Lisp error: (lsp-unknown-message-type #<hash-table equal 3/65 0x15723f3ee85d>)
  signal(lsp-unknown-message-type (#<hash-table equal 3/65 0x15723f3ee85d>))
  (if (gethash "method" json-data nil) 'notification (signal 'lsp-unknown-message-type (list json-data)))
  (if (gethash "id" json-data nil) (if (gethash "error" json-data nil) 'response-error (if (gethash "method" json-data nil) 'request 'response)) (if (gethash "method" json-data nil) 'notification (signal 'lsp-unknown-message-type (list json-data))))
  lsp--get-message-type(#<hash-table equal 3/65 0x15723f3ee85d>)
[...]
```

And messages:

```
>>> rls:12883(async)
Content-Length: 65

{
  "jsonrpc": "2.0",
  "method": "shutdown",
  "params": null
}

>>> rls:12883(async)
Content-Length: 61

{
  "jsonrpc": "2.0",
  "method": "exit",
  "params": null
}

<<<< rls:12883
{
  "jsonrpc": "2.0",
  "error": {
    "code": -32600,
    "message": "Invalid request"
  },
  "id": null
}
```

hpdeifel:
Previously it was sent as a notification, which tripped over servers like rls.
The spec says it should be a request, so make it one.

Fixes #685

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1749,7 +1749,7 @@ If NO-MERGE is non-nil, don't merge the results but return alist workspace->resu
 (defun lsp--shutdown-workspace ()
   "Shut down the language server process for ‘lsp--cur-workspace’."
   (with-demoted-errors "LSP error: %S"
-    (lsp-request "shutdown" nil :no-wait t)
+    (lsp-request "shutdown" nil)
     (lsp-notify "exit" nil))
   (lsp--uninitialize-workspace))
 


yyoncho:
Thank you!

Issue #395, opened 2018-07-11
Code action command argument that is an array gets executed as an object


bubba:
When the client receives a code action response containing an array as an argument like so:
```json
{
  "result": [
    {
      "command": "10621:hsimport:import",
      "arguments": [
        [
          "file:///Users/luke/Desktop/test/src/Lib.hs",
          "Test.Hspec"
        ]
      ],
      "title": "Import module Test.Hspec"
    }
  ],
  "jsonrpc": "2.0",
  "id": 264
}
```

The client executes the command with the first item in the array as an object key, and the rest as an array:
```json
{
  "jsonrpc": "2.0",
  "method": "workspace/executeCommand",
  "params": {
    "command": "10621:hsimport:import",
    "arguments": {
      "file:///Users/luke/Desktop/test/src/Lib.hs": [
        "Test.Hspec"
      ]
    }
  },
  "id": 268
}
```

Is this intentional? LanguageClient-neovim and VS Code keep the arguments as an array
```json
//vscode
{
  "jsonrpc": "2.0",
  "id": 14,
  "method": "workspace/executeCommand",
  "params": {
    "command": "34669:hsimport:import",
    "arguments": [
      [
        "file:///Users/luke/Desktop/test/src/Lib.hs",
        "Test.Hspec"
      ]
    ]
  }
}

// LangaugeClient-neovim
{
  "jsonrpc": "2.0",
  "method": "workspace/executeCommand",
  "params": {
    "arguments": [
      [
        "file:///Users/luke/Desktop/test/src/Lib.hs",
        "Test.Hspec"
      ]
    ],
    "command": "32954:hsimport:import"
  },
  "id": 31
}
```

yyoncho:
This is a bug - the LSP mode is deserializing arrays as lists but it is serializing lists as objects. It is not clear how this could be fixed. Change the deserializer to deserialize JSON arrays into vectors is a potential solution but this could break clients/extensions. @vibhavp any thoughts?

vibhavp:
Do any clients use the `arguments` field? IIRC, the field is supposed to be sent as is to the server.

yyoncho:
@vibhavp if we change the type to vector it will every result structure, not only the arguments field. I am reffering the following function: 

```elisp
(define-inline lsp--read-json (str)
  (inline-quote
    (let* ((json-array-type 'list)
            (json-object-type 'hash-table)
            (json-false nil))
      (json-read-from-string ,str))))
```

[...]

PR #691, opened 2019-03-04
Native json support

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/210
Fixes https://github.com/emacs-lsp/lsp-mode/issues/395

* Ajusted lsp-mode to use native json parsing
* introduced 2 new flags:

** lsp-use-native-json
** lsp-json-use-lists - could be used to switch to the list serialization in case
  there is a problem with the vector serialization.

* Addressed few bottlenecks in the code.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -121,6 +121,16 @@
   :group 'lsp-mode
   :type 'boolean)
 
+(defcustom lsp-use-native-json nil
+  "If non-nil, use native json parsing if available."
+  :group 'lsp-mode
+  :type 'boolean)
+
+(defcustom lsp-json-use-lists nil
+  "If non-nil, use lists instead of vectors when doing json deserialization."
+  :group 'lsp-mode
+  :type 'boolean)
+
 (defcustom lsp-log-max message-log-max
   "Maximum number of lines to keep in the log buffer.
 If nil, disable message logging.  If t, log messages but don’t truncate
@@ -623,7 +633,7 @@ FORMAT and ARGS i the same as for `message'."
   "Merge RESULTS by filtering the empty hash-tables and merging the lists.
 METHOD is the executed method so the results could be merged
 depending on it."
-  (pcase  (seq-remove #'null results)
+  (pcase (--map (if (vectorp it) (append it nil) it) (-filter 'identity results))
     (`() ())
     ;; only one result - simply return it
     (`(,fst) fst)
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -3326,15 +3338,19 @@ WORKSPACE is the active workspace."
...
          (json-false nil))
     (if use-native-json
         (with-no-warnings
-          (json-parse-string str :object-type 'hash-table
-                             :null-object nil :false-object nil))
+          (with-temp-buffer
+            (json-parse-string str
+                               :object-type 'hash-table
+                               :null-object nil
+                               :false-object :json-false)))
```
hpdeifel: Switching from `nil` to `:json-false` leads to conditionals with json values to suddenly succeed. For example, `(gethash "resolveProvider" ...)` in `lsp--sort-completions` now returns a truth-value (`:json-false`) for ccls, leading to errors during completion. There may be other similar instances.

hpdeifel:
By the way, is `lsp-json-use-lists` intended to go away at some point? Why do you do all this seq generic stuff instead of committing to vectors?

lsp-mode.el
```diff
@@ -2725,17 +2735,18 @@ RENDER-ALL - nil if only the signature should be rendered."
...
                  (param (seq-elt parameters active-parameter))
                  (selected-param-label (-some->> param (gethash "label")))
                  (start (if (stringp selected-param-label)
                             (s-index-of selected-param-label result)
-                          (seq-first selected-param-label)))
+                          (seq-first (append selected-param-label nil))))
                  (end (if (stringp selected-param-label)
                           (+ start (length selected-param-label))
-                        (seq-elt selected-param-label (1- (seq-length selected-param-label))))))
+                        (last (append selected-param-label nil)))))
```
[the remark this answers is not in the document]
yyoncho: it should be `cl-second`

lsp-mode.el
yyoncho: Thanks for spotting this. We had to evaluate whether to keep the current(incorrect) behaviour of fix all places which do if over a lsp result.

hpdeifel: I'd vote for `:false`/`:null`, as in the defaults of the emacs27 json library, or even `nil` instead of `:false`, that way you preserve truth value.

yyoncho: We already use `:json-false` for false when doing serialization.

yyoncho:
> By the way, is lsp-json-use-lists intended to go away at some point? 

Yes, we might not add it at all.

> Why do you do all this seq generic stuff instead of committing to vectors?

IMO using seq will make the code more robust.  I am not sure whether there is vector-filter or similar functions or we should resort to smt like cl-loop.

lsp-mode.el
vibhavp: But that doesn't affect us much as we don't reuse the objects we create during serialization. On the other hand, using a non-`nil` value while working with values we get from the language server would inadvertently result in a lot of broken code. Every `if` and conditional construct would have to be rewritten to check for the new "false", which would just result in inorganic and unidiomatic lisp code, prone to breaking.

yyoncho: @vibhavp there are few places we reuse the object, e. g. - lenses, diagnostics, code actions.

> which would just result in inorganic and unidiomatic lisp code, prone to breaking.

I agree that it will be ugly, error-prone, difficult to migrate and support. And I agree with you that it is better to keep it as it is at least for now...

A link to a code that deals with that issue - I am converting the nil to `:json-false` to make it work. 
https://github.com/emacs-lsp/lsp-java/blob/master/lsp-java-boot.el#L68

PR #703, opened 2019-03-13
Elixir-ls executable depends on system version

axelson:
Windows should use `language_server.bat`
*nix should use `language_server.sh`

Here is a link to the end of the thread where an end user encountered the issue: https://elixirforum.com/t/emacs-elixir-setup-configuration-wiki/19196/43?u=axelson


yyoncho:

axelson:
Thanks!

PR #707, opened 2019-03-13
[RUST] add progress indicator modeline

yyoncho:


yyoncho:
@hpdeifel you might be interested to try/test this PR. I think this will improve user experience, especially after https://github.com/emacs-lsp/lsp-mode/pull/708 is in (Cargo.toml changes will trigger recompiling of the project.)

hpdeifel:
So i've finally looked at this PR, here are my comments:

1. I think it needs a `force-mode-line-update`. Without that, nothing at all showed up for me.
2. It doesn't work well together with `display-time-mode`, as there is no space between the `rls` status and the time. `display-time-mode` doesn't add a space to its mode-line string, I've no idea if other users of `global-mode-string` are expected to do so.
3. I filed [this issue](https://github.com/joaotavora/eglot/issues/236) against `eglot` some time ago: I'm easily distracted by moving things :-) and this PR adds the same problem to `lsp-mode`. I'd admit the status is very handy when rls compiles lots of dependencies, but it gets in the way during normal editing. Note that this point is very subjective, so a config option could be a good idea.

yyoncho:
@hpdeifel 

Thank you for the feedback.

1. and 2.  Which modeline do you use? I have tested only with doom-modeline. 

3. What is your suggestion - adding a flag to disable it or smarter handling of these progress notifications e. g. display them 10ms after they are received unless there is no other notification which will result in not showing the quick messages? We might create some not so noisy controls like spinner which will changes its colour. I want to establish a general approach so we could use it for each server that displays progress stuff.

hpdeifel:
I use [moody](https://github.com/tarsius/moody/commits/master) and it does seem to update correctly with the default mode-line. I assumed that you always need `force-mode-line-update`, as the Elisp info manual states:

> If you alter any of the variables referenced by `mode-line-format` or `header-line-format`, or any other data structures that affect how text is displayed, you should use the function `force-mode-line-update` to update the display.

Regarding 3, I'm not sure. VSCode shows a spinner and the currently building crate name (without "Building", "Indexing"), which feels less distracting on first glance but I haven't used it for longer periods. I'm not sure a delay is useful. Either you're interested in immediate feedback about progress (e.g. when building dependencies) or not at all (e.g. when recompiling the current file because you made a change).

hpdeifel:
I use [moody](https://github.com/tarsius/moody) and it indeed works with the default mode line. I just assumed you'd need `force-mode-line-update`, as the Elisp info manual states:

> If you alter any of the variables referenced by `mode-line-format` or `header-line-format`, or any other data structures that affect how text is displayed, you should use the function `force-mode-line-update` to update the display.

Regarding 3, I'm not sure. VSCode shows a spinner and the currently building crate name (without "Indexing" or "Building"), which feels a bit less distracting on first glance, but I haven't used it for longer periods.

I'm not sure a delay would work. Either I'm interested in immediate feedback about progress (e.g. when compiling large dependency trees) or not at all (e.g. when `rls` recompiles the current file after making a change).

yyoncho:
@hpdeifel I have changed the code to display the message only if there is a message(I guess similar thing is performed by vscode). Also, I have enabled 

> Note that this point is very subjective, so a config option could be a good idea.

There is no need from option since you could remove `lsp-clients-rust-progress-string` from `global-mode-string`. 

I will merge the PR and I will address additional feedback separately (if any).

PR #716, opened 2019-03-16
Improve Flow files detection support

danielmartin:
Enable the Flow LSP client when we encounter a file with the following
comment:

```
/**
 * Documentation about an abstraction.
 *
 * @flow
 */

```
Previously, we enabled the TS client automatically in this case.

Also, this commit adds some unit testing for the Flow detection logic.

### DIFF
--- a/test/lsp-clients-test.el
+++ b/test/lsp-clients-test.el
@@ -1,0 +1,78 @@
+;;; lsp-clients-test.el --- unit tests for lsp-clients.el -*- lexical-binding: t -*-
+
+;; Copyright (C) 2019 Daniel Martín <mardani29@yahoo.es>.
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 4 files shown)


test/lsp-clients-test.el
```diff
@@ -0,0 +1,78 @@
...
+(ert-deftest lsp-flow-wrong-tag-detection ()
+  (should (not (lsp-clients-flow-tag-string-present-p lsp-flow-wrong-tag))))
+
+(ert-deftest lsp-flow-but-not-in-comment-tag-detection ()
+  (should (not (lsp-clients-flow-tag-string-present-p lsp-flow-but-not-in-comment-tag))))
+
+(ert-deftest lsp-flow-should-activate-on-flow-project ()
+  ;; Set `js-mode' ON and check that a Flow project activates the Flow
+  ;; LSP client.
+  (js-mode)
```
yyoncho: for the future reference - haven't tested it but I think that `(let ((major-mode 'js-mode)) ...)` would do the job and it will be faster.

PR #718, opened 2019-03-17
Fix xref-backend-identifier-at-point in case of no identifier

erikarvstedt:
Previously, with no identifier present at point, "nil" was returned
instead of "".


yyoncho:

Issue #715, opened 2019-03-16
mccabe configuration is not respected


raxod502:
**Describe the bug**
Changing settings in `lsp-clients-python-settings` does not affect the operation of `mccabe`.

**To Reproduce**
Open any Python file which has a function whose cyclomatic complexity is greater than 15. Here is an example:

<details>
<summary>Click to expand code</summary>
<p>

```python
def complex_function():
    if True:
        if True:
            if True:
                if True:
                    pass
                else:
                    pass
            else:
                if True:
                    pass
                else:
                    pass
        else:
            if True:
                if True:
                    pass
                else:
                    pass
            else:
                if True:
                    pass
                else:
                    pass
    else:
        if True:
            if True:
                if True:
                    pass
                else:
                    pass
            else:
                if True:
                    pass
                else:
                    pass
        else:
            if True:
                if True:
                    pass
                else:
                    pass
            else:
                if True:
                    pass
                else:
                    pass
```

</details>

Set up `lsp-mode` with `pyls` and enable `lsp-ui` with Flycheck.

**Expected behavior**
No error should be reported as `:plugins.mccabe.enabled` is set to `nil` in `lsp-clients-python-settings`.

**Actual behavior**
I see an error `mccabe: Cyclomatic complexity too high: 16 (threshold 15)` in the sideline from `lsp-ui-sideline`. Furthermore, if I change `:plugins.mccabe.threshold` in `lsp-clients-python-settings` from its default of 15 to, say, 50 and then run `M-x lsp-restart-workspace`, I see no difference at all—the threshold is still reported as 15. Clearly, the configuration in `lsp-clients-python-settings` is not being respected by the LSP server.

**Which Language Server did you use**
`lsp-python`

**OS**
macOS 10.14.3

**Logging**

```json
>>> pyls:36303(async)
Content-Length: 793

{
  "jsonrpc": "2.0",
  "method": "textDocument/codeAction",
  "params": {
    "textDocument": {
      "uri": "file:///Users/raxod502/files/code/web/hyperschedule-scraper/libportal.py"
    },
    "range": {
      "start": {
        "line": 174,
        "character": 0
      },
      "end": {
        "line": 174,
        "character": 7
      }
    },
    "context": {
      "diagnostics": [
        {
          "severity": 2,
          "message": "Cyclomatic complexity too high: 35 (threshold 15)",
          "range": {
            "end": {
              "character": 8,
              "line": 174
            },
            "start": {
              "character": 0,
[...]

PR #719, opened 2019-03-17
Should not include the comma in the enum list.

nasyxx:
Although I'm not sure if it's a good idea to remove the comma here, it can reduce the confusion when setting up and avoid directly using `["pycodestyle", "pyflakes"]` instead of the correct `["pycodestyle" "pyflakes"]` or `'("pycodestyle" "pyflakes")`.

`["pycodestyle", "pyflakes"]` will cause exception in pyls, no actions will be released then. But, the other two will not be like that.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -52,7 +52,7 @@
   '(
     :plugins.jedi_completion.enabled t
     :plugins.jedi_definition.follow_imports t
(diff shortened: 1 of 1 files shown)


yyoncho:

You might be interested in https://github.com/emacs-lsp/lsp-mode/issues/715#issuecomment-473502818 (current default value is wrong).

PR #722, opened 2019-03-19
Use lambda in server paths

danielmartin:
As discussed in Gitter.

To make sure users can configure server paths per project (maybe via `.dir-locals.el` or
from a major mode hook), it may be better to construct the final server path at runtime via a lambda.


yyoncho:
Can you fix the cons calls - https://travis-ci.org/emacs-lsp/lsp-mode/jobs/508649050 ?

yyoncho:
Thank you!

PR #728, opened 2019-03-22
Separate symbol highlighting from eldoc

sebastiencs:
Rebased of #237


yyoncho:
Looks good! I was planning to do a similar thing. 

1. Can you introduce a defcustom to allow disabling highlighting?
2. Can you store and cancel the timer when a new request is performed to avoid additional calls?

sebastiencs:
@yyoncho Done :)

PR #732, opened 2019-03-25
Fix command for IPH server.

tigersoldier:


yyoncho:
Thank you, I must have been sleeping while doing this commit.

PR #734, opened 2019-03-26
Fix rendering markdown string

tigersoldier:
Insert string to temporary buffer before calling the major mode for the languge.
Some modes such as gfm-view-mode changes the temporary buffer to be read-only
and throws error if inserting after.


yyoncho:
Thank you for finding and fixing this! 

cc @Khady

PR #737, opened 2019-03-27
Add client for gopls.

seagle0128:
Pre-install: go get -u golang.org/x/tools/cmd/gopls.


yyoncho:
Looks good, can you add it to the readme page supported servers table?

yyoncho:
Thank you!

petergardfjall:
From what I understand this registers `gopls` as the default (it has higher priority than `bingo`).
Is this a deliberate choice? Isn't `bingo` still considered more stable/feature-rich?

yyoncho:
@petergardfjall yes, it will be the default. It is up to the golang users to determine which ls should be the default(change the definitions locally if you dislike the defaults).

seagle0128:
@petergardfjall right. But if you didn't install gopls, it wouldn't be enabled.

thomasf:
I always install update all of my x/tools... go commands in one step by going ` go get -u golang.org/x/tools/cmd/... ` so even if I have it installed I might not prefer to use it over bingo since it's not really working yet., 

Maybe the actual setting that already exist could be used to decide which one is chosen?

```el
(defcustom lsp-clients-go-server "bingo"
  "The go language server executable to use."
  :group 'lsp-clients-go
  :risky t
  :type 'file)
```

yyoncho:
@thomasf - it is possible but IMO it will be easier to distinguish what you are running if they are under different server ids. If you all agree that `gopls` is not ready to be the default we may change the priority until it has feature parity with bingo ls. I am not golang user so it is up to you to determine what is the most convenient `lsp-mode` configuration for it.

shuxiao9058:
gopls not work with my project(gomod) now.

yyoncho:
Just for reference, to unregster gopls: 
```
(remhash 'gopls lsp-clients)
```

yyoncho:
Can you all take a look at https://github.com/emacs-lsp/lsp-mode/pull/742 and comment/test?

seagle0128:
It looks good, except two minor issues.

1. The implementation is different from other clients. e.g. `typescript-language-server` and `javascript-typescript-stdio`. 
2. The server-id looks not good for `go-go-langserver`.

IMO, for consistency, should remove the configuration `sp-clients-go-server` and keep `lsp-clients-go-server-args`. Or, make other clients configurable as #742 , which is not as original design, right?

dsvensson:
@petergardfjall The author of bingo has started porting over bingo features to gopls. Not upstreamed yet, but available here, https://github.com/saibing/tools

petergardfjall:
@dsvensson That is good news, but I suppose gopls won't be up to par with bingo for some time. I've tried using gopls a couple of times but it hasn't been a good experience. I look forward to an official language server though!

seagle0128:
@petergardfjall yes, it's a really good news! But, if you don't want to use `gopls` for now, the simplest way is to delete or rename `gopls` binary. `bingo` will be used automatically.

thomasf:
> Just for reference, to unregster gopls:
> 
> ```
> (remhash 'gopls lsp-clients)
> ```

When am I supposed to do that,, I added that as an eval-after-load after lsp-mode but gopls is still chosen...

thomasf:
ok, I guess I'm going to try to add it after lsp-clients instead.

corpix:
> Just for reference, to unregster gopls:
> 
> ```
> (remhash 'gopls lsp-clients)
> ```

This is not working with `use-package` for me:

```lisp
(use-package lsp-mode
    :ensure t
    :config (remhash 'gopls lsp-clients)
            (add-hook 'go-mode-hook #'lsp))
```

Why make unstable things default? This "update" broke all development environment for me :(

yyoncho:
Add `(require 'lsp-clients)` before remhash.

thomasf:
Or like this if you want to use use-package for everything

```elisp
(use-package lsp-mode
  :ensure t
  :commands (lsp lsp-mode)
  :config
  (progn
    (use-package lsp-clients
      :defer
      :config
      (progn
        (remhash 'gopls lsp-clients)))))

```

PR #741, opened 2019-03-29
Migrate lsp-hack package to lsp-clients

jra3:
Summary: This simply moves the current lsp-hack package into lsp-clients as an out-of-the-box option

Hack is an evolution of PHP that was developed at Facebook and offers much better performance and enhanced language features like static typing. HHVM, the language runtime that runs hack, supports LSP natively.

Test Plan: Loaded a .php file in hack-mode, turned on lsp, got the magic


lsp-clients.el
```diff
@@ -702,6 +702,34 @@ finding the executable with `exec-path'."
...
+  (interactive)
+  nil)
+
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-stdio-connection (lambda () lsp-clients-hack-command))
+                  :major-modes '(hack-mode)
+                  :priority -1
+                  :server-id 'hack
+                  ;; ignore some unsupported messages from Nuclide
+                  :notification-handlers (lsp-ht ("telemetry/event" 'lsp-hack--ignore)
```
yyoncho: nit: there is a standard emacs function `ignore`.

jra3: Yes! I actually copied the code and the comment from there. I got confused since it was in the subr package which was not required in this package yet. You are right though that it is a standard library distributed with emacs. I'll update it to use ignore

yyoncho:

PR #742, opened 2019-04-01
Make lsp-clients-go-server configurable

nanjj:
Current logic ignored `lsp-clients-go-server` and fixed to `gopls`,
while `gopls` does not work all the time.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -242,7 +242,8 @@ particular FILE-NAME and MODE."
   :tag "Go language")
 
 (defcustom lsp-clients-go-server "bingo"
-  "The go language server executable to use."
+  "The go language server executable to use.  One of `bingo',
+`gopls', `go-langserver' to use. Defaults to `bingo'."
   :group 'lsp-clients-go
   :risky t
   :type 'file)
@@ -314,31 +315,13 @@ defaults to half of your CPU cores."
 
 
 (lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection "gopls")
-                  :major-modes '(go-mode)
-                  :priority 0
-                  :initialization-options 'lsp-clients-go--make-init-options
-                  :server-id 'gopls
-                  :library-folders-fn (lambda (_workspace)
-                                        lsp-clients-go-library-directories)))
-
-(lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection
                                    (lambda () (cons lsp-clients-go-server
                                                     lsp-clients-go-server-args)))
                   :major-modes '(go-mode)
                   :priority -1
                   :initialization-options 'lsp-clients-go--make-init-options
-                  :server-id 'go-bingo
-                  :library-folders-fn (lambda (_workspace)
-                                        lsp-clients-go-library-directories)))
-
-(lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection "go-langserver")
-                  :major-modes '(go-mode)
-                  :priority -2
-                  :initialization-options 'lsp-clients-go--make-init-options
-                  :server-id 'go-ls
+                  :server-id (intern lsp-clients-go-server)
                   :library-folders-fn (lambda (_workspace)
                                         lsp-clients-go-library-directories)))
 


yyoncho:
@seagle0128 

Moving the discussion into the pull request.

> 
> It looks good, except two issues.
> 
>     The implementation is different from other clients. e.g. typescript-language-server and javascript-typescript-stdio.

My understanding is that the 3 servers are fork of one and the same language servers and that they have the same configuration options, functions, and so on which is not the same as js-ts/ts servers AFAIK. 

>     The server-id looks not good for go-go-langserver.

I agree.

> 
> IMO, for consistency, should remove the configuration sp-clients-go-server and keep lsp-clients-go-server-args. Or, make other clients configurable as #742 , which is not as original design, right?

The priority is the user experience, we are not coupled to a particular design decision. One of the problems is that if you do:

``` elisp
(setq ... "bingo")
(require 'lsp-clients)
(setq ... "gopls")
```

The server still will be "bingo".

seagle0128:
> My understanding is that the 3 servers are fork of one and the same language servers and that they have the same configuration options, functions, and so on which is not the same as js-ts/ts servers AFAIK.

They have different configurations actually. To me they are same as js-ts/ts servers.

For example:

```shell
$ gopls --help
The Go Language source tools.

Usage: gopls [flags] <command> [command-flags] [command-args]

Available commands are:
  serve : run a server for Go code using the Language Server Protocol
  query : answer queries about go source code

gopls flags are:
  -listen string
    	address on which to listen for remote connections
  -logfile string
    	filename to log to. if value is "auto", then logging to a default output file is enabled
  -mode string
    	no effect
  -port int
    	port on which to run gopls for debugging purposes
  -profile.cpu string
    	write CPU profile to this file
  -profile.mem string
    	write memory profile to this file
  -profile.trace string
    	write trace log to this file
  -remote string
    	*EXPERIMENTAL* - forward all commands to a remote lsp
```

```shell
$ bingo -h
Usage of bingo:
  -addr string
    	server listen address (tcp) (default ":4389")
  -build-tags string
    	build tags, separated by spaces.
  -cache-style string
    	set global cache style: none, on-demand, always. Can be overridden by InitializationOptions. (default "always")
  -diagnostics
    	enable diagnostics. no actual effect, just for compatible with ide-go
  -diagnostics-style string
    	diagnostics style: none, instant, onsave. Can be overridden by InitializationOptions. (default "instant")
  -disable-func-snippet
    	disable argument snippets on func completion. Can be overridden by InitializationOptions.
  -enhance-signature-help
    	enhance signature help with return result. Can be overridden by InitializationOptions.
  -format-style string
    	which format style is used to format documents. Supported: gofmt and goimports. Can be overridden by InitializationOptions. (default "goimports")
  -format-tool string
    	which tool is used to format documents. no actual effect, just for compatible with ide-go (default "goimports")
  -freeosmemory int
    	the interval time that aggressively free memory back to the OS, unit is second, default value is 0, means no free memroy back to the OS
  -func-snippet-enabled
    	enable argument snippets on func completion. no actual effect, just for compatible with ide-go (default true)
  -gocodecompletion
    	enable completion. no actual effect, just for compatible with ide-go
  -goimports-prefix string
    	set '--local' flag for the goimports invocation. Can be overridden by InitializationOptions.
  -logfile string
    	also log to this file (in addition to stderr)
  -maxparallelism int
    	use at max N parallel goroutines to fulfill requests. Can be overridden by InitializationOptions.
  -mode string
    	communication mode (stdio|tcp) (default "stdio")
  -pprof string
    	start a pprof http server (https://golang.org/pkg/net/http/pprof/)
  -trace
    	print all requests and responses
  -version
    	print version and exit
```

seagle0128:
FYI
https://github.com/saibing/bingo/issues/13#issuecomment-478351293

@saibing who is the author of `bingo`, is starting developing `gopls`.

yyoncho:
> They have different configurations actually. To me they are same as js-ts/ts servers.

Ok, I was fooled by the fact that lsp-clients-go--make-init-options is used for the three language servers. 

If @nanjj concern is that gopls is not ready to be the default language server we might switch the priorities temporary.

fjl:
It's also good to allow overriding the path regardless of whether it'll always be one server or the other.

yyoncho:
Closing in favour of https://github.com/emacs-lsp/lsp-mode/pull/771 .

PR #743, opened 2019-04-02
Fix issue that the role of the `-client-activation-fn` is offset in `…

twlz0ne:
…lsp--find-clients`

If the `-client-activation-fn` exists, then should not check the `-major-modes`.

Flowing is a minimal testing configuration:
https://gist.github.com/twlz0ne/a34bee2f5b2b62772ebe42756a1fb469


yyoncho:
Thank you. We may issue a warning in register client to make that more evident.

Issue #747, opened 2019-04-05
workspace/configuration response missing an ID


stamblerre:
I'm one of the developers working on the Go language server, [gopls](https://golang.org/x/tools/cmd/gopls), and it seems that one of our recent changes has exposed a bug in the `workspace/configuration` message.

We first noticed this bug because of the reports here: https://github.com/golang/go/issues/31178.

It seems that lsp-mode's response to the `workspace/configuration` request from the server doesn't actually have a message ID associated with it, so we ignore the response.

I've been able to reproduce this bug by using gopls with these settings:

```
(require 'lsp-mode)
(add-hook 'go-mode-hook #'lsp)
```

cc @ianthehat

hbar-x-2pi:
The response to `workspace/configuration` request should be a valid `JSON RPC` message.

```diff
diff --git a/lsp-mode.el b/lsp-mode.el
index 9dc731d..5bf6b5a 100644
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3714,8 +3714,7 @@ WORKSPACE is the active workspace."
                        (lsp--apply-workspace-edit (gethash "edit" params))
                        empty-response)
                       ("workspace/configuration"
-                       (or (lsp--create-initialization-options (lsp-session) client)
-                           empty-response))
+                       (lsp--make-response request (lsp--create-initialization-options (lsp-session) client)))
                       (other
                        (-if-let (handler (gethash other (lsp--client-request-handlers client) nil))
                            (lsp--make-response request (funcall handler workspace params))
```

yyoncho:
@hbar-x-2pi thanks, would you mind providing a PR?

yyoncho:
Fixed with https://github.com/emacs-lsp/lsp-mode/commit/7c7e40bdb2529b79b75356ca93b1a2f714b7e58c by @hbar-x-2pi

stamblerre:
Thanks all for the quick resolution!

PR #749, opened 2019-04-05
Fix response to `workspace/configuration` (#747)

hbar-x-2pi:
The response to `workspace/configuration` request should be a valid `JSON RPC` message.


yyoncho:

hbar-x-2pi:
Welcome and thanks for merging it so quickly!

PR #754, opened 2019-04-08
lsp-mode: avoid passing server error message to elisp format

fjl:
This fixes a bug where server error messages containing '%' would be
interpreted by the emacs format function.


yyoncho:
Thank you!

PR #756, opened 2019-04-09
Unify behavior of lsp-format-buffer and lsp-format-region (#755).

### LINKED ISSUE #755: lsp-format-region jumps to the beginning of the buffer
This bug is similar to #253 (fixed by #258), but for lsp-format-region, invoked by M-x indent-region with some servers, for example pyls in python-mode.

After running indent-region the region is reformatted, but point moves to the beginning of the buffer. It would be nice if point would stay as close as possible to the original position.

(In addition, sometimes lsp-format-region formats one more line above and/or below the marked section, worth it's own bug report I suppose).

bastianbeischer:
This fixes #755 and makes lsp-format-region interactive, which allows it to be used independently of indent-region.


yyoncho:
And congrats for your first PR! :tada:

(I do not get notifications for force pushes so it is good idea to ping for rereview)

PR #759, opened 2019-04-10
Add lsp-organize-imports

gexplorer:
Adds support for **organize-imports** command

### DIFF
--- a/README.org
+++ b/README.org
@@ -140,6 +140,7 @@
    - ~lsp-goto-implementation~ - Go to implementation
    - ~lsp-goto-type-definition~ - Go to type definition
    - ~lsp-rename~ - Rename symbol at point
+   - ~lsp-organize-imports~ - Organize imports
    - ~lsp-restart-workspace~ - Restart project
    - ~lsp-symbol-highlight~ - Highlight all relevant references to the symbol under point.
    - ~lsp-workspace-folders-add~ - Add workspace folder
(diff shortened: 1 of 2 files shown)


yyoncho:
Thanks for the PR! I am unable to find information for this method in lsp spec so I suspect that it might be an extension method. In case it is extension method for a particular server it should be prefixed like lsp-LANGUAGE-organize-imports and moved to the proper file. Can you tell me which language server did you use to test the method or point me to the lsp spec documenting in case I have missed it?

gexplorer:
I did it for the [typescript server](https://github.com/sourcegraph/javascript-typescript-langserver/), based on the same command implementation in tide. I had to add the command to the server which has the pull request pending.

Anyway found the definition of the command in this [specification](https://microsoft.github.io/language-server-protocol/specification), isn't it the official one?

yyoncho:
Yes, it is the official one, I was unable to find `codeActionsProvider`. There is `source.organizeImports` in the spec but I think that it should be implemented through [workspace/executeCommand](https://microsoft.github.io/language-server-protocol/specification#workspace_executeCommand)  not by introducing a new method. I am not sure how the server announces that it supports particular CodeActionKind though but I am going to check jdt.ls implementation. I will hold merging this PR untill we sort this out.

fjl:
I need this for Go editing as well.

yyoncho:
@fjl

This method should be moved in lsp-mode from lsp-java 
https://github.com/emacs-lsp/lsp-java/blob/master/lsp-java.el#L982

If go server supports organize imports it will work.

yyoncho:
Closing in favour of https://github.com/emacs-lsp/lsp-mode/pull/810 .

PR #765, opened 2019-04-14
Force Imenu rescan when `imenu-create-index-function' is updated

d125q:
If `imenu--make-index-alist` is somehow invoked _before_ `lsp-enable-imenu` (either via `imenu` or if there is a menu-bar entry) and `imenu-auto-rescan` is nil, then Imenu will use the "pre-LSP" alist until a rescan occurs.  This PR makes the rescan occur ASAP.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4185,7 +4185,13 @@ Return a nested alist keyed by symbol names. e.g.
 
 (defun lsp-enable-imenu ()
   "Use lsp-imenu for the current buffer."
-  (setq-local imenu-create-index-function #'lsp--imenu-create-index))
+  (setq-local imenu-create-index-function #'lsp--imenu-create-index)
+  (imenu--cleanup)
+  ;; Make sure imenu-update-menubar redoes everything.
+  (setq imenu-menubar-modified-tick -1)
+  (setq imenu--index-alist nil)
+  (setq imenu--last-menubar-index-alist nil)
+  (imenu-update-menubar))
 
 (defun lsp-resolve-final-function (command)
   "Resolve final function COMMAND."


lsp-mode.el
```diff
@@ -4185,7 +4185,13 @@ Return a nested alist keyed by symbol names. e.g.
...
 (defun lsp-enable-imenu ()
   "Use lsp-imenu for the current buffer."
-  (setq-local imenu-create-index-function #'lsp--imenu-create-index))
+  (setq-local imenu-create-index-function #'lsp--imenu-create-index)
+  (imenu--cleanup)
+  ;; Make sure imenu-update-menubar redoes everything.
+  (setq imenu-menubar-modified-tick -1)
+  (setq imenu--index-alist nil)
+  (setq imenu--last-menubar-index-alist nil)
+  (imenu-update-menubar))
```
yyoncho: I am not very familiar with the imenu related code: is this needed when you do not have menubar enabled? If not, I would prefer skipping this call and let imenu update the cache when imenu is called explicitly.

d125q: Thank you for the prompt reply!

The key thing to note is that there is a delay in the reporting od the client's Imenu capability.  Therefore, `imenu-create-index-function` is not set "early enough," so to say.

Since Imenu updatea its alist under certain conditions, the current code can result in Imenu using a "pre-LSP" alist.

This happens in two scenarios:

1. Imenu has a menu-bar entry, which will compute the alist very likely before LSP does anything since it runs on `menu-bar-update-hook`, or
2. The user forces the alist to be computed by invoking `imenu` while `lsp-mode` is waiting for capability response.

This basically clears any alist that may have been computed before the LSP client indicated its Imenu capability.  Lastly, `imenu-update-menubar` does nothing if there is no menu-bar entry for Imenu.

yyoncho:
> The key thing to note is that there is a delay in the reporting of the servers' Imenu capability. Therefore, imenu-create-index-function is not set "early enough," so to say.

Yeah, I got that. We set it right after the server is initialized and we know that the server has that capability. If we set it earlier the code will throw.

> Lastly, imenu-update-menubar does nothing if there is no menu-bar entry for Imenu.

That's good - I will do some manual testing tomorrow and will proceed with the merge.

yyoncho:
Closing in favour of https://github.com/emacs-lsp/lsp-mode/pull/780

PR #771, opened 2019-04-17
Add some gopls configuration settings

muirrn:
- add "usePlaceholders" config value which causes gopls to give full
  function parameter snippets rather than just "()" snippet
- add lsp-gopls-server-args for adding more cli args. the most common
  is '("-logfile" "/tmp/gopls.log" "-rpc.trace") to make it log out
  all rpc activity. i didn't reuse the existing
  lsp-clients-go-server-args since gopls doesn't accept the same
  arguments as the previous go lsp servers.
- don't use the other go initialization-options with gopls since gopls
  doesn't support those
- move go client stuff into separate lsp-go package

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -49,3 +49,4 @@ Christian Kellner
 benjaminor
 Amos Bird
 AmaiKinono
+Muir Manders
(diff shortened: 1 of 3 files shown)


yyoncho:

PR #776, opened 2019-04-19
support rustic-mode to connect rls.

lurdan:
Using rustic-mode, lsp-mode doesn't set languageId. It makes rls panic.


yyoncho:
Thank you!

PR #779, opened 2019-04-21
render first line if not all

scturtle:
Specifically to make rust-analyzer more usable.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3247,6 +3247,24 @@ When language is nil render as markup if `markdown-mode' is loaded."
    ((stringp content) (lsp--render-string content "markdown"))
    (t (error "Failed to handle %s" content))))
 
+(defun lsp--render-simple (content)
+  "Render first line of code block in CONTENT element.
+Fallback to lsp--render-element if not found."
+  (let* ((value (if (hash-table-p content) (gethash "value" content) content))
+         (lang (if (hash-table-p content) (gethash "language" content) "markdown"))
+         (lines (split-string value "\n"))
+         (start (--drop-while (not (string-prefix-p "```" it)) lines))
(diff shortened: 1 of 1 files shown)


yyoncho:
Can you comment whether reverting the change(c1df55e) I reverted due to unknown lsp-ui issue works fine with rust analyzer? I am a bit concerned our PR is too custom and it will potentially break due to minor mismatches(see below).  If revering c1df55e is not fine for you, I would prefer to create an extension point (e. g. lsp-render-function) which will be used to customize the hover rendering. This part of the code has been changing like from the start of the project due to the fact that onHover info is not designed to work with eldoc. 

I tested it with a couple of servers and it does not work fine with at least one server. 

```
<<<< ruby-ls:8927
{
  "jsonrpc": "2.0",
  "id": 110,
  "result": {
    "contents": {
      "kind": "markdown",
      "value": "[#ordinal](solargraph:/document?query=%23ordinal)\n\n(a, b, c)"
    }
  }
}
```

scturtle:
I think reverting c1df55e is the same as turning on `lsp-eldoc-render-all` which is too much for most of the time. I will try the custom function way (with my little elisp knowledge). Is another option like `lsp-eldoc-render-first-line` acceptable to provide the proposed heuristic?

yyoncho:
> Is another option like lsp-eldoc-render-first-line acceptable to provide the proposed heuristic?

I guess it is fine, as long as we make the extract logic a bit more robust and fallback to the full text when the input format does not match our expectations.

(I will use ' instead of \`)

I think that if the logic is to extract the first code block(the content between ''' and ''') and then use render markdown function will work in more cases. E. g.

> '''
> without language
> '''

> ''' single line '''

scturtle:
Not quite useful.

PR #780, opened 2019-04-21
Async Imenu menu-bar

d125q:
Only vanilla `imenu` has been tested, and not `helm-imenu` or similar.  I tested all four combinations:

1. ```elisp
   (menu-bar-mode +1)
   (imenu-add-menubar-index)
   (setq imenu-auto-rescan nil)
    ```
2. ```elisp
   (menu-bar-mode +1)
   (imenu-add-menubar-index)
   (setq imenu-auto-rescan t)
    ```
3. ```elisp
   (menu-bar-mode -1)
   (setq imenu-auto-rescan nil)
    ```
4. ```elisp
   (menu-bar-mode -1)
   (setq imenu-auto-rescan t)
    ```

When `imenu-auto-rescan` is non-nil, both interactive results and menu-bar entry are always up-to-date; however, the menu-bar entry is updated asynchronously.  When it is nil, forcing a `*Rescan*` either through interactive use or the menu-bar entry brings everything up to date.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3606,10 +3606,9 @@ perform the request synchronously."
       lsp--document-symbols
     (let ((method "textDocument/documentSymbol")
 	  (params `(:textDocument ,(lsp--text-document-identifier))))
+      (setq lsp--document-symbols-tick (buffer-chars-modified-tick))
       (if (not lsp--document-symbols-request-async)
-	  (setq lsp--document-symbols-tick (buffer-chars-modified-tick)
-		lsp--document-symbols (lsp-request method params))
-	(setq lsp--document-symbols-tick (buffer-chars-modified-tick))
+	  (setq lsp--document-symbols (lsp-request method params))
 	(lsp-request-async method params
 			   (lambda (document-symbols)
 			     (setq lsp--document-symbols document-symbols)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3604,19 +3604,18 @@ Else (e.g., due to intereactive use of `imenu' or `xref'),
 perform the request synchronously."
   (if (= (buffer-chars-modified-tick) lsp--document-symbols-tick)
       lsp--document-symbols
-    (if (not lsp--document-symbols-request-async)
-	(setq lsp--document-symbols-tick (buffer-chars-modified-tick)
-	      lsp--document-symbols (lsp-request
-				     "textDocument/documentSymbol"
-				     `(:textDocument ,(lsp--text-document-identifier))))
-      (lsp-request-async "textDocument/documentSymbol"
-			 `(:textDocument ,(lsp--text-document-identifier)
-					 :mode alive)
-			 (lambda (document-symbols)
-			   (setq lsp--document-symbols-tick (buffer-chars-modified-tick)
-				 lsp--document-symbols document-symbols)
-			   (lsp--imenu-refresh)))
-      lsp--document-symbols)))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3581,8 +3591,37 @@ A reference is highlighted only if it is visible in a window."
...
+  (if (= (buffer-chars-modified-tick) lsp--document-symbols-tick)
+      lsp--document-symbols
+    (if (not lsp--document-symbols-request-async)
+	(setq lsp--document-symbols-tick (buffer-chars-modified-tick)
+	      lsp--document-symbols (lsp-request
+				     "textDocument/documentSymbol"
+				     `(:textDocument ,(lsp--text-document-identifier))))
+      (lsp-request-async "textDocument/documentSymbol"
+			 `(:textDocument ,(lsp--text-document-identifier)
+					 :mode alive)
```
yyoncho: This mode should be param of lsp-request-async?

lsp-mode.el
```diff
@@ -3581,8 +3591,37 @@ A reference is highlighted only if it is visible in a window."
...
+			 (lambda (document-symbols)
+			   (setq lsp--document-symbols-tick (buffer-chars-modified-tick)
```
yyoncho: You should capture  (buffers-chars-modified-tick) before that since it could have changed after async call.

lsp-mode.el
```diff
@@ -3581,8 +3591,37 @@ A reference is highlighted only if it is visible in a window."
...
+If the buffer has not been modified since symbols were last
+retrieved, simply return the latest result.
+
+Else, if the request was initiated by Imenu updating its menu-bar
+entry, perform it asynchronously; i.e., give Imenu the latest
+result and then force a refresh when a new one is available.
+
+Else (e.g., due to intereactive use of `imenu' or `xref'),
+perform the request synchronously."
+  (if (= (buffer-chars-modified-tick) lsp--document-symbols-tick)
```
yyoncho: You may consider using cond

d125q: I would say it's good the way it is now.  Let me know if you have any further suggestions.

lsp-mode.el
```diff
@@ -3581,8 +3591,36 @@ A reference is highlighted only if it is visible in a window."
...
+      lsp--document-symbols
+    (let ((method "textDocument/documentSymbol")
+	  (params `(:textDocument ,(lsp--text-document-identifier))))
+      (if (not lsp--document-symbols-request-async)
+	  (setq lsp--document-symbols-tick (buffer-chars-modified-tick)
+		lsp--document-symbols (lsp-request method params))
+	(setq lsp--document-symbols-tick (buffer-chars-modified-tick))
```
yyoncho: It should be like:
```
(let ( (tick (buffe-modi...)
         (buff (current-buffer))
  (.... 
     (lambda (result) 
         (with-current-buffer buff
               (setq lsp--document-symbol-tick tick)
                ....))
```
otherwise, the stale results might be used as a real one.

d125q: But doesn't `:mode 'alive` already capture the buffer?

d125q: Scratch that.  I see what you mean.

yyoncho: Actually, you are right, buffer part is not needed, only the tick part.

yyoncho:
Thank you and congratulations on your first `lsp-mode` PR! :tada:

PR #783, opened 2019-04-22
Fix symlink handling in project root recognition

aureumleonis:
Current file-to-project folder recognition doesn't work when the project is located within a symlinked directory

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2864,7 +2864,7 @@ Applies on type formatting."
 
 (defun lsp-workspace-root (&optional path)
   "Find the workspace root for the current file or PATH."
-  (when-let (file-name (or path (buffer-file-name)))
+  (when-let (file-name (f-canonical (or path (buffer-file-name))))
     (->> (lsp-session)
          (lsp-session-folders)
          (--first (f-ancestor-of? it file-name)))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -2864,7 +2864,7 @@ Applies on type formatting."
 
 (defun lsp-workspace-root (&optional path)
   "Find the workspace root for the current file or PATH."
-  (when-let (file-name (or path (buffer-file-name)))
+  (when-let (file-name (f-canonical (or path (buffer-file-name))))
```
yyoncho: f-canonical will throw if called with nil

aureumleonis: I'll update to handle the nil case

aureumleonis: Just checked and f-ancestor-of? throws when file-name is nil so this change doesn't change ny behavior here

yyoncho:

kelvie:
It looks like this PR: https://github.com/emacs-lsp/lsp-mode/pull/1696 broke this fix.

the new `lsp-f-canonical` does *not* resolve symlinks.

ericdallo:
@yyoncho

kiennq:
I wonder what's expectation for this. If a project is located inside a sym-linked folder, shouldn't it be treated as new path (not resolve symlinks)?

PR #788, opened 2019-04-23
Extend JS/TS LS to work with any web-mode derived mode

d4ncer:
This PR checks if the passed `mode` is derived from `web-mode` rather than `web-mode` itself.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -101,7 +101,7 @@ finding the executable with variable `exec-path'."
   "Check if the javascript-typescript language server should be enabled
 based on FILE-NAME and MAJOR-MODE"
   (or (member mode '(typescript-mode typescript-tsx-mode js-mode js-jsx-mode js2-mode js2-jsx-mode rjsx-mode))
-      (and (eq major-mode 'web-mode)
+      (and (provided-mode-derived-p mode 'web-mode)
            (or (string-suffix-p ".tsx" filename t)
                (string-suffix-p ".jsx" filename t)))))
 
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -97,12 +97,9 @@ finding the executable with variable `exec-path'."
   :risky t
   :type '(repeat string))
 
-(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename _)
+(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
-  (or (string-suffix-p ".tsx" filename t)
-      (string-suffix-p ".jsx" filename t)
-      (string-suffix-p ".js" filename t)
-      (string-suffix-p ".ts" filename t)))
+  (not (null (string-match-p (rx (one-or-more char) "." (or "ts" "js") (opt "x") string-end) filename))))
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -99,7 +99,7 @@ finding the executable with variable `exec-path'."
 
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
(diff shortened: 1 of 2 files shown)


d4ncer:
I could also extend the checks on the other branch of this fn: 

```elisp
(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename mode)
  "Check if the javascript-typescript language server should be enabled
based on FILE-NAME and MAJOR-MODE"
  (or (provided-mode-derived-p mode 'typescript-mode 'typescript-tsx-mode 'js-mode 'js-jsx-mode 'js2-mode 'js2-jsx-mode 'rjsx-mode)
      (and (provided-mode-derived-p mode 'web-mode)
           (or (string-suffix-p ".tsx" filename t)
               (string-suffix-p ".jsx" filename t)))))
```

d4ncer:
Ah sorry, I just realised `provided-mode-derived-p` is only available 26.1+. I can introduce a helper fn with the same definition to be used in case the function isn't available?

yyoncho:
> I can introduce a helper fn with the same implementation to be used in case the function isn't available?

Sound good, just add a comment that we should delete the function once we drop support for 25.x

yyoncho:
An alternative approach is to delete the major mode match and keep only the file suffix. VScode is matching the language server only by file extension so should be good too.

d4ncer:
Went with the file suffix check only. The implementation feels a bit naïve to me, any refactoring you would suggest?

yyoncho:
> Went with the file suffix check only. The implementation feels a bit naïve to me, any refactoring you would suggest?

Two options:

1. Use regexp for files, e. g. ".js$\\|.ts$"
2. Introduce a function which accepts filename + list of regexps which we can later reuse for the other configuration.

d4ncer:
👍 I've gone with the regexp approach and I've added some tests too (first time ever writing tests, so let me know if they're not right!). I went with an `rx` based regex for readability, but I can use the raw form if you'd prefer.

d4ncer:
Side note: looks like there is already a fn that takes a list of regexes and a string and returns if any match - `lsp--string-match-any`.

test/lsp-clients-test.el
```diff
@@ -75,4 +75,12 @@
   (let ((major-mode 'js-mode))
     (should (not (lsp-clients-flow-activate-p (concat test-location "fixtures/SampleTypeScriptProject/src/sample.ts") nil)))))
 
+(ert-deftest lsp-typescript-javascript-activates-based-on-file-extension ()
+  (should (lsp-typescript-javascript-tsx-jsx-activate-p "abc.js"))
```
yyoncho: can you add the following tests `a1.ts` and `a1.d.ts`?

lsp-clients.el
```diff
@@ -97,13 +97,9 @@ finding the executable with variable `exec-path'."
...
-(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename mode)
-  "Check if the javascript-typescript language server should be enabled
-based on FILE-NAME and MAJOR-MODE"
-  (or (member mode '(typescript-mode typescript-tsx-mode js-mode js-jsx-mode js2-mode js2-jsx-mode rjsx-mode))
-      (and (eq major-mode 'web-mode)
-           (or (string-suffix-p ".tsx" filename t)
-               (string-suffix-p ".jsx" filename t)))))
+(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
+  "Check if the javascript-typescript language server should be enabled based on FILENAME."
+  (not (null (string-match-p (rx (one-or-more char) "." (or "ts" "js") (opt "x") string-end) filename))))
```
yyoncho: I think `(not (null ` is not needed.

d4ncer: Cool, will remove it. I was following the implementation of `s-matches?` from `s.el`. The only thing to note is that `s-matches-p` does not return `t`/`nil`, but rather the index (if found) / `nil`.

test/lsp-clients-test.el
d4ncer: Done

lsp-clients.el
yyoncho: I guess it is a convention of the library to return t/nil for each predicate function (there is such convention in clojure) although they do not follow it everywhere.

yyoncho:

PR #791, opened 2019-04-24
Add support for Clojure/ClojureScript via `clojure-lsp`

benedekfazekas:
Additionally enable jumping into jarred dependencies if the server is
configured to return the URI in the right format


lsp-clojure.el
```diff
@@ -0,0 +1,55 @@
...
+        (insert content)
+        (set-visited-file-name name)
+        (setq-local buffer-read-only t)
+        (set-buffer-modified-p nil)
+        (set-auto-mode)
+        (current-buffer)))
+    name))
+
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-stdio-connection '("bash" "-c" "clojure-lsp"))
```
yyoncho: Can you extract the command into defcustom variable for windows users? Even better will be if you can provide default value that will work for windows as well.

benedekfazekas: good shout, looking into it

yyoncho:
I did some testing and it looks promising, there is functionality that is not currently present in cider(e. g. finding references) and the fact that it performs static analysis is nice too. There are several extension methods what we might want to implement (introducing locals, inlining locals, etc). You may 

On the bad side it seems like it is not supported very well by company-lsp (cc @tigersoldier). The server requires client-side filtering but company-lsp fails most of the time probably due to incorrect prefix calculation.

benedekfazekas:
yeah indeed. finding references is there in `clojure-emacs/refactor-nrepl` but only available for clj (not cljs), see details in this thread https://github.com/clojure-emacs/refactor-nrepl/issues/195 `clojure-emacs` team member, clj-refactor/refactor-nrepl maintainer here btw ;)

also it seems that `rewrite-clj` based static analyzers are on the rise nowadays also see https://github.com/borkdude/clj-kondo

next step would be to look at integrating the refactorings `clojure-lsp` offers

benedekfazekas:
Added the `defcustom` for the command. It seems that there is an issue tracking the state of windows support https://github.com/snoe/clojure-lsp/issues/25 @snoe, @sparkofreason can you please chip in what would the start server command for windows?

yyoncho:

I will merge it as it is and we could work on exposing the extension methods/windows support and fixing company-lsp. 


> yeah indeed. finding references is there in clojure-emacs/refactor-nrepl but only available for clj (not cljs), see details in this thread clojure-emacs/refactor-nrepl#195 clojure-emacs team member, clj-refactor/refactor-nrepl maintainer here btw ;)

I am actually clojure/java programmer in real life so thank you for your work on refactor-nrep!. As a side note, is there an issue tracking the effort on making refactor-nrepl work with the latest cider?

PS: By the way, this is not the only possibility for collaboration between emacs-lsp and clojure-emacs. I think that even at this point, dap-mode supports debugging clojurescript and also it won't be hard to be adjusted to work for clojure.

benedekfazekas:
thanks for the merge! I am using latest clj-refactor, refactor-nrepl 2.4.0 with latest cider. feel free to file an issue if you have a problem with latest versions or maybe you have a slightly older one?

yyoncho:
I had some issues in the past and I am checking clj-refactor readme compatibility table from time to time - https://github.com/clojure-emacs/clj-refactor.el . It still says that it supports only cider 0.17/0.18. Is this info out of date?

benedekfazekas:
i suppose it is, my versions
```
CIDER 0.21.0 (New York)
clj-refactor 2.4.0 (package: 20190117.1235), refactor-nrepl 2.4.0
```

yyoncho:
Thank you, I will try it when I am back at work.

benedekfazekas:
let me know if you have any probs, ta!

PR #793, opened 2019-04-25
Fix content-length parsing for case without space after Content-Length:

zichun:
Certain LSP server doesn't have space in between Content-Length: header and the numeric value after (Content-Length:42), e.g cppextensions for vscode (https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools).

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3916,7 +3916,7 @@ WORKSPACE is the active workspace."
       ;; screwed up with a previous Content-Length
       (error "No Content-Length header"))))
 
-(defun lsp--trim-left (s)
+(defun s-trim-left (s)
   "Remove whitespace at the beginning of S."
   (if (string-match "\\`[ \t\n\r]+" s)
       (replace-match "" t t s)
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -3916,14 +3916,20 @@ WORKSPACE is the active workspace."
       ;; screwed up with a previous Content-Length
       (error "No Content-Length header"))))
 
+(defun lsp--trim-left (s)
```
yyoncho: why not using `s-trim-left`, we depend on `s`?

jokem59: still haven't got cpptools to work.  we can launch the server, but when we run `lsp-describe-thing-at-point`, it shows the server stuck in "Initializing...".  Hoping we can get it working!

yyoncho: I had a question about it a couple of weeks ago and when I searched in the cpptools repo the maintainers were basically saying that it will work only with vscode and to try ccls/cquery/clangd. If you need help you may ping me in it gitter.

lsp-mode.el
```diff
@@ -3916,14 +3916,20 @@ WORKSPACE is the active workspace."
       ;; screwed up with a previous Content-Length
       (error "No Content-Length header"))))
 
+(defun s-trim-left (s)
```
yyoncho: There is already such function(in `s` package), you could delete it.

zichun: oh oops doh. removed

yyoncho:

And congratulations you are our :100:  contributor! :tada:

PR #795, opened 2019-04-26
Add config option to disable pylint when using lsp-pyls.

sakshamsharma:
It appears this plugin's configuration wasn't added to lsp-mode.
[pylint_lint.py](https://github.com/palantir/python-language-server/blob/18f3c3f06f36b5e5ff3dae1436e85ef3eb5d7264/pyls/plugins/pylint_lint.py)

With this patch in my lsp-mode checkout, if I change the invocation to `pyls -vvv`, I can see pylint in the `Disabled checkers` log statement in the `*pyls::stderr*` buffer. Also, don't see the pylint errors anymore.

### DIFF
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -103,6 +103,10 @@ complexity."
   "List of modules to import on startup"
   :type '(repeat string))
 
+(defcustom lsp-pyls-plugins-pylint-enabled t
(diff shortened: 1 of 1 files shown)


yyoncho:

sakshamsharma:
Wow that was quick :smile:

PR #798, opened 2019-04-27
Prevent recursive file watches due to symlinks.

### LINKED ISSUE #796: Recursive file watches in presence of recursive symlinks
**Describe the bug**
lsp-mode fails to detect symlinks in a workdir that point to the same directory as before, and continues creating a huge watchlist of files.

Specifically this function:
```
(defun lsp-watch-root-folder (dir callback &optional watch)
  "Create recursive file notificaton watch in DIR.
CALLBACK is the will be called when there are changes in any of
the monitored files. WATCHES is a hash table directory->file
notification handle which contains all of the watch that
already have been created."
  (lsp-log "Creating watch for %s" dir)
```

I can't count how deep it traversed the symlink in my case, but I'm guessing it stopped when the file name length got hit (65k or so characters). I can see entries like:
```
/...../dir1/dir1/dir1/dir1/dir1/file1
/...../dir1/dir1/dir1/dir1/file1
/...../dir1/dir1/dir1/file1
/...../dir1/dir1/file1
.... so on ....
```

**Expected behavior**
It should keep a running stack of directories, and only visit symlinks that aren't pointing to a directory that has already been visited. Also, perhaps symlinks should be visited last in the DFS for files, after all directories have been visited in that directory (not sure if this will be very consequential or not)

pyls / ccls both

**OS**
Ubuntu

sakshamsharma:
Fixes #796

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1019,7 +1019,7 @@ already have been created."
     (condition-case err
         (progn
           (puthash
-           dir
+           (file-truename dir)
            (file-notify-add-watch
             dir
             '(change)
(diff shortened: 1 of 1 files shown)


yyoncho:

PR #799, opened 2019-04-27
Allow filtering clients using lsp-enabled-clients optionally.

sakshamsharma:
Setting this variable will allow users to only enable lsp clients
selectively, and modify what clients can be used without having to
restart emacs with a different set of registered clients.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -227,6 +227,10 @@ When set to t `lsp-mode' will auto-configure `lsp-ui' and `company-lsp'."
   "Hash table server-id -> client.
 It contains all of the clients that are currently registered.")
 
+(defvar lsp-enabled-clients nil
+  "List of clients allowed to be used for projects.
+When nil, all registered clients are considered candidates.")
+
 (defvar lsp-last-id 0
   "Last request id.")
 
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -227,6 +227,10 @@ When set to t `lsp-mode' will auto-configure `lsp-ui' and `company-lsp'."
   "Hash table server-id -> client.
 It contains all of the clients that are currently registered.")
 
+(defvar lsp-enabled-clients (list)
```
yyoncho: nit: replace (list) with nil.

lsp-mode.el
```diff
@@ -4587,11 +4591,13 @@ remote machine and vice versa."
     (-when-let (matching-clients (->> lsp-clients
                                       hash-table-values
                                       (-filter (-lambda (client)
-                                                 (and (or (-some-> client lsp--client-activation-fn (funcall buffer-file-name buffer-major-mode))
-                                                          (and (not (lsp--client-activation-fn client))
-                                                               (-contains? (lsp--client-major-modes client) buffer-major-mode)
-                                                               (eq (---truthy? remote?) (---truthy? (lsp--client-remote? client)))))
-                                                      (-some-> client lsp--client-new-connection (plist-get :test?) funcall))))))
+                                                 (and (or (eq lsp-enabled-clients nil)
```
yyoncho: nit: use `(null lsp-enabled-clients)

yyoncho: can you move this as a second condition and add logging that the client is not enabled? E. g.
1. Check whether it is present, matches mode, etc.
2. Check whether it is enabled
3. If it is matching and not enabled print a message similar to what we do for blacklisted dirs
https://github.com/emacs-lsp/lsp-mode/blob/42b1e5c1f82e56734777529fadba9291e8a9dc15/lsp-mode.el#L4906

yyoncho:
Thank you!

PR #803, opened 2019-04-28
Make method public

danielmartin:
This is used from `lsp-ui` package so it should be public (https://github.com/emacs-lsp/lsp-ui/pull/277)


yyoncho:
Thank you!

PR #807, opened 2019-04-29
[WIP] Show workspace diagnostics in the mode line

danielmartin:
Adds a new minor mode to show workspace diagnostics in the Emacs mode line.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1131,6 +1131,25 @@ PARAMS - the data sent from WORKSPACE."
                  (lsp--workspace-diagnostics it))))
     result))
 
+(defun lsp-diagnostics-by-severity (file-diagnostics)
+  "Return diagnostics from FILE-DIAGNOSTICS, grouped by
+  diagnostic severity."
+  (-group-by 'lsp-diagnostic-severity file-diagnostics))
+
+(defun lsp-calculate-diagnostic-statistics (file-diagnostics)
+  "Calculate FILE-DIAGNOSTICS statistics in format
+  'error/warning/information/hint'."
+  (string-join
+   (-map (-lambda ((severity . diagnostics))
+           (number-to-string (length diagnostics)))
+         (lsp-diagnostics-by-severity file-diagnostics))
+   "/"))
+
+(defun lsp--calculate-workspaces-diag-statistics ()
+  "Calculate diagnostic statistics for all workspaces in a session."
+  (let ((workspace-diagnostics (-flatten (ht-values (lsp-diagnostics)))))
+    (lsp-calculate-diagnostic-statistics workspace-diagnostics)))
+
 (cl-defstruct lsp-diagnostic
   (range nil :read-only t)
   ;; range has the form (:start (:line N :column N) :end (:line N :column N) )
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -67,6 +67,12 @@
     (3 . ,compilation-message-face)
     (4 . ,compilation-info-face)))
 
+(defconst lsp--message-type-help-echo
+  `((1 . "Number of errors in open LSP workspaces.")
+    (2 . "Number of warnings in open LSP workspaces.")
+    (3 . "Number of messages in open LSP workspaces.")
+    (4 . "Number of information in open LSP workspaces.")))
+
 (defconst lsp--errors
   '((-32700 "Parse Error")
     (-32600 "Invalid Request")
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -568,6 +568,12 @@ must be used for handling a particular message.")
   :type 'number
   :group 'lsp-mode)
 
+(defcustom lsp-show-mode-line-workspace-diagnostics t
+  "Whether to show a summary of diagnostics for all open
+  workspaces in the mode line."
+  :type 'boolean
+  :group 'lsp-mode)
+
 (defvar lsp-custom-markup-modes
   '((rust-mode "no_run" "rust,no_run" "rust,ignore" "rust,should_panic"))
   "Mode to uses with markdown code blocks.
@@ -1141,7 +1147,8 @@ PARAMS - the data sent from WORKSPACE."
   'error/warning/information/hint'."
   (string-join
    (-map (-lambda ((severity . diagnostics))
-           (number-to-string (length diagnostics)))
+           (propertize (f-filename (number-to-string (length diagnostics)))
+                       'face (cl-rest (assoc severity lsp--message-type-face))))
          (lsp-diagnostics-by-severity file-diagnostics))
    "/"))
 
@@ -1149,6 +1156,13 @@ PARAMS - the data sent from WORKSPACE."
   "Calculate diagnostic statistics for all workspaces in a session."
   (let ((workspace-diagnostics (-flatten (ht-values (lsp-diagnostics)))))
     (lsp-calculate-diagnostic-statistics workspace-diagnostics)))
+
+(defun lsp--update-mode-line-workspaces-diag-statistics ()
+  "Update the workspace diagnostic statistics and sets the
+summary information in
+`lsp--workspace-diagnostics-mode-line-string' so that it's shown
+in the mode line."
+  (setq lsp--workspace-diagnostics-mode-line-string (format "[%s] " (lsp--calculate-workspaces-diag-statistics))))
 
 (cl-defstruct lsp-diagnostic
   (range nil :read-only t)
@@ -1179,6 +1193,10 @@ PARAMS - the data sent from WORKSPACE."
      :source source
      :message (if source (format "%s: %s" source message) message)
      :original diag)))
+
+(defvar lsp-workspace-diagnostics-mode-line-string ""
+  "String with a summary of workspace diagnostic information,
+  optionally shown as part of the Emacs mode line.")
 
 (defun lsp--on-diagnostics (workspace params)
   "Callback for textDocument/publishDiagnostics.
@@ -5041,7 +5059,12 @@ argument ask the user to select which language server to start. "
 
     (lsp--info "Connected to %s."
                (apply 'concat (--map (format "[%s]" (lsp--workspace-print it))
-                                     lsp--buffer-workspaces)))))
+                                     lsp--buffer-workspaces))))
+
+  (when lsp-show-mode-line-workspace-diagnostics
+    (progn
+      (add-to-list 'global-mode-string (list '(t lsp--workspace-diagnostics-mode-line-string)))
+      (add-hook 'lsp-after-diagnostics-hook 'lsp--update-mode-line-workspaces-diag-statistics nil t))))
 
 (provide 'lsp-mode)
 ;;; lsp-mode.el ends here
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -574,12 +574,6 @@ must be used for handling a particular message.")
   :type 'number
   :group 'lsp-mode)
 
-(defcustom lsp-show-mode-line-workspace-diagnostics t
-  "Whether to show a summary of diagnostics for all open
-  workspaces in the mode line."
-  :type 'boolean
-  :group 'lsp-mode)
-
 (defvar lsp-custom-markup-modes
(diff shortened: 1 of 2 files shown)


danielmartin:
@yyoncho Do you like the minor mode approach? I have decided to add this as a minor mode that enables/disables a custom mode line string with some properties (colors, similarly to a `*compilation*` buffer). I need to check how it looks in a customized mode line (I use a pretty much stock one).

yyoncho:
@danielmartin yes seems like a good idea.

I think that we should make sure that the errors appear only in lsp enabled buffers - if we do not do that this will be the first issue reported by the users. I am not sure whether it will work but possibly we need to define lsp-workspace-diagnostics-mode-line-string as buffer-local and update it in each workspace buffer. That means that the minor mode you have defined should be a global minor mode.

One minor issue: can we ensure that the order of the diagnostics is (error warning info message)? (lsp-treemacs has that bug too)

We should think about a way to enable that mode by default and have an option to turn that off. IMO we should have all new features enabled by default because I suspect that nobody is re-reading the readme.

yyoncho:
An alternative approach I guess is to calculate the diagnostics, store them in a global var/session scoped var and then have a second hook like `diagnostics statistics updated` + minor mode like in this version of the PR. Otherwise, we will have to recalculate the diagnostic statistics per buffer. Or look for global-mode-string alternative...

Forgot to mention that it works fine with doom modeline so good job! :)

yyoncho:
CC @seagle0128 since he has expertise on modelines and can advice on how to make it work only in lsp-mode.

seagle0128:
Hi all,
Thanks to involve me in this PR.

In my understanding, this PR is to leverage `global-mode-string` to display diagnostics information. `flycheck` or `flymake` just displays the buffer diagnostics information. Correct?

I reviewed the codes and added some comments. It seems work, I suppose. But I still have some questions here. 

1. `lsp-after-diagnostics-hook` is called by `lsp--on-diagnostics`. Will `lsp--on-diagnostics` update all buffers always, even no diagnostics change for new buffers? `lsp--workspace-diagnostics-mode-line-string` is a buffer-local variable, that means should update for all buffers, thus may bring big impacts to the performance. Perhaps it's better store them in a global var/session scoped var as @yyoncho suggest. 
1. When working on two or more workspaces at the same time, how to update the workspace diagnostics after switching the buffers? 
1. `lsp-workspace-diagnostics-mode` is a global minor mode, but `lsp--workspace-diagnostics-mode-line-string` is a buffer-local var. So, why not make `lsp-workspace-diagnostics-mode` a local minor mode, and enable/disable it in `lsp-mode-hook`?
1. Following the question above, why just integrate these information in `lsp-mode-line` with an option to enable/disable? I know `lsp-mode-line` is in `minor-mode`, and `lsp--workspace-diagnostics-mode-line-string` is in `global-mode-string`, but actually they are not different. If I design it for `doom-modeline`, I just want to integrate into `lsp` segment while not provide a separate one just for additional information. If it's a separate one, it should have more functionalities (already done or implement in future). e.g. list diagnostics for all buffer in the workspace/project, filter and jump to specific files, etc. Otherwise the information is not too valuable. 

I don't know the background and details of this feature, just my opinions. Please correct me if I am wrong. Thanks!

yyoncho:
> In my understanding, this PR is to leverage `global-mode-string` to display diagnostics information. `flycheck` or `flymake` just displays the buffer diagnostics information. Correct?
> 

Yes.

> I reviewed the codes and added some comments. It seems work, I suppose. But I still have some questions here.
> 
> 1. `lsp-after-diagnostics-hook` is called by `lsp--on-diagnostics`. Will `lsp--on-diagnostics` update all buffers always, even no diagnostics change for new buffers? `lsp--workspace-diagnostics-mode-line-string` is a buffer-local variable, that means should update for all buffers, thus may bring big impacts to the performance. Perhaps it's better store them in a global var/session scoped var as @yyoncho suggest.

> 2. When working on two or more workspaces at the same time, how to update the workspace diagnostics after switching the buffers?

We need another hook which will run when the diagnostics change no matter whether the buffer is active. Otherwise, if an issue is fixed in а buffer that is not opened the modeline wont be updated.


> 3. `lsp-workspace-diagnostics-mode` is a global minor mode, but `lsp--workspace-diagnostics-mode-line-string` is a buffer-local var. So, why not make `lsp-workspace-diagnostics-mode` a local minor mode, and enable/disable it in `lsp-mode-hook`?

This is one of the options but we should decide whether we need to avoid calculating the statistics for each buffer.

>  If I design it for `doom-modeline`, I just want to integrate into `lsp` segment while not provide a separate one just for additional information. If it's a separate one, it should have more functionalities (already done or implement in future). e.g. list diagnostics for all buffer in the workspace/project, filter and jump to specific files, etc. Otherwise the information is not too valuable.

Separate what? Can you give an example? If I have understood you correctly you are suggesting global-mode-string alternative.

yyoncho:
We may use a buffer switch event to update the mode string as well (https://stackoverflow.com/questions/47456134/emacs-lisp-hooks-for-detecting-change-of-active-buffer )

seagle0128:
> Separate what? Can you give an example? If I have understood you correctly you are suggesting global-mode-string alternative.


I mean this new modeline string is not within the default lsp-mode-line. Useing global-mode-string is fine but not necessary in this case unless there will be more features.

yyoncho:
@danielmartin it seems like @seagle0128 is right - it will be more convenient if we use new minor mode(in addition to the global mode that you have created) + modeline(see lsp-mode-line) instead of the global string. This way we will be able to turn on that mode only in lsp-mode buffers. The global major mode will add a hook to lsp-mode-hook which enables that minor mode. Also, it will be good to have the statistic data raw somewhere. So modelines like doom-modeline could use it and format it it a way that is consistent with the look of the modeline. 

Or we could use lsp-mode-line as you initially suggested(it is hard for me to say what will be better).

danielmartin:
@yyoncho Yeah, my initial idea was to simply expand the `lsp-mode` lighter by extending `lsp-mode-line` with additional statistics data. The problem that was mentioned was that some custom mode lines "massage" well-known mode lines like the LSP one, and they show a rocket icon instead. They could adapt their mode line if we expose the raw diagnostics data, though.

> it will be more convenient if we use new minor mode(in addition to the global mode that you have created) + modeline(see lsp-mode-line) instead of the global string.

Why introducing a new minor mode? As the only feature is to add more information for the mode line (no custom keymap, etc.), do we really need to introduce a new minor mode? (Maybe yes, I want to be sure.)

I'd also want to get the statistics update right. By listening to the `lsp-on-diagnostics` hook we will show current information when we get information from the server. But a workspace file may have been modified externally (in another editor, for example) and the hook won't be called in that case. Isn't the `publishDiagnostics` notification sent per-file? Wouldn't that require that the client opens that modified file? 

I have to double check but I think that what VS Code shows is the amount of diagnostics _from the open files_. This means that if you open a file from the workspace with errors, the count will increase. It will decrease when you close that file. So VS Code is not showing you the total count of errors from the workspace, only what you have open, so the problem of externally modified files does not exist for them (and I think it'd difficult or impossible to solve with the current LSP protocol implementation).

danielmartin:
>  So VS Code is not showing you the total count of errors from the workspace, only what you have open, so the problem of externally modified files does not exist for them

You may have the file in a non-active editor tab (or a non-active Emacs buffer). But I assume that the `publishDiagnostics` notification will be sent in that case because the server knows about every open file and should reparse them if they change anyhow. I'm not 100% sure about that.

yyoncho:
> @yyoncho Yeah, my initial idea was to simply expand the `lsp-mode` lighter by extending `lsp-mode-line` with additional statistics data. The problem that was mentioned was that some custom mode lines "massage" well-known mode lines like the LSP one, and they show a rocket icon instead. They could adapt their mode line if we expose the raw diagnostics data, though.

It was me that raised that concern but the exposing the raw data is solving it.

> Why introducing a new minor mode? As the only feature is to add more information for the mode line (no custom keymap, etc.), do we really need to introduce a new minor mode? (Maybe yes, I want to be sure.)

I cannot think of a strong reason to introduce new minor mode instead of reusing the lsp-mode one.

>
> I'd also want to get the statistics update right. By listening to the `lsp-on-diagnostics` hook we will show current information when we get information from the server. But a workspace file may have been modified externally (in another editor, for example) and the hook won't be called in that case. Isn't the `publishDiagnostics` notification sent per-file? Wouldn't that require that the client opens that modified file?

Our hook won't be called if the file is not open, so we should add a another hook e. g. `lsp-on-diagnostics-received-hook` which will be executed no matter whether the file is open(and yes, lsp-treemacs/lsp-ui-flycheck-error-list have that bug too - they do not refresh if diagnostics change in a file that is not open.)

>
> I have to double check but I think that what VS Code shows is the amount of diagnostics _from the open files_. This means that if you open a file from the workspace with errors, the count will increase. It will decrease when you close that file. So VS Code is not showing you the total count of errors from the workspace, only what you have open, so the problem of externally modified files does not exist for them (and I think it'd difficult or impossible to solve with the current LSP protocol implementation).

No, it is not(or maybe the server you are checking is doing that) - check the screenshot.


Untitled (Workspace) - Visual Studio Code_131

seagle0128:
Is it possible to leave this functionality to `flycheck` or `flymake`? And an option to show diagnostics of current buffer or current project.

yyoncho:
> Is it possible to leave this functionality to flycheck or flymake?

There are two problems that have to be solved:

1. Last time I checked both do not allow reporting errors for files that are not open(part of their design). One of the reasons we have `lsp-treemacs-errors-list`. 
2. It is not nessesery lsp-mode root to match projectile/project.el one.

> And an option to show diagnostics of current buffer or current project.

We are going to have 3 modes `current buffer` / `current project` / `all projects`.

CC @fmdkdd @joaotavora in case they are interested in working on extending flycheck/flymake to support grouping errors by project or refute 1) .

fmdkdd:
> do not allow reporting errors for files that are not open(part of their design

You are correct that Flycheck is not project-aware.  It's more a consequence of the design rather than a conscious decision, AFAIK.  Flycheck was built for quick feedback on individual files, not full diagnostics for an entire project.  

Although it's been a year now that we display errors from other files in the error list and on the first line of the current buffer if the checker gives them (https://github.com/flycheck/flycheck/pull/1427), and I believe the error count in the modeline should include these other errors.  This only works for error-level diagnostics right now, but we could make it configurable.

yyoncho:
@fmdkdd 

> Although it's been a year now that we display errors from other files in the error list and on the first line of the current buffer if the checker gives them

Yes, but the `flycheck-error-new`'s  buffer param is mandatory which limit's us to report errors that are for files that are not opened.

fmdkdd:
> Yes, but the flycheck-error-new's buffer param is mandatory which restricts us - no way to report errors that are for files that are not opened.

The `buffer` param merely indicates which buffer triggered the check.  The `filename` param indicates which file the diagnostics is reported for.  You can have errors for unopened files, but they just have to be attached to a specific buffer.  Think of it this way: with Flycheck without lsp, when you edit a file `foo.c`, Flycheck executes an external checker *for this current file*, gathers all diagnostics, and adds overlays and populate the error list.  The `flycheck-error` objects are all attached to the buffer that triggered the check.  However, the checker may have returned errors for other files (`bar.c`), in which case we can still display them, but *only in the current buffer*, since the file `bar.c` has no associated buffer.  When you open `bar.c`, it won't have any error because flycheck-mode is not global, and thus does not "pass" errors from one buffer to the other.   However, if you had jumped from the `bar.c` error shown in the `foo.c` buffer, we would have triggered a check to make the errors appear there as well.

Looking at [this bit in lsp-ui](https://github.com/emacs-lsp/lsp-ui/blob/master/lsp-ui-flycheck.el#L183-L186), you are adding only the diagnostics for the current buffer as flycheck-error objects.  If instead you added all the diagnostics to the buffer, and used the `:filename` param to point to the diagnostic file, you would get the total number of errors in the modeline.

When no files for the project are opened, this is more tricky, since there's no buffer to report to!  That's where lsp-tree comes in I guess.

yyoncho:
We already have an implementation in master for the modeline diagnostics.

PR #810, opened 2019-04-30
lsp-mode: add lsp-organize-imports

fjl:
Improved version of #759 which wraps the `source.organizeImports` code action.


yyoncho:

We should do the same for the rest of the defined code actions.

shuxiao9058:
How to disable message organize-imports?


image

yyoncho:
@shuxiao9058 
``` 
(setq lsp-ui-sideline-show-code-actions nil)
```

FTR I am planning to add a filter in lsp-ui-sideline to remove source level code actions(similar to what vscode does).

PR #822, opened 2019-05-10
Fix return value of lsp-eldoc-function

Khady:
The documentation of `eldoc-documentation-function` says:

> It should return nil if there's no doc appropriate for the context.

`lsp-eldoc-function` was always returning eldoc-last-message. Because
of this, the last message displayed by eldoc would not disappear when
the point is moved in a place where no doc is available or
lsp-eldoc-hook returns nil and doesn't call lsp-hover.

The downside is that moving the point inside the same identifier might
trigger a refresh of the eldoc message, even if it is the same. Maybe
it is possible to fix this problem by changing `lsp-hover`. So that it
returns the documentation string instead of calling `eldoc-message`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3198,9 +3198,7 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
    (lambda (fn)
      (condition-case nil
          (funcall fn)
-       (lsp-capability-not-supported nil))
-     nil))
-  eldoc-last-message)
+       (lsp-capability-not-supported nil)))))
 
 (defvar-local lsp--highlight-bounds nil)
 (defvar-local lsp--highlight-timer nil)


yyoncho:
Can we go the other way - ensure that it set to nil in the problematic cases (e. g. when hooks are nil). I want to avoid the blinking since it is not bearable when you have multiline eldoc.

Khady:
I'm using multiline eldoc and it is fine for me. Maybe because the lsp server I use is fast. Btw, `lsp-hover` does not respect `eldoc-echo-area-use-multiline-p`.

> ensure that it set to nil in the problematic cases (e. g. when hooks are nil)

With this PR, `lsp-eldoc-function` already returns nil if `lsp-eldoc-hook` returns nil. Before that it was never returning nil.

yyoncho:
> With this PR, lsp-eldoc-function already returns nil if lsp-eldoc-hook returns nil. Before that it was never returning nil.

What I am saying is that we may fix the problem you are seeing by keeping lsp-eldoc-function as it is and ensuring that the on-hover will cleanup the eldoc-message. As a side note, I am unable to reproduce the stale eldoc message and I have tested with ~10 different language servers.

Khady:
> As a side note, I am unable to reproduce the stale eldoc message and I have tested with ~10 different language servers.

Probably because you don't have a custom `lsp-eldoc-hook` that returns nil at some positions.

> What I am saying is that we may fix the problem you are seeing by keeping lsp-eldoc-function as it is and ensuring that the on-hover will cleanup the eldoc-message.

Given the documentation of `eldoc-documentation-function`, I really have the impression that the current situation is a hack that works by accident. And it duplicates some work that eldoc is already doing when used properly.

yyoncho:
> Given the documentation of eldoc-documentation-function, I really have the impression that the current situation is a hack that works by accident. And it duplicates some work that eldoc is already doing when used properly.

All this is needed since eldoc-documentation-function is sync while in lsp-mode we update the eldoc text asynchronously. Returning nil is as much hack as returning the old eldoc message because in both cases we do not know the real value with the only difference that returning old eldoc message does not cause blinking.

Khady:
> All this is needed since eldoc-documentation-function is sync while in lsp-mode we update the eldoc text asynchronously.

hmm that's a problem indeed. Then you are right. The blinking I noticed was probably because of this.

> Returning nil is as much hack as returning the old eldoc message

But unless I missed something about run-hook-wrapped and condition-case, with this diff the value returned is the one returned by lsp-eldoc-hook? Anyway, it doesn't matter much, as there is the async problem.

PR #829, opened 2019-05-12
Enhancements to lsp-file-watch-ignored

madscientist:
The first three of these commits are pretty straightforward.

The final one does a bit of rearrangement and adds new ignored directories for autotools generated output such as the .deps directory etc.  I don't know if this is appropriate or whether we should be requesting people update this themselves with their own personal stuff, but there're a number of relatively esoteric regexps in the list already so...

But I purposefully put this commit last so it could be left out of the merge if desired.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,6 +4,6 @@ all:
 	cask build
 
 test: all
-	cask exec ert-runner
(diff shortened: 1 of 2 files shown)


yyoncho:

PR #834, opened 2019-05-14
Add Emacs 26.1 and 26.2 to CI Build (#812)

chasinglogic:


chasinglogic:
I am investigating the TLS build failures

yyoncho:
I had them green on my machine using emacs26.2.

chasinglogic:
I don't think there's anything on the code side that needs to change. I'm suspicious it was a network fluke and so I'm going to close and re-open this PR to trigger a rebuild.

chasinglogic:
Ah it looks like we aren't pulling in the install_gnutls target for some reason from the emacs-travis makefile:

https://github.com/flycheck/emacs-travis/blob/master/emacs-travis.mk#L89

chasinglogic:
@yyoncho It looks like my WIP commit (which I will squash) fixed the cask install issue but the tests are failing. I'm not very familiar with the LSP, do you have any pointers on where to start to fix the test failures?

yyoncho:
@chasinglogic 

Thank you for helping out with such a boring issue!

> I'm not very familiar with the LSP, do you have any pointers on where to start to fix the test failures?

The ecukes test is failing. The test consists of several interactive steps like: interactively opening a python file, starting the language server and performing a few common operations. I think that this would help - https://github.com/ecukes/ecukes#debugging Here it is the output with --debug option. https://travis-ci.org/yyoncho/lsp-mode/jobs/533976224 which doesn't reveal anything usefull. IMO it might be even ecukes bug. I am done for today, you may try the second option. Hopefully, it will reveal the call stack of this failure... If it doesn't we might ask in ecukes repo for help.

chasinglogic:
Cool beans I will follow up on this probably next week.

chasinglogic:
I can't repro this on my local Linux machine but it's a much newer version of Ubuntu (19.04). I started working on repro'ing this in a trusty docker container (trying to emulate Travis) but got interrupted.

I probably won't get back to this until Friday evening / Sunday but I am still working on it.

seagle0128:
I encounter the ci issue in Centaur Emacs, with lsp-mode. I noticed the same errors in ci of lsp-mode for snapshot.

```elisp
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

What's the problem here?

chasinglogic:
Hey @seagle0128 

That's a slightly different CI issue than this one. I'm not sure how to help.

@yyoncho I'm still working on this particular problem, sorry for the radio silence life has gotten in the way on this one for a bit.

seagle0128:
Thank you, @chasinglogic . Hope @yyoncho can help on this issue.

yyoncho:
@chasinglogic no worries. But IMO don't spend that much time into this issue - we dont have that much manpower. We could live without it and really there are other issues/features that you would enjoy much more.

yyoncho:
@seagle0128 seems like the language server is timing out but I need more context on that...

seagle0128:
@yyoncho I see the same error in https://travis-ci.org/emacs-lsp/lsp-mode/jobs/540188796.

yyoncho:
@seagle0128 IMO that error is issue with the infrastructure, e. g. initialization of the tests took more than 10 seconds and travis cancelled the build. But that is just a guess...

seagle0128:
@yyoncho It's over 10 mins... I disable `lsp-mode` then all work well.

yyoncho:
@seagle0128 can you give me a link to what the tests are doing? The error is the same but could be caused by a different issue.

seagle0128:
@yyoncho You can find the details on https://travis-ci.org/seagle0128/.emacs.d/jobs/540445105.

The test case is 
```shell
$ ${EMACS:=emacs} -nw --batch --eval '(let ((debug-on-error (>=  emacs-major-version 26)) (url-show-status nil) (user-emacs-directory default-directory) (early-init-file (expand-file-name "early-init.el")) (user-init-file (expand-file-name "init.el")) (load-path (delq default-directory load-path))) (when (>= emacs-major-version 27) (load-file early-init-file)) (setq centaur-lsp nil) (load-file user-init-file) (when (>= emacs-major-version 26) (run-hooks (quote after-init-hook)) (run-hooks (quote emacs-startup-hook))))'
```

yyoncho:
@seagle0128 pushed https://github.com/emacs-lsp/lsp-mode/commit/1f91a7c7b8d026ace82835197ed9ea8192ed56e8

seagle0128:
> @seagle0128 pushed [1f91a7c](https://github.com/emacs-lsp/lsp-mode/commit/1f91a7c7b8d026ace82835197ed9ea8192ed56e8)

It works now. Thanks a lot!

chasinglogic:
So I've been unable to reproduce this in either an Ubuntu Trusty VM (using Vagrant) or Container.

It's something specific about the image that Travis is using that causes this problem. I'm going to read some more on ecukes so I can get an Emacs stack trace and debug from there but the only thing I can figure is that the pyls setup code is failling and that the scenarios actually run.

yyoncho:
@chasinglogic I also think that the issue is caused by the image... Alternatively, the issue might be caused by old bytecode available on 26.2 image but I am not sure how that is possible. I wonder whether there are alternative images that provide emacs 26?

chasinglogic:
@yyoncho 

I don't think it's caused by old byte code, while working on this I discovered that the cache directive is being ignored by Travis. This means that the bytecode should be fresh every run unless I'm missing something?

yyoncho:
> @yyoncho
> 
> I don't think it's caused by old byte code, while working on this I discovered that the cache directive is being ignored by Travis. This means that the bytecode should be fresh every run unless I'm missing something?

I am just guessing because the error looks like there is an issue with the bytecode.

chasinglogic:
Some more interesting information:

1. Trying to use newer Ubuntu (16.04 Xenial) doesn't seem to work on Travis. It recognizes that I asked for xenial then puts me on a Trusty Tahr image anyway.
2. I believe this is because the `language: emacs-lisp` is not a valid value so we get some default image that Travis provides when it doesn't know what to do.
3. I'm trying to specify `language: ruby` to actually get xenial because if we can get Travis to give us a newer Ubuntu there is an Emacs table PPA that we can use to access vanilla builds of Emacs. It's also possible that on a newer Ubuntu it would just work with the existing makefile from flycheck.

If the existing makefile can be used I will do so.

yyoncho:
In order to move this forward, I would propose altering the script to skip equkes tests for emacs 26+

chasinglogic:
I have made the requested change.

yyoncho:

PR #836, opened 2019-05-15
Add scala to lsp-language-id-configuration.

dsyzling:
As discussed - add scala to list of lsp language configurations so we can maintain and refactor in one place.


yyoncho:
Thank you!

PR #848, opened 2019-05-22
lsp-rust: correct types for settings

kurnevsky:
- Use `choice` for optional settings instead of `repeat` (otherwise it might be confusing e.g. for `lsp-rust-rustflags` - it always has to be single string).
- Add `nil` variant to `lsp-rust-wait-to-build` and use it as default as it means automatic inferred time (default in rls).
- Correct some boolean settings.

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -32,21 +32,27 @@
   :link '(url-link "https://github.com/rust-lang/rls")
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-rust-sysroot ""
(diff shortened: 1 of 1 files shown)


yyoncho:
@GoingFast can you verify the PR?

kurnevsky:
Also set `sysroot` and `target` to nil as empty string is not correct value - rls doesn't check for it.

GoingFast:
@yyoncho Yes now flycheck works

yyoncho:
@GoingFast @kurnevsky thank you!

PR #849, opened 2019-05-22
Detect dart executable from dart-sdk-dir

itome:
Dart analysis server's version must be same as `dart` executable.
So we should detect dart executable from dart-sdk-dir.


yyoncho:

PR #853, opened 2019-05-24
Stop sending textDocument/didSave when server not support

itome:
dart_analysis_server return `Unknown method textDocument/didSave`.
https://github.com/dart-lang/sdk/issues/36464
We should stop sending `textDocument/didSave` when server capabilities doesn't include `textDocumentSync > save`.
https://github.com/microsoft/language-server-protocol/issues/288

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2377,6 +2377,12 @@ in that particular folder."
     (and (hash-table-p sync)
          (gethash "willSaveWaitUntil" sync))))
 
+(defun lsp--send-did-save-p ()
+  "Return whether did save notifications should be sent to the server."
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

jtpavlock:
Looks like I'm running into this issue again. I can confirm it's fixed with the referenced patch, but with the following code in the version I'm running it persists:

```dart
(defun lsp--send-did-save-p ()
  "Return whether didSave notifications should be sent to the server."
  (let ((sync (lsp:server-capabilities-text-document-sync? (lsp--server-capabilities))))
    (or (memq sync '(1 2))
        (lsp:text-document-sync-options-save? sync))))
```

Should I open a new issue up?

Issue #661, opened 2019-02-20
[Feature Request] Support for haxe language


mundusnine:
Following could be used to integrate the haxe language
https://github.com/vshaxe/haxe-languageserver

yyoncho:
@mundusnine I checked the language server - it seems like provides a lot of extension methods and requires a relatively high amount of work in order to make it work(I am judging based on the size of the client)  . Based on this it seems like the issue could be addressed only by Haxe user. Do you want to give it a try?

mundusnine:
For now, I don't really have the time. But I will put it in my stack for the futur. @fullofcaffeine also wanted this feature maybe he would be interested.

Gama11:
The size of the client might be a bit deceptive, we implement some things like task providers, custom tree views etc that aren't really necessary for the language server. People have gotten it to work with Neovim (https://github.com/vshaxe/vshaxe/issues/328) and Sublime Text before.

zaethan:
I was able to make it (somewhat) work, by adding this to my config:

``` 
  (defgroup lsp-haxe nil
    "LSP support fer Haxe"
    :group 'lsp-mode
    :link '(url-link "https://github.com/vshaxe/vshaxe"))

  ;; haxe lsp is executed via node
  (defcustom lsp-clients-haxe-executable "node"
    "haxe lsp is executed via node"
    :group 'lsp-haxe
    :risky t
    :type 'file)

  ;; arguments for node to start the haxe lsp
  (defcustom lsp-clients-haxe-args '("/mnt/RamDisk/vshaxe/server/bin/server.js")
    "arguments for node to start the haxe lsp"
    :group 'lsp-haxe
    :risky t
    :type 'file)

  (defun lsp-clients--haxe-command ()
    "Haxe LS startup command"
    `(,lsp-clients-haxe-executable ,@lsp-clients-haxe-args))

  ;; adapted from go / https://github.com/emacs-lsp/lsp-mode/pull/668/files
  (defcustom lsp-clients-haxe-arguments ["build.hxml"]
    "https://github.com/vshaxe/vshaxe/issues/328#issuecomment-472053630."
    :type 'string
    :group 'lsp-clients-haxe)

  (defun lsp-clients-haxe--make-init-options()
    "Init options for the haxe lsp"
    `(:displayArguments ,lsp-clients-haxe-arguments))

[...]

PR #856, opened 2019-05-24
added partial Haxe integration

zaethan:
Partial integration of the [haxe language server](https://github.com/vshaxe/haxe-language-server).
Also added a description to the readme.org about how to use the server. Since the server seems to be only available as a vscode extension and pure source code, lsp-haxe.el cannot know the path of the server.js file. Thus it provides a defcustom to be set by the user.
Emacs does not have a haxe mode on its own, but there is a package in melpa, which provides that. It is unmaintained, but works well. Thus I set the major modes variable just to haxe-mode.

As for issues:
* There is the code completion problem here https://github.com/emacs-lsp/lsp-mode/issues/661#issuecomment-495399421
  * It does actually print an error in the message buffer which I just didn't notice before, because of other messages from the lsp-modend: `company-call-backend-raw: Company: backend company-lsp error "Wrong type argument: hash-table-p, nil" with args (post-completion import)`
  * in this ([log.txt](https://github.com/emacs-lsp/lsp-mode/files/3218757/log.txt)) log I:
    1. opened the empty file HXDB2.hx
    2. typed `imp` for import
    3. hit enter for completion
    4. killed the buffer
* Remove unused imports (and unused variables) is actually quite weird:
  * Adding a useless import and trying to remove it with this action does not do anything.
  * After saving the file (and restarting emacs), it does remove the line with the unused import.
  * Clicking Remove unused import repeatedly always removes the line where the import was, regardless of what's there now.
  * In this ([log_rem.txt](https://github.com/emacs-lsp/lsp-mode/files/3218771/log_rem.txt)) log I:
    1. Opened the file Test.hx [Test.hx.txt](https://github.com/emacs-lsp/lsp-mode/files/3218776/Test.hx.txt)
    2. moved down the cursor a little and clicked on `Remove ...` until nothing was left (I moved the cursor down some more somewhere in between)
    3. killed the buffer
* Maybe more, maybe not, this is all I've tested for now.

To reproduce, it probably be best to get haxe here https://haxe.org/download/version/4.0.0-rc.2/ and the server here https://marketplace.visualstudio.com/items?itemName=nadako.vshaxe and set the lsp-clients--haxe-server-path variable to the extracted server.js. If you want to avoid installing haxe, set the HAXE_STD_PATH environment variable to the extracted std folder prior to starting emacs.


yyoncho:
I got some trouble making all this work, but finally, I did it after upgrading to haxe 4.0.

>  company-call-backend-raw: Company: backend company-lsp error "Wrong type argument: hash-table-p, nil" with args (post-completion import)

This one is caused by haxe ls not following the language server protocol: textEdit should have range populated: 


```
  {
    "data": {
      "origin": 1
    },
    "documentation": {
      "value": "```haxe\nenum abstract Main(T) {\n\t|\n}\n```",
      "kind": "markdown"
    },
    "insertTextFormat": 2,
    "textEdit": {
      "newText": "enum abstract ${1:Main}(${2:T}) {\n\t$0\n}"
    },
    "sortText": "~~",
    "kind": 15,
    "detail": "enum abstract Main",
    "label": "enum abstract"
  },
```

and the definition: 

``` javascript
interface TextEdit {
	/**
	 * The range of the text document to be manipulated. To insert
	 * text into a document create a range where start === end.
	 */
	range: Range;

	/**
	 * The string to be inserted. For delete operations use an
	 * empty string.
	 */
	newText: string;
}
```


> Clicking Remove unused import repeatedly always removes the line where the import was, regardless of what's there now.

This is a minor/usability bug on our side.

yyoncho:
There is one more issue - the haxe server does not declare that it wants to receive didSave events but at the same time vscode is sending them...

yyoncho:
> Adding a useless import and trying to remove it with this action does not do anything.
After saving the file (and restarting emacs), it does remove the line with the unused import.
Clicking Remove unused import repeatedly always removes the line where the import was, regardless of what's there now.


That same bug exists in vscode - the issue is that haxe server updates the code actions once you save the file. So, if you havent saved the file you could execute the code action multiple times and it will delete the current line. If you checkout the following branch - https://github.com/yyoncho/lsp-mode/tree/zaethan-master and you do save between the actions the issue would not appear.

Gama11:
>There is one more issue - the haxe server does not declare that it wants to receive didSave events but at the same time vscode is sending them...

I think in older versions of the LSP, it wasn't possible to explicitly specify this (and it defaulted to just sending it?). Anyway, https://github.com/vshaxe/haxe-language-server/commit/062bcba1a6c929047aed8b10ad25c3a6fc91234b should help? :)

yyoncho:
@Gama11 

Looks like it should be ok after this PR. But what about the missing range in the textEdit?

Gama11:
Well, that's a separate issue. :) (tracked here https://github.com/vshaxe/vshaxe/issues/350)

yyoncho:
merged as https://github.com/emacs-lsp/lsp-mode/commit/3bec696e15cf7008ebafd5798733a0b464164ed1 . In addition, I did several other config changes in fae4544 .

PR #860, opened 2019-05-26
The client and server were split, so there is no more server directory

Daniel-V1:
in this repo. This fixes the readme and the run command


yyoncho:
Thank you!

PR #869, opened 2019-06-04
Add lsp--validate-client.

vibhavp:


yyoncho:
Looks good to me. IMO the correct way to verify the PR is to package-install every lsp-mode related package (`Needed by` section in https://melpa.org/#/lsp-mode ) and verify that register call is passing. After doing that feel free to merge the PR.

yyoncho:
Cleanup.

PR #871, opened 2019-06-05
Add lsp-scala

kurnevsky:
As discussed here: https://github.com/rossabaker/lsp-scala/issues/15


yyoncho:
Looks good but I am not quite sure whether we should go with that name and what will happen if lsp-scala user downloads a new version of lsp-mode which contains another lsp-scala package into it. I would propose naming the lsp-mode package `lsp-metals` and registering it with higher priority than lsp-scala and different server id. Then you could add a warning in the original lsp-scala package to let the users know that they should migrate to lsp-metals. WDYT?

kurnevsky:
I already used different server id `metals` (old `lsp-scala` uses `scala`) so it seems they shoudn't conflict? Regarding priorities - I'd left it `-1` (old `lsp-scala` has 0) so that users will still use old `lsp-scala` unless they switch it explicitly. There is one major difference - I added configs as defvars instead of plain list so manual intervention might be needed anyway.

yyoncho:
> I already used different server id metals (old lsp-scala uses scala) so it seems they shoudn't conflict? 

My main point is that if someone has `(require 'lsp-scala)` in his config it will pick the first `lsp-scala` on the load-path.

kurnevsky:
ok, that makes sense. At least there is a precedent when the package is called after lsp server - lsp-pyls.

yyoncho:

PR #874, opened 2019-06-07
More precise link for metals installation

kurnevsky:
Sorry, wasn't aware of these docs :)

There is an instruction for metals installation for emacs. I put this link to the table from readme so it might be helpful to have it in the docs as well.


yyoncho:
Thank you!

PR #875, opened 2019-06-08
Add missing shellscript language id

kurnevsky:


yyoncho:

Issue #831, opened 2019-05-13
Can we support lua language?


drwx:
Not know well about elisp, but hope lua support by default. Thanks.

yyoncho:
which lua language server is the best?

shuxiao9058:
https://github.com/EmmyLua/EmmyLua-LanguageServer

https://github.com/sumneko/lua-language-server

A6GibKm:
There is also https://github.com/Alloyed/lua-lsp, I don't know which one is better but the later can be easily installed with `luarocks`.

yyoncho:
https://github.com/sumneko/lua-language-server seems to be the most complete but it is not very clear how to do the build outside of vscode.

https://github.com/EmmyLua/EmmyLua-LanguageServer - already has emacs configuration in its readme.

yyoncho:
Added builtin configuration for https://github.com/EmmyLua/EmmyLua-LanguageServer

PR #879, opened 2019-06-12
Add missing lua language id.

drwx:


yyoncho:
Thank you!

Can you link the lua configuration in https://github.com/emacs-lsp/lsp-mode/issues/831 ?

PR #880, opened 2019-06-13
Add F# Language support

TOTBWF:
This PR adds support for the F# language server: https://github.com/fsharp/FsAutoComplete.


lsp-fsharp.el
```diff
@@ -0,0 +1,80 @@
...
+(defun lsp-fsharp--fsac-runtime-cmd ()
+  "Get the command required to run fsautocomplete based off of the current runtime."
+  (pcase lsp-fsharp-server-runtime
+    ('net-core '("dotnet"))
+    ('mono '("mono"))
+    ('net-framework nil)))
+
+(defun lsp-fsharp--make-launch-cmd ()
+  "Build the command required to launch fsautocomplete."
+  (if lsp-fsharp-server-install-dir (lsp-warn "Cannot locate fsautocomplete"))
```
baronfel:
```suggestion
  (if lsp-fsharp-server-path (lsp-warn "Cannot locate fsautocomplete"))
```

should this be `lsp-fsharp-server-path` instead of `lsp-fsharp-server-install-dir`, since you renamed it a few commits ago?

TOTBWF: 🤦‍♂ Nice catch, fixed.

yyoncho:

kqr:
@TOTBWF Are you sure this is correct? It tests if lsp-fsharp-server-path is set, and if it is, it warns about it not being able to find lsp-fsharp-server. That sounds backwards to me.

iNecas:
@kqr I've hit this as well, seems just like an issue with `if/unless`, IMO this fixes the issue https://github.com/emacs-lsp/lsp-mode/pull/889

TOTBWF:
Yeah, that PR fixes the issue. I totally brainfarted here, nice catch 🙂

PR #889, opened 2019-06-15
Fix warnings produced by lsp-fsharp

iNecas:
While testing the lsp-fsharp, I was getting multiple warnings. Seems
like there was wrong usage of `if` instead of `unless` + couple of
notifications produced by the fsautocomplete that were not configured
and caused the warnings to produced.

I think marking the notifications to be ignored until we have some use
for them is better then showing the annoying warning messages.

### DIFF
--- a/lsp-fsharp.el
+++ b/lsp-fsharp.el
@@ -61,7 +61,7 @@
 
 (defun lsp-fsharp--make-launch-cmd ()
   "Build the command required to launch fsautocomplete."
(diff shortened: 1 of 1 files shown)


yyoncho:
@kqr / @TOTBWF - can you validate the PR?

TOTBWF:
Looks good to me! Sorry about this, must've gotten myself into some weird state when I was testing.

yyoncho:
> Looks good to me! Sorry about this, must've gotten myself into some weird state when I was testing.

No worries. 

@iNecas 

PR #891, opened 2019-06-16
fix: .ts files should be treated as typescript

### LINKED ISSUE #887: .ts files are handled as javascript files, when using js2-mode
**To Reproduce**
Open a `.ts` file with js2/rjsx-mode (that has support for typescript), and you'll see a lot of errors like `8006`, `8011`, `8010`.

**Expected behavior**
As a typescript file is open, it should be handled as a typescript file.

`typescript-language-server`

**OS**
macOS

**Do you have any suggestions?**
The solution would be the same as the solution for #781, fixed in #782.
I solved this for my configuration using this:
```
(after! lsp-mode
  (add-to-list 'lsp-language-id-configuration '(".*.ts" . "typescript")))
```
I may open a PR adding this line `lsp-language-id-configuration` if the maintainers think that this solution should be generalized (all `ts` files should be treated as typescript files no matter the mode)

gustavo-depaula:
fixes #887

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -531,6 +531,7 @@ If set to `:none' neither of two will be enabled."
 
 (defvar lsp-language-id-configuration '((".*.vue" . "vue")
                                         (".*.tsx" . "typescriptreact")
+                                        (".*.ts" . "typescript")
                                         (sh-mode . "shellscript")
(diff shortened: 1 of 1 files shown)


yyoncho:

PR #892, opened 2019-06-17
.jsx files should be treated as javascriptreact

isamert:
Just another PR in the spirit of #782 and #891.


yyoncho:
Thank you congrats for your first `lsp-mode` PR!

PR #893, opened 2019-06-17
Add language id config for hack/hack-mode

jra3:
Summary: there is already a hack client in lsp-mode, this adds an appropriate language id for it.

Test Plan: load a file in hack mode, M-x lsp => no longer get the warning about missing language id


jra3:
thanks!

PR #897, opened 2019-06-18
Update FAQ section formatting

### LINKED ISSUE #896: Ability to set LSP server for a given project
The README does not call this out anywhere and it's worth adding, assuming it is possible, for how to tell LSP which server I want for a given project.

For example, on a given project one Typescript/Javascript server works better than another and it would be nice to document how to tie into something like projectile to make this work.

jsmestad:
Have the questions separated from the answers.

Closes #896

### DIFF
--- a/README.org
+++ b/README.org
@@ -265,20 +265,20 @@
 
 ** FAQ
    - How to configure a server with local variables?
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

PR #902, opened 2019-06-21
Add new lsp-deferred entry point.

### LINKED ISSUE #792: LSP server takes a long time to start with a lot of files open
If I have a lot (~100) of files open when I (re)start gopls, it takes a while (~10+ seconds) before lsp commands start working. This is because we send many "didOpen" requests which makes the server busy for a while. Would it be possible to delay didOpen request until you visit the buffer for the first time? We would still need to send didOpen request for any buffers that had unsaved changes, of course. This appears to be how VSCode works (i.e. it only sends a didOpen request for the file you are currently viewing when you open VSCode).

muirrn:
You can use lsp-deferred instead of lsp in your major mode hooks. This
will defer starting the LSP server and adding the file to the LSP
server until the buffer is actually visible in Emacs. This can make
startup a lot faster because it avoids sending all open files at once
to the LSP server when starting Emacs or restarting the workspace.

Fixes #792

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1904,7 +1904,7 @@ CALLBACK - callback for the lenses."
   ;; the Language Server specific messages.
   (status-string nil)
 
-  ;; ‘shutdown-action’ flag used to mark that workspace should not be restarted(e. g. it
+  ;; ‘shutdown-action’ flag used to mark that workspace should not be restarted (e.g. it
   ;; was stopped).
   (shutdown-action)
 
@@ -4676,6 +4676,10 @@ returns the command to execute."
     (setf (lsp--parser-workspace parser) workspace)
     workspace))
 
+
+(defvar-local lsp--buffer-deferred nil
+  "Whether buffer was loaded via `lsp-deferred'.")
+
 (defun lsp--restart-if-needed (workspace)
   "Handler restart for WORKSPACE."
   (when (or (eq lsp-restart 'auto-restart)
@@ -4688,8 +4692,10 @@ returns the command to execute."
     (--each (lsp--workspace-buffers workspace)
       (when (buffer-live-p it)
         (with-current-buffer it
-          (lsp--info "Restarting LSP in buffer %s" (buffer-name))
-          (lsp))))))
+          (if lsp--buffer-deferred
+              (lsp-deferred)
+            (lsp--info "Restarting LSP in buffer %s" (buffer-name))
+            (lsp)))))))
 
 (defun lsp--update-key (table key fn)
   "Apply FN on value corresponding to KEY in TABLE."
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -5322,6 +5328,42 @@ argument ask the user to select which language server to start. "
                (apply 'concat (--map (format "[%s]" (lsp--workspace-print it))
                                      lsp--buffer-workspaces)))))
 
+(defvar lsp--emacs-started nil
+  "A boolean value indicating whether Emacs has finished starting.")
+
+(add-hook 'emacs-startup-hook #'lsp--emacs-startup)
```
yyoncho: can this be part of use-package configuration? AFAICS if this hook is attached after emacs has started lsp--emacs-started will never be set to t. Can we document that if you want to use this functionality you will have to eagerly load lsp-mode on startup or you should manually call `lsp--emacs-startup` in your config (if this is the case it should be public)?

lsp-mode.el
```diff
@@ -5322,6 +5328,42 @@ argument ask the user to select which language server to start. "
...
+(defun lsp--emacs-startup ()
+  "Run `lsp' for deferred buffers that are visible."
+  (setq lsp--emacs-started t)
+  ;; Now that Emacs is done starting, check if there are visible buffers
+  ;; that were deferred during startup.
+  (dolist (buf (buffer-list))
+    (with-current-buffer buf
+      (when lsp--buffer-deferred (lsp--if-visible)))))
+
+(defun lsp--if-visible ()
```
yyoncho: can you rename it to something like lsp--init-if-visible

lsp-mode.el
muirrn: Yes, you are right of course. I had some stupid config that was always loading lsp-mode. Anyway I found that it is desktop-mode itself that calls `switch-to-buffer` for each buffer as it is restoring the desktop file. That triggers `window-configuration-change-hook` and gives us false positives during startup.

I'd like to find some way to detect this so users don't have to configure anything other than `lsp-deferred`. Can you think of a good way to tell the difference between `lsp-deferred` being called via desktop-read and other cases? So far my only idea is to add a timer in `lsp--init-if-visible` so it only calls `lsp` if the buffer is still visible after 0.1 seconds (but I don't really like that idea).

yyoncho: I checked the code in startup.el and it seems like we may use `inhibit-startup-hooks` to check whether the startup hook has already been run.

muirrn: I noticed `inhibit-startup-hooks`, too, but it doesn't actually get set for me after starting up.

muirrn: I think I found a better solution. Setting a timer for 0 seconds seems to avoid the false positives from desktop mode.

yyoncho: Then the only solution that I could thick of is to ask the users to use emacs-startup-hook when lsp-mode is loaded on startup but call the lsp-mode `lsp--emacs-startup` explicitly if lsp-mode is lazy loaded.

lsp-mode.el
muirrn: Done.

lsp-mode.el
```diff
@@ -5322,6 +5328,30 @@ argument ask the user to select which language server to start. "
...
+`lsp-deferred' will wait until the buffer is visible before invoking `lsp'.
+This avoids overloading the server with many files when starting Emacs."
+  ;; Workspace may not be initialized yet. Use a buffer local variable to
+  ;; remember that we deferred loading of this buffer.
+  (setq lsp--buffer-deferred t)
+  (let ((buffer (current-buffer)))
+    ;; Avoid false positives as desktop-mode restores buffers by deferring
+    ;; visibility check until the stack clears.
+    (run-with-timer 0 nil (lambda ()
+                            (with-current-buffer buffer
```
yyoncho: need to check whether the buffer is still alive.

muirrn: Done

PR #913, opened 2019-07-01
Change `lsp-print-io` to `lsp-log-io` in README.org

LEXUGE:
Due to the obsolete of `lsp-print-io`, I supplanted it by `lsp-log-io` in order to prevent the `README.org` from outdated version


yyoncho:
Thank you!

PR #918, opened 2019-07-04
Prevent void-function call when rendering markdown

redcap1:
Hi there,

I ran into an issue with lsp-mode not rendering markdown properly. Not much of an elisp developer, but i traced it down to a call of gfm-view-mode which is not provided by markdown v 2.3.


yyoncho:

Issue #914, opened 2019-07-01
Automatic and irreversible switch to `*Warnings*` buffer when pyls (Python language server) returns certain errors


gcv:
**Describe the bug**

The following source file causes `lsp-mode` to switch to the `*Warnings*` buffer when the cursor is moved on the line containing `self.study_name` and makes it impossible to go back to the source buffer unless the pyls connection is killed:

```python
class TestClass:

    def initialize_optuna(self):
        try:
            train_env = Alpha(
                [lambda: Bravo(self.feature_df)])
            model = self.Model(self.Policy, train_env, nminibatches=1)
            self.study_name = f'{model.__class__.__name__}__{model.act_model.__class__.__name__}__{self.reward_strategy}'
        except:
            self.study_name = f'UnknownModel__UnknownPolicy__{self.reward_strategy}'
```

**To Reproduce**

I use Tramp remote lsp-mode integration, but I suspect the error will occur on local pyls installations also.

Make a root project containing the source file above (I called it `test.py`). Connect to `pyls` with `lsp-mode` (vanilla configuration). Position the cursor on the `study_name` identifier on line 8. It instantly dumps me into the `*Warnings*` buffer.

**Expected behavior**

Ideally, the error goes away altogether, but that might be a pyls problem. At minimum, `lsp-mode` should not autodump me back in the `*Warnings*` buffer.

**Which Language Server did you use**

python-language-server 0.27.0, used through `lsp-mode`'s Tramp remote connection. `pip list` output is attached to make the environment reproducible.

**OS**

Emacs 26.2 on macOS, pyls runs on a Linux docker container:

```
$ uname -a
Linux mlc3 4.9.125-linuxkit #1 SMP Fri Sep 7 08:20:28 UTC 2018 x86_64 GNU/Linux
```

**Error callstack**

The `*Warnings*` and `*lsp-log*` buffer contents are attached. The error seems to be related to pyls blowing up with `'FStringStart' object has no attribute 'children'` error.

[lsp-log.txt](https://github.com/emacs-lsp/lsp-mode/files/3346092/lsp-log.txt)
[pip-list.txt](https://github.com/emacs-lsp/lsp-mode/files/3346093/pip-list.txt)
[warnings.txt](https://github.com/emacs-lsp/lsp-mode/files/3346094/warnings.txt)

yyoncho:
Thanks for the detailed bug report, can you share your pyls configuration(the remote one)?

gcv:
[...]

PR #919, opened 2019-07-04
Update documentation for TRAMP configuration wrt stderr redirection.

gcv:
Documentation to help the next person who hits the problem I reported here: https://github.com/emacs-lsp/lsp-mode/issues/914


yyoncho:
Thank you, let's keep the original issue open so someone might try to do the stderr handing automatically.

PR #920, opened 2019-07-05
Implement auto-installation of fsharp language server

TOTBWF:
This PR implements auto-installation of the F# language server.


lsp-fsharp.el
```diff
@@ -57,15 +66,39 @@
...
+  (let ((fsac (lsp-fsharp--fsac-cmd)))
+    (unless (file-exists-p fsac)
+      (if (yes-or-no-p "Server is not installed. Do you want to install it?")
+          (lsp-fsharp--fsac-install)
+        (error "LSP F# cannot be started without FsAutoComplete Server")))
+  fsac))
+
+(defun lsp-fsharp--fsac-install ()
+  "Downloads the latest version of fsautocomplete, and set `lsp-fsharp-server-path'."
+  (let ((temp-file (make-temp-file "fsautocomplete" nil ".zip")))
```
yyoncho: Take a look at https://github.com/emacs-lsp/lsp-python-ms/pull/37 which also handles unzipping but it fallbacks to powershell on windows. We probably need to push up that logic in `lsp-mode` and then reuse it in `lsp-python-ms` and `dap-mode`.

lsp-fsharp.el
```diff
@@ -77,6 +110,7 @@
                   :major-modes '(fsharp-mode)
                   :notification-handlers (ht ("fsharp/notifyCancel" #'ignore)
```
yyoncho: not part of this review but you may take a look how do we handle the notifications in `lsp-rust` and `lsp-java` and implement the same logic here.

README.org
```diff
@@ -121,44 +121,44 @@
    ~lsp-clients.el~ while others require installing additional packages which provide
    server specific functionality.
 
-   | Language              | Language Server                           | Built-in       | Installation command                                                                                                 | Debugger                     |
```
yyoncho: The formatting changes are caused by the fact that one is performing the formatting with links display enabled and the other is performing the alignment with link display disabled(use `org-toggle-link-display`). Can you do the formatting so we will have only the required changes?

TOTBWF: Ah, nice catch. Fixed!

lsp-fsharp.el
TOTBWF: Good point, fixed.

TOTBWF:
I've been hacking on `dap-mode` support on and off, but it's a bit of a tricky situation. The debugger that ships with the C# extension for VSCode has a weird license, which only allows for usage inside of Microsoft products. There is [netcoredbg](https://github.com/Samsung/netcoredbg), but I've been experiencing some odd behavior with it.

lsp-fsharp.el
TOTBWF: Maybe I'm missing something really obvious, but they both look rather similar to what I'm doing here. Could you point me to the code that you are referencing?

yyoncho: Here it is: https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-rust.el#L296 (although I am not 100% sure that the FSharp notifications are progress notifications).

PR #924, opened 2019-07-05
pyls rope completion is disabled by default

### LINKED ISSUE #923: Rope completion in pyls should be False by default
I am not sure on how the settings sync between rope and pyls, but according to pyls code, rope completion should be disabled by default (https://github.com/palantir/python-language-server/blob/develop/pyls/plugins/rope_completion.py#L14  it is also very slow and trashy), so `lsp-pyls-plugins-rope-completion-enabled` should be nil?

mpanarin:
Closes #923 
according to https://github.com/palantir/python-language-server/commit/af4dcb325d70629435456ce253275edbb84877e1
rope completion should be disabled by default. As well as Jedi completion
works generally faster.

### DIFF
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -256,7 +256,7 @@ dot."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.1"))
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

PR #929, opened 2019-07-09
Pipe stderr for tramp connections

### LINKED ISSUE #914: Automatic and irreversible switch to `*Warnings*` buffer when pyls (Python language server) returns certain errors

yyoncho:
- Fixes #914

- when using lsp over tramp stderr and stdout are not separated, so pipe stderr
  locally to avoid breaking the parser

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -345,6 +345,8 @@ This flag affects only server which do not support incremental update."
   :group 'lsp-mode
   :package-version '(lsp-mode . "6.1"))
 
+(defvar lsp--stderr-index 0)
+
 (defvar lsp--delayed-requests nil)
 (defvar lsp--delay-timer nil)
 
@@ -4704,15 +4706,22 @@ should return the command to start the LS server."
                 (cons tcp-client-connection cmd-proc)))
    :test? (lambda () (executable-find (cl-first (funcall command-fn 0))))))
 
-(defun lsp-tramp-connection (local-command)
+(defun lsp-tramp-connection (local-command &optional generate-error-file-fn)
   "Create LSP stdio connection named name.
 LOCAL-COMMAND is either list of strings, string or function which
 returns the command to execute."
   (list :connect (lambda (filter sentinel name)
                    (let* ((final-command (lsp-resolve-final-function local-command))
                           ;; wrap with stty to disable converting \r to \n
-                          (wrapped-command (append '("stty" "raw" ";") final-command))
-                          (process-name (generate-new-buffer-name name)))
+                          (process-name (generate-new-buffer-name name))
+                          (wrapped-command (append '("stty" "raw" ";")
+                                                   final-command
+                                                   (list
+                                                    (concat "2>"
+                                                            (or (when generate-error-file-fn
+                                                                  (funcall generate-error-file-fn name))
+                                                                (format "/tmp/%s-%s-stderr" name
+                                                                        (cl-incf lsp--stderr-index))))))))
                      (let ((proc (apply 'start-file-process-shell-command process-name
                                         (format "*%s*" process-name) wrapped-command)))
                        (set-process-sentinel proc sentinel)
--- a/README.org
+++ b/README.org
@@ -237,7 +237,7 @@
                          :server-id 'pyls-remote))
      #+END_SRC
 **** Dealing with stderr
-    With TRAMP, Emacs does not have an easy way to distinguish stdout and stderr, so when the underlying LSP process writes to stderr, it breaks the ~lsp-mode~ parser. As a workaround, redirect stderr to ~/dev/null~ in the shell code, e.g. use the following ~:new-connection~ argument in the ~make-lsp-client~ call: ~(lsp-tramp-connection "/path/to/server 2>/dev/null")~.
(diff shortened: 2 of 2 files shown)


lsp-mode.el
```diff
@@ -4690,8 +4690,10 @@ returns the command to execute."
   (list :connect (lambda (filter sentinel name)
                    (let* ((final-command (lsp-resolve-final-function local-command))
                           ;; wrap with stty to disable converting \r to \n
-                          (wrapped-command (append '("stty" "raw" ";") final-command))
-                          (process-name (generate-new-buffer-name name)))
+                          (process-name (generate-new-buffer-name name))
+                          (wrapped-command (append '("stty" "raw" ";")
+                                                   final-command
+                                                   (list (format "2>%s" (make-temp-file process-name))))))
```
gcv: `make-temp-file` produces a tempfile specific to the Emacs host, but the LSP server runs remotely. Edebug shows the value of `wrapped-command` at this point is:

```
("stty" "raw" ";" "pyls" "2>/var/folders/2w/fgshrnj48x01s7pr7x00010gn/T/pyls-remote4kwB7c")
```

That path (almost certainly) does not exist on the LSP server host, which causes the attempt to start the LSP server to fail on startup.

yyoncho: Yep, it will be even more problematic if you do tramp Windows->Linux. Updated PR to use /tmp as a default path and I added an option to specify a function to generate the temp fine on the remote machine.

lsp-mode.el
```diff
@@ -4689,15 +4689,21 @@ should return the command to start the LS server."
...
-                          (wrapped-command (append '("stty" "raw" ";") final-command))
-                          (process-name (generate-new-buffer-name name)))
+                          (process-name (generate-new-buffer-name name))
+                          (wrapped-command (append '("stty" "raw" ";")
+                                                   final-command
+                                                   (list
+                                                    (concat "2>"
+                                                            (or (when generate-error-file-fn
+                                                                  (funcall generate-error-file-fn name))
+                                                                (format "/tmp/%s-stderr" name)))))))
```
gcv: We should try a little harder to prevent these stderr files from stomping on each other. I suggest something like this instead:

```
(format "/tmp/%s-%04x%04x-stderr" name (random (expt 16 4)) (random (expt 16 4)))
```

It just adds a poor man's uuid to the filename. It's completely portable and works well enough in practice.

lsp-mode.el
gcv: Cool. With your latest update, the code works. I suggested a change to make the stderr redirect files more unique.

lsp-mode.el
yyoncho: Thank you, I did it via simple counter because I am not quite sure whether the `%04x` format identifier works on the older emacs versions(we recently had a bug about that).

yyoncho:
Updated incorporating the latest comment.

lsp-mode.el
gcv: Seems like it should work. 👍

gcv:
It might be good to remove the now-outdated "Dealing with stderr" section from the README before merging this branch.

yyoncho:
> It might be good to remove the now-outdated "Dealing with stderr" section from the README before merging this branch.

Thank you, fixed and merged.

PR #936, opened 2019-07-13
Handle the metals/executeClientCommand extension notification.

kurnevsky:
This will allow to go to the definition from the tree view and also will show doctor html in emacs buffer.
screenshot-2019-07-13-19:49:30


yyoncho:
@dsyzling - would you test the PR?

kurnevsky:
@dsyzling note that metals has hardcoded settings for emacs that should be changed. Or you can just specify vscode option when installing metals :)

lsp-metals.el
```diff
@@ -109,11 +109,54 @@ more customizations like using environment variables."
...
+  "Move the cursor focus to the provided location."
+  (let ((xrefs (lsp--locations-to-xref-items (list location))))
+    (if (boundp 'xref-show-definitions-function)
+      (with-no-warnings
+        (funcall xref-show-definitions-function
+          (-const xrefs)
+          `((window . ,(selected-window)))))
+      (xref--show-xrefs xrefs nil))))
+
+(defun lsp-metals--execute-client-command (workspace params)
```
yyoncho: to remove the duplication, you may structure the code like that:
```
(when-let (command (pcase ....
                                     ("metals-doctor-run" 'lsp-metals--doctor-run)
                                      ....
                                   ))
       (funcall command workspace (ht-get ...)))
```

kurnevsky:
Also just found out that it can be configured via cli: https://scalameta.org/metals/docs/editors/new-editor.html#dmetalsexecute-client-command
But for some reason if both this setting and `metals.http` are turned on metals doesn't send client commands.

dsyzling:
Sorry guys for the late reply - been away for a week. I thought I could customise the lsp-metals-server-args and override the metals.client along with metals.execute-client-command=on. I tried combinations including -Dmetals.http=off but couldn't seem to get this to work. I triggered a new download of the metals binary and removed the http.client=emacs whilst adding -Dmetals.execute-client-command=on and that seemed to do the trick.

On my Arch installation the ticks do not render - I may be missing a font icon package. This is what the doctor screen looks like - by default the doctor screen appears in a split window so the column titles have word wrapped:
Metals-Doctor

yyoncho:
@dsyzling are we ok to merge it as it is? The issue with the symbols not rendering seems to be environment-specific.

dsyzling:
@yyoncho yes seems good to me - I'll take a look a little later to see if I can track down why the ticks aren't rendering in case someone else has the same problem - no doubt I won't be the last.

yyoncho:
@dsyzling ok, thank you.

PR #947, opened 2019-07-18
Add support for Erlang Language Server

robertoaloi:
This PR introduces support for the Erlang Language Server.

A few gotchas:

* The Erlang Language Server is a very early prototype, far from be useful at this stage, but this PR simplifies the interaction with it, making the development workflow for the Server nicer
* It's not clear to me why certain languages are under `lsp-clients`, whilst others have their own `lsp-LANGUAGE.el` file. I opted for the latter, but please advise if I should do something different
* I'm not familiar with either elisp or this project, so please advise if something is incorrect/missing


yyoncho:
Thank you!

> It's not clear to me why certain languages are under lsp-clients, whilst others have their own lsp-LANGUAGE.el file. I opted for the latter, but please advise if I should do something different

We initially had all of the client definitions in single fine but it turned out problematic when the definition became bigger so now we split them in separate files.

PR #958, opened 2019-08-01
[fix] wrong default value in pylint-args

mpanarin:
Mistakenly put wrong value for pylint-params in pyls

### DIFF
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -151,7 +151,7 @@ complexity."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.1"))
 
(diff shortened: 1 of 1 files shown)


lsp-pyls.el
```diff
@@ -151,7 +151,7 @@ complexity."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-pyls-plugins-pylint-args t
```
yyoncho: it should be [] if you want to pass an empty array. () is the same as nil.

mpanarin: nil was okay as well, but this makes sense. Fixed

PR #959, opened 2019-08-05
Fix misleading doc string

juergenhoetzel:


juergenhoetzel:
`lsp-fsharp-server-path` isn't set

PR #968, opened 2019-08-12
Update lsp-elm

Daniel-V1:
Remove treesitter-runtime option as that is now loaded with the server as a wasm module. Add diagnosticsOnSaveOnly option

----

#


yyoncho:

PR #974, opened 2019-08-15
Add customisation option to run solargraph with bundle exec

codeasone:
Reference: https://github.com/castwide/solargraph#solargraph-and-bundler


yyoncho:

PR #976, opened 2019-08-19
Use get-file-buffer instead of find-buffer-visiting

muirrn:
Except on Windows, use get-file-buffer instead of
find-buffer-visiting. find-buffer-visiting was resulting in 500ms+
completion latency for me in some projects. In particular, after every
edit, my LSP server (gopls) was sending many diagnostic notifications,
and each one of them was spending 20ms in find-buffer-visiting.

Based on previous similar
change (3243cb2263ebac8921eebbd707fe0d7d0f739883) and subsequent
reversion (019e99672249535e440cbd00bb10fd228d1afa9c) it seems like
get-file-buffer can be used instead of find-buffer-visiting, except on
Windows. Perhaps the filename from the URI can be transformed so
get-file-buffer works on Windows as well, but I don't have a Windows
machine right now, so preserve existing functionality for Windows.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1321,6 +1321,11 @@ PARAMS - the data sent from WORKSPACE."
      :message (if source (format "%s: %s" source message) message)
      :original diag)))
 
+(defalias 'lsp--buffer-for-file (if (eq system-type 'windows-nt)
+                                  #'find-buffer-visiting
+                                #'get-file-buffer))
+
+
 (defun lsp--on-diagnostics (workspace params)
   "Callback for textDocument/publishDiagnostics.
 interface PublishDiagnosticsParams {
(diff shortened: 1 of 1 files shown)


muirrn:
lsp-mode performance is abysmal in some projects for me due to this issue, so I think we should address it one way or another. Let me know if you have any better ideas.

yyoncho:
> Perhaps the filename from the URI can be transformed so
get-file-buffer works on Windows as well, but I don't have a Windows
machine right now, 

This does not seem to be feasible: consider the fact that in windows C:\foo\bar is the same as C:\FOO\BAR and the buffer-file-name depend on the case you have used when opening the file.

I was thinking about the following solution: reimplement get-file-buffer but make it work only for lsp-mode managed buffers. E. g: 
1. Keep a mapping file-name->buffer when lsp-mode is enabled.
2. Implement lsp-get-file-buffer which looks up into that map. On windows - downcase the files before writing and reading.

PR #981, opened 2019-08-21
Fix invalid quote of defcustom const value

juergenhoetzel:
While the default value is correct:
```
ELISP> lsp-fsharp-server-runtime
net-core
ELISP> (type-of lsp-fsharp-server-runtime)
symbol
```

Trying to set `(customize-set-variable lsp-fsharp-server-runtime)` to `Mono` results in

```
ELISP> lsp-fsharp-server-runtime
'mono
ELISP> (type-of lsp-fsharp-server-runtime)
cons
``` 
 
And thus failing to start the server


yyoncho:
Thank you!

PR #989, opened 2019-08-27
Provide path to binary instead of install dir

jfacorro:
@robertoaloi

### DIFF
--- a/lsp-erlang.el
+++ b/lsp-erlang.el
@@ -31,15 +31,15 @@
   :group 'lsp-mode
   :link '(url-link "https://github.com/erlang-ls/erlang_ls"))
 
-(defcustom lsp-erlang-server-install-dir
-  "."
-  "Path to the Erlang Language Server installation dir."
+(defcustom lsp-erlang-server-path
+  "erlang_ls"
+  "Path to the Erlang Language Server binary."
   :group 'lsp-erlang
   :risky t
-  :type 'directory)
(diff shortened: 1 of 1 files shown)


lsp-erlang.el
```diff
@@ -31,15 +31,15 @@
   :group 'lsp-mode
   :link '(url-link "https://github.com/erlang-ls/erlang_ls"))
 
-(defcustom lsp-erlang-server-install-dir
-  "."
-  "Path to the Erlang Language Server installation dir."
+(defcustom lsp-erlang-server-path
```
robertoaloi: I like the change, but please update the README at https://github.com/erlang-ls/erlang_ls/ to reflect the new variable.

robertoaloi: Also, thanks for the explicit ping on the PR. Is there any way to be automatically pinged if this file is modified?

yyoncho: I guess you may subscribe to all notifications and filter only the notifications containing the filename.

robertoaloi: @yyoncho This sounds like a more powerful alternative: https://help.github.com/en/articles/about-code-owners

robertoaloi: @jfacorro I updated the README: https://github.com/erlang-ls/erlang_ls/commit/6477ec23b75e992f2dec5db39c8898be0bcc38d7

PR #996, opened 2019-09-01
Add support for hover to metals

DhashS:
I think this adds support for hover to metals. New to emacs, so please give this a look to see if it's correct. This is based off my copy-and-paste from the rust hover.

### DIFF
--- a/lsp-metals.el
+++ b/lsp-metals.el
@@ -78,12 +78,20 @@ more customizations like using environment variables."
   :group 'lsp-metals
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-metals-show-hover-context t
+  "Show additional context in hover tooltips when available. This
+is often the type local variable declaration."
+  :type 'boolean
+  :group 'lsp-metals
+  :package-version '(lsp-mode . "6.1"))
+
 
 (lsp-register-custom-settings
  '(("metals.java-home" lsp-metals-java-home)
    ("metals.scalafmt-config-path" lsp-metals-scalafmt-config-path)
    ("metals.sbt-script" lsp-metals-sbt-script)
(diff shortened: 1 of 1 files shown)


lsp-metals.el
```diff
@@ -78,12 +78,20 @@ more customizations like using environment variables."
   :group 'lsp-metals
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-metals-show-hover-context t
+  "Show additional context in hover tooltips when available. This
+is often the type local variable declaration."
+  :type 'boolean
+  :group 'lsp-metals
+  :package-version '(lsp-mode . "6.1"))
```
yyoncho: 6.1 was released so change it to 6.2

kurnevsky:
There is no such option neither in [docs](https://scalameta.org/metals/docs/editors/new-editor.html#metals-user-configuration) nor in [src](https://github.com/scalameta/metals/search?q=show-hover-context). @DhashS where did you find it and what is it supposed to do?

dsyzling:
Agree with @kurnevsky - @DhashS have you tested this because we can't see this option available within the Metals source code base?

DhashS:
Spent some time last week reading into metals, it looks like all the flags are set [here](
https://github.com/scalameta/metals/blob/9ded96c368ec4197047586b41adb18dac5de82c0/metals/src/main/scala/scala/meta/internal/metals/MetalsServerConfig.scala). I now have some idea of what configuration to pass to metals, though it seems like i'll have to get a PR into metals first. I'm working on this on and off on the weekends, but i'll keep updating this PR as I get stuff working

yyoncho:
@DhashS I think that the hover information for which we use eldoc/lsp-ui works by default for scala.

yyoncho:
Closing the PR based on the fact that the hover works out of the box when using metals server.

PR #997, opened 2019-09-02
add custom variable to control skip hover on empty line

ezchi:
One of my LSP server apps will crash if received hover request on empty line. This change will add the feature to skip sending hover request to LSP server on empty lines.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3694,9 +3694,8 @@ RENDER-ALL - nil if only the signature should be rendered."
   (if (and lsp--hover-saved-bounds
            (lsp--point-in-bounds-p lsp--hover-saved-bounds))
       (lsp--eldoc-message lsp--eldoc-saved-message)
-    (let* ((whitespace-or-newline (looking-at "[[:space:]\n]"))
-           (skip-hover (and lsp-hover-skip-whitespace-or-newline whitespace-or-newline)))
-      (unless skip-hover
+    (let* ((whitespace-or-newline (looking-at "[[:space:]\n]")))
+      (unless whitespace-or-newline
         (setq lsp--hover-saved-bounds nil
               lsp--eldoc-saved-message nil)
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for the PR. What is the language server that is crashing? I have some concerns with the approach (only on new lines). I think that it will be better to change the code to run only if it is over a non-space. What do you think?

ezchi:
Ivan, do you mean you expect to skip over any whitespaces, tabs and newlines in anywhere?

yyoncho:
@ezchi no. I mean instead of checking the line content we could check what is under the cursor. If it is a symbol(letter, punctuation) - then go ahead and request onHover. If it is space do not run it.

lsp-mode.el
```diff
@@ -3686,47 +3691,50 @@ RENDER-ALL - nil if only the signature should be rendered."
...
-             (when hover
-               (when-let (range (gethash "range" hover))
-                 (setq lsp--hover-saved-bounds (lsp--range-to-region range)))
-               (-let (((&hash "contents") hover))
-                 (when-let (message
-                            (and contents (lsp--render-on-hover-content contents lsp-eldoc-render-all)))
-                   (when (or (and (not lsp-eldoc-prefer-signature-help) (setq pending 1))
+  (let* ((whitespace-or-newline (looking-at "[[:space:]\n]"))
+         (skip-hover (and lsp-hover-skip-whitespace-or-newline whitespace-or-newline)))
+    (unless skip-hover
```
yyoncho: what happens if you move from identifier to `space`? I think that it will preserve the original value so you should have else leg and call lsp--eldoc-message with an empty message.

ezchi:
@yyoncho New commit is coming and hope I understand you correctly.

yyoncho:
I tested the last version - can you make sure that when you switch from something with eldoc to empty space the old eldoc message is cleared?

yyoncho:

I merged the PR and did small adjustments to make sure the eldoc message is cleared right away.

PR #998, opened 2019-09-02
[untested] Add basic support for clangd-style semantic highlighting

sebastiansturm:
Hi,

this PR is meant to be a basis for discussion, though it already seems to work pretty well on my machine (using Emacs 27 trunk, clang/clangd 10 also built from trunk). Some real and potential disadvantages I'm aware of:

- fontification is done exclusively through overlays (as opposed to emacs-ccls, which allows the user to choose between overlays and `font-lock`-based fontification). I didn't notice any speed issues, but older Emacs versions may not fare so well (cf. https://lists.gnu.org/archive/html/emacs-devel/2018-03/msg00886.html) (BTW, does anyone know what happened to the overlay branch? I thought it had been merged a long time ago but was unable to find evidence for that)

- highlights pushed by the server are always applied, even if the buffer in question isn't currently visible. Again, currently not a problem for me but may be a problem for others or with other language servers

- `lsp--facemap` should probably be a per-server thing, not per-buffer (?)

- so far I haven't tried to optimize the code; especially that `bindat` thing is so convenient I wonder how much I'm paying for that. That being said, the initial fontification run on a ~1000-line C++ file takes about 23ms on my 2014 MacBook Pro; an autogenerated 10000-line file takes ~250ms. Subsequent updates are incremental and do not introduce perceptible delay

- I believe the highlighting protocol is still in flux. Since (AFAIK) there's no currently released version of clangd with support for semantic highlighting, I'd suggest to have lsp-mode track clangd trunk builds as closely as possible

Also, this is the first time I've tried to extend `lsp-mode` in any way and I'm amazed how easy it was to add a useful feature, even if my approach may violate some best practices or style guides. Many thanks to the lsp authors!

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -46,6 +46,7 @@
 (require 'json)
 (require 'network-stream)
 (require 'pcase)
+(require 's)
 (require 'seq)
 (require 'spinner)
 (require 'subr-x)
@@ -172,6 +173,14 @@ the buffer when it becomes large."
   :group 'lsp-mode
   :type 'boolean
   :package-version '(lsp-mode . "6.1"))
+
+(defcustom lsp-enable-semantic-highlighting nil
+  "Enable/disable semantic highlighting as proposed at
+ https://github.com/microsoft/vscode-languageserver-node/pull/367.
+This feature is not yet part of the official LSP spec and may
+occasionally break as language servers are updated."
+  :group 'lsp-mode
+  :type 'boolean)
 
 (defcustom lsp-folding-range-limit nil
   "The maximum number of folding ranges to receive from the language server."
@@ -2375,7 +2384,7 @@ disappearing, unset all the variables related to it."
                                         (hierarchicalDocumentSymbolSupport . t)))
                      (formatting . ((dynamicRegistration . t)))
                      (rename . ((dynamicRegistration . t)))
-                     (semanticHighlightingCapabilities . ((semanticHighlighting . t)))
+                     (semanticHighlightingCapabilities . ((semanticHighlighting . ,lsp-enable-semantic-highlighting)))
                      (codeAction . ((dynamicRegistration . t)
                                     (codeActionLiteralSupport . ((codeActionKind . ((valueSet . [""
                                                                                                  "quickfix"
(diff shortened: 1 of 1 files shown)


yyoncho:
Welcome to the project!

> Since (AFAIK) there's no currently released version of clangd with support for semantic highlighting, 

AFAIK JDT LS has an implementation of the feature so we could use it for testing.

One question - is semantic highlight full replacement of font-lock-mode? 

> highlights pushed by the server are always applied, even if the buffer in question isn't currently visible. Again, currently not a problem for me but may be a problem for others or with other language servers

I guess the alternative solution is to store the information somewhere and write custom font-lock-fontify-region-function but given the performance numbers, I think it will be better to work on json serialization optimization if we are looking for speedup. 

I would propose the following action plan:
1. Make the PR work with the other implementation of semantic highlight (JDT LS) which means we should revisit lsp-semantic-highlighting-faces.
2. Disable it by default and mark it as a experimental feature. 

What do you think?

sebastiansturm:
> One question - is semantic highlight full replacement of font-lock-mode?

almost, but not quite. Some very basic but important elements are not covered by semantic highlighting, such as strings, comments, or keywords such as `const`, `for`, or `while`.

> > highlights pushed by the server are always applied, even if the buffer in question isn't currently visible. Again, currently not a problem for me but may be a problem for others or with other language servers
> 
> I guess the alternative solution is to store the information somewhere and write custom font-lock-fontify-region-function but given the performance numbers, I think it will be better to work on json serialization optimization if we are looking for speedup.
> 
> I would propose the following action plan:
> 
> 1. Make the PR work with the other implementation of semantic highlight (JDT LS) which means we should revisit lsp-semantic-highlighting-faces.
> 2. Disable it by default and mark it as a experimental feature.
> 
> What do you think?

sounds great---the sooner I can switch back to canonical `lsp-mode` instead of using my own fork, the better for me. I'm not using Java at all though, so I'm not well qualified to try out the JDT language server

yyoncho:
> I'm not using Java at all though, so I'm not well qualified to try out the JDT language server

I tested it with replacing cpp -> java in your PR. It seems like it is working fine. 

Faces: In your PR you are reusing faces for different keys. I think that we should change that. I would suggest making all static variants to be italic.

lsp-mode.el
```diff
@@ -3905,6 +3907,79 @@ A reference is highlighted only if it is visible in a window."
                       (push overlay buf-overlays)
                       (puthash (current-buffer) buf-overlays overlays))))))))))))
 
+(defvar lsp-semantic-highlighting-faces
```
yyoncho: As I already mentioned the code should be changed so the suffix("cpp") is not part of the keys.

lsp-mode.el
```diff
@@ -2374,6 +2375,7 @@ disappearing, unset all the variables related to it."
                                         (hierarchicalDocumentSymbolSupport . t)))
                      (formatting . ((dynamicRegistration . t)))
                      (rename . ((dynamicRegistration . t)))
+                     (semanticHighlightingCapabilities . ((semanticHighlighting . t)))
```
yyoncho: Please introduce a variable. The default value should be false for some time.

lsp-mode.el
```diff
@@ -3905,6 +3907,79 @@ A reference is highlighted only if it is visible in a window."
...
+  (unless lsp--facemap
+    (let* ((capabilities (lsp--workspace-server-capabilities workspace))
+           (semanticHighlighting (gethash "semanticHighlighting" capabilities))
+           (scopes (or (gethash "scopes" semanticHighlighting) [])))
+      (setq lsp--facemap
+            (mapcar #'lsp--semantic-highlighting-find-face scopes))))
+  (let* ((file (lsp--uri-to-path (gethash "uri" (gethash "textDocument" params))))
+         (lines (gethash "lines" params))
+         (buffer (lsp--buffer-for-file file))
+         (t0 (current-time)))
```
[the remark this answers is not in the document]
sebastiansturm: ah yes, used that for benchmarking. Will include your suggested changes this evening when I'm back home, thanks!

yyoncho:
And one more thing - it will be great if you update the readme mentioning the new functionality and a screenshot before/after.

yyoncho:
Here are some elements that are not defined in the faces list(and probably there will be a lot more):

```
(invalid.deprecated.java meta.class.java source.java variable.other.autoboxing.java meta.method.body.java meta.method.java meta.class.body.java storage.modifier.static.java storage.modifier.final.java variable.other.definition.java meta.definition.variable.java 
meta.function.call.java constant.other.key.java meta.declaration.annotation.java storage.type.generic.java meta.definition.class.implemented.interfaces.java 
storage.type.annotation.java entity.name.type.interface.java meta.class.identifier.java constant.numeric.decimal.java keyword.other.var.java)
```

So I would propose to log when the there is no face for the key and skip it instead of using the warning face. I think that even when the feature is implemented there still will be scopes that are not mapped to face.

sebastiansturm:
> And one more thing - it will be great if you update the readme mentioning the new functionality and a screenshot before/after.

sure, will do

lsp-mode.el
```diff
@@ -3905,6 +3916,178 @@ A reference is highlighted only if it is visible in a window."
...
+  :group 'lsp-faces)
+
+(defface lsp-face-semhl-type-primitive
+  '((t :inherit font-lock-type-face :italic t))
+  "Face used for semantic highlighting scopes matching storage.type.primitive.*,
+unless overriden by a more specific face association."
+  :group 'lsp-faces)
+
+(defvar lsp-semantic-highlighting-faces
+  '(("^variable\\.parameter\\(\\..*\\)?$" . lsp-face-semhl-variable-parameter)
```
yyoncho: Is there something that forces us to use regex but not prefix and then test with starts-with? My main concern is that the regexp will be much slower.

sebastiansturm: regexp matching is only done on the first fontification run (within the `(unless lsp--facemap...` part in `lsp--on-semantic-highlighting`), and only needs to traverse the `scopes` vector sent by the language server (may have a dozen members or so).
For the fontification itself, faces are retrieved directly from `lsp--facemap` using `scopeIndex`

yyoncho: Fair enough.

yyoncho:

sebastiansturm:
thanks for your help and the quick merge!

Issue #966, opened 2019-08-11
How to use/register @angular/language-service for angular project.


JSiebelink:
**Describe the bug**
For Angular projects, it should be possible to use the official @angular/language-service for error diagnostics, go to definition functionality, auto completion and such, when opening html templates related to typescript components/files. For the official language service, see: https://angular.io/guide/language-service and for the vscode implementation of the service: https://github.com/angular/vscode-ng-language-service

**To Reproduce**
- Clone: https://gitlab.com/Jorrieporrie/lsp-mode-angular-service
- Run npm install in root folder
- Open app.component.ts to see the component's controller with a variable called 'typoVar' and a method called 'aMethodTheTemplateShouldFind'
- Open app.component.html to see the variable and the method correctly and incorrectly referenced in the top h1 tag.

**Expected behavior**
There should be diagnostics stating that 'tyypoVar' and 'aMethooodTheTemplateShouldFind()' are not defined.

**OS**
MacOS

yyoncho:
I think that at this point `vscode-ng-language-service` supports only node ipc connections(cc @kyliau to confirm). So the first step to make this support work is to add stdio support.

kyliau:
the LSP server for Angular language service uses the standard vscode-languageserver package.
If you start the server with `--stdio` flag, it'd read from stdin and stdout.
See https://github.com/angular/vscode-ng-language-service/blob/2558440c8fb9a14ee7aa927777aa3be5332829b7/server/src/server.ts#L17 and https://github.com/microsoft/vscode-languageserver-node/blob/e9f3e3cdfb0fb761654bed9154e248acade7417c/server/src/main.ts#L1691
Please let me know if you need further clarification =)

yyoncho:
@kyliau thank you. I thought that it has to be explicitly enabled. 

@JSiebelink I managed to make it work.

```
;;; Angular
(defcustom lsp-angular-language-server-command
  '("node" "/home/kyoncho/Downloads/angular/extension/server/server.js" "--stdio")
  "The command that starts the docker language server."
  :group 'lsp-angular
  :type '(choice
          (string :tag "Single string value")
          (repeat :tag "List of string values"
                  string)))

(lsp-register-client
 (make-lsp-client :new-connection (lsp-stdio-connection
                                   (-const lsp-angular-language-server-command))
                  :activation-fn (lambda (&rest _args)
                                   (string-match-p ".*\.html$" (buffer-file-name)) )
                  :add-on? t
[...]

PR #1003, opened 2019-09-04
Start angular-ls only in angular projects

AloisJanicek:
If `angular.json` in project's root is reliable indicator of angular project and If `(project-roots (project-current))` works on all supported versions of emacs, this could fix issue mentioned here https://github.com/emacs-lsp/lsp-mode/issues/966#issuecomment-528068671

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -504,7 +504,7 @@ responsiveness at the cost of possibile stability issues."
  (make-lsp-client :new-connection (lsp-stdio-connection
                                    (lambda () lsp-clients-angular-language-server-command))
                   :activation-fn (lambda (&rest _args)
-                                   (when (file-exists-p (concat (car (project-roots (project-current))) "angular.json"))
+                                   (when (file-exists-p (f-join (lsp-workspace-root) "angular.json"))
                                      (string-match-p ".*\.html$" (buffer-file-name))))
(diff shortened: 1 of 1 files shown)


lsp-clients.el
```diff
@@ -504,7 +504,8 @@ responsiveness at the cost of possibile stability issues."
  (make-lsp-client :new-connection (lsp-stdio-connection
                                    (lambda () lsp-clients-angular-language-server-command))
                   :activation-fn (lambda (&rest _args)
-                                   (string-match-p ".*\.html$" (buffer-file-name)))
+                                   (when (file-exists-p (concat (car (project-roots (project-current))) "angular.json"))
```
yyoncho: you should use f-join instead of concat and lsp-workspace-root instead of project-current since project-current does not necessarily match the project-root (e. g. if user has picked interactively project root).

yyoncho: One more thing - lsp-workspace-root does not work when the project is yet to be added(e. g. you are just opening the project). In this case the angular ls wont start on the initial startup. A workaround would be the user to use lsp-workspace-folder-add before importing the project. This is a limitation of lsp-mode design - we have to first find whether there are servers matching the file and then we look for a root folder.

yyoncho:

PR #1004, opened 2019-09-04
Fix missing face attr val + prefer new attr names

dmringo:
This fixes a small bug resulting from [omitting a `t` value for the `:italic` attribute of `lsp-face-semhl-static-method`](https://github.com/emacs-lsp/lsp-mode/blob/d4b94e4629495dc102960f82c26a71d5ea6c9b2d/lsp-mode.el#L3986) and prefers `:slant` and `:weight` attributes to `:italic` and `:bold`.  The bug can manifest when [`face-attr-match-p`](https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/faces.el?h=emacs-26.3#n1777) is called with `lsp-face-semhl-static-method` (`plist-put` throws an error because the face spec fails to satisfy `plistp` since it is missing a value for the `:italic` key).  I've not made an issue for this since the bug is small and trivial to fix.

The change to use `:slant` and `:weight:` is maybe debatable, but its worth noting that `:italic t` and `:bold t` are essentially aliases for `:slant italic` and `:weight bold`, and are only kept around for backward compatibility. `:slant` and `:weight` have been available since at least Emacs 20.4 and admit a wider range of values.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -648,7 +648,7 @@ must be used for handling a particular message.")
   :group 'lsp-faces)
 
 (defface lsp-face-highlight-write
-  '((t :inherit highlight :italic t))
+  '((t :inherit highlight :slant italic))
   "Face used for highlighting symbols being written to."
   :group 'lsp-faces)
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for detailed description and congrats for your first `lsp-mode` PR! :tada:

PR #1012, opened 2019-09-05
Add customization for gopls server path

psanford:
Add custom variable 'lsp-gopls-server-path to allow you to set the
server to something other than "gopls" or to a binary that isn't in
your path.


yyoncho:
Thank you!

PR #1013, opened 2019-09-06
Add custom client capability registration.

dsyzling:
Metals requires a set of experimental client capabilities to be sent to the lsp server during initialisation to enable the treeview provider. In order to decouple the implementation within lsp-treemacs from lsp-mode a new lsp--client field allows the caller to specify additional client capabilities that can be sent along with the standard lsp-mode defined set. Currently these are simply additional capabilities appended to the existing fixed caps.


yyoncho:
Thank you.

PR #1014, opened 2019-09-07
Use featurep to check for yasnippet

jguenther:
Also clarify the grammar of the warning message (this was the original reason for this PR) but I do think that `featurep` would be better than checking if a function is bound, since the name could potentially change


yyoncho:

PR #1015, opened 2019-09-08
Match TextMate scopes against string sets instead of raw regexps

sebastiansturm:
this isn't well-tested though it seems to work with my clangd setup. With this change it should be somewhat easier to accommodate different scope sets used by different language servers


yyoncho:

PR #1017, opened 2019-09-09
Minor semantic highlighting improvements [NFC]

sebastiansturm:
this PR changes `lsp--facemap` to a vector (since its length is fixed anyway) and mentions semantic highlighting in the readme


yyoncho:

PR #1019, opened 2019-09-10
Check hosts while looking for a session folder

### LINKED ISSUE #760: Stored projects are opened in order
**Describe the bug**
When opening a project through tramp (e.g., /ssh:hostname:path/to/source.c), and adding the project to LSP, the project will be stored in ~/.emacs/.lsp-session-v1. Later, when moving to a different network, and opening a different project, lsp-mode will try to connect to the remote repo first, which may not be accessible at that time.

**To Reproduce**
1. Configure any LSP server for tramp (e.g. clangd)
2. Open a remote project that is supported by the configured server (e.g. /ssh:hostname:path/to/source.c)
3. When prompted, confirm to create the project - lsp-mode will add an entry to ~/.emacs/.lsp-session-v1
4. Open another local project (i.e. not through tramp)
5. When prompted, confirm to create the project
6. Close emacs
7. Disconnect network
8. Open emacs
9. Open the second, local project again
10. lsp-mode will go through the project list, and open them in order. If any of the projects are remote and not reachable at that time, lsp-mode will open the tramp connection, fail to do so, and then fail to open the local project

**Expected behavior**
I should be able to open local projects, even if I opened tramp projects before that may not be reachable anymore.

clangd

**OS**
Ubuntu 18.10

acowley:
This avoids the problem of attempting to connect to an unreachable
host associated with a different session. The logic is that we should
only consider sessions whose host is the same as that of the file used to
initiate the search for a relevant session.

Fixes #760

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5667,13 +5667,22 @@ Returns nil if the project should not be added to the current SESSION."
           (t nil)))
     ('quit)))
 
+(defun lsp--files-same-host (f1 f2)
+  "Predicate on whether or not two files are on the same host."
+  (or (not (or (file-remote-p f1) (file-remote-p f2)))
+      (and (file-remote-p f1) 
+           (file-remote-p f2) 
+           (string-equal (tramp-file-name-host (tramp-dissect-file-name f1)) 
+                         (tramp-file-name-host (tramp-dissect-file-name f2))))))
+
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for contributing to `lsp-mode` and congrats for your first `lsp-mode` PR :tada:!

PR #1025, opened 2019-09-14
Add support for LaTeX and plain TeX via Digestif

astoff:
See https://github.com/astoff/digestif for details on the server. It seems to work well enough, so I'm suggesting to add built-in support in lsp-mode.


yyoncho:

lsp-clients.el
```diff
@@ -511,6 +511,25 @@ responsiveness at the cost of possibile stability issues."
...
+  :link '(url-link "https://github.com/astoff/digestif/"))
+
+(defcustom lsp-clients-digestif-executable "digestif"
+  "Command to start the Digestif language server."
+  :group 'lsp-tex
+  :risky t
+  :type 'file)
+
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-stdio-connection lsp-clients-digestif-executable)
```
yyoncho: for future reference it is a bit better to wrap `lsp-clients-digestif-executable` in lambda otherwise the changes to the variable won't take effect unless lsp-clients.el is evaluated again.

astoff:
Thanks! Let me also point out an outstanding issue in company-lsp that is relevant to this server: https://github.com/tigersoldier/company-lsp/issues/118.

yyoncho:
@astoff proposed fix at https://github.com/yyoncho/lsp-mode/tree/fix-annotate

PR #1038, opened 2019-09-22
Improve wording of some user prompts

alypeng:


yyoncho:

PR #1039, opened 2019-09-23
lsp-go: support adding modules to library folders

psanford:
Recent versions of go added the concept of modules as a first class
system for managing third party dependencies. Modules are installed to
$GOPATH/pkg/mod and are set to have read-only file permissions. The
same module+version will be shared across multiple projects.

Jumping to these files should not trigger a prompt for starting a new
lsp server.

This adds logic to detect the modules directory and adds it to
library-folders automatically. This behavior can be disabled with the
customization `lsp-clients-go-library-directories-include-go-modules'.

We invoke the go tool to determine the path of the modules directory
instead of looking at the GOPATH environment variable directly. This
is necessary to determine the default GOPATH directory when the
environment variable isn't set.

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -56,8 +56,7 @@ completing function calls."
                   :major-modes '(go-mode)
                   :priority 0
                   :server-id 'gopls
-                  :library-folders-fn (lambda (_workspace)
-                                        lsp-clients-go-library-directories)))
+                  :library-folders-fn 'lsp-clients-go--library-default-directories))
 
 (defgroup lsp-clients-go nil
   "LSP support for the Go Programming Language."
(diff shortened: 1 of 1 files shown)


lsp-go.el
```diff
@@ -120,6 +119,33 @@ defaults to half of your CPU cores."
...
+If `lsp-clients-go-library-directories-include-go-modules' is non-nil
+and the environment variable GOPATH is set this function will return
+$GOPATH/pkg/mod along with the value of
+`lsp-clients-go-library-directories'."
+  (let ((library-dirs lsp-clients-go-library-directories))
+    (when (and lsp-clients-go-library-directories-include-go-modules
+               (or (and (not (file-remote-p default-directory)) (executable-find "go"))
+                   (and (version<= "27.0" emacs-version) (with-no-warnings (executable-find "go" (file-remote-p default-directory))))))
+              (with-temp-buffer
+                (when (= 0 (call-process "go" nil t nil "env" "GOPATH"))
```
yyoncho: nit: use zerop, also you may consider using `shell-command-to-string`

lsp-go.el
```diff
@@ -143,17 +169,15 @@ defaults to half of your CPU cores."
                   :priority -1
                   :initialization-options 'lsp-clients-go--make-init-options
                   :server-id 'go-bingo
-                  :library-folders-fn (lambda (_workspace)
-                                        lsp-clients-go-library-directories)))
+                  :library-folders-fn 'lsp-clients-go--library-default-directories))
```
yyoncho: Does bingo support this new feature?

psanford: Bingo does seem to have support for go modules (it itself is implemented as a module).

lsp-go.el
psanford: Does shell-command-to-string work properly for remote buffers?

I'll switch this to use zerop.

yyoncho: Not sure, if it doesn't I am fine with the code as it is.

psanford: I'll do some testing with tramp and let you know.

yyoncho: Just to clarify - I am fine to have it as it is even without testing whether it works over tramp. It is up to you whether you want to invest more time into that...

psanford: I've updated the pr with proper support for remote sessions. (Remote sessions is one of the biggest complaints in go-mode.el with the legacy non-lsp tools, so we really want it to work for people using gopls).

muirrn:
Looks good to me!

yyoncho:
Thank you @muirrn  and @psanford !

PR #1046, opened 2019-09-26
Switch from autoload to fsharp/workspaceLoad

TOTBWF:
Previously, we were using `automaticWorkspaceInit` to get fsautocomplete to load up all of the workspaces. This works fine for simple projects, but fails on more complicated ones.

This PR implements the `fsharp/workspacePeek` and `fsharp/workspaceLoad` protocol, documented [here](https://github.com/fsharp/FsAutoComplete). In the future, it would probably be useful to expose a UI for selecting project files to load, but this is a good first start.


TOTBWF:
Just realized that `let-alist` isn't a dep we have, going to have to rework things a bit

yyoncho:
@TOTBWF if it is not a problem for you, could you open an issue before working on lsp-mode related tasks to avoid collisions?

TOTBWF:
Of course, will do :)

PR #1049, opened 2019-09-26
Try rangeFormatting when formatting is not supported

### LINKED ISSUE #1048: lsp-format-buffer when server supports textDocument/rangeFormatting but not textDocument/formatting
Some servers have support for the `textDocument/rangeFormatting` capability but not the `textDocument/formatting` capability so `lsp-format-buffer` fails.

Examples:
- [vscode-html-languageserver](https://github.com/vscode-langservers/vscode-html-languageserver)
- [vscode-json-languageserver](https://github.com/vscode-langservers/vscode-json-languageserver)

Would it make sense for `lsp-format-buffer` to apply a range format to the content of the buffer when this is the case?

alypeng:
Resolves #1048

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3924,7 +3924,9 @@ If ACTION is not set it will be selected from `lsp-code-actions'."
            (lsp--apply-formatting edits)))
         ((or (lsp--capability "documentRangeFormattingProvider")
              (lsp--registered-capability "textDocument/rangeFormatting"))
-         (lsp-format-region (point-min) (point-max)))
+         (save-excursion
+           (widen)
+           (lsp-format-region (point-min) (point-max))))
         (t (signal 'lsp-capability-not-supported (list "documentFormattingProvider")))))
 
 (defun lsp-format-region (s e)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3917,19 +3917,22 @@ If ACTION is not set it will be selected from `lsp-code-actions'."
...
-                            (lsp--make-document-formatting-params))))
-    (lsp--apply-formatting edits)))
+  (cond ((or (lsp--capability "documentFormattingProvider")
+             (lsp--registered-capability "textDocument/formatting"))
+         (let ((edits (lsp-request "textDocument/formatting"
+                                   (lsp--make-document-formatting-params))))
+           (lsp--apply-formatting edits)))
+        ((or (lsp--capability "documentRangeFormattingProvider")
+             (lsp--registered-capability "textDocument/rangeFormatting"))
+         (lsp-format-region (point-min) (point-max)))
```
yyoncho: Usually, the pattern when using point-min/point-max is 
```
(save-excursion
  (widen
     ... use point-min/point-max
```
in order to handle the case when the user has narrowed the buffer.

yyoncho:

PR #1053, opened 2019-09-27
Swap RLS register to crate_blacklist

ventlark:
According to RLS: option `use_crate_blacklist` is deprecated: use
`crate_blacklist` instead. This change implements just that.

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -114,7 +114,7 @@ the latest build duration."
   :group 'lsp-rust
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-rust-use-crate-blacklist t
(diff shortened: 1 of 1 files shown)


ventlark:
Note: This also changes the variable `lsp-rust-use-crate-blacklist` to `lsp-rust-crate-blacklist`. This may cause people who have configured `lsp-rust-use-crate-blacklist` to have broken configs.

A few possible solutions to this I can think of would be: leave the variable name as-is, redirect the value of `lsp-rust-use-crate-blacklist` to `lsp-rust-crate-blacklist`, or have a message printed out to the user prompting them to change it.

yyoncho:

PR #1056, opened 2019-10-02
Enable dynamic registration for rangeFormatting

iquiw:
VSCode JSON Language Server, YAML Language Server, etc. need `textDocument.rangeFormatting.dynamicRegistration` to use formatting feature.


yyoncho:
Thank you, it will be great if you can provide a PR so we can support these servers out of the box.

iquiw:
Yes, I am planning to send PRs for JSON and YAML langservers.

Thanks!

PR #1057, opened 2019-10-03
Fix invalid string escapes in lsp-language-id-configuration

CyberShadow:
`"\."` is the same as `"."`, i.e. "any character" as a regular expression.


yyoncho:
Thank you!

PR #1058, opened 2019-10-04
Fix default typescript-plugins in typescript language server.

isubasti:
lsp typescript is expected a vector as its plugins if specified, the previous default `nil` is causing lsp typescript to stuck at initialization step.


yyoncho:
Thank you!

PR #1059, opened 2019-10-04
Fix hover text not updating

CyberShadow:
```
When moving point from one position with one hover text directly to
another position with a different hover text, the echo area was not
being updated; instead, the previous hover text was being redisplayed.

This was due to the previous message being saved in
lsp--eldoc-saved-message, and redisplayed from there. The only
circumstance in which the variable was cleared was when point was
moved to whitespace and "hovered" there.

Fix this by clearing lsp--eldoc-saved-message as soon as it is clear
that its contents is no longer valid, i.e., before sending a
textDocument/hover request.
```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3813,12 +3813,11 @@ RENDER-ALL - nil if only the signature should be rendered."
   (if (and lsp--hover-saved-bounds
            (lsp--point-in-bounds-p lsp--hover-saved-bounds))
       (lsp--eldoc-message lsp--eldoc-saved-message)
+    (setq lsp--hover-saved-bounds nil
+          lsp--eldoc-saved-message nil)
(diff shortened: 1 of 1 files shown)


yyoncho:
> Fix this by clearing lsp--eldoc-saved-message after displaying it in
response to a textDocument/hover reply.

This will break the case when there are lsp--hover-saved-bounds. I think that

```
          (setq lsp--hover-saved-bounds nil
                  lsp--eldoc-saved-message nil)
```

should be before ``` (if whitespace-or-newline``` but I do not have the time to test it right now.

CyberShadow:
> This will break the case when there are lsp--hover-saved-bounds. I think that

Right, thanks.

CyberShadow:
> should be before ` (if whitespace-or-newline` but I do not have the time to test it right now.

It seems to work. Though, I haven't tested the interaction with `lsp-eldoc-prefer-signature-help`.

PR #1060, opened 2019-10-04
Consider hover contents consisting of a single line as signatures

CyberShadow:
```
When textDocument/hover returns a single string consisting of a single
line, show it in the minibuffer regardless of lsp-eldoc-render-all.
```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3754,10 +3754,16 @@ When language is nil render as markup if `markdown-mode' is loaded."
   "Render the content received from 'document/onHover' request.
 CONTENTS  - MarkedString | MarkedString[] | MarkupContent
 RENDER-ALL - nil if only the signature should be rendered."
-  (if (and (hash-table-p contents) (gethash "kind" contents))
-      ;; MarkupContent, deprecated by LSP but actually very flexible.
-      ;; It tends to be long and is not suitable in echo area.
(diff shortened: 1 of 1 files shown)


yyoncho:
Can you share which language servers you have used for testing?

CyberShadow:
I'm writing my own LSP server, though I think the change makes sense regardless of language. Perhaps the logic could be mode more general if that's something worth discussing.

CyberShadow:
If you have reservations about this change, I'm fine with addressing this locally (i.e. setting `lsp-eldoc-render-all`).

yyoncho:
I will have to test it with some of the popular language servers to see how it works.

Related: https://github.com/microsoft/language-server-protocol/issues/772

yyoncho:
Merging as it is and if someone complains(e. g. if there is huge single line text) we will consider make it configurable. 

Thank you.

jsravn:
@yyoncho what would you think about displaying the first line if MarkupContent is returned? As-is most language servers are returning MarkupContent in my experience - so eldoc is never rendered. There isn't a great solution since onHover is also used for documentation popups.

yyoncho:
@jsravn either this or take N chars. I am fine with this change.

jsravn:
@yyoncho https://github.com/emacs-lsp/lsp-mode/pull/1607

Issue #994, opened 2019-08-30
`lsp-log-io` buffer does not scroll to end when appending


jesperes:
**Describe the bug**
Setting `lsp-log-ui` to `t` causes `lsp-mode` to log all LSP commands to the `*lsp-log: ...*` buffer, but does not scroll to end when new commands are appended. This makes it quickly very difficult to see what is happening, without manually having to scroll to the end of the file. Also, `auto-revert-tail-mode` cannot be used since the buffer is not visiting a file (perhaps a fix would be to just write the logs to a file, then let user do `tail -f ...` or similar).

**To Reproduce**
Do `(setq lsp-log-io t)`, observe output in `*lsp-log: ...*` buffer being appended, but buffer not scrolling to end.

**Expected behavior**
I would have expected, or at least liked, to be able to have some way of observing the output as it is appended to the buffer without having to scroll to the end of the buffer manually.

**Which Language Server did you use**
`erlang_ls`, but that probably doesn't matter.

**OS**
Linux

yyoncho:
If you place the cursor at end of the file the upcoming messages will move it.

jesperes:
> If you place the cursor at end of the file the upcoming messages will move it.

I tried this, but the cursor stays at the same place. Appended messages does not move it.

yyoncho:
Sorry, I totally misread your comment. The feature works only in \*lsp-log\* buffer.

yyoncho:
https://stackoverflow.com/questions/12663061/emacs-auto-scrolling-log-buffer + placing the cursor at the end should work.

jesperes:
Sorry, but `(set (make-local-variable 'window-point-insertion-type) t)` does not help.

jesperes:
> Sorry, but `(set (make-local-variable 'window-point-insertion-type) t)` does not help.

But now it actually does. Weird.

Anyway, this should IMO really be the default for the log buffer.

robertoaloi:
The above method seems to require `evil-mode`, or?

yyoncho:
I tested with emacs -q and loading only lsp related packages so it should work without evil mode.

jesperes:
> The above method seems to require `evil-mode`, or?

I think that comment should be interpreted as "if you are using evil-mode, this only works in evil-insert-mode".

jfacorro:
Is there a better way of "setting a variable in the buffer" than this?
```
(with-current-buffer "*lsp-log: erlang-ls:48439*"
[...]

PR #1061, opened 2019-10-04
[#994] Enable scrolling by default in lsp-log-io buffers

jfacorro:


jfacorro:
Build fails because spinner is not available when installing through cask. 😞

yyoncho:
Thank you!

PR #1062, opened 2019-10-04
Improved support for textDocument/prepareRename

CyberShadow:


CyberShadow:
After a bit of rebasing, GitHub is now for some reason showing commits in the wrong order. This makes it harder to read the diffs.

The correct order (and the order in which they actually appear in the git history) is:

1. "Fix handling textDocument/prepareRename returning an object"
2. "Handle invalid renames"
3. "Use the "placeholder" field from textDocument/prepareRename"

yyoncho:
Looks good!

PR #1065, opened 2019-10-05
Import new GPG key to download ELPA package on Travis CI

iquiw:
This solves `spinner` download error from ELPA.

Emacs 26.1 and 26.2 are failing by other reason.


yyoncho:
Thank you for doing that.

Would you mind filing an issue before starting to work on tasks like that to avoid someone else working on it?

iquiw:
> Would you mind filing an issue before starting to work on tasks like that to avoid someone else working on it?

OK, I will do it.

PR #1067, opened 2019-10-05
Use non-versioned TextDocumentIdentifier in didClose

CyberShadow:
According to [the specification](https://microsoft.github.io/language-server-protocol/specifications/specification-3-14/#textDocument_didClose), the `textDocument` parameter for `textDocument/didClose` is `TextDocumentIdentifier`, not `VersionedTextDocumentIdentifier`.

RFC, maybe some servers rely on this information? I had a look at the git history and it was there since the beginning, I'm guessing not intentionally.

Eglot seems to be correctly sending just a `TextDocumentIdentifier`.


yyoncho:
Thank you! I verified with VScode which is the reference implementation(most of the servers test what works with vscode, not what the spec says).

PR #1070, opened 2019-10-05
Update intelephense configuration options

alypeng:
See https://github.com/bmewburn/vscode-intelephense/blob/master/package.json


yyoncho:
Thank you!

PR #1071, opened 2019-10-06
Display signature help regardless of where point is

alypeng:
It should be possible to see signature help when arguments are being split across multiple lines.

For example:
```php
// php
$this->foo(
    $arg1,
    $arg2,
    |
)
```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -413,13 +413,6 @@ This flag affects only server which do not support incremental update."
   :type 'boolean
   :group 'lsp-mode)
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you for taking care of than. Recently I am not using lsp-mode but only developing it and I haven't noticed the regression. What do you think about changing the code to skip only onHover when over whitespace and run document signature help always no matter where the cursor is?

alypeng:
Yep, sounds good to me. I'll update the PR.

yyoncho:
Thank you!

PR #1072, opened 2019-10-06
Added support for VHDL

chrbirks:
Added support for vhdl-mode via the LSP server from www.vhdltool.com


yyoncho:

lsp-vhdl.el
```diff
@@ -0,0 +1,70 @@
...
+  :risky t
+  :type 'file)
+
+(defun lsp-vhdl--create-connection ()
+  "Returns lsp-stdio-connection or an error if server not found"
+  (if (file-exists-p lsp-vhdl-server-path)
+      (lsp-stdio-connection
+       (lambda ()
+         (list lsp-vhdl-server-path "lsp")))
+    (error "VHDL LSP server binary not found: `%s'" lsp-vhdl-server-path)
```
iquiw: @chrbirks @yyoncho This error is already thrown with default setting, isn't it?

https://travis-ci.org/emacs-lsp/lsp-mode/jobs/594199984#L1371

chrbirks: You are right, it is.

yyoncho: I already pushed a fix + travis is now green.

PR #1076, opened 2019-10-07
Improve wording of the repository file watching warning

danielmartin:
- (lsp--ask-about-watching-repo): New function. Warning message is
shorter to make it less scary for users. The available configuration
variables are presented not as part of the message, but after the user
has answered. Also, per `yes-or-no-p' documentation, the string now ends in
a space.

- (lsp-watch-root-folder): Modify code to
call (lsp--ask-about-watching-repo).

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1273,7 +1273,7 @@ Do you want to watch all files in %s? "
         number-of-files
         dir
         dir))
-    (message
+    (lsp--info
      (concat "You can configure this warning with the `lsp-enable-file-watchers' "
              "and `lsp-file-watch-threshold' variables"))))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1260,6 +1260,23 @@ DELETE when `lsp-mode.el' is deleted.")
             (push (concat dir "/" file) files)))))
     (nconc result (nreverse files))))
 
+(defun lsp--ask-about-watching-big-repo (number-of-files dir)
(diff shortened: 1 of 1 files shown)


yyoncho:
@alypeng you are the author of the current wording, willing to review the proposed one?

danielmartin:
Build failure should be fixed by https://github.com/emacs-lsp/lsp-mode/commit/4ee6808e0e1853399ef78f0cff01f28683ec9ca1

@yyoncho What do think about this change? One of my worries is if customers understand the message well. For example, it's still not clear what the benefits of watching all files from a repo are, and many users may ask "will LSP work if I answer 'no'? I'm not sure how we can express that in an easy to understand way.

yyoncho:
@danielmartin I think that we might create a wiki page(or a configuration section in readme) with more details what the file watchers do and what scenarios won't work if file watches are disabled. Other than that I am fine with the change.

lsp-mode.el
```diff
@@ -1260,6 +1260,23 @@ DELETE when `lsp-mode.el' is deleted.")
...
+the entire repository, nil otherwise."
+  (prog1
+      (yes-or-no-p
+       (format
+        "There are %s files in folder %s so watching the repo may slow Emacs down.
+Do you want to watch all files in %s? "
+        number-of-files
+        dir
+        dir))
+    (message
```
yyoncho: can you use to lsp--info instead of message?

PR #1080, opened 2019-10-09
display rust-analyzer command instead of rls

brotzeit:


yyoncho:
IMO we should have separate configurations since the differences are much more than the common stuff. For example, the configuration sections are different for both servers and have different settings: 
https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/package.json#L165 vs https://github.com/rust-lang/rls-vscode/blob/master/package.json#L149

We could use this as a base:
https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/emacs/ra-emacs-lsp.el

brotzeit:
That's actually what I had in mind. Currently this code https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/emacs/ra-emacs-lsp.el is in rustic. But I would like to also provide rust-mode users a way to easily use the code and remove it from rustic.

yyoncho:
@brotzeit 
> But I would like to also provide rust-mode users a way to easily use the code and remove it from rustic.

When we have the 2 registration you could have functions that change the priorities of the server and control which server will be selected, like this: `(setf (lsp--client-priority (gethash 'rls lsp-clients)) 1)`

PR #1086, opened 2019-10-10
Update `gopls` initialization options

Gastove:
`gopls`, wrapped by `lsp-go`, supports a variety of settings not currently
exposed as `defcustom`s. This commit adds them. For full docs, see:

https://github.com/golang/tools/blob/master/gopls/doc/settings.md

Note that a set of these are marked experimental in the doc, and so I've
included the word `experimental` in the corresponding `defcustom`.

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -47,13 +47,13 @@ completing function calls."
   :type '(repeat string)
   :group 'lsp-gopls)
 
-(defcustom lsp-gopls-build-flags '("-tags")
-  "A set of flags passed on to the build system when invoked,
+(defcustom lsp-gopls-build-flags ["-tags"]
+  "A vector of flags passed on to the build system when invoked,
   applied to queries like `go list'."
   :type '(repeat string)
   :group 'lsp-gopls
   :risky t
-  :package-version '(lsp-mode "6.3"))
+  :package-version '(lsp-mode "6.2"))
 
 (defcustom lsp-gopls-env "{}"
   "`gopls' has the unusual ability to set environment variables,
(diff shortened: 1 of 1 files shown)


lsp-go.el
```diff
@@ -47,8 +47,93 @@ completing function calls."
   :type '(repeat string)
   :group 'lsp-gopls)
 
+(defcustom lsp-gopls-build-flags '("-tags")
```
yyoncho: Lists must be vectors otherwise it wont work when using emacs 27 native json serialization which threats all lists as objects. We should update that in the settings generator.

lsp-go.el
```diff
@@ -47,8 +47,93 @@ completing function calls."
...
+  "A set of flags passed on to the build system when invoked,
+  applied to queries like `go list'."
+  :type '(repeat string)
+  :group 'lsp-gopls
+  :risky t
+  :package-version '(lsp-mode "6.3"))
```
yyoncho: Untill 6.2 is released we must keep 6.2 versin.

lsp-go.el
Gastove: :+1: 

Is there a better `:type` to use for vectors?

lsp-go.el
Gastove: Can do :+1:

yyoncho:

lsp-go.el
yyoncho: Yes, we could use:


```
 :type '(restricted-sexp :match-alternatives (lambda (xs)	  :type '(repeat string)
                                                (and	
                                                 (vectorp xs)	
                                                 (seq-every-p #'stringp xs))))
```

This must be done in the whole project, not only in this PR.

Gastove:
Huzzah! :tada: Thanks for the help, and the swift review!

lsp-go.el
Gastove: Ah, very good. I'll make a note to start tidying some of those up next.

yyoncho: Please if you want to work on a particular task, file an issue and ask me to assign it to you to avoid collisions.

PR #1084, opened 2019-10-09
[Rust] Fix unactionable config

SamirHafez:
It's possible https://github.com/emacs-lsp/lsp-mode/pull/1063 broke correct LSP initialization with the Rust Language Server.

Among other issues, this resulted in https://github.com/emacs-lsp/lsp-mode/issues/1083

By enabling debugging in the Rust Language Server we can identify the issue:
```
[2019-10-09T21:26:38Z DEBUG rls::server] Language Server starting up. Version: rls 1.39.0 (8dc9ba9 2019-10-01)

[2019-10-09T21:26:41Z WARN  rls::actions::notifications] Received unactionable config:
Object({"rust": Object(
{"all_features": Bool(false),
"all_targets": Bool(true),
"build_bin": Null,
"build_command": Null,
"build_lib": Bool(false),
"build_on_save": Bool(false),
"cfg_test": Bool(false),
"clear_env_rust_log": Bool(true),
"clippy_preference": String("on"),
"crate_blacklist": Array([String("cocoa"), Array([String(","), String("gleam")]), Array([String(","), String("glium")]), Array([String(","), String("idna")]), Array([String(","), String("libc")]), Array([String(","), String("openssl")]), Array([String(","), String("rustc_serialize")]), Array([String(","), String("serde")]), Array([String(","), String("serde_json")]), Array([String(","), String("typenum")]), Array([String(","), String("unicode_normalization")]), Array([String(","), String("unicode_segmentation")]), Array([String(","), String("winapi")])]),
"features": Array([]),
"full_docs": Bool(false),
"jobs": Null,
"no_default_features": Bool(false),
"racer_completion": Bool(true),
"rustflags": Null,
"rustfmt_path": Null,
"show_hover_context": Bool(true),
"show_warnings": Bool(true),
"sysroot": Null,
"target": Null,
"target_dir": Null,
"unstable_features": Bool(false),
"wait_to_build": Null})}) (error: ())
```

Notice the `Received unactionable config` `notification`. This fails because `crate_blacklist` is an `array` of tuples (the separation `,`s are being handled as array items) when it should have been an array of `string`s.

This PR fixes that by changing the `crate_blacklist` field type to a `(repeat string)`. This is the result from the Rust Language Server log after applying this patch:

```
[2019-10-09T21:23:30Z DEBUG rls::server] Language Server starting up. Version: rls 1.39.0 (8dc9ba9 2019-10-01)
[...]

PR #1087, opened 2019-10-10
[Rust] Fix unactionable config

SamirHafez:
Addresses problem in https://github.com/emacs-lsp/lsp-mode/pull/1084

@yyoncho adds missing commit from previous fix

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -133,7 +133,9 @@ the latest build duration."
   :type '(restricted-sexp :match-alternatives (lambda (xs)
                                                 (and
                                                  (vectorp xs)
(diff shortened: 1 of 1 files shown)


SamirHafez:
https://github.com/emacs-lsp/lsp-mode/pull/1084#issuecomment-540409120

yyoncho:
Already pushed it.

SamirHafez:
This addresses missing information from the one you pushed. This adds `group` and `package-version` that was missing from before

SamirHafez:
Ah, you've reverted it. OK, thanks

PR #1088, opened 2019-10-11
lsp-vetur add default emmet setting

thanhvg:
vls will throw a error inside style tag:
```
 "Request textDocument/completion failed with message: Cannot read property'showExpandedAbbreviation' of undefined"
```
This should be handled by vls but until they do we need this work around.


yyoncho:

PR #1089, opened 2019-10-11
change url allowed chars to default, which is url-unreserved-chars

mgourd:
I encountered "Unable to find file" problem when using cquery as language server, I did some search and find that it's because the "+" in my file path didn't get hexified. Changing "+" to a letter solves my problem.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1196,8 +1196,7 @@ DELETE when `lsp-mode.el' is deleted.")
 
 (defun lsp--path-to-uri-1 (path)
   (concat lsp--uri-file-prefix
-          (url-hexify-string (expand-file-name (or (file-remote-p path 'localname t) path))
-                             url-path-allowed-chars)))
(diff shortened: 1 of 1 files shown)


mgourd:
#227 #238

yyoncho:
Thanks for the PR. I am on a vacation and I will do the testing on Monday. This PR touches core part of the project so we should make sure that it works fine in all of the cases.

yyoncho:
Can you rebase to check whether the travis will be green? ATM it is hard to tell whether the travis failure is related to the PR.

mgourd:
CI still fails. But I'm not sure if I'm doing the rebase in a right way because I'm not new to contributing on GitHub. Sorry if my unfamiliarity with git and GitHub cased any problems.

yyoncho:
@mgourd do worries, you did the rebase correctly. Unfortunately, it seems like the proposed code change breaks emacs 25.x . We should investigate that before proceeding to merge it.

kiennq:
I think this can be close now. We're switching to use `url-unreserved-chars` + `/` already

PR #1092, opened 2019-10-13
Add support for merlni lsp server for ocaml

rgrinberg:
Merlin is the primary backend of for all editor suppoert in OCaml and in
fact ocaml-language-server is based on it. The newest version of merlin
will provide its own lsp server, so it should be preferred to
ocaml-language-server.


yyoncho:
Thank you for your contribution! Can you update the readme with the information about this server?

rgrinberg:
Done. I fixed the capitalization of OCaml as well.

yyoncho:
Thank you for contributing to `lsp-mode` and congrats for you first PR! :tada:

PR #1093, opened 2019-10-13
Load yasnippet as an optional dependency

aidalgol:
`require` yasnippet with the NOERROR flag so that we can use it if it is installed without the user having to explicitly load it before using lsp-mode.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -57,6 +57,7 @@
 (require 'tree-widget)
 (require 'markdown-mode)
 (require 'ewoc)
+(require 'yasnippet nil t)
 
 (declare-function company-mode "company")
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -2522,7 +2523,7 @@ disappearing, unset all the variables related to it."
                                                                                 (featurep 'yasnippet)
                                                                                 (warn (concat
                                                                                        "Yasnippet is not yet loaded, but `lsp-enable-snippet' is set to `t'. "
```
yyoncho: loaded -> installed.

aidalgol: Oops.  Fixed.

yyoncho:

PR #1097, opened 2019-10-15
Correct `gopls` defcustom types

### LINKED ISSUE #1096: Invalid type <nil> for []string option "experimentalDisabledAnalyses" in lsp-go
Recently many people got this error while using `lsp-go`. 

```elisp
LSP :: Invalid type <nil> for []string option "experimentalDisabledAnalyses"
```

Gastove:
Hokay. I've been running this locally for a few hours, and it's looking a lot better. Notes about this:

1. In my testing, using either any kind of list of pairs or a `plist` all resulted in `lsp-mode` serializing an empty data structure as a JSON `null`, which makes `gopls` very angry. It _must_ be sent an empty object (`{}`). An empty `hash-table`, rather than being `nil` (that is, `null`), fills this requirement. I'm open to doing this differently, but it's working for me for now. 

2. I've updated JSON `array` types to use vectors. There was only one I missed before, but I also corrected the `:type` key on `lsp-gopls-build-flags` to clarify. 

Closes #1096

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -50,17 +50,17 @@ completing function calls."
 (defcustom lsp-gopls-build-flags ["-tags"]
   "A vector of flags passed on to the build system when invoked,
   applied to queries like `go list'."
-  :type '(repeat string)
+  :type '(vector string)
   :group 'lsp-gopls
   :risky t
   :package-version '(lsp-mode "6.2"))
 
-(defcustom lsp-gopls-env "{}"
+(defcustom lsp-gopls-env (make-hash-table)
   "`gopls' has the unusual ability to set environment variables,
   intended to affect the behavior of commands invoked by `gopls'
-  on the user's behalf. This variable takes a JSON object
-  literal, mapping env var name to value."
-  :type 'string
+  on the user's behalf. This variable takes a hash table of env
+  var names to desired values."
+  :type '(restricted-sexp :match-alternatives (hash-table-p))
   :group 'lsp-gopls
   :risky t
   :package-version '(lsp-mode "6.2"))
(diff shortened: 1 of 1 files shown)


Gastove:
@seagle0128 this should help, I hope.

lsp-go.el
```diff
@@ -50,17 +50,17 @@ completing function calls."
 (defcustom lsp-gopls-build-flags ["-tags"]
   "A vector of flags passed on to the build system when invoked,
   applied to queries like `go list'."
-  :type '(repeat string)
+  :type '(vector string)
```
yyoncho: there is no vector type, can you change this to restricted expression?

seagle0128: Agree.

Gastove: Hi! I'm happy to change it. That said, there _is_ a vector type (at least, the docs list one): https://www.gnu.org/software/emacs/manual/html_node/elisp/Composite-Types.html#Composite-Types

```
(vector element-types...)
    Like list except that the value must be a vector instead of a list. The elements work the same as in list. 
```

I will be guided by y'all. Do you prefer `vector`, or restricted expression?

yyoncho: You are right. I guess in the past we have used vector without a type parameter. Merging as it is.

seagle0128:
I tested the patch, and it works well now, except one: `LSP :: unexpected config staticcheck`. Frankly I am not sure if it's related to the PR.

yyoncho:
You have to verify that you are using the same version of the server.

Gastove:
Ugh, and `gopls` makes that extremely tricky -- especially given that `gopls@latest` currently has a pretty nasty memory leak (see [this issue](https://github.com/golang/go/issues/34906) on their tracker).

Gastove:
Why don't I open a separate issue for this? Track the details somewhere easier to find than a closed PR.

seagle0128:
Right! Thanks, @Gastove .

lsp-go.el
iquiw: IIUC, `vector` is fixed length type. i.e. `(vector string)` is a vector of one string element.

> (list element-types...)
>    The value must be a list with exactly as many elements as the element-types given; and each element must fit the corresponding element-type.
>
>    For example, (list integer string function) describes a list of three elements; the first element must be an integer, the second a string, and the third a function. 

> (vector element-types...)
>    Like list except that the value must be a vector instead of a list. The elements work the same as in list.

yyoncho: @iquiw thank you. I remember that there was a reason not to use the vector type but I didn't remember what was the case.

Gastove: Oh neat. 

Tomorrow I'll open an issue to track fixing that. Thanks, @iquiw -- would *not* have guessed it worked that way :man_facepalming:

PR #1103, opened 2019-10-16
add basic rust-analyzer support

brotzeit:


brotzeit:
> Can you add the entry in the README servers list and eventually recommend a server? 

I'm actually not sure what's currently the better solution.

PR #1107, opened 2019-10-16
Implement `lsp-string-vector` 

### LINKED ISSUE #1106: Clean up use of `:type (vector string)`
In a recent set of PRs (#1086, #1097), there was a little confusion about the correct `:type` key for representing vectors of strings. To summarize: `:type '(vector string)` is valid elisp, but not the type we want, as it specifies a _vector of length 1_, specifically. The correct `:type` is a `restricted-sexp`, using a function that `and`s `vectorp` with checking that all elements are of type `string`. 

Just to add flavor: the two _existing_ usages of `restricted-sexp`, both in `lsp-rust`, are incorrect. (They use the form `restricted-sexp :match-alternatives (lambda (xs)`, which doesn't work -- the argument to `:match-alternatives` is a list, not a function.)

To make this easier to get right, particularly for newcomers, I propose defining a new `:type`, `lsp-vector`. [New types](https://www.gnu.org/software/emacs/manual/html_node/elisp/Defining-New-Types.html#Defining-New-Types) are declared as new "widgets" (thanks Emacs, that's very obvious). As a first pass, this could look like so: 

```emacs-lisp
(defun lsp--string-vector-p (candidate)
  "Returns true if CANDIDATE is a vector data structure, and
every element of it is of type string."
  (and
   (vectorp candidate)
   (seq-every-p #'stringp candidate)))

(define-widget 'lsp-string-vector 'lazy
       "A vector of any length, every element of which is a string."
       :offset 4
       :tag "Vector"
       :type '(restricted-sexp
               :match-alternatives (list #'lsp--string-vector-p)))
```
This could then be used like: 
```emacs-lisp
(defcustom lsp-cool-variable ["arg"] "a variable"
    :type lsp-string-vector)
```

Gastove:
This PR creates a new type for use in `defcustom`s that need to represent vectors of strings. 

Very up for debate: what should this new type be _called_. I currently favour `lsp-string-vector`, because I feel it conveys to _users_ of the type what they should use as values for a custom using this type. Putting `json` or `array` in the name are also up for discussion! The name `lsp-json-string-array` certainly reflects what the type _becomes_ much more clearly. 

Closes #1106

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -50,7 +50,7 @@ completing function calls."
 (defcustom lsp-gopls-build-flags ["-tags"]
   "A vector of flags passed on to the build system when invoked,
   applied to queries like `go list'."
-  :type '(vector string)
+  :type 'lsp-string-vector
   :group 'lsp-gopls
   :risky t
   :package-version '(lsp-mode "6.2"))
@@ -85,7 +85,7 @@ completing function calls."
 ;; -- @gastove 2019-10-09
 (defcustom lsp-gopls-experimental-disabled-analyses []
   "A list of names of analysis passes that should be disabled."
-  :type '(vector string)
+  :type 'lsp-string-vector
   :group 'lsp-gopls
   :risky t
   :package-version '(lsp-mode "6.2"))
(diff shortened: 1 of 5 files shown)


yyoncho:
I tested locally, works fine. We should update the lsp-generate-settings.el to accommodate the new type.

yyoncho:
Thank you!

PR #1109, opened 2019-10-17
update lsp-rust

brotzeit:


lsp-rust.el
```diff
@@ -380,8 +380,9 @@ PARAMS progress report notification data."
   "Switch priorities of lsp servers."
   (interactive)
   (dolist (server '(rls rust-analyzer))
-    (setf (lsp--client-priority (gethash server lsp-clients))
-          (* (lsp--client-priority (gethash server lsp-clients)) -1))))
+    (when (natnump (setf (lsp--client-priority (gethash server lsp-clients))
+                         (* (lsp--client-priority (gethash server lsp-clients)) -1)))
+      (message (format "Switched to server %s." server)))))
```
brotzeit: I wonder if we should also offer to shutdown the running server and ask if the active server should be started.

yyoncho: if you are going to put the effort to implement that you may make it generic since we have several "duplicate-servers". E. g. implement function:
``` elisp
(defun lsp-switch-server (target-server-id) ...
;; calculate what servers do support that file
;; ask user if more than one
;; switch active server/new server priorities.
;; call lsp-restart
```

IMO this scenario is not that common(usually users use lsp-enabled-clients/lsp-disabled-clients using often in combination with dir-locals) but if you want to implement it - go ahead.

As a side note, rust analyzer seems to have a lot of extension methods and I think that you might be interested in exposing them first.

brotzeit: Ok. I will take a look at it.

yyoncho: For reference:

https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/src/commands/syntaxTree.ts 

https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/src/commands/inlay_hints.ts

I haven't tried them in vscode but they look interesting.

Issue #506, opened 2018-12-05
Feature: automatic server installation


yyoncho:
We could use https://gitlab.com/jabranham/system-packages to automatically download and install language servers that have published their packages(npm/gem/pip).

NB: the servers that does not have installation package won't be handled by this feature.

(cc @jabranham)

seagle0128:
Some comments form my experience. I tried this package with `use-package` a few months ago. It's unable to handle all scenarios on all platforms, especially for Windows and `sudo` with Linux. So it's a good feature but need more cautions.

jabranham:
@seagle0128 What issues did you run into with sudo on linux? Would you mind opening a new issue for it?

seagle0128:
@jabranham  I remember that some commands need `sudo` privileges while others don't. During Emacs startup, there will be many prompts to interrupt, and really annoying! Especially on remote development environment which didn't install all packages :-(

yyoncho:
We may use the vscode package for the installation, I am planning to do that for `dap-mode`.

eubarbosa:
@jabranham 
Ok. I will open a issue at gitlab! 

@seagle0128 
Just tried system-packages today, it is simpler (no prompt at startup), and it correctly guess by checking the distro package if its needs sudo to install packages! 

I had just a minor annoyance, if you have guix or nix installed in your distro it prefer those two package-manager rather than the distro default (eg: instead of apt it installs with guix)...

Opening a issue late!

seagle0128:
@eubarbosa No password prompt for `sudo` on Linux?

I used `system-packages` for legacy `lsp-mode` to try installing all servers at startup. I think the current situation is different. If the servers are installed and started on-demand, the annoyances I met should not exist.

tam5:
I'm all for auto installation of language servers, with the condition that a user could customize this to turn it off and specify the language server they want to use directly.

Also, I think a quicker/easier fix, should this take a little while to get done, would be to just add more detailed and exact instructions for installing each language server, so someone new to the idea of lsp would still be able to get up and running quickly

alanz:
I think providing a pointer to install instructions makes more sense. Some language servers are fairly complex to install, and may involve interaction with the system package manager.

yyoncho:
@alanz that's correct, I will update the feature description. In general I was referencing the following servers:

1. Servers that are available as a package like (pyls)
2. Servers that are packaged in VSCode extension(php).

Pitometsu:
[...]

PR #1112, opened 2019-10-18
A handy utility for updating fsac

Gastove:
A tiny commit to add a tiny function to make it easier to update `FSharpAutoComplete`. Handles the two-step of delete-old-dir, reinstall.


yyoncho:
Not something to address in this PR but long term we should make it generic across language servers. For example, we may add another field in lsp--client - `download-function` and then we can create a generic function - `lsp-update-server` which will look for all language servers that have that field specified and it will ask the user to select one. This is related to https://github.com/emacs-lsp/lsp-mode/issues/506

Gastove:
@yyoncho I really like this idea. Agreed, not for this PR, but having a language-agnostic mechanism for this would be _delightful_.

TOTBWF:
Just tested this out and it LGTM!

Issue #1113, opened 2019-10-18
(file-error "Read error" "Is a directory" "/path/to/a/directory")


JohnC32:
When running lsp-find-references on a heavily used C++ class (I'm using lsp-mode with ccls), sometimes I see:

    Debugger entered--Lisp error: (file-error "Read error" "Is a directory" "/path/to/a/directory")
      insert-file-contents("" nil nil nil nil)
      insert-file-contents-literally("")
      #[514 .....
      apply(#[514 .....
      #[128 .....
      mapcar(#[128 .....
      #[514 .....
      apply(#[514 .....
      seq-map(#[128 .....
      lsp-find-references()
      funcall-interactively(lsp-find-references)
      call-interactively(lsp-find-references nil nil)
      command-execute(lsp-find-references)

lsp--locations-to-xref-items, is calling (file-readable-p filename) and (insert-file-contents-literally filename), with filename equal to "". This causes it to insert the current directory. 

Why is filename empty?

I can work around the problem by modifying lsp--locations-to-xref-items to protect against an empty file name using the following "FIX". However this seems like it is masking a more fundamental issue. Can you help provide more insight as to what I can do to get to the root cause of the empty filename?

Thanks

````emacs-lisp
(defun lsp--locations-to-xref-items (locations)
  "Return a list of `xref-item' from Location[] or LocationLink[]."
  (setq locations (if (sequencep locations) locations (list locations)))

  (unless (seq-empty-p locations)
    (cl-labels ((get-xrefs-in-file
                 (file-locs location-link)
                 (let ((filename (seq-first file-locs)))
                   (when (not (equal filename ""))        ;;;          FIX?????
                     (let ((visiting (lsp--buffer-for-file filename))
                           (fn (lambda (loc)
                                 (lsp--xref-make-item filename
                                                      (if location-link (or (gethash "targetSelectionRange" loc)
                                                                      (gethash "targetRange" loc))
                                                        (gethash "range" loc))))))
                       (if visiting
                           (with-current-buffer visiting
                             (seq-map fn (cdr file-locs)))
                         (when (file-readable-p filename)
                           (with-temp-buffer
                             (insert-file-contents-literally filename)
                             (seq-map fn (cdr file-locs))))))))))
      (apply #'append
             (if (gethash "uri" (seq-first locations))
                 (seq-map
                  (-rpartial #'get-xrefs-in-file nil)
                  (seq-group-by
[...]

PR #1115, opened 2019-10-18
Be robust to empty uri's returned in textDocument/references response

JohnC32:
See issue,https://github.com/emacs-lsp/lsp-mode/issues/1113 where ccls is returning an invalid
entry in the response. I plan on also working with the ccls maintainers to fix the root cause.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3586,20 +3586,25 @@ https://microsoft.github.io/language-server-protocol/specification#textDocument_
                  (file-locs location-link)
                  (let ((filename (seq-first file-locs)))
                    (if (equal filename "")
-                       (lsp-log "lsp--locations-to-xref-items, unexpected empty filename")
-                     (let ((visiting (lsp--buffer-for-file filename))
-                           (fn (lambda (loc)
-                                 (lsp--xref-make-item filename
-                                                      (if location-link (or (gethash "targetSelectionRange" loc)
-                                                                      (gethash "targetRange" loc))
-                                                        (gethash "range" loc))))))
-                       (if visiting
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3584,20 +3584,22 @@ https://microsoft.github.io/language-server-protocol/specification#textDocument_
...
-                                                     (gethash "range" loc))))))
-                   (if visiting
-                       (with-current-buffer visiting
-                         (seq-map fn (cdr file-locs)))
-                     (when (file-readable-p filename)
-                       (with-temp-buffer
-                         (insert-file-contents-literally filename)
-                         (seq-map fn (cdr file-locs))))))))
+                 (let ((filename (seq-first file-locs)))
+                   (if (equal filename "")
```
yyoncho: Instead of checking whether the file is "" which looks more like a patch can you change the code to a more generic solution, something like:

``` emacs-lisp
(conditional-case err 
  ;;;   do processing 
 (error (lsp--warn "Failed to process xref entry with message: %s" (error-message-string err))
 (file-error (lsp--warn ...
```

the point is here to guard against any invalid files not only "".

JohnC32: I updated the function to be more generic.

lsp-mode.el
```diff
@@ -3584,20 +3584,27 @@ https://microsoft.github.io/language-server-protocol/specification#textDocument_
...
```
yyoncho: Can you delete this if - it will be handled from the conditional-case bellow ("" is also invalid file name).

yyoncho:
Thank you!

PR #1118, opened 2019-10-19
bump version

brotzeit:
The rust-analyzer support of rustic now depends on lsp-mode so a new release would be good.


yyoncho:
This is fine, but AFAIK the melpa versions do not work as expected since the version generated by melpa(e. g. 20190311.2152) is always bigger - https://github.com/magnars/dash.el/pull/277#issuecomment-482494706 

As a side note feel free to link rustic from lsp-mode README.

brotzeit:
I would add some notes about rust-mode/rustic and the rust related lsp-mode commands to the wiki ?

yyoncho:
Sure.

PR #1119, opened 2019-10-20
add lsp-rust-analyzer-syntax-tree

brotzeit:


brotzeit:
A review is always helpful.

PR #1120, opened 2019-10-20
Handle inactive lsp-lens-mode in lsp-avy-lens. Fix a few typos.

JohnC32:


lsp-mode.el
```diff
@@ -6005,36 +6005,38 @@ This avoids overloading the server with many files when starting Emacs."
...
-                                                  new-str
-                                                (concat new-str "\n"))))
-                                (overlay-put ov 'before-string new-str)))
-                            (lambda ()
-                              (--map (overlay-put it 'before-string
-                                                  (overlay-get it 'lsp-original))
-                                     lsp--lens-overlays))))))
-    (funcall-interactively action)))
+  (if (not lsp-lens-mode)
+      (message "lsp-lens-mode not active")
```
yyoncho: I will accept this as it is but it still needs more work. We should iterate over all visible buffers and collect the lenses from them (this check will go there).

PR #1121, opened 2019-10-21
Remove experimental gopls flags.

muirrn:
It is nice to have first-class config knobs, but once people start
using these experimental settings it will be hard to change or remove
them from from lsp-go.el as gopls changes.

To use experimental settings, users can configure them directly:

(lsp-register-custom-settings '(("gopls.awesomeFeature" "awesomeValue")))

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -78,62 +78,11 @@ completing function calls."
   :risky t
   :package-version '(lsp-mode "6.2"))
 
-;; EXPERIMENTAL! ヽ(⌐■_■)ノ♪♬ At time of writing, these options are considered
-;; experimental, and might either change or become durable. For more, see the
-;; official docs:
-;; https://github.com/golang/tools/blob/master/gopls/doc/settings.md#experimental
-;; -- @gastove 2019-10-09
-(defcustom lsp-gopls-experimental-disabled-analyses []
-  "A list of names of analysis passes that should be disabled."
-  :type 'lsp-string-vector
-  :group 'lsp-gopls
-  :risky t
-  :package-version '(lsp-mode "6.2"))
-
-(defcustom lsp-gopls-experimental-staticcheck nil
-  "If true, enables the use of staticcheck.io analyzers. Note
-  that users also have the option of using staticheck.io
-  analyzers through flycheck."
-  :type 'bool
-  :group 'lsp-gopls
-  :risky t
-  :package-version '(lsp-mode "6.2"))
-
-(defcustom lsp-gopls-experimental-completion-documentation nil
-  "If true, documentation is returned alongside completion candidates"
-  :type 'bool
-  :group 'lsp-gopls
-  :risky t
-  :package-version '(lsp-mode "6.2"))
-
-(defcustom lsp-gopls-experimental-complete-unimported nil
-  "If true, the completion engine is allowed to make suggestions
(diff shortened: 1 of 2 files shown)


seagle0128:
Looks fine

yyoncho:
@Gastove can you share your opinion?

I am not using go so leaving up to you to decide how to handle the PR. My 2c - if vscode is exposing them it is fine for us to expose them too and we could manage them the same way vscode does. E. g. Experimental -> Deprecated -> Removed or Experimental -> Stable.

muirrn:
> if vscode is exposing them

As far as I know in VSCode you manually add them to your JSON config file. They are not predefined anywhere.

yyoncho:
> > if vscode is exposing them
> 
> As far as I know in VSCode you manually add them to your JSON config file. They are not predefined anywhere.

AFAIK @Gastove has generated the settings from the vscode package.json which just unifies what vscode will send. So, if vscode delete certain property - we will delete it too. If someone has this property in his personal config it will work the same in both vscode and emacs. If I was using go(and I was a beginner) I would like to have an easy way to find out what are all of the settings without the need to inspect the server code. So, if we are going to delete the properties IMO we should have a wiki with the experimental properties or link to the corresponding server resource.

Gastove:
@yyoncho ah, I did _not_ generate the setting from VS Code; I had already finished generating them from `gopls`'s own documentation before I found out we had a VS Code converter. 

Personally, I prefer keeping the `defcustoms` as is, but that's a loosely-help preference; I'm not going to fight very hard to protect them. My argument in favour of keeping them goes: `lsp-register-custom-settings` is _not_ richly documented. Would I specify an array as a JSON literal? A vector? Does any kind of serialization occur? If I need to pass a var, now it's on me to set one up correctly and hope I get the types right. 

So, I'd say: if we're going to drop these, we should provide solid docs on how users should set them themselves, and perhaps pursue @yyoncho's proposal for a first-class adhoc system. I'm happy to help, say, draft a wiki page describing how to use `lsp-register-custom-settings` in a users own configs; I don't know that I'll have enough time to help with the adhoc config mechanism _soon_, but I'm glad to pitch in on that too.

muirrn:
> generated the settings from the vscode package.json

VSCode does not have any knowledge of these options, so I don't think that is possible.

The options are documented here: https://github.com/golang/tools/blob/master/gopls/doc/settings.md

There is an emacs page here: https://github.com/golang/tools/blob/master/gopls/doc/emacs.md

We should update the emacs page to include instructions for setting custom settings and then we can link to it in lsp-go.el and the README.

yyoncho:
> VSCode does not have any knowledge of these options, so I don't think that is possible.

Yes, you are right. Currently the vscode package.json does not expose these settings. What you are proposing makes sense to me. @Gastove?

Gastove:
Fine by me :+1: I also still think both a doc in the internal wiki and a richer adhoc config system are good ideas.

yyoncho:
Thank you all, I am glad that we have productive discussions.

seagle0128:
I tested with the latest build, and all error messages disappear.

yyoncho:
cc @vibhavp regarding https://github.com/emacs-lsp/lsp-mode/commit/a6ec2006400518e5c7851c67ba11483de07709b5

PR #1122, opened 2019-10-21
Fix deduping in lsp-register-custom-settings

muirrn:
Dedupe by the setting's path rather than the entire tuple. This avoids
ending up with duplicating settings that end up in the wrong order so
the newer value doesn't actually take effect. I swapped the arg order
to "append" so that the new setting in "props" replaces any existing
setting.


yyoncho:
Thank you!

PR #1124, opened 2019-10-22
Add test for PHP Language Server to check the existence of the php script

### LINKED ISSUE #1123: Incorrect determination of the availability of php-language-server
The default of `lsp-clients-php-server-command` is   `("php" ,(expand-file-name "~/.composer/vendor/felixfbecker/language-server/bin/php-language-server.php"))`, but in lsp-mode.el, its availability is determined by the existence of `php`. In other words, for someone who has php installed but has no interest in lsp for php, this will always lead to a false positive.

I understand that it can be customized out, but it is more like a hack and by default the false positive shouldn't be present for most people who only occasionally opens php files and do not want to set up lsp for php. IMO, the customization should be for someone who installs php-language-server.php in a different location.

This can be reproduced by opening a php file, where `php` is available but `php-language-server.php` is unavailable. `*lsp-log*` buffer shows:

```
Command "intelephense --stdio" is not present on the path.
Command "php /home/hong/.composer/vendor/felixfbecker/language-server/bin/php-language-server.php" is present on the path.
Found the following clients for /home/hong/tmp/a.php: (server-id php-ls, priority -2)
The following clients were selected based on priority: (server-id php-ls, priority -2)
```

xuhdev:
Currently only the existence of the executable, which is likely "php", is checked to determine
whether PHP Language Server is available. This commit adds check for the existence of
php-language-server.php when applicable.

Close #1123

cc @yyoncho @yevgnen

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -300,11 +300,26 @@ particular FILE-NAME and MODE."
   `("php" ,(expand-file-name "~/.composer/vendor/felixfbecker/language-server/bin/php-language-server.php"))
   "Install directory for php-language-server."
   :group 'lsp-php
-  :type 'file)
+  :type '(repeat string))
+
+(defun lsp-php--create-connection ()
+  "Create lsp connection."
+  (plist-put
+   (lsp-stdio-connection
+    (lambda () lsp-clients-php-server-command))
+   :test? (lambda ()
+            (if (and (cdr lsp-clients-php-server-command)
+                     (eq (string-match-p "php[0-9.]*\\'" (car lsp-clients-php-server-command)) 0))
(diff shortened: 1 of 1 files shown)


yyoncho:
This change is breaking the CI byte compilation tests: 

```
In toplevel form:
lsp-clients.el:309:16:Error: reference to free variable ‘lsp-php-server-command’
...............Test lsp-byte-compilation-test backtrace:
```

xuhdev:
Allow me to update this PR? :)

yyoncho:
> Allow me to update this PR? :)

Sorry, that was misclick.

xuhdev:
@yyoncho I've updated the PR. Please review. Thanks!

yyoncho:

PR #1128, opened 2019-10-23
Improve lsp-describe-thing-at-point, fix typo in error handler

JohnC32:
1. Make lsp-describe-thing-at-point to be less obtrusive
 - window height is now min(msg-height,10)
 - use it's own *lsp-mode* buffer to not cause confusion with the *Help* window

2. Fix missing arg to lsp-warn in error handler

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -495,7 +495,7 @@ finding the executable with `exec-path'."
   :group 'lsp-kotlin
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-kotlin-compiler-jvm-target "default"
+(defcustom lsp-kotlin-compiler-jvm-target "1.8"
   "Specifies the JVM target, e.g. \"1.6\" or \"1.8\""
   :type 'string
   :group 'lsp-kotlin


lsp-mode.el
```diff
@@ -3682,11 +3682,11 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
...
                             (gethash "contents"))))
-
     (if (and contents (not (equal contents "")))
-        (with-current-buffer (help-buffer)
-          (with-help-window (current-buffer)
-            (insert (lsp--render-on-hover-content contents t))))
+        (with-current-buffer (get-buffer-create "*lsp-help*")
+          (with-help-window "*lsp-help*"
+            (insert (string-trim-right (lsp--render-on-hover-content contents t))))
+          (fit-window-to-buffer (get-buffer-window "*lsp-help*") 10 1))
```
yyoncho: This is personal preference, and it should go in the lsp core(e. g. someone might prefer windows being always the same size, and so on). There are packages like purpose-mode that can handle this kind of customizations.

JohnC32: Proposal:

1. Use a new "\*lsp-help\*" buffer that has the standard actions, e.g. q to quit and doesn't get confused with the regular "\*Help\*" buffer. As I've done in the pull request.
2. Give a customization option to fit the "\*lsp-help\*" buffer to the height of the text and have this off by default, i.e. disable the call to fit-window-to-buffer.

would that be in alignment with LSP / Emacs general guidelines?

Thanks

yyoncho: > What about the change to not confuse with the standard Help system? 

This make sense. 

> would that be in alignment with LSP / Emacs general guidelines?

IMO the emacs way here (although I am not considering myself as an elisp expert) is to do not provide any customization and let the user do that using the emacs customization mechanisms. On the other hand, at lsp-mode side, we favour usability over emacs traditions. In this particular case, I believe that introducing settings will do more hurt. For example, in spacemacs there is customization help windows to display on the bottom with a particular size using purpose-mode. If we introduce these settings on our side this will introduce an inconsistency on spacecemacs side. Yes, we could have defaults nil which by default will ignore the width/height settings. I would suggest you to use defadvice and customize the help window on your side and eventually add FAQ/wiki entry

PR #1134, opened 2019-10-24
default lsp-gopls-build-flags to an empty list

jmhodges:
With the `['-tags']` default and a non-default GOPACKAGESDRIVER for gopls,  I was getting errors like "invalid options syntax -tags" because `-tags` isn't a legit option for the build system that driver integrates with.


psanford:
Yes this change makes sense.

muirrn:
Looks great, thanks!

PR #1140, opened 2019-10-26
Ignore files and blacklisted folders that do not belong to the same host

nicber:
This avoids a potential tramp connection to a host that may be unreachable.

Related to #1019


yyoncho:

I had that piece in my local copy but somehow I have forgotten to push it.

PR #1141, opened 2019-10-26
README.md: Update gopls download URL, remove old link

0x5d:
Remove the broken link pointing to the Go wiki. 
Update the `go get` command URL


psanford:
The official gopls install instructions moved to here: https://github.com/golang/tools/blob/master/gopls/doc/user.md. They still say to use the '@latest' version to get the latest stable version: `go get golang.org/x/tools/gopls@latest`.

I think we want to update the url instead of deleting it and keep the `go get` command as it was.

muirrn:
Can you update the link to point to https://github.com/golang/tools/blob/master/gopls/doc/user.md#installation?

The instructions there say

> If you do want to get the latest stable version of gopls, change to any directory that is both outside of your GOPATH and outside of a module (a temp directory is fine), and run
>
> `go get golang.org/x/tools/gopls@latest`
>
> Do not use the -u flag, as it will update your dependencies to incompatible versions.

README.org
```diff
@@ -146,7 +146,7 @@
...
    | Fortran               | [[https://github.com/hansec/fortran-language-server][fortran-language-server]]                   | Yes            | pip install fortran-language-server                                                                     | Yes                          |
-   | Go                    | [[https://golang.org/x/tools/cmd/gopls][gopls]]                                     | Yes            | [[https://github.com/golang/go/wiki/gopls][gopls]] go get golang.org/x/tools/gopls@latest                                                            | Yes                          |
+   | Go                    | [[https://golang.org/x/tools/cmd/gopls][gopls]]                                     | Yes            | go get -u golang.org/x/tools/cmd/gopls                                                            | Yes                          |
```
seagle0128: According to https://github.com/golang/tools/blob/master/gopls/doc/user.md,
should use `go get golang.org/x/tools/gopls@latest`.

> Do not use the -u flag, as it will update your dependencies to incompatible versions.

0x5d: Done! Sorry, I'm using GOPATH, so the @latest part fails with
```$ go get golang.org/x/tools/gopls@latest 
go: cannot use path@version syntax in GOPATH mode```

But I guess it should be ok for many people.

seagle0128:
Thank you!

PR #1142, opened 2019-10-26
Rust analyzer

brotzeit:


lsp-rust.el
```diff
@@ -406,6 +397,15 @@ PARAMS progress report notification data."
           (pop-to-buffer buf)))
     (message "rust-analyzer not running.")))
 
+(defun lsp-rust-analyzer-join-lines ()
```
yyoncho: Can you document what this method do?

PR #1146, opened 2019-10-30
Fix missing arg to lsp-warn, remove confusion with the help system.

JohnC32:
lsp-describe-thing-at-point now uses it's own *lsp-help* buffer to display content, rather than
using the *Help* buffer. This avoids confusion with the Emacs help system. For example, the prior
revision of this function, used the *Help* buffer and this would show a 'back' button which did not
work.


yyoncho:
Thank you!

PR #1147, opened 2019-10-31
Lsp Terraform

### LINKED ISSUE #878: Support for Terraform?
There is a Terraform LSP in development here: https://github.com/juliosueiras/terraform-lsp

It seems to be not super far along but it's something.

Gastove:
This PR brings in `lsp` configs for working with Terraform configuration files. This configuration is backed by [terraform-lsp](https://github.com/juliosueiras/terraform-lsp). 

`terraform-lsp` is... fairly limited. I've been using this at my day job, and it's still helpful, but it's nowhere near as fully featured as, say, `gopls` or `rls`. 

Also true: I cannot find a good way to auto-install `terraform-lsp`. It requires a fully-functional `go` install toolchain (a modern one, using `go mod`, meaning go version 1.11 or greater). I don't know if there's a typical or correct mechanism for reporting missing language servers, or if it's built in! Do let me know?

Closes #878.

### DIFF
--- a/lsp-terraform.el
+++ b/lsp-terraform.el
@@ -49,7 +49,8 @@
 (defun lsp-terraform--make-launch-cmd ()
   (-let [base `(,lsp-terraform-server)]
     (if lsp-terraform-enable-logging
-        (append base "-enable-log-file"))))
+        (push "-enable-log-file" base))
+    base))
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection #'lsp-terraform--make-launch-cmd)
--- a/lsp-terraform.el
+++ b/lsp-terraform.el
@@ -29,19 +29,30 @@
 (defgroup lsp-terraform nil
   "LSP support for Terraform, using terraform-lsp"
   :group 'lsp-mode
-  :link '(url-link "https://github.com/juliosueiras/terraform-lsp"))
+  :link '(url-link "https://github.com/juliosueiras/terraform-lsp")
+  :package-version `(lsp-mode . "6.2"))
 
 (defcustom lsp-terraform-server "terraform-lsp"
   "Path to the `terraform-lsp' binary."
   :group 'lsp-terraform
   :risky t
-  :type 'file)
+  :type 'file
+  :package-version `(lsp-mode . "6.2"))
 
-(defun lsp-terraform-server-start ()
-  ())
+(defcustom lsp-terraform-enable-logging nil
+  "If non-nil, enable `terraform-ls''s native logging."
+  :group 'lsp-terraform
+  :risky t
+  :type 'boolean
+  :package-version `(lsp-mode . "6.2"))
+
+(defun lsp-terraform--make-launch-cmd ()
+  (-let [base `(,lsp-terraform-server)]
+    (if lsp-terraform-enable-logging
+        (append base "-enable-log-file"))))
 
 (lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection #'lsp-terraform-server-start)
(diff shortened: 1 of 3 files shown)


lsp-terraform.el
```diff
@@ -0,0 +1,62 @@
...
+(defcustom lsp-terraform-enable-logging nil
+  "If non-nil, enable `terraform-ls''s native logging."
+  :group 'lsp-terraform
+  :risky t
+  :type 'boolean
+  :package-version `(lsp-mode . "6.2"))
+
+(defun lsp-terraform--make-launch-cmd ()
+  (-let [base `(,lsp-terraform-server)]
+    (if lsp-terraform-enable-logging
```
yyoncho: use `when` when there is one leg.

yyoncho:
If you are familiar with docker it will be great if you can add your server to lsp-docker image and add sample project to test with. It will be useful for both lsp developers(when fixing bugs) and for users(when someone wants to evaluate a server).

lsp-terraform.el
Gastove: Ah, happy to. Wanted to leave a clear place for more configurations in the future, but this total makes sense to me for now :+1:

Gastove:
@yyoncho I'm quite familiar with docker, as it happens; I don't have a good sample project around, but I'll try and knock one together soon. I'll open an issue for this, so it doesn't get lost. Very happy to get that taken care of pretty soon here :+1: 

As for the rest of your feedback:
- I haven't checked what VS Code does, but I'll have a look at that -- good point. I'd be surprised if it had a way around this, but I've been surprised before ;P  
- Happy to update the README :+1: 
- By "missing," I simply meant: if a user must manually install a given language server, and they haven't, do we have an existing way to tell them what they need to do, display instructions, etc?

Gastove:
Ahhh, shoot -- the `terraform-lsp` README doesn't mention or link to their releases, but they _are_ doing normal binary releases after all. This can absolutely be automated (which is what VS Code does). I'll get that wired up.

Gastove:
@yyoncho do we have an existing pattern for how to compute and download platform-specific versions? That is: terraform-lsp doesn't appear to publish a "newest" artefact, a la fsautocomplete; instead it's building platform-specific binaries with version numbers. VS Code is handling this using a Github library called Octokit; do we have any equivalent tooling that I'm just not finding, or should I write something for us? (All the info we need is available from the Github API, it'll just take a bit of parsing.)

yyoncho:
> do we have an existing way to tell them what they need to do, display instructions, etc?

No, we don't - we expect the user to check the info in the README. IMO it makes sense to have it at least for most popular language servers.

> do we have an existing pattern for how to compute and download platform-specific versions? 

Something similar to what FSharp and MS Pyls do?

I will move the rest of the discussion in https://github.com/emacs-lsp/lsp-mode/issues/506 .

Gastove:
Ah, at least for F# and `fsautocomplete`, the `fsautocomplete` team publishes a "latest" version, which isn't platform dependent. I believe MS pyls is similar, but I'm less familiar with it. 

In any case, it sounds like a broader solution is needed, so for now, I'm going to document up the steps to manually install the terraform server and leave it at that until #506 is resolved.

yyoncho:
ms python seems to be platform-dependent: https://github.com/emacs-lsp/lsp-python-ms/blob/master/lsp-python-ms.el#L152

juliosueiras:
@Gastove I added vendor folder, and tested against go 1.10/1.11 , 1.9 doesn't work due to one of the deps uses strings.Builder which is 1.10+ only

P.S. I am working on adding more features and fixing any existing issues with it

PR #1149, opened 2019-10-31
Fix #1138

### LINKED ISSUE #1138: Bad response to rangeFormatting/textEdits?
I'm seeing occasional clangd crashes/aborts where clangd complains about invalid offsets within some source file. This just happened again and looking at the `lsp-log` turned up the following pattern:

1. I'm editing some part of the source file, say I'm inserting a semicolon or pasting code from the clipboard
2. in response to the change just made, `lsp-mode` emits a `rangeFormatting` request referring to version 45 of the source file (I had previously unsuccessfully tried disabling as many formatting options as I could find, but perhaps a similar issue had been caused by other factors, such as change-hook using minor modes; will reinvestigate once this issue is resolved)
3. clangd responds with a `rangeFormatting` response containing 3 TextEdits
4. `lsp-mode` sends out a `didChange` message, the content and location of which are perfectly correct when taking the previous 3 TextEdits into account (I checked by logging all intermediate buffer versions to a temporary variable); also, the reported version is now at 49.

At this point clangd aborts, I guess due to `lsp-mode` failing to send out the 3 intermediate `onChange` messages (?)
When stepping through `lsp--apply-text-edit-replace-buffer-contents` in a similar situation, `lsp-mode` appears to behave correctly but I noticed that it emits the modified buffer wholesale because `lsp--apply-formatting` temporarily sets `lsp--server-sync-method` to `'full`.

So I wonder:
- why temporarily switching to full syncs is necessary? Are incremental changes somehow unreliable when used for buffer formatting?
- if this, in combination with full sync debouncing, causes a race condition between the 3 textEdit responses and whatever `didChange` message should have been sent out in response to my original editing operation? Or maybe switching back to incremental is supposed to flush pending full syncs and something's wrong with my setup (didn't check the code yet to see if that is the case)?

sebastiansturm:
this PR should fix #1138 by flushing delayed changes if the server sync method had temporarily been changed to `'full`. Issue #1138 contains a small test script that can be used to verify the efficacy of this change (tested on a trunk build of clangd-10)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1376,6 +1376,18 @@ already have been created."
            (indent 1))
   `(when-let (lsp--cur-workspace ,workspace) ,@body))
 
+(defmacro with-full-sync (&rest body)
+  "Execute BODY with lsp--server-sync-method set to 'full."
+  (declare (debug (form body))
+           (indent 1))
+  `(let ((no-flush-needed (eq 'lsp--server-sync-method 'full))
+         (lsp--server-sync-method 'full))
+     ,@body
+     ;; the next didChange message to be sent out in incremental mode might
+     ;; overtake the changes caused by BODY due to sync debouncing, so we need
+     ;; to flush before switching back to 'incremental
+     (unless no-flush-needed (lsp--flush-delayed-changes))))
+
 (defun lsp--window-show-message (_workspace params)
   "Send the server's messages to log.
 PARAMS - the data sent from _WORKSPACE."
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1376,6 +1376,18 @@ already have been created."
            (indent 1))
   `(when-let (lsp--cur-workspace ,workspace) ,@body))
 
+(defmacro with-full-sync (&rest body)
```
yyoncho: can you prefix this with `lsp-`?

Issue #1148, opened 2019-10-31
Unify the jobs that run on timer


yyoncho:
We have several things that we do on timer:

1. Highlight
2. Links
3. Lenses
4. Lsp-ui sideline stuff.
5. lsp-treemacs symbols view
6. ?

It will be better if the unify all these events to happen in one service event. This will give us more flexibility and it will allow us to turn off all the background updates in certain cases. For example, we do not want to update the sideline if company is visible.

PR #1152, opened 2019-11-01
display inlay hints with rust-analyzer

brotzeit:
fixes brotzeit/rustic#73


flodiebold:
Note that the inlay hints interact badly with the lsp-ui sideline, because it doesn't seem to consider the overlays in its width calculation, which often leads to lines wrapping around :confused:

brotzeit:
Oh. Thanks for the info. I will add a note.

lsp-rust.el
```diff
@@ -425,6 +430,57 @@ PARAMS progress report notification data."
...
       (message (format "Switched to server %s." server)))))
 
+;; inlay hints
+
+(defvar-local lsp-rust-analyzer-inlay-hints-timer nil)
+
+(defun lsp-rust-analyzer-update-inlay-hints (buffer)
+  (if (and (lsp-rust-analyzer-initialized?)
+           (eq buffer (current-buffer)))
+      (lsp-send-request-async
```
yyoncho: can you switch to lsp-request-async? lsp-send-request-async is deprecated(although there is no indication for that). This is an artefact from old lsp-mode. I will add deprecation notice.

lsp-rust.el
```diff
@@ -425,6 +430,57 @@ PARAMS progress report notification data."
                          (* (lsp--client-priority (gethash server lsp-clients)) -1)))
       (message (format "Switched to server %s." server)))))
 
+;; inlay hints
+
+(defvar-local lsp-rust-analyzer-inlay-hints-timer nil)
+
+(defun lsp-rust-analyzer-update-inlay-hints (buffer)
+  (if (and (lsp-rust-analyzer-initialized?)
```
yyoncho: here we should update only if the current buffer has changed(I assume that the hints won't change if there is no change in the buffer content). I filed a task to unify all tasks that run on timer - https://github.com/emacs-lsp/lsp-mode/issues/1148 so you could leave it as it is and once we solve #1148 we may change the code to use the newly introduced hook.

yyoncho:
I would suggest adding also a functions to show/hide the hints one time.

marcbowes:
> 
> 
> Note that the inlay hints interact badly with the lsp-ui sideline, because it doesn't seem to consider the overlays in its width calculation, which often leads to lines wrapping around 😕

Is there a issue tracking this? I was not able to find one. Seems like this is one of the biggest pieces of functionality missing in the out-of-the-box experience (for example, comparing what VSCode gives vs. Emacs).

Issue #1160, opened 2019-11-05
Support stopped servers


MaskRay:
https://github.com/MaskRay/ccls/wiki/Debugging If you set the environment variable `CCLS_TRACEME=s` (`raise(SIGSTOP)`), ccls will stop itself before handling client requests.

If you do this, however, after `M-x lsp`, you will immediately receive a prompt `Server ccls:16154 status:starting exited with status signal. Do you want to restart it? (y or n) `. This makes lsp-mode unsuitable for debugging initialization problems of a language server.

Any tips debugging lsp-mode code will be highly appreciated. I try to do:

* ` C-u C-M-x` at `lsp--restart-if-needed`  (emacs will automatically set a breakpoint at the entry. I don't want this implicit behavior, but I don't know how to suppress it)

and use edebug, but my edebug session is frequently interrupted by the ` Do you want to restart it? (y or n) .` prompt.

Note: a stopped server used to work (more than 1 year ago) when the auto restarting feature was not implemented.

yyoncho:
Isn't `(setq lsp-restart 'ignore)` what you are looking for?

MaskRay:
but I want to auto-restarting functionality. I just don't want it to consider a stopped process exited/signaled.

yyoncho:
I see - seems like an issue in `lsp--create-sentinel` - we probably misinterpret sentinel event.

yyoncho:
I added tracing of the signal and the signal received when traceme variable is set are `'stop` and `'signal` so I believe the proper fix is https://github.com/yyoncho/lsp-mode/commit/785497ce2d59e9721f7ae4831682ca934c190f1f

PR #1161, opened 2019-11-05
Make exit handling more robust

yyoncho:
- fixes 1160

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4656,10 +4656,10 @@ textDocument/didOpen for the new file."
 
 (defun lsp--send-no-wait (message proc)
   "Send MESSAGE to PROC without waiting for further output."
-  (when (memq (process-status proc) '(stop exit closed failed nil))
-    (error "%s: Cannot communicate with the process (%s)" (process-name proc)
-           (process-status proc)))
-  (process-send-string proc message))
+  (condition-case err
+      (process-send-string proc message)
+    ('error (lsp--error "Sening to process failed with the following error: %s"
+                        (error-message-string err)))))
 
 (define-error 'lsp-parse-error
   "Error parsing message from language server" 'lsp-error)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -5388,41 +5388,41 @@ returns the command to execute."
 (defun lsp--create-sentinel (workspace)
   "Create sentinel handler for WORKSPACE."
   (lambda (process exit-str)
-    (let* ((status (process-status process))
-           (folder->workspaces (lsp-session-folder->servers (lsp-session)))
-           (stderr (-> workspace lsp--workspace-proc process-name get-buffer)))
+    (unless (process-live-p process)
```
MaskRay: Is there any early-return mechanism to reduce indentation below?

lsp-mode.el
```diff
@@ -5388,41 +5388,41 @@ returns the command to execute."
...
+          ;; in the workspace.
+          (--each (lsp--workspace-buffers workspace)
+            (when (buffer-live-p it)
+              (with-current-buffer it
+                (setq lsp--buffer-workspaces (delete workspace lsp--buffer-workspaces))
+                (lsp--uninitialize-workspace)
+                (lsp--spinner-stop)
+                (lsp--remove-cur-overlays))))
+
+          ;; cleanup session from references to the closed workspace.
```
MaskRay: We can capitalize comments since we are updating them.

MaskRay:
Thanks for the patch. I confirm it addresses #1160. Yesterday while I was debugging, I noticed that `lsp--send-no-wait` also checks process-status. For a language server that is stopped after it has run for a while, does `lsp--send-no-wait` or somewhere else need a similar update?

lsp-mode.el
yyoncho: AFAIK there is early return using cl-lib, I refactored the code to use closure.

lsp-mode.el
yyoncho: Done, but I think we do not follow this style guide.

yyoncho:
> I noticed that lsp--send-no-wait also checks process-status. For a language server that is stopped after it has run for a while, does lsp--send-no-wait or somewhere else need a similar update?

I changed the code to use conditional-case instead of checking the status of the process upfront. It this should have some positive effect on performance and it will handle your concern.

lsp-mode.el
```diff
@@ -1536,33 +1536,33 @@ WORKSPACE is the workspace that contains the diagnostics."
       "Report new diagnostics to flymake."
       (funcall lsp--flymake-report-fn
                (-some->> (lsp-diagnostics)
-                 (gethash buffer-file-name)
```
seagle0128: Bad formats here 😄

yyoncho: They are correct but it is GH rendering the indentation in an odd way. Anyway, I removed the irrelevant changes.

PR #1163, opened 2019-11-08
Provide option to configure which clients to be required

yyoncho:
This will simplify client configuration.


yyoncho:
cc @alanz @seagle0128 @MaskRay

alanz:
I guess I should move `lsp-haskell` in to the mothership.

MaskRay:
I find myself quite happy with the old lsp-mode + lsp-foobar approach. If a user wants to use LSP for a specific language, enable that explicitly. The new layout reminds me of the llvm svn -> git monorepo transition. Can you list a few advantages about the new layout?

yyoncho:
> The new layout reminds me of the llvm svn -> git monorepo transition. Can you list a few advantages about the new layout?

I don't want to change the layout, I want to keep it as it is but I want to avoid the `(require 'lsp-foobar)` line that user has to add since it is a source of issues/bugs. WDYT?

yyoncho:
I think we could do better: We could introduce a variable `lsp-required-packages` and set it to `(list lsp-clients ccls lsp-pyls...)`. Then, the user could enable/disable the packages to be loaded via adding/removing and at the same time it will work out of the box without any configuration and at the same time, we will allow customizing what will be loaded.

noctuid:
If the user doesn't have any of the external packages installed, it's not going to add any extra time to loading `lsp-clients` (and I imagine loading all of them wouldn't take a ton of time anyway). If it's configurable I think this makes sense as a default.

That said, have you considered making `lsp-auto-require-clients` smarter? It doesn't make sense to me that calling `lsp` loads every client. Is the reason for this that it would be error-prone to, for example, only load `lsp-pyls` (and now `lsp-python-ms`) if the `major-mode` was `python-mode`? Or is it just not worth the extra work to do something like this for a bunch of small files?

yyoncho:
@noctuid 

ATM the association buffer -> client happens either by :activation-fn or by :major-modes. The activation-fn is used in order to handle the clients like flowjs which are activated when the buffer contains a particular string. I guess what you are proposing is to simplify that and introduce major-mode -> client association and keep it in lsp-mode? We could also try to generate autoloads, e. g. register the client when lsp-mode is loaded and when a particular major mode is loaded(not sure whether this is technically possible). Do you have a concrete proposal on how to achieve the smartness?

What do you think about introducing another variable controlling which packages to be automatically required? This is simple to implement, works out of the box for beginners and also advanced users could turn off particular clients. I think that this will be an improvement over what we currently have because lsp-clients is autoloading everything and you could not load one client from lsp-clients and lsp-pyls.el. Of course, this is not as good as having this working out of the box but as you mentioned it might not pay the effort.

brotzeit:
> I want to keep it as it is but I want to avoid the (require 'lsp-foobar) line that user has to add since it is a source of invalid bug reports

I think lsp-foobar repos should have installation instructions to avoid this kind of bug reports. Most users who don't know how to figure this out usually use doom-emacs(or whatever) ? In my experience it's enough to document it.

yyoncho:
@brotzeit I agree with you that it is enough to document it for the traditional emacs users which are used to read the README before starting to use a package. But based on the fact that we have a lot of users that just start out with emacs and even don't know what require is I still want to provide something that works out of the box for those who don't use spacemacs/doom. I have created lsp-ide project but still, one year later it is not clear to me how to structure it. Let say we want to:

1. Have everything working out of the box for the beginners and be as user-friendly as possible(be like vscode).
2. Keep the traditional emacs users who want to be in control of everything happy.

brotzeit:
IMO this is a tradeoff. We can't be _like_ vscode since emacs has a different philosophy, but we can try to take some good ideas of vscode and add it in an emacs way. People who want absolutely everything working out of the box won't use emacs anyway.

yyoncho:
I have implemented what I have described in https://github.com/emacs-lsp/lsp-mode/pull/1163#issuecomment-552079966 . I think it is improvement over what we currently have. Please share your opinion.

seagle0128:
I like the new approach. It's out of box, meanwhile the user can customize the clients.

I remember the author of `eglot` claimed its advantage is out-of-the-box.

MaskRay:
The way that eglot works out-of-the-box is that it contains a minimum configuration for each client, just the server command line. That is all.

I have some difficult time to understand why we consider the following 

```elisp
(use-package lsp-mode :commands lsp)
(use-package foobar :hook (foobar-mode) . (lambda () (require 'foobar) (lsp))))
```

to be frowned upon by beginners. 1) For Emacs distributions, users get mostly out-of-box experience. 2) For users who don't use a distribution, the `README.md` in the foobar repository gives the minimum configuration. As brotzeit said, "People who want absolutely everything working out of the box won't use emacs anyway." One nice thing with an independent repository is that its homepage can be rendered in a separate page. I think it might be useful to do a survey: how many people considered lsp-mode/README.md too long to read.

If a client has a minimum configuration, putting its id into a list, say, `lsp-auto-require-clients`, will not look too bad. I'll take it as simplifying `(use-package foobar :hook (foobar-mode) . (lambda () (require 'foobar) (lsp))))` to `(setq lsp-auto-require-clients '(... foobar))`

For out-of-tree clients (e.g. lsp-haskell, ccls, and cquery), users will still customize `lsp-haskell-*` or `ccls-*`, I don't follow how `lsp-auto-require-clients` can simplify things.

Keeping things simple and stupid is that, but I wonder whether we actually made things aligned with our good intention:

1. Long `README.md`
2. lsp-foobar repository -> `lsp-mode/lsp-foobar.el`
2a. `lsp-clients.el`
3. Forwarding initialization options as `(defcustom lsp-foobar-* ...)`

yyoncho:
I disagree with the statement that it is a tradeoff between traditional users and those who want everything to work out of the box. I think that if we organize the packages in a good way we could keep both groups happy. 

I disagree also with the statement that emacs cannot provide the similar(or better?) the out of the box experience as vscode. I agree with all of the points in @zaiste talk at `Emacs Conf 2019` starting https://youtu.be/V-CX3CHsw5E?t=36559 . I think that providing the out of the box experience comparable to vscode is much easier than providing the comparable to vim's modal editing experience. 

I believe that we emacs-lsp organization should provide a package (not necessary in lsp-mode repo) that provides the same lsp experience that you get in spacemacs and doom for those who don't want to use doom/spacemacs and at the same time allow advanced users to glue their own version of the package. Of course, this PR is a very small step in that direction and it might not be even needed at all.

@MaskRay 

> Long README.md

IMO there is no need from a poll - If you can suggest a better structure we will implement it. 

> lsp-foobar repository -> lsp-mode/lsp-foobar.el
> 2a. lsp-clients.el

IMO we could extract a package `lsp-core`, and keep the auto-configuration code + client definitions in `lsp-mode` so if someone does not want to use the clients he/she could use the lsp-core package. If you are suggesting to split lsp-mode into separate into 43 repos(number of clients in lsp-mode) and additionally break the configs of all lsp-mode users I am not willing to do that work unless there is very strong reason to do so.

myrgy:
I has to mention that many times there was a problem with lsp language plugins that depend on lsp-core api. And when we have all that modules in different repos it became very difficult to maintain api changes.

kiennq:
I wonder what's the status of this PR?
I would really love to this being merged. Having require all clients for just `typescript` seems really overkill to me.

yyoncho:
> I wonder what's the status of this PR?

I will resurrect the PR once https://github.com/emacs-lsp/lsp-mode/pull/1384 is merged.

yyoncho:
@kiennq synched with latest lsp-mode.

PR #1164, opened 2019-11-09
Check for TextDocumentSyncKind in sync checks. Fix #1162

### LINKED ISSUE #1162: Invalid save notification support checks.
**Describe the bug**
`didSave`/`willSave` (functions like `lsp--send-did-save-p`) server capabilities checks not follow current specification.

`textDocumentSync` allows `TextDocumentSyncOptions` (used in this checks right now) or number from 0 to 2 (`TextDocumentSyncKind`). For example elm-language-server sending `1` here which means Full sync so functions like `lsp--send-did-save-p` should return `t`. 

**Expected behavior**
Handle `TextDocumentSyncKind` numbers in this checks.

elm-language-server

**OS**
MacOS

ssbb:
Fixes #1162

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2722,34 +2722,34 @@ in that particular folder."
 (defun lsp--send-open-close-p ()
   "Return whether open and close notifications should be sent to the server."
   (let ((sync (gethash "textDocumentSync" (lsp--server-capabilities))))
-    (and (or (memq sync '(1 2))
-             (and (hash-table-p sync) (gethash "openClose" sync))))))
+    (or (memq sync '(1 2))
+        (and (hash-table-p sync) (gethash "openClose" sync)))))
 
 (defun lsp--send-will-save-p ()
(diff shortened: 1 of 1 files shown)


ssbb:
I have pretty basic Emacs Lisp knowledge, can you please check everything is good here?

lsp-mode.el
```diff
@@ -2722,33 +2722,34 @@ in that particular folder."
 (defun lsp--send-open-close-p ()
   "Return whether open and close notifications should be sent to the server."
   (let ((sync (gethash "textDocumentSync" (lsp--server-capabilities))))
-    (and (hash-table-p sync)
-         (gethash "openClose" sync))))
+    (and (or (memq sync '(1 2))
```
yyoncho: It will work as it is but I think that the outside `and` is not needed(it has only one argument).

ssbb: Just removed it 👍

yyoncho:

PR #1166, opened 2019-11-10
Add lsp-erlang-server-connection-type custom var to specify between 'tcp and 'stdio

jfacorro:
erlang-ls/erlang_ls#16

### DIFF
--- a/lsp-erlang.el
+++ b/lsp-erlang.el
@@ -38,12 +38,25 @@
   :risky t
   :type 'file)
 
+(defcustom lsp-erlang-server-connection-type
+  'tcp
+  "Type of connection to use with the Erlang Language Server: tcp or stdio"
+  :group 'lsp-erlang
+  :risky t
+  :type 'string)
+
 (defun lsp-erlang-server-start-fun (port)
   `(,lsp-erlang-server-path
-    ,(number-to-string port)))
+    "--transport" "tcp"
+    "--port" ,(number-to-string port)))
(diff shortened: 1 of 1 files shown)


jfacorro:
@robertoaloi

lsp-erlang.el
```diff
@@ -38,12 +38,25 @@
...
+  :risky t
+  :type 'string)
+
 (defun lsp-erlang-server-start-fun (port)
   `(,lsp-erlang-server-path
-    ,(number-to-string port)))
+    "--transport" "tcp"
+    "--port" ,(number-to-string port)))
+
+(defun lsp-erlang-server-connection ()
```
yyoncho: Here you could return a function that will return the function to create the connection based on the connection-type.

yyoncho: (not necessary to address).

yyoncho:
Looks good to me, I will wait for @robertoaloi to review.

lsp-erlang.el
```diff
@@ -38,12 +38,25 @@
   :risky t
   :type 'file)
 
+(defcustom lsp-erlang-server-connection-type
+  'tcp
+  "Type of connection to use with the Erlang Language Server: tcp or stdio"
+  :group 'lsp-erlang
+  :risky t
+  :type 'string)
```
stormwatch, line 46: type should be a symbol, not a string.

PR #1167, opened 2019-11-10
Added support for HDL Checker server to lsp-vhdl

chrbirks:
Defaults to use vhdl-tool server but added defcustom to select between the two. Using path to server if explicitly set and assumes it's in PATH otherwise.
Added extra check for hdl_checker binary since it's not automatically executable when downloading a new version.

### DIFF
--- a/lsp-vhdl.el
+++ b/lsp-vhdl.el
@@ -34,31 +34,69 @@
 
 (require 'lsp-mode)
 
-(defgroup lsp-vhdl nil
-  "LSP support for VHDL using the server from http://www.vhdltool.com"
-  :group 'lsp-mode
-  :link '(url-link "http://www.vhdltool.com"))
+(defvar vhdl-tool-bin-name "vhdl-tool"
+  "Name of the VHDL Tool binary.")
 
-(defcustom lsp-vhdl-server-path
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

yyoncho:
I have reverted the PR due to https://github.com/emacs-lsp/lsp-mode/issues/1168 . In this case, the user-errors should be removed from  `lsp-vhdl-vhdl-tool-server` and you should let `lsp-mode` do the check  whether the server exists or not.

chrbirks:
Thanks, I'll make a new pull request without `user-error`. It was to check for execute permissions on the server binary since lsp-mode just reports something along the lines `"Command ... is not present on the path"`.

yyoncho:
The check is fine, but it should not happen on load time but when lsp-mode has matched a vhdl file. 

You could use `:test?` returned by the command itself, similar to what we do here: https://github.com/yyoncho/lsp-mode/blob/refactoring/lsp-xml.el#L199

Let me know if you need a help with that.

chrbirks:
No, I didn't realize that code was run on load time. I'll try to get something working with `:test?`.

PR #1171, opened 2019-11-11
Implement unified way to handle recurring lsp features

### LINKED ISSUE #1148: Unify the jobs that run on timer

yyoncho:
Fixes #1148

Introduced:

- lsp-on-idle-hook - used for recurring tasks that are not related to editing (e.
g. highlights, lenses)
- lsp-on-change-hook - used for tasks that must run after the buffer is
changed(lenses, links, document color TBA)

Additionally performed the following fixes:

- highlights are not deleted if you switch from highlighted symbol to another
highlighted symbol.
- highlights symbol under point is now correctly calculated based on the server
response, not on thing-at-point which may not be the same.
- simplified the cleanup of highlights

Next steps:

* We should migrate lsp-ui to use the same hooks
* We should change company-lsp to disable the hooks when completion is in
progress.
* Simplify the code lenses facilities


yyoncho:
I pushed a few more commits on top of the restructuring of recurring tasks including a rewrite of jsonrpc handling in https://github.com/emacs-lsp/lsp-mode/pull/1171/commits/253a565841b99edfea8bdba279af0cbfa8ad2e18 . It will be great if it can get more testing before it is merged.

PR #1173, opened 2019-11-12
change logging stuff

brotzeit:
I think `special-mode` is better than `view-mode`

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -27,6 +27,7 @@
 (require 'lsp)
 (require 'dash)
 (require 'dash-functional)
+(require 'rx)
 (require 'lsp-pyls)
 (require 'lsp-rust)
 (require 'lsp-solargraph)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5589,6 +5589,14 @@ SESSION is the active session."
 
 ;; lsp-log-io-mode
 
+(defvar lsp-log-io-mode-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "M-n") #'lsp--log-io-next)
+    (define-key map (kbd "M-p") #'lsp--log-io-prev)
+    (define-key map (kbd "k") #'lsp--erase-log-buffer)
+    map)
+  "Keymap for lsp log buffer mode.")
+
 (define-derived-mode lsp-log-io-mode special-mode "LspLogIo"
   "Special mode for viewing IO logs.")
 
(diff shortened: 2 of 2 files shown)


lsp-mode.el
```diff
@@ -5592,36 +5587,58 @@ SESSION is the active session."
...
+      (with-current-buffer (lsp--get-log-buffer-create w)
+        (erase-buffer)))))
+
+(defun lsp--log-io-next (arg)
+  "Move to next log entry."
   (interactive "P")
   (ewoc-goto-next lsp--log-io-ewoc (or arg 1)))
 
-(defun lsp-log-io-prev (arg)
+(defun lsp--log-io-prev (arg)
```
yyoncho: Why do you rename the function? We previously decided to make this public so a user could eventually re-map them.

lsp-mode.el
```diff
@@ -5592,36 +5587,58 @@ SESSION is the active session."
...
+(defun lsp--get-log-buffer-create (workspace)
+  "Return the lsp log buffer of WORKSPACE, creating a new one if needed."
+  (let ((server-id (-> workspace lsp--workspace-client lsp--client-server-id symbol-name))
+        (pid (format "%s" (process-id (lsp--workspace-cmd-proc workspace)))))
+    (get-buffer-create (format "*lsp-log: %s:%s*" server-id pid))))
+
+(defun lsp--erase-log-buffer ()
+  "Delete contents of the project's lsp log buffers."
+  (interactive)
+  (let* ((session (lsp-session))
```
yyoncho: session is not used.

lsp-mode.el
```diff
@@ -5592,36 +5587,58 @@ SESSION is the active session."
...
 
-(defun lsp-log-io-next (arg)
+(defun lsp--get-log-buffer-create (workspace)
+  "Return the lsp log buffer of WORKSPACE, creating a new one if needed."
+  (let ((server-id (-> workspace lsp--workspace-client lsp--client-server-id symbol-name))
+        (pid (format "%s" (process-id (lsp--workspace-cmd-proc workspace)))))
+    (get-buffer-create (format "*lsp-log: %s:%s*" server-id pid))))
+
+(defun lsp--erase-log-buffer ()
+  "Delete contents of the project's lsp log buffers."
```
yyoncho: project -> session.

Binding this to `k` feels a bit unnatural to me. What do you think about having `k` to delete only the current buffer logs and `K` to delete all?

lsp-clients.el
```diff
@@ -27,6 +27,7 @@
 (require 'lsp)
 (require 'dash)
 (require 'dash-functional)
+(require 'rx)
```
MaskRay: rx is not used?

lsp-mode.el
```diff
@@ -5592,36 +5587,58 @@ SESSION is the active session."
...
     (lsp--spinner-stop)))
 
+;; lsp-log-io-mode
+
+(defvar lsp-log-io-mode-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "M-n") #'lsp--log-io-next)
+    (define-key map (kbd "M-p") #'lsp--log-io-prev)
+    (define-key map (kbd "k") #'lsp--erase-log-buffer)
+    map)
```
seagle0128: The key bindings seem not conventional for Emacs.

lsp-mode.el
seagle0128: Agree. Don't change the public APIs.

lsp-clients.el
brotzeit:
```elisp
(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
  "Check if the javascript-typescript language server should be enabled based on FILENAME."
  (or (string-match-p (rx (one-or-more anything) "." (or "ts" "js") (opt "x") string-end) filename)
      (derived-mode-p 'js-mode 'js2-mode 'typescript-mode)))
```

lsp-mode.el
brotzeit: I don't use regular emacs bindings. What would you suggest ?

seagle0128: I would suggest `M-k` to erase the whole buffer.
Or as @yyoncho's suggestion:

>Binding this to k feels a bit unnatural to me. What do you think about having k to delete only the current buffer logs and K to delete all?

lsp-mode.el
```diff
@@ -5592,36 +5587,67 @@ SESSION is the active session."
       (lsp--start-workspace session client project-root (lsp--create-initialization-options session client))
     (lsp--spinner-stop)))
 
+;; lsp-log-io-mode
+
+(defvar lsp-log-io-mode-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "M-n") #'lsp--log-io-next)
```
yyoncho: you have forgotten to update the function names.

PR #1175, opened 2019-11-12
Added support for HDL Checker server to lsp-vhdl

chrbirks:
Second attempt of adding support for new server. Now without any `user-error`.

### DIFF
--- a/lsp-vhdl.el
+++ b/lsp-vhdl.el
@@ -70,9 +70,10 @@ HDL Checker: A wrapper for third party tools such as GHDL, ModelSim, Vivado Simu
   :risky t
   :type 'file)
 
+(defvar lsp-vhdl--params nil)
+
 (defun lsp-vhdl--create-connection ()
   "Returns lsp-stdio-connection based on the selected server"
-  (defvar lsp-vhdl--params nil)
   (lsp-vhdl--set-server-path)
   (lsp-vhdl--set-server-args)
   (plist-put
--- a/lsp-vhdl.el
+++ b/lsp-vhdl.el
@@ -21,44 +21,84 @@
 
 ;;; Commentary:
 
-;; LSP support for VHDL using the language server from http://www.vhdltool.com/download.
-;; Set the following symbol to specify the path of the language server:
+;; LSP support for VHDL using using an external language server. Currently
+;; the supported servers are:
 ;;
-;; (setq lsp-vhdl-server-path "/path/to/server_binary")
-;;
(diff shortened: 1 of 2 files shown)


lsp-vhdl.el
```diff
@@ -21,44 +21,84 @@
...
   :risky t
   :type 'file)
 
 (defun lsp-vhdl--create-connection ()
-  "Returns lsp-stdio-connection or an error if server not found"
-  (lsp-stdio-connection
-   (lambda ()
-     (list lsp-vhdl-server-path "lsp"))))
+  "Returns lsp-stdio-connection based on the selected server"
+  (setq params nil)
```
yyoncho: Having `params` defined globally is not ok for elisp(you may have collisions). You may:

1. Change [ara,s to local - use let and change it to hashtable. 
2. Create a defvar which is prefixed with `lsp-vhdl--` to avoid the collisions. 
3. Do not have the params and calculate the command on demand(e. g. return functions :test? and :connect which will calculate the command on-demand two times.

chrbirks: Right, I'll change it to a defvar with name `lsp-vhdl--params` instead - when I figure out how to update a PR with a new commit.

lsp-vhdl.el
```diff
@@ -72,26 +72,26 @@ HDL Checker: A wrapper for third party tools such as GHDL, ModelSim, Vivado Simu
 
 (defun lsp-vhdl--create-connection ()
   "Returns lsp-stdio-connection based on the selected server"
-  (setq params nil)
+  (defvar lsp-vhdl--params nil)
```
yyoncho: defvar should be on the top level and then you should use setq in the method.

chrbirks: What exactly is top level in Lisp? Is it the top of the file or right after `(lsp-register-client` in line 95 in this case? Or can you point me to another example where support for multiple servers have been implemented? I feel like I'm reimplementing the wheel where others might have done something similar.

yyoncho: > What exactly is top level in Lisp?

It is outside of the function, like `defcustom`.

> Or can you point me to another example where support for multiple servers have been implemented? 

We usually have lsp-register-client per server(check lsp-rust.el for example).

chrbirks: I have moved the defvar outside the function for now. I will look into refactoring the code at some point so it will be more similar to how other people have done it.

yyoncho:
Thank you for contributing to `lsp-mode` and for the patience!

PR #1180, opened 2019-11-13
lsp-organize-imports: suppress "no actions" error

### LINKED ISSUE #1177: lsp-organize-imports errors out if action not available
Servers such as gopls and java only offer the "organizeImports" action if there are changes to be made. This means `lsp-organize-imports` errors out when there is nothing to organize, which makes it not work very well as a before-save-hook.

I propose we change lsp-organize-imports to hide the "No actions to select from" error, or at least hide it if invoked non-interactively.

muirdm:
Change lsp-organize-imports to catch the "no code actions" error.
Servers only offer the organizeImports action when there are imports
to organize. If lsp-organize-imports raises an error in normal cases
then it is not usable as a before-save-hook.

When called interactively, lsp-organize-imports still produces an
informative message.

Fixes #1177.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -224,6 +224,8 @@ occasionally break as language servers are updated."
   "Unsupported file scheme" 'lsp-error)
 (define-error 'lsp-client-already-exists-error
   "A client with this server-id already exists" 'lsp-error)
+(define-error 'lsp-no-code-actions
+  "No code actions" 'lsp-error)
 
 (defcustom lsp-auto-guess-root nil
   "Automatically guess the project root using projectile/project.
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4056,9 +4058,13 @@ If ACTION is not set it will be selected from `lsp-code-actions'."
...
 (defun lsp-organize-imports ()
-  "Perform the source.organizeImports code action."
+  "Perform the source.organizeImports code action, if available."
   (interactive)
-  (lsp-execute-code-action-by-kind "source.organizeImports"))
+  (condition-case nil
+      (lsp-execute-code-action-by-kind "source.organizeImports")
+    (lsp-no-code-actions
+     (when (called-interactively-p 'any)
+       (message "source.organizeImports action not available")))))
```
yyoncho: Nit: can you change `message` to `lsp--info`?

muirdm: Done.

PR #1184, opened 2019-11-17
Add support for passing environment variables to language servers

danielmartin:
Some language servers support setting some environment variables in
their main process to offer additional functionality. For example, the
Bash language server supports a couple of environment variables to
enable using Explainshell to answer "onHover" queries, or to consider
parsing errors as problems.

* (lsp-clients.el): Add lsp-bash group for Bash-specific
customizations.
* (lsp-clients.el): Add lsp-bash-explainshell-endpoint and
lsp-bash-highlight-parsing-errors defcustoms.
* (lsp-mode.el): Add function environment-fn to lsp-client.
* (lsp-clients.el): Set environment-fn when registering the Bash
client.
* (lsp-mode.el): Use environment-fn when starting the different server
processes by appending the alist to a let-bound process-environment.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1074,14 +1074,6 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
   (server-id)
   ;; defines whether the client supports multi root workspaces.
   (multi-root)
-  ;; Function that returns an environment structure that will be used
-  ;; to set some environment variables when starting the language
-  ;; server process. These environment variables enable some
-  ;; additional features in the language server. The environment
-  ;; structure is an alist of the form (KEY . VALUE), where KEY is a
-  ;; string (regularly in all caps), and VALUE may be a string, a
-  ;; boolean, or a sequence of strings.
-  (environment-fn)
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -1074,6 +1074,14 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
...
   ;; defines whether the client supports multi root workspaces.
   (multi-root)
+  ;; Function that returns an environment structure that will be used
+  ;; to set some environment variables when starting the language
+  ;; server process. These environment variables enable some
+  ;; additional features in the language server. The environment
+  ;; structure is an alist of the form (KEY . VALUE), where KEY is a
+  ;; string (regularly in all caps), and VALUE may be a string, a
+  ;; boolean, or a sequence of strings.
+  (environment-fn)
```
yyoncho: can you put that in the end?

yyoncho: (If it is not in the end it will break the packages byte-compiled against the previous version.)

yyoncho:
Thank you!

PR #1185, opened 2019-11-18
[imp] default pylint to off

mpanarin:
https://github.com/palantir/python-language-server/pull/698

### DIFF
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -145,7 +145,7 @@ complexity."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.1"))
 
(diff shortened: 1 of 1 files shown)


seagle0128:
It does make sense and looks fine.

yyoncho:
Merging based on @seagle0128's comment.

AndreaOrru:
Even though this variable is set to nil, pylint sporadically runs. It's really annoying :(

AndreaOrru:
I found that Pylint may still be run by `flycheck` as the next checker in chain even when you use `flake8`.

I solved it by doing:
```
(add-to-list 'flycheck-disabled-checkers 'python-pylint)
```

PR #1186, opened 2019-11-18
add flake8 parameters to lsp-pyls

mpanarin:
pyls supports `flake8` checks.

### DIFF
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -290,6 +290,54 @@ at all."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-pyls-plugins-flake8-enabled nil
+  "Enable or disable the plugin."
+  :type 'boolean
+  :group 'lsp-pyls
+  :package-version '(lsp-mode . "6.2"))
+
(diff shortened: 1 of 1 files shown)


lsp-pyls.el
```diff
@@ -290,6 +290,54 @@ at all."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-pyls-plugins-flake8-enabled nil
+  "Enable or disable the plugin."
+  :type 'boolean
+  :group 'lsp-pyls
+  :package-version '(lsp-mode . "6.1"))
```
yyoncho: these should be 6.2

brotzeit: I think this should be 6.3. We bumped the version a few weeks ago.

yyoncho: My understanding is that we are 6.2 until we tag release - https://github.com/emacs-lsp/lsp-mode/releases . I am not sure when we should bump the package header version - when we start new release when we release it. I am fine with both approaches. But still, if we tag 6.2 we should update the changelog for the release :).

mpanarin: updated to 6.2 )

PR #1190, opened 2019-11-20
Moving lsp-powershell from kiennq/lsp-powershell

kiennq:
I'm the author of [lsp-powershell](https://github.com/kiennq/lsp-powershell) and would like to move it under the umbrella of lsp-mode.


lsp-pwsh.el
```diff
@@ -0,0 +1,307 @@
...
+  :priority 1
+  :initialization-options #'lsp-pwsh--extra-init-params
+  :notification-handlers (lsp-ht ("powerShell/executionStatusChanged" 'ignore))
+  :initialized-fn (lambda (w)
+                    (with-lsp-workspace w
+                      (lsp--set-configuration
+                       (lsp-configuration-section "powershell"))))
+  ))
+
+(add-to-list 'lsp-language-id-configuration '(powershell-mode . "powershell"))
```
yyoncho: can you move that in lsp-mode file?

lsp-pwsh.el
```diff
@@ -0,0 +1,307 @@
...
+  "The path to powershell vscode extension."
+  :group 'lsp-pwsh
+  :type 'string)
+
+(defcustom lsp-pwsh-exe (or (executable-find "pwsh") (executable-find "powershell"))
+  "PowerShell executable."
+  :type 'string
+  :group 'lsp-pwsh)
+
+(defcustom lsp-pwsh-dir (expand-file-name "PowerShellEditorServices" lsp-pwsh-ext-path)
```
yyoncho: can you add `:package-version '(lsp-mode . "6.2")` to all new defcustoms?

lsp-pwsh.el
```diff
@@ -0,0 +1,307 @@
...
+                  (not (memq major-mode lsp-pwsh--major-modes)))
+                '((name . --force-post-completion))))
+
+(defun lsp-pwsh--filter-cr (str)
+  "Filter CR entities from STR."
+  (when (and (eq system-type 'windows-nt) str)
+    (replace-regexp-in-string "\r" "" str)))
+
+(advice-add 'lsp-ui-doc--extract :filter-return #'lsp-pwsh--filter-cr)
+(advice-add 'lsp-ui-sideline--format-info :filter-return #'lsp-pwsh--filter-cr)
```
yyoncho: This should go in lsp-ui package.

lsp-pwsh.el
```diff
@@ -0,0 +1,307 @@
...
+FORCED if specified with prefix argument."
+  (interactive "P")
+  (let ((parent-dir (file-name-directory lsp-pwsh-dir)))
+    (unless (and (not forced) (file-exists-p parent-dir))
+      (lsp-pwsh--get-extension
+       (format lsp-pwsh-github-asset-url "PowerShell" "PowerShellEditorServices" "PowerShellEditorServices.zip")
+       parent-dir))))
+
+;; Download extension
+(lsp-pwsh-setup)
```
yyoncho: This should be moved from top-level to probably to `lsp-pwsh--command`. FYI At some point, we will have a field in the `lsp--client` for download commands and lsp-mode will automatically manage that process.

lsp-pwsh.el
```diff
@@ -0,0 +1,307 @@
...
+  :initialized-fn (lambda (w)
+                    (with-lsp-workspace w
+                      (lsp--set-configuration
+                       (lsp-configuration-section "powershell"))))
+  ))
+
+(add-to-list 'lsp-language-id-configuration '(powershell-mode . "powershell"))
+
+;; Compatibility
+(if (fboundp 'company-lsp)
```
yyoncho: Are you able to make lsp-mode work with company-tng using this piece of code? Still, we have to move it somewhere.

yyoncho: cc @dgutov - we were discussing the company-tng and this might be helpful.

kiennq: At this point I think this is kind of like language specific fix to get it works with other popular package. 
For language servers that suffer from `company-tng` is not expanding the post completion, I think PowerShellEditorService got hit the hardest. Since it actually erases part of text at beginning of completion candidate and need the post completion to get back.
I've already check if the advice is applied only for powershell though.
Is that good to still keep this here?

lsp-pwsh.el
```diff
@@ -0,0 +1,307 @@
+;;; lsp-pwsh.el --- lsp-mode client for PowerShellEditorServices  -*- lexical-binding: t; -*-
```
brotzeit: I think you can remove the lsp-mode here.

lsp-pwsh.el
yyoncho: > Is that good to still keep this here?

ATM company-tng is considered unsupported(or company-tng does not support company-lsp). IMO even if we submit it as it is we should work with upstream no fixing it.

In your experience, is this sufficient to make company-tng work with company-lsp? 

As a side note `(with-eval-after-load` might be a bit better than `fboundp`.

kiennq: > In your experience, is this sufficient to make company-tng work with company-lsp?

Yes, this's sufficient to make `company-tng` work with `lsp-pwsh`.

I really cannot say anything for other language servers, but in my experience it works acceptable.

yyoncho: Merging as it is but I am reserving the right to reverse it in case of a problem.

kiennq: Great, thank you.

dgutov: @yyoncho I see the hack, it's... interesting. But if you have a specific suggestion how company-tng can be improved, I'm all ears.

PR #1191, opened 2019-11-20
indent elisp

brotzeit:


yyoncho:
I was about to do the same. Have you indented lsp-mode.el as well? I think it also has misaligned stuff?

PR #1202, opened 2019-11-26
Better implementation of displaying textDocument/signatureHelp

yyoncho:
This implementation is more accurate and closer to the lsp protocol idea.

- the signatures could be browsed using `M-n`/`M-p` and `C-c C-k` to switch back
to hover info

- the signatureHelp is triggered when

- added two new variables: `lsp-signature-render-documentation` - to control
whether to include signature documentation and `lsp-signature-auto-activate`.
When `lsp-signature-auto-activate` is true the signatureHelp will be displayed
when you press the signature trigger char(e. g. `(` in java).

``` elisp
(add-hook 'company-completion-finished-hook (lambda (&rest _) (lsp-signature-activate)))
```

### DIFF
--- a/README.org
+++ b/README.org
@@ -202,6 +202,10 @@
    - ~lsp-lens-hide~ - Hide lenses in the current file
    - ~lsp-lens-mode~  - Turn on/off lenses in the current file.
    - ~lsp-avy-lens~ - Click lens using ~avy~
+   - ~lsp-signature-activate~ - activate signature help. Bound to ~C-M-SPC~ by default.
+   - ~lsp-signature-next~ - show next signature. Bound to ~M-n~ when signature help is activated.
+   - ~lsp-signature-previous~ - show previous signature. Bound to ~M-p~ when signature help is activated.
+   - ~lsp-signature-stop~ - cancel signature help. Bound to ~C-c C-k~ when signature help is activated.
 ** Settings
    - ~lsp-log-io~ - If non-nil, print all messages to and from the language server to ~*lsp-log*~.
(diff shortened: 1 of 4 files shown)


README.org
```diff
@@ -202,6 +202,10 @@
    - ~lsp-lens-hide~ - Hide lenses in the current file
    - ~lsp-lens-mode~  - Turn on/off lenses in the current file.
    - ~lsp-avy-lens~ - Click lens using ~avy~
+   - ~lsp-signature-activate~ - activate signature help. Bound to ~C-M-SPC~ by default.
```
seagle0128: The keybinding is `C-S-SPC` in `lsp-mode-map`.

lsp-mode.el
```diff
@@ -125,18 +125,6 @@
   :type 'boolean
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-use-native-json t
```
ocharles: This seems to be the PR that removes this option. Do those of us on Emacs HEAD need to do anything to get native JSON parsing?

yyoncho: No, it will be picked automatically. The flag was there temporary as a fallback option when we introduced native json support.

ocharles: Thank you for the very fast response! I look forward to trying `lsp-mode` in Emacs HEAD then! :smile:

yyoncho: Make sure you remove the `elc` generated from the old emacs - we do some evaluations compile time depending on emacs version to make the runtime faster. (The easiest option is to delete the package and install it again).

honnix: I wish I could see this comment earlier. Thanks!

PR #1203, opened 2019-11-26
Make sure experimental features are passed to gopls as boolean

lucianjon:
Make sure when the new boolean experimental gopls features are `nil` they are passed to gopls as `false`

Currently there is an error on initialisation if they are nil


yyoncho:
Thank you but I reverted the commit introducing the experimental features based on https://github.com/emacs-lsp/lsp-mode/pull/1121 . @vibhavp the go language maintainers(@muirdm  and @seagle0128 ) agreed that the experimental features won't be part the default configuration.

PR #1206, opened 2019-12-01
Quote the powershell path

r-darwish:
Failing to do so resulted the following error in Windows:

```
'c:/Program' is not recognized as an internal or external command,
operable program or batch file.
```


yyoncho:
Looks good! I will wait for @kiennq as the author of the code to approve it too.

kiennq:
Approve. Thanks for fixing the path

PR #1210, opened 2019-12-01
Prefer previous workspace when jumping to library

### LINKED ISSUE #1178: Prefer current workspace when jumping to library code
Currently when jumping into library code not under an active workspace we seem to randomly choose a workspace to open the file in. This causes things to break for gopls since gopls will not work properly for the library package unless it is a dependency of the current workspace.

I propose we change the library workspace selection to prefer the current buffer's workspace if it can't otherwise find a suitable workspace.

psanford:
Remember what the last active workspace(s) was when opening a file in
a library directory. When jumping to a library definition you most
likely want to open that library from the same workspace as your
current buffer.

Fixes #1178

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -618,6 +618,11 @@ Changes take effect only when a new session is started."
                                         (crystal-mode . "crystal"))
   "Language id configuration.")
 
+(defvar lsp--last-active-workspaces nil
+  "Keep track of last active workspace.
+We want to try the last workspace first when jumping into a library
+directory")
+
 (defvar lsp-method-requirements
   '(("textDocument/onTypeFormatting" :capability "documentOnTypeFormattingProvider")
     ("workspace/executeCommand"
@@ -2490,6 +2495,7 @@ If NO-MERGE is non-nil, don't merge the results but return alist workspace->resu
                                     cancel-token))
              (id (cl-incf lsp-last-id))
              (body (plist-put body :id id)))
+        (setq lsp--last-active-workspaces target-workspaces)
         (when cancel-token
           (puthash cancel-token (cons id target-workspaces) lsp--cancelable-requests))
         (seq-doseq (workspace target-workspaces)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1105,6 +1110,15 @@ to the beginning and ending points in the range correspondingly."
       (cons start end))
      (parent (lsp--find-wrapping-range parent)))))
 
+(defun lsp--focus-buffer (&rest args)
+  "Focus buffer handler.
+
+Remember the last active workspace set so that we can try that first
```
yyoncho: typo: workspace -> workspaces.

psanford:
> I wonder what would lead to better behaviour using `lsp--send-request-async` to set `lsp--last-active-workspaces`(using `target-workspaces`) or current solution. WDYT?

Interesting. You are suggesting we keep track based on the last lsp request we issued instead of the last active buffer. That seems like a better solution than advising 'select-window like the current diff does. 

I'll give that a try.

lsp-mode.el
psanford: Ha. When I used the word "set" I meant in the discrete math sense of the word. I can see how this comment is confusing and will update it.

yyoncho: Ah, it was actually me misreading the comment. It is fine to stay as it is.

lsp-mode.el
```diff
@@ -612,6 +612,11 @@ Changes take effect only when a new session is started."
                                         (ess-r-mode . "r"))
   "Language id configuration.")
 
+(defvar lsp--last-active-workspaces nil
+  "Keep track of last active workspace.
+We want to try the last workspace first when jumping into library
+a library directory")
```
muirdm: > jumping into library a library directory

```
  paris
  in the
the springtime
```

psanford: Fixed.

lsp-mode.el
psanford: Fixed.

psanford:
I've updated the pr to set the last active workspaces from (lsp--send-request-async).

yyoncho:
Thank you!

PR #1213, opened 2019-12-02
Modify declare-function invocations for external plugins.

leungbk:
For functions from external packages, we must specify "ext:".


yyoncho:

PR #1215, opened 2019-12-02
lsp-pwsh: support code formatting

### LINKED ISSUE #1212: Powershell formatting is not supported
**Describe the bug**
`lsp-format-buffer` does not work in Powershell even though the server supports formatting

**To Reproduce**
Run `lsp-format-buffer` in a Powershell buffer

**Expected behavior**
Formatting should work like in Visual Studio Code

Powershell

**OS**
macOS

**Error callstack**
```
Debugger entered--Lisp error: (lsp-capability-not-supported "documentFormattingProvider")
  signal(lsp-capability-not-supported ("documentFormattingProvider"))
  (cond ((or (lsp--capability "documentFormattingProvider") (lsp--registered-capability "textDocument/formatting")) (let ((edits (lsp-request "textDocument/formatting" (lsp--make-document-formatting-params)))) (lsp--apply-formatting edits))) ((or (lsp--capability "documentRangeFormattingProvider") (lsp--registered-capability "textDocument/rangeFormatting")) (save-restriction (widen) (lsp-format-region (point-min) (point-max)))) (t (signal (quote lsp-capability-not-supported) (list "documentFormattingProvider"))))
  lsp-format-buffer()
  funcall-interactively(lsp-format-buffer)
  call-interactively(lsp-format-buffer record nil)
  command-execute(lsp-format-buffer record)
  execute-extended-command(nil "lsp-format-buffer" "lsp-format-bu")
  funcall-interactively(execute-extended-command nil "lsp-format-buffer" "lsp-format-bu")
```

kiennq:
Fix #1212:
- force enable `formattingProvider` capability of PSES since it already supports that.
- filter `\r` out of edit text since they're annoying and logically should never used as replaced text.

I'm not really happy with the hack of removing `\r`, @yyoncho, do you have any suggestion?

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3129,6 +3129,8 @@ The method uses `replace-buffer-contents'."
                                     beg (+ beg (length newText))
                                     length)))))))))
 
+(defvar-local lsp--filter-cr? t)
+
 (defun lsp--apply-text-edits (edits)
   "Apply the edits described in the TextEdit[] object."
   (unless (seq-empty-p edits)
@@ -3146,6 +3148,11 @@ The method uses `replace-buffer-contents'."
                            'lsp--apply-text-edit)))
         (unwind-protect
             (->> edits
+                 (mapc (lambda (edit)
+                         (when lsp--filter-cr?
+                           ;; filter \r out of text since it's mostly useless
+                           (ht-set edit "newText"
+                                   (s-replace "\r" "" (ht-get edit "newText" ""))))))
                  (nreverse)
                  (seq-sort #'lsp--text-edit-sort-predicate)
                  (mapc (lambda (edit)
--- a/lsp-pwsh.el
+++ b/lsp-pwsh.el
@@ -307,7 +307,10 @@ Must not nil.")
   :initialized-fn (lambda (w)
                     (with-lsp-workspace w
                       (lsp--set-configuration
-                       (lsp-configuration-section "powershell"))))
+                       (lsp-configuration-section "powershell")))
+                    (let ((caps (lsp--workspace-server-capabilities w)))
+                      (ht-set caps "documentRangeFormattingProvider" t)
+                      (ht-set caps "documentFormattingProvider" t)))
   ))
 
 ;; Compatibility


lsp-mode.el
```diff
@@ -3146,6 +3146,10 @@ The method uses `replace-buffer-contents'."
                            'lsp--apply-text-edit)))
         (unwind-protect
             (->> edits
+                 (mapc (lambda (edit)
```
yyoncho: having this unconditionally seems to be opinionated and this might be a problem in mixed Windows/Linux teams with coding conventions. Are there functional impact when the ^Ms are present?  As a side note it is possible to hide ^M https://stackoverflow.com/questions/730751/hiding-m-in-emacs .

kiennq: Hiding `\r` in Emacs doesn't seem good to me. If we should use that code, it should be in user config file.

With powershell language server, whenever user send `\n` and ask for formatting, the server will return `\r\n` as it's the standard EOL in Windows.
Now in Emacs, when the file coding is set to `-dos`, the `\r\n` will automatically converted to `\n` (at least that happens when you call `(buffer-string)`). 

So we have an Emacs specific issue here:
- Whenever we request formatting, powershell server will answer with `\r\n`.
- Now go to Emacs, even if the file encoding is set to `-dos`, if we're replacing text EOL with `\r\n`, the `\r` will be treated as extra character, and `^M` will be shown.
- Worse, next time we request formatting with `\r\n`, the server will return `\r\r\n` and inflate the buffer indefinitely.

Once the buffer is in its correct encoding (`-dos`, `-linux` and `-mac`), the newline will always be represented as `\n` instead of `\r\n`, so removing `\r` would be fine IMO.
However, if this change is risky, can you suggest a better scoped change for this?

yyoncho: > Hiding \r in Emacs doesn't seem good to me.

Yes, I am not suggesting it. Just mentioning that if there is no functional impact but only visual there is an option to hide it. 

> if we're replacing text EOL with \r\n, the \r will be treated as extra character, and ^M will be shown.

Why do we need to do that?

> Worse, next time we request formatting with \r\n, the server will return \r\r\n and inflate the buffer indefinitely.

Can we avoid that? This sounds like an issue in lsp-mode.

> Once the buffer is in its correct encoding (-dos, -linux and -mac), the newline will always be represented as \n instead of \r\n, so removing \r would be fine IMO.

Can you elaborate on that?

> However, if this change is risky, can you suggest a better scoped change for this?

We may have a flag to enable it and enable it for powershell buffers via buffer local.

kiennq: > Why do we need to do that?

That's what powershell language server does. `lsp-mode` sends EOL as `\n` and the language server will answer that with `\r\n` as it's standard EOL in Windows. Now when we try to replace the current buffer with what language server sent, `\r` will be left there alone, visible.

> Can we avoid that? This sounds like an issue in lsp-mode.

That why this change try to remove `\r` from text returned from language server, to match with what Emacs expects (I guess).

>Can you elaborate on that?

For example, a buffer contains "a\r\n". If we encode it using `utf-8-dos`, we receive "a\n" with `buffer-string`. If we encode it using `utf-8-unix`, we get "a\r\n".
Both cases, the EOL is just "\n". After we fix the buffer encoding and do text replacement, any extra "\r" will be visible and should be safely removed.

yyoncho: So, what you are saying is that removing \r is practically noop since it will be either removed by emacs if it is not needed or automatically added if needed?

kiennq: > So, what you are saying is that removing \r is practically noop since it will be either removed by emacs if it is not needed or automatically added if needed?

That's not I meant. Sorry.
What I mean is that in Emacs, once we encoding the buffer correctly, the EOL will always be `\n`, having extra `\r`s around are mostly redundant and distracting.

I'm curious, did we convert `\n` to `\r\n` when sync document with server in Windows?

yyoncho: > That's not I meant. Sorry.
What I mean is that in Emacs, once we encoding the buffer correctly, the EOL will always be \n, having extra \rs around are mostly redundant and distracting.

What will happen if the user does want dos encoding? Will he get \r\n?

> I'm curious, did we convert \n to \r\n when sync document with server in Windows?

AFAIK no. 

@emacs-lsp/lsp-mode does anybody have more knowledge regarding the matter?

kiennq: > What will happen if the user does want dos encoding? Will he get \r\n?

The replaced text and file encoding is separated matter. 
For example, if the file encoding is `-dos`, and user save string "a\n" into file. It will be saved as "a\r\n" and user will only see "a\n" with `buffer-string`.
Similarly, if the file encoding is `-unix`, and user save string "a\n" into file, It will be saved as "a\n" and user will only see "a\n" as `buffer-string`

Now since language server will try to replace "\n" with "\r\n", the `buffer-string` will become something like "a\r\n" in both `-dos` and `-unix` encoding. And respectively, "a\r\r\n" and "a\r\n" in file content.
You can see that the "\r" is redundant as it interfere with the current file encoding

yyoncho: I tested mssql server with that change and it solved some issues. I will wait for some time before merging it so others can share their opinion. Meanwhile, can you add a variable to disable this behaviour (don't use defcustom but defvar).

kiennq: Added

PR #1238, opened 2019-12-10
Do not enable lsp-ui by default

yyoncho:
_Note_: this PR will be merged based on the up-votes and down-votes.

### DIFF
--- a/README.org
+++ b/README.org
@@ -50,6 +50,7 @@
    - 🚀 *Easy to configure* - works out of the box and automatically upgrades if additional packages are present
    - 🌟 *Flexible* - could be configured as full-blown IDE with flashy UI or minimal distraction free
 ** Overview
+   _Configuration change: starting from v6.2.1 [[https://github.com/emacs-lsp/lsp-ui][lsp-ui]] won't be enabled by default. Refer to configuration section on how to enable it._
    Client for [[https://github.com/Microsoft/language-server-protocol/][Language Server Protocol]] (v3.14). [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] aims to provide IDE-like experience by providing optional integration with the most popular Emacs packages like ~company~, ~flycheck~ and ~projectile~.
 
    - Non-blocking asynchronous calls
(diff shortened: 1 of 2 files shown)


seagle0128:
IMO, maybe lsp-ui can be enabled by default, but not all features should be enabled by default, e.g. `sideline`. I know many people don't like hover and diagnose sideline.

kiennq:
I would second the option of @seagle0128, `lsp-ui` can be enabled by default, but not all of it features.
Personally, I only enable `lsp-ui-doc` and wire to display hover only when pressing `K` (`evil-lookup-func`).

Here is what I configured
``` elisp
(use-package lsp-ui
  :ensure t
  :diminish
  :hook (lsp-mode . lsp-ui-mode)
  :init
  (setq lsp-ui-peek-enable nil
        lsp-ui-sideline-enable nil
        lsp-ui-imenu-enable nil
        lsp-ui-doc-enable t
        lsp-ui-doc-position 'at-point
        lsp-ui-doc-include-signature t)
  (setq evil-lookup-func
        (lambda () (cond
                    ((and (not (display-graphic-p)) (boundp 'lsp-mode) lsp-mode)
                     (lsp-hover))
                    ((and (boundp 'lsp-ui-doc-mode) lsp-ui-doc-mode)
                     (lsp-ui-doc-show))
                    (t (woman)))))
  :config
  (lsp-ui-mode t)
  (advice-add #'lsp-ui-doc-hide
              :before-while
              (lambda (&rest _)
                  (or (overlayp lsp-ui-doc--inline-ov)
                      (let ((frame (lsp-ui-doc--get-frame)))
                        (and frame (frame-visible-p frame)))))
              '((name . --p)))
  (add-hook 'lsp-ui-doc-mode-hook
            (lambda ()
              (if lsp-ui-doc-mode
                  (progn
                    (add-hook 'post-command-hook #'lsp-ui-doc-hide nil t)
                    (remove-hook 'post-command-hook #'lsp-ui-doc--make-request t))
                (remove-hook 'post-command-hook 'lsp-ui-doc-hide t))))
  )
```

dennismayr:
(I like `sideline` enabled :confused: )

yyoncho:
> _Note_: this PR will be merged based on the up-votes and down-votes.

Issue #1073, opened 2019-10-06
Support for JSON Language Server


iquiw:
I'm [working on this](https://github.com/iquiw/lsp-mode/commits/lsp-json).

Language server: https://github.com/vscode-langservers/vscode-json-languageserver

There are 2 issues I am facing.

1. Completion does not work while server responses `textDocument/completion`. I did not check the detail yet and it could be taken up separately.
2. With default setting (`lsp-clients-json-schemas` is `nil`), the language server does not detect any schema.
  In VSCode, its client also does not send schemas in default, but its server knows some well-known files, such as `package.json`, `tsconfig.json`, etc.
  Should we provide some schemas as default setting?

yyoncho:
IMO we should have as much stuff working as possible by default and usually what vscode does by default can be used as a reference.

AFAICS vscode sends the following notification which is not present in your code:

```
[Trace - 7:11:12 PM] Sending notification 'json/schemaAssociations'.
Params: {
    "vscode://defaultsettings/keybindings.json": [
        "vscode://schemas/keybindings"
    ],
    "/User/keybindings.json": [
        "vscode://schemas/keybindings"
    ],
    "vscode://defaultsettings/settings.json": [
        "vscode://schemas/settings/default"
    ],
    "/User/settings.json": [
        "vscode://schemas/settings/user"
    ],
    "/Machine/settings.json": [
        "vscode://schemas/settings/machine"
    ],
    "/Workspaces/*/workspace.json": [
        "vscode://schemas/workspaceConfig"
    ],
    "/**/*.code-workspace": [
        "vscode://schemas/workspaceConfig"
    ],
    "/User/locale.json": [
        "vscode://schemas/locale"
    ],
    "/.vscode/settings.json": [
        "vscode://schemas/settings/folder"
    ],
    "/.vscode/launch.json": [
        "vscode://schemas/launch"
    ],
    "/.vscode/tasks.json": [
        "vscode://schemas/tasks"
    ],
    "/User/snippets/*.json": [
        "vscode://schemas/snippets"
    ],
    "/**/*.code-snippets": [
        "vscode://schemas/global-snippets"
    ],
    "/.vscode/extensions.json": [
        "vscode://schemas/extensions"
    ],
    "/.devcontainer/devcontainer.json": [
        "file:///usr/share/code/resources/app/extensions/configuration-editing/schemas/devContainer.schema.json"
    ],
    "/.devcontainer.json": [
[...]

PR #1243, opened 2019-12-11
Support for JSON Language Server 

kiennq:
Per #1073, adding `lsp-json` with server automatically installation capability.


lsp-json.el
```diff
@@ -0,0 +1,136 @@
...
+
+(require 'lsp-mode)
+(require 'ht)
+(require 'url-util)
+
+(defgroup lsp-json nil
+  "LSP support for JSON, using vscode-json-languageserver."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/vscode-langservers/vscode-json-languageserver")
+  :package-version '(lsp-mode . "6.2"))
```
yyoncho: 6.2 was just released. We are now 6.3

kiennq:
> 1. Is it possible to associate the current buffer with a specific schema?

Yes, either via `"$schema": "..."` in json file or special mapping with `lsp-json-schemas`.
Please refer [here](https://github.com/vscode-langservers/vscode-json-languageserver#settings) for latter settings.

> 2. Do we need a method like lsp-json-update-schemas which should be called when schema mappings are changed?

Not really, if we update schema then just run `lsp-restart-workspace` would be easier :).
I don't want to provide additional method (we can try to do it through `:set` for `defcustom`) unless it's necessary.

lsp-json.el
kiennq: Updated!

yyoncho:
Ok, if `lsp-json--schema-associations` is intended to be used by the users it should be renamed to `lsp-json-schema-associations`.

kiennq:
> Ok, if `lsp-json--schema-associations` is intended to be used by the users it should be renamed to `lsp-json-schema-associations`.

No, user should use `lsp-json-schemas`, the `lsp-json--schema-associations` is internal.

yyoncho:
Thank you!

yyoncho:
> No, user should use `lsp-json-schemas`, the `lsp-json--schema-associations` is internal.

Yes, you are right.

yyoncho:
Do you have twitter? If yes, can you twit about it with a gif? IMO it will be very useful language server for a lot of people.

yyoncho:
@kiennq one more thing - if you have twitter account it will be great to tweet about the new functionality with a gif. IMO this server will be helpful for a lot of people.

kiennq:
Thanks, here is my twit https://twitter.com/neikmaster/status/1204859535610957824

lsp-json.el
```diff
@@ -0,0 +1,136 @@
+;;; lsp-pwsh.el ---  vscode-json-languageserver integration -*- lexical-binding: t; -*-
```
thisch: this filename in this comment is wrong

kiennq: Oh, the downfall of copy/paste :(. Will update it next time I touch this file again.

thisch: No worries.

PR #1245, opened 2019-12-12
Adding async-request-handlers for `lsp-json`

kiennq:
Since json language server requests emacs to provide file/remote resource content that may take time to retrieve, we need to make the request response being able to not block typing via async callback,

This is per discussion in #1073

### DIFF
--- a/README.org
+++ b/README.org
@@ -376,6 +376,8 @@ If the language server supports environment variables to control additional beha
    - I am getting "Package ‘spinner-1.7.3’ is unavailable" when trying to install =lsp-mode=.
      - This is caused by GPG keys used by the ELPA package manager not being up
       to date. You may fix by installing: [[https://elpa.gnu.org/packages/gnu-elpa-keyring-update.html][gnu-elpa-keyring-update]]
+   - Json completion doesn't seem working?
+     - The latest [[https://www.npmjs.com/package/vscode-json-languageserver/v/1.2.2][vscode-json-languageserver]] is broken. You will need to install the earlier version of it ~npm i vscode-json-languageserver@1.2.1~
 ** See also
    - [[https://github.com/emacs-lsp/lsp-docker/][lsp-docker]] - provide docker image with preconfigured language servers with corresponding emacs configuration.
    - [[https://github.com/sebastiencs/company-box/][company-box]] - =company= frontend with icons.
--- a/lsp-json.el
+++ b/lsp-json.el
@@ -1,4 +1,4 @@
-;;; lsp-pwsh.el ---  vscode-json-languageserver integration -*- lexical-binding: t; -*-
+;;; lsp-json.el ---  vscode-json-languageserver integration -*- lexical-binding: t; -*-
 
 ;; Copyright (C) 2019  Kien Nguyen
 
@@ -26,6 +26,7 @@
 
 (require 'lsp-mode)
 (require 'ht)
+(require 'url)
 (require 'url-util)
 
 (defgroup lsp-json nil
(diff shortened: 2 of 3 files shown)


yyoncho:
I had some troubles when testing on clean machine with emacs 27.

1. The automatic install does not work on linux systems since it requires super user privileges. I suggest removing automatic installation until we make it work for all systems.
2. The lists must be vectors because emacs 27 json serialization supports only vectors.

kiennq:
> 2\. The lists must be vectors because emacs 27 json serialization supports only vectors.

Are you talking about `lsp-json--schema-associations`? Also I wonder where can you get the emacs-snapshot version that support json serialization?
I'm on Ubuntu (via WSL) and install emacs with this PPA https://launchpad.net/~ubuntu-elisp/+archive/ubuntu/ppa


> 1. The automatic install does not work on linux systems since it requires super user privileges. I suggest removing automatic installation until we make it work for all systems.

Oh, it works great on Windows though.
In my Ubuntu, once I set `npm config set prefix '~/.npm-global'`, then install `npm i -g` **will** not require `sudo` anymore.
I'm not sure if there's anyone really install some packages for everyone on their machines.
For now I will scoop down the automatic installation to Windows, is that fine?

yyoncho:
> > 2. The lists must be vectors because emacs 27 json serialization supports only vectors.
> 
> Are you talking about `lsp-json--schema-associations`? Also I wonder where can you get the emacs-snapshot version that support json serialization?
> I'm on Ubuntu (via WSL) and install emacs with this PPA https://launchpad.net/~ubuntu-elisp/+archive/ubuntu/ppa

I don't know - I am compiling it from source - Another option is to use lsp-docker, which has emacs 27.

> lsp-json--schema-associations
>  lsp-json--extra-init-params
 
I fixed these locally but I am unable to get completion working. 

> In my Ubuntu, once I set `npm config set prefix '~/.npm-global'`, then install `npm i -g` **will** not require `sudo` anymore.
> I'm not sure if there's anyone really install some packages for everyone on their machines.
> For now I will scoop down the automatic installation to Windows, is that fine?

In general, I prefer having https://github.com/emacs-lsp/lsp-mode/issues/506 in before working on any automatic installations because it makes the task harder(e. g. whoever starts working on it will have to fix all automatic installations that are implemented). If you insist keep it for windows.

kiennq:
> I fixed these locally but I am unable to get completion working.

Can you post the locally fix here? I don't have emacs27 with native json now, so it would be better to reuse your fix.


> I fixed these locally but I am unable to get completion working.

The latest `vscode-json-languageserver` has problem, you may need to install the version before it.
`npm i vscode-json-languageserver@1.2.1`

yyoncho:
> > I fixed these locally but I am unable to get completion working.
> 
> Can you post the locally fix here? I don't have emacs27 with native json now, so it would be better to reuse your fix.

https://github.com/yyoncho/lsp-mode/commit/162c1d3b7f6276b71aa9f4e97b4e24ccbba88335
It works fine after this commit.

> 
> > I fixed these locally but I am unable to get completion working.
> 
> The latest `vscode-json-languageserver` has problem, you may need to install the version before it.
> `npm i -g vscode-json-languageserver@1.2.1`

Can you document that?

kiennq:
Fixed!!!

yyoncho:
Thank you!

As a side note, when you mark a server with `:multi-root t` lsp-mode will run only one instance of that server but I am not sure whether it could cause issues if the server does not support workspace folders.

kiennq:
> Thank you!
> 
> As a side note, when you mark a server with `:multi-root t` lsp-mode will run only one instance of that server but I am not sure whether it could cause issues if the server does not support workspace folders.

Oh, I didn't know about this. Certainly `lsp-json` could be benefit from this

Issue #1236, opened 2019-12-10
Handling backslash escape sequences in markdown.


machekku:
Hello,
I believe this is a followup to #1154.

I'm using PHP with intelephense.

PHP uses backslash character as namespace separator. Backslash is an escape character in markdown, so for example, hover docs for `Psr\Log\LoggerInterface::error` arrive like this:

[Trace - 02:54:41 PM] Received response 'textDocument/hover - (1682)' in 801ms.
```json
Result: {
  "range": {
    "end": {
      "character": 32,
      "line": 468
    },
    "start": {
      "character": 27,
      "line": 468
    }
  },
  "contents": {
    "value": "__Psr\\\\Log\\\\LoggerInterface::error__\n\nRuntime errors that do not require immediate action but should typically\nbe logged and monitored.\n\n```php\n<?php\npublic function error(\n    $message,\n    array $context = array()\n) { }\n```\n\n_@param_ `string $message`\n\n_@param_ `array $context`\n\n_@return_ `void`",
    "kind": "markdown"
  }
}
```

JSON decoder turns `Psr\\\\Log\\\\LoggerInterface` into `Psr\\Log\\LoggerInterface`, and then markdown parser should turn this into `Psr\Log\LoggerInterface`, but unfortunately it appears as `Psr\\Log\\LoggerInterface` in hover docs:
![image](https://user-images.githubusercontent.com/16875/70535944-33505580-1b5e-11ea-8d6d-cf4b51b45147.png)

A similar issue with escaped underscore character was fixed in #1154, however that fix was applied only for the underscore.

Original markdown spec (https://daringfireball.net/projects/markdown/syntax#backslash) gives this list of possible backslash escapes:

\\   backslash
\`   backtick
\*   asterisk
\_   underscore
\{}  curly braces
\[]  square brackets
\()  parentheses
\#   hash mark
\+   plus sign
\-   minus sign (hyphen)
\.   dot
\!   exclamation mark

yyoncho:
Willing to provide a PR? Unfortunately, we are forced to do these workarounds...

yyoncho:
Fixed by https://github.com/emacs-lsp/lsp-mode/pull/1247

PR #1247, opened 2019-12-13
fix markdown escape characters and add multi-root to lsp-json

kiennq:
Per discussion in #1236.
We should properly handling escaped characters in markdown.

Also add `:multiroot t` for `lsp-json` and `lsp-pwsh` since both of their server can work with multiple work spaces.

### DIFF
--- a/lsp-json.el
+++ b/lsp-json.el
@@ -125,6 +125,7 @@
   :major-modes lsp-json--major-modes
   :server-id 'json-ls
   :priority -1
+  :multi-root t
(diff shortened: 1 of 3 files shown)


seagle0128:
The regexp looks good to me. I am not sure if `:multi-root` is necessary. @yyoncho please confirm that.

yyoncho:
Thank you!

Merging as it is but IMO we should work with the upstream to fix it or look for alternative solution for rendering markdown.

PR #1249, opened 2019-12-15
Fix rust-analyzer inlay hints flickering

flodiebold:
Apparently the `'tick` mode got lost when copying the code for the rust-analyzer inlay hints, which causes them to flicker while editing.


brotzeit:
Thanks!

PR #1255, opened 2019-12-19
Use character U+25A0 instead of U+2B1B on document color overlay

### LINKED ISSUE #1254: Use character U+25A0 instead of U+2B1B on document color overlay
Looks like Emacs does not support the character between U+2B00 - U+2BFF on Windows properly. So I suggest use U+25A0 instead.

This is what U+2B1B showing in my overlay:
image

Emacs Version:
GNU Emacs 27.0.50 (build 1, x86_64-w64-mingw32)
 of 2019-11-12

My OS:
Windows 10

kevindragon:
fix #1254

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4205,6 +4205,11 @@ RENDER-ALL - nil if only the signature should be rendered."
                      :cancel-token :signature))
 
 
+(defcustom lsp-overlay-document-color-char "■"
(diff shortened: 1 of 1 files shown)


yyoncho:
I think it all depends on the font you are using and even U+25A0 might not be present. Can you make it configurable?

yyoncho:

PR #1258, opened 2019-12-20
Fix lsp-rust-server defcustom type

t-mw:
Trying to customize `lsp-rust-server` currently throws an error.


yyoncho:

PR #1260, opened 2019-12-21
ocmalmerlin-lsp moved to ocaml-lsp-server

rgrinberg:
The lsp server provided by the dev tool merlin, now moved to its own
repository.


yyoncho:
Thank you, some nitpicking:

1. Can you declare `define-obsolete-variable-alias` for each renamed variable?
2. Can you update the README.org as well in the Supported languages section?

rgrinberg:
Done. Thanks for the pointers

yyoncho:

PR #1261, opened 2019-12-21
Move `lsp-rust-server` to `lsp-rust` group so it's more discoverable

kanru:
Currently it's under `lsp-mode` directly. Naturally when one wants to customize `lsp-rust` they use `M-x customize-grup RET lsp-rust RET`.


yyoncho:
Thank you for contributing to `lsp-mode`, I cherry-picked your change into master branch.

PR #1262, opened 2019-12-21
lsp-csharp.el: ability to upgrade server via lsp-csharp-update-server

razzmatazz:
This adds ability to update roslyn server binary from github by checking for the latest version and initiating the download/extract procedure.

Server version to use is now determined enumerating and sorting by `version<` subdirectories on:
 - `~/.emacs.d/cache/omnisharp-roslyn`

..instead of referencing a (customizable) variable as we had before.

I think this makes the behaviour more flexible, with possibility to use an old version of the server (downgrade) by just deleting a new version from the cache dir. Also there is now no need to customize variables to update to newer version – just `M-x lsp-csharp-update-server`.

FYI @josteink

### DIFF
--- a/lsp-csharp.el
+++ b/lsp-csharp.el
@@ -34,14 +34,6 @@ Version 1.34.3 minimum is required."
   :group 'lsp-mode
   :link '(url-link "https://github.com/OmniSharp/omnisharp-roslyn"))
 
-(defcustom lsp-csharp-server-version
-  "v1.34.8"
-  "OmniSharp Roslyn server version to use
-Will use this value to download server binaries from the
-https://github.com/OmniSharp/omnisharp-roslyn/releases page."
-  :group 'lsp-csharp
-  :type '(string :tag "Version string starting with 'v', e.g. 'v1.34.8'"))
-
 (defcustom lsp-csharp-server-install-dir
   (locate-user-emacs-file ".cache/omnisharp-roslyn/")
   "Installation directory for OmniSharp Roslyn server."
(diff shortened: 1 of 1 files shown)


yyoncho:
@razzmatazz - not really related to this PR but can you provide a sample C# hello world application which I and anyone who wants to evaluate the server can use for testing the language server capabilities? I want to add it to lsp-docker demo projects.

lxbarbosa:
LGTM. 

installed latest version of the server correctly!

yyoncho:
@lxbarbosa thank you!

@razzmatazz you may consider also asking the user which version he/she wants to install and also a lsp-mode side validation whether the version is at least version X. This will be helpful when we have code that is depending on server fix.

Other than that I will wait for @josteink to validate the PR.

razzmatazz:
> @razzmatazz - not really related to this PR but can you provide a sample C# hello world application which I and anyone who wants to evaluate the server can use for testing the language server capabilities? I want to add it to lsp-docker demo projects.

Do you need a .zip file or instructions how to create a new C# projects with .net core SDK? Not sure if you can:
 - reference the `mcr.microsoft.com/dotnet/core/sdk:3.1` docker image from you demo project somehow..
 - or is it ok that the user needs to have https://dotnet.microsoft.com/download installed on his/her machine?

razzmatazz:
> @lxbarbosa thank you!
> 
> @razzmatazz you may consider also asking the user which version he/she wants to install and also a lsp-mode side validation whether the version is at least version X. This will be helpful when we have code that is depending on server fix.
> 
> Other than that I will wait for @josteink to validate the PR before merging it.

Mhm.. I wanted to reduce the amount of knobs (customizables) to a minimum, but it might be a good idea, to add `lsp-csharp-min-required-version` defvar later, once we get to a precedent where this is needed.

josteink:
@yyoncho I think 99.99% of the users will want latest version and we should optimize for that. 

Those who want a specific version, can still get through manually invoking the update-function with a specific version-argument. 

@razzmatazz looks good to me!

josteink:
Also if we want to respect the coverage-bit of the failing CI, we should quite easily be able to add some tests for lsp-csharp--version-list-latest, lsp-csharp--server-dir and lsp-csharp--server-bin.

yyoncho:
Thank you @josteink .

> @yyoncho I think 99.99% of the users will want latest version and we should optimize for that.
> 
> Those who want a specific version, can still get that through manually invoking the update-function with a specific version-argument.

Agreed - I was proposing the interactive dialogue to improve the user visibility over what is happening. 

> Also if we want to respect the coverage-bit of the failing CI, we should quite easily be able to add some tests for lsp-csharp--version-list-latest, lsp-csharp--server-dir and lsp-csharp--server-bin.

We should first fix the issues in lsp-mode.el which breaks the coverage generation tool - now we have coverage only for the lsp-mode.el extensions.

yyoncho:
@razzmatazz thank you for contributing to `lsp-mode`!

sonnguyen9800:
This should be put in the official doc, as this method is much more better and bug-free than manually install and build the server from source code.

ericdallo:
Agree @sonnguyen9800, we need a section on docs about auto install/download servers, and maybe one specific for lsp-csharp about updating the server.

josteink:
Added a PR to try to improve the documentation. Feel free to provide feedback.

PR #1270, opened 2019-12-25
[csharp] Use lsp-seq-first instead of seq-first

VPeruS:
Prevent errors on emacs 26


yyoncho:
Thank you. 

Apparently our byte-compilation tests are not picking all of the files I will fix them.

PR #1272, opened 2019-12-26
Implement automatic downloading facilities

### LINKED ISSUE #506: Feature: automatic server installation

yyoncho:
- Fixes #506

- Implemented base facilities for downloading and installing language servers
- added new field :download-server-fn in lsp--client structure which will be
called with (client callback update?).
- Implemented automatic installation via npm for jsts-ls and ts-ls.
- All servers(when feasible) should be installed under `lsp-server-install-dir`.

@akirak - check the cl-defgenerics in lsp-mode - let me know if they are
sufficient to implement the nix variands.

@razzmatazz - I have ported CSharp implementation but I havent converted it into
async.

@seagle0128 - lsp-python-ms is not ported. In general, it will work as it is but
it will ask to install the server even if you are about to start language server not being in python file. 

@TOTBWF - F# same as C#.

@kiennq powershell installation is converted to async one.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -153,18 +153,16 @@ This directory shoud contain a file matching groovy-language-server-*.jar"
                   :server-id 'groovy-ls))
 
 ;;; TypeScript/JavaScript
+
+(lsp-dependency javascript-typescript-langserver
+  (:system "javascript-typescript-stdio")
+  (:npm :package "javascript-typescript-langserver"
+        :path ".bin/javascript-typescript-stdio"))
+
 (defgroup lsp-typescript-javascript nil
   "Support for TypeScript/JavaScript, using Sourcegraph's JavaScript/TypeScript language server."
   :group 'lsp-mode
   :link '(url-link "https://github.com/sourcegraph/javascript-typescript-langserver"))
-
-(defcustom lsp-clients-javascript-typescript-server "javascript-typescript-stdio"
-  "The javascript-typescript-stdio executable to use.
-Leave as just the executable name to use the default behavior of
-finding the executable with variable `exec-path'."
-  :group 'lsp-typescript-javascript
-  :risky t
-  :type 'file)
 
 (defcustom lsp-clients-typescript-javascript-server-args '()
   "Extra arguments for the typescript-language-server language server."
@@ -179,13 +177,17 @@ finding the executable with variable `exec-path'."
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
-                                                          (cons lsp-clients-javascript-typescript-server
+                                                          (cons (lsp-package-path 'javascript-typescript-langserver)
                                                                 lsp-clients-typescript-javascript-server-args)))
                   :activation-fn 'lsp-typescript-javascript-tsx-jsx-activate-p
                   :priority -3
                   :completion-in-comments? t
-                  :ignore-messages '("readFile .*? requested by TypeScript but content not available")
-                  :server-id 'jsts-ls))
+                  :server-id 'jsts-ls
+                  :download-server-fn (lambda (_client callback error-callback _update?)
+                                        (lsp-package-ensure
+                                         'javascript-typescript-langserver
+                                         callback
+                                         error-callback))))
 
 
 ;;; TypeScript
@@ -193,14 +195,6 @@ finding the executable with variable `exec-path'."
   "LSP support for TypeScript, using Theia/Typefox's TypeScript Language Server."
   :group 'lsp-mode
   :link '(url-link "https://github.com/theia-ide/typescript-language-server"))
-
-(defcustom lsp-clients-typescript-server "typescript-language-server"
-  "The typescript-language-server executable to use.
-Leave as just the executable name to use the default behavior of
-finding the executable with variable `exec-path'."
-  :group 'lsp-typescript
-  :risky t
-  :type 'file)
 
 (defcustom lsp-clients-typescript-server-args '("--stdio")
   "Extra arguments for the typescript-language-server language server."
@@ -230,18 +224,37 @@ directory containing the package. Example:
                                                   (and name location))
                                                 xs)))))
 
+(lsp-dependency typescript-language-server
+  (:system "typescript-language-server")
+  (:npm :package "typescript-language-server"
+        :path ".bin/typescript-language-server"))
+
+(lsp-dependency typescript
+  (:system "tsserver")
+  (:npm :package "typescript"
+        :path ".bin/tsserver"))
+
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
-                                                          (cons lsp-clients-typescript-server
+                                                          (cons (lsp-package-path 'typescript-language-server)
                                                                 lsp-clients-typescript-server-args)))
                   :activation-fn 'lsp-typescript-javascript-tsx-jsx-activate-p
                   :priority -2
                   :completion-in-comments? t
                   :initialization-options (lambda ()
                                             (list :plugins lsp-clients-typescript-plugins
-                                                  :logVerbosity lsp-clients-typescript-log-verbosity))
+                                                  :logVerbosity lsp-clients-typescript-log-verbosity
+                                                  :tsServerPath (lsp-package-path 'typescript)))
                   :ignore-messages '("readFile .*? requested by TypeScript but content not available")
-                  :server-id 'ts-ls))
+                  :server-id 'ts-ls
+                  :download-server-fn (lambda (_client callback error-callback _update?)
+                                        (lsp-package-ensure
+                                         'typescript
+                                         (-partial #'lsp-package-ensure
+                                                   'typescript-language-server
+                                                   callback
+                                                   error-callback)
+                                         error-callback))))
 
 
 
@@ -338,19 +351,18 @@ particular FILE-NAME and MODE."
 
 (defun lsp-php--create-connection ()
   "Create lsp connection."
-  (plist-put
-   (lsp-stdio-connection
-    (lambda () lsp-clients-php-server-command))
-   :test? (lambda ()
-            (if (and (cdr lsp-clients-php-server-command)
-                     (eq (string-match-p "php[0-9.]*\\'" (car lsp-clients-php-server-command)) 0))
-                ;; Start with the php command and the list has more elems. Test the existence of the PHP script.
-                (let ((php-file (nth 1 lsp-clients-php-server-command)))
-                  (or (file-exists-p php-file)
-                      (progn
-                        (lsp-log "%s is not present." php-file)
-                        nil)))
-              t))))
+  (lsp-stdio-connection
+   (lambda () lsp-clients-php-server-command)
+   (lambda ()
+     (if (and (cdr lsp-clients-php-server-command)
+              (eq (string-match-p "php[0-9.]*\\'" (car lsp-clients-php-server-command)) 0))
+         ;; Start with the php command and the list has more elems. Test the existence of the PHP script.
+         (let ((php-file (nth 1 lsp-clients-php-server-command)))
+           (or (file-exists-p php-file)
+               (progn
+                 (lsp-log "%s is not present." php-file)
+                 nil)))
+       t))))
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-php--create-connection)
@@ -390,7 +402,9 @@ particular FILE-NAME and MODE."
   "LSP support for OCaml, using ocaml-lsp-server."
   :group 'lsp-mode
   :link '(url-link "https://github.com/ocaml/ocaml-lsp"))
+
 (define-obsolete-variable-alias 'lsp-merlin 'lsp-ocaml-lsp-server)
+(define-obsolete-variable-alias 'lsp-merlin-command 'lsp-ocaml-lsp-server-command)
 
 (defcustom lsp-ocaml-lsp-server-command
   '("ocamllsp")
@@ -400,7 +414,6 @@ particular FILE-NAME and MODE."
           (string :tag "Single string value")
           (repeat :tag "List of string values"
                   string)))
-(define-obsolete-variable-alias 'lsp-merlin-command 'lsp-ocaml-lsp-server-command)
 
 (lsp-register-client
  (make-lsp-client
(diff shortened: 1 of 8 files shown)


lsp-pwsh.el
```diff
@@ -292,26 +286,28 @@ Must not nil.")
 
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection #'lsp-pwsh--command)
+  :new-connection (lsp-stdio-connection #'lsp-pwsh--command
+                                        (lambda ()
+                                        (-> lsp-pwsh-dir file-name-directory f-exists?)))
```
kiennq: This line should be indented.

lsp-pwsh.el
```diff
@@ -321,40 +317,42 @@ Must not nil.")
...
+
+  (unless (and lsp-pwsh-exe (file-executable-p lsp-pwsh-exe))
+    (user-error "Use `lsp-pwsh-exe' with the value of `%s' is not a valid powershell binary"
+                lsp-pwsh-exe))
+
+  (let ((parent-dir (file-name-directory lsp-pwsh-dir))
+        (url (format lsp-pwsh-github-asset-url "PowerShell"
+                     "PowerShellEditorServices" "PowerShellEditorServices.zip"))
+        (temp-file (make-temp-file "ext" nil ".zip")))
+    (unless (and (not update) (file-exists-p parent-dir))
```
kiennq: Since you're merging `lsp-pwsh-setup` and `lsp-pwsh--get-extension`, it would be great if you can make some minor changes too.
This should check for `(f-exists? lsp-pwsh-dir)`

lsp-pwsh.el
```diff
@@ -321,40 +317,42 @@ Must not nil.")
...
+      ;; since we know it's installed, use powershell to download the file
+      ;; (and avoid url.el bugginess or additional libraries)
+      (lsp-async-start-process
+       (lambda (success? msg)
+         (if success?
+             (progn
+               (lsp--info "lsp-pwsh: Downloading done!")
+               (when (file-exists-p parent-dir) (delete-directory parent-dir 'recursive))
```
kiennq: This should be come `(when (f-exists? lsp-pwsh-dir) (delete-directory lsp-pwsh-dir 'recursive))`

kiennq:
Thanks, the changes for `lsp-pwsh` looks great to me.

seagle0128:
I will take a look after this is merged.
Since `lsp-python-ms` is working well for now, @yyoncho do you think it's necessary to port? And it's able to handle downloading and extracting the `mspyls` package? Of course, I can see `lsp-python-ms-dir` should be updated accordingly.

lsp-pwsh.el
```diff
@@ -32,7 +32,7 @@
 
 (defgroup lsp-pwsh nil
   "LSP support for PowerShell, using the PowerShellEditorServices."
-  :group 'lsp-mode
+  :group 'lsp
```
seagle0128: Why changed to `lsp` group?

yyoncho: will be reverted.

yyoncho:
> I will take a look after this is merged.
> Since `lsp-python-ms` is working well for now, @yyoncho do you think it's necessary to port? And it's able to handle downloading and extracting the `mspyls` package? Of course, I can see `lsp-python-ms-dir` should be updated accordingly.

It is not something urgent - it will make lsp-mode consistent since there will be a unified way to install the server. It will help also implementing the dashboard with available servers - ATM you cannot check whether the server is present because it will trigger installation. I will do a PR.

yyoncho:
Addressed @kiennq and @seagle0128 comments.

razzmatazz:
@yyoncho I think we should store lsp server files under ~/.emacs.d/.cache instead of ~/.emacs.d/lsp-servers (as defined in `lsp-server-install-dir`).

I am not 100% sure about convention on this but:
 - spacemacs saves all the "ephemeral/cache" data to that folder
 - here, https://ambrevar.xyz/emacs/#orgfcba11b it says that other modes save data there too; 
   - `Many modes store their cache files in ~/.emacs.d. I prefer to keep those ephemeral files in ~/.cache/emacs.`
 - at least on my installation I find a lot of "cache" stuff there so there is a grain of truth to it

razzmatazz:
Otherwise the changes look good to me, I tried both installing and updating the server.

I will try to implement async download mechanism via `url-retrieve` which seems to need some wrapper unlinke `url-copy-file` lsp-csharp and lsp-fsharp uses ATM.

yyoncho:
> @yyoncho I think we should store lsp server files under `~/.emacs.d/.cache/lsp-servers` instead of `~/.emacs.d/lsp-servers` (as defined in `lsp-server-install-dir`).

Make sense, I will update it.

> I will try to implement async download mechanism via url-retrieve which seems to need a wrapper unlike the url-copy-file function which both lsp-csharp and lsp-fsharp uses to download a binary ATM.

I wish we can have the download method in lsp-mode.el so we can reuse it across servers. E. g. lsp-url-retrieve(url store-location callback) which first tries pwsh, wget, curl and fallbacks to url-retrieve. 

The same about the extract method - based on the extension of the file checks for pwsh, unzip, untar, etc. If not present - shows an error.

lsp-pwsh.el
```diff
@@ -321,40 +317,41 @@ Must not nil.")
...
+       (lambda (success? msg)
+         (if success?
+             (progn
+               (lsp--info "lsp-pwsh: Downloading done!")
+               (when (file-exists-p lsp-pwsh-dir) (delete-directory lsp-pwsh-dir 'recursive))
+               (lsp-async-start-process
+                callback
+                lsp-pwsh-exe "-noprofile" "-noninteractive" "-nologo"
+                "-ex" "bypass" "-command" "Expand-Archive"
+                "-Path" temp-file "-DestinationPath" lsp-pwsh-dir))
```
kiennq: This needs to be `(file-name-directory lsp-pwsh-dir)`, the parent of `lsp-pwsh-dir`. The reason is because the archive is containing a folder that has same name as `lsp-pwsh-dir` so we need to expand it into parent folder for it to replace `lsp-pwsh-dir`.

akirak:
@yyoncho Thank you for this PR. I think it is great, and I am looking forward to seeing a new version of lsp-mode getting this facility.

I don't know why it needs an integration with Nix when it can download and install server executables in itself. Yet I like to install servers using Nix, so let me leave a few comments on this ticket.

> All servers(when feasible) should be installed under lsp-server-install-dir.

Is it mandatory for the nix wrapper to follow this convention? Nix installs executables in ~/.nix-profile/bin by default. Executables installed as part of npm packages won't have `.bin` prefix, so if it needs to suppoort Nix, `lsp-package-path` needs to have a slightly different API.

lsp-clients.el
```diff
@@ -179,13 +171,18 @@ finding the executable with variable `exec-path'."
...
                   :completion-in-comments? t
-                  :ignore-messages '("readFile .*? requested by TypeScript but content not available")
-                  :server-id 'jsts-ls))
+                  :server-id 'jsts-ls
+                  :download-server-fn (lambda (_client callback _update?)
+                                        (lsp-package-ensure
+                                         :npm
+                                         "javascript-typescript-langserver"
+                                         (lambda (success? msg)
+                                           (funcall callback success? msg))))))
```
akirak: Here each callback function takes a boolean status flag as an argument. In JavaScript, promises take success and error callbacks separately. I think this is a matter of preference, but I'd personally prefer `lsp-package-ensure` having two distinct callback functions. What do you think? Depending on your decision, I will extend the API of my `nix-env-install` to support callbacks.

yyoncho:
> Is it mandatory for the nix wrapper to follow this convention? Nix installs executables in ~/.nix-profile/bin by default. Executables installed as part of npm packages won't have `.bin` prefix, so if it needs to suppoort Nix, `lsp-package-path` needs to have a slightly different API.

No, let's keep it in that folder only if it makes sense. Can you suggest better interface for lsp-package-path? I am not npm/nix expert. Not sure whether striping/adding .bin to the path is an option

lsp-clients.el
yyoncho: I will experiment which interface result in better code.

akirak:
I am thinking of an API like this:

```emacs-lisp
(defcustom lsp-package-backends
  '(nix npm builtin
    ;; system simply finds an executable from PATH
    system))
  "Package managers that can be used to install server executables."
  :type '(repeat symbol))

(defun lsp-find-executable (command packages)
  (-some (lambda (backend)
           (cond
            ((eq backend 'system)
             (executable-find command))
            ((lsp-backend-available-p backend)
             (lsp-backend-find-executable backend command
                                       (alist-get backend packages)))))
         lsp-package-backends))

(cl-defgeneric lsp-backend-available-p (backend))
(cl-defgeneric lsp-backend-find-executable (backend command &optional package))

;; An example client definition depending on an external package registry
(lsp-register-client
 (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
                                                          (cons "javascript-typescript-stdio"
                                                                lsp-clients-typescript-javascript-server-args)))
                  :dependencies '(("javascript-typescript-stdio"
                                   ((npm "javascript-typescript-langserver"))))
                  :activation-fn 'lsp-typescript-javascript-tsx-jsx-activate-p
                  :priority -3
                  :completion-in-comments? t
                  :server-id 'jsts-ls))

;; An example client with builtin download function
(lsp-register-client
 (make-lsp-client ...
                  :dependencies '(("fsautocomplete" ((builtin lsp-fsharp--fsac-install))))
```

The executable path needs to be resolved by `lsp-find-executable`, and `:dependencies` field specifies possible packages for each executables. Many of the existing server installation functions will be handled by `builtin`. Each dependency consists of an executable name, a list of package specs, and an optional plist for future enhancements (e.g. to specify it is an optional dependency). It would probably need a lot of rewrites, but it can support multiple package backends.

> Not sure whether striping/adding .bin to the path is an option

`.bin` is specific to npm, and executables installed using nix don't contain the prefix. If you only need to find executables, I think you can omit it.

I personally find the API of [eglot](https://github.com/joaotavora/eglot) to be more concise yet satisfactory than this package. Unfortunately, it didn't work with bash-language-server, and lsp-mode seems to have more resources and velocity, so I will probably use this package.

lsp-clients.el
akirak: You might want to take a look at [emacs-async](https://github.com/jwiegley/emacs-async) for design of async APIs. It would be possible to support concurrent downloads, even if you probably don't need it in this package.

akirak:
> I think we should store lsp server files under ~/.emacs.d/.cache/lsp-servers instead of ~/.emacs.d/lsp-servers (as defined in lsp-server-install-dir).

This can be enforced by [no-littering](https://github.com/emacscollective/no-littering) package. It would be better to add this package to no-littering once this PR is merged. (If anyone doesn't, I will.)

Nonetheless, the author of the package [suggests](https://github.com/emacscollective/no-littering/pull/100#issuecomment-525041932) that you not add an extra prefix, i.e. `.cache`. Therefore `(locate-user-emacs-file "lsp-servers/")` would be usually sufficient.

Related discussion: https://github.com/syl20bnr/spacemacs/issues/5947

yyoncho:
@akirak Thank you.

Your suggestion looks fine except for the fact that we are not always looking for executables - the eglot implementation is oversimplified which leads to reports like this - https://github.com/eclipse/eclipse.jdt.ls/issues/748 and awkward workarounds. And yes, having all of the so-called providers under the same interface does not make sense and won't work. Following your example, I would modify it to include the parameters needed for each backend: 

``` elisp
; dependency declaration follows
((:system "typescript-language-server")
 (:nix :package "typescript-language-server" :binary-name "the-binary-name")
 (:npm :package "typescript-language-server" :path ".bin/typescript-language-server")) ;; here we don't always look for binary and we need the binary name.

((:custom :function ,lsp-pwsh--download))
```

I would also give up the generics for simple defcustom to make disabling of particular provider easier - see https://stackoverflow.com/questions/59201891/how-to-undefine-defmethods

yyoncho:
@razzmatazz @razzmatazz 

> > I think we should store lsp server files under ~/.emacs.d/.cache/lsp-servers instead of ~/.emacs.d/lsp-servers (as defined in lsp-server-install-dir).
> 

For me, it works either way.

akirak:
@yyoncho Thank you for your comment. I don't know much about how `lsp-mode` works, so please forgive me for not covering the whole requirements.

I'd prefer dependency definitions like this to avoid duplicates:

```emacs-lisp
(defun lsp-check-dependencies (dependencies)
  (dolist (dep dependencies)
    (pcase dep
      (`(executable ,name ,specs)
       (unless (lsp-find-executable name)
         ;; Install a package from one of the specs
         \.\.\.)))))

(defun lsp-find-executable (command packages)
  (-some (lambda (backend)
           (cond
            ((eq backend 'system)
             (executable-find command))
            ;; Nix can handle multiple package repositories
            ((eq backend 'nix)
             (lsp-backend-find-executable backend command
                                          packages))
            ((lsp-backend-available-p backend)
             (lsp-backend-find-executable backend command
                                          (alist-get backend packages)))))
         lsp-package-backends))

;; Define dependencies
'((executable "typescript-language-server"
              ((npm :package "typescript-language-server")
               ;; Add an explicit nix package once it is available in the official repo
               (nix :repo nixpkgs :package "typescript-language-server"))))
'((executable fsautocomplete-executable
              ((builtin :function lsp-fsharp--fsac-install))))
```

Nix can support various types of packages (e.g. npm, pip, etc.) through third-party wrappers as well as from its official package repository nixpkgs. I have already added support for [node2nix](https://github.com/svanderburg/node2nix) to support npm packages, and I may support other external package systems in the future. Thus the Nix provider will receive all package specs during installation. Yes, all backends/provides are not equal, and nix is treated as special in this example.

> I would also give up the generics for simple defcustom to make disabling of particular provider easier.

You can disable particular providers by customizing `lsp-package-backends` variable:

```emacs-lisp
(defcustom lsp-package-backends
  '(nix npm builtin
        ;; system simply finds an executable from exec-path
        system)
  "Package managers that can be used to install server executables."
  :type '(repeat symbol))

(use-package lsp-mode
  :custom
  ;; Use only specific backends/providers
  (lsp-package-backends '(npm builtin system)))

(use-package lsp-mode
  :config
  ;; Disable nix backend
  (cl-delete 'nix lsp-package-backends))
```

I personally don't want to let users write generics to customize a package. It might not be easy for some people to understand generics in (CL) Emacs Lisp. Also, it will become difficult to upgrade the API of generic methods if you allow users to redefine implementations. Customizing variables is the standard way in Emacs and hides the internals.

If you like this idea, I will send a PR to help with the implementation. I want to do that, because it will help me understand this package. Thanks.

yyoncho:
> I personally don't want to let users write generics to customize a package. It might not be easy for some people to understand generics in (CL) Emacs Lisp. Also, it will become difficult to upgrade the API of generic methods if you allow users to redefine implementations. Customizing variables is the standard way in Emacs and hides the internals.

Yes, I agree - we should have customization options for the common scenarios and empower the expert users to do advanced customizations via elisp. 

> If you like this idea, I will send a PR to help with the implementation. I want to do that, because it will help me understand this package. Thanks

I like the idea but I am already working on the change. So let me finish the initial version then we will let it evolve to make it easier to use (e. g. have common methods for extracting/downloading deps)  and complete (e. g. support nix, vscode market place, etc). Also, feel free to pick any of the open items or propose/discuss over the existing feature - we encourage contributions.

akirak:
@yyoncho I see. Please merge this PR as soon as it is finished. It looks already sufficient, and I don't think it needs to support Nix right now.

> I like the idea but I am already working on the change. So let me finish the initial version then we will let it evolve to make it easier to use (e. g. have common methods for extracting/downloading deps)

I wish to continue my experiments with Nix by manually installing servers, so I will probably disable the automatic downloading feature for some languages, but it should never hinder your development. Thank you for sharing your contexts.

Issue #726, opened 2019-03-22
lsp-before-uninitialized-hook is not called


seagle0128:
I am looking for the solution to do something after shutdown server. I found `lsp-before-uninitialized-hook` is not called by any functions. But in `lsp--uninitialize-workspace`, `lsp-workspace-uninitialized-hook` is called.

Is it as designed or an issue?

yyoncho:
It is has regressed, I guess during the big restructuring, I will fix it.

yyoncho:
Added `lsp-after-uninitialized-hook` which I think makes more sense. Please let me know if you need another hook `lsp-before-uninitialized-hook`.

seagle0128:
`lsp-after-uninitialized-hook` is fine for me, but it seems `lsp-workspace-uninitialized-hook` is still undefined. Should it be removed?

yyoncho:
> Should it be removed?

Done.

seagle0128:
Thank you!

PR #1279, opened 2019-12-29
Rename lsp-after-uninitialized-hook and lsp-workspace-folders-changed-hook to follow "normal hook" convention

### LINKED ISSUE #1257: lsp-after-uninitialized-hook not following "abnormal hook" conventions
`lsp-after-uninitialized-hook` name ends with `-hook` and is documented as "List of functions to be called after a Language Server has been uninitialized."

Based on this, one would try to write a regular hook function, i.e. one that takes no args and whose return value is ignored. But this will fail with a "wrong number of arguments" error, because the `lsp-after-uninitialized-hook` passes workspace arg to hook functions.

Based on [Emacs Lisp manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Hooks.html), hooks that take args are called "abnormal hooks", their naming convention is to end with `-functions` instead of `-hook`, and the documentation should describe what is passed (or what is done with the return value).

Grepping for `run-hook-with-args` suggests that there is one more abnormal hook like this one: `lsp-workspace-folders-changed-hook`. This one is documented to take args, but is not suffixed with `-functions` neither.

MaskRay:
By convention, `-hook` variables mean normal hooks, which contain lists
of functions to be called with no arguments.

Fixes #1257

-----

`lsp-after-uninitialized-hook` is young (#726) and is probably only used by doom-modeline. @seagle0128 may make the change to the downstream package and we can save a `define-obsolete-variable-alias`.

### DIFF
--- a/doc/src/lsp-mode-vars.adoc
+++ b/doc/src/lsp-mode-vars.adoc
@@ -211,15 +211,15 @@ List of regexps matching directory paths which won’t be monitored when creatin
 NOTE: Introduced in `lsp-mode` 6.1
 
 ____
-[id="lsp-after-uninitialized-hook"]
-- `lsp-after-uninitialized-hook`
+[id="lsp-after-uninitialized-functions"]
+- `lsp-after-uninitialized-functions`
 ____
 Default value: `pass:[(doom-modeline-update-lsp)
 ]`
 
 List of functions to be called after a Language Server has been uninitialized.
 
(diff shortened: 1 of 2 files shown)


seagle0128:
@MaskRay Okay. The patch is ready and will be submitted as song as this PR is merged.

yyoncho:
Merging as it is. IMO in the future, we should add these fixes into the changelog right away.

PR #1285, opened 2019-12-30
Make lsp-completion-at-point search symbol instead of word-strictly

### LINKED ISSUE #1284: `lsp-completion-at-point` is broken for function contains symbol in `rust-mode` with `company-capf`
**Describe the bug**
With 45cb3c70, the `bounds` of current symbol are set to `(bounds-of-thing-at-point 'word-strictly)`, this makes `company-capf` become broken for function contains symbol in name in `rust-mode`.

For example, when type `test_` and trigger capf completion, since `_` is symbol and not a part of word in `rust-mode`, the `bounds` will return `nil`.
This makes suggestion for prefix `test_` not correct anymore.

**To Reproduce**
With following rust code where `|` is the cursor position, trigger completion at point

``` rust
fn test_a() {
    println!("Hello!");
}

fn main() {
    test_|
}
```
Result 
image

You can see that `test_test_a` is inserted, the reason is because the bounds are `nil`, thus trigger `company-capf` to insert the whole `test_a` candidate without knowing that the prefix is `test_`.

**Expected behavior**
Candidates for prefix `test_` should be shown
image

`lsp-rust`

**OS**
Windows 10

kiennq:
This should fix #1284 

Also, make `:company-prefix-length` to return `t` or `nil` depends on the previous chars are `trigger-chars` or not.
Currently, it returns (point) which is pretty pointless since with point far enough, that value will always larger than `company-minimum-prefix-length` and make the completion always being triggered.
We should respect `company-minimum-prefix-length` and only forcefully trigger completion if the `trigger-chars` are typed.

Also, respect `lsp--client-completion-incomments?`

Also allow a degree of flex filtering if set via `completion-styles` and language that support it (example json language server)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3702,77 +3702,105 @@ and the position respectively."
   "Return non-nil if text before point matches any of the trigger characters."
   (looking-back (regexp-opt (cl-coerce trigger-characters 'list)) (line-beginning-position)))
 
-(add-to-list 'completion-category-defaults '(lsp-substring (styles . (substring))))
-
 (defun lsp-completion-at-point ()
   "Get lsp completions."
-  (let* ((bounds (bounds-of-thing-at-point 'word-strictly))
-         (trigger-chars (->> (lsp--server-capabilities)
-                             (gethash "completionProvider")
-                             (gethash "triggerCharacters")))
-         start-point result done?)
-    (list
-     (if bounds (car bounds) (point))
-     (point)
-     (lambda (_probe _pred action)
-       (cond
-        ((eq action 'metadata)
-         `(metadata . ((display-sort-function . (lambda (candidates)
-                                                  (--sort (string-lessp (get-text-property 0 'lsp-sort-text it)
-                                                                        (get-text-property 0 'lsp-sort-text other))
-                                                          candidates)))
-                       (category . lsp-substring))))
-        ((eq (car-safe action) 'boundaries) nil)
-        (t (if done?
-               result
-             (setf start-point (point))
-             (let* ((resp (lsp-request-while-no-input "textDocument/completion"
-                                                      (plist-put (lsp--text-document-position-params)
-                                                                 :context (ht ("triggerKind" 1)))))
-                    (items (lsp--sort-completions (cond
-                                                   ((seqp resp) resp)
-                                                   ((hash-table-p resp) (gethash "items" resp))))))
-               (setf done? (or (seqp resp)
-                               (not (gethash "isIncomplete" resp)))
-                     result (seq-into (seq-map (-lambda ((item &as &hash "label"
-                                                               "insertText" insert-text
-                                                               "sortText" sort-text))
-                                                 (propertize (or label insert-text)
-                                                             'lsp-completion-item item
-                                                             'lsp-sort-text sort-text))
-                                               items)
-                                      'list)))))))
-
-     :annotation-function #'lsp--annotate
-     :company-require-match 'never
-     :company-prefix-length (point)
-     :exit-function
-     (lambda (candidate _status)
-       (-let [(&hash "insertText" insert-text
-                     "textEdit" text-edit
-                     "insertTextFormat" insert-text-format
-                     "additionalTextEdits" additional-text-edits)
-              (lsp--resolve-completion (plist-get (text-properties-at 0 candidate) 'lsp-completion-item))]
+  (when (or (--some (lsp--client-completion-in-comments? (lsp--workspace-client it))
+                    (lsp-workspaces))
+            (not (nth 4 (syntax-ppss))))
+    (let* ((bounds (bounds-of-thing-at-point 'symbol))
+           (trigger-chars (->> (lsp--server-capabilities)
+                               (gethash "completionProvider")
+                               (gethash "triggerCharacters")))
+           start-point result done?)
+      (list
+       (if bounds (car bounds) (point))
+       (point)
+       (lambda (_probe pred action)
          (cond
-          (text-edit
-           (delete-region start-point (point))
-           (lsp--apply-text-edit text-edit))
-          (insert-text
-           (delete-region (- (point) (length candidate)) (point))
-           (insert insert-text)))
+          ((eq action 'metadata)
+           `(metadata . ((display-sort-function
+                          . (lambda (candidates)
+                              (--sort (string-lessp
+                                       (get-text-property 0 'lsp-sort-text it)
+                                       (get-text-property 0 'lsp-sort-text other))
+                                      candidates))))))
+          ((eq (car-safe action) 'boundaries) nil)
+          (t (if done?
+                 result
+               (setf start-point (point))
+               (let* ((resp (lsp-request-while-no-input "textDocument/completion"
+                                                        (plist-put (lsp--text-document-position-params)
+                                                                   :context (ht ("triggerKind" 1)))))
+                      (items (lsp--sort-completions (cond
+                                                     ((seqp resp) resp)
+                                                     ((hash-table-p resp) (gethash "items" resp)))))
+                      (probe
+                       (-if-let ((start . end)
+                                 (or (-some-> (lsp-elt items 0)
+                                       (lsp--ht-get "textEdit" "range")
+                                       lsp--range-to-region)
+                                     bounds))
+                           (buffer-substring-no-properties start end)
+                         "")))
+                 (setf done? (or (seqp resp)
+                                 (not (gethash "isIncomplete" resp)))
+                       result (-some--> items
+                                (-map (-lambda ((item &as &hash
+                                                      "label"
+                                                      "filterText" filter-text))
+                                        (propertize (or filter-text label)
+                                                    'lsp-completion-item item))
+                                      it)
+                                (seq-into it 'list)
+                                (completion-all-completions probe it pred (length probe))
+                                ;; completion-all-completions may return a list in form (a b . x)
+                                ;; the last cdr is not important and need to be removed
+                                (let ((tail (last it)))
+                                  (if (consp tail) (setcdr tail nil))
+                                  it)
+                                (-map (-partial #'get-text-property 0 'lsp-completion-item) it)
+                                (-map (-lambda ((item &as &hash
+                                                      "label"
+                                                      "insertText" insert-text
+                                                      "sortText" sort-text))
+                                        (propertize (or label insert-text)
+                                                    'lsp-completion-item item
+                                                    'lsp-sort-text sort-text))
+                                      it))))))))
 
-         (when (eq insert-text-format 2)
-           (yas-expand-snippet (buffer-substring start-point (point))
-                               start-point
-                               (point)))
-         (when additional-text-edits
-           (lsp--apply-text-edits additional-text-edits)))
+       :annotation-function #'lsp--annotate
+       :company-require-match 'never
+       :company-prefix-length
+       (save-excursion
+         (when (car bounds) (goto-char (car bounds)))
+         (lsp--looking-back-trigger-characters-p trigger-chars))
+       :exit-function
+       (lambda (candidate _status)
+         (-let [(&hash "insertText" insert-text
+                       "textEdit" text-edit
+                       "insertTextFormat" insert-text-format
+                       "additionalTextEdits" additional-text-edits)
+                (lsp--resolve-completion (plist-get (text-properties-at 0 candidate) 'lsp-completion-item))]
+           (cond
+            (text-edit
+             (delete-region start-point (point))
+             (lsp--apply-text-edit text-edit))
+            (insert-text
+             (delete-region (- (point) (length candidate)) (point))
+             (insert insert-text)))
 
-       (when lsp-signature-auto-activate
-         (lsp-signature-activate))
+           (when (eq insert-text-format 2)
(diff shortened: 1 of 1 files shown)


yyoncho:
Using symbol breaks other language servers, e. g. In java if you have @foobar the prefix should be @foobar. I suggest using symbol or word-strictly whatever is bigger.

kiennq:
> Using symbol breaks other language servers, e. g. In java if you have @foobar the prefix should be @foobar. I suggest using symbol or word-strictly whatever is bigger.

Do you mean the prefix should be **foobar**? Since that's what `word-strictly` returned.

muirdm:
The `insertText` in the response from the server has the correct prefix, right? Can `lsp-completion-at-point` just use the response instead of `thing-at-point` to determine the text to replace?

yyoncho:
@kiennq I will be with no access to emacs for next 3-4 days, when I am back I will test that scenario. In general the filtering and the caching in capf does not work.

kiennq:
> The `insertText` in the response from the server has the correct prefix, right? Can `lsp-completion-at-point` just use the response instead of `thing-at-point` to determine the text to replace?

From what I saw the `insertText` can contain snippet depend on `insertTextFormat`, so using the response may not work. Worse, there is `textEdits` inside response that may affect how to calculate prefix too.



> @kiennq I will be with no access to emacs for next 3-4 days, when I am back I will test that scenario. In general the filtering and the caching in capf does not work.

No worries, happy new year!

yyoncho:
Both symbol and word-strictly are wrong as a calculation but based on your findings symbol is less wrong since it handles properly the _. 

The correct way to handle client-side filtering is:

1. Do not calculate the prefix upfront
2. Use textEdit range to calculate the prefix and do the filtering if it is present.

> * *Note:* The range of the edit must be a single line range and it must contain the position at which completion
> * has been requested.

3. If textEdit is not present - we should use the same "logic" that vscode uses for calculating the prefix.

few examples of the bizarre logic in vscode: 
* If you are in html and you have `<foo` the prefix is `foo`
* If you are in xml and you have `<foo` the prefix is `<foo` 

4. We cannot use trigger chars to determine the range like vscode does since some of the servers(e. g. spring-boot have the whole alphabet as trigger chars.

Let me know how do you want to proceed:

1. Merge it as it is since it works better (e. g. only java annotations will be broken).
2. Discuss and implement what I have described in this comment.

kiennq:
@yyoncho Can you point me to the logic where vscode is determining the prefix?

For now let's just discuss and try to implement what you suggested.

I'm thinking of something along the line of passing the prefix determining logic to the lsp--client itself, when we cannot determine the prefix through `textEdit` range

yyoncho:
> Can you point me to the logic where vscode is determining the prefix?

It is in `src/vs/editor/contrib/suggest/suggest.ts` - I am not 100% that vscode does the first part with textEdit - the point is that everyone is testing with vscode thus part 3) is sufficient. 

The filtering is here: https://github.com/Microsoft/vscode/blob/master/src/vs/base/common/filters.ts#L524 - I am not sure whether the elisp will be fast enough to implement that so we might do a feature request against Emacs core to implement it in the core. We could go with what company-lsp which has simplified version of it.

> I'm thinking of something along the line of passing the prefix determining logic to the lsp--client itself, when we cannot determine the prefix through textEdit range

I want to avoid any per client adjustments because they usually don't work - I would rather go with: use textEdit range and fallback `(thing-at-point 'symbol)` - it is good enough most of the time. If does not work in a certain case - we might report that in the server repo because insertText is deprecated and everyone should use textEdit. E. g. for example we are doing client side filtering for client configured servers - this is incorrect - we should do that for **all** language servers like vscode does - https://github.com/microsoft/language-server-protocol/issues/105

kiennq:
> I want to avoid any per client adjustments because they usually don't work - I would rather go with: use textEdit range and fallback `(thing-at-point 'symbol)` - it is good enough most of the time. If does not work in a certain case - we might report that in the server repo because insertText is deprecated and everyone should use textEdit. For example we are doing client side filtering for client configured servers - this is incorrect - we should do that for **all** language servers like vscode does - [microsoft/language-server-protocol#105](https://github.com/microsoft/language-server-protocol/issues/105)

Interesting, let me try implement this.

kiennq:
Updated. Now we're determining prefix by `textEdit` or fallback to symbol.
@yyoncho

lsp-mode.el
```diff
@@ -3706,73 +3706,94 @@ and the position respectively."
...
-     (lambda (candidate _status)
-       (-let [(&hash "insertText" insert-text
-                     "textEdit" text-edit
-                     "insertTextFormat" insert-text-format
-                     "additionalTextEdits" additional-text-edits)
-              (lsp--resolve-completion (plist-get (text-properties-at 0 candidate) 'lsp-completion-item))]
+  (when (or (--some (lsp--client-completion-in-comments? (lsp--workspace-client it))
+                    (lsp-workspaces))
+            (not (nth 4 (syntax-ppss))))
+    (let* ((trigger-chars (->> (lsp--server-capabilities)
```
yyoncho: We are not allowed to do that here, as per docs:

```
NOTE: These functions should be cheap to run since they're sometimes
run from `post-command-hook'; and they should ideally only choose
which kind of completion table to use, and not pre-filter it based
on the current text between START and END (e.g., they should not
obey `completion-styles').
```

We should either by creating a new completion style or by doing that in the `(lambda (probe pred action)` function.

lsp-mode.el
```diff
@@ -3706,73 +3706,94 @@ and the position respectively."
...
-     :company-require-match 'never
-     :company-prefix-length (point)
-     :exit-function
-     (lambda (candidate _status)
-       (-let [(&hash "insertText" insert-text
-                     "textEdit" text-edit
-                     "insertTextFormat" insert-text-format
-                     "additionalTextEdits" additional-text-edits)
-              (lsp--resolve-completion (plist-get (text-properties-at 0 candidate) 'lsp-completion-item))]
+  (when (or (--some (lsp--client-completion-in-comments? (lsp--workspace-client it))
```
yyoncho: Did you happen to find a server which does not want completion in comments and strings? Because I want to change the default.

lsp-mode.el
```diff
@@ -3706,73 +3706,94 @@ and the position respectively."
...
+                                       (get-text-property 0 'lsp-sort-text it)
+                                       (get-text-property 0 'lsp-sort-text other))
+                                      candidates)))
+                         (category . lsp-substring))))
+          ((eq (car-safe action) 'boundaries) nil)
+          (t
+           (let* ((str (replace-regexp-in-string
+                        (rx (and bos (1+ (not alnum))))
+                        "" probe))
+                  (candidates (completion-all-completions
```
yyoncho: here we should match against filterText property of CompletionItem, not against the label.

lsp-mode.el
kiennq: I think most of the server don't want completion in comment. Inside string, having language aware completion is fine.

lsp-mode.el
kiennq: Hmm, we need server response to determine the bound, and it's not recommended to have long running code here.
I don't know how we can progress from here.
We need to return `(start end collection . props)` quickly for every call to completion at point, so determine `start` and `end` using server response seems to be out of reach now.

kiennq:
I'm more curious about the case of java being broken with symbol bounds. Can you elaborate more about that? From what I saw, the symbols bound correctly return `@foobar` for java.

yyoncho:
> I'm more curious about the case of java being broken with symbol bounds. Can you elaborate more about that? From what I saw, the symbols bound correctly return `@foobar` for java.

It actually should be `foobar` to work (my first comment was wrong). This is the reason company-lsp is using trigger chars to calculate the prefix. But this is not the only problematic case: consider that you are completing namespace in Java, e. g. `java.util` and `java.collections`. When you have `java.foo` symbol at point will calculate `foo` but it should be `java.foo`.

lsp-mode.el
yyoncho: It needs the bounds to calculate the prefix to filter against, but since we will create our completion style we could ignore the prefix provided by completion framework and filter against the properly calculated prefix by ourselves which is calculated before the filtering has started. So we could return `(point) (point)` for bounds and then do the right thing when performing filtering.

lsp-mode.el
yyoncho: leave it as it is but we should definitely revisit the logic and investigate what vscode is doing in this case.

lsp-mode.el
kiennq: I see, since we've provided completion table as lambda prior to this PR, that function return value will be used as it without intervene of completion style functions. We have to call `completion-do-completions` by ourselves.

With that being said, we still need some reasonable  bounds when `textEdit` is not available (like `rls`)

yyoncho: I am not sure I understand 100% your comment:

Prior to this PR we are using this to perform the filtering:

```
(add-to-list 'completion-category-defaults '(lsp-substring (styles . (substring))))
```

We could create our filtering styles, e.g. lsp-fuzzy/lsp-substring/lsp-prefix.

As a side note, if you remove that line from lsp-mode company will display the items without any filtering, so we could do the filtering in the lambda itself and bypass the completion styles. 


> With that being said, we still need some reasonable bounds when textEdit is not available (like rls)

For these we will use `(thing-at-point 'symbol)` but this could be determined after the results are received. Do you mean that you could pass it as bounds by default and then use it only if you don't have textEdit? If yes, this seems to be ok since it leads to the same behaviour.

lsp-mode.el

lsp-mode.el
```diff
@@ -3702,77 +3702,101 @@ and the position respectively."
...
+               (setf start-point (point))
+               (-let* ((resp (lsp-request-while-no-input "textDocument/completion"
+                                                         (plist-put (lsp--text-document-position-params)
+                                                                    :context (ht ("triggerKind" 1)))))
+                       (items (lsp--sort-completions (cond
+                                                      ((seqp resp) resp)
+                                                      ((hash-table-p resp) (gethash "items" resp)))))
+                       ((start . end) (or
+                                       (-when-let* ((item (lsp-elt items 0))
+                                                    ((&hash "textEdit" text-edit) item)
```
yyoncho: Nit: you could simplify using &hash?

lsp-mode.el
```diff
@@ -3702,77 +3702,101 @@ and the position respectively."
...
+                                                           it]
+                                                      (propertize (or filter-text label)
+                                                                  'lsp-completion-item it))
+                                                    items)
+                                             'list)
+                                   (completion-all-completions probe it pred (length probe))
+                                   (let ((tail (last it)))
+                                     (if (consp tail) (setcdr tail nil))
+                                     it)
+                                   (--map (-let* ((item (get-text-property 0 'lsp-completion-item it))
```
yyoncho: Can you avoid using nested anamorphic forms(replace --map with -map)? Then make the code a bit harder to read since it changes it meanings in very narrow block of code.

lsp-mode.el
```diff
@@ -3702,77 +3702,101 @@ and the position respectively."
...
+                                                    ((&hash "textEdit" text-edit) item)
+                                                    ((&hash "range") text-edit))
+                                         (lsp--range-to-region range))
+                                       bounds))
+                       (probe
+                        (if (and start end) (buffer-substring-no-properties start end)
+                          "")))
+                 (setf done? (or (seqp resp)
+                                 (not (gethash "isIncomplete" resp)))
+                       result (--> (seq-into (--map (-let [(&hash "label"
```
yyoncho: possible to move the inner --map in the --> and reduce the nesting here (e. g. start with `(--> items ...`?

lsp-mode.el
```diff
@@ -3702,77 +3702,101 @@ and the position respectively."
...
+                                                                  "filterText" filter-text)
+                                                           it]
+                                                      (propertize (or filter-text label)
+                                                                  'lsp-completion-item it))
+                                                    items)
+                                             'list)
+                                   (completion-all-completions probe it pred (length probe))
+                                   (let ((tail (last it)))
```
yyoncho: The let block looks a bit odd, can you drop a comment?

kiennq:
> Can you elaborate on that?

Since we're passing lambda as programmed completion table, the `completion-all-completions` (`company-capf` is using that to get candidates) will pass the result of lambda as it without any client side filtering.
So in this change, I'm applying `completion-all-completions` to the list of computed candidates, that will apply client filtering with respect to `completion-styles` to the it.

Now since we're getting the candidates from language server, sometimes the language server has already been filtered candidate, i.e: return only the candidates that has prefix of the current context, however some language server that does not do filtering (e.x: json language server), at that time we will have additional client filtering.

For example with following json
``` jsonc
{
    "$schema": "http://json.schemastore.org/bowerrc",
    "sct|
}
```

The language server will return all possible candidates regardless of `sct` prefix. Now with our additional filtering, it can be fuzzily filtered
image
(The visual is off, but I don't know how to simply fix that now)

lsp-mode.el

lsp-mode.el

lsp-mode.el

lsp-mode.el

yyoncho:
Thank you!

> (The visual is off, but I don't know how to simply fix that now)

We should pass :company-match  (see also company-lsp).

PR #1292, opened 2020-01-04
New VHDL language server

chrbirks:
Added support for new VHDL language server.

### DIFF
--- a/lsp-vhdl.el
+++ b/lsp-vhdl.el
@@ -43,14 +43,9 @@
 (defvar vhdl-tool-bin-name "vhdl-tool"
   "Name of the VHDL Tool binary.")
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Looks good! As a side note, lsp-mode allows running several language servers at the same time(but you need separate registrations and mark it as `:add-on?`).

chrbirks:
I didn't know about `:add-on`. How would that work with servers with identical reported errors? Or for finding definitions when multiple servers are active?

yyoncho:
Identical errors will be duplicated. For definitions, you would be asked to select from the list of target locations since there is no deduplication on the client-side for that particular usecase. In the cases I tested one of the servers was complementary - e. g. typescript-language-server and eslint language server.

PR #1295, opened 2020-01-05
lsp-json: update lsp-json-server to new binary name

lxbarbosa:
Hi,

Json server upstream project has renamed its binary to `json-languageserver` and lsp-mode wont find it.

Snippet from https://github.com/vscode-langservers/vscode-json-languageserver

> Start the language server with the **json-language-server** command. Use a command line argument to specify the prefered communication channel:

This PR fixes it!

`

Thanks!

### DIFF
--- a/lsp-json.el
+++ b/lsp-json.el
@@ -65,7 +65,7 @@
    ("http.proxy" lsp-http-proxy)
    ("http.proxyStrictSSL" lsp-http-proxyStrictSSL)))
 
-(defcustom lsp-json-server "vscode-json-languageserver"
+(defcustom lsp-json-server "json-languageserver"
   "Json language server executable."
   :type 'string
   :group 'lsp-json


yyoncho:

kiennq:
@yyoncho, @lxbarbosa I think the snippet from https://github.com/vscode-langservers/vscode-json-languageserver is not that accurate.
Actually I noticed this before when I was working on `lsp-json`. The command mentioned in the link `npm install -g json-language-server` is not working currently, because there's no `json-language-server` package.
The closest one is `vscode-json-languageserver` that we're using now to `lsp-json` https://www.npmjs.com/search?q=json-language-server.

With that, the command `json-languageserver` is not existed.

Can you double check again to see if we need to revert this?

lxbarbosa:
@kiennq  Indeed, to install you still need `vscode-json-languageserver`, but the binary is installed as `json-languageserver`

yyoncho:
@lxbarbosa hm, on my system with clean installation it is named vscode-json-languageserver. Is it possible that you have old version of the server?

kiennq:
The package manifest also says the name is `vscode-json-languageserver` https://github.com/vscode-langservers/vscode-json-languageserver/blob/master/package.json

lxbarbosa:
@yyoncho @kiennq 

Oh just found why, there are more than one json-language-server extracted from vscode:

https://github.com/vscode-langservers/vscode-json-languageserver-bin
and
https://github.com/vscode-langservers/vscode-json-languageserver

its a subtle difference but yep, different servers.

maybe lsp-mode could search for any of those servers.



Its a Debian fresh install, can't say why it did not worked

kiennq:
They shared the same code base for the server (even same github repo). However, the former one is much older (released 2 years ago https://www.npmjs.com/package/vscode-json-languageserver-bin).
For now I think we should use the newer one.

Also, currently, `lsp-mode` readme is recommending to use `vscode-json-languageserver`, so users without vscode may not be able to find the language server with this change.
I think we should revert this.

Please let me know what you think.

lxbarbosa:
My bad, a revert is indeed needed!

PR #1299, opened 2020-01-06
Correct npm installing commands and path for Windows

kiennq:
According to https://docs.npmjs.com/files/folders#executables, we should install using global opyion with prefix and correct the path to have support for Windows

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5982,10 +5982,14 @@ Check `*lsp-install*' and `*lsp-log*' buffer."
 
 ;; npm handling
 
+;; https://docs.npmjs.com/files/folders#executables
 (cl-defun lsp--npm-dependency-path (&key package path &allow-other-keys)
-  (let ((path (f-join lsp-server-install-dir "npm" package "node_modules" path)))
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -5982,17 +5982,22 @@ Check `*lsp-install*' and `*lsp-log*' buffer."
...
+;; https://docs.npmjs.com/files/folders#executables
 (cl-defun lsp--npm-dependency-path (&key package path &allow-other-keys)
-  (let ((path (f-join lsp-server-install-dir "npm" package "node_modules" path)))
+  (let ((path (f-join lsp-server-install-dir "npm" package
+                      (cond ((eq system-type 'windows-nt) "")
+                            (t "bin"))
+                      path)))
     (unless (f-exists? path)
-      (error "The package %s is not installed. Unable to find %s." package path))
+      (error "The package %s is not installed.  Unable to find %s" package path))
```
yyoncho: Do we need double space here?

kiennq: Yes, `checkdoc` suggest that.

PR #1302, opened 2020-01-07
Fix ts-ls with correct tsServerPath

kiennq:
The `initialization-options` of `ts-ls` is doing nothing with passing `tsServerPath`.
From https://github.com/theia-ide/typescript-language-server, it seems that we actually need to pass the path via command line options.

Also, fix `lsp--npm-dependency-path` to return full executable name, else the `ts-ls` will crash in Windows.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -237,8 +237,10 @@ directory containing the package. Example:
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
-                                                          (cons (lsp-package-path 'typescript-language-server)
-                                                                lsp-clients-typescript-server-args)))
+                                                          `(,(lsp-package-path 'typescript-language-server)
+                                                            "--tsserver-path"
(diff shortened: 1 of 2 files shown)


lsp-clients.el
```diff
@@ -237,8 +237,10 @@ directory containing the package. Example:
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
-                                                          (cons (lsp-package-path 'typescript-language-server)
-                                                                lsp-clients-typescript-server-args)))
+                                                          `(,(lsp-package-path 'typescript-language-server)
+                                                            "--tsserver-path"
+                                                            ,(lsp-package-path 'typescript)
+                                                            ,@lsp-clients-typescript-server-args)))
```
yyoncho: Do we need to delete line 250?

kiennq: Maybe, I'm not sure, I would just keep both of them.

PR #1307, opened 2020-01-07
Fix  $schema insertion in lsp-json

kiennq:
In `lsp-json` when we try to select candidate and insert `$schema` into json file, with lsp-capf, `\$schema` will be insert.
This is due to `start-point` is not correctly calculated (it can be changed when lambda block is invoked so it's may not correct).
We may need to use text edit range as `start-point` and fallback to `(point)` if needed.

Since we will reverting text in `textEdit` range anyway, this change is actually safe and works better for snippet expanding.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3727,23 +3727,24 @@ and the position respectively."
           ((eq (car-safe action) 'boundaries) nil)
           (t (if done?
                  result
-               (setf start-point (point)
-                     prefix-line (buffer-substring-no-properties (point-at-bol) start-point))
-               (let* ((resp (lsp-request-while-no-input "textDocument/completion"
-                                                        (plist-put (lsp--text-document-position-params)
-                                                                   :context (ht ("triggerKind" 1)))))
-                      (items (lsp--sort-completions (cond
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -3727,23 +3727,24 @@ and the position respectively."
...
-                           (buffer-substring-no-properties start end)
-                         "")))
-                 (setf done? (or (seqp resp)
+               (-let* ((resp (lsp-request-while-no-input "textDocument/completion"
+                                                         (plist-put (lsp--text-document-position-params)
+                                                                    :context (ht ("triggerKind" 1)))))
+                       (items (lsp--sort-completions (cond
+                                                      ((seqp resp) resp)
+                                                      ((hash-table-p resp) (gethash "items" resp)))))
+                       ((start . end)
```
yyoncho: Not part of this review but here we should ignore end and use (point). E. g.  you may be completing at | and have a line `foo| bar`. In this case, the prefix is not (begin . end) but (begin . point).

kiennq: it makes sense

PR #1309, opened 2020-01-08
Fix #641 Set shutdown action properly

### LINKED ISSUE #641: lsp-mode prompts if pyls should be restarted on emacs exit
**Describe the bug**
When editing a python buffer using lsp-mode with the pyls language server, emacs will prompt to restart pyls on exit.

**To Reproduce**
1. Open a buffer python buffer (`lsp` is called through mode hook)
2. Exit emacs using `C-x C-c`
3. Emacs will show a prompt in the minibuffer: `Server pyls:xxxx exited with status signal. Do you want to restart it? (y or n)`

**Expected behavior**
Emacs should exit and kill any active language servers without any prompts.

pyls

**OS**
Fedora 29

wyuenho:
The `lsp--workspace-shutdown-action` in the `kill-emacs-hook` has never been set properly so the expected result when quitting emacs should show the workspace restart prompt by default. The fact that `lsp-mode` has been working so long out of the box without showing the prompt when quitting Emacs is more mysterious than the cause of this issue.

See: https://github.com/emacs-lsp/lsp-mode/issues/641#issuecomment-495744002

P.S. Please improve your code style. Reasonable people have learned not to use global variables since the 70s. Please refactor all these `lsp-foreach-workspace` `with-lsp-workspace` and `lsp--cur-workspace` etc etc nonsense. You don't need macros that write to some global variables to pass variables along to functions. Not knowing what writes to globals makes debugging super painful.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2560,7 +2560,7 @@ If NO-MERGE is non-nil, don't merge the results but return alist workspace->resu
   "Unload working workspaces."
   (lsp-foreach-workspace (lsp--shutdown-workspace)))
 
-(defun lsp--shutdown-workspace ()
+(defun lsp--shutdown-workspace (&optional restart)
   "Shut down the language server process for ‘lsp--cur-workspace’."
   (with-demoted-errors "LSP error: %S"
     (let ((lsp-response-timeout 0.5))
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

PR #1312, opened 2020-01-10
use (point) when calculating prefix-line

yyoncho:
- it is used to restore the original state of the buffer and it is not affected
by the start-point


kiennq:
LGTM

PR #1313, opened 2020-01-10
Improve edge case library folder detection.

### LINKED ISSUE #1217: Read only mode
Not sure if this is the intended behavior, but I noticed that ever since installing this, all my buffers have been opening in read-only mode. Is there a way to disable that?

muirdm:
Don't open a file as a library file if it is within a project root.
This makes things work better if your project root is within a library
folder for some reason.

I also reverted d8e99484ab9bbb211e00f09aff1088ef151f1653 and restored
the Go library paths as just '("/usr"). This fixes standard library
detection on Mac. I don't think we should adjust the default config to
handle the rare case of a user putting their home directory under
/usr. In this case, the user should just customize
lsp-clients-go-library-directories to accommodate their setup.

I also added an info message when opening a library file in read only
mode so it is easy to figure out why your file is in read only mode.
Note that this message is superseded in the minibuffer by the "Connected
to" message, but at least it shows up in *Messages*.

Fixes #1217.

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -147,7 +147,7 @@ defaults to half of your CPU cores."
   :type 'boolean
   :group 'lsp-clients-go)
 
-(defcustom lsp-clients-go-library-directories '("/usr/lib" "usr/local/lib")
+(defcustom lsp-clients-go-library-directories '("/usr")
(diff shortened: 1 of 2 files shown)


yyoncho:
Thank you!

PR #1317, opened 2020-01-14
Add mypy plugin configuration in lsp-pyls

itome:
See https://github.com/tomv564/pyls-mypy#configuration

### DIFF
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -347,6 +347,21 @@ parameters referenced in config."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-pyls-plugins-mypy-enabled t
+  "Enable or disable the plugin."
+  :type 'boolean
+  :group 'lsp-pyls
+  :package-version '(lsp-mode . "6.3"))
+
+(defcustom lsp-pyls-plugins-mypy-live-mode-enabled nil
+  "Enable or disable the mypy live-mode. live_mode provides type checking as you type,
+but have some limitation, 1) Imports cannot be followed correctly.
+2) Stub files are not validated correctly. Turning off live_mode means you must save
+your changes for mypy diagnostics to update correctly."
+  :type 'boolean
+  :group 'lsp-pyls
+  :package-version '(lsp-mode . "6.3"))
+
 (lsp-register-custom-settings
  '(("pyls.rope.ropeFolder" lsp-pyls-rope-rope-folder)
    ("pyls.rope.extensionModules" lsp-pyls-rope-extension-modules)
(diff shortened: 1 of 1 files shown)


yyoncho:
Looks good to me!

@mpanarin willing to take a look?

mpanarin:
I don't think this is a good idea.

This is a 3rd party plug-in which is obviously not guaranteed to be installed with pyls (although a lot of `pyls` users probably do use it)
1. This defaults are not very good as making it enabled by default implies that it is installed, which can lead to errors in pyls when it is not.
2. pyls is in progress of implementing its own mypy support (although slow) (https://github.com/palantir/python-language-server/pull/392) so author efforts should probably be directed there.
3. When the plugin is installed with pyls it is enabled by default, as well as `live_mode` - no reason to set it from IDE.

But this rises a different question. @yyoncho is there a way to add parameters to ls dynamically? 
As in this case, imagine there would be a reason to enable/disable a feature in the pyls plugin. Maybe some kind of functionality to add parameters like this through dir-locals would be worth it?

yyoncho:
> @yyoncho is there a way to add parameters to ls dynamically?

Can you elaborate?

> Maybe some kind of functionality to add parameters like this through dir-locals would be worth it?

You could have dir local with `lsp-client-settings`. The properties that a server needs are determined by the first token of the path (so-called section).

itome:
Take the situation into consideration, I change my mind and think this change confuse users. I'll close this PR if the discussion is finished.

yyoncho:
> When the plugin is installed with pyls it is enabled by default, as well as live_mode - no reason to set it from IDE.

Just want to mention that  if the plugin is present these properties could be useful to disable it per project. In general it is up to you(the pyls users) to decide what make sense.

mpanarin:
> Just want to mention that if the plugin is present these properties could be useful to disable it per project. In general it is up to you(the pyls users) to decide what make sense.
You have to install pyls with all of its plugins in each environment anyway, so I don't think it is really usefull here.
> You could have dir local with `lsp-client-settings`. The properties that a server needs are determined by the first token of the path (so-called section).
That is a good idea, thanks!
Although, it can be a bit confusing as it requires adding values to the list in dir-locals instead of settings some variable. Isn't it unsafe?
Maybe it is better to a variable like `lsp-ls-additional-params` which are passed to language server additionaly to everything else?
This way in dir-locals we would simply set a var and that's it

yyoncho:
> Although, it can be a bit confusing as it requires adding values to the list in dir-locals instead of settings some variable. Isn't it unsafe?

You could set variables as well. The list is useful if you want to alter the set of variables to using dir locals.

mpanarin:
> You could set variables as well. The list is useful if you want to alter the set of variables to using dir locals.

I understand that. I mean that list manipulations are unsafe in dir-locals, while setting vars is not. And it is a bit more strightforward.

yyoncho:
@mpanarin can you merge/reject this PR as the owner of lsp-pyls.el?

yyoncho:
@mpanarin thank you.

ztlevi:
By seting `lsp-clients-settings`, I found pyls-mypy enabled is working but live_mode is not working. A better workaround is create a `.mypy.ini` config file in HOME or `mypy.ini` in your repo.

Hi-Angel:
To have a bit of documentation here: after having `pyls-mypy` installed, you need to change parameters [that are decribed in pyls-mypy README](https://github.com/tomv564/pyls-mypy/blob/master/README.rst#configuration). For `emacs-lsp` you can do this like:

```lisp
(add-to-list 'lsp-client-settings '("pyls.plugins.pyls_mypy.enabled" t))
(add-to-list 'lsp-client-settings '("pyls.plugins.pyls_mypy.live_mode" nil))
```

Issue #1305, opened 2020-01-07
company-capf endgame


yyoncho:
We miss the following features:

1. Caching - we should cache and filter right after `isIncomplete` is nil
2. Fuzzy/Partial matching - now we match only exact string match. There are categories in emacs completion framework and they have flex matching in emacs 27. We should decide whether to reuce the emacs builtin completion style facilities or create ad-hoc solution. 
3. Fix gui to display matches (see :company-match in company-capf docs).

yyoncho:
@kiennq

kiennq:
> 2\. Fuzzy/Partial matching - now we match only exact string match. There are categories in emacs completion framework and they have flex matching in emacs 27. We should decide whether to reuce the emacs builtin completion style facilities or create ad-hoc solution.

I think we are kind of done for this already. We've already using `completion-all-completion` to filter candidates based on `completion-styles`.
Once we have caching and reuse cache, we will have some degree of fuzzy/partial matching regardless of language server.
For now only language server that return all candidates (like `lsp-json`) can have fuzzy/partial matching.

yyoncho:
Ah, now I got it what do you mean - your implementation is a great way to handle the fact that we should filter based on filterText!

I am not sure we are supposed to use `completion-styles` directly but we should define a completion category and then configure that category with the corresponding style. The completion-styles are global and even if we change it buffer-locally it will probably affect other stuff. This could be easily fixed in the future though.

Let me know if you want to work on this issue.

kiennq:
> Let me know if you want to work on this issue.

Sure, let me work on this issue.

Although, as a side note https://www.gnu.org/software/emacs/manual/html_node/elisp/Completion-in-Buffers.html said that
>Additionally, the collection should generally not be pre-filtered based on the current text between start and end, because that is the responsibility of the caller of completion-at-point-functions to do that according to the completion styles it decides to use.

This actually only true for collection that is not a function (like `elisp-completion-at-point`), in case we're using programmed completion with a function, we're responsible for filtering candidate inside the collection lambda.

yyoncho:
This is ready thanks to @kiennq  - `company-capf` is feature complete and now is in a testing period. After it is over we will update the docs to recommend it over company-lsp.

PR #1334, opened 2020-01-19
Cache completion candidates in lsp-capf

kiennq:
This finish fixing #1305.
Caching candidates from language server when language server said they're completed (`isIncomplete == nil`).

The cache is a list that contains `(cache-prefix cache-items additional-args)`.
It will be applied when `cache-prefix` is prefix of currently input string. And will be invalidated/updated immediately otherwise.

Pairing with `completion-styles` we can archive flex matching for `lsp-mode` completion at point function.
a2d69604-18a8-430e-a99d-cea0740e985f

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3705,6 +3705,39 @@ and the position respectively."
   "Return non-nil if text before point matches any of the trigger characters."
   (looking-back (regexp-opt (cl-coerce trigger-characters 'list)) (line-beginning-position)))
 
+(defvar lsp--capf-cache nil
+  "Cached candidates for completion at point function.
+In the form of (prefix items args).")
+
+(defun lsp--capf-filter-candidates (pred string items &rest plist)
+  "List the possible completion of STRING in candidates ITEMS.
+Only the elements that satisfy predicate PRED are considered.
+PLIST is the additional data to attach to each candidate."
+  (--> items
+       (-map (-lambda ((item &as &hash
+                             "label"
+                             "filterText" filter-text))
+               (propertize (or filter-text label) 'lsp-completion-item item))
+             it)
+       (seq-into it 'list)
+       (completion-all-completions string it pred (length string))
+       ;; completion-all-completions may return a list in form (a b . x)
+       ;; the last cdr is not important and need to be removed
+       (let ((tail (last it)))
+         (if (consp tail) (setcdr tail nil))
+         it)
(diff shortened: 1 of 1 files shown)


yyoncho:
Looks good to me, I will spend some time testing with different servers.

One question about cache cleanup which was causing issues in the past: how does it happen? Is it possible to end up with stale cache if I trigger completion on line N and then reuse the cache on line M which starts with the same prefix?

kiennq:
> One question about cache cleanup which was causing issues in the past: how does it happen? Is it possible to end up with stale cache if I trigger completion on line N and then reuse the cache on line M which starts with the same prefix?

It's possible the stale cache will happens. However, in that case, the prefix that we used to filter the candidates from the cache will be quite long and I doubt it can easily match any candidate inside the cache. We will probably not getting any cache in that case.
I can probably fallback to server response in that case. What do you think?

yyoncho:
> I can probably fallback to server response in that case. What do you think?

I do not like this idea because some of the language servers send a huge response(e. g. 3mb) and it will have a performance hit. I could try to do:

1. Cleanup on :exit-function
2. Opportunistically use `company-completion-finished-hook` and `company-completion-cancelled-hook`  if company is used.

Point 2 will be sufficient to handle the company-capf case. We could also look for a function emacs core which we could put advice on and perform the cleanup(e. g. `completion-at-point` ?!)

kiennq:
> Point 2 will be sufficient to handle the company-capf case. We could also look for a function emacs core which we could put advice on and perform the cleanup(e. g. `completion-at-point` ?!)

Ah, I see, it certainly helps.
Also I look at `completion-at-point` function and didn't see anything promising to advise against cancel action. Also I don't think `:exit-function` is called in that case either.

yyoncho:
> completion-at-point

We could add `:before` advice to cleanup the cache upfront instead of looking for cancel action.

kiennq:
> > completion-at-point
> 
> We could add `:before` advice to cleanup the cache upfront instead of looking for cancel action.

That mean the cache will never be applied when `completion-at-point` is called, is that okay?
I never using `completion-at-point` function alone, so I would rely on you to decide that.

yyoncho:
> That mean the cache will never be applied when `completion-at-point` is called, is that okay?
> I never using `completion-at-point` function alone, so I would rely on you to decide that.

I think it is invoked one time and it cannot reuse cache either way.

kiennq:
> I think it is invoked one time and it cannot reuse cache either way.

Done

PR #1335, opened 2020-01-19
Add server locations, special-mode for stderr-buf, fix types

JohnC32:
1. Add defcustom lsp-server-locations which lets one specify server locations.
   With this, an organization can install language servers on a central filer
   e.g. npm install -g --prefix /central/location/lsp/SERVER SERVER,
   then Emacs can access them via NFS or Samba which makes Emacs work
   cleanly for all with the desired version language servers.

2. Add a cache for lsp--npm-dependency-path to support lsp-server-locations
   and avoid unnecessary stats after the location to a server has been found.

3. Use special-mode for stderr-buf which makes it readonly, lets one type
   'q' to bury the buffer, etc.

4. Fix a couple typos. Note, one of the typo fixes is in code. The two
   functions
     lsp--cleanup-highlights-if-needed
     lsp--document-highlight
   had typo ":highlihts" which should be ":highlights"
   I suspect this fixes a bug, but wasn't sure what these functions are for.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6074,10 +6074,10 @@ nil."
   ;; child process spawn command that is invoked by the
   ;; typescript-language-server). This is why we check for existence and not
   ;; that the path is executable.
-  (or (and (f-absolute? path)
-           (f-exists? path)
-           path)
-      (executable-find path)))
+  (if (and (f-absolute? path)
+           (f-exists? path))
+      path
+    (executable-find path))
 
 (defun lsp-package-path (dependency)
   "Path to the DEPENDENCY each of the registered providers."
@@ -6103,25 +6103,17 @@ nil."
 
 ;; npm handling
 
-(defvar lsp--npm-dependency-path-cache nil
-  "Cache of server packages to the executable paths")
-
 ;; https://docs.npmjs.com/files/folders#executables
-(cl-defun lsp--npm-dependency-path (&key package exe-name &allow-other-keys)
-  "Return npm dependency executable path for EXE-NAME for node server PACKAGE."
-  ;; Did we already load exe-path? If so use it, otherwise see if it has been installed into
-  ;; `lsp-server-install-dir'.
-  (let ((exe-path (gethash package lsp--npm-dependency-path-cache)))
-    (when (not exe-path)
-      (setq exe-path (executable-find
-                      (f-join lsp-server-install-dir "npm" package
-                              (cond ((eq system-type 'windows-nt) "")
-                                    (t "bin"))
-                              exe-name))))
-    (if (and exe-path (f-exists? exe-path))
-        (puthash package exe-path lsp--npm-dependency-path-cache)
-      (error "The package %s is not installed.  Unable to find %s" package exe-name))
-    exe-path))
+(cl-defun lsp--npm-dependency-path (&key package path &allow-other-keys)
+  "Return npm dependency PATH for PACKAGE."
+  (let ((path (executable-find
+               (f-join lsp-server-install-dir "npm" package
+                       (cond ((eq system-type 'windows-nt) "")
+                             (t "bin"))
+                       path))))
+    (unless (and path (f-exists? path))
+      (error "The package %s is not installed.  Unable to find %s" package path))
+    path))
 
 (cl-defun lsp--npm-dependency-download  (callback error-callback &key package &allow-other-keys)
   (if-let (npm-binary (executable-find "npm"))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5943,27 +5943,6 @@ SESSION is the active session."
   :type 'directory
   :package-version '(lsp-mode . "6.3"))
 
-(defcustom lsp-server-locations nil
-  "List of server PACKAGE / LOCATION pairs of pre-installed
-server package names and their corresponding executable path
-locations.  The calls to `lsp-dependency' define the PACKAGE name
-key you can use and the corresponding string value should
-location of the executable. For example, if you install the 
-Typescript/JavaScript language servers in a team shared read-only
-Unix location, you can instruct lsp-mode to use the by adding
-the adding two pairs:
-
- key                         string
- ---                         ---------
- typescript-language-server  /net/lsp/typescript-language-server/bin/typescript-language-server
- typescript                  /net/lsp/typescript//lib/node_modules/typescript/lib/tsserver.js
-
-Note, the typescript compiler is invoked by the typescript-language-server and thus we pass
-the path to the tsserver.js file to ensure it works when spawned on Windows."
-  :risky t
-  :type '(alist :value-type (group string))
-  :package-version '(lsp-mode . "6.3"))
-
 (defvar lsp--dependencies (ht))
 
 (defmacro lsp-dependency (name &rest definitions)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6077,7 +6077,7 @@ nil."
   (if (and (f-absolute? path)
            (f-exists? path))
       path
-    (executable-find path))
+    (executable-find path)))
 
 (defun lsp-package-path (dependency)
   "Path to the DEPENDENCY each of the registered providers."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1408,8 +1408,8 @@ already have been created."
      ,@body
      (lsp--flush-delayed-changes)))
 
-(defun lsp-cannonical-file-name  (file-name)
-  "Return the cannonical FILE-NAME."
+(defun lsp-canonical-file-name  (file-name)
+  "Return the canonical FILE-NAME."
   (f-canonical (directory-file-name (f-expand file-name))))
 
 (defun lsp--window-show-message (_workspace params)
@@ -2800,7 +2800,7 @@ in that particular folder."
   (interactive
    (list (read-directory-name "Select folder to add: "
                               (or (lsp--suggest-project-root) default-directory) nil t)))
-  (cl-pushnew (lsp-cannonical-file-name project-root)
+  (cl-pushnew (lsp-canonical-file-name project-root)
               (lsp-session-folders (lsp-session)) :test 'equal)
   (lsp--persist-session (lsp-session))
 
@@ -2812,7 +2812,7 @@ in that particular folder."
                                       (lsp-session-folders (lsp-session)) nil t
                                       (lsp-find-session-folder (lsp-session) default-directory))))
 
-  (setq project-root (lsp-cannonical-file-name project-root))
+  (setq project-root (lsp-canonical-file-name project-root))
 
   ;; send remove folder to each multiroot workspace associated with the folder
   (dolist (wks (->> (lsp-session)
@@ -3599,7 +3599,7 @@ Applies on type formatting."
 (defun lsp-workspace-root (&optional path)
   "Find the workspace root for the current file or PATH."
   (-when-let* ((file-name (or path (buffer-file-name)))
-               (file-name (lsp-cannonical-file-name file-name)))
+               (file-name (lsp-canonical-file-name file-name)))
     (->> (lsp-session)
          (lsp-session-folders)
          (--first (and (lsp--files-same-host it file-name)
@@ -3835,7 +3835,7 @@ and the position respectively."
 
 (defun lsp--to-yasnippet-snippet (text)
   "Convert VS code snippet TEXT to yasnippet snippet."
-  ;; VS code snippet doesn't esscape "{", but yasnippet requires escaping it.
+  ;; VS code snippet doesn't escape "{", but yasnippet requires escaping it.
   (let (parts
         (start 0))
     (dolist (range (s-matched-positions-all (regexp-quote "{") text))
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -5939,6 +5942,27 @@ SESSION is the active session."
   :type 'directory
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-server-locations nil
```
kiennq: This's only used for `npm` backend, perhaps you want to change it into another name that reflect that?
Or this can be used by other backends too?

kiennq: Also what's the advantage of adding a new custom variable instead of just calling `lsp-dependency` with new values for `npm` backend?
User doesn't need one more customizable variable and we don't need to change big chunks of code.

In that case you can probably make `lsp-dependency` autoload.

JohnC32: It would be nice to have a way to eliminate the customize I proposed. I couldn't see how to inject the location of the language server without adding a defcustom. How would I go about doing this with a call to lsp-dependency? Alternatively, if you'd like I can try making the defcusom more general to handle all types of language servers, though would need some pointers.

In our environment we configure Emacs for a large number of programmers. They get all of the benefits of lsp, etc. without having to set anything up. The add-on packages are placed on NFS / Samba for Linux, Mac, Windows users. This is a significant time saver, especially when many people are not comfortable with lisp.

kiennq: I see, taking a closer look at the code again, how about changing `lsp-server-install-dir` to the shared storage path?
You can probably put that as local-dirs variable inside project folder so everyone who clone the project can benefit from that.

JohnC32:
Yes, that would work - thanks. How about we keep the code involving lsp--npm-dependency-path-cache to avoid extra stat's of the file system when new files are opened?

Would you know anything about the "highlihts" typo? I was trying to test this and think neither the incorrect or the correct spelling is right. If I disable lsp-eldoc-enable-hover, then explicitly call lsp-document-highlight on a source code variable, I see all instances highlighted. Then I move to end of line and call lsp-document-highlight, it doesn't remove the highlight. In an older version of lsp mode from Oct, 951766b0, that we were using. This sequence works as expected in this older version of lsp-mode.

yyoncho:
> Would you know anything about the "highlihts" typo? I was trying to test this and think neither the incorrect or the correct spelling is right. If I disable lsp-eldoc-enable-hover, then explicitly call lsp-document-highlight on a source code variable, I see all instances highlighted.

It should work either way(as long as we use  the same "token"). 

> Then I move to end of line and call lsp-document-highlight, it doesn't remove the highlight. In an older version of lsp mode from Oct, 951766b, that we were using. This sequence works as expected in this older version of lsp-mode.

This is probably a regression - we should add `(lsp--remove-overlays 'lsp-highlight)` in `lsp-document-highlight`

yyoncho:
> Would you know anything about the "highlihts" typo? I was trying to test this and think neither the incorrect or the correct spelling is right. If I disable lsp-eldoc-enable-hover, then explicitly call lsp-document-highlight on a source code variable, I see all instances highlighted.

It should work either way(as long as we use  the same "token"). 

> Then I move to end of line and call lsp-document-highlight, it doesn't remove the highlight. In an older version of lsp mode from Oct, 951766b, that we were using. This sequence works as expected in this older version of lsp-mode.

This is probably a regression - we should add `(lsp--remove-overlays 'lsp-highlight)` in `lsp-document-highlight`

JohnC32:
Thanks - adding `(lsp--remove-overlays 'lsp-highlight)` works.

kiennq:
LGTM now

lsp-mode.el
```diff
@@ -5592,6 +5593,9 @@ returned by COMMAND is available via `executable-find'"
                                    :noquery t)))
                        (set-process-query-on-exit-flag proc nil)
                        (set-process-query-on-exit-flag (get-buffer-process stderr-buf) nil)
+                       (with-current-buffer (get-buffer stderr-buf)
+                         ;; Make the *NAME::stderr* buffer buffer-read-only, q to bury, etc.
+                         (special-mode))
```
kiennq: We should check the `major-mode` before activating new `major-mode`. This is done so the initialization logic inside major-mode activation function will not run too much, which can affect perf.
Something along `(unless (eq 'special-mode major-mode) (special-mode))`.

lsp-mode.el
```diff
@@ -6065,17 +6069,25 @@ When UPDATE? is t force installation even if the server is present."
 
 ;; npm handling
 
+(defvar lsp--npm-dependency-path-cache nil
```
yyoncho: is this cache really needed?

lsp-mode.el
JohnC32: Since we are creating the `stderr-buf` we know it won't be in special-mode, thus the condition will always be true, so I don't believe adding an always true condition is needed (or am I missing something an there's a way to have make-process create the buffer in special-mode?).

lsp-mode.el
JohnC32: Strictly speaking the cache isn't needed, when HOME is a local file system the redundant stat's for newly opened files aren't noticeable. In our environment we map HOME (~) for Linux, Windows, Mac to the same NFS / CIFS filer location. stat's against these can stall and do in general take much longer than a local file system. So, it does help in this scenario.

Customizing `lsp-server-install-dir` to a shared location didn't end up working for us. The reason is we want the "organization standard" languages servers in a shared read-only network location. People still need to be able to download non-organization standard servers for languages we haven't baked into our default emacs setup. The solution I took was to use `(puthash "server-name" "/net/location" lsp--npm-dependency-path-cache)` which lets us customize lsp without having to maintain our tweaks. This probably isn't the best way to state where "organization standard" language servers come from without having to modify `lsp-server-install-dir`. If you have something I could tryout, that would be nice.

Thanks

yyoncho: Got it. But still, I do not think that the caching is part of npm provider. Imagine that we add several installers(yarn, pip, composert, etc). Each one of them will need caching for that particular use-case. The caching is a better fit for the `:system` provider.  But before jumping into implementing a cache for the system, I think that the following will do the job for you:

``` elisp
(with-eval-after-load 'lsp-clients
  (lsp-dependency typescript-language-server
    (:system "/home/kyoncho/.emacs.d/.cache/lsp/npm/typescript-language-server/bin/typescript-language-server")))
``` 

Note also the you could create providers optimized for your organization and register them in `lsp-deps-providers` and then do `(lsp-dependency ...`. (We could introduce a method `lsp-dependency-add-provider` to make it more consistent. 

Let me know what do you think.

yyoncho: And one more thing that you may consider for your organization - lsp-mode supports running the language server in docker container over a local source tree. This may yield better performance and might be more convenient to manage the dependencies - e. g. everyone will have a consistent experience.

JohnC32: I tried modifying `lsp--dependencies` via

            (ht-set lsp--dependencies 'typescript-language-server
                    (list `(:system ,tls-exe)))
            (ht-set lsp--dependencies 'typescript
                    (list `(:system ,ts-exe))))

and validated the `lsp--dependencies` paths for :system are correct. However, this doesn't work? This is stretching my knowledge of lisp. I'm not familiar with macros, ht, etc. thus having a little difficulty following the flow of the code.

Note, I can't call `lsp-dependency` because the paths to the executables are dynamically computed and are in a variable. Calling `lsp-dependency` doesn't expand the variable.

Nice to know lsp supports docker containers, but I don't think we need to use a docker container in our environment because we manage the locations of the servers, etc. and rely on central filers to update them on a regular clip. However, I need to investigate more to see if the containers are a better approach.

yyoncho: > However, this doesn't work? 

:system uses executable-find which wont work for .js file I think. You may do `M-x trace-function RET executable-find RET` and track down the invocations of executable-find. 

> Note, I can't call lsp-dependency because the paths to the executables are dynamically computed and are in a variable. Calling lsp-dependency doesn't expand the variable.

We might replace the macro here with a function since it looks like it is causing more harm than good.

lsp-mode.el
kiennq: I see, I though that buffer can be set to `(special-mode)` multiple times.

lsp-mode.el
JohnC32: Yes, `executable-find` when called from `lsp-package-path`, doesn't work on a *.js file, yet the tsserver requires a js file on Windows. This is because the typescript compiler ,tsserver, is invoked by the typescript-language-server and thus we pass the path to  tsserver.js to ensure typescript-language-server is able to spawn the tsserver. On UNIX, you can pass a path to a bash script that runs the tsserver and this works. However, on Windows if you pass a CMD script that runs the tsserver compiler, the typescript-language-server fails when spawning. The language server is called with the tsserver path and when passed a CMD file, when the typescript-language-server calls `cp.spawn(tsserverPath, args)`, it fails. If you pass the path to the tsserver.js things work on all platforms.

It seems to me that if I'm providing the location of the necessary items, that lsp shouldn't be doing `executable-find`, etc. It should just use what was specified.

Would it make sense to add a `defcustom lsp-package-locations` that lets one map package dependencies to absolute locations and this could be used to configure all language servers? This `defcustom` would be used by `lsp-package-path`. If this idea makes sense, I can try it out.

Another alternative is that I just put advice on `lsp-package-path`, but this is somewhat working around the missing capability to specify an absolute path to the packages, which to me seems like a useful feature for others.

yyoncho: > It seems to me that if I'm providing the location of the necessary items, that lsp shouldn't be doing executable-find, etc. It should just use what was specified.

Haven't tested that but something like that should work:

```
(defvar lsp-deps-providers
  (list :npm (list :path #'lsp--npm-dependency-path
                   :install #'lsp--npm-dependency-download)
        :system (list :path #'lsp--system-path)))

(defun lsp--system-path (path)
  (or (executable-find path)
      (and (f-absolute? path)
           (f-exists? path)
           path)))
```
and the other change is to convert lsp-dependency from macro to function to allow params evaluation without doing tricks.

akirak: You probably can use this:

```emacs-lisp
(defun lsp--system-path (path)
  (or (and (file-executable-p path)
           path)
      (executable-find path)))
```

JohnC32: Thanks for the pointers. Sorry I've been a little slow in responding (been busy with other items). Hopefully, all is good now. I removed the cache code since it probably unique to our environment and I can manage this on our end.

Note, we can't use `executable-find` because some of the dependencies are not executables. For example the typescript compiler, tsserver.js is a JavaScript file. Passing the path to the tsserver.cmd file on Windows causes the typescript-language-server to fail. See comment above.

lsp-mode.el
```diff
@@ -6039,7 +6060,24 @@ When UPDATE? is t force installation even if the server is present."
...
+nil."
+  ;; For node.js 'sub-packages' PATH may point to a *.js file. Consider the
+  ;; typescript-language-server. When lsp invokes the server, lsp needs to
+  ;; supply the path to the typescript compiler, tsserver.js, as an argument. To
+  ;; make code platform independent, one must pass the absolute path to the
+  ;; tsserver.js file (Windows requires a *.js file - see help on the JavaScript
+  ;; child process spawn command that is invoked by the
+  ;; typescript-language-server). This is why we check for existence and not
+  ;; that the path is executable.
+  (or (and (f-absolute? path)
```
kiennq: nit: this can be rewrite in `if` term, which is easier to understand.

lsp-mode.el
kiennq: > However, on Windows if you pass a CMD script that runs the tsserver compiler, the typescript-language-server fails when spawning. The language server is called with the tsserver path and when passed a CMD file, when the typescript-language-server calls `cp.spawn(tsserverPath, args)`, it fails. If you pass the path to the tsserver.js things work on all platforms.

Out of curious, which version of `typescript-language-server` are you using? I'm using the latest one and it seems fine for me to pass cmd file path as parameter of `--tsserver-path`.
Maybe that's the limitation of `cmd` in Windows that is not run well with network path.

yyoncho: > I removed the cache code since it probably unique to our environment and I can manage this on our end.

AFAICS the latest version still has cache?

lsp-mode.el
yyoncho: I suggested that version and I usually favour the functional version over the imperative one, e. g. 
```
(or (foo) (bar) (foo-bar) )
```
over
```
(if-let (foo (foo))
    foo
  (if-let (bar (bar))
      bar
    (foo-bar)))
```

kiennq: I agree with your case of writing (using `if-let`), the `if-let` look overly verbose, however in this case, the functional version of 
``` elisp
(or (and (f-absolute? path)
         (f-exists? path)
         path)
    (executable-find path))
```
can be reduce to 
``` elisp
(if (and (f-absolute? path)
         (f-exists? path))
    path
  (executable-find path))
```
Which doesn't riddle me of shouldn't the last `path` in `and` clause being check before used in `f-absolute?`

yyoncho: I agree for that particular case - `if` statement looks fine too.

> Which doesn't riddle me of shouldn't the last path in and clause being check before used in f-absolute?

Can you elaborate? You mean like that 
```
(and path
         (f-absolute? path)
         (f-exists? path)
         path)
```

kiennq: I means when I look at
``` elisp
(and (f-absolute? path)
         (f-exists? path)
         path)
```
First thought come to me as: "Oh, they're checking for path nullness too". Then "wait why don't path nullness being checked before use?". Then I finally realize that they just want to return `path` here, since if path is `nil`, the another clause of `or` will throw too

lsp-mode.el
JohnC32: Now it's really removed. Thanks for the catch.

JohnC32: Regarding the tsserver-path issue. Here's the info on spawning child processes https://nodejs.org/api/child_process.html#child_process_spawning_bat_and_cmd_files_on_windows 

We have in typescript-language-server/lib/node_modules/typescript-language-server/lib/tsp-client.js, `cp.spawn(tsserverPath, args);` and it's not setup to invoke the cmd file. 

I'm using typescript-language-server version 0.4.0 with typescript compiler (tsserver) 3.7.4

lsp-mode.el
JohnC32: I updated the code to be 

```emacs-lisp
(if (and (f-absolute? path)
         (f-exists? path))
    path
  (executable-find path))
```
which to me is a little easier to parse because lisp isn't my native coding language, though the functional version is also fairly clean to me.

yyoncho:
@JohnC32 seems like you are missing a bracket somewhere in the code - https://travis-ci.org/emacs-lsp/lsp-mode/jobs/641024850#L1461

JohnC32:
Oops - that's embarrassing. I did test the change but I committed the wrong version of the file. The missing paren is no longer.

yyoncho:

PR #1336, opened 2020-01-20
Should not cache completion candidates when it's empty

kiennq:
This fix a corner case of PR #1334.
We should not cache completion candidates when it's empty, else the cache cleanup logic will not be run.


lsp-mode.el
```diff
@@ -3794,7 +3794,7 @@ PLIST is the additional data to attach to each candidate."
                   (prefix-line (buffer-substring-no-properties (point-at-bol) (point))))
             (setf done? (or (seqp resp)
                             (not (gethash "isIncomplete" resp)))
-                  lsp--capf-cache (when (and done? (car bounds))
+                  lsp--capf-cache (when (and done? (car bounds) (not (seq-empty-p items)))
```
yyoncho: Do we need `(car bounds)` here?

kiennq: Yes, just in case the bounds is `nil` or we cannot get the start point of current symbol.
In that case, we don't have the prefix to make sure that the cache is for current input and should not risk caching it.

yyoncho: Isn't prefix-line + start point + cache cleanup sufficient? Otherwise, we won't cache after trigger character and when triggering the completion with no symbol. AFAICS after  `(car bounds)` is put in the cache it is not used later.

kiennq: The `bounds` is used to calculate the prefix for caching. That may different from the prefix we used for filtering.
The reason is to get the prefix for filtering, we actually have to wait for server response (textEdit), whereas the bounds can be calculated directly from `thing-at-point`.
And we put prefix, not `(car bounds)` into cache.

yyoncho: Hm, why do we use `(buffer-substring-no-properties (car bounds) (point))` to calculate the prefix but not `(buffer-substring-no-properties start (point))` and later reuse that `start` point if the cache prefix-line is a prefix of current prefix-line? Am I missing something? Also, seems like we should default bounds to `(list (point) (point))` when symbol at point returns nil. This will also remove the need from checking for `(car bounds)` which will be guaranteed to be not nil.

kiennq: The prefix here is used just for cache checking purpose. When we do filtering, we're using cached `start-point`.
We can certainly use `prefix-line` instead of prefix for cache checking. But I doubt it would be perf hit when the line is long. So it's more like the longer we take, the slower completion become even with cache, which is not good.

Bound `bounds` to default value also good for me, will update that.

yyoncho:
For some reason, flex completion does not work with jdtls but it works with htmlls, ts-ls. I am investigating.

lsp-mode.el
yyoncho: > That means the cache is not working for jdtls, I can only imagine that they never send `isIncomplete` with `nil`

Not cache - I do not get any results when I do `(setq completion-styles '(flex))` but it works with default `completion-styles`

yyoncho:
> That means the cache is not working for jdtls, I can only imagine that they never send `isIncomplete` with `nil`

Not cache - I do not get any results when I do `(setq completion-styles '(flex))` but it works with default `completion-styles`

kiennq:
> Not cache - I do not get any results when I do `(setq completion-styles '(flex))` but it works with default `completion-styles`

What is the value of `lsp--capf-cache` at that time?
Without cache, the flex matching will not work.

yyoncho:
jdtls itself provide flex completion and also it works with company-lsp.

kiennq:
Is there any error in `*Message*`?
Can you try to dump out the value of `lsp--capf-cache`, `pretty-print` it and show me?

yyoncho:
I think it is not related to cache - I suspect that it filters the results two times - once in our implementation, once we return the result to completion facilities. I will investigate further and let you know.

kiennq:
> I think it is not related to cache - I suspect that it filters the results two times - once in our implementation, once we return the result to completion facilities. I will investigate further and let you know.

`company-capf` is using `completion-all-completions` to get all the completion candidates. Since we're passing a lambda as completion table, it will just show whatever that lambda returned as candidates.
Because of that, we have to filter result in our implementation with a list of completion items as completion table. In this case the filter will be properly applied.

yyoncho:
> Because of that, we have to filter result in our implementation with a list of completion items as completion table. In this case the filter will be properly applied.

This doesn't seem to be true:

I set the following advice on completion-all-completions:

```
(defun my/fn (orig string table pred point &optional metadata)
  (let ((result (funcall orig string table pred point metadata)))
    (message "input string=%s pred=%s point=%s length=%s" string pred point (condition-case err
                                                                                (length result)
                                                                              (error 1)))
    result))
(add-function :around (symbol-function 'completion-all-completions)
              'my/fn)
```

Then I saw in the messages buffer:

```
input string=De pred=nil point=2 length=1
input string=@De pred=nil point=3 length=1
```

Note that the second call is using the miscalculated prefix.

We could skip the second filtering by:

1. Define category for the source
2. Use define that category only when we are performing the filtering(you could use lsp-substring example which was added just before we started refactoring).

kiennq:
Hmm, first the `(length result)` will always fail since what `completion-all-completions` returned is in the form of `(,@candidates . number)`. It is a full list with last cdr point to some data instead of `nil`.
`length` will fail on that.

`company-capf` will try to call `completion-all-completions` to get candidates.
With your `add-function`, since the completion table pass to `company-capf` is a lambda, it will be called to resolve to completion list.
When the lambda is called, it triggers `lsp--capf-filter-candidates`, which will then trigger our call of `completion-all-completions` with completion list get from language server.
So the order of log appear in will be our `completion-all-completions`, then `company-capf`.
**Note** that, the order of calling is `company-capf`'s `completion-all-completions` triggers ours.

From this
```
input string=De pred=nil point=2 length=1
input string=@De pred=nil point=3 length=1
```

Our call to `completion-all-completions` is with correct prefix, we're getting that prefix from `textEdit`.
However, the prefix from `company-capf` is not correct. That's fine since we're not using that prefix for anything.

Can you try to advice with these functions instead, it will print correct number of completion items

``` elisp
(advice-add #'completion-all-completions :around
            (lambda (orig string table pred point &optional metadata)
              (let* ((result (funcall orig string table pred point metadata))
                     (last (last result))
                     (base-size (if (consp last) (cdr last))))
                (if (consp last) (setcdr last nil))
                (message "completion-all-completions string=%s length=%s" string (length result))
                (append result base-size)))
            '((name . --a1)))

(advice-add #'lsp--capf-filter-candidates :before
            (lambda (pred string _ &rest _)
              (message "lsp--capf-filter-candidates prefix: %s" string))
            '((name . --a1)))

```

yyoncho:
Here it is what I get. I am a bit busy in real live and dont have the time to investigate further based on your comments.

With flex:
``` 
lsp--capf-filter-candidates prefix: 
completion-all-completions string= length=0
completion-all-completions string=@ length=0
lsp--capf-filter-candidates prefix: D
completion-all-completions string=D length=0
completion-all-completions string=@D length=0
lsp--capf-filter-candidates prefix: De
completion-all-completions string=De length=0
completion-all-completions string=@De length=0
lsp--capf-filter-candidates prefix: Dep
completion-all-completions string=Dep length=0
completion-all-completions string=@Dep length=0
```

With `(basic partial-completion emacs22)`
```
lsp--capf-filter-candidates prefix: D
completion-all-completions string=D length=3
completion-all-completions string=@D length=3
lsp--capf-filter-candidates prefix: De
completion-all-completions string=De length=2
completion-all-completions string=@De length=2
lsp--capf-filter-candidates prefix: Dep
completion-all-completions string=Dep length=1
completion-all-completions string=@Dep length=1
lsp--capf-filter-candidates prefix: De
completion-all-completions string=De length=2
completion-all-completions string=@De length=2
```

kiennq:
Don't worry, if it's possible could you try with this advice and post the result here?
``` elisp
(advice-add #'completion-all-completions :around
            (lambda (orig string table pred point &optional metadata)
              (let* ((result (funcall orig string table pred point metadata))
                     (last (last result))
                     (base-size (if (consp last) (cdr last))))
                (if (consp last) (setcdr last nil))
                (message "completion-all-completions string=%s table-len=%s length=%s"
                         string
                         (if (and (not (functionp table)) (listp table)) (length table))
                         (length result))
                (append result base-size)))
            '((name . --a1)))
```

yyoncho:
here it is what I get:

```
completion-all-completions string= table-len=24 length=24
completion-all-completions string=@ table-len=nil length=24
completion-all-completions string=D table-len=24 length=3
completion-all-completions string=@D table-len=nil length=3
completion-all-completions string=De table-len=24 length=2
completion-all-completions string=@De table-len=nil length=2
completion-all-completions string=Dep table-len=24 length=1
completion-all-completions string=@Dep table-len=nil length=1
completion-all-completions string=De table-len=24 length=2
completion-all-completions string=@De table-len=nil length=2
(flex)
completion-all-completions string= table-len=24 length=0
completion-all-completions string=@ table-len=nil length=0
completion-all-completions string=D table-len=24 length=0
completion-all-completions string=@D table-len=nil length=0
completion-all-completions string=De table-len=24 length=0
completion-all-completions string=@De table-len=nil length=0
completion-all-completions string=Dep table-len=24 length=0
completion-all-completions string=@Dep table-len=nil length=0
completion-all-completions string=Depr table-len=24 length=0
completion-all-completions string=@Depr table-len=nil length=0
completion-all-completions string=Depre table-len=24 length=0
completion-all-completions string=@Depre table-len=nil length=0
completion-all-completions string=Deprea table-len=24 length=0
completion-all-completions string=@Deprea table-len=nil length=0
completion-all-completions string=Deprec table-len=24 length=0
```

As a side note you probably will be able to reproduce it on your side if you offset `(thing-at-point)` with 1 point and to generate the discrepancy between textEdit.range and thing-at-point

kiennq:
Btw, if you set `completion-styles` to `(basic flex)` do you get any fuzzy candidate now?

kiennq:
@yyoncho, I found out the reason.
All boil down to the implementation of `all-completions` in C code.
Although it states that
> COLLECTION can also be a function to do the completion itself.
It receives three arguments: the values STRING, PREDICATE and t.
Whatever it returns becomes the value of ‘all-completions’.

This is actually not entirely correct since
> Additionally to this predicate, ‘completion-regexp-list’
is used to further constrain the set of candidates.

So if only `flex` remains in `completion-styles`, we will have double filtering.
Most of the case, that's fine.
However in your case, since the prefix that `completion-at-point` returns include `@`, which is not present in the real prefix that language server returns, the filtering of `company-capf` will actually filter out all candidate.

To fix this, we need:
- Implement a better prefix detecting logic instead of using just symbol at point. This is independent from language server prefix and need to be a suffix of language server prefix.
Perhaps, we can walk from `bounds-start` to the first alphanumeric character and declare that character as start.
- Or switch to use some other trivial completion styles before `flex`.
For example, set `completion-styles` to `(basic flex)`. With this the `company-capf` will trigger `basic` completion (which returns whatever our completion table returns) while our filtering will handle both `(basic flex)` filtering.

What do you think?
The result is we have fuzzy completion

yyoncho:
> I agree with your points. But before that while testing this functionality I noticed that after setting completion-styles to flex the completion in elisp buffers became so slow so it is practically unusable(it takes . Which makes me think that the flex implementation in the core 

If we do that, we will still do 2 filtrations but it will work since the correct one will be bigger, right?

> Or switch to use some other trivial completion styles before flex.
For example, set completion-styles to (basic flex). With this the company-capf will trigger basic completion (which returns whatever our completion table returns) while our filtering will handle both (basic flex) filtering.

It seems like there is one more problem that we have to solve - while testing this functionality I noticed that flex completion is extremely slow in elisp buffer completion.

Here it is my cpu report:

```
- company-post-command                                                                                              396  50%
 - company--perform                                                                                                 378  48%
  - company--continue                                                                                               368  47%
   - company-calculate-candidates                                                                                   365  46%
    - company--fetch-candidates                                                                                     330  42%
     - company-call-backend-raw                                                                                     330  42%
      - apply                                                                                                       330  42%
       - company-capf                                                                                               330  42%
        - company-capf--candidates                                                                                  330  42%
         - completion-all-completions                                                                               288  37%
          - completion--nth-completion                                                                              288  37%
           - completion--some                                                                                       288  37%
            - #<compiled 0x157cabdefa55>                                                                            288  37%
             - completion-flex-all-completions                                                                      288  37%
              - completion-substring--all-completions                                                               247  31%
               - completion-pcm--all-completions                                                                    247  31%
                  all-completions                                                                                     1   0%
              - completion-pcm--hilit-commonality                                                                    41   5%
               - mapcar                                                                                              39   5%
                  #<compiled 0x157cabdefad9>                                                                         28   3%
         - #<compiled 0x157cabdefaf1>                                                                                42   5%
          - sort                                                                                                     40   5%
             #<compiled 0x1fcfb2b84b3f>                                                                              23   2%
```

Also, it generates unacceptable amount of memory, here it is the memory profile:

```
- company-post-command                                     52,514,135  58%
 - company--perform                                        49,556,195  55%
  - company--continue                                      48,354,767  54%
   - company-calculate-candidates                          48,343,239  54%
    - company--fetch-candidates                            47,550,623  53%
     - company-call-backend-raw                            47,550,623  53%
      - apply                                              47,550,623  53%
       - company-capf                                      47,550,623  53%
        - company-capf--candidates                         47,549,567  53%
         - completion-all-completions                      47,549,567  53%
          - completion--nth-completion                     47,549,567  53%
           - completion--some                              47,549,567  53%
            - #<compiled 0x157cad058e49>                   47,548,511  53%
             - completion-flex-all-completions             47,548,511  53%
              - completion-pcm--hilit-commonality          46,876,887  52%
               - mapcar                                    45,658,319  51%
                - #<compiled 0x157cad148d39>               41,980,679  47%
                   #<compiled 0x1fcfb2b85243>                 492,632   0%
```

I think that if (basic flex) is sufficient to make it work we should go this way. But considering that there is no flex completion in emacs 26 and that flex implementation in emacs 27 has poor performance IMO we should implement our flex implementation(there is one already in company-lsp) at least until emacs 27 flex implementation is fast enough. 

My emacs version is maybe a month old, so I will pull the latest version, and redo the perf tests and then I will report emacs bugs.

Note that, the completion is the most performance-critical part of lsp-mode and here we should aim for optimal performance. 

Let me know what do you think.

kiennq:
I think the flex filtering is implemented inside `all-completions`, which is quite efficient.
The part that consume both CPU and memory is `completion-pcm--hilit-commonality`, which calculate score and highlight part of text.

Nevertheless, I think setting `flex` to the end of `completion-styles` (i.e.: `(basic flex)`) is the best choice now. Fuzzy should only be served after there's no exact match.

For implementing our own completion styles, I'm a little bit skeptical for that. Given that Emacs 27 should be shipped soon. And user can probably using some thing like `company-flx` for that

yyoncho:
> CPU and memory is completion-pcm--hilit-commonality, which calculate score and highlight part of text.

I will test with servers that have huge completion results (e. g. dart) to check if it is fast enough. 

> And user can probably using some thing like company-flx for that

We have to verify that this is sufficient for pre emacs 27 users.

kiennq:
Also how about we just use `do-completions` and preserve the order from language server? It should be much faster compare to call `completion-all-completions`

yyoncho:
> Also how about we just use do-completions and preserve the order from language server? It should be much faster compare to call completion-all-completions

If we have prefix `Foo` and candidates `FooBar` and `BarFoo` we should have `FooBar` as a first candidate no matter what the server order is.

kiennq:
I'm not really worry about that case, since we will have `basic` completion which is favorite prefix match goes first.
The case that matching characters is spreading around, I'm not really sure if calculated score is making any different

yyoncho:
It will hurt usability, e. g. if you have prefix `F` and `aFoo`, `barfoo`, `barbarbarbarFoo` - IMO pretty much every editor does the weight-based sorting. If we have performance issues we should work with emacs-devel to solve them.

PR #1339, opened 2020-01-21
Respect XDG_CACHE_HOME for lsp-server-install-dir

tomoya:
Hi, The lsp-mode is a very cool package for me.

Emacs will support the XDG convention for init files from 27. So, I migrated configuration file management to the XDG convention.
I want the XDG convention to support lsp-mode.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5969,8 +5969,11 @@ SESSION is the active session."
 
 ;; download server
 
-(defcustom lsp-server-install-dir (expand-file-name
-                                   (locate-user-emacs-file (f-join ".cache" "lsp")))
+(defcustom lsp-server-install-dir (expand-file-name "lsp"
+                                     ;; Using xdg-cache-home for variable name occured package-lint error
+                                     (or (if-let (xdg-cache-home-dir (getenv "XDG_CACHE_HOME"))
+                                             (concat xdg-cache-home-dir "/emacs"))
+                                         (locate-user-emacs-file ".cache")))
   "Directory in which the servers will be installed."
   :risky t
   :type 'directory
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5971,7 +5971,7 @@ SESSION is the active session."
 
 (defcustom lsp-server-install-dir (expand-file-name "lsp"
                                      ;; Using xdg-cache-home for variable name occured package-lint error
-                                     (or (if-let (xdg-cache-home-dir (getenv "XDG_CACHE_HOME"))
+                                     (or (when-let (xdg-cache-home-dir (getenv "XDG_CACHE_HOME"))
                                              (concat xdg-cache-home-dir "/emacs"))
                                          (locate-user-emacs-file ".cache")))
   "Directory in which the servers will be installed."


lsp-mode.el
```diff
@@ -5969,8 +5969,11 @@ SESSION is the active session."
 
 ;; download server
 
-(defcustom lsp-server-install-dir (expand-file-name
-                                   (locate-user-emacs-file (f-join ".cache" "lsp")))
+(defcustom lsp-server-install-dir (expand-file-name "lsp"
+                                     ;; Using xdg-cache-home for variable name occured package-lint error
+                                     (or (if-let (xdg-cache-home-dir (getenv "XDG_CACHE_HOME"))
```
yyoncho: nit: use when-let when there is only then clause.

tomoya: Thank you for your advice.
My emacs lisp knowledge is stopping from 10 years ago. I learned if-let from lsp-mode code.

akirak: Emacs now has `xdg-cache-home` function built-in, but I am not sure if it is supported on other platforms and in older versions of Emacs. I suggest you use the function when it's available, because it falls back to `~/.cache` when the environment variable is undefined.

tomoya: Thank you for your suggestion.

https://emba.gnu.org/emacs/emacs/blob/emacs-27/lisp/xdg.el#L56
I found `xdg-cache-home` func in xdg.el. This library included from emacs-26.
The package-lint error also seems to be due to this.

I impliment to use `xdg-cache-home` like this.

```lisp
(unless (version< emacs-version "26")
  (require 'project)
  (require 'flymake)
  (require 'xdg))

...

(defcustom lsp-server-install-dir (expand-file-name "lsp"
                                     (or (when (getenv "XDG_CACHE_HOME")
                                           (concat (xdg-cache-home) "/emacs"))
                                         (locate-user-emacs-file ".cache")))

```

But package-lint returns this error.

```
5975:52: error: You should depend on (emacs "26.1") if you need `xdg-cache-home'.
```

image

I guess that lsp-mode should not change `(emacs "25.1")` for this pull-request.
How about think?

akirak: I am not sure if you need this kind of stuff. You can simply put data files directly under `~/.emacs.d` or `~/.config/emacs` using `locate-user-emacs-file` like most other packages do, and it is a recommended practice for package developers.

I use [no-littering](https://github.com/emacscollective/no-littering) to manage config and data files of Emacs. I already [sent a PR](https://github.com/emacscollective/no-littering/pull/112) for this package. If you want to conform to XDG personally, you can do so by setting `no-littering-var-directory` to some value.

yyoncho: @akirak do you have objections to submitting as it is? Using conditionly xgd when emacs version is bigger than 26 seems like overkill considering that the function it provides is pretty much single linner.

akirak: Yes, I don't think this feature is necessary. At present, I have never seen any recommendation for individual package developers to follow the XDG conventions. According to [this discussion](https://github.com/emacscollective/no-littering/pull/100#issuecomment-525041932), `(locate-user-emacs-file YOUR-PACKAGE-NAME)` is usually sufficient, and `no-littering` package lets you organize package data if you want. Note that `lsp-mode` is already supported by no-littering.

I think `xdg.el` is meant for working with external (non-Emacs) tools and programs. A single user can have multiple installations of Emacs on a machine, and it can be troublesome for storing package data in somewhere outside of `user-emacs-directory`. If anyone wants to store  server executables in a centralized location, he/she can customize `lsp-server-installation-dir` (and it is why the variable is customizable).

yyoncho: @emacs-lsp/lsp-mode can others share their opinion too?

sebastiansturm: While I don't have a strong opinion about this, I find @akirak's argument convincing

seagle0128:
curious. Doesn't locate-user-emacs-file respect XDG_CACHE_HOME?

lsp-mode.el
severen: Storing files _that are not configuration_ in `$XDG_CONFIG_HOME` flies directly in the face of the entire purpose of the XDG BaseDir spec, which is to provide useful and pragmatic separation of 3 disparate types of data: config files, data files (the most ambiguous category, admittedly), and cache files. The advantages of this separation should hopefully be obvious, but to make it explicit the two main advantages are (1) easily backupable config files _without_ possibly megabytes to gigabytes of transient cruft (2) easily manageable cached data.

While you can make the argument that the user should "just change the location themselves", it doesn't change the fact that by default lsp-mode would be grossly violating the etiquette around the BaseDir spec and reducing it to "dotfiles but in another directory".

yyoncho: @severen in your opinion what should be the default setting?

severen: The default should be to store it in `$XDG_DATA_HOME/emacs`, with appropriate an fallback for other operating systems or older versions of Emacs (which would be to just store it in `user-emacs-directory`).

yyoncho: @tomoya what do you think about @severen's proposal?

tspiteri: Also, `$XDG_DATA_HOME`, `$XDG_CONFIG_HOME`, `$XDG_CACHE_HOME` are often not set, in which case the XDG spec specifies the defaults `$HOME/.local/share`, `$HOME/.config`, `$HOME/.cache` respectively.

ericdallo:
@tomoya, any news on that? what do you think about severen's proposal?

yyoncho:
Cleaning up staled PRs.

PR #1341, opened 2020-01-21
Update lsp-groovy-vars.adoc

CsBigDataHub:
updated according to https://github.com/emacs-lsp/lsp-mode/blob/f90ba14b0574d2fac31b6c797b49113e4de632dd/lsp-clients.el#L134


yyoncho:

I think that all these are supposed to be autogenerated in some way but I am not familiar with how it is supposed to work @vibhavp is very busy lately.

PR #1345, opened 2020-01-23
Updates installation command for digestif

zeroset:
The installation command for the Latex LSP Server was broken. This
change escapes the double dash server and the order of command line
options.


yyoncho:

PR #1346, opened 2020-01-23
Implement metals decoration protocol

kurnevsky:
https://scalameta.org/metals/docs/editors/decoration-protocol.html
screenshot-2020-01-23-16:07:38

### DIFF
--- a/lsp-metals.el
+++ b/lsp-metals.el
@@ -149,6 +149,30 @@ Should be ignored if there is no open doctor window."
   (with-lsp-workspace workspace
     (lsp-send-execute-command command)))
 
+(defvar-local lsp-metals--decorations nil
+  "Overlays used by `metals/publishDecorations' handler.")
+
+(defun lsp-metals--publish-decorations (workspace params)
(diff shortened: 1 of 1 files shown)


lsp-metals.el
```diff
@@ -163,7 +182,9 @@ Should be ignored if there is no open doctor window."
  (make-lsp-client :new-connection (lsp-stdio-connection 'lsp-metals--server-command)
                   :major-modes '(scala-mode)
                   :priority -1
+                  :custom-capabilities '((experimental (decorationProvider . t)))
```
kurnevsky: It seems it'll conflict with https://github.com/emacs-lsp/lsp-treemacs/blob/d097670f1d77e0c806b0094df4ebb575fb83ed87/lsp-metals-treeview.el#L811-L822

yyoncho: Yes. lsp-treemacs must be changed to append in :experimental.

thinkiny: Is this working now ? Do i have to configure something ???

PR #1347, opened 2020-01-23
Don't pass empty argument to vhdl_ls

sebastiansturm:
I tried to use `lsp-mode` with `vhdl_ls` today and found that `vhdl_ls` crashes because it is being passed a single, empty command line argument. This change should resolve the `vhdl_ls` issue (at least it does on my machine), but I haven't checked if it breaks one of the other VHDL language servers


chrbirks:
It seems that some recent changes to `vhdl_ls` is now causing it to crash when giving it an empty parameter. I just tested your changes with all three servers and it seems to be working fine.

yyoncho:
Thank you @sebastiansturm and @chrbirks!

PR #1349, opened 2020-01-24
Add default keybindings and which-key integration

yyoncho:
- added `lsp-command-keymap` which will contain all `lsp-mode` shortcuts.
  Introduced `lsp-define-conditional-key` which can be used to define keys that
  are enabled only if certain condition is true.

- added `which-key` integration use ...
``` emacs-lisp
  (add-hook 'lsp-mode-hook #'lsp-enable-which-key-integration)
```
... to enable it.

- implemented workarround for https://debbugs.gnu.org/cgi/bugreport.cgi?bug=39149


yyoncho:
how it looks like:

which-key-demo

lsp-mode.el
```diff
@@ -2028,6 +2030,242 @@ CALLBACK - callback for the lenses."
...
+                         ,@body)
+                   item)))))
+
+(defvar lsp--describe-buffer nil)
+
+(defun lsp-describe-buffer-bindings-advice (buffer &optional _prefix _menus)
+  (setq lsp--describe-buffer buffer))
+
+(add-function :before
+              (symbol-function 'describe-buffer-bindings)
```
kiennq: nit: this can be just `#'describe-buffer-bindings`

lsp-mode.el
```diff
@@ -2028,6 +2030,242 @@ CALLBACK - callback for the lenses."
...
+                 (when (with-current-buffer lsp--describe-buffer
+                         ,@body)
+                   item)))))
+
+(defvar lsp--describe-buffer nil)
+
+(defun lsp-describe-buffer-bindings-advice (buffer &optional _prefix _menus)
+  (setq lsp--describe-buffer buffer))
+
+(add-function :before
```
kiennq: nit: why use `add-function` when `advice-add` covers more cases?

lsp-mode.el
```diff
@@ -2028,6 +2030,242 @@ CALLBACK - callback for the lenses."
...
+  (declare (indent 3)
+           (debug (form form form &rest sexp)))
+
+  `(define-key ,keymap ,key
+     '(menu-item
+       ,(format "maybe-%s" (or (car (cdr-safe def)) def))
+       ,def
+       :filter (lambda (item)
+                 (when (with-current-buffer lsp--describe-buffer
+                         ,@body)
```
kiennq: I think we can extend this handle multiple key bindings at once, like what `define-key` is doing.
``` elisp
(defmacro lsp-define-conditional-key (keymap key def cond &rest bindings)
  "In KEYMAP, define key sequence KEY as DEF conditionally.
This is like `define-key', except the definition disappears
whenever COND evaluates to nil.
BINDINGS is a list of (key def cond)."
  (declare (indent defun)
           (debug (form form form form &rest sexp)))
  (let (expr)
    (while key
      (setq expr
            (append expr `((define-key ,keymap ,key
                             '(menu-item
                               ,(format "maybe-%s" (or (car (cdr-safe def)) def))
                               nil
                               :filter (lambda (&optional _)
                                         (if ,cond ,def)))))))
      (setq key (pop bindings)
            def (pop bindings)
            cond (pop bindings)))
    `(progn ,@expr)))
```
Then we can do something like
``` elisp
(lsp-define-conditional-key "==" lsp-format-buffer (or (lsp-feature? "textDocument/rangeFormatting")
                                                                                        (lsp-feature? "textDocument/formatting"))
                                            "=r" lsp-format-region (lsp-feature? "textDocument/rangeFormatting"))
```

lsp-mode.el
yyoncho: Thanks, I didn't know that.

lsp-mode.el
yyoncho: I guess the answer is that it covers the requirements for this case. Is there any particular benefit from using advice-add in this case?

lsp-mode.el
yyoncho: Thank you, will refactor to that - I was following https://endlessparentheses.com/define-context-aware-keys-in-emacs.html .

lsp-mode.el
kiennq: https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Named-Functions.html
> they know how to deal with macros and autoloaded functions, they let describe-function preserve the original docstring as well as document the added advice, and they let you add and remove advice before a function is even defined.

lsp-mode.el
yyoncho: Done, with a bit different implementation:

``` emacs-lisp
(defmacro lsp-define-conditional-key (keymap key def cond &rest bindings)
  "In KEYMAP, define key sequence KEY as DEF conditionally.
This is like `define-key', except the definition disappears
whenever COND evaluates to nil.
BINDINGS is a list of (key def cond)."
  (declare (indent defun)
           (debug (form form form form &rest sexp)))
  (->> (cl-list* key def cond bindings)
       (-partition 3)
       (-map (-lambda ((key def cond))
               `(define-key ,keymap ,key
                  '(menu-item
                    ,(format "maybe-%s" def)
                    ,def
                    :filter (lambda (item)
                              (when (with-current-buffer lsp--describe-buffer
                                      ,cond)
                                item))))))
       macroexp-progn))
```

lsp-mode.el
yyoncho: Done - I guess we may benefit if the function is not loaded.

JohnC32:
On a related topic, I've been meaning to propose a PR to update the LSP menu and add keybindings. Below is screen shot of the menu. You get to pick what prefix you want (defaults to C-c t, t in honor of the classic "tags" system). Before proposing the PR, I need to figure out how to make items like the CCLS extensions optional, i.e. they appear if you are using CCLS. I also need to think about how to make the "dynamic views" switchable from auto mode to on demand mode. We have a large number of people using LSP and opinions vary on the dynamic views, e.g. showing the doc comments on the fly or explicitly asking for them. The dynamic views menu solves this problem by making it easy to pick between the two. The slight issue is that the checkboxes that influence this behavior are currently saving the customizations which is non-standard.

snip

yyoncho:
> Before proposing the PR, I need to figure out how to make items like the CCLS extensions optional, i.e. they appear if you are using CCLS.

This PR has examples of how to make a menu optional - check `:filter` element. And I think that it covers what you are looking for (except for the CCLS part).

lsp-mode.el
```diff
@@ -2028,6 +2030,223 @@ CALLBACK - callback for the lenses."
...
+       (-map (-lambda ((key def cond))
+               `(define-key ,keymap ,key
+                  '(menu-item
+                    ,(format "maybe-%s" def)
+                    ,def
+                    :filter (lambda (item)
+                              (when (with-current-buffer lsp--describe-buffer
+                                      ,cond)
+                                item))))))
+       macroexp-progn))
```
kiennq: This looks nice. Do you need `(require 'macroexp)` too?

yyoncho: Probably something that we require already loads this package or the byte-compilation tests would fail. I guess it won't be an issue if we add it.

PR #1352, opened 2020-01-24
Fix type argument error when using xref-find-references

0x28:
Hi and thank you for this awesome mode.

I recently had the problem that I was getting `Wrong type argument: number-or-marker-p` when using `xref-find-references` in combination with the Rust Language Server (RLS).

The problem was that the range in the LSP message wasn't parsed correctly. This PR fixes this problem.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4882,8 +4882,10 @@ perform the request synchronously."
 
 (defun lsp--xref-elements-index (symbols path)
   (-mapcat
-   (-lambda ((&hash "name" "children" "selectionRange" (&hash? "start") "location"))
-     (cons (cons (concat path name) (lsp--position-to-point (or start location)))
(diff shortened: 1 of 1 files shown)


yyoncho:
Can provide an example which was failing before your fix?

0x28:
Sure thing. I created a small test project: https://github.com/0x28/lsp_test

When putting point on `inc` and calling `xref-find-references` I get the type error.

yyoncho:
@0x28 thank you, I will test it tomorrow my time and proceed with the merge.

yyoncho:
Thank you for contributing to `lsp-mode` and congrats for your first `lsp-mode` PR!

As a side note, you may use `(setq xref-prompt-for-identifier nil)` which will always use the identifier at point.

PR #1353, opened 2020-01-26
Send metals/didFocusTextDocument notification on buffer change.

kurnevsky:
https://scalameta.org/metals/docs/editors/new-editor.html#metalsdidfocustextdocument

Should be sent for decorations to work properly: https://github.com/scalameta/metals/issues/1258#issuecomment-578398187

### DIFF
--- a/lsp-metals.el
+++ b/lsp-metals.el
@@ -183,6 +183,15 @@ Should be ignored if there is no open doctor window."
                         (c (ignore (lsp-warn "Unknown metals client command: %s" c))))))
     (apply command (append (list workspace) (ht-get params "arguments") nil))))
 
+(defvar lsp-metals--current-buffer nil
+  "Current buffer used to send `metals/didFocusTextDocument' notification.")
+
+(defun lsp-metals--did-focus ()
+  "Send `metals/didFocusTextDocument' when current buffer changes."
+  (unless (eq lsp-metals--current-buffer (current-buffer))
+    (setq lsp-metals--current-buffer (current-buffer))
+    (lsp-notify "metals/didFocusTextDocument" (lsp--buffer-uri))))
+
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection 'lsp-metals--server-command)
                   :major-modes '(scala-mode)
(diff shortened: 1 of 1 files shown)


lsp-metals.el
```diff
@@ -183,6 +183,15 @@ Should be ignored if there is no open doctor window."
                         (c (ignore (lsp-warn "Unknown metals client command: %s" c))))))
     (apply command (append (list workspace) (ht-get params "arguments") nil))))
 
+(defvar lsp-metals--current-buffer nil
+  "Current buffer used to send `metals/didFocusTextDocument' notification.")
+
+(defun lsp-metals--did-fucus ()
```
yyoncho: typo: focus

lsp-metals.el
```diff
@@ -183,6 +183,15 @@ Should be ignored if there is no open doctor window."
...
+  "Send `metals/didFocusTextDocument' when current buffer changes."
+  (when (not (eq lsp-metals--current-buffer (current-buffer)))
```
yyoncho: nit: instead of `(when (not` use `(unless`

lsp-metals.el
```diff
@@ -183,6 +183,15 @@ Should be ignored if there is no open doctor window."
                         (c (ignore (lsp-warn "Unknown metals client command: %s" c))))))
     (apply command (append (list workspace) (ht-get params "arguments") nil))))
 
+(defvar lsp-metals--current-buffer nil
```
kurnevsky: Maybe better to store it in workspace metadata?

yyoncho: Is it working fine with this version of the code?

kurnevsky: Yes, but it will send unnecessary notifications when switching between different metals workspaces.

yyoncho: Before sending a message you may check if the buffer is managed by metals instance. As a side note. E. g.
```
(with-current-buffer "HelloSpec.scala"
  (lsp--workspace-buffers (lsp-find-workspace 'metals (buffer-file-name)) ))
```

kurnevsky: I add local hooks only to metals buffers, so no need to check it.

yyoncho: So IIUC 

> it will send unnecessary notifications when switching between different metals workspaces.

is not a problem?

kurnevsky: Yes, not a problem at all :)

Issue #1280, opened 2019-12-29
Test lsp--uri-to-path--handle-utf8 fails in Debian build


thkoch2001:
I'm packaging version 6.2.1 for Debian and this test fails:

```
Test lsp--uri-to-path--handle-utf8 condition:
    (ert-test-failed             
     ((should                                              
       (equal                                   
        (lsp--uri-to-path "file:///c:/Users/%E4%BD%A0%E5%A5%BD/")
        "c:/users/你好/"))                                
      :form                                                         
      (equal "c:/users/\344\275\240\345\245\275/" "c:/users/你好/")
      :value nil :explanation                                                 
      (arrays-of-different-length 16 12 "c:/users/\344\275\240\345\245\275/" "c:/users/你好/" first-mismatch-at 9)))
   FAILED  14/29  lsp--uri-to-path--handle-utf8
```

The test was introduced to fix #646.

Relevant packages installed to run the test:

elpa-dash_2.16.0+dfsg-1
elpa-dash-functional_1.2.0+dfsg-6
elpa-f_0.20.0-2
elpa-ht_2.2-2
elpa-lv_0.15.0-2
elpa-markdown-mode_2.3+210-1
elpa-s_1.12.0-3
elpa-spinner_1.7.3-2
emacs-nox_1:26.3+1-1

CC @wenxin-wang

wenxin-wang:
It seems that variable `locale-coding-system` is not set to utf-8 in your test environment, can you help me verify that? It seems that I overlooked such cases in that PR.

thkoch2001:
emacs -Q --batch --eval "(print locale-coding-system)"

in the build schroot gives `nil`.

wenxin-wang:
Thank you for the confirmation. A [quick and dirty fix](https://github.com/wenxin-wang/lsp-mode/commit/ecd8c7d64c7f0abc6a2f276e9ffe7d0751f640c3) would be falling-back to `utf-8` when `locale-coding-system` is `nil`. If you have time, can you try it?

I really feel there should be a better way. Can emacs try and detect encoding in a string? If not, this is the best solution I could think of.

yyoncho:
@wenxin-wang would you mind creating a PR from the proposed fix?

wenxin-wang:
Of course! Please check https://github.com/emacs-lsp/lsp-mode/pull/1354

PR #1354, opened 2020-01-26
lsp--uri-to-path: use utf-8 when locale-coding-system unset

wenxin-wang:
Try to fix https://github.com/emacs-lsp/lsp-mode/issues/1280


yyoncho:

PR #1355, opened 2020-01-26
Don't prompt about opening a web page if server can't be installed

thomasf:
With python and the pyls language server I get this prompt every time I open
any python file when I haven't already set up the language server. It is very
annoying so this patch demotes it from a yes or no prompt to a warn log message.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6774,14 +6774,13 @@ The server(s) will be started in the buffer when it has finished."
        ((setq clients (unless matching-clients
                         (lsp--filter-clients (-andfn #'lsp--matching-clients?
                                                      (-not #'lsp--server-binary-present?)))))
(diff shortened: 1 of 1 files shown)


thomasf:
I guess a possible downside with this change if is a newbie Emacs user don't know about the `*Messages*` buffer and don't know where to find the message/URL after it's being displayed in the mini buffer.

thomasf:
As one would suspect, this is how it looks after the change:

image

yyoncho:
> I guess a possible downside with this change if is a newbie Emacs user don't know about the `*Messages*` buffer and don't know where to find the message/URL after it's being displayed in the mini buffer.

We assume that lsp-mode users will at least know about it.

thomasf:
I see an error, the message still has a question about opening the url. I'll fix it later this evening;

yyoncho:

PR #1357, opened 2020-01-28
Implemented debug adapter protocol support for metals

yyoncho:
- fixes https://github.com/emacs-lsp/dap-mode/issues/196

### DIFF
--- a/lsp-metals.el
+++ b/lsp-metals.el
@@ -180,6 +180,7 @@ Should be ignored if there is no open doctor window."
                         (`"metals-doctor-reload" #'lsp-metals--doctor-reload)
                         (`"metals-goto-location" #'lsp-metals--goto-location)
                         (`"metals-echo-command" #'lsp-metals--echo-command)
+                        (`"metals-model-refresh" #'lsp-metals--model-refresh)
                         (c (ignore (lsp-warn "Unknown metals client command: %s" c))))))
     (apply command (append (list workspace) (ht-get params "arguments") nil))))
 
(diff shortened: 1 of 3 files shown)


yyoncho:
I could do a bit better with this PR(I have mirrored what vscode does): The port could be obtained in the debug provider which would allow using `dap-debug-last`.

cc @kurnevsky @dsyzling @rchallapalli

lsp-metals.el
```diff
@@ -192,21 +193,51 @@ Should be ignored if there is no open doctor window."
...
-                                             ("metals/publishDecorations" #'lsp-metals--publish-decorations)
-                                             ("metals/treeViewDidChange" #'ignore))
-		  :server-id 'metals
-                  :initialized-fn (lambda (workspace)
-                                    (with-lsp-workspace workspace
-                                      (lsp--set-configuration
-                                       (lsp-configuration-section "metals"))))
-                  :after-open-fn (lambda ()
-                                   (add-hook 'lsp-on-idle-hook #'lsp-metals--did-focus nil t))))
+ (make-lsp-client
```
kurnevsky: Did you reformat it intentionally?

lsp-metals.el
```diff
@@ -192,21 +193,51 @@ Should be ignored if there is no open doctor window."
...
+                            cl-third
+                            string-to-number)
+           :type "scala"
+           :name (gethash "name" session-params)
+           :host "localhost"
+           :request "launch"
+           :noDebug no-debug))))
+
+(defun lsp-metals--model-refresh (_params)
+  (when (and (buffer-file-name)
```
kurnevsky: First argument for such handlers is a workspace, so no need to find it.

yyoncho: I used it as a guard for the case in which the user has switched to a buffer not managed by metals. But now I think that I should change the code to refresh all buffers that are managed by metals - or at least this is what vscode seems to be doing.

kurnevsky: But we receive this notification for particular workspace, so we should refresh it only there, regardless of where the user is. Shouldn't we?

yyoncho: Yes, I meant that we should refresh all buffers that are managed by the current metals workspace.

yyoncho:
Addressed comments.

lsp-metals.el
```diff
@@ -192,15 +193,46 @@ Should be ignored if there is no open doctor window."
     (setq lsp-metals--current-buffer (current-buffer))
     (lsp-notify "metals/didFocusTextDocument" (lsp--buffer-uri))))
 
+(declare-function dap-debug (conf))
+(declare-function dap-register-debug-provider (name conf))
```
kiennq: I think this should be
``` elisp
(declare-function dap-debug "ext:dap-mode" (conf))
(declare-function dap-register-debug-provider "ext:dap-mode" (name conf))
```

yyoncho: Yes, you are right - interestingly the byte-compilation tests do not fail...

lsp-metals.el
```diff
@@ -192,15 +193,46 @@ Should be ignored if there is no open doctor window."
...
+    (dap-debug
+     (list :debugServer (-> (gethash "uri" session-params)
+                            (split-string ":")
+                            cl-third
+                            string-to-number)
+           :type "scala"
+           :name (gethash "name" session-params)
+           :host "localhost"
+           :request "launch"
+           :noDebug no-debug))))
```
kiennq: Just asking but can we do something similar to `lsp-rust` for `:action-handlers`?
``` elisp
:action-handlers (ht ("rls.run" 'lsp-rust--rls-run))
```
Then we can define a `cl-defmethod` for `metals-debug-session-start` and `metals-run-session-start` inside `dap-metals`.
That way the `lsp-mode` doesn't need to be aware of `dap-mode` and we can keep clean dependency between `lsp-mode` and `dap-mode`.

yyoncho: We can achieve that in multiple ways, e. g. editing the client itself - ATM we are having a discussion to move lsp-metals in a separate package to avoid that kind of external deps so I went for the simplest solution .

PR #1362, opened 2020-01-30
Metals refresh lens only if lsp-lens-mode is enabled

zhenwenc:
The code lens should not be rendered when `lsp-lens-mode` is disabled.


yyoncho:

PR #1367, opened 2020-01-31
Include ghdl-ls as vhdl native, direct server

csantosb:
No need to go through hdl-checker


yyoncho:
@chrbirks willing to take a look?

chrbirks:
Sure, I'll take a look at it tomorrow.

csantosb:
Yeah, ghdl ls is bleeding edge: I'm using arch linux and installed aur/ghdl-llvm-git with no issue. 
Anyway, I guess the same applies when using hdl checker, right ?

yyoncho:
@csantosb 
Thank you for the PR!

@chrbirks 
Thank you for the review!

chrbirks:
I was using the ghdl package from AUR (version 0.36-1) but didn't have ghdl ls installed in any way, so now that you mention it it was strange why HDL Checker was working for me.

Looking at https://github.com/suoto/hdl_checker/blob/master/hdl_checker/builders/ghdl.py it looks like HDL Checker just calls `ghdl` with the `-a` analyze source and `-s` syntax check flags. So you probably just get basic linting when using HDL Checker.

Issue #1363, opened 2020-01-31
lsp breaks magit-log-trace-definition


agzam:
`magit-log-definition` depends on `(which-function)` to figure out name of the function. Because lsp-mode changes `imenu--index-alist` (by appending things like `(Function)` or `(Variable)` to each item) and `(which-function)` depends on it, this breaks aforementioned function. 

And who knows, maybe it's not only this specific function, perhaps many other things that depend on `(which-function)` to return strictly the name of the function, now would break as well.

Can someone suggest a workaround, please?

agzam:
For now I've settled by doing `(setq lsp-enable-imenu nil)`. But that's more like a temp. workaround and not the fix.

yyoncho:
lsp-mode does not just append `(Function)` and `(Variable)` but implements imenu provider. It seems like there is either a which-function bug or lsp-mode is not providing the correct imenu format.

yyoncho:
@agzam can you tell me which mode is broken?

agzam:
it breaks `(which-function)`, when you run it with the cursor being anywhere inside a function body it should return just the name of the function. But when `lsp-enable-imenu` is set to true and lsp gets connected to a lsp-server (in my case that's clojure-lsp), it returns something like: `foo (Function)`. I believe I have explained it in the top comment. Let me know if it remains unclear even after reviewing it.

yyoncho:
I see - most likely those tags `(Function)` are not needed. I will take a look if they can be removed at all.

agzam:
I think it's useful for imenu to show distinction between variables and functions, I don't know if this feature necessarily has to be removed [in order for (which-function) to work properly]. I haven't had time to actually dig a bit more. Maybe I will get a chance to look into this over the weekend. I'll let you know if I find anything interesting. Thank you!

yyoncho:
Proposed fix at https://github.com/emacs-lsp/lsp-mode/pull/1368

PR #1368, opened 2020-02-01
Do not append (kind) to imenu items

yyoncho:
- Fixes https://github.com/emacs-lsp/lsp-mode/issues/1363

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5646,10 +5646,9 @@ an alist
   (let* ((start-point (lsp--symbol-get-start-point sym))
          (name (gethash "name" sym)))
     (if (seq-empty-p (gethash "children" sym))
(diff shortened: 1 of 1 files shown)


yyoncho:
> Writing our own wrapper over imenu should do it, but I'm not sure it's worth the effort for such a small addition.

IMO we should bypass `imenu` because it does not match the lsp protocol, e. g.:

1. You could have only symbols from the current document.
2. You cannot mark members that are deprecated. 

What I was planning to investigate but for `helm` was the ability to put the `lsp-treemacs-symbols` in helm buffer combine both features. 

IMO we should integrate with emacs facilities but if they are limiting us (e. g. do not allow us to expose everything what lsp protocol provides) we create our own tools(and eventually file feature requests against core). 

Another way to achieve that is to create tools similar to `lsp-ivy-workspace-symbols` but for document symbols.

PR #1375, opened 2020-02-02
Remember buffer for scheduled lsp-lens-refresh

kurnevsky:


lsp-mode.el
```diff
@@ -1899,16 +1900,17 @@ BUFFER-MODIFIED? determines whether the buffer is modified or not."
           (delete-overlay it)))
       (setq-local lsp--lens-overlays overlays))))
 
-(defun lsp-lens-refresh (buffer-modified?)
+(defun lsp-lens-refresh (buffer-modified? &optional buffer)
```
kurnevsky: Is it ok to modify this function or should I create another internal one?

yyoncho: as long as we are backward compatible it is fine.

PR #1384, opened 2020-02-06
Move flycheck integration in lsp-mode

yyoncho:
- fixes https://github.com/emacs-lsp/lsp-ui/issues/301

- Introduced `lsp-diagnostic-package` which will replace `lsp-prefer-flymake`
with the following improvement: default value is :auto which means pick flycheck
if it is present, otherwise use flymake. The nil values and t are kept for
backward compatibility.

- I will do a separate PR for lsp-ui to make everything backward in lsp-ui too.

### DIFF
--- a/README.org
+++ b/README.org
@@ -53,18 +53,16 @@
    - 🌟 *Flexible* - could be configured as full-blown IDE with flashy UI or minimal distraction free
 ** Overview
    Client for [[https://github.com/Microsoft/language-server-protocol/][Language Server Protocol]] (v3.14). [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] aims to provide IDE-like experience by providing optional integration with the most popular Emacs packages like ~company~, ~flycheck~ and ~projectile~.
-
    - Non-blocking asynchronous calls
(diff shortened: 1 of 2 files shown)


README.org
```diff
@@ -53,18 +53,16 @@
    - 🌟 *Flexible* - could be configured as full-blown IDE with flashy UI or minimal distraction free
 ** Overview
    Client for [[https://github.com/Microsoft/language-server-protocol/][Language Server Protocol]] (v3.14). [[https://github.com/emacs-lsp/lsp-mode][lsp-mode]] aims to provide IDE-like experience by providing optional integration with the most popular Emacs packages like ~company~, ~flycheck~ and ~projectile~.
-
    - Non-blocking asynchronous calls
-   - Real-time Diagnostics/linting (via builtin ~flymake~ when Emacs > 26, requires flymake>=1.0.5 or [[https://github.com/flycheck/flycheck][flycheck]]/[[https://github.com/emacs-lsp/lsp-ui][lsp-ui]])
-   - Code completion - using [[https://github.com/tigersoldier/company-lsp][company-lsp]] or builtin ~completion-at-point~
+   - Real-time Diagnostics/linting (via ~flymake~ when Emacs > 26, requires flymake>=1.0.5 or builtin [[https://github.com/flycheck/flycheck][flycheck]])
```
laurynas-biveinis: It is flymake not flycheck which is builtin?

s-kostyaev: Flymake is builtin and flycheck is external

PR #1388, opened 2020-02-07
Fix formatting lsp-method-requirements

### LINKED ISSUE #1387: lsp-feature? does not seem to work, so indent-region-function gets set incorrectly
**Describe the bug**
In ruby files with solargraph, indent-region-function is getting set. It looks like this code is supposed to prevent that if the server does not support it:

https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L3226-L3228

Unfortunately, `(lsp-feature? "xxxx")` always returns something in ruby files (and others, it seems), regardless if what string you give it.

I'm on Emacs 27, if that matters.

**To Reproduce**

Open any file with lsp enabled and `M-: (lsp-feature? "textDocument/rangeFormatting")`. It should return nil, but does not.

**Expected behavior**
Should return nil

solargraph, ts-ls, likely others

**OS**
macos Catalina

**Error callstack**
None

aaronjensen:
Fixes #1387


aaronjensen:
I don't know if this was intentional or not, but without it, the feature detection is broken for range formatting.

yyoncho:

> I don't know if this was intentional or not, but without it, the feature detection is broken for range formatting.

it was a typo

CC @axelson

aaronjensen:
Thank you for the quick merge!

axelson:
It works! Thanks @aaronjensen and @yyoncho.

PR #1393, opened 2020-02-08
Support pyenv in pyls

15cm:
Different environment is supported for pyls in
palantir/python-language-server#680. Try to get the
environment specified by pyenv and pass it in.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5418,7 +5418,7 @@ WORKSPACE is the active workspace."
                                         (lsp--error "Failed to apply edits with message %s" (error-message-string err))
                                         :json-false))))
                      ((string= method "workspace/configuration")
-                      (lsp--build-workspace-configuration-response params))
+                      (with-lsp-workspace workspace (lsp--build-workspace-configuration-response params)))
                      ((string= method "workspace/workspaceFolders")
                       (let ((folders (or (-> workspace
                                              (lsp--workspace-client)
@@ -6490,10 +6490,11 @@ changing the value of `foo'."
 
 (defun lsp-register-custom-settings (props)
   "Register PROPS.
-PROPS is list of triple (path value boolean?) where PATH is the
+PROPS is list of triple (path value boolean? ignore-on-nil?) where PATH is the
 path to the property, VALUE is either a literal value or symbol
-used to retrieve the value, and BOOLEAN? is an optional flag that
-should be non-nil for boolean settings."
+used to retrieve the value, BOOLEAN? is an optional flag that should be non-nil
+for boolean settings, and IGNORE-ON-NIL? is an optional flag that ignores the
+setting if its evaluated value is nil."
   (let ((-compare-fn #'lsp--compare-setting-path))
     (setq lsp-client-settings (-uniq (append props lsp-client-settings)))))
 
@@ -6516,15 +6517,17 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
 (defun lsp-configuration-section (section)
   "Get settings for SECTION."
   (let ((ret (ht-create)))
-    (mapc (-lambda ((path variable boolean?))
+    (mapc (-lambda ((path variable boolean? ignore-on-nil?))
             (when (s-matches? (concat section "\\..*") path)
               (let* ((symbol-value (if (symbolp variable)
                                        (symbol-value variable)
-                                     variable))
+                                     (if (functionp variable)
+                                         (funcall variable) variable)))
                      (value (if (and boolean? (not symbol-value))
                                 :json-false
                               symbol-value)))
-                (lsp-ht-set ret (s-split "\\." path) value))))
+                (unless (and ignore-on-nil? (not value))
+                  (lsp-ht-set ret (s-split "\\." path) value)))))
           lsp-client-settings)
     ret))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6490,11 +6490,11 @@ changing the value of `foo'."
 
 (defun lsp-register-custom-settings (props)
   "Register PROPS.
-PROPS is list of triple (path value boolean? ignore-on-nil?) where PATH is the
-path to the property, VALUE is either a literal value or symbol
-used to retrieve the value, BOOLEAN? is an optional flag that should be non-nil
-for boolean settings, and IGNORE-ON-NIL? is an optional flag that ignores the
-setting if its evaluated value is nil."
+PROPS is list of triple (path value boolean?) where PATH is the path to the
+property; VALUE can be a literal value, symbol to be evaluated, or lambda
+function to be called; BOOLEAN? is an optional flag that should be non-nil for
+boolean settings, when it is nil the property will be ignored if the VALUE is
+nil."
   (let ((-compare-fn #'lsp--compare-setting-path))
     (setq lsp-client-settings (-uniq (append props lsp-client-settings)))))
 
@@ -6517,7 +6517,7 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
 (defun lsp-configuration-section (section)
   "Get settings for SECTION."
   (let ((ret (ht-create)))
-    (mapc (-lambda ((path variable boolean? ignore-on-nil?))
+    (mapc (-lambda ((path variable boolean?))
             (when (s-matches? (concat section "\\..*") path)
               (let* ((symbol-value (if (symbolp variable)
                                        (symbol-value variable)
@@ -6526,7 +6526,7 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
                      (value (if (and boolean? (not symbol-value))
                                 :json-false
                               symbol-value)))
-                (unless (and ignore-on-nil? (not value))
+                (when (or boolean? value)
                   (lsp-ht-set ret (s-split "\\." path) value)))))
           lsp-client-settings)
     ret))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6491,10 +6491,10 @@ changing the value of `foo'."
 (defun lsp-register-custom-settings (props)
   "Register PROPS.
 PROPS is list of triple (path value boolean?) where PATH is the path to the
-property; VALUE can be a literal value, symbol to be evaluated, or lambda
-function to be called; BOOLEAN? is an optional flag that should be non-nil for
-boolean settings, when it is nil the property will be ignored if the VALUE is
-nil."
+property; VALUE can be a literal value, symbol to be evaluated, or either a
+function or lambda function to be called without arguments; BOOLEAN? is an
+optional flag that should be non-nil for boolean settings, when it is nil the
+property will be ignored if the VALUE is nil."
   (let ((-compare-fn #'lsp--compare-setting-path))
     (setq lsp-client-settings (-uniq (append props lsp-client-settings)))))
 
@@ -6520,7 +6520,9 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
     (mapc (-lambda ((path variable boolean?))
             (when (s-matches? (concat section "\\..*") path)
               (let* ((symbol-value (if (symbolp variable)
-                                       (symbol-value variable)
+                                       (if (fboundp variable)
+                                           (funcall variable)
+                                         (symbol-value variable))
                                      (if (functionp variable)
                                          (funcall variable) variable)))
                      (value (if (and boolean? (not symbol-value))
--- a/lsp-pyls.el
+++ b/lsp-pyls.el
@@ -365,8 +365,16 @@ parameters referenced in config."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-pyls-plugins-jedi-environment nil
+  "Specify the environment that jedi runs on where <environmeent>/bin/python
+should be the python executable. This option will be prioritized over
+`lsp-pyls-plugins-jedi-use-pyenv-environment'."
+  :type 'string
+  :group 'lsp-pyls
+  :package-version '(lsp-mode . "6.3"))
+
 (lsp-register-custom-settings
- (append '(("pyls.rope.ropeFolder" lsp-pyls-rope-rope-folder)
+ '(("pyls.rope.ropeFolder" lsp-pyls-rope-rope-folder)
            ("pyls.rope.extensionModules" lsp-pyls-rope-extension-modules)
            ("pyls.plugins.autopep8.enabled" lsp-pyls-plugins-autopep8-enabled t)
            ("pyls.plugins.yapf.enabled" lsp-pyls-plugins-yapf-enabled t)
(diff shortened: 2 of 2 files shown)


lsp-mode.el
```diff
@@ -6516,15 +6517,17 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
 (defun lsp-configuration-section (section)
   "Get settings for SECTION."
   (let ((ret (ht-create)))
-    (mapc (-lambda ((path variable boolean?))
+    (mapc (-lambda ((path variable boolean? ignore-on-nil?))
```
yyoncho: I am not sure if this has to be conditional - can we make this the default behavior and strip all properties that are nil and that are not `boolean?`.

lsp-pyls.el
```diff
@@ -353,53 +359,77 @@ parameters referenced in config."
...
+                      (when root
+                        (setenv "PYENV_VERSION" nil)
+                        (let ((python-env (f-parent
+                                           (f-parent
+                                            (shell-command-to-string
+                                             (format "PYENV_DIR='%s' %s which python"
+                                                     root lsp-pyls-pyenv-command-path))))))
+                          (message (format "Configure pyls with environment: %s" python-env))
+                          (setenv "PYENV_VERSION" pyenv-version)
+                          python-env)
```
yyoncho: nit: as per elisp coding guidelines(which we mostly follow) the brackets must be closed on the same line.

lsp-pyls.el
```diff
@@ -353,53 +359,77 @@ parameters referenced in config."
...
+           ("pyls.plugins.jedi_definition.follow_imports" lsp-pyls-plugins-jedi-definition-follow-imports t)
+           ("pyls.plugins.jedi_definition.enabled" lsp-pyls-plugins-jedi-definition-enabled t)
+           ("pyls.plugins.jedi_completion.include_params" lsp-pyls-plugins-jedi-completion-include-params t)
+           ("pyls.plugins.jedi_completion.enabled" lsp-pyls-plugins-jedi-completion-enabled t)
+           ("pyls.configurationSources" lsp-pyls-configuration-sources))
+         (when lsp-pyls-plugins-jedi-use-pyenv-environment
+           (let ((pyenv-get-python-executable
+                  (lambda ()
+                    (let* ((pyenv-version (getenv "PYENV_VERSION"))
+                           (root (nth 0 (lsp-find-roots-for-workspace lsp--cur-workspace (lsp-session)))))
```
yyoncho: nit: please use lsp-seq-first or cl-first instead of `(nth 0`

lsp-pyls.el
```diff
@@ -353,53 +359,77 @@ parameters referenced in config."
...
+                      (when root
+                        (setenv "PYENV_VERSION" nil)
+                        (let ((python-env (f-parent
+                                           (f-parent
+                                            (shell-command-to-string
+                                             (format "PYENV_DIR='%s' %s which python"
+                                                     root lsp-pyls-pyenv-command-path))))))
+                          (message (format "Configure pyls with environment: %s" python-env))
```
yyoncho: nit: use `lsp--info` instead of `message`.

lsp-pyls.el
```diff
@@ -353,53 +359,77 @@ parameters referenced in config."
...
-   ("pyls.plugins.jedi_signature_help.enabled" lsp-pyls-plugins-jedi-signature-help-enabled t)
-   ("pyls.plugins.jedi_references.enabled" lsp-pyls-plugins-jedi-references-enabled t)
-   ("pyls.plugins.jedi_hover.enabled" lsp-pyls-plugins-jedi-hover-enabled t)
-   ("pyls.plugins.jedi_definition.follow_builtin_imports" lsp-pyls-plugins-jedi-definition-follow-builtin-imports t)
-   ("pyls.plugins.jedi_definition.follow_imports" lsp-pyls-plugins-jedi-definition-follow-imports t)
-   ("pyls.plugins.jedi_definition.enabled" lsp-pyls-plugins-jedi-definition-enabled t)
-   ("pyls.plugins.jedi_completion.include_params" lsp-pyls-plugins-jedi-completion-include-params t)
-   ("pyls.plugins.jedi_completion.enabled" lsp-pyls-plugins-jedi-completion-enabled t)
-   ("pyls.configurationSources" lsp-pyls-configuration-sources)))
+ (append '(("pyls.rope.ropeFolder" lsp-pyls-rope-rope-folder)
```
yyoncho: consider splitting this to two separate `lsp-register-custom-settings` instead of using one with `append`.

lsp-mode.el
```diff
@@ -6490,10 +6490,11 @@ changing the value of `foo'."
 
 (defun lsp-register-custom-settings (props)
   "Register PROPS.
-PROPS is list of triple (path value boolean?) where PATH is the
+PROPS is list of triple (path value boolean? ignore-on-nil?) where PATH is the
 path to the property, VALUE is either a literal value or symbol
-used to retrieve the value, and BOOLEAN? is an optional flag that
-should be non-nil for boolean settings."
+used to retrieve the value, BOOLEAN? is an optional flag that should be non-nil
```
yyoncho: can you update the docs mentioning that it may be also a function?

lsp-pyls.el
```diff
@@ -353,53 +359,77 @@ parameters referenced in config."
...
+           ("pyls.plugins.jedi_references.enabled" lsp-pyls-plugins-jedi-references-enabled t)
+           ("pyls.plugins.jedi_hover.enabled" lsp-pyls-plugins-jedi-hover-enabled t)
+           ("pyls.plugins.jedi_definition.follow_builtin_imports" lsp-pyls-plugins-jedi-definition-follow-builtin-imports t)
+           ("pyls.plugins.jedi_definition.follow_imports" lsp-pyls-plugins-jedi-definition-follow-imports t)
+           ("pyls.plugins.jedi_definition.enabled" lsp-pyls-plugins-jedi-definition-enabled t)
+           ("pyls.plugins.jedi_completion.include_params" lsp-pyls-plugins-jedi-completion-include-params t)
+           ("pyls.plugins.jedi_completion.enabled" lsp-pyls-plugins-jedi-completion-enabled t)
+           ("pyls.configurationSources" lsp-pyls-configuration-sources))
+         (when lsp-pyls-plugins-jedi-use-pyenv-environment
+           (let ((pyenv-get-python-executable
```
yyoncho: Do we need here a variable as well? I mean the logic might be - use the variable if the user has configured it if he hasn't try to automatically find the env. Also, it seems like the check for `lsp-pyls-plugins-jedi-use-pyenv-environment` should go in the lambda itself because you could have 2 projects - both of them have pyenv but you want to use pyenv for only one of them.

lsp-mode.el
15cm: Make sense. Done

lsp-pyls.el
15cm: Agree. With the variable you mentioned, added as `lsp-pyls-plugins-jedi-environment`, I am considering an approach that let users get the environment by `pyenv` and pass/set it for `lsp-mode`. However, since the custom-settings are for each `workspace` and users don't have information of `workspace`, they will have get it by `buffer-name`, `projectile-project-root` etc. As the `workspace` is defined inside `lsp-mode` and configured once, I prefer to calling `pyenv` on our side instead of on users side.

How do you think about it?

yyoncho: Can you elaborate? I suggested the variable because the logic for automatic calculation might fail and in this case IMO it will be useful for the user to be able to set it manually. Also, the user might configure it using dir locals per project.

15cm: I totally agree that users should be able to manually set the environment and added `lsp-pyls-plugins-jedi-environment` for this purpose. What I want to discuss is the "automatic" part, i.e. do we want to find the environment by `pyenv` in lsp-pyls or it should be the users responsiblity to do that and set the `lsp-pyls-plugins-jedi-environment` themselves. My concerns and opinions have been stated in the previous comment.

yyoncho: I do not have enough python knowledge and I do not know what is the typical workflow is and what is the expected interaction `lsp-mode - pyenv - pyls`. I will let you and @mpanarin discuss it. From my point of view, the lsp-mode behaviour should be expected and as much stuff as possible should be working out of the box. For example, AFAIK pyls must be installed in the pyenv of the current project - I would like this to happen automatically - lsp-mode should detect that pyls is missing and offer to install it.

yyoncho: > As the workspace is defined inside lsp-mode and configured once, I prefer to calling pyenv on our side instead of on users side.

I don't understand here what do you mean by "users side"?

15cm: The approach I took would let stuff works out of the box. Let me discuss with @mpanarin since he have more context about how `plys` works. Thanks for your follow-on.

lsp-pyls.el
```diff
@@ -353,53 +359,84 @@ parameters referenced in config."
...
+           ("pyls.plugins.jedi_references.enabled" lsp-pyls-plugins-jedi-references-enabled t)
+           ("pyls.plugins.jedi_hover.enabled" lsp-pyls-plugins-jedi-hover-enabled t)
+           ("pyls.plugins.jedi_definition.follow_builtin_imports" lsp-pyls-plugins-jedi-definition-follow-builtin-imports t)
+           ("pyls.plugins.jedi_definition.follow_imports" lsp-pyls-plugins-jedi-definition-follow-imports t)
+           ("pyls.plugins.jedi_definition.enabled" lsp-pyls-plugins-jedi-definition-enabled t)
+           ("pyls.plugins.jedi_completion.include_params" lsp-pyls-plugins-jedi-completion-include-params t)
+           ("pyls.plugins.jedi_completion.enabled" lsp-pyls-plugins-jedi-completion-enabled t)
+           ("pyls.configurationSources" lsp-pyls-configuration-sources)
+           ("pyls.plugins.jedi.environment"
+            (lambda ()
```
mpanarin: Can this lambda be turned into a separate function to allow advising/overriding?

I am not sure if this is a good idea on project/elisp side of things @yyoncho .

yyoncho: You could do 

``` emacs-lisp
(with-eval-after-load 'lsp-pyls 
  (lsp-register-custom-settings (("pyls.plugins.jedi.environment" . my-function) ...
```

FWIW advising any public function is acceptable from elisp perspective.

lsp-pyls.el
```diff
@@ -353,53 +359,84 @@ parameters referenced in config."
   :group 'lsp-pyls
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-pyls-plugins-jedi-use-pyenv-environment t
```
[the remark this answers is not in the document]
15cm: `pyenv` only manages python versions. For project management, you probably use things like `poetry` or `pipenv`  and combined them with `pyenv virtualenv`(that's my workflow though). So having `pyenv` installed doesn't mean the user would definitely like to use it in pyls. I think it's necessary to keep this flag to leave the control for users.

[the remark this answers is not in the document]
mpanarin: What I mean, is that you either manage all of your python versions with `pyenv` (through `.python_version` for project of through `pyenv global`) or you don't use `pyenv` at all. This is different from `virtualenvs`.

In my opinion, virtualenvs should not be managed  in lsp-mode. If you use virtualenvs, there are great emacs packages for them, that set `exec-path` properly. So, imo, the flow should be like this:
1. use `lsp-pyls-plugins-jedi-environment` if it is set.
2. Check if we are in virtualenv by `(getenv "VIRTUAL_ENV")` if yes -> send nil
3. Check if pyenv is present. If yes -> send it
4. In other situations send nil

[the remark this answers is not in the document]
mpanarin: Also, currently you don't check that pyenv is actually installed.

If `lsp-pyls-plugins-jedi-use-pyenv-environment` is `t`, and it is by default, user without pyenv installed will get initialization errors here: https://github.com/emacs-lsp/lsp-mode/pull/1393/files#diff-ce06d48fb3afc95bf7eafe4b4a9f2b11R434

lsp-pyls.el
15cm: Added function support for values in lsp-client-settings and extracted the lambda to a separate function for "pyls.plugins.jedi.environment".

lsp-pyls.el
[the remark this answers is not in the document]
15cm: I agree with that if `pyenv` is installed then we should at least fallback to `pyenv global`.

I didn't mentioned `virtualenvs` but https://github.com/pyenv/pyenv-virtualenv. I am not familiar how `VIRTUAL_ENV` is set by other emacs packages so I stick with getting the current environment by `pyenv`.

Then for the `exec-path`, it won't help in a use case, which is the goal of this PR: `pyls` is installed globally instead of per python version. Different python environments will be passed into it when executing this global `pyls`. So `exec-path` won't help as the `pyls` is always like `/usr/bin/pyls` and that's why `pyls.plugins.jedi.environment` comes.

I'd like to propose a flow like this(only change the 2nd point):
1. use `lsp-pyls-plugins-jedi-environment` if it is set.
~~2. Check if we are in virtualenv by (getenv "VIRTUAL_ENV") if yes -> **send it**~~(Out of the scope of this PR. Leave if to people who knows how `VIRTUAL_ENV` is set per project in Emacs)
3. Check if pyenv is present. If yes -> send it
4. In other situations send nil

Added the check to see if `pyenv` is installed. Thanks for pointing out.(I assumed that when `lsp-pyls-plugins-jedi-use-pyenv-environment` is `t` user should have `pyenv` installed but forgot it's set to `t` by default.)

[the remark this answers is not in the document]
mpanarin: Will pyenv use its virtualenv by default?

If you don't set `lsp-pyls-plugins-jedi-environment` and you have `pyenv-virtualenv` with pyls installed inside, will this work?
If no, this is a problem, as it will break existing user configurations.
"VIRTUAL_ENV" is set by `virtualenv`, not by specific emacs packages. If `pyenv-virtualenv` wraps around it, it sets it as well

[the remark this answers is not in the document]
15cm: 1. `pyenv` won't use `virtualenv`.
2. `pyenv-virtualenv` is managed by `pyenv`, like `pyenv global`, `.python_version`, etc.
3. I meant the `VIRTUAL_ENV` of the shell env inside Emacs.

Let's say we have a work flow like this:
1. We have only one pyls binary and `lsp-pyls-server-command` is set to it, i.e. `/usr/bin/pyls`
2. We have two project `proj1` and `proj2` and have `.python_version` specifying `2.7` and `3.7` respectively.(If using `pyenv-virtualenv`, the only difference is `.python_version` will be `env1` and `env2`)
3. We open to these projects in Emacs and call `lsp` on after another

What we expect is that the `.python_version` is regarded for each project automatically and passed to pyls as `pyls.plugins.jedi.environment`. Shell ENV won't help with this. If we want to pass the ENV as `pyls.plugins.jedi.environment`, the ENV need to be set automatically by an Emacs package when we switch buffer. Such packages that I've played with do things manually.

In summary, considering that users have their own ways to config `pyls`, such as:
Write some script to dynamically config `lsp-pyls-server-command`, according to any information from such as the buffer, `(shell-command-to-string "pyenv which)`", `PYENV_VERSION` or `VIRTUAL_ENV` that's populated by https://github.com/pythonic-emacs/pyenv-mode or whatever. It's hard to predict the intended behavior.

Hence I'd like to protect this PR by the `lsp-pyls-plugins-jedi-environment` flag and set it to `nil` by default. Athough I think it's a good way to detect the environment if `pyenv` is installed, let's leave it for users to enable. Python environment management is such a mess. I don't want to think of all the cases that the users want to hack on it.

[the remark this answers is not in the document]
15cm: The flow with latest commit will be:
```python
if lsp-pyls-plugins-jedi-environment is set:
  send `pyls.plugins.jedi.environment`
else if lsp-pyls-plugins-jedi-environment is enabled:
  send `pyls.plugins.jedi.environment`
else:
  do nothing and the behavior won't change
```

PR #1394, opened 2020-02-08
Add missing paren to which-key config

credmp:
The documentation is one paren short in the declaration of use-package for which-key. Users who copy this config will have a parse error in their configuration.


yyoncho:
Thank you, good catch!

PR #1398, opened 2020-02-09
Use lsp-workspace-status instead of custom variable

kurnevsky:
Turned out there is a special function in `lsp-mode` to report current status, and it works across multiple workspaces.

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -282,20 +282,15 @@ is often the type local variable declaration."
    ("rust.target" lsp-rust-target)
    ("rust.sysroot" lsp-rust-sysroot)))
 
-(defvar lsp-clients-rust-progress-string ""
-  "Rust progress status as reported by the RLS server.")
-
-(put 'lsp-rust-progress-string 'risky-local-variable t)
-(add-to-list 'global-mode-string (list '(t lsp-clients-rust-progress-string)))
-
-(defun lsp-clients--rust-window-progress (_workspace params)
+(defun lsp-clients--rust-window-progress (workspace params)
   "Progress report handling.
 PARAMS progress report notification data."
   (-let (((&hash "done" "message" "title") params))
     (if (or done (s-blank-str? message))
-        (setq lsp-clients-rust-progress-string nil)
-      (setq lsp-clients-rust-progress-string (format "%s - %s" title (or message "")))
-      (lsp-log lsp-clients-rust-progress-string))))
+        (lsp-workspace-status nil workspace)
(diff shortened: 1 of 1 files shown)


lsp-rust.el
```diff
@@ -282,20 +282,15 @@ is often the type local variable declaration."
...
-(defun lsp-clients--rust-window-progress (_workspace params)
+(defun lsp-clients--rust-window-progress (workspace params)
   "Progress report handling.
 PARAMS progress report notification data."
   (-let (((&hash "done" "message" "title") params))
     (if (or done (s-blank-str? message))
-        (setq lsp-clients-rust-progress-string nil)
-      (setq lsp-clients-rust-progress-string (format "%s - %s" title (or message "")))
-      (lsp-log lsp-clients-rust-progress-string))))
+      (lsp-workspace-status nil workspace)
```
brotzeit: The line isn't indented.

kurnevsky: It actually is, with `indent-region`. It places `then` and `else` blocks on the same level. Or what what indentation do you use?

brotzeit: When I run an indent function on this function I get this:

```
(defun lsp-clients--rust-window-progress (workspace params)
  "Progress report handling.
PARAMS progress report notification data."
  (-let (((&hash "done" "message" "title") params))
    (if (or done (s-blank-str? message))
        (lsp-workspace-status nil workspace)
      (let ((status-string (format "%s - %s" title (or message ""))))
        (lsp-workspace-status status-string workspace)
        (lsp-log status-string)))))
```

kurnevsky: Is it `indent-region` (C-M-\) or something else? Did you configure it somehow?

kurnevsky: Hm, interesting, with `emacs -Q` I get:
```
(defun lsp-clients--rust-window-progress (workspace params)
  "Progress report handling.
PARAMS progress report notification data."
  (-let (((&hash "done" "message" "title") params))
        (if (or done (s-blank-str? message))
	    (lsp-workspace-status nil workspace)
	  (let ((status-string (format "%s - %s" title (or message ""))))
	    (lsp-workspace-status status-string workspace)
	    (lsp-log status-string)))))
```

kurnevsky: ok, it seems `dash.el` has it's own rules which are ignored with `emacs -Q`, and your option is correct, so changed it.

brotzeit: Interesting =)

PR #1404, opened 2020-02-11
Ignore .ccls-cache for file watching

bennofs:
.ccls-cache is an internal directory created by the ccls lang server.


yyoncho:
Thank you!

PR #1406, opened 2020-02-11
Add clang-tidy specific Flycheck error explainer for the lsp checker

danielmartin:
New feature for C++ language servers that emit clang-tidy messages (like Clangd).

Press e on a clang-tidy error and you'll see an explanation from the LLVM website, so that you can learn about the error message before applying a fix-it (if available).

**Screenshot:**

clangd-clang-tidy-integration


lsp-clients.el
```diff
@@ -51,6 +51,7 @@
 (require 'lsp-json)
 (require 'lsp-verilog)
 (require 'lsp-eslint)
+(require 'lsp-cpp)
```
yyoncho: do we need this require when we are autoloading the lsp-ccp entryfunction?

yyoncho:
Thank you!

Issue #1240, opened 2019-12-10
Intermittent connection in typescript-language-server via Docker-tramp


MokkeMeguru:
I want to use typescript-language-server in docker's project via tramp-mode.

So I try open docker container in eshell and evaluate this function.
```
@598d6202b669 /docker:598d6202b669:/usr/app λ (lsp-register-client
 (make-lsp-client :new-connection
                  (lsp-tramp-connection "typescript-language-server --stdio")
                  :major-modes '(typescript-mode)
                  :remote? t
                  :server-id 'typescript-remote))
```
And also, I installed typescript-language-server typescript in the docker.

When I open container's file with lsp-mode, but I got an error message "Server typescript-remote:32228 exited with status exit. Do you want to restart it? (y/n)" **every few** seconds.

I can get some lsp's refactoring between **the few** seconds.

So I want to know how to keep a lsp's connection.

Thanks!

yyoncho:
There might be another problem but `"typescript-language-server --stdio"` should be `(list "typescript-language-server"  "--stdio")`.

MokkeMeguru:
No. lsp-tramp-connection's argument is below
```
commands & optional GENERATE-ERROR-FILE-FN
```

MokkeMeguru:
if we run this command instead of previous one, we receive same error message at soon.
```
(lsp-register-client
 (make-lsp-client :new-connection
                  (lsp-tramp-connection "typescript-language-server")
                  :major-modes '(typescript-mode)
                  :priority 2
                  :remote? t
                  :server-id 'typescript-remote))
```

I know this tramp connection is succeed in pyls. (https://qiita.com/MeguruMokke/items/2188bd32b253233bdf73 sorry I wrote this article in Japanese... ) .

If you run pyls in command line,  you found pyls is running something like a server.
So I think typescript-language-server will be running as something like a server too.

Let's run typescript-language-server in command line.

```
$ npm install typescript-language-server typescript
$ typescript-language-server
Connection type required (stdio, node-ipc, socket). Refer to --help for more details.
```

So we should use **stdio** or **node-ipc** or **socket**

I think there are two solution to connect lsp server
- **stdio**
this is my solution. I think we can connect via tramp.
- **socket**
I don't know how to set the connection port number.

yyoncho:
As I already mentioned, the params must be list, not string when the command has arguments, please test this one:

```
(lsp-register-client
[...]

PR #1408, opened 2020-02-12
sending nil pid

rnikoopour:
Language servers who receive a PID check the PID.  If it's no longer running then the language server exits.

Sending the PID works when all language servers are run directly on the host.  However, when using `lsp-docker` sending emacs PID causes certain servers to exit since the emacs PID doesn't exist on the container.

The spec allows for this to be null.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6141,7 +6141,7 @@ SESSION is the active session."
       (lsp-request-async
        "initialize"
        (append
-        (list :processId (emacs-pid)
+        (list :processId nil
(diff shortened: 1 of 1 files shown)


kiennq:
Out of curious, why we want to send nil `pid`?

yyoncho:
@kiennq fixes https://github.com/emacs-lsp/lsp-mode/issues/1240 . We had a long discussion in gitter chat.

yyoncho:
Thank you for contributing to `lsp-mode` and congrats for your first PR in the repo! :tada:

PR #1415, opened 2020-02-13
Quote status returned by server.

kurnevsky:
Server can return % in the status string which lead to errors when
displaying it.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2447,7 +2447,8 @@ BINDINGS is a list of (key def cond)."
 (defun lsp-workspace-status (status-string &optional workspace)
   "Set current workspace status to STATUS-STRING.
 If WORKSPACE is not specified defaults to lsp--cur-workspace."
-  (setf (lsp--workspace-status-string (or workspace lsp--cur-workspace)) status-string))
+  (let ((status-string (when status-string (replace-regexp-in-string "%" "%%" status-string))))
+    (setf (lsp--workspace-status-string (or workspace lsp--cur-workspace)) status-string)))
 
 (defun lsp-session-set-metadata (key value &optional _workspace)
   "Associate KEY with VALUE in the WORKSPACE metadata.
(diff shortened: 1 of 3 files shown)


kurnevsky:
@yyoncho probably you should do the same for lsp-java:
https://github.com/emacs-lsp/lsp-java/blob/dbeeee9c74db25db217059011935b882f7ec1257/lsp-java.el#L513
https://github.com/emacs-lsp/lsp-java/blob/dbeeee9c74db25db217059011935b882f7ec1257/lsp-java.el#L673

yyoncho:
I suspect that this is an issue in your modeline - see https://github.com/seagle0128/doom-modeline/issues/121

kurnevsky:
I have the same behavior with standard modeline.

yyoncho:
Which emacs version do you use? When I do `emacs -q` and then eval the following in the scratch:

```
(defcustom test-var2 "Test %" "")
(add-to-list 'global-mode-string (list '(t test-var2)))
(setq test-var2 "This % is displayed")
```

The `%` is displayed.

kurnevsky:
It's `GNU Emacs 26.3`. Hm, interesting, your code works but if I set it with plain string like `(add-to-list 'global-mode-string (list '(t "a%va")))` it doesn't...

kurnevsky:
ok, got it. It stops working if you use eval like this (which lsp-mode does):
```
(defcustom test-var2 "Test %" "")
(add-to-list 'global-mode-string (list '(t (:eval test-var2))))
(setq test-var2 "This % is displayed")
```

lsp-mode.el
```diff
@@ -2447,7 +2447,8 @@ BINDINGS is a list of (key def cond)."
 (defun lsp-workspace-status (status-string &optional workspace)
   "Set current workspace status to STATUS-STRING.
 If WORKSPACE is not specified defaults to lsp--cur-workspace."
-  (setf (lsp--workspace-status-string (or workspace lsp--cur-workspace)) status-string))
+  (let ((status-string (if status-string (replace-regexp-in-string "%" "%%" status-string) nil)))
```
yyoncho: nit: use when instead of if when then clause is nil.

kurnevsky: fixed

PR #1418, opened 2020-02-15
Change typescript activation function

rnikoopour:
Because javascript and typescript used the same activation function it would only activate the javascript language server. 

This differentiates typescript and javascript activation functions.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -166,17 +166,17 @@ See instructions at https://marketplace.visualstudio.com/items?itemName=mads-har
   :risky t
   :type '(repeat string))
 
-(defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
+(defun lsp-typescript-javascript-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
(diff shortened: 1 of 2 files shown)


kiennq:
Since `javascript-typescript-language-server` and `typescript-language-server` support both `javascript` and `typescript`, why we need to differentiate between them?
Also, if you could keep the `rx` syntax, it would be great, the `rx` syntax is easier to see at a glance than the full-blown regex

rnikoopour:
When speaking with @yyoncho I shared the following with him related to `lsp-docker`:

```(make-client :server-id 'jsts-ls :docker-server-id 'jstsls-docker :server-command "typescript-language-server --stdio")```

He responded by saying this used the wrong server.  When trying to fix that I had to make the change to `lsp-mode` to make sure that javascript used the javascript server and typescript used the typescript server.

rnikoopour:
Going to close this for now.

yyoncho:
> When speaking with @yyoncho I shared the following with him related to `lsp-docker`:
> 
> `(make-client :server-id 'jsts-ls :docker-server-id 'jstsls-docker :server-command "typescript-language-server --stdio")`

I meant that you use the `ts-ls` binary for `jsts-ls` server in your config.

Issue #1412, opened 2020-02-13
Flycheck checking all buffer


gjm:
**Describe the bug**
When `flycheck` activates it checks all open buffers and not just the one being edited.

**To Reproduce**
1. Open several buffers 
2. Edit one of them
3. See checker processes for each of the open buffers

**Expected behavior**
Should only trigger checker processes for the edited buffer.

**Which Language Server did you use**
PHP with intelephense language server.

**OS**
Ubuntu 19.10

PR #1419, opened 2020-02-15
Improve lsp-mode completion performance

yyoncho:
ATM lsp-mode completion slows down sometimes because we receive diagnostics or
related results while we are performing completion, as a result, emacs spends
time in:
* Flycheck
* Sideline (see https://github.com/emacs-lsp/lsp-ui/issues/398)
* Lenses
* Links

This change inhibits all of the actions that are performed on idle hook and also
disables the flycheck during the completion.

Fixes https://github.com/emacs-lsp/lsp-mode/issues/1412

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -197,7 +197,7 @@ occasionally break as language servers are updated."
 
 (defvar-local lsp--cur-workspace nil)
 
-(defvar-local lsp--cur-version nil)
+(defvar-local lsp--cur-version 0)
 
 (defvar lsp--uri-file-prefix (pcase system-type
                                (`windows-nt "file:///")
@@ -1417,14 +1417,6 @@ already have been created."
   (declare (debug (form body))
            (indent 1))
   `(when-let (lsp--cur-workspace ,workspace) ,@body))
-
-(defmacro lsp--with-full-sync (&rest body)
-  "Execute BODY with lsp--server-sync-method set to 'full."
-  (declare (debug (form body))
-           (indent 1))
-  `(let ((lsp-document-sync-method lsp--sync-full))
-     ,@body
-     (lsp--flush-delayed-changes)))
 
 (defun lsp-canonical-file-name  (file-name)
   "Return the canonical FILE-NAME."
@@ -3181,7 +3173,9 @@ in that particular folder."
 
       (setq-local global-mode-string (if (-contains? global-mode-string status)
                                          global-mode-string
-                                       (cons status global-mode-string))))
+                                       (cons status global-mode-string)))
+      (when (bound-and-true-p company-mode)
+        (lsp--setup-company)))
      (t
       (setq-local indent-region-function nil)
       (remove-function (local 'eldoc-documentation-function) #'lsp-eldoc-function)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -6005,9 +5979,28 @@ returns the command to execute."
                        (cons proc proc))))
         :test? (lambda () (-> local-command lsp-resolve-final-function lsp-server-present?))))
 
+(defun lsp--setup-company ()
+  (company-mode 1)
+  (add-hook 'company-completion-started-hook
```
kiennq: Can all these hooks be performed with `(with-eval-afterload)`?
Reason: I'm not using `lsp--auto-configure` and have more sophisticated company backends configuration to handle capf.
Moreover, the hook should be added whether auto configured is set or not.

yyoncho: Isn't `(add-hook 'lsp-managed-mode-hook 'lsp--setup-company)` sufficient to cover this case?

In general, executing stuff on the top level is considered a flaw so I want to not have it where possible.

seagle0128: It's not a good idea to use `eval-after-load` in the package.

kiennq: Yes, the `add-hook` is enough to cover that case. However, are we asking the user to additionally add hook to a private `lsp--setup-company` function? That's kind of breaking change for user who don't use auto configure.
Instead of executing stuff on the top level, perhaps we can add the `with-eval-afterload` to minor mode initialization function?
The part of enable `(company-mode 1)` can stay in its places before this change though.

kiennq: Instead of usinig `eval-after-load` forward declaring hook and add hook immediately works for me too.

yyoncho: > However, are we asking the user to additionally add hook to a private lsp--setup-company function?

 We can reorganize and make each of the functions from the auto configure public and reusable. E. g. lsp-setup-flycheck, lsp-setup-company so the user could pick only the functions that he/she wants. 

> Instead of usinig eval-after-load forward declaring hook and add hook immediately works for me too.

This means that we will set these hooks for all modes. Although our handlers are very lightweight it doesn't feel right to have them unconditionally added.

kiennq: > We can reorganize and make each of the functions from the auto configure public and reusable. E. g. lsp-setup-flycheck, lsp-setup-company so the user could pick only the functions that he/she wants.

That would be great. Albeit I still need to modify my config file for `lsp-mode` to work again. In case if user is not aware of those setup function, does that means `lsp-mode` is working incorrectly? Do we want that?

> > Instead of usinig eval-after-load forward declaring hook and add hook immediately works for me too.
> 
> This means that we will set these hooks for all modes. Although our handlers are very lightweight it doesn't feel right to have them unconditionally added.

I mean we can add hook immediately without waiting for the feature to be loaded. We can add hook in mode enable function and clean it up in mode disable function. That will ensure that the hook will only affect when `lsp-mode` is running.

yyoncho: > I mean we can add hook immediately without waiting for the feature to be loaded. We can add hook in mode enable function and clean it up in mode disable function. That will ensure that the hook will only affect when `lsp-mode` is running.

This sounds ok - I will add them directly to the managed mode hook to make sure they are executed even with autoconfigure disabled. But generally, I consider disabling autoconfigure options to be only for the most advanced users/contributors which follow lsp-mode development.

kiennq: > But generally, I consider disabling autoconfigure options to be only for the most advanced users/contributors which follow lsp-mode development.

I see. I've tried `lsp-auto-configure` before in the hope of not getting burn if there're breaking change in `lsp-mode` that require my config change to get it works.
However I found out my `companymode` setup is become broken instead.

Should we add another custom `lsp-enable-company-integration` to help auto configure to automatically changing company mode configuration or not?

yyoncho: > Should we add another custom `lsp-enable-company-integration` to help auto configure to automatically changing company mode configuration or not?

I don't want to go in that direction - I will follow your advice and use the managed mode hook unconditionally since we need this to make company work.

yyoncho: I pushed the new version - moved the setup company in managed mode.

kiennq:
LGTM

PR #1421, opened 2020-02-15
Added lsp-after-client-package-hook

rnikoopour:
`lsp-docker` hooked `lsp-clients`.  `lsp-docker` needs a hook after the clients are loaded.  This PR adds a new hook that `lsp-docker` can use.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -296,6 +296,11 @@ When nil, all registered clients are considered candidates.")
   :type 'hook
   :group 'lsp-mode)
 
+(defcustom lsp-after-client-packages-hook nil
+  "List of functions to be called after client packages are opened."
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -6982,6 +6987,7 @@ argument ask the user to select which language server to start. "
   (when (and lsp-auto-configure)
     (seq-do (lambda (package) (require package nil t))
             lsp-client-packages))
+  (run-hooks 'lsp-after-client-packages-hook)
```
yyoncho: that would run every time you run M-x lsp. So we should put in the `when` and also make sure that the require block is called only once.

rnikoopour:
closing per discussion with @yyoncho on gitter

PR #1423, opened 2020-02-16
flymake-backend: Re-report diagnostics using new flymake report-fn.

vibhavp:
This fixes an edge case bug in where sometimes newer diagnostics are
reported using an older version of the report function, using
"zombie" diags on the buffer.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1554,10 +1554,8 @@ WORKSPACE is the workspace that contains the diagnostics."
 
     (defun lsp--flymake-backend (report-fn &rest _args)
       "Flymake backend."
-      (let ((first-run (null lsp--flymake-report-fn)))
-        (setq lsp--flymake-report-fn report-fn)
-        (when first-run
-          (lsp--flymake-update-diagnostics))))
+      (setq lsp--flymake-report-fn report-fn)
+      (lsp--flymake-update-diagnostics))
 
     (defun lsp--flymake-update-diagnostics ()
       "Report new diagnostics to flymake."


yyoncho:
Can you explain how this works? AFACS the only difference between the new code and the old code is that the old code wasn't calling the flymake function only if it was the first flymake call which will avoid re-reporting the diagnostics if they are already reported.

vibhavp:
Sorry for the late followup, but I've observed this issue with gopls, where the language server may at times send diagnostics before flymake changes the reporting function for a particular edit. In this case, new diagnostics end up being not reported at all, and the buffer still reports the older errors.

yyoncho:
Which flymake version do you use? I think that the older versions do not work well with lsp-mode.

vibhavp:
It's the one bundled with Emacs 26.3. That should be the latest stable version, right?

yyoncho:
No, there is newer in elpa.

> Real-time Diagnostics/linting (via flymake when Emacs > 26, requires flymake>=1.0.5 

I tried make that hard dependency in Required-Packages but it breaks the 25.* clients since the new flymake requires emacs26+.

vibhavp:
Ah, makes sense. Lets merge this to fix this issue for older flymake versions, then?

yyoncho:
I am not 100% sure if this fixes the issue. Current code is:
```
if this is first run -> report what we have.
else -> store the handler and wait lsp-mode to report diagnostics.
```
after the change the code is:

```
report diagnostics right away even if they are already reported.
```

I am not the author of this code (@hpdeifel did it) but AFAICS the second variant will read to redundant diagnostics refresh calls.

hpdeifel:
Hi! It's been a while since I've looked at the code, but the current behaviour should be the correct one for current flymake versions (eglot also does the same). It would be good to know which flymake version fixes the issue you experience and then maybe find a more specific work around that doesn't affect users of newer flymake versions.

Also, it's very unfortunate that `lsp-mode` can't declare the correct flymake dependency, since this results in a buggy default experience for all users who don't read the readme very carefully.

Issue #750, opened 2019-04-07
[Feature] create modeline indicator for buffer/project/workspace errors statistics


yyoncho:
We already do that in lsp-treemacs but it would be nice if we have an overview of the errors in the modeline which is visible all of the time.

![Selection_123](https://user-images.githubusercontent.com/13259670/55678279-a6366c00-58ff-11e9-8172-81fd4dcd944d.png)

Here it is what vscode has (by default it lists the errors/warnings/info for all projects in the workspace)
![Selection_124](https://user-images.githubusercontent.com/13259670/55678387-72f4dc80-5901-11e9-8e4d-637a710280a5.png)

dangom:
Isn't this what flymake/flycheck do already?

yyoncho:
@dangom AFAIK they display the information only for the current buffer since they do not have the notion of what project is. Also none of them can list the errors for files that are not opened.

Issue #382, opened 2018-06-18
Add command for next/previous workspace error


yyoncho:
Investigate whether this could be accomplished using Flycheck multibuffer support and implement it on LSP side if it is not possible.

PR #1425, opened 2020-02-16
Show flycheck errors from other files

mrBliss:
If there are no errors in the current buffer, it might be because a dependent file/buffer contains errors. Instead of not showing any errors, show errors (but not warnings/info) from other files. This is supported by flycheck, see https://github.com/flycheck/flycheck/pull/1427.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7109,15 +7109,13 @@ reported according to `flycheck-check-syntax-automatically'."
   "Start an LSP syntax check with CHECKER.
 
 CALLBACK is the status callback passed by Flycheck."
-  (->> (or (gethash (lsp--fix-path-casing buffer-file-name)
-                    (lsp-diagnostics))
-           (gethash (lsp--fix-path-casing (file-truename buffer-file-name))
-                    (lsp-diagnostics)))
-       (-map (-lambda (diag)
+  (cl-labels ((make-flycheck-error
+               (file diag)
+               ;; when FILE is nil, use the current buffer
                (flycheck-error-new
-                :buffer (current-buffer)
+                :buffer (if file (find-file-noselect file) (current-buffer))
                 :checker checker
-                :filename buffer-file-name
+                :filename (or file buffer-file-name)
                 :line (1+ (lsp-diagnostic-line diag))
                 :column (1+ (lsp-diagnostic-column diag))
                 :message (lsp-diagnostic-message diag)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7109,15 +7109,13 @@ reported according to `flycheck-check-syntax-automatically'."
...
-                    (lsp-diagnostics))
-           (gethash (lsp--fix-path-casing (file-truename buffer-file-name))
-                    (lsp-diagnostics)))
-       (-map (-lambda (diag)
+  (cl-labels ((make-flycheck-error
+               (file diag)
+               ;; when FILE is nil, use the current buffer
                (flycheck-error-new
-                :buffer (current-buffer)
+                :buffer (if file (find-file-noselect file) (current-buffer))
```
yyoncho: That would mean we open all files for which we have received diagnostics which is no-go for performance reasons(imagine a legacy project with 1000+ files with warnings).

mrBliss:
> Thank you for the PR. Unfortunately, it is not that simple and we were doing that in the past and we had to revert it back for performance reasons(see comment). The alternative approach is to implement:
> 
> #750
> #382
> 
> See also:
> [flycheck/flycheck#1631](https://github.com/flycheck/flycheck/issues/1631)

Ah, I was already thinking that I didn't search hard enough for issues related to this. Thanks for the pointers!

lsp-mode.el
mrBliss: Oh, I have been too long in Haskell land, I was thinking it would be a thunk or a monadic action (deferred computation in this case) :slightly_smiling_face:.

This function would only be called for *errors* in other buffers, not warnings.

Yeah, I wrote this with Haskell in mind, where the compiler typically stops when it encounters an error and doesn't move on to the next file. So I'd never run the risk of getting tons of errors in other buffers. In my workflow, you fix errors one by one, buffer per buffer, starting from the top of the module dependency graph. So you nearly always would want to switch to the other buffer containing errors.

yyoncho:
We could still achieve what you want avoiding the perf issue if we open only the first file with error but IMO this will be just a workaround for the lack of #750 . This is what VScode has. Let me know if you want to work on #750 - there is stale PR in progress - https://github.com/emacs-lsp/lsp-mode/pull/807

lsp-mode.el
yyoncho: I had similar issues with Java and I created lsp-treemacs-errors-list. The general issue is that both flycheck and flymake are designed after single file liters and they don't have the notion of a project which does not fit in lsp world.

PR #1426, opened 2020-02-16
Make sure signature is automatically cleaned and a bit of refactoring

kiennq:
## Issue
When signature is displayed and user switch to another buffer which doesn't managed by `lsp-mode`, initially the signature is not being cleaned (1)
After that if user is switching back to previous buffer, the signature will be displayed again. However after that, the signature is not **being automatically cleaned**.
The reason is that when switching  to other buffer, the `lsp-signature-stop` will remove the `lsp-signature` from `lsp-on-idle-hook` **locally**, which is a wrong removal. It should remove from `lsp-on-idle-hook` of the last buffer with signature.

## What changed
- [x] Remove `lsp-signature` from correct buffer
- [x] Rename some to `lsp-signature` to have more consistent naming
- [x] `setq-local` for package `defvar` is bad, since once the `setq-local` is called, the variable effectively become local value. The subsequence call using `setq` will just set the local value.
  I found out that `lsp-signature--last-buffer` is being the victim of that call pattern. Which explains (1). So I think we should probably refactor to use `setq` for the variable that's `defvar-xxx` in this package. With that we will ensure all global variable with `defvar` and local variable with `defvar-local` works as intended.

I've separated to two separated commit for easier to review.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1184,8 +1184,8 @@ to the beginning and ending points in the range correspondingly."
                      "textDocument/selectionRange"
                      (list :textDocument (lsp--text-document-identifier)
                            :positions (vector (lsp--cur-position)))))))
-     (setq-local lsp--document-selection-range-cache
-                 (cons response (buffer-modified-tick)))
+     (setq lsp--document-selection-range-cache
+           (cons response (buffer-modified-tick)))
      response)))
 
 (defun lsp-extend-selection ()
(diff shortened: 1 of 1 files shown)


yyoncho:
Looks good to me! (Although I am not sure the rename lsp--signature- to lsp-signature-- fits with elisp coding guides and also makes it inconsistent with the rest of the code which does not follow that convention.)

I will try to do manual testing tomorrow and proceed with the merge.

kiennq:
> Although I am not sure the rename lsp--signature- to lsp-signature-- fits with elisp coding guides and also makes it inconsistent with the rest of the code which does not follow that convention

Thanks, I will revert back from `lsp-signature--` to `lsp--signature-`.

yyoncho:
Thank you!

PR #1428, opened 2020-02-17
Add Elixir to supported language table in README

wtfleming:
https://github.com/JakeBecker/elixir-ls has become inactive for an extended period of time. https://github.com/elixir-lsp/elixir-ls is an active fork that has merged in pull requests and fixed issues


yyoncho:
Thank you!

PR #1429, opened 2020-02-17
Do not call the server if additionalTextEdits are present

yyoncho:
- as per spec textEdit/insertTextFormat and insertText are mandatory so we have
to call resolve only if additionalTextEdits are not present.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4065,7 +4065,9 @@ PLIST is the additional data to attach to each candidate."
                          "textEdit" text-edit
                          "insertTextFormat" insert-text-format
                          "additionalTextEdits" additional-text-edits)
-                  (lsp--resolve-completion item)))
+                  (if (gethash "additionalTextEdits" item)
+                      item
+                    (lsp--resolve-completion item))))
            (cond
             (text-edit
              (delete-region (point-at-bol) (point))


lsp-mode.el
```diff
@@ -4065,7 +4065,9 @@ PLIST is the additional data to attach to each candidate."
                          "textEdit" text-edit
                          "insertTextFormat" insert-text-format
                          "additionalTextEdits" additional-text-edits)
-                  (lsp--resolve-completion item)))
+                  (if (gethash "additionalTextEdits" item)
```
kiennq: Just asking, which language server that requires `completionItem/resolve` for `additionalTextEdits`?
From the specs, I saw that only `documentation` and `detail` are subjects of delay computation, can we perhaps drop the `lsp--resolve-completion` in this case entirely?

kiennq: Otherwise, the change LGTM

yyoncho: JDT LS was populating the additionalTextEdits in resolve call but they fixed that behavior lately. I checked the spec and it seems like additionalTextEdits are not mentioned. The spec says:

```
The returned completion item should have the documentation property filled in. The request can delay the computation of the detail and documentation properties. However, properties that are needed for the initial sorting and filtering, like sortText, filterText, insertText, and textEdit must be provided in the textDocument/completion response and must not be changed during resolve.
```

The issue is that additonalTextEdit is not mentioned in any of the lists. I will ask it the lsp repo for clarification. It will be great if we can drop that resolve call.

yyoncho:
I merged as it is since it is an improvement either way. Once https://github.com/microsoft/language-server-protocol/issues/924 is resolved we will decide what to do.

yyoncho:
Thank you @kiennq for bringing up this discussion - https://github.com/emacs-lsp/lsp-mode/commit/f6ec81b07039b5279bfd6dcd906e284d44f4a035 

In the past, JDT LS was not returning additionalTextEdits as part of the textDocument/completion result and we implemented the resolve as part of the insert procedure. FTR It works fine in vscode since vscode is automatically resolving the documentation of the selected item.

PR #1432, opened 2020-02-18
Omit unneeded calls to `remove-overlays`

sebastiansturm:
As discussed on gitter, this commit introduces a buffer-local flag to track whether document highlights (may) have been set. Checking this flag on `lsp--cleanup-highlights-if-needed` allows us to avoid unnecessary, and potentially expensive, calls to `remove-overlays`. Some superficial testing hasn't brought up any problems, but I'll keep using this and report back later today


yyoncho:
Thank you!

PR #1434, opened 2020-02-18
Hide code lenses before revert-buffer

zainab-ali:
Firstly, thank you for adding code lens support :slightly_smiling_face: 

I've found that the code lens overlays aren't removed when `revert-buffer` is called.  This results in the following odd behaviour : https://drive.google.com/file/d/1-kl5Cbi6ie2WBg7GoW0gCy2RkVwScMM7/view

When a buffer is reverted with `revert-buffer`, the `lsp--lens-overlays` variable is reset to it's default value of `nil`.  However, the code lens overlays it previously referred to still exist.  This PR calls `lsp-lens-hide` before a buffer is reverted, deleting any code lens overlays.

I'm not well versed in overlays, so am not sure if this is the best possible solution.  Please let me know if you think of a better approach.  I'm also not too sure where or how to add tests to this project - please feel free to point me to a place to do so.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2036,14 +2036,14 @@ CALLBACK - callback for the lenses."
     (add-hook 'lsp-on-idle-hook #'lsp--lens-idle-function nil t)
     (add-hook 'lsp-on-change-hook (lambda () (lsp--lens-schedule-refresh t)) nil t)
     (add-hook 'after-save-hook (lambda () (lsp--lens-schedule-refresh t)) nil t)
-    (add-hook 'before-revert-hook (lambda () (lsp-lens-hide)) nil t)
+    (add-hook 'before-revert-hook #'lsp-lens-hide nil t)
     (lsp-lens-refresh t))
    (t
     (lsp-lens-hide)
     (remove-hook 'lsp-on-idle-hook #'lsp--lens-idle-function t)
     (remove-hook 'lsp-on-change-hook (lambda () (lsp--lens-schedule-refresh nil)) t)
     (remove-hook 'after-save-hook (lambda () (lsp--lens-schedule-refresh t)) t)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -2036,12 +2036,14 @@ CALLBACK - callback for the lenses."
     (add-hook 'lsp-on-idle-hook #'lsp--lens-idle-function nil t)
     (add-hook 'lsp-on-change-hook (lambda () (lsp--lens-schedule-refresh t)) nil t)
     (add-hook 'after-save-hook (lambda () (lsp--lens-schedule-refresh t)) nil t)
+    (add-hook 'before-revert-hook (lambda () (lsp-lens-hide)) nil t)
```
yyoncho: nit: there is no need to wrap it in lambda, you could write directly `#'lsp-lens-hide`

lsp-mode.el
```diff
@@ -2036,12 +2036,14 @@ CALLBACK - callback for the lenses."
...
     (lsp-lens-refresh t))
    (t
     (lsp-lens-hide)
     (remove-hook 'lsp-on-idle-hook #'lsp--lens-idle-function t)
     (remove-hook 'lsp-on-change-hook (lambda () (lsp--lens-schedule-refresh nil)) t)
     (remove-hook 'after-save-hook (lambda () (lsp--lens-schedule-refresh t)) t)
+    (remove-hook 'before-revert-hook (lambda () (lsp-lens-hide)) nil t)
```
yyoncho: remove hook has fewer params.

yyoncho: see also https://travis-ci.org/emacs-lsp/lsp-mode/jobs/652126618#L1347

yyoncho:

PS: It is a good idea to ping after you push a change - it seems like GH doesn't send a notification all the time.

zainab-ali:
Thanks!  I'll keep that in mind

PR #1440, opened 2020-02-21
lsp--line-character-to-point: `character`  'nil?

tomterl:
Fixes https://github.com/emacs-lsp/lsp-origami/issues/5, but I don't know what this might implicate in other places.
In certain, but not all, php-files, ```lsp--line-character-to-point``` is called with ```character``` passed as ```nil```, don't use numerical comparison in this case.  Simply check that character is not nil prior to the comparison

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1144,7 +1144,7 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
       ;; server may send character position beyond the current line and we
       ;; should fallback to line end.
       (let ((line-end (line-end-position)))
-        (if (and (not (equal nil character)) (> character (- line-end (point))))
+        (if (and character (> character (- line-end (point))))
             line-end
           (forward-char character)
           (point))))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1144,7 +1144,7 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
       ;; server may send character position beyond the current line and we
       ;; should fallback to line end.
       (let ((line-end (line-end-position)))
-        (if (> character (- line-end (point)))
+        (if (and (not (equal nil character)) (> character (- line-end (point))))
```
yyoncho: could be simply (not character) instead of (not (equal nil character)). This shouldn't affect the other code since it is extending the contract.

tomterl: Hmm - it would have to be ```(not (not character))```, that's not that much better, is it? Let me check if simply ```character``` does work.

tomterl: It does - I'll mend the PR

lsp-mode.el
```diff
@@ -1144,7 +1144,7 @@ INHERIT-INPUT-METHOD will be proxied to `completing-read' without changes."
       ;; server may send character position beyond the current line and we
       ;; should fallback to line end.
       (let ((line-end (line-end-position)))
-        (if (> character (- line-end (point)))
```
yyoncho: I think should be `(or (not character) ...`

yyoncho:

PR #1445, opened 2020-02-21
lsp--capf: Improve highlights of matched characters in candidates

kiennq:
This fix a corner case of matches highlighting when the `prefix` contains some extra start characters compare to `label`.
In that case we shift to try to find match by cutoff some start characters of `prefix` until an highlight matches is found.


yyoncho:
Thank you!

PR #1446, opened 2020-02-23
update rust-analyzer binary

### LINKED ISSUE #1431: Rename binary to rust-analyzer
We need to change the default value of `lsp-rust-analyzer-server-command`(https://github.com/rust-analyzer/rust-analyzer/pull/3216).

brotzeit:
close #1431

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -318,7 +318,7 @@ PARAMS progress report notification data."
 
 ;; rust-analyzer
 
-(defcustom lsp-rust-analyzer-server-command '("ra_lsp_server")
+(defcustom lsp-rust-analyzer-server-command '("~/.cargo/bin/rust-analyzer")
   "Command to start rust-analyzer."
   :type '(repeat string)
   :package-version '(lsp-mode . "6.2"))


brotzeit:
I expect there will be issues by users who updated lsp-mode, but didn't update rust-analyzer.

yyoncho:
> I expect there will be issues by users who updated lsp-mode, but didn't update rust-analyzer.

IMO the only scenario we can support given the resources we have is the latest lsp-mode+latest version of the server(unless explicitly mentioned).

Is there one-liner for installing RA? If yes, can you put it in the readme?

brotzeit:
`git clone https://github.com/rust-analyzer/rust-analyzer && cd rust-analyzer && cargo xtask install --server`

Too long for the readme I guess ?

yyoncho:
> Too long for the readme I guess ?

Yes, let us focus on creating the automatic installation.

contrun:
We should not assume rust-analyzer lies in "~/.cargo/bin/rust-analyzer". We should assume "~/.cargo/bin" is in the PATH instead. People like me install rust-analyzer with our package manager.

brotzeit:
That was the value before, but I changed this since there were several issues about rust-analyzer not starting. I thought rust-analyzer is still not stable so it probably won't be in package repos yet. If you open a PR I will merge it.

How do we want the automatic installation to work ? Do we assume that users want to install the server from source or through a package manager ?

yyoncho:
> How do we want the automatic installation to work ? Do we assume that users want to install the server from source or through a package manager ?

The way it works in vscode.

contrun:
I created a new pr to change the path. https://github.com/emacs-lsp/lsp-mode/pull/1453


I think we may safely assume the users want to install the server from source as it does not super user privilege. Maybe we can change this line https://github.com/emacs-lsp/lsp-mode/blob/9982b7ed5d96eb46cc1de0501de3b66944b90ff4/lsp-rust.el#L304  to check also check the file under path `~/.cargo/bin/`.

PR #1453, opened 2020-02-25
update rust-analyzer path

contrun:


brotzeit:
For reference, I will update rustic as well as soon as the pending PRs are merged.

Issue #1451, opened 2020-02-24
lsp-prefer-capf doesnt work well


taigacute:
**Describe the bug**
Set the `lsp-prefer-capf` t,  It performs exact matching and does not perform filtering well.
Bug 1: I can get completion items , but when input `p` cant get any completion items.
![image](https://user-images.githubusercontent.com/41671631/75134708-0f8ebb80-571a-11ea-88df-b81f83e33104.png)
input `p`
![image](https://user-images.githubusercontent.com/41671631/75134717-14ec0600-571a-11ea-97eb-793c668223aa.png)

Bug 2: doesnt filter  for gopls.
![image](https://user-images.githubusercontent.com/41671631/75134786-495fc200-571a-11ea-94a7-5355ba550577.png)

**To Reproduce**
1. open a go file
2. use this lsp config and company config
```elisp
(use-package lsp-mode
  :commands (lsp-install-server lsp lsp-deferred)
  :hook (prog-mode . (lambda ()
                          (unless (derived-mode-p 'emacs-lisp-mode 'lisp-mode)
                            (lsp-deferred))))
  :init
  (setq  lsp-auto-guess-root t
        lsp-prefer-capf  t
        lsp-print-io t
        lsp-keep-workspace-alive nil)
 :config
 (when lsp-auto-configure
    (mapc (lambda (package) (require package nil t))
          lsp-client-packages)))
(use-package company
  :commands company-complete-common company-manual-begin company-grab-line
  :after-call pre-command-hook after-find-file
  :init
  (setq company-tooltip-align-annotations t
        company-tooltip-limit 14
        company-idle-delay 0
        company-echo-delay (if (display-graphic-p) nil 0)
        company-minimum-prefix-length 2
        company-require-match 'never
        company-dabbrev-ignore-case nil
        company-dabbrev-downcase nil
        company-global-modes '(not erc-mode message-mode help-mode gud-mode eshell-mode shell-mode)
        company-backends '(company-capf)
        company-frontends '(company-pseudo-tooltip-frontend
                            company-echo-metadata-frontend))
  :config
  (global-company-mode +1))

```

**Expected behavior**
1. doesn't  exact match
2. filter the response.

**Which Language Server did you use**
lsp-go

**OS**
macos 10.15.3

**Error callstack**

kiennq:
Can you provide the language server log according to [this](https://github.com/emacs-lsp/lsp-mode)?
I cannot repro bug 1:
[...]

PR #1454, opened 2020-02-25
lsp-capf: convert from LSP snippet to yasnippet

kiennq:
This fix one of problem mentioned in #1451 
Specifically problem mentioned [here](https://github.com/emacs-lsp/lsp-mode/issues/1451#issuecomment-590695358)

We need to convert from LSP snippet format to yasnippet format when expanding it.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4106,9 +4106,10 @@ PLIST is the additional data to attach to each candidate."
              (insert insert-text)))
 
            (when (eq insert-text-format 2)
-             (yas-expand-snippet (buffer-substring start-point (point))
-                                 start-point
-                                 (point)))
+             (yas-expand-snippet
+              (lsp--to-yasnippet-snippet (buffer-substring start-point (point)))
+              start-point
+              (point)))
            (when additional-text-edits
              (lsp--apply-text-edits additional-text-edits)))
          (lsp--capf-clear-cache)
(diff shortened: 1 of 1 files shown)


s-kostyaev:
I've just checked it out. It works fine, thanks!

yyoncho:
This breaks the CI, can you take a look?

lsp-mode.el
```diff
@@ -4021,6 +4021,14 @@ PLIST is the additional data to attach to each candidate."
         (setq label-pos 0)))
     matches))
 
+(defun lsp--capf-to-yasnippet-snippet (text)
+  "Convert LSP snippet TEXT to yasnippet snippet."
+  ;; Need to escape "{"
+  (s-replace-regexp (rx (not (or "$" "\\")) (group "{"))
```
s-kostyaev: I think here should be:
```elisp
(s-replace-regexp (rx (not (or "\$" "\\")) (group "{"))
```
It can fix CI for emacs < 27

s-kostyaev:
Sorry, it was wrong tip. Correct one:
```elisp
(defun lsp--capf-to-yasnippet-snippet (text)
  "Convert LSP snippet TEXT to yasnippet snippet."
  ;; Need to escape "{"
  (s-replace-regexp (rx (not (any "$" "\\")) (group "{"))
                    (rx "\\" (backref 1))
                    text
                    nil nil 1))
```

thanks @larsbrinkhoff for https://hub.docker.com/r/larsbrinkhoff/emacsen/

PR #1460, opened 2020-02-27
Adds `lsp-intelephense-multi-root` custom flag

polaris64:
This flag is used during `make-lsp-client` for the `multi-root` flag.
Default value is set which matches the previous setting.  Unsetting this
flag forces lsp-mode to start a separate language server per project.


yyoncho:
Can you change the property version to 6.3?

yyoncho:

PR #1464, opened 2020-03-01
[WIP] Rework semantic highlighting

sebastiansturm:
this fixes a typo and adds support for clangd's highlighting of disabled lines based on the `isInactive` property. If anyone else is using clangd-based semantic highlighting, I'd welcome feedback or bug reports. Otherwise I'll just merge this in a few days as I don't think it could break anything not related to semantic highlighting

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5120,7 +5120,7 @@ unless overridden by a more specific face association."
 unless overridden by a more specific face association."
   :group 'lsp-faces)
 
-(defface lsp-face-semhl-disable
+(defface lsp-face-semhl-disabled
   '((t :inherit font-lock-comment-face))
   "Face used for semantic highlighting scopes matching meta.disabled,
 unless overridden by a more specific face association."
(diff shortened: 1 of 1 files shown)


sebastiansturm:
following up on https://github.com/emacs-lsp/lsp-mode/issues/1465, I've disabled field boundary checking within semantic highlighting and within `lsp--line-character-to-point`. Which turns this PR into a bit of a mixed bag, but at least everything contained herein is in some way related to overlays (and I don't see a convenient way to open and maintain two separate PRs at once)

sebastiansturm:
despite some other overlay-related performance improvements, I'm still finding situations where my implementation of semantic highlighting offers awful performance. In particular, creating or deleting lines at the top of a large buffer apparently causes clangd to reemit highlighting information for pretty much everything following that newly created or deleted line (whereas I had previously assumed clangd would not reemit highlighting data for unchanged lines but leave it up to the frontend to correctly move highlights along with the underlying text). For this reason I'm now experimenting with a naive deferred implementation of overlay-based semantic highlighting that keeps a cache of line highlighting information and renders it on an idle timer. Obviously that method can get out of sync with the actual buffer contents, so whether that works in practice depends on the updates sent out by the server. The old (slow) method is still available, but I haven't tested yet if it still works (and even if it does, there is some code duplication right now that I should get rid of).
Since all these performance issues should go away once the `noverlay` branch is merged, I don't think a perfect solution is required. I'll test my current implementation during the next few days and see if it works well with clangd. Merging this PR into master would not be a good idea at this point though

yyoncho:
Have you considered putting highlighting code on top of font-lock mode? AFAIK It will call you to apply highlights only for the visible/changed areas?

sebastiansturm:
> Have you considered putting highlighting code on top of font-lock mode? AFAIK It will call you to apply highlights only for the visible/changed areas?

no but that sounds like something I should be using. Will have a look, thanks

sebastiansturm:
it will be a few days before I can get back to this, so I think I should extract the "inhibit field text motion" commit from this PR (which I should have done anyway) so it can be merged separately. By inhibiting field text motion, `lsp--line-character-to-point` does not only get a lot faster if we have many overlays or text properties around, I'd also think the current way is incorrect, because when we're talking to language servers about line boundaries, surely they won't care about Emacs's notion of "fields"?

Maybe I'm missing something here and there is no haste in getting this change merged, so I'll wait a bit longer to see if anybody knows a good reason why we should be respecting field boundaries after all

lsp-mode.el
```diff
@@ -167,13 +167,21 @@ the buffer when it becomes large."
...
 This feature is not yet part of the official LSP spec and may
 occasionally break as language servers are updated."
   :group 'lsp-mode
-  :type 'boolean)
+  :type '(choice
+          (const :tag "Disable" nil)
+          (const :tag "Immediate" :immediate)
+          (const :tag "Deferred" :deferred)))
+
+(defcustom lsp-semantic-highlighting-window-size 70
```
danielmartin: I'd rename this to `lsp-semantic-highlighting-context-lines`, following a similar naming convention as occur's `list-matching-lines-default-context-lines` or compile's `compilation-context-lines`.

sebastiansturm: thanks for testing this! I've tried to address these changes.
Following a proposal by @yyoncho, I also added a commit that uses font-lock callbacks to refontify changed or newly visited buffer parts. Doesn't work very well yet though (triggering refontification on buffer content changes doesn't really seem like a good idea to me as long as we don't use the new semantic tokenization protocol, as the first fontification run may often precede the arrival of the new highlighting information we need).
Maybe I should just get rid of the last commit, merge the non-font-lock state of this PR and revisit font-lock integration as soon as support for the new highlighting protocol gets implemented?

yyoncho: > Maybe I should just get rid of the last commit, merge the non-font-lock state of this PR and revisit font-lock integration as soon as support for the new highlighting protocol gets implemented?

Make sense to me.

sebastiansturm: I now changed back to a non-font-lock implementation. I suggest to give this a few more days of testing, then I'd squash all commits except for the `inhibit-field-motion` thing and merge this PR

PR #1469, opened 2020-03-01
generic flycheck checker issues

brotzeit:
I'm trying to run the generic syntax checker 

```elisp
(flycheck-define-generic-checker 'lsp
    "A syntax checker using the Language Server Protocol (LSP)
provided by lsp-mode.
See https://github.com/emacs-lsp/lsp-mode."
    :start #'lsp--flycheck-start
    :modes '(python-mode)
    :error-explainer (lambda (e)
                       (cond ((string-prefix-p "clang-tidy" (flycheck-error-message e))
                              (lsp-cpp-flycheck-clang-tidy-error-explainer e))
                             (t (flycheck-error-message e)))))
```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7248,18 +7248,21 @@ CALLBACK is the status callback passed by Flycheck."
                 :line (1+ (lsp-diagnostic-line diag))
                 :column (1+ (lsp-diagnostic-column diag))
                 :message (lsp-diagnostic-message diag)
-                :level (lsp--flycheck-calculate-level diag)
+                :level (if (string-match "clippy" (lsp-diagnostic-message diag))
+                           'info
+                        (lsp--flycheck-calculate-level diag))
                 :id (lsp-diagnostic-code diag)
-                :end-column (-> diag
-                                lsp-diagnostic-range
-                                (plist-get :end)
-                                (plist-get :column)
-                                (1+))
-                :end-line (-> diag
-                              lsp-diagnostic-range
-                              (plist-get :end)
-                              (plist-get :line)
-                              (1+)))))
+                ;; :end-column (-> diag
+                ;;                 lsp-diagnostic-range
+                ;;                 (plist-get :end)
+                ;;                 (plist-get :column)
+                ;;                 (1+))
+                ;; :end-line (-> diag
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7248,18 +7248,21 @@ CALLBACK is the status callback passed by Flycheck."
...
+                ;; :end-column (-> diag
+                ;;                 lsp-diagnostic-range
+                ;;                 (plist-get :end)
+                ;;                 (plist-get :column)
+                ;;                 (1+))
+                ;; :end-line (-> diag
+                ;;               lsp-diagnostic-range
+                ;;               (plist-get :end)
+                ;;               (plist-get :line)
+                ;;               (1+))
```
brotzeit: But I can't build lsp-mode with these lines.

lsp-mode.el
```diff
@@ -7248,18 +7248,21 @@ CALLBACK is the status callback passed by Flycheck."
                 :line (1+ (lsp-diagnostic-line diag))
                 :column (1+ (lsp-diagnostic-column diag))
                 :message (lsp-diagnostic-message diag)
-                :level (lsp--flycheck-calculate-level diag)
+                :level (if (string-match "clippy" (lsp-diagnostic-message diag))
+                           'info
+                        (lsp--flycheck-calculate-level diag))
```
brotzeit: The problem with clippy is that it returns error code 1. I don't like this solution so maybe somebody has a but idea ?

yyoncho:
Did you update your flycheck? They introduced the :end-column and :end-line a week ago.

brotzeit:
Yeah, I did but I forgot to restart emacs =)
How to handle the clippy thing ?

lsp-mode.el
yyoncho: I don't understand why returning error code 1 is a problem.

yyoncho:
> How to handle the clippy thing ?

can you elaborate?

brotzeit:
IMO clippy lints should be info types.

yyoncho:
> IMO clippy lints should be info types.

I guess this should be fixed on the server-side?

brotzeit:
I guess you're right.

PR #1470, opened 2020-03-02
Add metals experimental capabilities.

kurnevsky:
They were added in https://github.com/scalameta/metals/pull/1414


yyoncho:
Thank you!

PR #1472, opened 2020-03-04
use q to kill metals doctor buffer

contrun:
@dsyzling For review. This PR is not much, but it surely improve my life quality a little bit.

### DIFF
--- a/lsp-metals.el
+++ b/lsp-metals.el
@@ -117,7 +117,9 @@ more customizations like using environment variables."
   (erase-buffer)
   (insert html)
   (shr-render-region (point-min) (point-max))
-  (setq-local buffer-read-only t))
+  (goto-char (point-min))
+  (view-mode 1)
+  (setq view-exit-action 'kill-buffer))
 
 (defun lsp-metals--generate-doctor-buffer-name (workspace)
   (format "*Metals Doctor: %s*" (process-id (lsp--workspace-cmd-proc workspace))))


lsp-metals.el
```diff
@@ -114,6 +114,7 @@ more customizations like using environment variables."
   (require 'shr)
   (setq-local show-trailing-whitespace nil)
   (setq-local buffer-read-only nil)
+  (local-set-key (kbd "q") 'kill-this-buffer)
```
kiennq: Should we make this buffer as `special-mode` so the `q` key bind will work automatically? Also that makes buffer read-only too.

contrun: Your concern is reasonable. It seems to be a overkill in this scenario. I think we don't have any other read only buffer for metals. As for lsp-mode in general, I am not knowledgeable enough.

kiennq: I surely recommend to use existing mode key binds (in this case, `special-mode`) instead of adding new key bind randomly. It allows user customized key binds to work while still provide a good default

contrun: Yeah. We should do it. What about `(view-mode 1)`?

lsp-metals.el
```diff
@@ -117,7 +117,9 @@ more customizations like using environment variables."
   (erase-buffer)
   (insert html)
   (shr-render-region (point-min) (point-max))
-  (setq-local buffer-read-only t))
+  (beginning-of-buffer)
```
kiennq: This function is for interactive use only;
  in Lisp code use `(goto-char (point-min))' instead.

lsp-metals.el
```diff
@@ -117,7 +117,9 @@ more customizations like using environment variables."
   (erase-buffer)
   (insert html)
   (shr-render-region (point-min) (point-max))
-  (setq-local buffer-read-only t))
+  (goto-char (point-min))
+  (view-mode 1)
+  (setq view-exit-action 'kill-buffer))
```
yyoncho: setq-local?

kurnevsky: `view-exit-action` automatically becomes buffer-local when set. So it should be fine unless you think it's clearer to use `setq-local` explicitly :)

yyoncho: That's true.

PR #1473, opened 2020-03-04
add texlab as tex lsp server

contrun:
I also made [texlab](https://github.com/latex-lsp/texlab) as the default lsp sever, as it seems to be well-maintained than [digestif](https://github.com/astoff/digestif).


yyoncho:
Thank you!

PR #1475, opened 2020-03-05
lsp-capf: prevent double filtering

kiennq:
Prevent double filtering by:
- Define new category for `lsp-capf` which will always have style `basic`. This will prevent `company-capf` from performing additional filtering.
- Define `lsp-completion-styles` to allow user to customize completion styles to use when filtering candidates.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4168,11 +4168,7 @@ Also, additional data to attached to each candidate can be passed via PLIST."
          (cond
           ((eq action 'metadata)
            `(metadata ((category . lsp-capf)
-                       (display-sort-function . (lambda (candidates)
-                                                  (--sort (string-lessp
-                                                           (get-text-property 0 'lsp-sort-text it)
-                                                           (get-text-property 0 'lsp-sort-text other))
-                                                          candidates))))))
+                       (display-sort-function . #'identity))))
           ((eq (car-safe action) 'boundaries) nil)
           ;; retrieve candidates
           (done? result)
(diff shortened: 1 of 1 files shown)


yyoncho:
Works great! Few questions:

Will we skill all client-side filtering if we do `(setq lsp-completion-styles nil)`?
What do you think about helm flex filtering? Does it make sense to use/recommend it or at least mention it in readme?
What do you think about setting substring strategy as a default for emacs 26? 

I will prepare a documentation update and ask you for review.

kiennq:
> Will we skill all client-side filtering if we do `(setq lsp-completion-styles nil)` (e. g. useful for very smart servers like clangd)?

No, it's actually the opposite, if we set the `lsp-completion-styles` to `nil` all the items will be filtered out. Same behavior with `completion-styles`.
Note that if the server returns `isIncomplete` as `true` we will stop filtering and re-ranking anyway.



> What do you think about helm flex filtering? Does it make sense to use/recommend it or at least mention it in readme

I haven't try `helm-flex` filtering yet. It seems need more configure to work.



> What do you think about setting substring strategy as a default for emacs 26?

This is good too, I will set it as default for emacs 26.

yyoncho:
> I haven't try `helm-flex` filtering yet. It seems need more configure to work.

I guess we have to investigate it as an option for emacs 26.

yyoncho:
Hm, I noticed that the completions are not sorted backwards sometimes?

sorting

kiennq:
It seems the items returned from filtering is not properly sort
Fixed in #1483

PR #1483, opened 2020-03-06
lsp-capf: sort candidate after filtering

kiennq:
We need to sort candidates based on filtering score.
Fix issue mention in [#1445(comment)](https://github.com/emacs-lsp/lsp-mode/pull/1475#issuecomment-595884077)


yyoncho:
Thank you!

PR #1492, opened 2020-03-08
dls: configurable server path and args

kubo39:
This commit allows to custom server's path and arguments by using
`lsp-dls-path` or `lsp-dls-args`.

### DIFF
--- a/lsp-dls.el
+++ b/lsp-dls.el
@@ -31,7 +31,7 @@
   :group 'lsp-mode
   :link '(url-link "https://github.com/d-language-server/dls"))
 
-(defcustom lsp-dls-path "~/.dub/packages/.bin/dls-latest/dls"
+(defcustom lsp-dls-path "dls"
   "Path to dls server binary."
   :type 'string
   :group 'lsp-dls)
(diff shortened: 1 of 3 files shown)


lsp-dls.el
```diff
@@ -0,0 +1,56 @@
+;;; lsp-dls.el --- dls configuration               -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 Hiroki Noda
```
kubo39: Is it better to use `lsp-mode developers` ?

yyoncho: It is fine to have it like that.

lsp-dls.el
```diff
@@ -0,0 +1,56 @@
...
+;;; Code:
+
+(require 'lsp-mode)
+
+(defgroup lsp-dls nil
+  "LSP support for D, using https://github.com/d-language-server/dls."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/d-language-server/dls"))
+
+(defcustom lsp-dls-path "~/.dub/packages/.bin/dls-latest/dls"
```
kiennq: I think it should be just `dls` here, since even in other language servers, it would just expose the server binary to `$PATH`.
That will help with user who install server into custom path and then expose the server to global PATH

yyoncho:
Thank you!

Issue #1172, opened 2019-11-11
Create refcard 


yyoncho:
We could use cider's as a guide: 

https://github.com/clojure-emacs/cider/blob/master/refcard/cider-refcard.pdf

alanz:
Perhaps include an evil mode one too.

PR #1493, opened 2020-03-08
Add reference card for essential commands

innerout:
This pull request introduces a basic reference card for essential commands 
and is inspired by cider as mentioned in #1172 .


yyoncho:
Thank you for doing that, looks good to me! Let others share their opinion as well.

chrbirks:
What about including the name of the function that each shortcut calls? It could help those who have conflicting shortcuts or want to write a hydra with some of these.

innerout:
If I add too much text in a single command there is wrapping in the line then but I could make another reference card that contains the commands instead of the description.

yyoncho:
@chrbirks we are using cider as a guide for this refcard. I guess the answer to your issue is using M-x describe-key.

yyoncho:
Thank you!

PR #1503, opened 2020-03-10
Support "only" param when requesting code actions

### LINKED ISSUE #1502: Support "only" option when asking for code actions
**Describe the bug**
lsp-organize-imports can take a long time if it ends up waiting for other expensive code actions 

**To Reproduce**
gopls has a very expensive code action that sometimes locks up my lsp-organize-imports before-save-hook for a few seconds

**Expected behavior**
lsp-organize-imports should specify the CodeActionParams.context.only flag so the server knows to only prepare the organizeImports code action

gopls

**OS**
mac

muirdm:
This lets the server know we only care about, for example,
source.organizeImports. The server can avoid preparing other
potentially slow actions. gopls in particular currently has a
sometimes-slow quickfix action that makes my lsp-organize-imports save
hook get stuck for a few seconds.

Fixes #1502.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4834,21 +4834,23 @@ It will show up only if current point has signature help."
             (gethash command))
           (lsp-workspaces)))
 
-(defun lsp--text-document-code-action-params ()
+(defun lsp--text-document-code-action-params (&optional kind)
   "Code action params."
   (list :textDocument (lsp--text-document-identifier)
         :range (if (use-region-p)
                    (lsp--region-to-range (region-beginning) (region-end))
                  (lsp--region-to-range (point) (point)))
-        :context (list :diagnostics (lsp-cur-line-diagnostics))))
+        :context (list
+                  :diagnostics (lsp-cur-line-diagnostics)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4834,21 +4834,23 @@ It will show up only if current point has signature help."
...
+(defun lsp--text-document-code-action-params (&optional kind)
   "Code action params."
   (list :textDocument (lsp--text-document-identifier)
         :range (if (use-region-p)
                    (lsp--region-to-range (region-beginning) (region-end))
                  (lsp--region-to-range (point) (point)))
-        :context (list :diagnostics (lsp-cur-line-diagnostics))))
+        :context (list
+                  :diagnostics (lsp-cur-line-diagnostics)
+                  :only (if kind (list kind) nil))))
```
yyoncho: nit: use `when` instead of if/else/nil and also we should send a vector not a list, otherwise it will break emacs 27.

muirdm: Done.

yyoncho:
Thank you!

PR #1506, opened 2020-03-11
lsp-pwsh: automatically create logs directory

kiennq:
As discussed in gitter, we should create logs directory automatically in case the language server is not creating it first.


yyoncho:
Thank you!

PR #1510, opened 2020-03-12
Install "vls" instead of "vue-language-server"

yhiraki:
https://vuejs.github.io/vetur/CHANGELOG.html#_0-24-0-2020-03-04-vsix
> vue-language-server deprecated and will be published as vls.


yyoncho:
Thank you!

PR #1519, opened 2020-03-14
Update rust-analyzer runnable support to work with newest RA

flodiebold:
The args after the `--` have been separated into a new field to support debugging.


yyoncho:
Thank you!

PR #1526, opened 2020-03-17
add support robot-framework

dalanicolai:
Here is a rudimentary implementation of the robot-framework lsp-client. It can already be used in a quite straightforward way, so you might agree to merge it already. However, it would be nice to at least implement the right menu's for the log and trace variables. So for now I am sending this PR mainly so you can have a look at it. I can update the documentation too, but I don't know how to instal the server without vscode, I never really worked with npm or javascript. Would be great to get your feedback.

Additionally I could update the spacemacs lsp layer later

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -28,6 +28,8 @@
 (require 'dash)
 (require 'dash-functional)
 (require 'rx)
+(require 'cl-lib)
+(require 'em-glob)
 
 ;;; Ada
 (defgroup lsp-ada nil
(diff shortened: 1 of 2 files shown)


lsp-clients.el
```diff
@@ -828,5 +830,120 @@ responsiveness at the cost of possible stability issues."
...
+  :type 'boolean
+  :group 'lsp-rf)
+
+(defcustom lsp-rf-language-server-dir "~/.vscode/extensions/tomiturtiainen.rf-intellisense-2.8.0/server/library-docs/"
+  "Libraries directory for libraries in lsp-rf-language-server-libraries"
+  :type 'string
+  :group 'lsp-rf)
+
+(defcustom lsp-rf-language-server-libraries '("BuiltIn-3.1.1" "Collections-3.0.4")
+  "Libraries whose keywords are suggested with auto-complete"
```
yyoncho: should be vector. Note that emacs27+ json serialization treats list as objects

lsp-clients.el
```diff
@@ -828,5 +830,120 @@ responsiveness at the cost of possible stability issues."
...
+
+(lsp-register-custom-settings
+ '(
+   ("rfLanguageServer.trace.server" lsp-rf-language-server-trace-server)
+   ("rfLanguageServer.logLevel" lsp-rf-language-server-log-level)
+   ("rfLanguageServer.libraries" lsp-rf-language-server-libraries)
+   ("rfLanguageServer.pythonKeywords" lsp-rf-language-server-python-keywords t)
+   ("rfLanguageServer.excludePaths" lsp-rf-language-server-exclude-paths)
+   ("rfLanguageServer.includePaths" lsp-rf-language-server-include-paths)
+   ))
```
yyoncho: close on  the same line

lsp-clients.el
```diff
@@ -828,5 +830,120 @@ responsiveness at the cost of possible stability issues."
...
+                  :initialized-fn (lambda (workspace)
+                                    (with-lsp-workspace workspace
+                                      (lsp--set-configuration
+                                       (lsp-configuration-section "rfLanguageServer"))
+                                      (lsp-request "buildFromFiles"
+                                                   (list :files
+                                                         (vconcat
+                                                          (rf-language-server-exclude-paths
+                                                           (directory-files-recursively
+                                                            (projectile-project-root)
```
yyoncho: You should not use projectile here but `lsp--workspace-root`

lsp-clients.el
```diff
@@ -828,5 +830,120 @@ responsiveness at the cost of possible stability issues."
...
+   ("rfLanguageServer.logLevel" lsp-rf-language-server-log-level)
+   ("rfLanguageServer.libraries" lsp-rf-language-server-libraries)
+   ("rfLanguageServer.pythonKeywords" lsp-rf-language-server-python-keywords t)
+   ("rfLanguageServer.excludePaths" lsp-rf-language-server-exclude-paths)
+   ("rfLanguageServer.includePaths" lsp-rf-language-server-include-paths)
+   ))
+
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-stdio-connection
+                                   `("node" ,(expand-file-name
```
yyoncho: please make `node` part of the command to allow customizing it.

lsp-clients.el
```diff
@@ -828,5 +830,120 @@ responsiveness at the cost of possible stability issues."
...
+  "An array of files that should be included by the parser. Glob patterns as strings are accepted (eg. *.robot between double quotes)"
+  :type 'lsp-string-vector
+  :group 'lsp-rf)
+
+(defcustom lsp-rf-language-server-exclude-paths []
+  "An array of files that should be ignored by the parser. Glob patterns as strings are accepted (eg. *bad.robot between double quotes)"
+  :type 'lsp-string-vector
+  :group 'lsp-rf)
+
+(defcustom lsp-rf-language-server-python-keywords nil
```
yyoncho: if it is deprecated consider not adding it at all.

lsp-clients.el
dalanicolai: This means that for now it works correctly, but from emacs 27+ on it will fail?

yyoncho: yes.

lsp-clients.el
dalanicolai: I've added lsp-workspace-root. I looked at lsp--workspace-root but I am not familiar with cl structures yet. I don't know what to use as argument (CL-X). I tried lsp--cur-workspace but its value is nil

lsp-clients.el
```diff
@@ -828,5 +830,120 @@ responsiveness at the cost of possible stability issues."
...
+                  :initialized-fn (lambda (workspace)
+                                    (with-lsp-workspace workspace
+                                      (lsp--set-configuration
+                                       (lsp-configuration-section "rfLanguageServer"))
+                                      (lsp-request "buildFromFiles"
+                                                   (list :files
+                                                         (vconcat
+                                                          (parse-rf-language-server-exclude-paths
+                                                           (directory-files-recursively
+                                                            (lsp-workspace-root)
```
yyoncho: pass `workspace`

dalanicolai: What do you mean? You mean pass that to the `lsp--workspace-root`? (Probably not because that doesn't work)

yyoncho: should be `(lsp--workspace-root workspace)`

dalanicolai: Okay, replaced that line. However, I tested this before from a buffer that should be "inside" the workspace but then workspace was void. Now I see the workspace is passed by the initialized-fn "keyword".

yyoncho:
Thank you for contributing to `lsp-mode` and congrats for your first PR! :tada:

dalanicolai:
Great! Thanks... I will soon have a look where to update some documentation and probably also add some configuration for the spacemacs layer...

PR #1527, opened 2020-03-18
lsp-metals: Support `/**` completion

### LINKED ISSUE #1525: Support scaladoc auto completion on type `/**`
Scala metals added support for auto completion on type `/**`. The relevant PR can be found here: https://github.com/scalameta/metals/pull/1250

However, on Emacs this doesn't seem to work. It would be nice to add it.

kiennq:
This fix #1525 
There're two issues:
- The prefix length is `nil` due to symbol boundary is `nil` in this case. This the completion request is not triggered.
  We should always calculate `company-prefix-length` so it will automatically trigger completion even if the current position is not belong to a symbol.
  We new approach, the prefix length will be decided as the length to closest previous whitespace boundary.
- `lsp-metals` doesn't support `completion-in-comments`, thus make completion is not triggered while inside comment. We should fix that too.

### DIFF
--- a/lsp-metals.el
+++ b/lsp-metals.el
@@ -275,7 +275,8 @@ server via `window/logMessage'."
                                       (lsp--set-configuration
                                        (lsp-configuration-section "metals"))))
                   :after-open-fn (lambda ()
-                                   (add-hook 'lsp-on-idle-hook #'lsp-metals--did-focus nil t))))
+                                   (add-hook 'lsp-on-idle-hook #'lsp-metals--did-focus nil t))
+                  :completion-in-comments? t))
 
 (provide 'lsp-metals)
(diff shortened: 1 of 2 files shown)


yyoncho:
Looks good, I will spend some time manually testing the PR before merging.

yyoncho:
As a side note, the 25.x failures do not look false negative: https://travis-ci.org/github/emacs-lsp/lsp-mode/jobs/663796560?utm_medium=notification&utm_source=github_status

kiennq:
> As a side note, the 25.x failures do not look false negative: https://travis-ci.org/github/emacs-lsp/lsp-mode/jobs/663796560?utm_medium=notification&utm_source=github_status

Should be fixed now

PR #1542, opened 2020-03-27
Use `*${name}::stderr*` template for tcp-connection process stderr buffer

seriyps:
Right now when server uses TCP transport (eg, erlang_ls), we launch a server process and this creates a new buffer for process's stderr with the name of the language server (eg, `erlang-ls`).

But "stdio" transport uses "*${name}::stderr*" template for subprocess stderr output.

I think it's better fit, because, according to https://www.emacswiki.org/emacs/Buffer, buffers that are not associated with files are usually have names enclosed with asterisks.

With this PR both stderr buffers for both TCP and stdio transports will look uniform.


yyoncho:

PR #1544, opened 2020-03-28
bug: just use symbol prefix and trigger characters for completions

### LINKED ISSUE #1536: Completion error
It will show completion items when input `(` or `{`, Seems like this commit `2472231de1bebbd3236a3a3c8ba0cbfe58364faf` caused . please check @kiennq  @yyoncho 
image

kiennq:
This should fix #1536 
We revert back to using symbol boundary prefix and trigger characters to trigger completion request.


yyoncho:
Thank you!

Issue #1543, opened 2020-03-28
lsp-fsharp using old version of FsAutoComplete


mpardalos:
I believe lsp-fsharp is downloading an old version of fsac. If I run `lsp-fsharp-update-fsac` and then run  `dotnet fsautocomplete.dll` in the directory where it was downloaded it gives:
```
FsAutoComplete 0.38.2 (git sha 7db211bec017e54fdd4a0e10ead2de6760dcaab5)
```
Where the latest version that should be available is 0.41.1.

Deleting and re-running `lsp-install-server` also gives the same result.

This is causing #1508 to occur even though the bug causing it was fixed in fsac.

We are currently downloading fsac from the following appveyor link:

https://ci.appveyor.com/api/projects/fsautocomplete/fsautocomplete/artifacts/bin/pkgs/fsautocomplete.netcore.zip?branch=master

According to [fsautocomplete's README](https://github.com/fsharp/FsAutoComplete), they are currently using Azure pipelines for their CI. Does this mean we are using an old download link?

The latest downloads are available on their [github releases](https://github.com/fsharp/FsAutoComplete/releases). Maybe we should be downloading it from there?

yyoncho:
> The latest downloads are available on their [github releases](https://github.com/fsharp/FsAutoComplete/releases). Maybe we should be downloading it from there?

I suggest checking where vscode download it from.

mpardalos:
ionide/ionide-vscode-fsharp seems to build fsac master at build time: 

https://github.com/ionide/ionide-vscode-fsharp/blob/9f2aa197bcadcfa665be959bbf50c1fd3bc605bd/build.fsx#L33.

mpardalos:
I was looking to see if there's a way to get the latest github release. It turns out there is no stable link, however they do have an API endpoint for getting a list of releases where we can get the link to the last release. You can get the link with this jq expression:

```
curl https://api.github.com/repos/fsharp/fsautocomplete/releases/latest \
| jq '.assets | .[] | select(.name == "fsautocomplete.netcore.zip") | .browser_download_url'
```

I am not sure if it's easy to integrate this into the current `lsp-install-server`. Do we have to provide a direct link to the download or can we directly provide a download function?

If that's possible I could submit a PR to get the zip from there.

razzmatazz:
I was hit by this issue too.

[...]

PR #1546, opened 2020-03-29
lsp-fsharp.el: use github instead of appveyor to retrieve server

razzmatazz:
As @mpardalos has noted in #1543, fsautocomplete no longer has the latest version built and available on appveyor CI. I tried to copy a couple of functions from `lsp-csharp.el` to do the same github api calls to query the latest version of the project and retrieve the latest one.

Related to #1543

### DIFF
--- a/lsp-fsharp.el
+++ b/lsp-fsharp.el
@@ -48,14 +48,6 @@ The slash is expected at the end."
   :type 'directory
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-fsharp-server-download-url "https://ci.appveyor.com/api/projects/fsautocomplete/fsautocomplete/artifacts/bin/pkgs/fsautocomplete.netcore.zip?branch=master"
-  "Fsautocomplete download url.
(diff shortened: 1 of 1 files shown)


lsp-fsharp.el
```diff
@@ -179,14 +171,46 @@ disable if `--backgorund-service-enabled' is not used"
...
+      (json-read))))
+
+(defun lsp-fsharp--latest-version-from-github ()
+  "Return latest version of the server available from github."
+   (lsp-fsharp--version-list-latest
+    (seq-map (lambda (elt) (s-trim (cdr (assq 'name elt))))
+             (lsp-fsharp--fetch-json "https://api.github.com/repos/fsharp/FsAutoComplete/releases"))))
+
+(defun lsp-fsharp--server-download-url (version)
+  "Return url for .zip file to download for given VERSION, depending on lsp-fsharp-server-runtime."
```
kiennq: `https://github.com/%s/%s/releases/latest/download/%s`
If you just want latest version, this url is enough.
It save one web query which can be quite slow.

PR #1548, opened 2020-03-29
Remember current buffer on lsp--document-links

sebastiansturm:
The way `lsp--document-links` is currently implemented, it may accidentally generate link highlights in the wrong buffer. This PR should fix that issue by remembering the current buffer and switching back to it once the async request gets resolved

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2879,6 +2879,11 @@ If NO-MERGE is non-nil, don't merge the results but return alist workspace->resu
                                            local))
                                         hooks)))
                               (remhash cancel-token lsp--cancelable-requests)))
+             (callback (pcase mode
+                         ((or 'alive 'tick) (lambda (&rest args)
+                                              (with-current-buffer buf
+                                                (apply callback args))))
+                         (_ callback)))
              (callback (lsp--create-async-callback callback
                                                    method
                                                    no-merge


lsp-mode.el
```diff
@@ -3913,24 +3913,27 @@ Applies on type formatting."
...
-                            (lsp--position-to-point (gethash "end" range))
-                            'action (lsp--document-link-keymap link)
-                            'keymap (let ((map (make-sparse-keymap)))
-                                      (define-key map [M-return] 'push-button)
-                                      (define-key map [mouse-2] 'push-button)
-                                      map)
-                            'help-echo "mouse-2, M-RET: Visit this link")
-          (overlay-put 'lsp-link t)))
-      links))
-   :mode 'tick))
```
yyoncho: I think that all this is a regression from the lsp-request-async refactoring. In the past, we were wrapping the callback in lsp--create-async-callback for `tick/alive` and we were doing with-current-buffer in it. Would you mind to fix the issue it that way to avoid the code duplication?

sebastiansturm: sure, should have checked the commit history first. Thanks for the info

sebastiansturm: I've now had a look at this and I agree that `'tick` should use `(with-current-buffer ...)`, though I guess that most current users of `'tick` should probably switch to `'unchanged` anyway? Legitimate uses of a buffer-changing `'tick` implementation would probably be `lsp--on-type-formatting` and `lsp--before-save`, but `lsp--document-links`, `lsp-hover` and `lsp--document-highlight` not so much. Do you agree?

yyoncho: Yes.

sebastiansturm: Since all modes cancel the request if the current buffer is killed, I'd just do the `(with-current-buffer ...)` unconditionally, or would you prefer not to needlessly wrap the callback if mode is `'current` or `'unchanged`?

yyoncho: > Since all modes cancel the request if the current buffer is killed, 

`detached` is supposed not to do that?

sebastiansturm: oh yes, I was looking at the values listed in the `(pcase mode`, not at the docstring. So, I'll just wrap it for `alive/tick` then as you originally proposed

sebastiansturm: do you see any issues with the change? I've been using it for about a week, seemed to work for me so far

yyoncho:
Thank you!

> do you see any issues with the change? I've been using it for about a week, seemed to work for me so far

I haven't seen a notification - it is good to ping when you perform force push.

sebastiansturm:
ah I see. Thanks!

PR #1554, opened 2020-04-01
Fix npm instructions in documentation

lovebug356:
installing with npm is done with `npm i` or `npm install` (not `npm -i`)


yyoncho:
Thank you!

Issue #1560, opened 2020-04-03
[Dart] Feature request: new variables for initialization options


jpprime:
I would really love to see variables to control the new three initialization option on the analysis server. Specifically you can now toggle closingLabels, outline and flutterOutline on or off. More information on these options can be found [here](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md).

ericdallo:
@jpprime, [this](https://github.com/emacs-lsp/lsp-mode/pull/1561) add the variables, but Im not sure if lsp has the support for these features currently.

ericdallo:
Currently `lsp-mode` has support only for the `closingLabels` feature. I added support on the PR :point_up:

jpprime:
The `closingLabels` is actually the feature I was mostly missing, since the `outline` and `flutterOutline` features are available in the DevTools, although it would be really nice if I could not use Chromium just for that and have it available on Emacs. Is it possible to add the outline features? I am willing to help.

ericdallo:
I don't know how to implement these features as `lsp-treemacs` currently does not has this support like said it here.
@yyoncho, do you have any idea if we can implement these features in any way?

yyoncho:
> I don't know how to implement these features as `lsp-treemacs` currently does not has this support like said it here.
> @yyoncho, do you have any idea if we can implement these features in any way?

Yes, we can - we have created multiple tree views in lsp-treemacs/lsp-java/dap-mode/ccls/. You may take a look at `lsp-treemacs-call-hierarchy`. Feel free to ping me if you need more clarification.

jpprime:
I guess this should be closed and maybe moved to `lsp-treemacs` then, since the option that could be implemented on the main `lsp-mode` has been implemented thanks to @ericdallo.

ericdallo:
@jpprime we are thinking on implement those temporary on lsp-mode then move to a separated `lsp-dart` repository.
The PR is here: https://github.com/emacs-lsp/lsp-mode/pull/1573

ericdallo:
I think we can close this given the above merged PR @yyoncho

yyoncho:
Yes, thanks.

PR #1561, opened 2020-04-04
[lsp-dart] Add more variables to initializationOptions

ericdallo:
Add support for the [closingLabels initializationOptions](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md#initialization-options).
image

Fixes #https://github.com/emacs-lsp/lsp-mode/issues/1560

### DIFF
--- a/lsp-dart.el
+++ b/lsp-dart.el
@@ -94,20 +94,21 @@ PARAMS closing labels notification data sent from WORKSPACE."
   (-let* (((&hash "uri" "labels") params)
           (buffer (lsp--buffer-for-file (lsp--uri-to-path uri))))
     (remove-overlays (point-min) (point-max) 'lsp-dart-closing-labels t)
-    (seq-doseq (label-ht labels)
-      (save-excursion
-        (-let* ((label (gethash "label" label-ht))
-                (range (gethash "range" label-ht))
-                ((beg . end) (lsp--range-to-region range))
-                (end-line (progn
-                            (goto-char end)
-                            (line-end-position)))
-                (overlay (make-overlay beg end-line buffer)))
-          (overlay-put overlay 'lsp-dart-closing-labels t)
-          (overlay-put overlay 'after-string (propertize (concat lsp-dart-closing-labels-prefix " " label)
-                                                         'display `((height ,lsp-dart-closing-labels-size))
-                                                         'cursor t
-                                                         'font-lock-face 'font-lock-comment-face)))))))
+    (when buffer
+      (seq-doseq (label-ht labels)
+        (save-excursion
+          (-let* ((label (gethash "label" label-ht))
+                  (range (gethash "range" label-ht))
+                  ((beg . end) (lsp--range-to-region range))
+                  (end-line (progn
+                              (goto-char end)
+                              (line-end-position)))
+                  (overlay (make-overlay beg end-line buffer)))
+            (overlay-put overlay 'lsp-dart-closing-labels t)
+            (overlay-put overlay 'after-string (propertize (concat lsp-dart-closing-labels-prefix " " label)
+                                                           'display `((height ,lsp-dart-closing-labels-size))
+                                                           'cursor t
+                                                           'font-lock-face 'font-lock-comment-face))))))))
 
 (lsp-register-client
  (make-lsp-client :new-connection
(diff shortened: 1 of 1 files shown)


ericdallo:
@yyoncho, Can you help review it and confirm please if `lsp-treemacs` has support for the `outlines` and `flutterOutlines` [features](https://dartcode.org/releases/v3-3/#preview-flutter-outline)?.

I'm also don't know if we can use this `closingLabels` feature like [vscode](https://user-images.githubusercontent.com/17878/40066976-4b9cf28c-5833-11e8-812f-13ef733bc172.png).

yyoncho:
> @yyoncho, Can you help review it and confirm please if `lsp-treemacs` has support for the `outlines` and `flutterOutlines` [features](https://dartcode.org/releases/v3-3/#preview-flutter-outline)?.

It does not have support for those. Note that these are dart protocol extension and non-standard features so they should be handled separately. 

> I'm also don't know if we can use this closingLabels feature like vscode.

We can - but we have to create handlers in lsp-dart.el for them. If I understand the feature correctly, lsp-rust.el has similar feature so you could take a look.

Let me know if you want to work on implementing those extensions. If you want, feel free to ask for help/pointers/etc here or in gitter channel.

ericdallo:
@yyoncho, Thanks for the help!
I followed `lsp-rust` and searched a lot on net and finally make it works :)
Can you review it please?

ericdallo:
I realized that emacs behavior on overlay is different from vscode. Do you know how fix this @yyoncho?


Emacs:
closing-labels-emacs
Vscode
closing-labels-vscode

yyoncho:
> I realized that emacs insert mode behavior on overlay is different from vscode. Do you know if we can fix this @yyoncho?

I would try using `intangible` as per https://www.gnu.org/software/emacs/manual/html_node/elisp/Special-Properties.html#Special-Properties . 

If it doesn't work I suggest asking in emacs-devel mailing list. Generally, we could create ad-hoc solution but I think that emacs might have a way to handle that case.

Other than that, the PR looks good to me.

ericdallo:
@yyoncho, I tried `intangible`, `cursor-intangible` but it didn't fix it, but I got it to work with `cursor` property :D
I think it's good to go now :rocket:

lsp-dart.el
```diff
@@ -65,6 +88,27 @@ imported into the current file. Defaults to true"
      ,(expand-file-name (f-join lsp-dart-sdk-dir "bin/snapshots/analysis_server.dart.snapshot"))
      "--lsp")))
 
+(defun lsp-dart--handle-closing-labels (_workspace params)
+  "Closing labels notification handling.
+PARAMS closing labels notification data sent from WORKSPACE."
+  (-let* (((&hash "uri" "labels") params)
+          (buffer (get-file-buffer (string-remove-prefix "file://" uri))))
```
yyoncho, line 95: use `lsp--uri-to-path`. Also, I suggest checking if the buffer is still open to avoid reopening the buffer if it has been closed.

ericdallo: Do you mean checking if the buffer still exists?

ericdallo: I found `lsp--buffer-for-file`, I think this is the right t apply here, right?

yyoncho: > I found `lsp--buffer-for-file`, I think this is the right t apply here, right?

Yes. You should check if the buffer is nil before using it.

ericdallo: Done!

lsp-dart.el
```diff
@@ -65,6 +88,28 @@ imported into the current file. Defaults to true"
      ,(expand-file-name (f-join lsp-dart-sdk-dir "bin/snapshots/analysis_server.dart.snapshot"))
      "--lsp")))
 
+(defun lsp-dart--handle-closing-labels (_workspace params)
+  "Closing labels notification handling.
+PARAMS closing labels notification data sent from WORKSPACE."
+  (-let* (((&hash "uri" "labels") params)
+          (buffer (lsp--buffer-for-file (lsp--uri-to-path uri))))
+    (remove-overlays (point-min) (point-max) 'lsp-dart-closing-labels t)
```
yyoncho, line 96: Sorry for not providing all information in the review. It should be:

```
(when buffer
  (with-current-buffer buffer 
     (remove-overlays ...
```
otherwise, the call will be executed in the context of the current-buffer which is not necessary to be the same as `uri`.

ericdallo: It makes sense, Done!

yyoncho:

lsp-dart.el
yyoncho: Thank you!

PR #1562, opened 2020-04-04
lsp-capf: apply cache when prefix is before (point)

kiennq:
This is done to prevent cache is applied wrongly and make `lsp-completion-at-point` throws.
- Only apply cache when the cached prefix point is before (point)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4240,7 +4240,8 @@ Also, additional data to attached to each candidate can be passed via PLIST."
           (done? result)
           ((and lsp--capf-cache
                 (s-prefix? (car lsp--capf-cache)
-                           (buffer-substring-no-properties bounds-start (point))))
+                           (buffer-substring-no-properties bounds-start (point)))
+                (< (caar (cadr lsp--capf-cache)) (point)))
            (apply #'lsp--capf-filter-candidates (cdr lsp--capf-cache)))
           (t
            (-let* ((resp (lsp-request-while-no-input "textDocument/completion"


lsp-mode.el
```diff
@@ -4240,7 +4240,9 @@ Also, additional data to attached to each candidate can be passed via PLIST."
           (done? result)
           ((and lsp--capf-cache
                 (s-prefix? (car lsp--capf-cache)
-                           (buffer-substring-no-properties bounds-start (point))))
+                           (buffer-substring-no-properties bounds-start (point)))
+                (< (car (caadr lsp--capf-cache)) (point)))
+           (message "from cache")
```
yyoncho, line 4245: leftover?

lsp-mode.el
```diff
@@ -4255,9 +4257,9 @@ Also, additional data to attached to each candidate can be passed via PLIST."
                                                 item)
                                        item))))
                    (prefix-line (buffer-substring-no-properties (point-at-bol) (point))))
-             (setf done? (or (seqp resp)
-                             (not (gethash "isIncomplete" resp)))
-                   lsp--capf-cache (when (and done? (not (seq-empty-p items)))
+             (setf done? (or (seqp resp) (not (gethash "isIncomplete" resp)))
+                   lsp--capf-cache (when (and done? (not (seq-empty-p items))
+                                              (not (equal bounds-start (point))))
```
yyoncho, line 4262: can you elaborate on that? What will happen if I do not have a prefix and manually invoke the completion? Will it reuse the cache after the first invocation?

lsp-mode.el
kiennq: Removed

lsp-mode.el
kiennq: No, hmm think again this one is not really needed, I will remove this.

yyoncho:
Thank you!

PR #1563, opened 2020-04-04
lsp-capf: sending completion context

### LINKED ISSUE #1555: [lsp-clangd] Strange completion after right angle bracket
**My Config**
```elisp
(use-package company
  :ensure t
  :hook (prog-mode . company-mode)
  :custom
  (company-idle-delay 0)
  (company-echo-delay 0)
  (company-show-numbers t)
  (company-require-match nil)
  (company-minimum-prefix-length 3)
  (company-tooltip-align-annotations t)
  (company-backends '(company-capf)))

(use-package lsp-mode
  :ensure t
  :hook (prog-mode . lsp-deferred)
  :custom
  (lsp-keymap-prefix "C-c l")
  (lsp-enable-indentation nil)
  (lsp-enable-on-type-formatting nil)
  (lsp-auto-guess-root t)
  (lsp-prefer-capf t)              
  (lsp-keep-workspace-alive nil)
  (lsp-eldoc-enable-hover nil)
  (lsp-signature-auto-activate nil))
```

**Describe the bug**
Code completion after a single `>` character.

image

**To Reproduce**
Open a `a.cpp` with the following content:

```cpp
int main() {
    
}
```

type `>` in the `main` function.

A single `:` meets the same issue.

image

**Expected behavior**

No completion.

`lsp-clangd`

# Background

It has been discussed in 

- https://github.com/MaskRay/ccls/issues/144
- https://github.com/joaotavora/eglot/issues/173

kiennq:
Sending `CompletionContext`, fix #1555

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4062,6 +4062,11 @@ and the position respectively."
 
 (defalias 'lsp--cur-line-diagnotics 'lsp-cur-line-diagnostics)
 
+(defconst lsp--completion-trigger-kinds
+  `((invoked . 1)
+    (character . 2)
+    (incomplete . 3)))
+
 (defun lsp--make-completion-item (item &rest plist)
   "Make completion item from lsp ITEM and PLIST."
   (-let (((&hash "label"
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4217,6 +4228,20 @@ Also, additional data to attached to each candidate can be passed via PLIST."
...
+(defun lsp--capf-get-context (trigger-characters)
+  "Get completion context with provided TRIGGER-CHARACTERS."
+  (let* (trigger-char
+         (trigger-kind (cond
+                        ((setq trigger-char (lsp--looking-back-trigger-characters-p
+                                             trigger-characters))
+                         'character)
+                        ((equal lsp--capf-cache 'incomplete) 'incomplete)
+                        (t 'invoked))))
+    (ht<-alist
```
yyoncho, line 4240: do we need this conversion?

kiennq: Yes, the idea is to only include `triggerCharacter` when it existed. Some server like `clangd` basically just return nil if we pass `triggerCharacter` with nil

yyoncho: I don't understand how converting alist to ht will help for that?

kiennq: The `append` will not add the `nil` element to `alist`. Then we we can convert that list to `ht` with only non-nil member.
Is there a better way to write that? I don't think converting from a list to ht take more time, especially since these operations are not run that frequently

yyoncho: I have approved the PR as it is. 

> Then we we can convert that list to ht with only non-nil member.

Append part is fine - my question is solely about the alist->ht conversion. AFAICS this conversion does not strip nils. 

(But in general, feel free to merge it as it is).

kiennq: Thanks, the main part of this is `append` to not include nil. then `ht<-alist` just goes ahead and do the conversion like what `ht` does.

Issue #1559, opened 2020-04-03
Rust analyzer: out-of-date initialization options


hghwng:
**Describe the bug**
`lsp-rust-analyzer--make-init-options` in [lsp-rust.el](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-rust.el#L399) does not match [rust-analyzer's config parser](https://github.com/rust-analyzer/rust-analyzer/blob/47a48505a54a08939ef7958243bf8d87aa045bd4/crates/rust-analyzer/src/config.rs#L113).

See also: [VSC's config keys](https://github.com/rust-analyzer/rust-analyzer/blob/09b8ee30bab7040ec983968aaca90814029f4dc0/editors/code/package.json)

**To Reproduce**
For example, set `lsp-rust-all-targets` to `t`, and functions in `cfg(test)` won't get diagnostics.

**Expected behavior**
Unit tests get diagnostics.

**Which Language Server did you use**
lsp-rust

**OS**
Arch Linux; emacs 26.3-1; Spacemacs 1d797ffa4a57ec9b4f46f0cfd22652141c2559d1; lsp-mode 20200315.

brotzeit:
Thanks for the bug report! Do you want to open a pull request ?

brotzeit:
@hghwng Test #1568 please.

hghwng:
@brotzeit It works! The lsp-log buffer (with `lsp-log-io`) also agrees with me.

(And ... sorry for the late reply :-( － I could have opened a pull request)

brotzeit:
No problem. You can always open pull requests to improve stuff.

PR #1564, opened 2020-04-05
lsp-rust WIP: Add rust-analyzer trace server option

apiraino:
Hi @brotzeit can you help me refine and finish this PR? I can't code in lisp :-)

I'd love to contribute a bit to #1559 , I need to trace the rust-analyzer server but I dont know how to go on from here.

My reference for the trace option is [here](https://github.com/rust-analyzer/rust-analyzer/blob/09b8ee30bab7040ec983968aaca90814029f4dc0/editors/code/package.json#L363)

thank you so much!

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -396,6 +396,15 @@ PARAMS progress report notification data."
   :group 'lsp-rust
   :package-version '(lsp-mode . "6.2.2"))
 
+(defcustom lsp-rust-analyzer-server-trace "off"
+  "Trace rust-analyzer server."
+  :type '(choice
+          (const "off")
+          (const "messages")
+          (const "verbose"))
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "6.2.2"))
+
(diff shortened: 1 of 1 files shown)


kiennq:
What's the additional information this one will log compare to set `lsp-log-io` to `t`?

apiraino:
> What's the additional information this one will log compare to set `lsp-log-io` to `t`?

Could be but I have no idea :-) I'm still figuring out all the available options. I've been suggested to do so to increase the amount of info when I need to debug and provide useful info what's happening. 

But now that I look closely at the standard `lsp-mode` options I see a `lsp-server-trace` flag and it seems to do what I need. If yes, I can happily close this PR as it's not really needed.

apiraino:
ok, after some more tests, I've understood that this feature is already implemented (silly me that I did not read more carefully the documentation)

brotzeit:
@apiraino No problem. Can you do me a favor and test  #1568

PR #1568, opened 2020-04-07
update and add new rust-analyzer customization options

brotzeit:


cassandracomar:
I think the chain hints need to be rendered here: https://github.com/emacs-lsp/lsp-mode/pull/1568/files#diff-e196a72bcbed4c56a4d30daf13708f64R569-R575

apiraino:
Hi @brotzeit I've tried the lsp-rust.el from this branch. I think I didn't configure properly my lsp-mode because I don't see any difference with a previous lsp-rust.el (from `lsp-mode-20200406.615`).

The relevant parts in my `custom.el` are:
```
+ '(lsp-rust-analyzer-display-chaining-hints t)
+ '(lsp-rust-analyzer-display-parameter-hints t)
+ '(lsp-rust-analyzer-server-display-inlay-hints t)
```

but this is what I see when I enable and disable the inline hints
saved

I'd like to help testing but I'm missing some context. What am I supposed to see? Is even my test case correct?

Can you help me clarifying a bit? :-) thanks

brotzeit:
@cassandracomar I'm not sure I understand what you mean =)

@apiraino I guess you don't see any differences because your values are the default values of rust-analyzer, which we didn't configure before with lsp-mode ?

I will merge this now and if anybody encounters new issues please open another issue. Or even better, a pull request ;)

PR #1570, opened 2020-04-07
Remove dls, as deprecated

kubo39:
A bolt from the blue.
https://forum.dlang.org/post/usqhsdtwfhrterzpstst@forum.dlang.org


yyoncho:
Thank you!

PR #1573, opened 2020-04-08
[lsp-dart] Add outline and flutterOutline options

ericdallo:
image

### DIFF
--- a/lsp-dart.el
+++ b/lsp-dart.el
@@ -87,6 +87,13 @@ be sent with outline information for open files. Defaults to t"
   :group 'lsp-dart
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-dart-flutter-outline t
+  "When set to non-nil, dart/textDocument/publishFlutterOutline notifications will
+be sent with Flutter outline information for open files. Defaults to t"
+  :type 'boolean
+  :group 'lsp-dart
+  :package-version '(lsp-mode . "6.3"))
+
 (defcustom lsp-dart-outline-position-params
   `((side . ,treemacs-position)
     (slot . 2)
@@ -97,13 +104,15 @@ Defaults to side following treemacs default."
   :group 'lsp-dart
   :package-version '(lsp-mode . "6.3"))
 
-(defun lsp-dart--get-current-outline ()
-  "Returns the current outline data on current workspace."
-  (lsp-workspace-get-metadata "current-outline" (first (lsp-workspaces))))
-
-(defun lsp-dart--set-current-outline (workspace current-outline)
-  "Set CURRENT-OUTLINE on WORKSPACE metadata."
-  (lsp-workspace-set-metadata "current-outline" current-outline workspace))
+(defcustom lsp-dart-flutter-outline-position-params
+  `((side . ,treemacs-position)
+    (slot . 2)
+    (window-width . ,treemacs-width))
+  "The Flutter outline tree position params.
+Defaults to side following treemacs default."
+  :type 'list
+  :group 'lsp-dart
+  :package-version '(lsp-mode . "6.3"))
 
 (defun lsp-dart--outline-kind->icon (kind)
   "Maps an outline KIND to a treemacs icon symbol.
@@ -132,7 +141,8 @@ Kinds from https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool
     ("UNIT_TEST_GROUP" 'structure)
     ("UNIT_TEST_TEST" 'method)
     ("ENUM" 'enumerator)
-    ("ENUM_CONSTANT" 'enumitem)))
+    ("ENUM_CONSTANT" 'enumitem)
+    ("NEW-INSTANCE" 'flutter)))
 
 (defun lsp-dart--outline-tree-ret-action (uri range)
   "Build the ret action for and item in the outline tree view.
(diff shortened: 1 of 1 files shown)


lsp-dart.el
```diff
@@ -80,6 +80,154 @@ Defaults to 0.9"
...
+          (window (display-buffer-in-side-window buffer lsp-dart-outline-position-params)))
+    (when (null focus?)
+      (select-window window)
+      (set-window-dedicated-p window t))))
+
+(defun lsp-dart--show-flutter-outline (focus?)
+  "Shows an Flutter outline tree.
+Focus on it if FOCUS? is non-null."
+  (-let* (((&hash "uri" "outline" (&hash "children")) (lsp-workspace-get-metadata "current-flutter-outline"
+                                                                                  (first (lsp-workspaces))))
```
yyoncho, line 224: `(first (lsp-workspaces))` - `lsp-find-workspace`

lsp-dart.el
```diff
@@ -120,10 +284,26 @@ PARAMS closing labels notification data sent from WORKSPACE."
...
+                    (closingLabels . ,lsp-dart-closing-labels)
+                    (outline . ,lsp-dart-outline)
+                    (flutterOutline . ,lsp-dart-flutter-outline))
+                  :notification-handlers (ht ("dart/textDocument/publishClosingLabels" 'lsp-dart--handle-closing-labels)
+                                             ("dart/textDocument/publishOutline" 'lsp-dart--handle-outline)
+                                             ("dart/textDocument/publishFlutterOutline" 'lsp-dart--handle-flutter-outline))
                   :server-id 'dart_analysis_server))
 
+;;;###autoload
+(with-eval-after-load 'lsp-treemacs
```
yyoncho, line 296: all pieces that refer to lsp-treemacs should be wrapped in `(with-eval-after-load` . Without this, the travis will fail.

lsp-dart.el
```diff
@@ -120,10 +284,26 @@ PARAMS closing labels notification data sent from WORKSPACE."
...
-                  :notification-handlers (ht ("dart/textDocument/publishClosingLabels" 'lsp-dart--handle-closing-labels))
+                    (closingLabels . ,lsp-dart-closing-labels)
+                    (outline . ,lsp-dart-outline)
+                    (flutterOutline . ,lsp-dart-flutter-outline))
+                  :notification-handlers (ht ("dart/textDocument/publishClosingLabels" 'lsp-dart--handle-closing-labels)
+                                             ("dart/textDocument/publishOutline" 'lsp-dart--handle-outline)
+                                             ("dart/textDocument/publishFlutterOutline" 'lsp-dart--handle-flutter-outline))
                   :server-id 'dart_analysis_server))
 
+;;;###autoload
```
yyoncho, line 295: I don't think this autoload is needed.

lsp-dart.el
```diff
@@ -80,6 +80,154 @@ Defaults to 0.9"
...
+   "*Flutter Outline*"))
+
+(defun lsp-dart--show-outline (focus?)
+  "Shows an outline tree.
+Focus on it if FOCUS? is non-null."
+  (-let* (((&hash "uri" "outline" (&hash "children")) (lsp-workspace-get-metadata "current-outline"
+                                                                                  (first (lsp-workspaces))))
+          (buffer (lsp-dart--build-outline-tree uri children))
+          (window (display-buffer-in-side-window buffer lsp-dart-outline-position-params)))
+    (when (null focus?)
```
yyoncho, line 216: this `null` seems to be redundant.

lsp-dart.el
```diff
@@ -80,6 +80,154 @@ Defaults to 0.9"
...
+  (seq-map (-lambda ((&hash "children"
+                            "element" (&hash "kind" "name" "parameters" "range")))
+             (let ((label (concat name (when parameters
+                                         (propertize (concat " " parameters)
+                                                     'face 'lsp-lens-face)))))
+               (list :key label
+                     :label label
+                     :icon (lsp-dart--outline-kind->icon kind)
+                     :children (unless (seq-empty-p children)
+                                 (lsp-dart--outline->tree uri children))
```
yyoncho, line 169: in general, you can pass a function to :children to make the conversion lazy. You may skip this one if you do not expect this function to be called often with big data.

lsp-dart.el
```diff
@@ -80,6 +80,154 @@ Defaults to 0.9"
...
+    (slot . 2)
+    (window-width . ,treemacs-width))
+  "The outline tree position params.
+Defaults to side following treemacs default."
+  :type 'list
+  :group 'lsp-dart
+  :package-version '(lsp-mode . "6.3"))
+
+(defcustom lsp-dart-flutter-outline-position-params
+  `((side . ,treemacs-position)
```
yyoncho, line 108: you cannot use treemacs-position here like that - since we do not have a dependency on it.

ericdallo: Wrapped into a `with-eval-after-load`

lsp-dart.el
ericdallo: Done, I don't know if I need to wrap the "private" functions given that the public one is already wrapped with `with-eval-after-load`

yyoncho: everything that is undefined will cause test failures no matter if it is public or private.

yyoncho:
Thank you!

ericdallo:
Thank you for all the help @yyoncho !

PR #1576, opened 2020-04-09
[lsp-dart] Fix kind for flutter icon

ericdallo:
* Check if tree item is widget [just like VSCode](https://github.com/Dart-Code/Dart-Code/blob/97c7e1629889f1c9eda7ac4999958e06abf70a4c/src/extension/flutter/flutter_outline_view.ts#L342-L350)

Fix with https://github.com/emacs-lsp/lsp-treemacs/pull/38


yyoncho:
Thank you!

PR #1585, opened 2020-04-15
Activate flow language server if there is flow tag in file or .flowconfig in project

jaaso:


test/fixtures/SampleFlowProject/src/sameple-without-flow-comment.js
```diff
@@ -0,0 +1,5 @@
+function square(n: number): number {
```
danielmartin, line 1: There's a typo in this filename: `sameple-without-flow-comment.js`.

test/lsp-clients-test.el
```diff
@@ -69,6 +69,18 @@
   (let ((major-mode 'js-mode))
     (should (lsp-clients-flow-activate-p (concat test-location "fixtures/SampleFlowProject/src/sample.js") nil))))
 
+(ert-deftest lsp-flow-should-activate-on-flow-project-without-flow-file-comment ()
+  ;; Set `js-mode' ON and check that a Flow project activates the Flow but without flow tag
```
danielmartin, line 73: I think the test name is self-explanatory and we can just remove this comment. The same for the other test.

danielmartin:
Thanks for your contribution!

PR #1586, opened 2020-04-15
Process the $/progress messages from LSP 3.15

### LINKED ISSUE #1427: Add support for Progress Notifications
Are there any plans to add support for `$/progress` notifications, as introduced by version `3.15` of the protocol?

alanz:
Fixes #1427

This is a bit of a draft, it works for the [erlang_ls](https://github.com/erlang-ls/erlang_ls), which reports percentages, but not for [haskell-language-server](https://github.com/haskell/haskell-language-server) which does not.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1499,6 +1499,74 @@ PARAMS - the data sent from WORKSPACE."
         (completing-read (concat message " ") (seq-into choices 'list) nil t)
       (lsp-log message))))
 
+(defun lsp--on-progress (workspace params)
+  "Callback for $/progress.
+type ProgressToken = number | string;
+interface ProgressParams<T> {
+    token: ProgressToken;
+    value: T;
+}
+PARAMS contains the progress data.
+WORKSPACE is the workspace that contains the progress token."
+   (let* ((token (gethash "token" params))
+          (value (gethash "value" params))
+          (kind (gethash "kind" value))
+          (client (lsp--workspace-client workspace)))
+     (lsp-log "progress: %s-%S" token kind)
+     (pcase kind
+
+       ("begin"
+        (let* ((message (gethash "title" value))
+               (cur (gethash "percentage" value 0))
+               (reporter
+                (make-progress-reporter message 0 100 cur)))
+          (lsp-workspace-set-work-done-token token reporter workspace)))
+
+       ("report"
+        (let ((reporter (lsp-workspace-get-work-done-token token workspace)))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1498,6 +1498,74 @@ PARAMS - the data sent from WORKSPACE."
...
+interface ProgressParams<T> {
+    token: ProgressToken;
+    value: T;
+}
+PARAMS contains the progress data.
+WORKSPACE is the workspace that contains the progress token."
+   (let* ((token (gethash "token" params))
+          (value (gethash "value" params))
+          (kind (gethash "kind" value))
+          (client (lsp--workspace-client workspace)))
```
yyoncho, line 1513: This client is unused - I think this is causing the CI failures.

lsp-mode.el
```diff
@@ -2518,6 +2585,10 @@ BINDINGS is a list of (key def cond)."
   ;; current workspace in format filePath->file notification handle.
   (watches (make-hash-table :test 'equal))
 
+  ;; contains all the workDone progress tokens that have been created
+  ;; for the current workspace.
+  (work-done-tokens (make-hash-table :test 'equal))
```
yyoncho, line 2590: can you move that in the last position?

lsp-mode.el
```diff
@@ -5820,8 +5911,12 @@ WORKSPACE is the active workspace."
                                      (list :uri (lsp--path-to-uri folder))))
                              (apply #'vector))))
                      ((string= method "window/workDoneProgress/create")
-                      ;; Ignoring for now
-                      nil
+                      ;; Not sure why this request exists, the
```
yyoncho, line 5914: most likely this has to be deleted.

yyoncho:
Thank you!

robertoaloi:
@alanz It looks like you can invoke the progress reporter by setting `min-value` and `max-value` to `nil` to get a rotating spinner. That would be useful for cases (even in Erlang LS) where the total time is not known in advance (e.g. when running tests via a code lens). This is actually part of the protocol itself, where the percentage value in the `begin` message is optional and should be used exactly for this purpose. So I suggest that, if the `begin` message does not contain a percentage, we fall back to `nil` (instead of `0`), both for the `min-value` and `max-value`. 

https://www.gnu.org/software/emacs/manual/html_node/elisp/Progress.html

robertoaloi:
Created #1599 for this.

PR #1589, opened 2020-04-17
lsp-solargraph: use bundler per project

amake:
This is a follow up to [this discussion](https://www.reddit.com/r/emacs/comments/g2ayzc/lspmode_and_ruby/fnn6ike/).

I would like to be able to use lsp-solargraph with bundler on a per-project basis, e.g. via a dir-locals file.

This PR extracts the solargraph launch command building into a separate function so that overriding `lsp-solargraph-use-bundler` can take effect.

In addition it sets `lsp-solargraph-use-bundler` to "safe" if it is a boolean. I don't think this should be a security risk, but I also note that almost no `lsp-*` variables are marked as "safe" so if it's against policy then please let me know.

Also I don't know how this does or should affect the value of `"solargraph.useBundler"` passed to `lsp-register-custom-settings`. In my testing things worked fine, despite this value being potentially out of sync depending on the project.

### DIFF
--- a/lsp-solargraph.el
+++ b/lsp-solargraph.el
@@ -115,6 +115,13 @@
   :group 'lsp-solargraph
   :package-version '(lsp-mode . "6.3"))
 
+(defun lsp-solargraph--build-command ()
+  "Build solargraph command"
+  (let ((lsp-command '("solargraph" "stdio")))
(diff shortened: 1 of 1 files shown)


yyoncho:

Issue #1590, opened 2020-04-17
gopls panics because file paths are not properly converted to URIs


codesuki:
lsp-mode uses the the function url-hexify-string to convert a path to an URI.
Following down the rabbit hole you find this line
;; Allow % to avoid re-encoding %-encoded sequences.
https://github.com/emacs-mirror/emacs/blob/master/lisp/url/url-util.el#L415

The following PR would solve the issue I think, but it breaks with emacs < 26.x
https://github.com/emacs-lsp/lsp-mode/pull/1089

This PR to gopls that broke it.
https://github.com/golang/tools/commit/f8e42dc47720b31a0fedf0b422ee4cbd171ab163

yyoncho:
> The following PR would solve the issue I think, but it breaks with emacs < 26.x
> #1089

We can make that conditional.

yyoncho:
Can you include sample url which will break lsp-mode?

codesuki:
I don't have the log anymore but Bazel creates paths like `/var/.../architecture%/test_main.go`. Then gopls panicks with `%/t` is not a valid URI escape.

PR #1593, opened 2020-04-17
fix: golsp panicking because of % in Bazel paths

codesuki:
fixes: https://github.com/emacs-lsp/lsp-mode/issues/1590

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1299,12 +1299,19 @@ On other systems, returns path without change."
   "Implemented only to make `company-lsp' happy.
 DELETE when `lsp-mode.el' is deleted.")
 
+(defconst url-path-allowed-chars-without-percent
+  (let ((vec (copy-sequence url-path-allowed-chars)))
+    (aset vec ?% nil)
+    vec)
+  "url-path-allowed-chars without percentage sign. Percentage
+  sign breaks gopls.")
+
 (defun lsp--path-to-uri-1 (path)
   (concat lsp--uri-file-prefix
           (--> path
                (expand-file-name it)
                (or (file-remote-p it 'localname t) it)
-               (url-hexify-string it url-path-allowed-chars))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1299,12 +1299,19 @@ On other systems, returns path without change."
   "Implemented only to make `company-lsp' happy.
 DELETE when `lsp-mode.el' is deleted.")
 
+(defconst url-path-allowed-chars-without-percent
```
kiennq, line 1302: We should prefix this with `lsp--` to prevent global namespace collision

lsp-mode.el
```diff
@@ -1299,12 +1299,19 @@ On other systems, returns path without change."
...
+  (let ((vec (copy-sequence url-path-allowed-chars)))
+    (aset vec ?% nil)
```
kiennq, line 1304: I think we can go further and instead of just removing `%`, remove all other additional delimiter characters except `/` from `url-path-allowed-chars`.
Since we are just hexifying local path, the union of `url-unresvered-chars` and `\` should be enough.
Something like https://github.com/kiennq/lsp-mode/commit/60947473381eb39fa0153fe3bb8680e27a2035a5
And it passes the travis build too.

codesuki: I'll try if this works with gopls. I tried with the default value before and it didn't work.

lsp-mode.el
codesuki: agree

lsp-mode.el
kiennq: Yes, I think it's because it normalizes the `/` so the path is not correct anymore.

codesuki: Yes, that was the problem iirc.

codesuki:
@kiennq works! thanks for the improvements.

kiennq:
LGTM

PR #1595, opened 2020-04-18
fix: don't unnecessarily sending nil rangeLimit

### LINKED ISSUE #1591: rangeLimit is nil
In `lsp-julia` `(lsp--client-capabilities)` returns the following:
```elisp
...
 (textDocument
               ...
               (foldingRange (dynamicRegistration . t)
                             (rangeLimit)
                             (lineFoldingOnly . :json-false))
               ...)
...
```
`rangeLimit` should either have an integer value or should be missing.

https://github.com/julia-vscode/LanguageServer.jl/issues/631

kiennq:
Fix #1591


yyoncho:
Thank you!

PR #1597, opened 2020-04-19
Setting to disable adding company-lsp automatically

LouisKottmann:
Hello,

I was trying to setup company-backends to have a merged list for simplicity as I use few backends. For clarity and as an example, I want to keep my backends set up like so:

```lisp
(setq company-backends '((company-lsp company-keywords company-yasnippet company-files)
                         company-dabbrev))
```

When trying to set it up, I kept having `company-lsp` pushed at the top of `company-backends` whenever lsp was activated.

After a long time going through people's configurations I finally set to read _lsp-mode.el_, and I found that line:

https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L6525

Locally commenting it out proved that it was the culprit.

I then naturally set out to disable that line, browsing for lsp options. But there is no option to do so.

Hence this PR :)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -275,6 +275,13 @@ links, and so on. For finer granularity you may use `lsp-enable-*' properties."
   :type 'boolean
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-auto-configure-company-backends t
+  "When auto-configuring `lsp-mode', 
+  `company-lsp' is pushed at the top of the `company-backends'.
+  If that is not desirable, set this variable to nil"
+  :group 'lsp-mode
+  :type 'boolean)
+
 (defcustom lsp-disabled-clients nil
   "A list of disabled/blacklisted clients.
 Each entry in the list can be either:
@@ -6522,7 +6529,8 @@ returns the command to execute."
          (not lsp-prefer-capf))
     (progn
       (company-mode 1)
-      (add-to-list 'company-backends 'company-lsp)
+      (when lsp-auto-configure-company-backends
+        (add-to-list 'company-backends 'company-lsp))
       (setq-local company-backends (remove 'company-capf company-backends))))
 
    ((and (fboundp 'company-mode))


yyoncho:
Thank you for the PR. I would suggest a bit different approach to achieve the same which is symmetrical to what we do for flycheck/flymake. Here it is - instead of introducing one more setting reuse the existing one lsp-prefer-capf and use the value :none for indicating that no company backend should be configured. We could rename the variable to something like lsp-completion-backend. We did the same refactoring for diagnostics - we started with lsp-prefer-flymake and then changed it to lsp-diagnostics-package. What do you think?

LouisKottmann:
I think it works for flycheck/flymake because there is a finite number of options (2) and they are highly related: it is one or the other. I would say that company-capf & company-lsp are less related.

But you know better than me, if that's the way to go I can change the PR.

LouisKottmann:
Honestly I'm not sure how to best set it up.
What would you suggest the options be?

`:none` => lsp auto-config does not touch company-backends
`:auto` => If the company-lsp backend is present, add it and remove company-capf. Otherwise add company-capf (default, how it currently works)
`:no-company-lsp` => like `:auto` but does not add company-lsp
`:company-capf` => like `lsp-prefer-capf t`
`:no-company-capf` => like `lsp-prefer-capf nil`

That seems overly complicated to me, and some variations are not represented. I think company-capf and company-lsp are not related enough.

yyoncho:
I prefer to have fewer configuration variables when that is possible.

```
:none => lsp auto-config does not touch company-backends
:auto | nil | :company-lsp => If the company-lsp backend is present, add it and remove company-capf. Otherwise add company-capf (default, how it currently works) 
:company-capf | t => use capf
```

Few notes:
1. company-lsp is deprecated in favour of company-capf
2. Technically, what you want to achive now is possible by modifying the backends in lsp-mode-managed-mode-hook.

LouisKottmann:
Mmmh I had not understood that company-lsp was being deprecated in favor of company-capf, I thought it was the opposite.

After switching them, it seems to work fine.

At the moment I use `lsp-after-open-hook` to re-setup my backends and it works fine, I just wanted something cleaner and an easier setup for others.

Does this PR still make sense?

yyoncho:
> Mmmh I had not understood that company-lsp was being deprecated in favor of company-capf, I thought it was the opposite.

It used to be like that but @kiennq rewrote capf integration. `company-lsp` author is currently not active.

> At the moment I use `lsp-after-open-hook` to re-setup my backends and it works fine, I just wanted something cleaner and an easier setup for others.
> 
> Does this PR still make sense?

From my point of view, it makes sense to fix the inconsistency between lsp-prefer-capf and lsp-diagnostics-package but introducing a new variable is not justified considering that you can achieve the customization using lsp-mode hooks.

LouisKottmann:
I'll keep on using lsp-mode hooks then.

The README could use an update to tell people to not bother with company-lsp though.

yyoncho:
> The README could use an update to tell people to not bother with company-lsp though.

Yes.

PR #1600, opened 2020-04-20
[#1599] Progress reporting: enable spinner if not percentage.

### LINKED ISSUE #1599: Support unbound progress reports
In #1586 support for progress reports has been introduced, but they only work with percentages.

It looks like you can invoke the progress reporter by setting min-value and max-value to nil to get a rotating spinner. That would be useful for cases where the total time is not known in advance (e.g. when running tests via a code lens). This is actually part of the protocol itself, where the percentage value in the `begin` message is optional and should be used exactly for this purpose. So I suggest that, if the `begin` message does not contain a percentage, we fall back to `nil` both for the `min-value` (instead of `0` and for the `max-value` (instead of `100`).

https://www.gnu.org/software/emacs/manual/html_node/elisp/Progress.html

alanz:
Closes #1599

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1516,19 +1516,21 @@ WORKSPACE is the workspace that contains the progress token."
    (let* ((token (gethash "token" params))
           (value (gethash "value" params))
           (kind (gethash "kind" value)))
-     (lsp-log "progress: %s-%S" token kind)
      (pcase kind
 
        ("begin"
         (let* ((message (gethash "title" value))
-               (cur (gethash "percentage" value 0))
+               (cur (gethash "percentage" value nil))
                (reporter
-                (make-progress-reporter message 0 100 cur)))
+                (if cur
+                    (make-progress-reporter message 0 100 cur)
+                  ;; Spinner only
+                  (make-progress-reporter message nil nil))))
           (lsp-workspace-set-work-done-token token reporter workspace)))
 
        ("report"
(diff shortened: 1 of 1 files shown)


yyoncho:
@robertoaloi willing to test this?

alanz:
I am not able to test it with the erlang language server, think my config is broken. But intending to check it with haskell-language-server / ghcide, when I make sure the progress reporting is correct there.

robertoaloi:
@yyoncho @alanz Of course!

lsp-mode.el
```diff
@@ -1516,19 +1516,21 @@ WORKSPACE is the workspace that contains the progress token."
...
+                  (make-progress-reporter message nil nil))))
           (lsp-workspace-set-work-done-token token reporter workspace)))
 
        ("report"
         (let ((reporter (lsp-workspace-get-work-done-token token workspace)))
-          (progress-reporter-update reporter (gethash "percentage" value 0))))
+          (progress-reporter-update reporter (gethash "percentage" value nil))))
 
        ("end"
         (let ((reporter (lsp-workspace-get-work-done-token token workspace)))
```
alanz, line 1536: Should the commented out messages after this function be deleted?

robertoaloi: I would say yes.

robertoaloi:
It seems to play nicely with Erlang LS.

alanz:
I think this is ready to land now

yyoncho:
Thank you!

PR #1602, opened 2020-04-22
Protect against a missing progress reporter

alanz:
If we receive unexpected progress messages, make sure we actually have
a progress reporter before using it.

Closes #1601

PR #1603, opened 2020-04-22
Use when-let to simplify progress reporter code

alanz:
As mentioned in https://github.com/emacs-lsp/lsp-mode/pull/1602#pullrequestreview-398270661


yyoncho:
Thank you, the review comment was just for future reference :)

alanz:
I know, but I am not sure when I will touch this code again, so figured I may as well just do it now.

Also the best way to learn a thing is to actually do it.

PR #1609, opened 2020-04-25
Allow lsp-enable-which-key-integration to work for all modes

### LINKED ISSUE #1598: lsp-enable-which-key-integration does not work with vue-mode
**Describe the bug**
vue-mode uses mmm-mode to activate different major modes depending on where your cursor is. This causes issues with `lsp-enable-which-key-integration` since that function sets which-key descriptions only for the activated major mode.

Is there any downside to changing [this line](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L2371) to use `which-key-add-key-based-replacements` instead? I tried this change locally and my which-key descriptions worked as expected.

**To Reproduce**
This can be seen by pressing the lsp keymap prefix when the cursor is in a spot where the major mode is not vue-mode

I made a small example project in this branch: https://github.com/bimbiltu/lsp-playground/tree/vue-which-key-prefixes-not-named

Note that I set the lsp-keymap-prefix to "C-c l" in this branch. If you move the cursor
outside of the <script> block, where the major mode is vue-mode, which-key looks like this:
```
= → +formatting F → +folders G → +peek T → +toggle a → +code actions g → +goto h → +help r → +refactor s → +sessions
```

and if you move the cursor inside the <script> block, where the major mode is js-mode, which-key looks like this:
```
= → +prefix F → +prefix G → +prefix T → +prefix a → +prefix g → +prefix h → +prefix r → +prefix s → +prefix`
```

**Expected behavior**
It would be nice to call `lsp-enable-which-key-integration` and have it work over the entire .vue file

lsp-vetur

bimbiltu:
Fixes #1598

### DIFF
--- a/README.org
+++ b/README.org
@@ -344,12 +344,19 @@
    | ~s-l G i~  | Peek implementation locations of the symbol at point. (requires ~lsp-ui~)                                                  |
    | ~s-l G s~  | Peek symbols in the worskpace. (requires ~lsp-ui~)                                                                         |
 *** which-key integration
-     To enable [[https://github.com/justbur/emacs-which-key/][which-key]] integration put that in your config.
+    To enable [[https://github.com/justbur/emacs-which-key/][which-key]] integration in the active major mode for ~lsp-mode-map~
+    you will need to call the ~lsp-enable-which-key-integration~ function. This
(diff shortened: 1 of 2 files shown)


yyoncho:

PR #1613, opened 2020-04-25
Use new flycheck fringe bitmaps property if available

tspiteri:
flycheck's `:fringe-bitmap` can now be a cons of two bitmaps: the old one for a normal fringe and a new larger one for a 16-pixel fringe. The new cons can be obtained using `(get level 'flycheck-fringe-bitmaps)`, while `(get level 'flycheck-fringe-bitmap-double-arrow)` only returns the single bitmap. This PR will use the new system while falling back to the old system for older flycheck where getting `'flycheck-fringe-bitmaps` returns nil.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7685,12 +7685,14 @@ g. `error', `warning') and list of LSP TAGS."
                (category (--doto (intern (format "lsp-%s-category" name))
                            (setf (get it 'face) face
                                  (get it 'priority) 100)))
-               (new-level (intern name)))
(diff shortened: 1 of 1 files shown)


tspiteri:
cc @cpitclaudel

cpitclaudel:
LGTM, thanks a lot.

yyoncho:
@tspiteri thank you for contributing to `lsp-mode` and congrats for your first `lsp-mode` PR! :tada:

PR #1614, opened 2020-04-25
Add some missing markdown escapes

jsravn:
These show up a lot in gopls MarkupContent.

This is what gopls `FullDocumentation` currently renders as:

image

After this change:

image


yyoncho:
Thank you!

PR #1616, opened 2020-04-25
lsp-mode: Eliminate quadratic-time index-building for imenu.

### LINKED ISSUE #948: more efficient conversion of point to lsp position
**Describe the bug**
After opening a large file (github.com/aws/aws-sdk-go/service/s3/api.go) my Emacs becomes unresponsive, using 100% CPU. I tried waiting a minute, but it didn't seem to recover.

**To Reproduce**
Open above file and run `lsp`.

**Expected behavior**
I expected Emacs to work normally.

gopls

**OS**
Mac

**Error callstack**
I set `debug-on-quit` and got a stack trace:

https://gist.github.com/muirrn/40a9a4782c7e07c50c58bdcfe693bcca

leungbk:
Closes #948.

I've tested the non-hierarchical index-building by running on some Python files of my own, though I'm not sure how to test the hierarchical index-building since I'm not comfortable with any languages that have `imenu` entries with LSP children.

I can't tell if this provides any speedup for the file mentioned in #948 since even basic movement in that file is slow for me when `lsp-mode` is on and since I don't really know Golang.

I'd appreciate any suggestions on how I can make a minimal test for the hierarchical index-building.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -549,7 +549,7 @@ are determined by the index of the element."
 (defconst lsp--imenu-compare-function-alist
   (list (cons 'name #'lsp--imenu-compare-name)
         (cons 'kind #'lsp--imenu-compare-kind)
-        (cons 'position #'lsp--imenu-compare-position))
+        (cons 'position #'lsp--imenu-compare-line-col))
   "An alist of (METHOD . FUNCTION).
 METHOD is one of the symbols accepted by
 `lsp-imenu-sort-methods'.
@@ -6146,13 +6146,17 @@ WORKSPACE is the active workspace."
                             err)))
                workspace))))))))
 
+(defvar-local lsp--line-col-to-point-hash-table nil
+  "Hash table with keys (line . col) and values that are either point positions or markers.")
+
 (defun lsp--symbol-to-imenu-elem (sym)
   "Convert SYM to imenu element.
 
 SYM is a SymbolInformation message.
 
 Return a cons cell (full-name . start-point)."
-  (let* ((start-point (lsp--symbol-get-start-point sym))
+  (let* ((start-point (ht-get lsp--line-col-to-point-hash-table
+                              (lsp--get-line-and-col sym)))
          (name (gethash "name" sym))
          (container (gethash "containerName" sym)))
     (cons (if (and lsp-imenu-show-container-name container)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -6204,15 +6209,64 @@ SYM can be either DocumentSymbol or SymbolInformation."
...
-  (let ((symbols (lsp--get-document-symbols)))
-    (if (lsp--imenu-hierarchical-p symbols)
-        (lsp--imenu-create-hierarchical-index symbols)
-      (seq-map (lambda (nested-alist)
-                 (cons (car nested-alist)
-                       (seq-map #'lsp--symbol-to-imenu-elem (cdr nested-alist))))
-               (seq-group-by #'lsp--get-symbol-type (lsp--imenu-filter-symbols symbols))))))
+  (let* ((symbols (lsp--get-document-symbols))
+         (filtered-symbols (lsp--imenu-filter-symbols symbols))
+         (line-col-to-point-map (lsp--convert-line-col-to-points-batch (lsp--collect-lines-and-cols filtered-symbols))))
```
yyoncho, line 6265: nit: line-col-to-point-map seems to be redundant and event better to have `(let ((lsp--line-col-to-point-hash-table ...)` which will remove the need lsp--line-col-to-point-hash-table to be cleared explicitly. Also, consider passing it as parameter.

lsp-mode.el
```diff
@@ -6141,13 +6141,17 @@ WORKSPACE is the active workspace."
...
+  "Hash table with keys (line . col) and values that are either point positions or markers.")
+
 (defun lsp--symbol-to-imenu-elem (sym)
   "Convert SYM to imenu element.
 
 SYM is a SymbolInformation message.
 
 Return a cons cell (full-name . start-point)."
-  (let* ((start-point (lsp--symbol-get-start-point sym))
+  (let* ((start-point (ht-get lsp--line-col-to-point-hash-table
```
yyoncho, line 6153: is lsp--symbol-get-start-point still used?

lsp-mode.el
leungbk: OK, I used `(let ((lsp--line-col-to-point-hash-table` as you've suggested.

About passing it as a formal parameter: I initially passed it as a formal parameter to `lsp--imenu-create-hierarchical-index` and `lsp--symbol-to-hierarchical-imenu-elem`, and `lsp--imenu-create-non-hierarchical-index`. But the first two functions are mutually recursive, and I think it's confusing to have recursive functions calling themselves with the same arguments everytime.

leungbk:
> Looks good! What is the resulting speedup?

I tested this from `pyls` with a file that looks like

``` python
def fun1(args):
    pass

[...]

def fun30000(args):
    pass
```

and there wound up being a timeout while waiting for the response from the server.

I tried again on a shorter file of actual Python code (https://github.com/python/mypy/blob/125728a547f80ede4ecedc0ff7f890407d309caf/mypy/checker.py#L1) and didn't notice any difference between my branch and the master branch.

So it's possible that my branch doesn't make any noticeable difference. Alternatively, it may be the case that the files that actually benefit from this alternate index-build approach are too big for `pyls` to handle quickly. I know `pyls` is slow, but I'm not sure how to install Microsoft's `python-language-server` on NixOS.

@muirdm does this branch appear to help with the file you mentioned in the original issue? I'm not knowledgeable with Golang, and my computer is too slow to smoothly perform even basic movement in that file when `lsp-mode` is running `gopls`, let alone building an `imenu` index.

muirdm:
> does this branch appear to help with the file you mentioned in the original issue

Yes, much faster. imenu and menu-bar-mode both load almost instantly compared to spinning indefinitely before. Thanks!

yyoncho:
Thank you @leungbk and congrats for your first `lsp-mode` PR! :tada:

Out of curiosity, why did you pick this issue?

PR #1619, opened 2020-04-27
Add lsp-latex in readme.

ROCKTAKEY:
Thanks for maintenance!

lsp-latex is merged into MELPA. As said in [PR](https://github.com/ROCKTAKEY/lsp-latex/issues/8), I would like to add lsp-latex link to README.


yyoncho:
@ROCKTAKEY can you keep the current line and add separate one? 

cc @contrun

ROCKTAKEY:
Sorry, I got it.

PR #1620, opened 2020-04-27
Add defcustom to disable automatic LSP server download

hectorhon:
This defcustom will help the user avoid the prompt when opening a file for which the language server is not installed.

As an example, I install the JavaScript language servers locally for my npm projects as a dev dependency. I also add lsp-mode to the js-mode hook. As a result. when I visit random js files elsewhere, lsp-mode always prompts me to choose a language server, which can be annoying.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6744,6 +6744,12 @@ SESSION is the active session."
 
 ;; download server
 
+(defcustom lsp-server-allow-automatic-install t
+  "Whether to allow automatic install of a language server when
+one is not found."
+  :group 'lsp
+  :type 'boolean)
+
 (defcustom lsp-server-install-dir (expand-file-name
                                    (locate-user-emacs-file (f-join ".cache" "lsp")))
   "Directory in which the servers will be installed."
@@ -7547,9 +7553,10 @@ The server(s) will be started in the buffer when it has finished."
                   (cl-pushnew (current-buffer) (lsp--client-buffers client)))
                 clients))
        ;; look for servers to install
-       ((setq clients (lsp--filter-clients (-andfn #'lsp--matching-clients?
-                                                   #'lsp--client-download-server-fn
-                                                   (-not #'lsp--client-download-in-progress?))))
+       ((and lsp-server-allow-automatic-install
+             (setq clients (lsp--filter-clients (-andfn #'lsp--matching-clients?
+                                                        #'lsp--client-download-server-fn
+                                                        (-not #'lsp--client-download-in-progress?)))))
         (let ((client (lsp--completing-read
                        (concat "Unable to find installed server supporting this file. "
                                "The following servers could be installed automatically: ")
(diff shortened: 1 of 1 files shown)


kiennq:
We can already blacklist project folder, why don't you blacklist the folder contain those random js files?

hectorhon:
The files might not come from a specific folder. I cannot blacklist them if I do not know where they will be from in advance.

yyoncho:
Removing `:npm` from `lsp-deps-providers` should work too?

hectorhon:
I'm sorry, that did not work, I am still getting the prompt.

lsp-deps-providers is a variable defined in ‘lsp-mode.el’.
Its value is (:system (:path lsp--system-path))

nbfalcon:
This could also simplify some code in Doom Emacs: they have their own defcustom, +lsp-auto-install-syervers, and use defadvice to override automatic server installation. ($DOOM_GIT_CHECKOUT/modules/tools/lsp/+lsp.el).

frobware:
This is now pretty old but is there a means of doing this in what's current today?

nbfalcon:
@frobware 
```emacs-lisp
(defun +lsp-dont-auto-install-a (orig-fn &rest args)
  (or (catch 'lsp-dont-install
        (cl-letf (((symbol-function #'lsp--completing-read)
                   (lambda (&rest _)
                     (throw 'lsp-dont-install))))
          (apply orig-fn args) t))
      ;; alternatively do nothing here
      (user-error "No servers available (could be installed).")))
(eval-after-load 'lsp-mode (advice-add #'lsp :around #'+lsp-dont-auto-install-a))
```
should work (untested).

frobware:
> @frobware
> 
> ```emacs-lisp
> (defun +lsp-dont-auto-install-a (orig-fn &rest args)
>   (or (catch 'lsp-dont-install
>         (cl-letf (((symbol-function #'lsp--completing-read)
>                    (lambda (&rest _)
>                      (throw 'lsp-dont-install))))
>           (apply orig-fn args) t))
>       ;; alternatively do nothing here
>       (user-error "No servers available (could be installed).")))
> (eval-after-load 'lsp-mode '(advice-add #'lsp :around #'+lsp-dont-auto-install-a))
> ```
> 
> should work (untested).

Thanks for this - will try it out.

yyoncho:
Alternative solution:

```
(with-eval-after-load 'lsp-steep ;; lsp-steep is the last client loaded by lsp-mode
  (maphash (lambda (_ client)
             (setf (lsp--client-download-server-fn client) nil))
           lsp-clients))
```

yyoncho:
Closing the PR, since the issue could be solved outside of lsp-mode.

nbfalcon:
@yyoncho my solution was not intended to solve this, but as a workaround. Both yours and mine modify private lsp-mode variables, so I wouldn't consider them real `solutions`. Mine is especially hacky.

lsp-mode.el
```diff
@@ -7563,13 +7570,15 @@ The server(s) will be started in the buffer when it has finished."
...
 (If you have already installed the server check *lsp-log*)."
-                   (mapconcat (lambda (client)
-                                (symbol-name (lsp--client-server-id client)))
-                              clients
-                              " ")))
+                       (mapconcat (lambda (client)
+                                    (symbol-name (lsp--client-server-id client)))
+                                  clients
+                                  " "))
+          (lsp--warn "LSP server is not present and automatic install of LSP servers has been disabled. Check the variable lsp-server-allow-automatic-install.")))
```
nbfalcon, line 7581: Perhaps this should be "The following servers could be installed but automatic installation was disabled: %s" matching-clients

yyoncho: > private lsp-mode variables,

FWIW lsp--client-* should be made public to allow further client customizations.

PR #1625, opened 2020-04-30
Fix kotlin-language-server command

sei40kr:
## Description

`lsp-kotlin-language-server-path` is ignored by the server starting command.

* Make `lsp-kotlin-language-server-path` obsolete
* Add `lsp-clients-kotlin-server-executable` instead and set it `kotlin-language-server` by default


yyoncho:
Thank you!

PR #1627, opened 2020-04-30
Fix LSP Dart documentation

ericdallo:
* Fix old documentation about LSP Dart.
* I'm not sure if duplicate the variables from [here](https://github.com/emacs-lsp/lsp-dart#supported-settings) is a good approach.


yyoncho:
> I'm not sure if duplicate the variables from here is a good approach.

Can you elaborate?

ericdallo:
I noticed that we add to the docs the `defcustom`s from the LSP's, and I removed it [here](https://github.com/emacs-lsp/lsp-mode/pull/1627/files#diff-2d46c09f5c8b610f68836db17ecb2946L1) because it's on the `lsp-dart` README already.

PR #1607, opened 2020-04-23
Display the first line of MarkupContent in eldoc

jsravn:
Many LSP servers, like gopls, return `document/onHover` as
MarkupContent. This modifies the eldoc behavior to support MarkupContent
by displaying the first rendered line. As the majority of LSP servers
return the signature, this will display the signature as most people expect.

In the future, https://github.com/microsoft/language-server-protocol/issues/772 should make this more reliable.

PR #1635, opened 2020-05-01
Show function prototype in the echo area when Clangd is active

danielmartin:
Since https://github.com/emacs-lsp/lsp-mode/pull/1607, we show the first line of MarkupContent `onHover` in the echo area. For Clangd, in particular, the first line is not very representative.

This patch improves the experience using Clangd by extracting the part of the doc that corresponds to the type signature and showing that instead. The information is always condensed in one line to avoid the echo area growing uncomfortably.

This is in part a limitation of LSP and the protocol should offer this flexibility by default. Until the protocol is revised, I think it's OK we handle this client-side.

**Screenshots:**

Clangd showing the signature of a function from a call site:

Screenshot 2020-05-02 at 1 10 16 AM

Clangd showing the type of an auto variable:

Screenshot 2020-05-02 at 1 11 01 AM

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4779,6 +4779,34 @@ When language is nil render as markup if `markdown-mode' is loaded."
                               (or title command))
                             nil t))))
 
+(defun lsp-join-region (beg end)
+  "Apply join-line from BEG to END.
+This function is useful when an indented function prototype needs
+to be shown in a single line."
+  (save-excursion
+    (let ((end (copy-marker end)))
+      (goto-char beg)
+      (while (< (point) end)
+        (join-line 1)))
+    (s-trim (buffer-string))))
+
+(defun lsp--workspace-server-id (workspace)
+  "Return the server ID of WORKSPACE."
+  (-> workspace lsp--workspace-client lsp--client-server-id))
+
+(defun lsp--handle-rendered-for-echo-area (contents)
+  "Return a single line from RENDERED, appropriate for display in the echo area."
+  (pcase (lsp-workspaces)
+    (`(,workspace)
+     (lsp-clients-extract-signature-on-hover contents (lsp--workspace-server-id workspace)))
+    ;; For projects with multiple active workspaces we also default to
+    ;; render the first line.
+    (_ (lsp-clients-extract-signature-on-hover contents nil))))
+
+(cl-defgeneric lsp-clients-extract-signature-on-hover (contents _server-id)
+  "Extract a representative line from CONTENTS, to show in the echo area."
+  (car (s-lines (lsp--render-element contents))))
+
 (defun lsp--render-on-hover-content (contents render-all)
   "Render the content received from 'document/onHover' request.
 CONTENTS  - MarkedString | MarkedString[] | MarkupContent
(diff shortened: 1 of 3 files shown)


kiennq:
This look great, can it be implemented in `cl-defgeneric/cl-defmethod` style to completely separate change in `lsp-mode` and `lsp-clients`?

lsp-mode.el
```diff
@@ -4779,6 +4779,23 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
+  "Return the server ID of WORKSPACE."
+  (-> workspace lsp--workspace-client lsp--client-server-id symbol-name))
+
+(defun lsp--handle-rendered-for-echo-area (contents)
+  "Return a single line from RENDERED, appropriate for display in the echo area."
+  (pcase (lsp-workspaces)
+    (`(,workspace)
+     (if (string-equal "clangd" (lsp--workspace-server-id workspace))
+         (progn
+           (require 'lsp-clients)
```
yyoncho, line 4792: Can we restructure to remove that? I mean, one might not want to include lsp-clients (we have settings to avoid that) and in the end, the require will be forced? Also, the require of files that have lsp mode client registrations is slow due to lsp-register-custom-settings calls.

danielmartin:
> I think that we have something like that in the past but I cant find it (maybe for lsp-ui-sideline?).

The closest function I've found is https://github.com/emacs-lsp/lsp-ui/blob/master/lsp-ui-sideline.el#L214, but it doesn't do anything special to filter the hover information. For clangd the information shown is a little bit noisy (see screenshot), so maybe it makes sense to put this new logic somewhere where `lsp-ui-sideline` could also use it. `lsp-ui-sideline` is a little bit special, though, because it may make sense to show the function comment as well. WDYT?

Screenshot 2020-05-02 at 11 32 09 AM

@yyoncho Where do you think is the best place to put this? We have `lsp-cpp.el` for some extra C++ functionality regarding Flycheck. Should I move this there, with an autoload for the entry point?

lsp-mode.el
```diff
@@ -4779,6 +4779,23 @@ When language is nil render as markup if `markdown-mode' is loaded."
                               (or title command))
                             nil t))))
 
+(defun lsp--workspace-server-id (workspace)
+  "Return the server ID of WORKSPACE."
+  (-> workspace lsp--workspace-client lsp--client-server-id symbol-name))
```
kiennq, line 4784: Why return string instead of symbol?

yyoncho:
> `lsp-ui-sideline` is a little bit special, though, because it may make sense to show the function comment as well. WDYT?

Sounds good.

> @yyoncho Where do you think is the best place to put this? We have lsp-cpp.el for some extra C++ functionality regarding Flycheck. Should I move this there, with an autoload for the entry point?

I was thinking about making it less cpp specific(e. g. remove the language in regex) and put it lsp-mode.el. Alternatively, we may try the cl-defmethod as @kiennq suggested.

lsp-mode.el
kiennq: I think we can use `cl-defgeneric/cl-defmethod` here.
For example
``` elisp
(cl-defgeneric lsp-clients-extract-signature-on-hover (server-id contents)
  "Extract a representative line from CONTENTS, to show in the echo area."
  (car (s-lines (lsp--render-element contents))))
```
And put this in `lsp-clients.el`
``` elisp
(cl-defmethod lsp-clients-extract-signature-on-hover ((_server-id (eql clangd)) contents)
  ""
  ;; ...
  )
```
User will have to require `lsp-clients` to use `lsp-clangd` anyway, so the correct specializer will be loaded.

yyoncho:
And one more option: make the function to extract the docs configurable and provide recipes for extracting the info. After all, this is a hack/patch and it is not necessary to make it into lsp-mode.el.

danielmartin:
> And one more option: make the function to extract the docs configurable and provide recipes for extracting the info. After all, this is a hack/patch and it is not necessary to make it into lsp-mode.el.

I like this approach, to not make `lsp-mode` a "kitchen sink". This is also something that could potentially be standardised in the protocol in the future, so we'd need to remove it, anyway.

I think a `defcustom` where the function is invoked with the `contents` along with the `server-id` is OK, and we can show a sample Clangd-specific implementation in the documentation.

PR #1640, opened 2020-05-02
Case-insensitive completion for code actions

tspiteri:
Since code actions are not usually dependent on case, it would be more convenient to have their completion case-insensitive. If somehow there happen to be actions that differ only by case, `completing-read` would still allow the proper case to select the required action.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4773,11 +4773,12 @@ When language is nil render as markup if `markdown-mode' is loaded."
    ((seq-empty-p actions) (signal 'lsp-no-code-actions nil))
    ((and (eq (seq-length actions) 1) lsp-auto-execute-action)
     (lsp-seq-first actions))
-   (t (lsp--completing-read "Select code action: "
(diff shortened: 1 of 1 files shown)


yyoncho:
This looks more like personal preference - I would say that there are people that prefer faster narrowing by having case sensitive match. In your case, you may set a buffer local setting specific for lsp-mode buffers using lsp-mode-managed-mode hook or related or you want it only for the actions, you could wrap the code actions functions. Let me know what do you think.

tspiteri:
My point is that with code actions, case matching doesn't help. For example my actions typically look something like this:
```
Add `#[derive]`
Add default constructor
Change visibility to pub(crate)
Implement Options
```
Are there any code actions that are case-sensitive?

(Of course, this is only for completing code actions, not for completing directories/workspaces/whatever where case might actually matter.)

yyoncho:
I will merge it as it is but I reserve the right to revert it anyone complains.

PR #1644, opened 2020-05-03
Update rust-analyzer settings

flodiebold:
`featureFlags` has been replaced by separate boolean settings, and there are a
bunch of new ones.


brotzeit:
Thanks!

yyoncho:
@flodiebold offtopic: since you are part of RA development it will be great if you can file issues for all of the stuff that RA has implemented but it is not supported in `lsp-mode`.

flodiebold:
@yyoncho Maybe I'm forgetting something, but I think it's mostly semantic highlighting.

PR #1647, opened 2020-05-04
Allow querying for highlights when point is on an active highlight.

### LINKED ISSUE #1239: Can't trigger nested highlighting
When a highlighted region has nested highlightable regions, moving the cursor to the smaller highlightable regions doesn't shrink the highlight. You can only get the small regions to trigger if you move outside the big highlight and then move directly to the small highlight.

highlight

It seems like we need to re-query for highlighting even if the cursor is currently in a highlit region.

leungbk:
This allows for highlightable elements nested within lower-priority
highlightable elements to be treated appropriately.

Related to #1239.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4507,10 +4507,21 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
     (setq lsp--have-document-highlights nil)
     (lsp-cancel-request-by-token :highlights)))
 
+(defvar-local lsp--symbol-bounds-of-last-highlight-invocation nil
+  "The bounds of the symbol from which `lsp--document-highlight'
+  was most recently invoked.")
+
 (defun lsp--document-highlight ()
   (unless (or (looking-at "[[:space:]\n]")
-              (lsp--point-on-highlight?)
-              (not lsp-enable-symbol-highlighting))
+              (not lsp-enable-symbol-highlighting)
+              (and (lsp--point-on-highlight?)
+                   (let ((curr-sym-bounds (bounds-of-thing-at-point 'symbol)))
+                     (prog1
+                         (and curr-sym-bounds
+                              (equal curr-sym-bounds
+                                     lsp--symbol-bounds-of-last-highlight-invocation))
+                       (setq lsp--symbol-bounds-of-last-highlight-invocation
+                             curr-sym-bounds)))))
     (lsp-request-async "textDocument/documentHighlight"
                        (lsp--text-document-position-params)
                        #'lsp--document-highlight-callback


yyoncho:
So this change is a bit controversial. We will handle the nested highlighting but at the same time if you move the cursor inside of highlighted area this will result in redundant calls to the server. Anyone from @emacs-lsp/lsp-mode willing to share their opinion?

alanz:
Based on the gif in the bug report, it is something that does need to be fixed.
A server should mostly be presenting cached results for high-frequency requests, so that should not be a major concern.

leungbk:
One thing I didn't consider was caching the previous response from the server, comparing that with the server's next response, and using that alongside a `buffer-modified-tick` to determine whether or not we need to re-apply the highlights. Would that be preferred?

ericdallo:
I think we can use a cache as we can't trust that all servers have a cache implementation on their side.

danielmartin:
Another approach may be to cache the result of `(bounds-of-thing-at-point 'word)` in the overlay, and only make a new request when the point is outside that range (ie. the user moved to another word). But needs to be careful if the buffer is modified in between.

yyoncho:
@leungbk you don't have to check buffer-modified-tick - the rest of the code handles that. When you are called back it will be the same buffer-modified-tick because the request is executed in mode 'tick. What you might try is to check before deleting/reapplying the highlights is whether the range of the highlight at point is the same as the range of the response. I believe that this might be sufficient.

leungbk:
> Another approach may be to cache the result of `(bounds-of-thing-at-point 'word)` in the overlay, and only make a new request when the point is outside that range (ie. the user moved to another word).

OK, I've pushed something like this. However...

> But needs to be careful if the buffer is modified in between.

I don't know how we can get away with caching if we need to account for things like asynchronous edits by the language server. It seems quite possible that a) an asynchronous edit pushes preceding text forward so that an already-highlighted region now ends up at point, and that b) if we were to ask the server for highlights with point at the pushed-forward text, the server would respond with a different set of highlights.

> @leungbk you don't have to check buffer-modified-tick - the rest of the code handles that. When you are called back it will be the same buffer-modified-tick because the request is executed in mode 'tick. What you might try is to check before deleting/reapplying the highlights is whether the range of the highlight at point is the same as the range of the response. I believe that this might be sufficient.

I don't know if this works in the following example:

https://imgur.com/bWSeRDC

When point is on `&foo` and I jump to the `foo` in the return type, the overlay at point already matches the portion at that region returned by the language server, but we must re-apply the highlights regardless.

yyoncho:
Can you rebase?

muirdm:
Thanks for working on this! Can you rebase so we can get this fix in?

leungbk:
> Thanks for working on this! Can you rebase so we can get this fix in?

I rebased, though this patch doesn't address one of Daniel's concerns:

> Another approach may be to cache the result of (bounds-of-thing-at-point 'word) in the overlay, and only make a new request when the point is outside that range (ie. the user moved to another word). **But needs to be careful if the buffer is modified in between.**

One sequence that might happen is

```
1) an asynchronous edit pushes preceding text forward so that an already-highlighted region now ends up at point
2) if we were to ask the server for highlights with point at the pushed-forward text, the server would respond with a different set of highlights.
```

Step 2 won't happen with this patch since we won't query the server again.

PR #1649, opened 2020-05-04
lsp-capf: allow customize showing completion detail

### LINKED ISSUE #1639: Rust autocomplete gives very long information
Here is an example. It shows the full path where the keywords are defined.
image
Any config to make the information more concise?

kiennq:
Fix #1639


yyoncho:
thank you!

PR #1654, opened 2020-05-04
lsp-capf: Support clangd score attribute extension

kiennq:
I think it would be nice to support `clangd` code completion scores [extension](https://clangd.llvm.org/extensions.html).
That multiplier will ensure we still be able to cache candidates and still get the good candidate order even with our own filtering.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4243,10 +4243,12 @@ When the heuristic fails to find the prefix start point, return DEFAULT value."
        (-map (-lambda ((item &as &hash
                              "label"
                              "filterText" filter-text
-                             "_emacsStartPoint" start-point))
+                             "_emacsStartPoint" start-point
+                             "score"))
                (propertize (or filter-text label)
                            'lsp-completion-item item
-                           'lsp-completion-start-point start-point))
+                           'lsp-completion-start-point start-point
+                           'lsp-completion-score score))
              it)
        (seq-into it 'list)
        (-group-by (-partial #'get-text-property 0 'lsp-completion-start-point) it)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4277,8 +4279,8 @@ Also, additional data to attached to each candidate can be passed via PLIST."
                                  it))))
                   (-flatten-n 1)
                   (-sort (-on #'> (lambda (o)
-                                    (or (get-text-property 0 'completion-score o)
-                                        0))))
+                                    (* (or (get-text-property 0 'completion-score o) 1.0)
+                                       (or (get-text-property 0 'lsp-completion-score o) 0)))))
```
yyoncho, line 4283: seems like the second 0 should be 1.0

kiennq: Ah, correct, I've changed my mind, setting to a small number as penalty in case the score is not existed instead

yyoncho: I am a bit concerned that these calculations are on the hot path(the completion) and they will be calculated N*LogN during the sorting. Probably, a bit more efficient will be to calculate the scores one time and use them during the sorting. Some language servers like dart do not filter and have 1000+(?) candidates. Another option is to reorganize the code so we use the old comparison function when the first candidate does not have 'score. WDYT?

lsp-mode.el
```diff
@@ -4277,8 +4279,13 @@ Also, additional data to attached to each candidate can be passed via PLIST."
                                  it))))
                   (-flatten-n 1)
                   (-sort (-on #'> (lambda (o)
-                                    (or (get-text-property 0 'completion-score o)
-                                        0))))
+                                    (or (get-text-property 0 'sort-score o)
```
yyoncho, line 4282: are we sure that this won't be persisted multiple completions when using cache?

lsp-mode.el
kiennq: I see. While I'm expecting calculating the multiply can be quite fast since it's implemented in C code, calculating score just one seems to be better optimization.
I've fixed it by caching the calculated value.

lsp-mode.el
kiennq: Yes, it won't be persisted. The list is created a new every time by previous block. The only thing persisted is `lsp-completion-item`.

kiennq: Wait, let me check again.

yyoncho:
Thank you! We should mention that in the README when clangd section is in-place.

lsp-mode.el
yyoncho: The multiplication is fast but the function call elisp->c is slow. So you have 2 `or`'s one `get` and one `*` and this one will be repeated N*LogN while the user is typing.

lsp-mode.el
kiennq: Nvm, the list is indeed created a new every time by `completion-all-completions`.

PR #1655, opened 2020-05-05
Add auto doc generation

ericdallo:
- Auto docs generation for all supported LSPs when deploying via `gh-pages`.
- Allow custom file edits like the existing `lsp-rust.md`.
- Move changelog to root.
- Remove old docs.

In a next PR, I'll try to add public interfaces to the docs

Screenshot from 2020-05-04 23-57-51


mkdocs.yml
```diff
@@ -4,14 +4,69 @@ site_url: https://emacs-lsp.github.io/lsp-mode
 nav:
   - Home: README.md
   - LSPs:
+      - Ada: lsp-ada.md
       - Apple Sourcekit: https://emacs-lsp.github.io/lsp-sourcekit
+      - Angular: lsp-angular.md
+      - Bash: lsp-bash.md
+      - C++: lsp-ccls.md
```
yyoncho, line 11: I would add the name of the server in the index for the languages that have more than one server.

mkdocs.yml
```diff
@@ -4,14 +4,69 @@ site_url: https://emacs-lsp.github.io/lsp-mode
...
+      - C++: lsp-clangd.md
+      - C++: lsp-cquery.md
```
yyoncho, line 13: Can you indicate that cquery is no longer supported?

ericdallo: maybe we can remove it for the navigation?

yyoncho: works for me. The emacs package is broken for a long time.

PR #1662, opened 2020-05-06
lsp-capf: properly check for trigger-chars condition

kiennq:
Up until now, we've detected if the trigger characters are there and continue to trigger the completion by:
- go to bound start (decided by `symbol` boundary)
- check if the previous char is trigger char.

This doesn't work well when the `symbol` boundary is also including the trigger character already.
This change is to mitigate that problem by:
- Only detect if previous character is trigger char, and will request completion list.
- If there's cache, that means we're still trying to complete a same symbol, thus continue to request for completions.
- Recalculate the boundary start point to include trigger character.

The first two changes ensure that once we detect the trigger character and still complete the same symbol, `company-mode` will continue to request for completions. Archive the same effect (minus bug) of the original approach.
The last change ensure that `company-mode` will not suddenly reset the cache due to boundaries are changed. Since the trigger character will trigger completion, it should be included in current symbol to prevent `company-mode` detecting different boundary start and then treat new completion as for different symbol with the one triggered by trigger character.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4192,7 +4192,7 @@ and the position respectively."
             (when-let (kind-name (and kind (aref lsp--completion-item-kind kind)))
               (format " (%s)" kind-name)))))
 
-(defun lsp--looking-back-trigger-characters-p (trigger-characters)
+(defun lsp--looking-back-trigger-characterp (trigger-characters)
   "Return trigger character if text before point matches any of the TRIGGER-CHARACTERS."
   (unless (= (point) (point-at-bol))
     (seq-some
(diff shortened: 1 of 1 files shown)


yyoncho:
Which language server to test this with?

kiennq:
> Which language server to test this with?

`texlab`, it registered `\` as trigger characters.
For example, turn on logging and type `\doc`, it should show the logging for the `\` as trigger character (`kind` property of the request) and the next `d` as `TriggerForIncompleteCompletions`

yyoncho:
Thank you!

Issue #1660, opened 2020-05-06
lsp-mode 6.3: company-capf aborts active yasnippet while company-lsp doesn't


netjune:
Before logging the bug, please make sure that:

1. You are using the latest version of `lsp-mode` related packages.
2. You may also try reproduce the issue using clean environment using the following command:
3. Check FAQ and Troubleshooting section (https://github.com/emacs-lsp/lsp-mode#faq and https://github.com/emacs-lsp/lsp-mode#troubleshooting)
``` bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` could be downloaded from `https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el`.
Alternatively, in will be great if you can reproduce the issue using [lsp-docker](https://github.com/emacs-lsp/lsp-docker/) which provides the minimal configurations for `lsp-mode` and ships with most of the language servers.
3. Consider providing a minimal project to reproduce the issue with. Note that `lsp-mode` supports 30+ languages and the maintainers of the project are not familiar with all of the languages and a sample project + full steps to reproduce will help a lot.

**Describe the bug**

In an active field of yasnippet, a simple word completion with company-capf aborts yasnippet. While company-lsp works well for this case.

**To Reproduce**
Steps to reproduce the behavior(sample project + file which can be used to reproduce the issue with.)

1. expand an yasnippet with more than one fields
2. do a simple word completion with company-capf
3. yasnippet aborted with a message:
[yas] Committing snippets. Action would destroy a protection overlay.

**Expected behavior**

company-capf should works the same as company-lsp.

**Which Language Server did you use**
Mention which server/language did you use (e. g. lsp-python, lsp-ccls, lsp-java, etc)
lsp-go

**OS**
Which os do you use?
macOS 10.13.6

**Error callstack**
If there is an error please reproduce the issue with `toggle-debug-on-error` and after setting `lsp-print-io` to t, and then include the callstack and attach the content of `*lsp-log*` buffer and the content of the server log(use `lsp-workspace-show-log`)

If `emacs` hangs please do `M-x` `toggle-debug-on-quit` and then do `C-g` when emacs hangs and include the callstack as part of the issue report.

``` emacs-lisp
(setq lsp-print-io t)
```

kiennq:
It's dup of #1642 
You can work-around by setting `yas-inhibit-overlay-modification-protection` to `t`.
Or install the latest `lsp-mode` version

[...]

Issue #1642, opened 2020-05-02
Company-capf breaks snippet's filling


tpadioleau:
**Describe the bug**
Cannot go to next field of a snippet when it is filled with result from `company-capf`.

**To Reproduce**
`lsp-start-plain.el`
```elisp
;;; lsp-start-plain.el --- LSP mode quick starter      -*- lexical-binding: t; -*-

;; Copyright (C) 2018 Ivan Yonchovski

;; Author: Zhu Zihao <all_but_last@163.com>
;; Keywords: languages

;; This program is free software; you can redistribute it and/or modify
;; it under the terms of the GNU General Public License as published by
;; the Free Software Foundation, either version 3 of the License, or
;; (at your option) any later version.

;; This program is distributed in the hope that it will be useful,
;; but WITHOUT ANY WARRANTY; without even the implied warranty of
;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
;; GNU General Public License for more details.

;; You should have received a copy of the GNU General Public License
;; along with this program.  If not, see <https://www.gnu.org/licenses/>.

;;; Commentary:

;; This file is a helper to start a minimal lsp environment.
;; To use this, start your Emacs with "emacs -q" and load this file.

;; It will install `lsp-mode', `lsp-ui' with their dependencies to start a
;; minimal lsp environment.

;; And it forces Emacs to load `.el' files rather than `.elc' files
;; for more readable backtrace.

;;; Code:

(require 'package)

(setq debug-on-error t)

(let* ((package-archives '(("melpa" . "https://melpa.org/packages/")
                           ("gnu" . "https://elpa.gnu.org/packages/")))
       (no-byte-compile t)
       (package-user-dir (expand-file-name (make-temp-name "lsp-tmp-elpa")
                                           user-emacs-directory))
       (custom-file (expand-file-name "custom.el" package-user-dir))
       (pkg-list '(lsp-mode yasnippet company)))

  (package-initialize)
  (package-refresh-contents)

  (mapcar (lambda (pkg)
            (unless (package-installed-p pkg)
              (package-install pkg))
            (require pkg))
          pkg-list)

  (add-hook 'prog-mode-hook 'yas-minor-mode)
  (add-hook 'prog-mode-hook 'lsp)

  (setq company-backends '((company-capf)))

  (add-hook 'kill-emacs-hook `(lambda ()
                                (delete-directory ,package-user-dir t))))

[...]

PR #1665, opened 2020-05-07
lsp-capf: try to restore minimal text when completion is finalized

kiennq:
This should help fixing #1642, #1660 and another problem of filling snippet with multiple same clauses like what @tpadioleau reported.
We will try to minimalize the text needs to be change when restore the state of current buffer to in sync with server completion items.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4178,12 +4178,12 @@ and the position respectively."
                  "sortText" sort-text
                  "_emacsStartPoint" start-point)
           item)
-         ((&plist :prefix-line) plist))
+         ((&plist :markers) plist))
     (propertize (or label insert-text)
                 'lsp-completion-item item
                 'lsp-sort-text sort-text
                 'lsp-completion-start-point start-point
-                'lsp-completion-prefix-line prefix-line)))
+                'lsp-completion-markers markers)))
 
 (defun lsp--annotate (item)
   "Annotate ITEM detail."
@@ -4209,6 +4209,11 @@ When the completion is incomplete, cache contains value of `incomplete'.")
 
 (defun lsp--capf-clear-cache (&rest _)
   "Clear completion caches."
+  (-some-> (and (listp lsp--capf-cache) lsp--capf-cache)
+    (cddr)
+    (plist-get :markers)
+    (cadr)
+    (set-marker nil))
   (setq lsp--capf-cache nil))
 
 (defun lsp--capf-guess-prefix (item &optional default)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4440,16 +4440,21 @@ Also, additional data to attached to each candidate can be passed via PLIST."
...
-            (text-edit
-             (delete-region (point-at-bol) (point))
-             (insert prefix-line)
-             (lsp--apply-text-edit text-edit))
-            ((or insert-text label)
-             (delete-region (point-at-bol) (point))
-             (insert prefix-line)
-             (delete-region start-point (point))
-             (insert (or insert-text label))))
+           (let ((bol (point-at-bol)))
```
yyoncho, line 4443: will that work if I do: 

type `f` | invoke completion | type `oo` | go back one char | auto-complete `foo()`

kiennq: I think the behavior will not change compare to before.
Before: we delete the region from `bol` to `(point)` and insert `prefix-line` back to it.
Now: we compare if the string from `bol` to `(point)` already include `prefix-line` and only delete the different part.

Both of them should resulted in same text, only different is that the later is changing smaller region than the former

yyoncho: It does not. The second one will result - `fooo()` because the `o` will be omitted from `(delete-region (+ bol (length prefix-line)) (point))`

yyoncho: Apparently this does not work even now... we could fix it probably by keeping suffix-line(or using markers?) or replace-bufffer-contents with the full line.

kiennq: The first one is not? I think it will resulted in the same thing, unless we include the `suffix-line` to also restore behind the point

tpadioleau: For snippets I am used to it works well.
Should it be considered to remove `(setq-local yas-inhibit-overlay-modification-protection t)` ?

kiennq: Using marker seems nice, let's me try it.

kiennq: Switch to use markers instead of `prefix-line`.

lsp-mode.el
```diff
@@ -4429,44 +4429,48 @@ Also, additional data to attached to each candidate can be passed via PLIST."
...
+        (apply #'delete-region markers)
+        (delete-region start-point (point))
+        (insert (or insert-text label)))))
+
+    (when (eq insert-text-format 2)
+      (yas-expand-snippet
+       (lsp--to-yasnippet-snippet (buffer-substring start-point (point)))
+       start-point
+       (point)))
+    (when (and lsp-completion-enable-additional-text-edit additional-text-edits)
```
yyoncho, line 4463: I wonder whether we could move this before the yas-expand which will solve the issue with overlays protection?

kiennq: Read the code for overlays protection again, I think additional text edits will not affect to yas's overlays. Since according to the specs, the additional text edits cannot interferes with the main edit. Thus even modify them, I think the overlay protection will not be triggered.

kiennq:
> Looks good!
> 
> The markers should be set to nil once they are used via set-marker

Ah, I forgot to clean markers. They will be clean by garbage collector anyway, but clean them earlier should be better.

lsp-mode.el
```diff
@@ -4429,44 +4429,51 @@ Also, additional data to attached to each candidate can be passed via PLIST."
...
+    (when (eq insert-text-format 2)
+      (yas-expand-snippet
+       (lsp--to-yasnippet-snippet (buffer-substring start-point (point)))
+       start-point
+       (point)))
+
+    (when (and lsp-completion-enable-additional-text-edit additional-text-edits)
+      (lsp--apply-text-edits additional-text-edits)))
+
+  (lsp--capf-clear-cache)
```
yyoncho, line 4467: seems like clear-cache is a better place to clean up the markers, isn't it?

kiennq: It's a little bit tricky since `post-completion` is being run after `company-completion-finished-hook`, thus required the marker is not being clean until after `post-completion`.
I've modified company hook to only clean cache in `company-after-completion-hook`.

yyoncho:
Thank you!

Sorry for being picky about the markers but not leaking the markers was very important because each maker makes the editing slower because AFAIK they require recalculating their position.

kiennq:
> Sorry for being picky about the markers but not leaking the markers was very important because each maker makes the editing slower because AFAIK they require recalculating their position.

No worry, I complete agree with cleaning markers ASAP

PR #1666, opened 2020-05-07
Remove unused variable.

kurnevsky:
It's defined but never used.


yyoncho:
Yes, there are a bunch of those(could be discovered using `(emr-el-find-unused-definitions)`)

PR #1669, opened 2020-05-08
rust-analyzer: Don't update inlay hints on idle

flodiebold:
It's unnecessary, and in large files can slow the editor to a halt.


brotzeit:
I wasn't sure if we really need this when I added it. What do you think @yyoncho ?

yyoncho:
Looks good

PR #1671, opened 2020-05-09
Switched redundant initialized-fn to initialization-options. 

alexjgriffith:
Replacing the redundant `initialized-fn key` with `initialization-options`, for `make-lsp-client`, properly imports `build.hxml`, making libraries referenced in `build.hxml` available for auto completion etc. 

Fixes bug where Flymake would be unable to find classes, even though the project compiles correctly using `haxe build.hxml`.

This pull request also pulls out `build.hxml` into it's own custom `lsp-haxe-hxml`, since most projects use other names, such as `compile.hxml`, `all.hxml`, etc.

edit: 
Addressed (read ignored) two new notifications 
- `haxe/didChangeRequestQueue`
- `haxe/didRunHaxeMethod`

### DIFF
--- a/lsp-haxe.el
+++ b/lsp-haxe.el
@@ -64,7 +64,8 @@
 ;; The build spec for the project.
 (defcustom lsp-haxe-hxml "build.hxml"
   "The compile file for the haxe project."
-  :type 'file)
+  :type 'file
+  :package-version '(lsp-mode . "6.4"))
 
 ;; https://github.com/emacs-lsp/lsp-mode/blob/150a933694349df960dc8fd7a15e04f5727e6433/lsp-rust.el#L251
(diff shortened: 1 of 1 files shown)


lsp-haxe.el
```diff
@@ -61,6 +61,11 @@
   :risky t
   :type '(repeat string))
 
+;; The build spec for the project.
+(defcustom lsp-haxe-hxml "build.hxml"
+  "The compile file for the haxe project."
+  :type 'file)
```
yyoncho, line 67: nit: can you add `:package-version '(lsp-mode . "6.4")`

alexjgriffith: Thanks for the feedback! 

I've included the `package-version` and addressed the new notifications that the haxe language server was sending.

yyoncho:

Issue #1409, opened 2020-02-12
Get the "wrong-type-argument stringp nil" error when I execute some commands


mahito1594:
Dear developpers,

I am very sorry for my poor English.  I get the `entered--Lisp error` when I execute some commands (e.g., `lsp-workspace-restart`).  I am not familiar with Emacs Lisp and Language Server at all, so I do not know whether this is due to me or not.  I am glad if you help me.

Thanks,

**Describe the bug**
I get `(wrong-type-argument stringp nil)` error when I execute some commands with lsp-mode.  I remark that sometimes I get no errors.

**To Reproduce**
Save the following as `my-mwe.el` (which is based on `lsp-start-plain.el`):

``` emacs-lisp
(require 'package)

(setq debug-on-error t)
(setq lsp-print-io t)

(let* ((package-archives '(("melpa" . "https://melpa.org/packages/")
                           ("gnu" . "https://elpa.gnu.org/packages/")))
       (no-byte-compile t)
       (package-user-dir (expand-file-name (make-temp-name "lsp-tmp-elpa")
                                           user-emacs-directory))
       (custom-file (expand-file-name "custom.el" package-user-dir))
       (pkg-list '(lsp-mode lsp-ui company-lsp yasnippet lsp-java lsp-python-ms lsp-haskell helm-lsp lsp-treemacs dap-mode lsp-origami
                            ;; for my MWE
                            ccls)))

  (package-initialize)
  (package-refresh-contents)

  ;; - avoid the error "bad-signature" if necessary -
  (setq package-check-signature nil)
  (package-install 'gnu-elpa-keyring-update)
  (gnu-elpa-keyring-update)
  (setq package-check-signature 'allow-unsigned)
  ;; - - - -

  (mapcar (lambda (pkg)
            (unless (package-installed-p pkg)
              (package-install pkg))
            (require pkg))
          pkg-list)

  (add-hook 'kill-emacs-hook `(lambda ()
                                (delete-directory ,package-user-dir t)))

  ;; my MWE
  (add-hook 'c-mode-hook
	    (lambda ()
	      (require 'ccls)
	      (lsp))))

(provide 'my-mwe)
```

After that, run

```
emacs -q -l /path/to/my-mwe.el
```

If you execute some commands (e.g., `M-x lsp-workspace-restart`) in any `*.c` file, then you get the error, for example

``` emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
[...]

PR #1676, opened 2020-05-11
Killed lsp--describe-buffer causes errors when type M-x

mahito1594:
Dear maintainers,

After we killed the buffer which is set to the variable `lsp--describe-buffer`, some errors occur when we type `M-x` (For detail of the error, Issue #1409 is helpful, I think).
Therefore, we should check whether `lsp--describe-buffer` is killed or not by using `buffer-live-p` (Note: `(buffer-live-p nil)` returns `nil`).

I am sorry for not being familiar with Emacs Lisp.  So, could you point out if something wrong.

Thanks,

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2250,7 +2250,8 @@ BINDINGS is a list of (key def cond)."
                     ,(format "maybe-%s" def)
                     ,def
                     :filter (lambda (item)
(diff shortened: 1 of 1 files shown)


yyoncho:

Sorry for the delay, I somehow missed it.

PR #1677, opened 2020-05-11
add ess-julia-mode to language-id list

seregaxvm:
Add `ess-julia-mode` to `lsp-language-id-configuration` list. This change is necessary to add support for `ess-julia-mode` in [lsp-julia](https://github.com/non-Jedi/lsp-julia).


yyoncho:
Thank you!

PR #1681, opened 2020-05-11
rust-analyzer: add chain to rendered inlay hints

cassandracomar:
Should make chaining hints render, like so.


cassandracomar:
testing this, will update if it works.

brotzeit:
Nice. Thanks!

PR #1685, opened 2020-05-13
Only show "truncated" if doc-lines > 0

bennofs:
When `lsp-signature-doc-lines` is zero, it does not make sense to show the
"Truncated" message in every case. With the new behaviour, setting
`lsp-signature-doc-lines` to zero becomes useful: only the signature is
shown by default, but `lsp-signature-toggle-full-docs` can be used to
view the full documentation.

If this is not the right way to do this, I'm happy about any feedback.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5018,7 +5018,9 @@ It will show up only if current point has signature help."
                             (length signatures)
                             (propertize "│ " 'face 'shadow)))
             (prefix-length (- (length prefix) 2))
-            (method-docs (when lsp-signature-render-documentation
+            (method-docs (when
+                             (and lsp-signature-render-documentation
+                                  (or (not (numberp lsp-signature-doc-lines)) (< 0 lsp-signature-doc-lines)))
                            (let ((docs (lsp--render-element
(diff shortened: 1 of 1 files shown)


yyoncho:
Isn't that equivalent to setting doc-lines to max int?

bennofs:
No, then you cannot make it only show the signature without any documentation.

yyoncho:
Hm, I still don't get it: AFAICS in both cases `lsp-signature-toggle-full-docs` can be used to
view the full documentation when. Can you describe how exactly to test the new option?

bennofs:
@yyoncho If `doc-lines` is set to max-int, then:

* in the signature view, all documentation lines are shown *at all times*
* running `toggle-full-docs` changes nothing at all (since all lines are already shown)

Whereas with `doc-lines` set to 0, the following happens with this PR:

* only the signature is shown by default
* running `toggle-docs-full` shows all of the documentation in addition to the signature

yyoncho:
Ok, now I got it - I have totally forgotten how I have implemented that feature.

yyoncho:

PR #1687, opened 2020-05-13
Fix typos

tarsius:


yyoncho:
Thank you (I really should run the spellchecker more regularly).

PR #1689, opened 2020-05-13
Call dap-autoconfigure-mode

ericdallo:
Enable dap-mode auto configuration when lsp has `lsp-auto-configure`

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3974,8 +3974,7 @@ in that particular folder."
       (add-hook 'lsp-on-idle-hook #'lsp--document-links nil t))
 
     (when (and lsp-enable-dap-auto-configure
-               (fboundp 'dap-mode)
-               (not (bound-and-true-p dap-mode)))
+               (featurep 'dap-mode))
       (dap-auto-configure-mode 1)))
 
   (let ((buffer (current-buffer)))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -65,6 +65,8 @@
 (declare-function evil-set-command-property "ext:evil-common")
 (declare-function projectile-project-root "ext:projectile")
 (declare-function yas-expand-snippet "ext:yasnippet")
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3474,7 +3482,12 @@ in that particular folder."
 
     (when (and lsp-enable-links
                (lsp-feature? "textDocument/documentLink"))
-      (add-hook 'lsp-on-idle-hook #'lsp--document-links nil t)))
+      (add-hook 'lsp-on-idle-hook #'lsp--document-links nil t))
+
+    (when (and lsp-enable-dap-auto-configure
+               (fboundp 'dap-mode)
+               (not (bound-and-true-p dap-mode)))
```
kiennq, line 3489: I think we can change this to use `featurep`. With current logic, even if user disable `dap-mode`, it will reenable itself again next time a buffer is opened. And user has no choice but set `lsp-enable-dap-auto-configure` to `nil`. Which can be tedious.
```suggestion
               (featurep 'dap-mode)
```

kiennq:
To fully enable `dap-mode` we also need language specific enabling, like `(require 'dap-go)`.
Do we have the same thing like `lsp-client-packages` for `dap-mode`? So, this auto-configure works for `lsp-mode`.

ericdallo:
> To fully enable `dap-mode` we also need language specific enabling, like `(require 'dap-go)`.
> Do we have the same thing like `lsp-client-packages` for `dap-mode`? So, this auto-configure works for `lsp-mode`.

I don't think we have something like this for `dap-mode` :pensive: 
Should we create it?
@yyoncho

yyoncho:
> Should we create it?

I think it makes sense.

ericdallo:
> To fully enable dap-mode we also need language specific enabling, like (require 'dap-go).
> Do we have the same thing like lsp-client-packages for dap-mode? So, this auto-configure works for lsp-mode.

We should do that in another PR.

PR #1692, opened 2020-05-14
Disable willSave and includeText when textDocumentSync is a number

harismandal:
## Specification
https://microsoft.github.io/language-server-protocol/specification#textDocument_didClose

## VSCode implementation
Default params when textDocumentSync is a number.
https://github.com/microsoft/vscode-languageserver-node/blob/03179b00aef101d4a46e705c048717e3f4560e08/client/src/client.ts#L3012-L3027

willSave not included when number kind.
https://github.com/microsoft/vscode-languageserver-node/blob/03179b00aef101d4a46e705c048717e3f4560e08/client/src/client.ts#L1131

includeText only if save.includeText.
https://github.com/microsoft/vscode-languageserver-node/blob/03179b00aef101d4a46e705c048717e3f4560e08/client/src/client.ts#L1211

There is also a small difference in didSave notifications which in vscode implementation are sent when textDocumentSync.save is a boolean and it's false, but I didn't change that.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3218,8 +3218,7 @@ in that particular folder."
 (defun lsp--send-will-save-p ()
   "Return whether will save notifications should be sent to the server."
   (let ((sync (gethash "textDocumentSync" (lsp--server-capabilities))))
-    (or (memq sync '(1 2))
-        (and (hash-table-p sync) (gethash "willSave" sync)))))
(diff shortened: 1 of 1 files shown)


yyoncho:

For which server it was causing problems?

harismandal:
It was causing some error messages with [jedi-language-server](https://github.com/pappasam/jedi-language-server).

PR #1694, opened 2020-05-15
cleanup: change to use encode-coding-string and equal

kiennq:
Related to emacs-lsp/dap-mode#279, we should use `encode-coding-string` instead of `string-as-unibyte`.
Also change from `string-equal` to `equal` for shortness.
I don't think we need to throw when compare thing is not string?

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1290,7 +1290,7 @@ On other systems, returns path without change."
          (type (url-type url))
          (encoding (or locale-coding-system 'utf-8))
          (file (decode-coding-string (url-filename url) encoding))
-         (file-name (if (and type (not (string= type "file")))
+         (file-name (if (and type (not (equal type "file")))
                         (if-let ((handler (lsp--get-uri-handler type)))
                             (funcall handler uri)
                           uri)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1290,7 +1290,7 @@ On other systems, returns path without change."
          (type (url-type url))
          (encoding (or locale-coding-system 'utf-8))
          (file (decode-coding-string (url-filename url) encoding))
-         (file-name (if (and type (not (string= type "file")))
```
yyoncho, line 1293: why is this? string= should be more performant in this case?

kiennq: When I try benchmark code with
``` elisp
(setq a '(("aaa" . "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
          ("aaa" . "aab")
          (#("aaa" 0 1 (s a)) . "aaa")
          (#("aaa" 0 1 (s a)) . "aab")))
(benchmark 1000000 '(--map (equal (car it) (cdr it)) a))
;; "Elapsed time: 1.544581s"
(benchmark 1000000 '(--map (string= (car it) (cdr it)) a))
;; "Elapsed time: 1.593952s"
```
It seems `equal` is a little bit faster?

Issue #1695, opened 2020-05-15
Using f-canonical is slow


kiennq:
When I first open a file that has `lsp-mode` enabled, it's slow.
![image](https://user-images.githubusercontent.com/2631472/82029780-6c26c480-96d2-11ea-9bf9-73053f9fcc7b.png)

It seems that `f-canonical` is using `file-truename` and it's slow (and also recursive).
Should we advice that function to use `lsp-canonical-file-name` instead?
Is there a special need of following symlinks for that?
I think we have several bug related to `f-canonical` before too.
#1300 #1523

kiennq:
@yyoncho

yyoncho:
@kiennq there is lsp-with-cached-filetrue-name but we should find a generic solution for the problem because a lot of function of `f` are using this method under the hood. We could try your solution as well - it will yield better performance but I think that we switched to using f-cannonical here and there to solve issues related to symbolic links.

kiennq:
From my profiling, they're coming from using `f-same?`, `f-ancestor-of?` which is calling `f-canonical` under the hood.
So, using the `filetrue-name` here is actually undesired. I don't think we want to follow symlink when calculating project root workspace, which can lead to weird situation.
For example, I have file `a.x` that is inside workspace `A` and sym-linked to workspace `B` to provide as a shared file. Now, whenever I open file `a.x` in workspace `B`, it calculates my workspace root to `A` and activate the wrong workspace.

yyoncho:
Ok, so the options are:

1. Reimplement those methods in lsp-mode
2. Make a macro from what you have implemented in your PR.
3. Use the macro about caching full name.

I like 1 the most then 2 and 3. 

(cc @rejeep )

rejeep:
Hey, I don't really do any Emacs development anymore (kids you know...), so I don't really have an opinion, but happy to accept any contributions to the repo.

kiennq:
I will implement 1, and explorer if this can be contributed back to `f.el`.

kiennq:
Done implement 1.

PR #1696, opened 2020-05-15
lsp: using lsp-canonical-file-name instead of f-canonical when init

kiennq:
This is an attempt of fixing #1695, we probably fix other places, but right now, just fix the `lsp` command running slow due to `f-canonical`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -967,6 +967,28 @@ FORMAT and ARGS i the same as for `message'."
       "Return the local name component of FILE."
       (or (file-remote-p file 'localname) file))))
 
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1482,10 +1502,6 @@ already have been created."
            (indent 1))
   `(when-let (lsp--cur-workspace ,workspace) ,@body))
 
-(defun lsp-canonical-file-name (file-name)
```
yyoncho, line 1485: it is public  - can you define alias?

kiennq: Ah, sorry, Defined alias

PR #1697, opened 2020-05-15
[pyls] catch up with v0.32.0 version of pyls

mpanarin:
Catching up with the new version of pyls.

in `lsp-register-custom-settings` I added 
```elisp
("pyls.plugins.rope_rename.enabled" (lambda () (eq lsp-pyls-rename-backend 'rope)) t)
("pyls.plugins.jedi_rename.enabled" (lambda () (eq lsp-pyls-rename-backend 'jedi)) t)
```


yyoncho:
Thank you!

PR #1698, opened 2020-05-15
Multiple servers: finish callbacks when error

### LINKED ISSUE #1659: Cannot rename symbol when eslint is configured
I've enabled eslint alongside typescript with by by setting `lsp-eslint-server-command` in my startup. When I do this and try to rename a symbol via `lsp-rename`, I get a timeout (maybe because it's trying to send the rename to both backends?).

Glad to try and fix this if it is indeed a bug. Pointers would be welcome.

jkdufair:
This is my naive attempt to gracefully complete callbacks when on of the servers returns an error.

Closes #1659

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2907,8 +2907,9 @@ ID is the request id. "
   (let (results errors)
     (lambda (result)
       (push (cons lsp--cur-workspace result)
-            (if result results errors))
-      (when (eq (+ (length errors) (length results)) (length workspaces))
+            (if (eq result :error) errors results))
+      (when (and (not (eq (length errors) (length workspaces)))
+                 (eq (+ (length errors) (length results)) (length workspaces)))
         (funcall callback
                  (if no-merge
                      results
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3001,6 +3002,7 @@ If NO-MERGE is non-nil, don't merge the results but return alist workspace->resu
                               nil
                               target-workspaces))
              (error-callback (lambda (error)
+                               (funcall callback nil)
```
yyoncho, line 3005: nil is valid return value so you cannot use it to indicate an error: use something like `:error`.

lsp-mode.el
```diff
@@ -2904,10 +2904,11 @@ MODE determines when the callback will be called depending on the
...
 If NO-MERGE is non-nil, don't merge the results but return alist workspace->result.
 ID is the request id. "
-  (let (results)
+  (let (results errors)
     (lambda (result)
-      (push (cons lsp--cur-workspace result) results)
-      (when (eq (length results) (length workspaces))
+      (push (cons lsp--cur-workspace result)
+            (if result results errors))
+      (when (eq (+ (length errors) (length results)) (length workspaces))
```
yyoncho, line 2911: here we should check if the server has returned only errors and in that case, we should not call the success callback

jkdufair:
@yyoncho I have applied your suggestions. Thank you for those.

yyoncho:

PR #1699, opened 2020-05-16
remove code coverage badge

brotzeit:
I'd like to remove it since I don't think it has any kind of value for us. Opinions ?


brotzeit:
Any objections ?

kurnevsky:
Using this badge you can easily check what's currently covered and what's not. Not sure, at least some information about test is better than none? :)

danielmartin:
As the coverage is still low, I agree with removing the badge and have the link somewhere else, away from the front page.

It's one of the first things you see when you visit the webpage, so people may incorrectly think the package quality is low.

brotzeit:
> It's one of the first things you see when you visit the webpage, so people may incorrectly think the package quality is low.

Yeah, that's also an argument that I had in mind.

brotzeit:
@kurnevsky I think @danielmartin is right and the current code coverage doesn't reflect the stability of the package. Maybe we can add it later again when we have a better test suite.

PR #1701, opened 2020-05-18
Rely on flycheck logic for flycheck integration

### LINKED ISSUE #1629: Flycheck not always executed due to lsp
If a user activates `lsp-flycheck-live-reporting` and uses a language server that does not support diagnostics, `flycheck` will not be executed.

The issue is in the following code (comment from me) :

```lisp
(defun lsp-flycheck-enable (&rest _)
  "Enable flycheck integration for the current buffer."
  (flycheck-mode 1)
  (when lsp-flycheck-live-reporting
    (setq-local flycheck-check-syntax-automatically nil)) ;; this prevent flycheck to run automatically
  (setq-local flycheck-checker 'lsp)
  (lsp-flycheck-add-mode major-mode)
  (add-to-list 'flycheck-checkers 'lsp)
  (add-hook 'lsp-after-diagnostics-hook #'lsp--flycheck-report nil t)) ;; if the server does not support diagnostics lsp--flycheck-report will never be called
```

Steps to reproduce:
  1. Create a bash script with:

```bash
#!/usr/bin/env bash

foo="test bar"
echo $foo
```
  2. ensure you have installed `bash-language-server` and that you have configured set `(flycheck-add-next-checker 'lsp 'sh-shellcheck)`
  3. open the script and start `lsp`
  4. modify the script (comment/uncomment lines) and see that you have to call `flycheck-buffer` explicitly to update lints

damienrg:
Instead of re-implementing the logic by hand and calling flycheck-buffer
directly, we rely on flycheck facility for the integration.

This provides the following benefits:
  - simpler user configuration, as users do not have to configure lsp
    options to control flycheck integration anymore
  - better integration with flycheck

Fix #1629.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7802,13 +7802,6 @@ This avoids overloading the server with many files when starting Emacs."
 (defvar flycheck-checker)
 (defvar flycheck-checkers)
 
-(defcustom lsp-flycheck-live-reporting t
-  "If non-nil, diagnostics in buffer will be reported as soon as possible.
-Typically, on every keystroke. If nil, diagnostics will be
-reported according to `flycheck-check-syntax-automatically'."
-  :type 'boolean
-  :group 'lsp-mode)
-
 (defun lsp--get-buffer-diagnostics ()
   (or (gethash (lsp--fix-path-casing buffer-file-name)
                (lsp-diagnostics))
@@ -7833,10 +7826,13 @@ reported according to `flycheck-check-syntax-automatically'."
         (lsp--flycheck-level level tags)
       level)))
 
+(defvar-local lsp--diagnostics-modified? nil)
+
 (defun lsp--flycheck-start (checker callback)
   "Start an LSP syntax check with CHECKER.
 
 CALLBACK is the status callback passed by Flycheck."
+  (setq lsp--diagnostics-modified? nil)
   (->> (lsp--get-buffer-diagnostics)
        (-map (-lambda (diag)
                (flycheck-error-new
@@ -7862,7 +7858,8 @@ CALLBACK is the status callback passed by Flycheck."
 
 (defun lsp--flycheck-buffer ()
   (remove-hook 'lsp-on-idle-hook #'lsp--flycheck-buffer t)
-  (flycheck-buffer))
+  (when lsp--diagnostics-modified?
+    (flycheck-buffer)))
 
 (defun lsp--buffer-visible? ()
   (or (get-buffer-window (current-buffer))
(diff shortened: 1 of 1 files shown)


damienrg:
I would like to add unit tests but I am not sure how to do that for now.

It seems that I should not add a dependency to flycheck and provide stub implementation.
Any guidance would be great to be sure I start in the right direction.

lsp-mode.el
```diff
@@ -7833,36 +7826,42 @@ reported according to `flycheck-check-syntax-automatically'."
...
-                :buffer (current-buffer)
-                :checker checker
-                :filename buffer-file-name
-                :line (1+ (lsp-diagnostic-line diag))
-                :column (1+ (lsp-diagnostic-column diag))
-                :message (lsp-diagnostic-message diag)
-                :level (lsp--flycheck-calculate-level diag)
-                :id (lsp-diagnostic-code diag)
-                :end-column (-> diag
+  (when lsp--diagnostics-modified?
```
yyoncho, line 7835: do we need to check that here? It seems to me that once we are at that point we should report unconditionally. What happens when I explicitly call M-x flycheck-buffer after modified?

lsp-mode.el
```diff
@@ -7833,36 +7826,42 @@ reported according to `flycheck-check-syntax-automatically'."
...
-                                (1+))
-                :end-line (-> diag
-                              lsp-diagnostic-range
-                              (plist-get :end)
-                              (plist-get :line)
-                              (1+)))))
-       (funcall callback 'finished)))
+                                (plist-get :line)
+                                (1+)))))))
+  (funcall callback 'finished)
```
yyoncho, line 7858: nit: keep it on the same line.

lsp-mode.el
damienrg: > do we need to check that here?

No, it will remove it. After submitting the PR, I thought about the case you mention and I do not think flycheck has a cache.

damienrg:
> you may take a look at lsp-integration-test.el but I am not 100% sure how to simulate on idle even in them(alternatively trigger on idle check programmatically?).

Can you give me a direction to implement the test:

1. Should I use `lsp-with-pyls`?
2. Should I try to call `lsp--on-diagnostics` directly?
3. Should I use a fake server?
4. something else

> And one more thing: please, drop a note when you start working on a particular bug. In the past, we had a lot of collisions so we are trying to establish this as a practice.

Noted. While we are here, is it ok to force push or do you prefer new commits?

yyoncho:
> 1. Should I use `lsp-with-pyls`?

It is up to you - you could create a suite for your language server.

> 2. Should I try to call `lsp--on-diagnostics` directly?

Depends on what you want to test/cover. IMO if we want to verify that this PR works we should simulate what the user is doing - e. g. 
1. Edit file - verify that the errors pop up (with big flycheck-on-idle)
2. Edit file - verify that the errors pop up (with small flycheck-on-idle)

> 3. Should I use a fake server?

It is up to you again, but I would recommend not doing that.

> Noted. While we are here, is it ok to force push or do you prefer new commits?

My personal preference is force push but it is not enforced before merging PRs.

lsp-mode.el
```diff
@@ -7871,19 +7868,13 @@ CALLBACK is the status callback passed by Flycheck."
...
-    (flycheck-buffer))
-   ;; not visible or inhibit hooks - schedule refresh
-   (t (add-hook 'lsp-on-idle-hook #'lsp--flycheck-buffer nil t))))
+from the language server."
+  (when (not lsp--diagnostics-modified?)
+    (setq lsp--diagnostics-modified? t)
+    (when (memq 'idle-change flycheck-check-syntax-automatically)
+      ;; make sure diagnostics are published even if the diagnostics
+      ;; have been received after idle-change has been triggered
+      (add-hook 'lsp-on-idle-hook #'lsp--flycheck-buffer nil t))))
```
yyoncho, line 7877: there is one more thing that we should do here, consider the following: User edits -> flycheck idle passes and it does not report the errors -> server is slow and takes 5 seconds to calculate and send diagnostics -> lsp-mode receives the messages and it will show them on in lsp-on-idle -> but the user does not perform any action, waiting for the errors to show up.

If the user does not do anything, the error won't pop up - I think you should call `(lsp--idle-reschedule (current-buffer))` here.

yyoncho:

I cherrypicked the commit on master and applied the comments from the review(and few other things).

PR #1707, opened 2020-05-20
Use the spinner.el spinner for progress

alanz:
This means that other activities requiring user input can happen at
the same time.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1560,7 +1560,13 @@ WORKSPACE is the workspace that contains the progress token."
                (cur (gethash "percentage" value nil))
                (reporter
                 (if lsp-progress-via-spinner
-                    (spinner-start 'progress-bar)
+                    ;; The progress relates to the server as a whole,
+                    ;; display it on all buffers.
+                    (mapcar (lambda (buffer)
+                              (with-current-buffer buffer
+                                (spinner-start 'progress-bar))
+                              buffer)
+                            (lsp--workspace-buffers workspace))
                   (if cur
                       (make-progress-reporter message 0 100 cur)
                     ;; No percentage, just progress
@@ -1575,7 +1581,11 @@ WORKSPACE is the workspace that contains the progress token."
        ("end"
         (when-let ((reporter (lsp-workspace-get-work-done-token token workspace)))
           (if lsp-progress-via-spinner
-              (spinner-stop)
+              (mapc (lambda (buffer)
+                      (when (buffer-live-p buffer)
+                        (with-current-buffer buffer
+                          (spinner-stop))))
+                    reporter)
             (progress-reporter-done reporter))
             (lsp-workspace-rem-work-done-token token workspace))))))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -221,6 +221,11 @@ the buffer when it becomes large."
   :group 'lsp-mode
   :type '(repeat symbol))
 
+(defcustom lsp-progress-via-spinner t
+  "If non-nil, display LSP $/progress reports via a spinner in the modeline."
+  :group 'lsp
+  :type 'boolean)
+
 (defvar-local lsp--cur-workspace nil)
 
 (defvar-local lsp--cur-version 0)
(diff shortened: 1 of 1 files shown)


alanz:
And it doesn't always stop at the end.  Any pointers?

yyoncho:
> And it doesn't always stop at the end. Any pointers?

Maybe stop not called in the correct buffer?

alanz:
> Maybe stop not called in the correct buffer?

No, trying it with a single buffer, with the haskell language server, which does  a lot of work on startup.

But I think it needs some logic like the start/stop of the main spinner, when all the progress indicators are stopped. Because it also creates and destroys a lot of those.

alanz:
Ok, if I then do a change to the file, the spinner activates again, then disappears.

For completeness, what happens is the spinner stops moving, but stays in the modeline, for my original problem. And it is a different style of spinner to the lsp startup one.

So maybe this is ok.

yyoncho:
> No, trying it with a single buffer, with the haskell language server, which does a lot of work on startup.

This does not guarantee you are calling the spinner in the same buffer. It might be called in the process buffer for example.

tspiteri:
I tried this and can easily reproduce the spinner not stopping. What I did then is call spinner-start in all buffers in the workspace, replacing `(spinner-start 'progress-bar)` with
```elisp
(mapcar (lambda (buffer)
	  (with-current-buffer buffer
	    (spinner-start 'progress-bar)))
	(lsp--workspace-buffers workspace))
```
The return value of spinner-start is a function to call to stop the spinner. To stop all the spinners, I replaced `(spinner-stop)` with
```elisp
(mapc (lambda (stopper)
	(funcall stopper))
      reporter)
```

Note that if it is undesirable to run the spinner in all workspace buffers, you can leave the spinner-start part unchanged, and simply replace `(spinner-stop)` with `(funcall reporter)` (I haven't tested this, but it is really a simplification).

yyoncho:
Do we need then to start spinner for all buffers in the workspace? As a side note, in an alternative implementation, we may introduce a slot in lsp--workspace and then show it global string just like we do for lsp--workspace-status-string.

tspiteri:
We don't need to, I only thought that it would help as in my case the progress relates to the server in all files, so better would be to only spin in the buffers for the server, not for the whole workspace, but I don't know how to do that.

In any case, we could leave the spinner-start unchanged, and simply replace `(spinner-stop)` with `(funcall reporter)`. That way the spinner is stopped no matter where we are when the end notification arrives. I tested this and the only issue is that the spinner is stopped, but the mode line itself is not refreshed, so the space is still there in the mode line until I go to that buffer once more. Then it is automatically refreshed.

tspiteri:
Another way to do it that only spins in a single buffer and stops more cleanly:

1. Save the buffer as `reporter`, so replace `(spinner-start ...)` with `(progn (spinner-start ...) (current-buffer))`.
2. Then, replace `(spinner-stop)` with `(with-current-buffer reporter (spinner-stop))`.

This cleans up better as spinner-stop calls `(force-mode-line-update)`.

alanz:
FYI this was a very quick attempt at solving this, if anyone wants to put a more comprehensive solution into place be my guest.  My knowledge here is a bit tangential.

yyoncho:
> Another way to do it that only spins in a single buffer

That would make sense only if the task is for the current buffer, right?

tspiteri:
>That would make sense only if the task is for the current buffer, right?

I don't really know: I only know about my use case where I have a server that affects all buffers in my workspace for its one progress notification, so I don't have any idea if there are other kinds of status notifications that affect only a single buffer, or if all notifications always affect all buffers in the workspace, or something else.

tspiteri:
What I've been using today, and seems to work, is the following modification on the pull request:
```elisp
;; replace
                    (spinner-start 'progress-bar)
;; with
                    (mapcar (lambda (buffer)
                              (cons buffer
                                    (with-current-buffer buffer
                                      (spinner-start 'progress-bar))))
                            (lsp--workspace-buffers workspace))
;; and replace              
              (spinner-stop)
;; with              
              (mapc (lambda (buffer-stopper)
                      (funcall (cdr buffer-stopper))
		      (when (buffer-live-p (car buffer-stopper))
			(with-current-buffer (car buffer-stopper)
                          (force-mode-line-update))))
                    reporter)
```
This puts the spinner in all buffers for the workspace. If I have two projects open at the same time, only the workspace in question is affected, as I would expect.

Starting the spinners will save a dotted pair for each workspace buffer: the car is the buffer itself and the cdr is the return of spinner-start.

When the end notification is received, for each dotted pair, the spinner is stopped by calling the cdr, and the mode line is updated using the car. I did this instead of simply calling spinner-stop on the buffer so that if another spinner has been started in the buffer in the meantime, we do not stop it. Also, before updating the mode line there is a check so that if a buffer is killed between the start and end notifications, we do not get an error.

yyoncho:
@alanz - willing to apply @tspiteri suggestion?

alanz:
@yyoncho yes, will update. Thanks.

yyoncho:
Thank you!

PR #1709, opened 2020-05-20
Fix mixed-up typescript server links

theosherry:
The typescript server links are pointing to the wrong servers.

# Reproduce
- goto https://emacs-lsp.github.io/lsp-mode/page/languages/
- click `JavaScript/TypeScript (sourcegraph)`
- verify the page is actually for the `theia-ide` server


yyoncho:
Thank you!

PR #1712, opened 2020-05-22
Filter out non-ranges where the beginning is greater than or equal...

leungbk:
...to the end of the folding range.

----

969817d obviates the need to call `lsp--line-character-to-point` for each range.
86a8f41 removes false ranges beginning at the same point as the end. I had a problem with these when working on https://github.com/emacs-lsp/lsp-origami/pull/2.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6336,15 +6336,15 @@ an alist
       (let ((sym (pop stack)))
         (push (lsp--get-line-and-col sym) line-col-list)
         (unless (seq-empty-p (gethash "children" sym))
-          (setf stack (append (lsp--imenu-filter-symbols (gethash "children" sym)) stack)))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1781,31 +1781,51 @@ WORKSPACE is the workspace that contains the diagnostics."
...
+                                   #'lsp--line-col-comparator
+                                   (apply
+                                    #'nconc
+                                    (seq-map
+                                     (lambda (range)
+                                       (-let [(&hash "startLine" start-line
+                                                     "startCharacter" start-character
+                                                     "endLine" end-line
+                                                     "endCharacter" end-character
+                                                     "kind" _kind)
```
yyoncho, line 1797: nit: if kind is not used just delete it.

yyoncho:
Thank you!

PR #1715, opened 2020-05-22
Remove Elixir (JakeBecker) support

wtfleming:
https://github.com/JakeBecker/elixir-ls is no longer maintained. The https://github.com/elixir-lsp/elixir-ls fork is being actively developed, and already supported by lsp-mode.


yyoncho:
Thank you!

PR #1716, opened 2020-05-23
[docs] Add mention to rustfmt on save

apiraino:
Hi,

I'm copypasting a nice suggestion received from @flodiebold.

Make sense to add it to the documentation until maybe a more official support for rustfmt is added to lsp-rust?

Thanks for reviewing!


brotzeit:
Thanks!

PR #1718, opened 2020-05-23
Change regexp for Angular language server

leungbk:
The second backslash is required. End-of-string pattern is preferred
here over end-of-line metacharacter. The ".*" isn't necessary.

----

#


yyoncho:
Thank you!

PR #1722, opened 2020-05-24
Set a tooltip for the lsp--on-progress spinners

tspiteri:
This enables users to see the notification title if the spinner is there for a long time and they hover the mouse over the spinner.


yyoncho:
Looks good. I would propose also exposing a command, e. g. lsp-show-current-tasks which can be used to show the same info but without using the mouse.

PR #1723, opened 2020-05-25
Speed up lsp--folding-range-build-trees

leungbk:
lsp--folding-range-insert-into-trees calls nconc on an accumulator
during each recursive invocation, so remove it and traverse with a
stack instead.

----

#


yyoncho:

PR #1724, opened 2020-05-25
Add fallback for flycheck error level when server does not report it

TOTBWF:
In the LSP spec, the diagnostic level on an error is optional. However,
we do not handle this case when translating to a flycheck error level, which
causes flycheck to not work.

See https://microsoft.github.io/language-server-protocol/specifications/specification-current/#diagnostic.

To solve this, I added a custom option to fall back on when the server does not provide a diagnostic level.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7867,6 +7867,13 @@ reported according to `flycheck-check-syntax-automatically'."
   :type 'boolean
   :group 'lsp-mode)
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

conao3:
Missing quote for :group argument.
please merge #1728! Now completely failed to load lsp-mode!
(This degrade can't detect on CI?)

PR #1725, opened 2020-05-25
lsp-rust-analyzer-update-inlay-hints: make overlay advances with text

kiennq:
Here is the issue:
45933196-276d-4f3c-81af-846d18434f04
You can see that when I try to insert `bar`, it's displayed after the inlay overlay `: i32`.

And here is after the fix
167f8d10-849a-4acc-8373-32ad5b4709d5
The text `bar` is display **before** inlay, and in correct place


brotzeit:
Thanks!

PR #1729, opened 2020-05-28
Update `lsp-language-id-configuration` for FORTRAN 77 and older

berquist:
I also removed an association that looked wrong. It's from https://github.com/emacs-lsp/lsp-mode/commit/3a7bb350f4e53e9d85399d49497a1bc0bb4f810b.


yyoncho:
Thank you!

PR #1730, opened 2020-05-28
Fix lsp-eslint custom option types

wyuenho:
The syntax on the choice types are wrong and so they don't display on emacs27. This PR fixes it.


yyoncho:
Thank you

PR #1732, opened 2020-05-29
Implement client side bindings for the protocol

yyoncho:
- I have extracted the scema from lsp4e, it covers all features from 3.15 except
the progress reporting stuff from 3.15

- Introduced destructoring function in the form of `(-let [&<LspStructureName>
:lsp-field...`. In order to make that more convenient the fields from the
structure are converter from camel-case to snake-case and also the optional
fields ends with ?. So failureReason will become failure-reason?

- Generated fields accessors, e. g. `(gethash "character" point)` will
become `(lsp:point-character point)`.

- Generated constants for the enums, e. g. `lsp/message-type-warning`.

- Generated methods lsp:<lsp interface>? which will check if the passed object
is satisfying the interface, e. g.

``` elisp
(lsp:position? (ht ("character" 10)
                   ("line" 10)))
=> t
```
Some sample code:

Before:

``` elisp
(defun lsp--window-log-message-request (params)
  "Display a message request to the user and send the user's selection back to the server."
  (let* ((type (gethash "type" params))
         (message (lsp--propertize (gethash "message" params) type))
         (choices (seq-map (-partial 'gethash "title")
                           (gethash "actions" params))))
    (if choices
        (completing-read (concat message " ") (seq-into choices 'list) nil t)
      (lsp-log message))))
```

After:

``` elisp
(-defun lsp--window-log-message-request ((&ShowMessageRequestParams :message :type :actions?))
  "Display a message request to the user and send the user's selection back to the server."
  (let* ((message (lsp--propertize message type))
         (choices (seq-map #'lsp:message-action-item-title actions?)))
    (if choices
        (completing-read (concat message " ") (seq-into choices 'list) nil t)
      (lsp-log message))))
```


ericdallo:
That's really awesome!

yyoncho:
Forgot to mention that:

``` elisp
(-let [(&Position :missing-property) nil]
  )
```
will fail compile time with

> if: Unknown key: :missing-property. Available keys: (:character :line)

lsp-protocol.el
```diff
@@ -0,0 +1,419 @@
...
+                                   (object value)
+                                 (puthash ,(substring (symbol-name name) 1) value object))))
+                           params)))))
+       (apply #'append )
+       (list* 'progn)))
+
+

+
+(defconst lsp/completion-item-kind-text 0)
```
kiennq, line 88:
```suggestion
```
This is defined twice

lsp-protocol.el
```diff
@@ -0,0 +1,419 @@
...
+;; begin autogenerated code
+
+(defconst lsp/completion-item-kind-text 0)
```
kiennq, line 91: Can we use `--` instead of `/`, are you afraid of name collision?

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,99 @@
...
+             (-map (-lambda ((name type-data))
+                     (let* ((key (intern (substring (symbol-name name) 1)))
+                            (required (-map (lambda (field-name)
+                                              (intern (concat ":" field-name))) (plist-get type-data :required)))
+                            (params (->> (plist-get type-data :properties)
+                                         (-partition 2)
+                                         (-map #'cl-first)
+                                         (-filter (-not (-partial #'-contains? required))))))
+                       (list key required params))))
+             (apply #'nconc '(lsp-interface ))))))
```
kiennq, line 87: No need for space
```suggestion
             (apply #'nconc '(lsp-interface))))))
```

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,99 @@
...
+                        (string= type "object")))
+             (-map (-lambda ((name type-data))
+                     (let* ((key (intern (substring (symbol-name name) 1)))
+                            (required (-map (lambda (field-name)
+                                              (intern (concat ":" field-name))) (plist-get type-data :required)))
+                            (params (->> (plist-get type-data :properties)
+                                         (-partition 2)
+                                         (-map #'cl-first)
+                                         (-filter (-not (-partial #'-contains? required))))))
+                       (list key required params))))
```
kiennq, line 86: I think with this `list`, the return value will be in the form of `(lsp-interface (key required params) ....)`? Not flattened list like in `lsp-protocol`?

lsp-protocol.el
```diff
@@ -0,0 +1,419 @@
...
+(defmacro lsp-interface (&rest interfaces)
+  "Generates LSP bindings from (Inteface name, Required Fields, Optional Fields) triplet.
+
+Example usage with dash.
+
+(-let [(&ApplyWorkspaceEditResponse
+  :failure-reason?) (ht (\"failureReason\" \"...\"))]
+  failure-reason?) "
+  (->> interfaces
+       (-partition 3)
```
kiennq, line 40: Can we generate list triplet instead of flatten list? This will save a `-partition` call. And easier to see the generated code.

lsp-protocol.el
```diff
@@ -0,0 +1,419 @@
...
+                                                       (substring (symbol-name label) 1)))
+                                   (object)
+                                 (gethash ,(substring (symbol-name name) 1) object))
+                              `(defun ,(intern (format "lsp:set-%s-%s"
+                                                       (s-dashed-words (symbol-name interface))
+                                                       (substring (symbol-name label) 1)))
+                                   (object value)
+                                 (puthash ,(substring (symbol-name name) 1) value object))))
+                           params)))))
+       (apply #'append )
```
kiennq, line 83:
```suggestion
       (apply #'append)
```

lsp-protocol.el
```diff
@@ -0,0 +1,419 @@
...
+       (list* 'progn)))
```
kiennq, line 84:
```suggestion
       (cl-list* 'progn)))
```

lsp-protocol.el
yyoncho: Fixed, I have accidentally duplicated it.

lsp-protocol.el
yyoncho: It is either `-` or something else since it is public. I used `:` and `/` to make the code a bit more readable because you will have a visual clue.

scripts/lsp-generate-bindings.el
yyoncho: The next line will merge the lists

lsp-protocol.el
yyoncho: Updated, not sure if  the result is better.

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,80 @@
...
+             (-partition 3)
+             (apply #'nconc '(lsp-interface))))))
```
kiennq, line 68: I think instead of `-partition` and `apply`, this can simply be `(funcall #'nconc '(lsp-interface))`

lsp-protocol.el
kiennq: Thanks

lsp-protocol.el
```diff
@@ -0,0 +1,439 @@
...
+                                      (cons
+                                       (intern (concat ":" (s-dashed-words (symbol-name param-name)) "?"))
+                                       param-name))
+                                    optional)
+                              (-map (lambda (param-name)
+                                      (cons (intern (concat ":" (s-dashed-words (symbol-name param-name))))
+                                            param-name))
+                                    required))))
+                 (cl-list*
+                  `(defun ,(intern (format "dash-expand:&%s" interface)) (key source)
```
kiennq, line 52: I think we're also exploring the possible to switch to use `plist` instead of hash table for better perf.
Would it make sense to generate `cl-defgeneric` and `cl-defmethod` here?
Something like
```suggestion
                  `(cl-defgeneric ,(intern (format "dash-expand:&%s" interface)) (key source))
                  `(cl-defmethod ,(intern (format "dash-expand:&%s" interface)) (key (source hash-table))
```

yyoncho: > Would it make sense to generate `cl-defgeneric` and `cl-defmethod` here?
> Something like

No, because that code is executed compile time while the type info is available runtime. And it won't work either way because we should first replace all gethash/puthash with the generated getters and setters and then we will switch to plists by changing the generated code.

scripts/lsp-generate-bindings.el
yyoncho: Fixed it should be simply cl-list*

lsp-protocol.el
kiennq: > No, because that code is executed compile time while the type info is available runtime.

Can you elaborate more?
I think it will dispatch to correct method at the runtime depend on the type of source?



> And it won't work either way because we should first replace all gethash/puthash with the generated getters and setters and then we will switch to plists by changing the generated code.

I see. just think that we can just generate hash table and plist version now by `cl-defgeneric`. And require only another PR to convert everything to use generated getters/setters + set correct `json` parsing type.

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,80 @@
...
+;;; Code:
+
+
+
+
+
+(with-temp-buffer
+  (insert-file-contents-literally "generated.protocol.schema.json")
+
+  (let ((defs (plist-get (json-parse-string (buffer-string)
```
kiennq, line 34: nit: This will only work in Emacs with native json.

lsp-protocol.el
yyoncho: > Can you elaborate more?
I think it will dispatch to correct method at the runtime depend on the type of source?

This function is called when `dash` is expanding the `-let` and generating`let` statements during the macro expansion phase. The type of the `source` will always be `symbol`(or elisp statement). E. g.

```
(let ((foo (ht)))
  (-let [(&ClientInfo :name) foo]
    name))
```

will emit the following calls to expand function.

```
1 -> (dash-expand:&ClientInfo :name foo)
1 <- dash-expand:&ClientInfo: (gethash "name" foo)
```

kiennq: I see, thanks

scripts/lsp-generate-bindings.el
ericdallo: Maybe we should move the `lsp--read-json` macro from `lsp-mode` to here, or on the future move this wrapper functions like `lsp--json-serialize` to a helper/utils `.el`.

ericdallo: Oh, this is a script for bindings generator, I think we will use this automatically, right? not by end-users, so I think is not a big deal use `json-parse-string` here.

lsp-protocol.el
```diff
@@ -0,0 +1,439 @@
+;;; lsp-protocol.el --- Language Sever Protocol Bindings  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020  Ivan Yonchovski
+
+;; Author: Ivan Yonchovski <yyoncho@gmail.com>
+;; Keywords: abbrev, abbrev, abbrev, abbrev
```
ericdallo, line 6: are these repeated keywords right?

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,80 @@
...
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;;
```
ericdallo, line 23:
```suggestion
;; Generates lsp-mode bindings
```

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,80 @@
...
+
+;;; Code:
+
+
+
+
```
ericdallo, line 29:
```suggestion
```

scripts/lsp-generate-bindings.el
```diff
@@ -0,0 +1,80 @@
...
+             (cl-list* 'lsp-interface)))))
+
+;; (lsp-interface CallHierarchyCapabilities nil (:dynamicRegistration))
+
+;; (-let [(&CallHierarchyCapabilities :failure-reason?) (ht ("failureReason" "..."))]
+;;   failure-reason?)
+
+;; (-let [(&CallHierarchyCapabilities :dynamic-registration?) (ht ("dynamicRegistration" "..."))]
+;;   dynamic-registration?)
```
ericdallo, line 76: leftover this, or would be an example?

scripts/lsp-generate-bindings.el
yyoncho: Yep, I converted it to use emacs 26 json parser.

lsp-protocol.el
yyoncho: will be deleted.

yyoncho:
> Is this ready or is it a first PR?

I think it is ready to be used. One thing that I wonder is whether to add -defun to lsp-mode until dash PR is merged. Of course, if we find something to be missing we should fix it and generally, it is still open for feedback.

ericdallo:
This feature from this PR was included by [this commit](https://github.com/emacs-lsp/lsp-mode/commit/ce405d0fdce7ea5a6fb902f6c1383c272b1e61bb)

PR #1733, opened 2020-05-29
Move customizable options back to lsp-mode group

wyuenho:
There is no such custom group called `lsp` defined.


yyoncho:
Thank you!

Issue #977, opened 2019-08-20
Abysmal performance in narrowed buffer


Fuco1:
**Describe the bug**

Unusable performance in a narrowed buffer, Emacs freezes and becomes unresponsive for minutes+

**To Reproduce**

1. Get a 1000+ line php buffer
2. Narrow down to a function
3. Insert any text

**Expected behavior**

There should be no noticable difference in a narrowed vs unnarrowed buffer.

I suspect you might be using the "standard idiom"

```elisp
(save-excursion
  (save-restriction
    (widen)
    ;; your code goes here
    ))
```

in some super-tight loop.  This has killed me several times before in both smartparens and Elsa.  Both `save-restriction` and `widen` kill performance if used in a loop.

**Which Language Server did you use**

https://github.com/felixfbecker/php-language-server latest commit

**OS**

```
Linux thales 4.15.0-20-generic #21-Ubuntu SMP Tue Apr 24 06:16:15 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

sebastiansturm:
I may be seeing something related now (though I believe that didn't use to be the case until recently): in a C++ file of about 2000 lines, `evil-insert-newline-below` takes ridiculously long to execute (~0.5 s per invocation). Doesn't seem to matter where I am within the buffer. I manually instrumented some lsp methods containing the save-widen-idiom mentioned by @Fuco1, but found no smoking gun. @Fuco1: do you have any further hints on how to find the root case? Instrumenting all of `lsp` using `elp-instrument-package` didn't turn up anything particularly slow either

sebastiansturm:
also, whatever I'm seeing is not deterministic. Restarting Emacs with identical settings has fixed the performance issue for now (I'm sure it will come back though). I noticed, however, that the buffer now contains 955 overlays as opposed to ~6500 just before restart. So maybe my experimental semantic-highlighting handler leaks overlays, will have to add some instrumentation I guess

yyoncho:
@sebastiansturm does it happen in a narrowed buffer? You may try also `profile-start`/`profile-report`. As a side note,  `gc-cons-threshold` not being high enough may cause such problems and it wont show if you have introspected only `lsp-mode` package.

sebastiansturm:
[...]

PR #1735, opened 2020-05-30
Remove lsp--extract-line-from-buffer

leungbk:
When lsp-find-references searches a file for references, it causes
lsp--line-char-to-point to be called once for each reference in that
file, which can be time-consuming since each invocation of
lsp--line-char-to-point requires a run
of (save-excursion (save-restriction (widen) (goto-char (point-min)
[...]))). Instead, collect all the references within Emacs with O(1)
passes per file using lsp--xref-make-items, which
uses (save-excursion (save-restriction (widen) (goto-char (point-min)
[...]))) once per file.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4601,34 +4601,53 @@ Others: TRIGGER-CHARS"
                 (error)))
       item))
 
-(defun lsp--extract-line-from-buffer (pos)
-  "Return the line pointed to by POS (a Position object) in the current buffer."
-  (let* ((point (lsp--position-to-point pos))
-         (inhibit-field-text-motion t))
+(defun lsp--xref-make-items (filename ranges)
+  "Return a list of `xref-item's from a list of corresponding RANGES in FILENAME."
+  (let ((inhibit-field-text-motion t)
+        (ranges (-sort (lambda (range-a range-b)
+                         (-let* ((pos-start-a (gethash "start" range-a))
+                                 (line-num-a (gethash "line" pos-start-a))
+                                 (col-num-a (gethash "character" pos-start-a))
+                                 (pos-start-b (gethash "start" range-b))
+                                 (line-num-b (gethash "line" pos-start-b))
+                                 (col-num-b (gethash "character" pos-start-b)))
+                           (lsp--line-col-comparator (cons line-num-a col-num-a)
+                                                     (cons line-num-b col-num-b))))
+                       ranges))
+        (curr-line 0)
+        accum)
     (save-excursion
-      (goto-char point)
-      (buffer-substring-no-properties (line-beginning-position)
-                                      (line-end-position)))))
-
-(defun lsp--xref-make-item (filename range)
-  "Return a xref-item from a RANGE in FILENAME."
-  (let* ((pos-start (gethash "start" range))
-         (pos-end (gethash "end" range))
-         (line (lsp--extract-line-from-buffer pos-start))
-         (start (gethash "character" pos-start))
-         (end (gethash "character" pos-end))
-         (len (length line)))
-    (add-face-text-property (max (min start len) 0)
-                            (max (min end len) 0)
-                            'highlight t line)
-    ;; LINE is nil when FILENAME is not being current visited by any buffer.
-    (xref-make (or line filename)
-               (xref-make-file-location filename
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4601,34 +4601,53 @@ Others: TRIGGER-CHARS"
...
-                            'highlight t line)
-    ;; LINE is nil when FILENAME is not being current visited by any buffer.
-    (xref-make (or line filename)
-               (xref-make-file-location filename
-                                        (1+ (gethash "line" pos-start))
-                                        (gethash "character" pos-start)))))
+      (save-restriction
+        (widen)
+        (goto-char (point-min))
+        (dolist (range ranges)
```
yyoncho, line 4623: is there a reason for preferring `dolist` over `-map`?

leungbk: I prefer to use `-map` with projection functions that do not mutate anything outside the scope they define, and to use `dolist` for impure functions.

leungbk:
> Can we unify somehow the logic here with the logic in imenu?

I can't think of a great way to do it.

1. `(save-excursion (save-restriction (widen) [...]` is so commonly used that making it part of a macro does not help with expressiveness.
2. Beyond the fact that the data of interest are shaped differently in imenu and in xref, their respective `lsp-` functions modify `point` differently: `lsp--convert-col-to-points-batch` used by `imenu` can change `point` to a different column within a line, while the analog in this PR does not consider the column number when mutating `point`

yyoncho:
> > Can we unify somehow the logic here with the logic in imenu?
> 
> I can't think of a great way to do it.
> 
> 1. `(save-excursion (save-restriction (widen) [...]` is so commonly used that making it part of a macro does not help with expressiveness.

As part of org-mode integration, I have created the following macro:

``` elisp
(defmacro lsp-save-restriction-and-excursion (&rest form)
  (declare (indent 0) (debug t))
  `(if lsp--already-widened
       (save-excursion ,@form)
     (widen)
     (-let [lsp--already-widened t]
       (save-restriction (save-excursion ,@form)))))
```
which aims to address https://github.com/emacs-lsp/lsp-mode/issues/977. Not sure if it can help for that particular case. 

> 2. Beyond the fact that the data of interest are shaped differently in imenu and in xref, their respective `lsp-` functions modify `point` differently: `lsp--convert-col-to-points-batch` used by `imenu` can change `point` to a different column within a line, while the analog in this PR does not consider the column number when mutating `point`

Isn't it possible to have cur-line as a global variable and use it as a pivot in a macro section? In that section, all position->point conversion will be relative to cur-line?

yyoncho:
One more thing to consider: as part of lsp-mode->org-mode integration we will have to go and adjust all point<>position transformations.

leungbk:
> As part of org-mode integration, I have created the following macro:

I'm fine with using that macro here since you seem to have a use for it elsewhere.

> Isn't it possible to have cur-line as a global variable and use it as a pivot in a macro section? In that section, all position->point conversion will be relative to cur-line?

Are there any benefits to making `cur-line` global beyond shaving off several lines of local declarations? I'm not sure I understand what you mean by pivot.

> One more thing to consider: as part of lsp-mode->org-mode integration we will have to go and adjust all point<>position transformations so every time we introduce section like that this increases the complexity of the code.

Do you have a branch of your org-mode extensions I can look at?

yyoncho:
> Are there any benefits to making `cur-line` global beyond shaving off several lines of local declarations? I'm not sure I understand what you mean by pivot.

The idea is that you could generalize both imenu/xrefs and any sorted position->point conversion. E. g. something like that:

``` elisp
(lsp-begin-sorted-position->point-transformation
   ;; code that uses lsp--position->point-using-cur-line
   )
```

Here it is the org-mode code, but it is still in progress:

https://github.com/yyoncho/lsp-mode/tree/org-mode-lsp-mode

leungbk:
Since there may well end up being streaming support for finding references, it's probably not worth continuing with this PR.

yyoncho:
> Since there may well end up being streaming support for finding references, it's probably not worth continuing with this PR.

IMO streaming is not going to happen. I asked 2 times in the LSP repo if there is server planning to implement streaming and nobody answered...

PR #1738, opened 2020-05-31
Change Serenata priority

Sasanidas:
It seems like the php-language server haven't received any recent update https://github.com/felixfbecker/php-language-server/commits/master

 Even that it doesn't have some feature Serenta is more mature, and less unstable, so I propose to change the lsp priority, also link here the Serenata author comment about it:
serenata  

Regards


yyoncho:
Thank you, I will leave it for some time open before merging so others could share their opinion as well.

PR #1740, opened 2020-06-01
extract signature for rust-analyzer

scturtle:
Add specific `lsp-clients-extract-signature-on-hover` function for rust-analyzer.

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -27,6 +27,7 @@
 (require 'lsp-mode)
 (require 'ht)
 (require 'dash)
+(require 's)
 
 (defgroup lsp-rust nil
   "LSP support for Rust, using Rust Language Server or rust-analyzer."
@@ -716,6 +717,18 @@ The command should include `--message=format=json` or similar option."
   (interactive (list (or lsp-rust-analyzer--last-runnable
                          (lsp-rust-analyzer--select-runnable))))
   (lsp-rust-analyzer-run (or runnable lsp-rust-analyzer--last-runnable)))
+
+(cl-defmethod lsp-clients-extract-signature-on-hover (contents (_server-id (eql rust-analyzer)))
+  (-let* (((&hash "value") contents)
+          (groups (--partition-by (s-blank? it) (s-lines value)))
+          (sig_group (if (s-equals? "```rust" (car (-third-item groups)))
+                         (-third-item groups)
+                       (car groups)))
+          (sig (--> sig_group
+                    (--drop-while (s-equals? "```rust" it) it)
+                    (--take-while (not (s-equals? "```" it)) it)
+                    (s-join "" it))))
+    (lsp--render-element (concat "```rust\n" sig "\n```"))))
 
 (provide 'lsp-rust)
 ;;; lsp-rust.el ends here


brotzeit:
@flodiebold can you help me out ?

flodiebold:
I don't really know what this does or why it's needed :thinking:

scturtle:
The problem these lines of code want to solve is that hover in lsp-mode hardly show useful signature with rust-analyzer.

截屏2020-06-03 08 12 59

截屏2020-06-03 08 15 13

Yes I know where this struct/function come from, please just show me the definition.

The value in response from rust-analyzer is like:

```
"value": "```rust\nkecc::opt::simplify_cfg\n```\n\n```rust\npub struct SimplifyCfgConstProp\n```\n___\n\nSimplifies block exits by propagating constants.",
```

or:

```
"value": "kecc::opt::simplify_cfg\n___\n\n```rust\npub struct SimplifyCfgConstProp\n```\n\nSimplifies block exits by propagating constants.",
```

On its first line it often shows where this struct/function come from and the format is casual and different between versions of rust-analyzer.

Yes `lsp-eldoc-render-all` can be turned on but I just need one line of signature like other lsp clients. Document is quite long for stdlib function.

In lsp-mode, `lsp--render-on-hover-content` calls `lsp--handle-rendered-for-echo-area` and use generic function `lsp-clients-extract-signature-on-hover` to extract one line of signature. If there is no specific function like this [clangd one](https://github.com/emacs-lsp/lsp-mode/blob/64864737116454f8a73e395c77d92404cb74d281/lsp-clients.el#L444) the default function will be use and only show the first rendered line which causes the problem above.

This PR extract the first two markdown code block from the response value and prefer the second one. Yes his heuristic is wield and I hope anyone could come out a better solution.

flodiebold:
This seems very coupled to the exact layout of the hover response though, which could easily change. Maybe it would be possible to add a setting to RA that makes it only return the signature -- you're not using the full hover response in another place with this, right?

yyoncho:
FWIW if the server uses `MarkedString` this issue won't exist.

scturtle:
Well another problem: `lsp--hover-saved-bounds` often caches ridiculous large range wrong result and make hover not usable any more.

yyoncho:
> Well another problem: `lsp--hover-saved-bounds` often caches ridiculous large range wrong result and make hover not usable any more.

Can you elaborate?

scturtle:
> Can you elaborate?

If you query the hover request when rust-analyzer is processing some new changes, it will return ridiculous result like:
```
[Trace - 01:16:39 PM] Received response 'textDocument/hover - (874)' in 3ms.
Result: {
  "range": {
    "start": {
      "line": 86,
      "character": 66
    },
    "end": {
      "line": 297,
      "character": 5
    }
  },
  "contents": {
    "value": "```rust\nbool\n```",
    "kind": "markdown"
  }
}
```
And then a large range of source code is shadowed by saved bounds result.

danielmartin:
What's the status of this PR? It needs a rebase as there are some conflicts. I think it's a good idea to try to extract a Rust signature line for eldoc, but unfortunately we need to rely on some heuristics. Do you know if there's a better way to do this?

> FWIW if the server uses MarkedString this issue won't exist. I think it is now deprecated but IMO it is much more useful for clients like emacs which want to show minimal information.

Yes, that's a bad thing. I think LSP should separate signature info from documentation so that they could be shown separately.

yyoncho:
> Do you know if there's a better way to do this?

The alternative approach is to get N symbols from the output by default and let users on their own to customize that with a function(which you already provided).

scturtle:
> What's the status of this PR? It needs a rebase as there are some conflicts.

I agree with the maintainer that there need be a better way. This PR is what I can do so far.
I hope to leave this PR open for somebody finding it useful. That function can be put into one's `config.el` and not necessarily `lsp-rust.el`.
Along with disabling `lsp--hover-saved-bounds` (to work around the rust-analyzer bug mentioned above), it works quite good for me at the time.

jiacai2050:
@scturtle I copied your PR to my lsp-rust.el file, and restart Emacs, but nothing changed, is there anything I'm missing?

image

scturtle:
@jiacai2050 Try deleting `lsp-rust.elc` or recompile it with `emacs --batch --eval '(byte-recompile-directory "." 0)'`.

jiacai2050:
@scturtle No luck.
I also try build lsp-mode from source, apply your changes, and `cask build`, but hove message doesn't change at all.

jiacai2050:
image

After set `lsp-eldoc-render-all nil`, hover message display signature! But minibuffer is still too large, with many blank lines. Any ideas?

scturtle:
@jiacai2050 Try replacing `(lsp--render-element (concat "```rust\n" sig "\n```"))` with just `sig`?

jiacai2050:
@scturtle After pull latest commit(7ce0d78), eldoc minibuffer only display one line! Thanks for your help.

jiacai2050:
@brotzeit It's kinds of stale. perhaps we can close this PR, and document this. Let users decide how to extract siganature.
cc @scturtle

chaoky:
did something change? this stopped working for me :\

scturtle:
@chaoky https://github.com/scturtle/dotfiles/blob/f1e087e247876dbae20d56f944a1e96ad6f31e0b/doom_emacs/.doom.d/config.el#L74-L85

gagbo:
> @chaoky https://github.com/scturtle/dotfiles/blob/f1e087e247876dbae20d56f944a1e96ad6f31e0b/doom_emacs/.doom.d/config.el#L74-L85

Would you mind if we upstreamed this in Doom? That looks like a very very useful snippet!

scturtle:
> disabling `lsp--hover-saved-bounds` (to work around the rust-analyzer bug mentioned above)

I found a easy way to do this by adding this line in config.el:

```lisp
(advice-add #'lsp-hover :after (lambda () (setq lsp--hover-saved-bounds nil)))
```

It clears the cached bounds after every call to `lsp-hover`.

dgutov:
Now that this showed up on Reddit, and Rust-Analyzer is the official and preferred LS for Rust (since 2022), maybe it would be a good idea to revisit this PR?

scturtle:
Just in case anyone need it, here is a revised version that fix issue with `lsp-use-plists`:
```emacs lisp
;; do not cache the shitty result from rust-analyzer
(advice-add #'lsp-eldoc-function :after (lambda (&rest _) (setq lsp--hover-saved-bounds nil)))
;; extract and show short signature for rust-analyzer
(cl-defmethod lsp-clients-extract-signature-on-hover (contents (_server-id (eql rust-analyzer)))
  (let* ((value (if lsp-use-plists (plist-get contents :value) (gethash "value" contents)))
         (groups (--partition-by (s-blank? it) (s-lines (s-trim value))))
         (sig_group (if (s-equals? "```rust" (car (-third-item groups)))
                        (-third-item groups)
                      (car groups)))
         (sig (--> sig_group
                   (--drop-while (s-equals? "```rust" it) it)
                   (--take-while (not (s-equals? "```" it)) it)
                   (--map (s-trim it) it)
                   (s-join " " it))))
    (lsp--render-element (concat "```rust\n" sig "\n```"))))
```

pallix:
This last function should be part of the mode? We can see how popular it is with the emojis and I have seen many user being confused by getting only the alignment from the type instead of the type in their mode-line when using Rust and lsp.

PR #1744, opened 2020-06-02
Use ranged semanticToken requests when available

sebastiansturm:
this patch uses `semanticTokens/range` requests when supported by the server. Some open issues related to semantic highlighting still haven't been adressed though:

1. it might be advisable to use `font-lock-face` instead of `face` for semantic highlighting. It should certainly lead to a cleaner implementation, but might incur some runtime overhead that I haven't measured yet (see https://github.com/emacs-lsp/lsp-mode/pull/1721)

2. semantic token requests can take a long time to complete if the server is still initializing, so in those cases a progress bar should be shown. Not sure how to best do this; I could add a set of internal progress tokens to go alongside the `work-done-tokens` that are already part of the lsp workspace, and update the spinner any time either of these sets changes. Or maybe there's some better way that I missed?

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5895,14 +5895,14 @@ or `(point)' lies outside `lsp--semantic-highlighting-region'.")
     (setq lsp--semantic-tokens-use-ranged-requests is-range-provider)
     (setq font-lock-extend-region-functions new-extend-region-functions)
     (add-function :around (local 'font-lock-fontify-region-function) #'lsp--semantic-tokens-fontify)
-    (add-hook 'lsp-on-change-hook #'lsp--semantic-tokens-request-update)
+    (add-hook 'lsp-on-change-hook #'lsp--semantic-tokens-request-update nil t)
     (lsp--semantic-tokens-request-update)
     (setq lsp--semantic-tokens-teardown
           (lambda ()
             (setq font-lock-extend-region-functions old-extend-region-functions)
             (remove-function (local 'font-lock-fontify-region-function)
                              #'lsp--semantic-tokens-fontify)
-            (remove-hook 'lsp-on-change-hook #'lsp--semantic-tokens-request-update)))))
+            (remove-hook 'lsp-on-change-hook #'lsp--semantic-tokens-request-update t)))))
 
 (defun lsp--semantic-tokens-fontify (old-fontify-region beg end &optional loudly)
   ;; TODO: support multiple language servers per buffer?
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5880,6 +5880,8 @@ or `(point)' lies outside `lsp--semantic-highlighting-region'.")
 
 (defvar-local lsp--semantic-tokens-teardown nil)
 
+(defvar-local lsp--semantic-tokens-use-ranged-requests nil)
+
 (defun lsp--semantic-tokens-request-update ()
   (lsp--semantic-tokens-request
    (when lsp--semantic-tokens-use-ranged-requests (cons (window-start) (window-end))) t))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -750,6 +750,12 @@ directory")
     ("textDocument/references" :capability "referencesProvider")
     ("textDocument/selectionRange" :capability "selectionRangeProvider")
     ("textDocument/semanticTokens" :capability "semanticTokensProvider")
+    ("textDocument/semanticTokensRangeProvider"
+     :check-command (lambda (workspace)
+                      (with-lsp-workspace workspace
+                        (let ((table (lsp--capability "semanticTokensProvider")))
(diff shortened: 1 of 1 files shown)


yyoncho:
Hi @sebastiansturm . I am currently working on porting lsp-mode to use lsp-protocol.el. Please let me know what in your opinion will be the best way to coordinate the effort. Also, I think that this is a good time to drop the legacy highlighting and limit the work on porting that code too.

sebastiansturm:
I'm planning to try out `font-lock-face` today or tomorrow, after that I'd gladly start converting the highlighting code to `lsp-protocol`. If you're fine with removing support for the Theia protocol already, I'm all for it - but please be aware that this will break semantic highlighting for users of the two most recent clangd releases (9 and 10). Not sure if that kind of user even exists, but if so, that might IMO give them the wrong impression that `lsp-mode` is a brittle tool (on the other hand, semantic highlighting was explicitly marked as experimental, so maybe it's OK to do this after all; I'm not sure)

yyoncho:
> If you're fine with removing support for the Theia protocol already, I'm all for it - but please be aware that this will break semantic highlighting for users of the two most recent clangd releases (9 and 10). 

I think it is fine to delete it - it was experimental and never made it to the spec and it is not something critical. After all, lsp-mode.el is targeting language server protocol. 

We probably should start adding notes in changelog right away for every breaking change, not only before doing a release.

sebastiansturm:
well, that's nice, I'll get rid of it right away then. AFAIU, `lsp-protocol` is generated from 3.15 and hence doesn't know about semanticTokens yet, is that correct?

yyoncho:
> well, that's nice, I'll get rid of it right away then. AFAIU, `lsp-protocol` is generated from 3.15 and hence doesn't know about semanticTokens yet, is that correct?

Yes, I used lsp4j. Once https://github.com/microsoft/language-server-protocol/issues/67 is solved we could generate the bindings directly from the spec. Creating a few definitions manually is not that hard though.

sebastiansturm:
sure, should only be a few lines; I was just assuming it should all be autogenerated already. I'll have a look; as for the `font-lock-face`, I won't pursue it further right now because I just realized I'll still need to advise/modify font-lock in some way to update highlights while scrolling (unless I went back to the awful workarounds used by the old implementation, which I'd rather avoid, or always decorate the full buffer which isn't such a great alternative either).

sebastiansturm:
I removed Theia support from my fork of lsp-mode. Would it make sense to open a PR for that and then do a PR to your `plists` branch that includes that change, and converts semanticTokens to `lsp-protocol`? On an unrelated note, I noticed that with a recent build of `lsp-mode`, I'd get `invalid-function` errors related to `lsp-save-restriction-and-excursion`. I haven't looked at it in detail yet, though I guessed it might be because `lsp-save-restriction-and-excursion` is defined before being used (I'm using gccemacs, I assume that might make a difference)?

yyoncho:
> I removed Theia support from my fork of lsp-mode. Would it make sense to open a PR for that and then do a PR to your plists branch that includes that change, and converts semanticTokens to lsp-protocol? 

Yes. I will stop force pushing in plists. When you do the cleanup PR I will rebase plists on top of it. 

> On an unrelated note, I noticed that with a recent build of lsp-mode, I'd get invalid-function errors related to lsp-save-restriction-and-excursion. I haven't looked at it in detail yet, though I guessed it might be because lsp-save-restriction-and-excursion is defined before being used (I'm using gccemacs, I assume that might make a difference)?

I guess so. I guess it makes sense to fix that on our side but probably it should be handled by native compilation to be backward compatible.

(cc @AndreaCorallo)

AndreaCorallo:
Hi guys,

I guess `lsp-save-restriction-and-excursion` is a macro, in case the correct require should be added to the code.   In normal emacs probably works because before being compiled the macro definition is already loaded for some reason but async-compilation starting a process from scratch is a bit more picky for this kind of issue.  You should be able to reproduce it also with a vanilla Emacs just byte compiling the file without lsp-mode already loaded.

I say this because other packages had similar findings :) 

Please let me know if the issue is a different one

Ciao
  Andrea

sebastiansturm:
thanks for the quick response! It is indeed a macro, and I just noticed that Emacs (flycheck) warns about `lsp-save-restriction-and-excursion` being defined too late, same for `lsp-with-filename` and `lsp-with-current-buffer`. Don't see those warnings with `cask build`, probably because other source files `require` lsp-mode.el before lsp-mode.el itself is getting compiled. @yyoncho: does it make sense to explicitly list `lsp-mode.el` as the first item of cask's `files` list so errors like this will show up in the CI?

yyoncho:
> @yyoncho: does it make sense to explicitly list `lsp-mode.el` as the first item of cask's `files` list so errors like this will show up in the CI?

If doing so will uncover those issues - then definitely yes.

PR #1745, opened 2020-06-03
Initial org-mode integration

yyoncho:
- check docs/page/lsp-org.md
- fixed diagnostics reporting (there will be additional PR to fix `lsp-treemacs-errors-list`).

### DIFF
--- a/docs/page/lsp-org.md
+++ b/docs/page/lsp-org.md
@@ -1,0 +1,45 @@
+# Literate programming using LSP and org-mode(alpha)
+`lsp-mode` provides **experimental** support for running the language servers
+inside of [org-mode](https://orgmode.org/) source blocks. `lsp-mode` is doing
+that by obtaining the information about the source block and then translating
+the point to the LSP positions back and forth so the server is which so the
+server actually thinks that Emacs has opened the original file. In order that to
+work the files has to be present on the disk as well because the server expects
+to find them.
+
+
+``` org
+#+BEGIN_SRC python :tangle "python.py"
+print "Hello!"
+#+END_SRC
+```
+
+## Commands
+* `lsp-org` start `lsp-mode` in source buffer.
+* `lsp-virtual-buffer-disconnect` turn off `lsp-mode`.
+
+## What works
+* `lsp-mode` core features (finding references, going to definitions, completion, lenses, highlighting, etc)
+* `company-mode`
+* `flycheck`
+* `lsp-treemacs-symbols`
+* `lsp-treemacs-errors-list`
+
+## Known limitations
+* Incremental updates are performed by sending the whole
+  buffer each time(they simulate full updates).
+* Deleting content that spans across the source block and the content of the org
+  document is not handled.
+* Opening and editing the actual file from the disk might cause undefined
+  behavour because the server expects file to be open only once.
+
+## What does not work
+* `dap-mode`
+* `lsp-ui`
+* `flymake`(?)
+
+## What's next
+- Finishing the virtual buffers `API`.
+- `lsp-ui` support
+- `dap-mode` support
+- `markdown-mode` integration
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -235,6 +235,9 @@ modifiers into account. Only applies if
 (defvar-local lsp--cur-workspace nil)
 
 (defvar-local lsp--cur-version 0)
+(defvar-local lsp--virtual-buffer-connections nil)
+(defvar-local lsp--virtual-buffer nil)
+(defvar lsp--virtual-buffer-mappings (ht))
 
 (defvar lsp--uri-file-prefix (pcase system-type
                                (`windows-nt "file:///")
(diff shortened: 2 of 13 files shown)


lsp-mode.el
```diff
@@ -8356,8 +8391,234 @@ See https://github.com/emacs-lsp/lsp-mode."
...
+  (mapc #'lsp-virtual-buffer-disconnect lsp--virtual-buffer-connections))
+
+(defun lsp--move-point-in-indentation (point indentation)
+  (save-excursion
+    (goto-char point)
+    (if (<= point (+ (point-at-bol) indentation))
+        (point-at-bol)
+      point)))
+
+(defun lsp-org ()
```
kiennq, line 8462: Can we move this into another file? `lsp-org.el` for example.
I imagine that we would have another one for `markdown-mode` too

kiennq: Another thought, since this function can be applied even for something not `org-mode`? Should we rename this to something more generic and passing a parameter / object that provides `(org-element-context)`, `(cl-third (org-babel-get-src-block-info 'light))` ... as generic functions or properties?
Then creating `markdown-mode` integration can be as simple as define those interfaces/functions

yyoncho: Once we settle down and this is tested enough my plan is to split `lsp-org` into 2 parts: common and mode-specific. The mode-specific will be building a `virtual-buffer` and pass to the common one. Then we could move lsp-org in a separate file if it is more than a dozen lines.

The issue with this PR is that I have touched a lot of places and I am constantly having to merge and reimplement once the master branch changes. I want to have it in and then focus on converting the code to use lsp-protocol.el. Once we move to plists and lsp-protocol.el and receive enough feedback for org-mode integration we could start addressing your comments.

WDYT?

docs/page/lsp-org.md
```diff
@@ -0,0 +1,41 @@
+# Literate programming using LSP and org-mode(alpha)
+`lsp-mode` provides **experimental** support for running the language servers
+inside of [org-mode](https://orgmode.org/) source blocks. `lsp-mode` is doing
+that by obtaining the information about the source block and then translating
+the point to the LSP positions back and forth so the server is which so the
+server actually thinks that Emacs has opened the original file. In order that to
+work the files has to be present on the disk as well because the server expects
+to find them.
```
ericdallo, line 8:
```suggestion
`lsp-mode` provides **experimental** support for running the language servers inside of [org-mode](https://orgmode.org/) source blocks. `lsp-mode` is doing that by obtaining the information about the source block and then translating the point to the LSP positions back and forth so the server is which so the server actually thinks that Emacs has opened the original file. In order to that work the files have to be present on the disk as well because the server expects to find them.
```

docs/page/lsp-org.md
```diff
@@ -0,0 +1,41 @@
...
+* `lsp-treemacs-symbols`
+* `lsp-treemacs-errros`
+
+## Known limitations
+* Incremental updates are performed by sending the whole
+  buffer each time(they simulate full updates).
+* Deleting content that spans across the source block and the content of the org
+  document is not handled.
+* Opening and editing the actual file from the disk might cause undefined
+  behavour because the server expects file to be open only once.
```
ericdallo, line 30:
```suggestion
* Incremental updates are performed by sending the whole buffer each time(they simulate full updates).
* Deleting content that spans across the source block and the content of the org document is not handled.
* Opening and editing the actual file from the disk might cause undefined behavior because the server expects file to be open only once.
```

lsp-mode.el
kiennq: Sure

lsp-mode.el
```diff
@@ -235,7 +235,13 @@ modifiers into account. Only applies if
 (defvar-local lsp--cur-workspace nil)
 
 (defvar-local lsp--cur-version 0)
+(defvar-local lsp--virtual-buffer-connections nil)
+(defvar-local lsp--virtual-buffer nil)
+(defvar lsp--virtual-buffer-mappings (ht))
 
+(defvar-local lsp--virtual-buffer-connections nil)
+(defvar-local lsp--virtual-buffer nil)
+(defvar lsp--virtual-buffer-mappings (ht))
```
leungbk, line 244: These three lines are duplicated.

yyoncho: Thank you, fixed. Apparently I have pasted them twice.

lsp-mode.el
```diff
@@ -6759,23 +6764,21 @@ an alist
...
-                 (forward-line (- line curr-line))
-                 (setq curr-line line)
-                 (let ((line-end (line-end-position)))
-                   (if (or (not col) (> col (- line-end (point))))
-                       (goto-char line-end)
-                     (forward-char col)))
-                 (ht-set! line-col-to-point-map (cons line col) (if imenu-use-markers
-                                                                    (point-marker)
-                                                                  (point))))))
+  (let* ((line-col-to-point-map (ht-create))
```
leungbk, line 6767: Is `let*` actually necessary here?

test/fixtures/org-mode/org_demo_file.py
```diff
@@ -0,0 +1,8 @@
+import org_demo_file_2
+
+org_demo_file_2.exrnal_function("call2")
```
leungbk, line 3: Typo: `exrnal` -> `external`

test/fixtures/org-mode/demo.org
```diff
@@ -0,0 +1,62 @@
+#+BEGIN_SRC python :tangle "org_demo_file.py"
+import org_demo_file_2
+
+org_demo_file_2.exrnal_function("call2")
```
leungbk, line 4: `exrnal` -> `external`

docs/page/lsp-org.md
```diff
@@ -0,0 +1,45 @@
...
+## Commands
+* `lsp-org` start `lsp-mode` in source buffer.
+* `lsp-virtual-buffer-disconnect` turn off `lsp-mode`.
+
+## What works
+* `lsp-mode` core features (finding references, going to definitions, completion, lenses, highlighting, etc)
+* `company-mode`
+* `flycheck`
+* `lsp-treemacs-symbols`
+* `lsp-treemacs-errros`
```
leungbk, line 26: Typo

lsp-mode.el
yyoncho: Nope, reverted that change.

stardiviner:
Can `lsp-org` support lsp-mode in org mode opened source block buffer with `[C-c ']` ? I know there there are some workaround solution already. I still hope `lsp-org` can support this officially for better maintaince in future.

PR #1749, opened 2020-06-05
rust-analyzer fixes

flodiebold:
- update joinLines request, which changed the name and the schema a bit
 - remove applySourceChange commands, they're not used anymore; there's now an extension that allows rust-analyzer to return snippets from assists, which we don't support yet
 - add support for "go to parent module" command


brotzeit:
Thanks!

yyoncho:
Not directly part of this review but I propose adding a note in lsp-rust.el which is the targeted rust analyzer version and mention in the readme that if you are updating lsp-mode you should update rust analyzer too. This comment applies to every language server that has a great variety of extension methods.

PR #1758, opened 2020-06-07
Modify declare-function calls for org-mode functions

leungbk:
1. Specify the exact package in which each function is found
2. Every org package used here comes with org-mode, which comes with
every recent installation of Emacs anyway, so don't write "ext:"

In both Ivy and Helm, the developers avoid using `ext:` in all their `org`-related declare-function invocations.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8509,10 +8509,10 @@ See https://github.com/emacs-lsp/lsp-mode."
 
 ;; org-mode/virtual-buffer
 
-(declare-function org-babel-get-src-block-info "ext:org")
-(declare-function org-do-remove-indentation "ext:org")
-(declare-function org-src-get-lang-mode "ext:org")
-(declare-function org-element-context "ext:org")
+(declare-function org-babel-get-src-block-info "ob-core")
+(declare-function org-do-remove-indentation "org-macs")
+(declare-function org-src-get-lang-mode "org-src")
+(declare-function org-element-context "org-element")
 
 (defun lsp--virtual-buffer-update-position ()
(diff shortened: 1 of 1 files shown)


yyoncho:
Thank you!

kiennq:
For sake of consistent code base, I think we should keep `ext:`
```
Sometimes a file will optionally use functions from an external package. If you prefix the filename in the declare-function statement with ‘ext:’, then it will be checked if it is found, otherwise skipped without error.
```
For every external package, against current package, I think it would be easier and more consistent to put `ext:` in front for external functions

yyoncho:
I don't have strong opinion on this - as long as it is passing the byte-compilation tests.

leungbk:
I tried grepping all of my packages and saw that they all went without the `ext:` in the specific case of `org`. Aside from `ivy` and `helm-org`, packages like `avy`, `evil-collection`, `elisp-demos`, and a few others went without the `ext:`, whereas I could not find any examples of any packages using `ext:` for `org` functions.

I'm not sure which is right. On the one hand, every Emacs has an Org installation out of the box, but on the other hand, people tend not to use the pre-packaged version of Org that ships with Emacs. The only thing I'm aware of that might serve as a test is running `M-x check-declare-file` on `lsp-mode.el`, which passes for me whether or not I put `ext:` in front.

kiennq:
I think it would make more sense to just simple put `ext:` on every package that's not included/dependencies of `lsp-mode`.
Thus, even when the Emacs maintainer decides that `org-mode` should stay in ELPA instead of `:core` (so everyone can using the latest version), we still working instead of having `check-declare-file` broken.

PR #1760, opened 2020-06-07
Remove support for Theia semantic highlighting protocol

sebastiansturm:


lsp-mode.el
```diff
@@ -3103,13 +3080,12 @@ disappearing, unset all the variables related to it."
...
                       (formatting . ((dynamicRegistration . t)))
                       (rangeFormatting . ((dynamicRegistration . t)))
-                      ,@(pcase lsp-semantic-highlighting
-                          ((or :immediate :deferred) '((semanticHighlightingCapabilities . ((semanticHighlighting . t)))))
-                          (:semantic-tokens `((semanticTokens
-                                               . ((tokenModifiers . ,(if lsp-semantic-tokens-apply-modifiers
-                                                                         (apply 'vector (mapcar #'car lsp-semantic-token-modifier-faces)) []))
-                                                  (tokenTypes . ,(apply 'vector (mapcar #'car lsp-semantic-token-faces)))))))
-                          (_ '()))
+                      ,@(if lsp-enable-semantic-highlighting
```
yyoncho, line 3083: if -> when?

sebastiansturm: > This is a lot of code!

Hope you mean a lot of deleted code? That's largely because I had two similar, but separate implementations for the old protocol living side by side. I'll also move the face definitions upwards so I can get rid of the "reference to free variable" warnings caused by the references to `lsp-semantic-token-modifier-faces` and `lsp-semantic-token-faces` in lines 3086 and 3087. Hope that doesn't lead to awful diffs

sebastiansturm: > if -> when?

thanks! Changed it

yyoncho: > > This is a lot of code!
> 
> Hope you mean a lot of deleted code? That's largely because I had two similar, but separate implementations for the old protocol living side by side. I'll also move the face definitions upwards so I can get rid of the "reference to free variable" warnings caused by the references to `lsp-semantic-token-modifier-faces` and `lsp-semantic-token-faces` in lines 3086 and 3087. Hope that doesn't lead to awful diffs

Yes, a lot of deleted code (like something good).

sebastiansturm: alright, I don't need to worry then :) I'm glad we could get rid of that deadwood

yyoncho:
Thank you!

I will rebase plists branch on master tomorrow my time and we will continue as you suggested in the other PR.

PR #1761, opened 2020-06-07
Modify declare-function calls

leungbk:
Using nonnil fileonly value (fourth argument to declare-function) works
around functions that declare-function can't understand, like
constructors and accessor functions for cl-defstructs.

When arguments are optional, they must be noted in the arglist (third
argument to declare-function).

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7960,8 +7960,8 @@ Returns nil if the project should not be added to the current SESSION."
           (t nil)))
     ('quit)))
 
-(declare-function tramp-file-name-host "tramp" (file))
-(declare-function tramp-dissect-file-name "tramp" (file))
+(declare-function tramp-file-name-host "ext:tramp" (file) t)
+(declare-function tramp-dissect-file-name "ext:tramp" (file &optional nodefault))
 
 (defun lsp--files-same-host (f1 f2)
   "Predicate on whether or not two files are on the same host."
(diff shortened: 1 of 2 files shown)


leungbk:
I learned this only recently: the way to check this is to run `M-x check-declare-directory` or `check-declare-file`.

lsp-mode.el
```diff
@@ -7960,8 +7960,8 @@ Returns nil if the project should not be added to the current SESSION."
           (t nil)))
     ('quit)))
 
-(declare-function tramp-file-name-host "tramp" (file))
-(declare-function tramp-dissect-file-name "tramp" (file))
+(declare-function tramp-file-name-host "tramp" (file) t)
```
kiennq, line 7963: I think we should also prefix `tramp` with `ext:`.
A simple reason is that it works even when `tramp.el` is not available.
It's true that `tramp.el` is included in Emacs, but it's completely out of this package control, and just simple add `ext:` every where provides a more future proof way.

leungbk: OK, I prefixed the `tramp` and `org` stuff with `ext:`.

I think another reason to use `ext:` here is when we depend on a version of `tramp` or `org` that is newer than what ships with the bundled Emacs.

It's weird to me that no one else does this for `tramp` or `org` though. Even `tramp` eschews `ext:` when using `declare-function` for `org` functions.

PR #1762, opened 2020-06-07
Remove metals client.

kurnevsky:
It's available as a separate package now:
https://melpa.org/#/lsp-metals


lsp-mode.el
```diff
@@ -220,8 +220,8 @@ modifiers into account. Only applies if
 (defcustom lsp-client-packages
   '(ccls lsp-clients lsp-clojure lsp-csharp lsp-css lsp-dart lsp-elm
          lsp-erlang lsp-eslint lsp-fsharp lsp-gdscript lsp-go lsp-haskell lsp-haxe
-         lsp-intelephense lsp-java lsp-json lsp-metals lsp-perl lsp-pwsh lsp-pyls
-         lsp-python-ms lsp-rust lsp-serenata lsp-solargraph lsp-terraform lsp-verilog lsp-vetur
+         lsp-intelephense lsp-java lsp-json lsp-perl lsp-pwsh lsp-pyls lsp-python-ms
```
yyoncho, line 223: Seems like this should be reverted?

kurnevsky: You are right.

kurnevsky:
> Looks good to me! Can you add a note in the changelog? I think it will be good to mention all breaking changes right away.

Added to 6.4 version.

docs/page/gallery.md
```diff
@@ -17,10 +17,6 @@ Gallery
 
 ![](../examples/call-hierarchy-ccls.png)
 
-## Metals Doctor
-
-![](../examples/metals-doctor.png)
-
```
ericdallo, line 23: I'd keep this, like "Flutter debugging" above from `lsp-dart`

kurnevsky: I thought to add this to lsp-metals readme, but can keep here as well.

ericdallo: yes, both are ok :)

Issue #1683, opened 2020-05-12
[discussion] refactor to remove hash-tables and improve the code quality


yyoncho:
There are multiple problems with using hashmaps in lsp-mode:

1. HT literal is unreadable
2. they cannot be compared (easily)
3. They are a bit slower when deserialized
4. They are slower with small number of elements (our usecase)
5. Has to be checked for nil
6. ?

The issue is that we have 20k+ lines of code in emacs-lsp that relies on maps. As part of https://github.com/emacs-lsp/dap-mode/pull/274 suggested the following code for since @adam-r-kowalski was looking for some typechecking and I suggested the following code:

``` elisp
(defun dash-expand:&lsp-point (key source)
  (unless (member key '(:line :character))
    (error "Unknown key:  %s" key))
  `(gethash ,(substring (symbol-name key) 1) ,source))

;; works 
(-let [(&lsp-point :line) (ht ("line" 10)
                              ("character" 10))]
  line)

;; fails compile time
(-let [(&lsp-point :line2) (ht ("line" 10)
                               ("character" 10))]
  line2)
```

I was thinking about this idea right after dash introduced the expand feature but I was looking for making the code more beautiful(e. g. remove the strings) but now IMO if we go into that direction we could - autogenerate the expand functions from the spec see https://github.com/microsoft/language-server-protocol/issues/67

The benefits will be: 

1. Code will be a bit safer (and more beautiful?)
2. We will get rid of the "strings!@#"
3. Once we fix all places - we could switch to plists/alists and be happy.

ATM I cannot think of any drawbacks.

WDYT?

kiennq:
Totally agree with this

yyoncho:
> I think with current destructuring syntax, we've already be able to get rid of string, and be able to use either quote symbol `'a` or colon symbol `:a` instead?

Yes, we will be able to use :fieldName and be warned if the field name is not present. 

I will create a small prototype. We could create a few more things on top of the example above:
1. The optional fields will end with `?`
2. The camel case will be automatically replaced with snake-case.

yyoncho:
I have autogenerated the dash bindings: https://gist.github.com/yyoncho/030ed484bd1a215335b97b8df2058a18 

Example ussage:

```
(-let [(&ApplyWorkspaceEditResponse :failure-reason?) (ht ("failureReason" "..."))]
  failure-reason?)
```

We may also try to generate bindings for the actual methods as well.

yyoncho:
[...]

PR #1766, opened 2020-06-08
Provide easier way to edit hash tables defcustom + add one for go codelens

ibizaman:
There are currently two `defcustom`s that are used to edit hash-tables:
https://github.com/emacs-lsp/lsp-mode/blob/485d8d23d096cb98aa9b79249a3c96c25ad61e6a/lsp-yaml.el#L89
and
https://github.com/emacs-lsp/lsp-mode/blob/53bc160f0337877f1314fa6eb4835f2dbf52da4b/lsp-go.el#L58

They provide subpar editing feature because they require the user to use the hash-table syntax as can be seen with this screenshot:
image

This PR changes that to allow editing those as "normal" `alist`:
image
image
while maintaining the underlying variable as a hash-table:
image

This PR also adds a third hash-table `defcustom` that allows the user to enable/disable go codelens features. Indeed, `gopls` provides [5 codelens](https://github.com/golang/tools/blob/4d5ea46c79fe3bbb57dd00de9c167e93d94f4710/internal/lsp/source/options.go#L102-L108) but only [enables 3 by default](https://github.com/golang/tools/blob/4d5ea46c79fe3bbb57dd00de9c167e93d94f4710/internal/lsp/source/options.go#L119-L123) and there's no way to configure that currently in `lsp-mode`.

As I'm writing the above paragraph, looks like this PR does two things although the second commit depends on the first. Feel free to tell me to split it in two.

Also, I tried my best at documentation but even I can tell it's lacking although I can't find better phrasing.

Finally, I saw #1683 and I don't think it goes against that work.

### DIFF
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -55,12 +55,14 @@ completing function calls."
   :risky t
   :package-version '(lsp-mode "6.2"))
 
-(defcustom lsp-gopls-env (make-hash-table)
+(defcustom lsp-gopls-env '()
   "`gopls' has the unusual ability to set environment variables,
   intended to affect the behavior of commands invoked by `gopls'
   on the user's behalf. This variable takes a hash table of env
   var names to desired values."
-  :type '(restricted-sexp :match-alternatives (hash-table-p))
+  :type '(repeat (cons (string :tag "env var name") (string :tag "value")))
+  :set 'lsp--defcustom-set-hashtable-from-set
+  :get 'lsp--defcustom-get-set-from-hashtable
   :group 'lsp-gopls
   :risky t
   :package-version '(lsp-mode "6.2"))
--- a/lsp-go.el
+++ b/lsp-go.el
@@ -80,11 +80,32 @@ completing function calls."
   :risky t
   :package-version '(lsp-mode "6.2"))
 
+(defvar lsp-gopls-available-codelens
+  '(("generate" . "Run `go generate` for a directory")
+	("test" . "Run `go test` for a specific test function")
+	("tidy" . "Run `go mod tidy` for a module")
+	("upgrade_dependency" . "Upgrade a dependency")
+	("regenerate_cgo" . "Regenerate cgo definitions"))
+  "Available codelens that can be further enabled or disabled
+  through `lsp-gopls-codelens'.")
+
+(defcustom lsp-gopls-codelens '(("generate" . t) ("test" . t))
+  "Select what codelens should be enabled or not.
+
+The codelens can be found at https://github.com/golang/tools/blob/4d5ea46c79fe3bbb57dd00de9c167e93d94f4710/internal/lsp/source/options.go#L102-L108."
+  :type (lsp--defcustom-available-as-hash-table-type lsp-gopls-available-codelens)
+  :group 'lsp-gopls
+  :set 'lsp--defcustom-set-hashtable-from-set
+  :get 'lsp--defcustom-get-set-from-hashtable
+  :risky t
+  :package-version '(lsp-mode "6.4"))
+
 (lsp-register-custom-settings
  '(("gopls.usePlaceholders" lsp-gopls-use-placeholders t)
    ("gopls.hoverKind" lsp-gopls-hover-kind)
    ("gopls.buildFlags" lsp-gopls-build-flags)
-   ("gopls.env" lsp-gopls-env)))
+   ("gopls.env" lsp-gopls-env)
+   ("gopls.codelens" lsp-gopls-codelens)))
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8441,6 +8441,26 @@ See https://github.com/emacs-lsp/lsp-mode."
         (lsp--info "Disconnected from buffer %s" file-name))
     (lsp--error "Nothing to disconnect from?")))
 
+(defun lsp--defcustom-available-as-hash-table-type (alist)
+  "Returns a list suitable for the `:type' field in a `defcustom' used to populate a hash table.
+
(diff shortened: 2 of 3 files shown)


ericdallo:
Cool, we are migrating hash-tables -> `plists`, but I think using `alists` are good too.

c/c @yyoncho

lsp-go.el
```diff
@@ -55,12 +55,14 @@ completing function calls."
...
-(defcustom lsp-gopls-env (make-hash-table)
+(defcustom lsp-gopls-env '()
   "`gopls' has the unusual ability to set environment variables,
   intended to affect the behavior of commands invoked by `gopls'
   on the user's behalf. This variable takes a hash table of env
   var names to desired values."
-  :type '(restricted-sexp :match-alternatives (hash-table-p))
+  :type '(repeat (cons (string :tag "env var name") (string :tag "value")))
+  :set 'lsp--defcustom-set-hashtable-from-set
+  :get 'lsp--defcustom-get-set-from-hashtable
```
yyoncho, line 65: `:get` and `:set` should not be needed.

yyoncho: Also :get/:set are introduced in emacs 26 and we target emacs 25 as well.

yyoncho:
> Cool, we are migrating hash-tables -> `plists`, but I think using `alists` are good too.
> 
> c/c @yyoncho

We are using even now alists here and there when sending data(e. g. lsp--client-capabilities mostly because alist align better than plists when there is deeply nested structure). As I mentioned above, the format that we use for sending is not an issue. 

If your suggestion is about using alists for the parsed stuff - we might benchmark that. But my guess is that plists will be faster but I might be wrong.

yyoncho:
> Thank you for your PR

Forgo to mention that I am all for this change. hts are awkward in elisp.

ibizaman:
@yyoncho What's your preferred method to modify my PR. Rebasing or adding new commits? I'll add commits in the meantime but can always rebase later.

lsp-go.el
ibizaman: Indeed! I can simply use the `alist` as-is. I saw in the trace output:
```
Params: {
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    {
      "env": {
        "asdf": "asdf"
      },
      "buildFlags": [],
      "hoverKind": "SynopsisDocumentation",
      "usePlaceholders": true
    },
    null
  ]
}
```
When setting env var and value to `asdf` and `asdf` respectively.

yyoncho: Yep. Not related to the PR but what is the server request? Because one of the sections is null.

ibizaman: Ah actually I still need to transform `nil` (false) to `:json-false` otherwise I get:
```
[Trace - 10:28:48 PM] Received notification 'window/showMessage'.
Params: {
  "message": "Invalid type %!d(string=test) for map key \"codelens\" in option %!q(MISSING)",
  "type": 1
}
```

yyoncho: You should be setting this in the actual value. I mean, you should not set it to nil but to :json-false. We cannot safely assume that nil is false.

ibizaman:
Regarding https://github.com/emacs-lsp/lsp-mode/pull/1766#discussion_r437146055 I'll stick with alists as there are idiomatic ways to map on those.

ibizaman:
This latest commit works fine without using hash-tables at all:
```
[Trace - 11:00:24 PM] Sending response 'workspace/configuration - (2)'. Processing request took 0ms
Params: {
  "jsonrpc": "2.0",
  "id": 2,
  "result": [
    {
      "codelens": {
        "test": true,
        "generate": false
      },
      "buildFlags": [],
      "hoverKind": "SynopsisDocumentation",
      "usePlaceholders": true
    },
    null
  ]
}
```

lsp-go.el
```diff
@@ -78,11 +80,32 @@ completing function calls."
...
+  "Available codelens that can be further enabled or disabled
+  through `lsp-gopls-codelens'.")
+
+(defcustom lsp-gopls-codelens '(("generate" . t) ("test" . t))
+  "Select what codelens should be enabled or not.
+
+The codelens can be found at https://github.com/golang/tools/blob/4d5ea46c79fe3bbb57dd00de9c167e93d94f4710/internal/lsp/source/options.go#L102-L108."
+  :type (lsp--defcustom-available-as-alist-type lsp-gopls-available-codelens)
+  :group 'lsp-gopls
+  :set 'lsp--defcustom-set-json-alist
```
yyoncho, line 98: we should not use :get and :set since they are not supported in emacs 25. I propose removing them and ask users to enter :json-false when they need false.

ibizaman: Ah sorry @yyoncho yes I read that but forgot when I implemented.
And in fact, I believe I can just change https://github.com/emacs-lsp/lsp-mode/pull/1766/files#diff-b1e9bc82c3b6bd6e95470e18ee41a19bR8460 to set `:json-false` instead of `nil`.

ibizaman:
There, _much_ better! Last commit produces expected output:
```
[Trace - 11:47:54 PM] Sending response 'workspace/configuration - (2)'. Processing request took 56ms
Params: {
  "jsonrpc": "2.0",
  "id": 2,
  "result": [
    {
      "codelens": {
        "test": true,
        "generate": false
      },
      "env": {
        "asdf": "asdfasdf"
      },
      "buildFlags": [],
      "hoverKind": "SynopsisDocumentation",
      "usePlaceholders": true
    },
    null
  ]
}
```

ibizaman:
FWIW I tried using `plist` but it won't work without the get/set methods as what's sent to the server is a list, not a hash:
```
      "env": [
        "asdf",
        "asdfasdf",
        "two",
        "three"
      ],
```
And that results in a type error on the server side `LSP :: invalid config gopls.env type []interface {}`

PR #1767, opened 2020-06-08
Fix some bytecompiler warnings/gccemacs errors

sebastiansturm:
among other things, this change explicitly lists `lsp-protocol.el` and `lsp-mode.el` as the first members of Cask's `files` list (which leads to some byte compiler errors being shown that may get masked if files are compiled in a different order, such as late-macro-definition warnings). I left the glob `*.el` unchanged, so with this change Cask would compile these two files twice. I guess a simple way to fix this would be to have Cask automatically ignore duplicates caused by overlapping globs

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3839,6 +3839,23 @@ in that particular folder."
            (lsp--apply-text-edits text-edits)))
        changes))))
 
+(defmacro lsp-with-current-buffer (buffer-id &rest body)
+  (declare (indent 1) (debug t))
+  `(if-let (wcb (plist-get ,buffer-id :with-current-buffer))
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -977,6 +977,32 @@ They are added to `markdown-code-lang-modes'")
...
   :group 'lsp-faces)
 
+(defcustom lsp-signature-render-documentation t
+  "Display signature documentation in `eldoc'."
+  :type 'boolean
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "6.2"))
+
+(defcustom lsp-signature-auto-activate t
+  "Auto-activate the documentation when  "
```
ericdallo, line 987: `when  "` ?

yyoncho:
Thank you!

PR #1769, opened 2020-06-09
Fix renaming when using multiple language servers

0x450x6c:
Emacs stuck when using renaming feature with two language servers (intelephense + psalm).

It seems lsp-mode tries to apply rename for both language servers, but psalm doesn't support renaming feature.


lsp-mode.el
```diff
@@ -897,6 +897,13 @@ directory")
                                            (lsp--registered-capability-options)))))
                           (and (hash-table-p table)
                                (gethash "prepareProvider" table))))))
+    ("textDocument/rename"
```
yyoncho, line 900: Isn't that sufficient:

```
("textDocument/rename" :capability "renameProvider")
```
?

0x450x6c: Tested, yes it is sufficient.

PR #1770, opened 2020-06-09
Fix go to definition when using multiple language servers

0x450x6c:


lsp-mode.el
```diff
@@ -4940,16 +4940,16 @@ Others: TRIGGER-CHARS"
                                     filename (error-message-string err)))
                    (file-error (lsp-warn "Failed to process xref entry, file-error, '%s': %s"
                                          filename (error-message-string err)))))))
-    (apply #'append
-           (if (-some->> locations (cl-first) (gethash "uri"))
```
0x450x6c, line 4944: When using multiple language servers (intelephense + psalm), `locations` can contain mixed (`Location` and `LocationLink`) elements.

yyoncho:
This looks good. I will do manual testing and proceed with the merge. As a side note, atm we have functions to check if an object is an instance of particular class. I will do that change as part of `plists` refactoring.

PR #1771, opened 2020-06-09
Fix signature help when using multiple language servers

0x450x6c:


yyoncho:
This means that signature help will have the signature of the last server that has responded to the signature request, right? If yes, I don't think that this is the desired behaviour. 

I am not sure what should be the proper handling of it, too. One option to convert both and merge them. But it looks like what we should do is actually ask the user to disable signature help for one of the servers. E. g.  Have a warning if 2 servers match signature help and provide a recipe for how to disable signature help for one of them. WDYT?

0x450x6c:
Option to disable looks good, but I implemented first option, because it was easy to implement.

PR #1754, opened 2020-06-07
[lsp-pwsh] Use lsp-protocol

ericdallo:
https://github.com/emacs-lsp/lsp-mode/issues/1751

PR #1780, opened 2020-06-11
[lsp-pwsh] Use lsp-protocol and plists

ericdallo:
Replaces https://github.com/emacs-lsp/lsp-mode/pull/1754


lsp-pwsh.el
```diff
@@ -256,25 +260,24 @@ Must not nil.")
...
-                                             ("character" (- StartColumnNumber 1))))
+                                         (ht ("line" (- start-line-number 1))
+                                             ("character" (- start-column-number 1))))
                                         ("end"
-                                         (ht ("line" (- EndLineNumber 1))
-                                             ("character" (- EndColumnNumber 1))))))
-                           ("newText" Text))]))
+                                         (ht ("line" (- end-line-number 1))
+                                             ("character" (- end-column-number 1))))))
+                           ("newText" text))]))
```
kiennq, line 274: These should be changed to `lsp-make-text-edit` and `lsp-make-range`

ericdallo: I made this change and really forgot to push :sweat_smile:
Thanks for noticing it

Issue #1237, opened 2019-12-10
[Feature Request]Provide not so intrusive alternative for lsp-ui-sideline actions


yyoncho:
We may follow what vscode does and display bulb to indicate that there is an code action either by using overlay or indicator in the modeline.

ericdallo:
I could help on that, just don't know the better UI for that, if we show as an overlay on the right fringe/on the end of the line/on minibuffer, or something else...

ericdallo:
Thinking better, minibuffer could not be a good choice, as emacs uses to present all kind of interactions with the user, and a simple `message` can replace the code actions icon.

yyoncho:
yes, seems like there are 3 options.

1. Modeline( the easiest to implement)
2. Just a bulb like vscode do
3. bulb + summary as @matklad suggested

ericdallo:
for the modeline, we can just add and remove the icon to the `global-mode-string` easily

PR #1781, opened 2020-06-11
Add modeline code actions support

ericdallo:
As suggested [here](https://github.com/emacs-lsp/lsp-ui/pull/456), this PR adds support for code actions via modeline enabling it via `lsp-modeline-code-actions-mode` or setting `lsp-modeline-code-actions-enable` to `t`.

* Single code action:
image
* Multiple code actions:
image
* Fallback to unicode if `all-the-icons` is not installed:
image

Closes https://github.com/emacs-lsp/lsp-mode/issues/1237

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1934,7 +1934,7 @@ WORKSPACE is the workspace that contains the progress token."
                                        (interactive)
                                        (if single-action?
                                            (lsp-execute-code-action (lsp-seq-first actions))
-                                         (call-interactively #'lsp-execute-code-action)))))))
+                                         (lsp-execute-code-action (lsp--select-action actions))))))))
 
 (defun lsp-modeline--update-code-actions (actions)
   "Update modeline with new code ACTIONS."
@@ -1965,7 +1965,6 @@ WORKSPACE is the workspace that contains the progress token."
   :group 'lsp-mode
   :global nil
   :lighter ""
-  :init-value nil
   (cond
    (lsp-modeline-code-actions-mode
     (add-hook 'lsp-on-idle-hook 'lsp--modeline-check-code-actions nil t))
@@ -2630,6 +2629,7 @@ BINDINGS is a list of (key def cond)."
       "Tl" lsp-lens-mode (lsp-feature? "textDocument/codeLens")
       "TL" lsp-toggle-trace-io t
       "Th" lsp-toggle-symbol-highlight (lsp-feature? "textDocument/documentHighlight")
+      "Ta" lsp-modeline-code-actions-mode (lsp-feature? "textDocument/codeAction")
       "TS" lsp-ui-sideline-mode (featurep 'lsp-ui-sideline)
       "Td" lsp-ui-doc-mode (featurep 'lsp-ui-doc)
       "Ts" lsp-toggle-signature-auto-activate (lsp-feature? "textDocument/signatureHelp")
@@ -2711,6 +2711,7 @@ active `major-mode', or for all major modes when ALL-MODES is t."
        "T h" "toggle highlighting"
        "T L" "toggle log io"
        "T s" "toggle signature"
+       "T a" "toggle modeline code actions"
        "T S" "toggle sideline"
        "T d" "toggle documentation popup"
        "T p" "toggle signature help"
@@ -2816,7 +2817,8 @@ active `major-mode', or for all major modes when ALL-MODES is t."
       ["Add" lsp-workspace-folders-add]
       ["Remove" lsp-workspace-folders-remove]
       ["Open" lsp-workspace-folders-open])
-     ["Toggle Lenses" lsp-lens-mode]))
+     ["Toggle Lenses" lsp-lens-mode]
+     ["Toggle modeline code actions" lsp-modeline-code-actions-mode]))
   "Menu for lsp-mode.")
 
 (defun lsp-mode-line ()
@@ -6954,6 +6956,10 @@ returns the command to execute."
   "Autoconfigure `company', `flycheck', `lsp-ui',  if they are installed."
   (when (functionp 'lsp-ui-mode)
     (lsp-ui-mode))
+
+  (when (and lsp-modeline-code-actions-enable
+             (lsp--capability "codeActionProvider"))
+    (lsp-modeline-code-actions-mode 1))
 
   (cond
    ((or
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -650,6 +650,26 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-modeline-code-actions-enable nil
+  "Wheter to show code actions on modeline."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-modeline-code-actions-delay 0.2
+  "Number of seconds to wait before checking for code actions to show on modeline."
+  :type 'number
+  :group 'lsp-mode)
+
+(defcustom lsp-modeline-code-actions-kind-regex "quickfix.*\\|refactor.*"
+  "Regex for the code actions kinds to show in the modeline."
+  :type 'string
+  :group 'lsp-mode)
+
+(defface lsp-modeline-code-actions-face
+  '((t :foreground "cyan2"))
+  "Face used to highlight code action text on modeline."
+  :group 'lsp-faces)
+
 (defcustom lsp-after-diagnostics-hook nil
   "Hooks to run after diagnostics are received.
 Note: it runs only if the receiving buffer is open. Use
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -650,6 +650,26 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-modeline-code-actions-enable nil
```
yyoncho, line 653: Let's have this enabled by default and eventually coordinate with lsp-ui-sideline.

lsp-mode.el
```diff
@@ -650,6 +650,26 @@ If this is set to nil, `eldoc' will show only the symbol information."
...
+  "Wheter to show code actions on modeline."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-modeline-code-actions-delay 0.2
```
yyoncho, line 658: we should never create timers but use `lsp-on-*` hooks.

lsp-mode.el
```diff
@@ -650,6 +650,26 @@ If this is set to nil, `eldoc' will show only the symbol information."
...
+  :type 'number
+  :group 'lsp-mode)
+
+(defcustom lsp-modeline-code-actions-kind-regex "quickfix.*\\|refactor.*"
+  "Regex for the code actions kinds to show in the modeline."
+  :type 'string
+  :group 'lsp-mode)
+
+(defface lsp-modeline-code-actions-face
+  '((t :foreground "cyan2"))
```
yyoncho, line 669: I guess it will be better to inherit `success` or another face from the core face? Otherwise, this might look bad in different themes?

lsp-mode.el
```diff
@@ -1877,6 +1897,101 @@ WORKSPACE is the workspace that contains the progress token."
...
+                             :face 'lsp-modeline-code-actions-face
+                             :v-adjust -0.0575)
+    (propertize "💡" 'face 'lsp-modeline-code-actions-face)))
+
+(defun lsp--modeline-build-code-actions-string (actions)
+  "Build the string to be presented on modeline for code ACTIONS."
+  (-let* ((icon (lsp--modeline-code-actions-icon))
+          (first-action-string (propertize (->> actions
+                                                lsp-seq-first
+                                                (gethash "title")
```
yyoncho, line 1921: please use either `lsp:*` or destructuring. Lets write the new code against `lsp-protocol.el`.

lsp-mode.el
```diff
@@ -1877,6 +1897,101 @@ WORKSPACE is the workspace that contains the progress token."
...
+                                   (if single-action?
+                                       first-action-string
+                                     "select from multiple code actions"))
+                'mouse-face 'mode-line-highlight
+                'local-map (make-mode-line-mouse-map
+                            'mouse-1 (lambda ()
+                                       (interactive)
+                                       (if single-action?
+                                           (lsp-execute-code-action (lsp-seq-first actions))
+                                         (call-interactively #'lsp-execute-code-action)))))))
```
yyoncho, line 1942: Can we reuse the already calculated code actions?

lsp-mode.el
```diff
@@ -1877,6 +1897,101 @@ WORKSPACE is the workspace that contains the progress token."
...
+
+(defun lsp-modeline--update-code-actions (actions)
+  "Update modeline with new code ACTIONS."
+  (when lsp-modeline-code-actions-kind-regex
+    (setq actions (seq-filter (-lambda ((&hash "kind"))
```
yyoncho, line 1947: ditto for "kind"

lsp-mode.el
```diff
@@ -1877,6 +1897,101 @@ WORKSPACE is the workspace that contains the progress token."
...
+                                (or (not kind)
+                                    (s-match lsp-modeline-code-actions-kind-regex kind)))
+                              actions)))
+  (if (seq-empty-p actions)
+      (setq-local global-mode-string (remove '(t (:eval lsp--modeline-code-actions-string)) global-mode-string))
+    (progn
+      (setq lsp--modeline-code-actions-string (lsp--modeline-build-code-actions-string actions))
+      (add-to-list 'global-mode-string '(t (:eval lsp--modeline-code-actions-string))))))
```
yyoncho, line 1955: this looks problematic: we are adding the string to the "global" `global-modeline-variable` but then we remove it from the "local" `global-modeline-variable`.

lsp-mode.el
```diff
@@ -1877,6 +1897,101 @@ WORKSPACE is the workspace that contains the progress token."
...
+   (lsp--text-document-code-action-params)
+   #'lsp-modeline--update-code-actions
+   :mode 'alive
+   :cancel-token :lsp-modeline-code-actions))
+
+(defvar-local lsp--modeline-code-actions--timer nil)
+
+(defun lsp--modeline-check-code-actions ()
+  "Check for code actions showing it on modeline."
+  (if t
```
yyoncho, line 1970: this `if t` looks redundant.

lsp-mode.el
```diff
@@ -1877,6 +1897,101 @@ WORKSPACE is the workspace that contains the progress token."
...
+      (progn
+        (when lsp--modeline-code-actions--timer
+          (cancel-timer lsp--modeline-code-actions--timer))
+        (let ((buf (current-buffer)))
+          (setq lsp--modeline-code-actions--timer
+                (run-with-idle-timer lsp-modeline-code-actions-delay
```
kiennq, line 1976: Can we reuse diagnostic information here?
Or only if current line has diagnostics information?
From the definition of `CodeAction` and related, it has
```

	/**
	 * The diagnostics that this code action resolves.
	 */
	diagnostics?: Diagnostic[];
```
so it is tied to diagnostics information.

yyoncho:
> Looks good.
> 
> Can we add this to Toggles in the default lsp-managed-mode map and assign a shortcut as well?

I think that we should follow what we do for the rest of the features that run on idle. Note that code actions just like highlights, lenses, etc. are an optional features. So we should have something like

```
(when (and enable-action-modeline-auto-activate (server supports actions))
   .. enable modeline
)
```

in the auto configure code.

lsp-mode.el
ericdallo: Should I point this branch to plists branch so?

lsp-mode.el
ericdallo: Yeah, what hook do you think I should use here?

lsp-mode.el
ericdallo: How? are they set in some variable already?

lsp-mode.el
ericdallo: Hum, will `add-to-list` set to the global variable?
Should I use `setq-local` with `append` instead so?

lsp-mode.el
ericdallo: my bad, the idea was to check for the `lsp-modeline-code-actions-enable` variable here.

lsp-mode.el
ericdallo: Changed to `lsp-on-idle-hook`

lsp-mode.el
ericdallo: Good idea, changed to use emacs face `homoglyph`

lsp-mode.el
ericdallo: Oh, `lsp:code-actions-title` is on master, done!

lsp-mode.el
ericdallo: I didn't find it anywhere

lsp-mode.el
yyoncho: yes.

lsp-mode.el
yyoncho: aren't the actions present in the variable `actions`? `lsp-execute-code-action` will go to the server and calculate them again.

ericdallo: Oh I see what do you mean, the problem is that `lsp-execute-code-action` receive the action that it will be executed or nothing to present to user select the desired action, currently there'is no way to pass actions to it

yyoncho: ``` (lsp-execute-code-action (lsp--select-action actions)) ```?

lsp-mode.el
ericdallo: @yyoncho I confirmed that `add-to-list` here, change the local variable of the buffer, not the global one, so  I think it is ok to keep it this way, what do you think?

lsp-mode.el
ericdallo: Hum, I didn't notice we can do that, thanks!

lsp-mode.el
yyoncho: > @yyoncho I confirmed that add-to-list here, change the local variable of the buffer, not the global one

If setq-local runs first it creates buffer-local variable. Before that add-to-list works on the global variable.

lsp-mode.el
```diff
@@ -1877,6 +1892,85 @@ WORKSPACE is the workspace that contains the progress token."
...
+                                     "select from multiple code actions"))
+                'mouse-face 'mode-line-highlight
+                'local-map (make-mode-line-mouse-map
+                            'mouse-1 (lambda ()
+                                       (interactive)
+                                       (if single-action?
+                                           (lsp-execute-code-action (lsp-seq-first actions))
+                                         (lsp-execute-code-action (lsp--select-action actions))))))))
+
+(defun lsp-modeline--update-code-actions (actions)
```
yyoncho, line 1939: you have to call force-mode-line-update here.

lsp-mode.el
```diff
@@ -1877,6 +1892,85 @@ WORKSPACE is the workspace that contains the progress token."
...
+                              actions)))
+  (if (seq-empty-p actions)
+      (setq-local global-mode-string (remove '(t (:eval lsp--modeline-code-actions-string)) global-mode-string))
+    (progn
+      (setq lsp--modeline-code-actions-string (lsp--modeline-build-code-actions-string actions))
+      (add-to-list 'global-mode-string '(t (:eval lsp--modeline-code-actions-string))))))
+
+(defun lsp--modeline-check-code-actions (&optional buffer)
+  "Request code actions to update modeline for given BUFFER."
+  (when (or (not buffer)
```
yyoncho, line 1954: `when` and `buffer` are no longer needed.

lsp-mode.el
```diff
@@ -1877,6 +1892,85 @@ WORKSPACE is the workspace that contains the progress token."
...
+            (eq (current-buffer) buffer))
+    (lsp-request-async
+     "textDocument/codeAction"
+     (lsp--text-document-code-action-params)
+     #'lsp-modeline--update-code-actions
+     :mode 'alive
```
yyoncho, line 1960: as we discussed offline the mode should be either 'changed or 'tick

yyoncho:
Thank you!

PR #1782, opened 2020-06-11
Enable plists and fix tests + byte-compile loading

kiennq:
- Enable `lsp-use-plists` by default
- Fix travis
- Fix loading from byte compile errors of `lsp-get/put/...` not defined.

### DIFF
--- a/lsp-protocol.el
+++ b/lsp-protocol.el
@@ -37,7 +37,7 @@
 (defun lsp-keyword->string (keyword)
   (substring (symbol-name keyword) 1))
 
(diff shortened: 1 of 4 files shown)


yyoncho:
Looks good! Do you observe perf improvement when using plists?

kiennq:
> Looks good! Do you observe perf improvement when using plists?

I do feel it's faster :)

kiennq:
I will merge this. Is that okay? We should really test with plists enabled in this branch

ericdallo:
Nice!

PR #1785, opened 2020-06-11
lsp--capf-filter: optimze filtering logic and using while-no-input

kiennq:


kiennq:
@yyoncho @ericdallo Can you try with this to see if perf get better for dart?

ericdallo:
Yes, I will try, thanks for this!

lsp-mode.el
```diff
@@ -4501,38 +4501,41 @@ When the heuristic fails to find the prefix start point, return DEFAULT value."
...
-                                                             0.001)
-                                                         (or (get-text-property 0 'lsp-completion-score o)
-                                                             0.001))))
-                                          (put-text-property 0 1 'sort-score score o)
-                                          score)))))
-                  ;; TODO: pass additional function to sort the candidates
-                  (-map (-partial #'get-text-property 0 'lsp-completion-item)))
-           lsp-items)))
-    (-map (apply #'-rpartial #'lsp--make-completion-item plist) filtered-items)))
+  (let ((res (while-no-input
```
yyoncho, line 4504: can we wrap the whole all-completion?

lsp-mode.el
```diff
@@ -4501,38 +4501,41 @@ When the heuristic fails to find the prefix start point, return DEFAULT value."
...
+                                                            start-point (point))))
+                                                    (setq queries (plist-put queries start-point s))
+                                                    s)))
+                                       (fuz-query (or (plist-get fuz-queries start-point)
+                                                      (let ((s (lsp--regex-fuzzy query)))
+                                                        (setq fuz-queries
+                                                              (plist-put fuz-queries start-point s))
+                                                        s))))
+                                  (when (string-match fuz-query cand)
+                                    (setq cand (copy-sequence cand))
```
yyoncho, line 4522: is this copy-sequence needed?

kiennq: I forgot to remove this.

lsp-mode.el
kiennq: I don't think we need and should wrap `all-completion`. Since it can be interrupted in middle of calculating some thing like cache, we may have to live with that broken cache until either user have to abort completion of complete completion with broken candidates.

Just wrap some thing that's expensive and keep the other preparing/cleaning up code intact is better IMO.

If you want to have `while-no-input` to interrupting always, you can try https://github.com/kiennq/company-capf-noblock, which will forcefully abort current completion at point with `while-no-input`

yyoncho: We should if can go upstream.

About the expensive parts - I think that the initial sort is slow too. But let's get benchmarks for that.

ericdallo:
Tested and it seems really fast with Dart/Flutter :)

PR #1787, opened 2020-06-11
Optimize lsp--capf-guess-prefix

yyoncho:
- do not use -some-> and lsp--range-to-region
- cache (point)

Probably we should look for a way to cache buffer-substring-no-properties across
different lsp--capf-guess-prefix calls.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4458,7 +4458,7 @@ When the completion is incomplete, cache contains value of `incomplete'.")
     (set-marker nil))
   (setq lsp--capf-cache nil))
 
-(lsp-defun lsp--capf-guess-prefix ((&CompletionItem :label :insert-text? :text-edit?) default)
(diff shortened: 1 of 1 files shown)


kiennq:
> Probably we should look for a way to cache buffer-substring-no-properties across
> different lsp--capf-guess-prefix calls.

We should probably using memorize pattern and pass an additional `&optional cache` to `lsp--capf-guess-prefix`. That cache can be used to store `:start-point` and prefix.

kiennq:
Also, can you rebase against latest `plists`?

yyoncho:
> Also, can you rebase against latest `plists`?

Done. 

> We should probably using memorize pattern and pass an additional &optional cache to lsp--capf-guess-prefix. That cache can be used to store :start-point and prefix.

Yes. But for these optimizations we need benchmarks. I will capture the completion result from lsp-dart + sample file so we can use it for benchmarks.

PR #1788, opened 2020-06-11
Change lsp--modeline-code-actions-icon

leungbk:
* If all-the-icons is not already loaded, it never gets detected with
featurep. So require it optionally.

* declare-function does not understand when functions are defined in
macros. all-the-icons-octicon is defined by the macro
all-the-icons-define-icon. So set the fourth argument of
declare-function to t.


ericdallo:
Thank you for noticing it!

PR #1789, opened 2020-06-11
Use lsp-keymap-prefix on code actions for modeline

ericdallo:


lsp-mode.el
```diff
@@ -1924,7 +1930,7 @@ WORKSPACE is the workspace that contains the progress token."
                                         'display `((height 0.9))
                                         'face lsp-modeline-code-actions-face)))))
     (propertize string
-                'help-echo (concat "Apply code actions (s-l a a)\nmouse-1: "
+                'help-echo (concat (format "Apply code actions (%s a a)\nmouse-1: " lsp-keymap-prefix)
```
kiennq, line 1933: You may want to use a combination of `where-is-internal` and `key-description` to automatically get the key bind.
Something like 
```suggestion
                'help-echo (concat (format "Apply code actions (%s)\nmouse-1: " )
                                   (->> 'lsp-execute-code-action
                                        (where-is-internal)
                                        (-find (lambda (o) (not (member (aref o 0) '(menu-bar normal-state)))))
                                        (key-description)))
```

ericdallo: Good idea :)

lsp-mode.el
```diff
@@ -1917,14 +1923,19 @@ WORKSPACE is the workspace that contains the progress token."
                                                 (replace-regexp-in-string "[\n\t ]+" " "))
                                            'face lsp-modeline-code-actions-face))
           (single-action? (= (length actions) 1))
+          (execute-code-actions-binding (->> 'lsp-execute-code-action
+                                             (where-is-internal)
+                                             (-find (lambda (o)
```
yyoncho, line 1928: can you test that with lsp-execute-code-action not bound to keybinding?

kiennq:
Also, while we're at it, can we optimize the `codeAction` request so it's only called when there is `(lsp--cur-line-diagnotics)`?
It will reduce much more wasted CPU since mode line is update quite frequently.

lsp-mode.el
kiennq: It will return empty string

yyoncho: then we dont need the () in the format call bellow.

ericdallo:
@kiennq Good idea, but I tested checking `(lsp--cur-line-diagnotics)` with lsp-dart, but always return an empty vector even when has some actions, is there any other function that should I use?

lsp-mode.el
ericdallo: Good idea, fixed it

yyoncho:
> Also, while we're at it, can we optimize the `codeAction` request so it's only called when there is `(lsp--cur-line-diagnotics)`?

A lot of the actions are present even without diagnostics.

kiennq:
@ericdallo So it seems that Dart is not using diagnostic information for calculating code actions? All of language server I've used so far always relied on diagnostic information, so I'm not sure

ericdallo:
Well, I'm not sure what's is the right, but I think we should support as many languages as possible, so I think we should keep the "extra" actions call, right?

yyoncho:
Here we are looking for feature parity with vscode and it calls the language servers even if there is no diagnostic at the point. What I noticed that vscode does not call the server if there is space under the cursor though.

Also, I think that we should use the 'preferred' flag from the code action and display that one in the modeline.

ericdallo:
@yyoncho but the in the spec is an boolean the `isPreferred`, what do you mean?

yyoncho:
> @yyoncho but the in the spec is a boolean the `isPreferred`, what do you mean?

Yes. I mean that we should show that action's title.

ericdallo:
I see, just filter the actions that we already presents with `isPreferred`

yyoncho:
> I see, just filter the actions that we already presents with `isPreferred`

No, I mean that we should threat the action with isPreferred with higher priority and display it in the modeline if it is present. The rest stays the same.

ericdallo:
Oh sorry, it makes sense, I'll do that

yyoncho:
Just a heads up: we should declare :isPreferredSupport in order to receive isPreferred flag.

ericdallo:
Done @yyoncho :)

lsp-mode.el
```diff
@@ -1911,20 +1917,27 @@ WORKSPACE is the workspace that contains the progress token."
 (defun lsp--modeline-build-code-actions-string (actions)
   "Build the string to be presented on modeline for code ACTIONS."
   (-let* ((icon (lsp--modeline-code-actions-icon))
-          (first-action-string (propertize (->> actions
-                                                lsp-seq-first
-                                                lsp:code-action-title
-                                                (replace-regexp-in-string "[\n\t ]+" " "))
+          (first-action-string (propertize (or (-first #'lsp:code-action-is-preferred? actions)
```
yyoncho, line 1920: the first leg of the or will return CodeAction, the second leg will return string.

ericdallo: good catch, done!

ericdallo:
Can we merge this @yyoncho ?

PR #1790, opened 2020-06-12
Improve lsp-diagnose messages

amake:
Byte compilation check message was misleading: if native JSON is not enabled, then recompiling Emacs (hopefully with correct config) may enable it, but then you still have to recompile lsp-mode itself once native JSON is available.


yyoncho:
Thank you!

PR #1795, opened 2020-06-13
Add github action to run tests

ibizaman:
I stole the makefile from https://github.com/emacs-lsp/lsp-dart/blob/master/Makefile and the github workflow from https://github.com/emacs-lsp/lsp-dart/blob/master/.github/workflows/test.yml

Emacs 27 is not yet available but it could come soon https://github.com/purcell/nix-emacs-ci/issues/22.
That being said, the snapshot version is Emacs 28.0.50 https://github.com/ibizaman/lsp-mode/runs/767702850?check_suite_focus=true#step:4:160

I left the first target `all` as it was on purpose, to avoid the default make target to change. I didn't remove the travis file but I don't see why we should keep it, so I can remove it if you want. The github actions is much faster so that's nice too.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,43 @@
+name: CI
+
+on:
+    push:
+        branches:
+            - master
+    pull_request:
+        branches:
+            - master
+
+jobs:
+    build:
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                emacs-version:
+                    - 25.1
+                    - 25.2
+                    - 25.3
+                    - 26.1
+                    - 26.2
+                    - 26.3
+                    # - 27.1
+                    - snapshot
+
+        steps:
+            - uses: actions/checkout@v1
+
+            - uses: actions/setup-python@v1.1.1
+              with:
+                  python-version: "3.6"
+                  architecture: "x64"
+
+            - uses: purcell/setup-emacs@master
+              with:
+                  version: ${{ matrix.emacs-version }}
+
+            - uses: conao3/setup-cask@master
+              with:
+                  version: 0.8.4
+
+            - name: Run tests
+              run: "make ci"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,12 +1,6 @@
 name: CI
 
-on:
-    push:
-        branches:
-            - master
-    pull_request:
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,43 @@
+name: CI
+
+on:
+    push:
+        branches:
+            - master
+    pull_request:
+        branches:
+            - master
```
kiennq, line 9: Can we remove the restriction of only running on `master`?
This should be run on all, or at least some special branches
@yyoncho

README.md
```diff
@@ -6,7 +6,8 @@ description: Language Server Protocol support with multiples languages support f
 [![](https://melpa.org/packages/lsp-mode-badge.svg)](https://melpa.org/#/lsp-mode)
 [![](https://stable.melpa.org/packages/lsp-mode-badge.svg)](https://stable.melpa.org/#/lsp-mode)
 [![](https://badges.gitter.im/emacs-lsp/lsp-mode.svg)](https://gitter.im/emacs-lsp/lsp-mode)
-[![](https://travis-ci.org/emacs-lsp/lsp-mode.svg?branch=master>)](https://travis-ci.org/emacs-lsp/lsp-mode)
+[![](https://github.com/ibizaman/lsp-mode/workflows/CI/badge.svg)](https://github.com/ibizaman/lsp-mode/actions?query=workflow%3ACI)
```
kiennq, line 9: The badge is pointing to your repo. This need to be fixed.
Also I think we can just remove the badge, Github already showing status anyway.
Like this
image

yyoncho: > Github already showing status anyway.

we might need it for https://emacs-lsp.github.io/lsp-mode/

.github/workflows/test.yml
yyoncho: make sense - let's run it on top of everything in emacs-lsp

ericdallo:
I added master as a restricted branch too (require tests to pass on PRs, or admin force merge). This will happen only for new PRs/commits, so for open PRs (currently 12), the tests will not run until a new commit on that branch.

ericdallo:
Removed Travis integration: https://github.com/emacs-lsp/lsp-mode/commit/7b2a346e65041b4e7bef5c4d5a58d258b3569c03

PR #1796, opened 2020-06-13
Fix link to installation instructions

wkirschbaum:
The #supported-languages anchor tag does not exist anymore and seems
like it moved to a different site. This change fixes the installation
instructions message.


yyoncho:
Thank you!

PR #1797, opened 2020-06-13
lsp-csharp.el: add handler for the "omnisharp/client/findReferences"

razzmatazz:
This will make it possible to use code lenses to find symbol references.

Previously only reference count was displayed but now it possible to invoke action on the references element too (via `lsp-avy-lens`)

The related PR on server side is:
 - https://github.com/OmniSharp/omnisharp-roslyn/pull/1807


lsp-csharp.el
```diff
@@ -244,6 +244,21 @@ tarball or a zip file (based on a current platform) to TARGET-DIR."
   "Resolves path and arguments to use to start the server."
   (list (lsp-csharp--language-server-path) "-lsp"))
 
+(defun lsp-csharp--action-client-find-references (action)
+  "Reads first argument from ACTION as Location and displays xrefs for that location
+using the `textDocument/references' request."
+  (if-let ((arguments (gethash "arguments" action))
```
yyoncho, line 250: can you rewrite it using lsp-protocol.el bindings?

razzmatazz: I am not entirely sure.. Can you point me to any example of the code that does similar thing? 

I do need to define schema for this particular command using lsp-protocol.el, right?

yyoncho: AFACS there are existing types - Command, Location should be sufficient. Check this:

https://github.com/emacs-lsp/lsp-mode/blob/plists/lsp-mode.el#L2316

razzmatazz: I am not sure I understand your pointer to `lsp--lens-create-interactive-command`.. that function returns a lambda that invokes a command with command and args..

But.. do you mean, I need to decorate (with a macro presumably?) `lsp-csharp--action-client-find-references` function so `lsp-protocol.el` could somehow parse the conents of `action` into subcomponents that i actually need (and currently extract within if-let expression?)

kiennq: It should be something like this
```suggestion
  (-if-let* (((&Command: arguments?) action))
```

Or you can even change the `defun` to `lsp-defun` and
``` elisp
(lsp-defun lsp-csharp--action-client-find-references ((&Command :arguments?))
```

razzmatazz: I have updated the code using -if-let* / destructuring. Sorry for the long delay

yyoncho (approved):

yyoncho:
> clients/lsp-csharp.el:238:1:Error: Unused lexical variable ‘command’

There is one warning which is breaking the CI, can you fix it?

razzmatazz:
> > clients/lsp-csharp.el:238:1:Error: Unused lexical variable ‘command’
> 
> There is one warning which is breaking the CI, can you fix it?

Done

yyoncho:
Thank you!

PR #1799, opened 2020-06-13
Add breadcrumb on headerline

ericdallo:
* Add breadcrumb on headerline via `lsp-headerline-breadcrumb-mode`:
Screenshot from 2020-06-13 19-07-53
* Support for deprecated symbols:
image
* Add missing documentation to modeline code actions and breadcrumb.

Full screenshot:
image

In a next PR, the idea is to add mouse support to breadcrumb.


lsp-mode.el
```diff
@@ -1968,7 +1983,97 @@ WORKSPACE is the workspace that contains the progress token."
...
+
+(defun lsp--headerline-check-breadcrumb (&rest _)
+  "Request for document symbols to build the breadcrumb."
+  (when (lsp-feature? "textDocument/documentSymbol")
+    (-if-let* ((lsp--document-symbols-request-async t)
+               (symbols (lsp--get-document-symbols))
+               (symbols-hierarchy (lsp--headerline-symbols->symbols-hierarchy symbols)))
+        (progn
+          (setq lsp--headerline-breadcrumb-string (lsp--headerline-build-string symbols-hierarchy))
+          (add-to-list 'header-line-format '(t (:eval lsp--headerline-breadcrumb-string))))
```
kiennq, line 2064: I think you should call `add-to-list` once in minor activation and remove it when deactivation.
No need to remove the eval phrase when there's no symbol, you can just set `lsp--headerline-breadcrumb-string` to `""`, which will gracefully not showing it.
If you want to completely hide the header line instead, I guess we should not do it, since it will cause buffer jumping when show/hide header line.

ericdallo: Yeah, my idea was to hide it completely, but I think you are right about buffer jumping...
I'll change to set it to empty string :)

ericdallo: Done, thanks!

lsp-mode.el
```diff
@@ -1968,7 +1983,96 @@ WORKSPACE is the workspace that contains the progress token."
...
+  "Toggle breadcrumb on headerline."
+  :group 'lsp-mode
+  :global nil
+  (cond
+   (lsp-headerline-breadcrumb-mode
+    (add-to-list 'header-line-format '(t (:eval lsp--headerline-breadcrumb-string)))
+    (add-hook 'lsp-on-idle-hook 'lsp--headerline-check-breadcrumb nil t))
+   (t
+    (remove-hook 'lsp-on-idle-hook 'lsp--headerline-check-breadcrumb t)
+    (setq-local header-line-format (remove '(t (:eval lsp--headerline-breadcrumb-string)) header-line-format)))))
```
kiennq, line 2075: It seems that you're adding to (potentially) global `header-line-format` but removing from local `header-line-format`.
I think you can just use `setq` here. It has similar effect if `header-line-format` is already buffer local, and working even if there's only one global `header-line-format`

lsp-mode.el
```diff
@@ -1968,7 +1983,96 @@ WORKSPACE is the workspace that contains the progress token."
...
+
+(defun lsp--headerline-document-symbols->symbols-hierarchy (document-symbols)
+  "Convert DOCUMENT-SYMBOLS to symbols hierarchy."
+  (-let (((symbol &as &DocumentSymbol? :children?) (seq-some (-lambda ((symbol &as &DocumentSymbol :range))
+                                                          (-let (((beg . end) (lsp--range-to-region range)))
+                                                            (and (<= beg (point) end)
+                                                                 symbol)))
+                                                        document-symbols)))
+    (if children?
+        (append (list symbol) (lsp--headerline-document-symbols->symbols-hierarchy children?))
```
yyoncho, line 2037: seems like you need cons here instead of append.

lsp-mode.el
```diff
@@ -1968,7 +1983,96 @@ WORKSPACE is the workspace that contains the progress token."
...
+      (when symbol
+        (list symbol)))))
+
+(defun lsp--headerline-symbols-informations->symbols-hierarchy (symbols-informations)
+  "Convert SYMBOL-INFORMATIONS to symbols hierarchy."
+  (seq-some (-lambda ((symbol &as &SymbolInformation :location (&Location :range)))
+              (-let (((beg . end) (lsp--range-to-region range)))
+                (and (<= beg (point) end)
```
yyoncho, line 2045: when tested manually, when I move outside of an element with one char it still stays the same, so probably something here is off by one.

lsp-mode.el
```diff
@@ -1968,7 +1983,96 @@ WORKSPACE is the workspace that contains the progress token."
...
+  (when-let (first-symbol (lsp-seq-first symbols))
+    (if (lsp-symbol-information? first-symbol)
+        (lsp--headerline-symbols-informations->symbols-hierarchy symbols)
+      (lsp--headerline-document-symbols->symbols-hierarchy symbols))))
+
+(defun lsp--headerline-check-breadcrumb (&rest _)
+  "Request for document symbols to build the breadcrumb."
+  (when (lsp-feature? "textDocument/documentSymbol")
+    (-if-let* ((lsp--document-symbols-request-async t)
+               (symbols (lsp--get-document-symbols))
```
yyoncho, line 2060: when the symbols are not actually present and they are retrieved asynchronously we should manually refresh the modeline to force building the breadcrumb with latest data.

lsp-mode.el
```diff
@@ -1968,7 +1983,96 @@ WORKSPACE is the workspace that contains the progress token."
...
+      (all-the-icons-material "chevron_right"
+                             :face lsp-headerline-breadcrumb-face)
+    (propertize "›" 'face lsp-headerline-breadcrumb-face)))
+
+(lsp-defun lsp--headerline-breadcrumb-symbol-icon ((&DocumentSymbol :kind))
+  "Build the SYMBOL icon for headerline breadcrumb."
+  (when (require 'lsp-treemacs t t)
+    (treemacs-get-icon-value (lsp-treemacs-symbol-kind->icon kind))))
+
+(defun lsp--headerline-build-string (symbols-hierarchy)
```
yyoncho, line 2011: I get 

```
Debugger entered--Lisp error: (wrong-type-argument sequencep #<hash-table equal 4/4 0x1578ae42e959>)
  seq-length(#<hash-table equal 4/4 0x1578ae42e959>)
  #f(compiled-function (sequence) #<bytecode 0x1578ab916e79>)(#<hash-table equal 4/4 0x1578ae42e959>)
  apply(#f(compiled-function (sequence) #<bytecode 0x1578ab916e79>) #<hash-table equal 4/4 0x1578ae42e959> nil)
  seq-empty-p(#<hash-table equal 4/4 0x1578ae42e959>)
  seq-reduce((closure ((symbols-hierarchy . #<hash-table equal 4/4 0x1578ae42e959>) cl-struct-lsp-watch-tags cl-struct-lsp--client-tags lsp--log-lines c-basic-offset company-backends t) (last-symbol-name symbol-to-append) (let ((symbol2-name (if (lsp:document-symbol-deprecated\? symbol-to-append) (propertize (lsp:document-symbol-name symbol-to-append) 'font-lock-face 'lsp-headerline-breadcrumb-deprecated-face) (propertize (lsp:document-symbol-name symbol-to-append) 'font-lock-face lsp-headerline-breadcrumb-face))) (symbol2-icon (lsp--headerline-breadcrumb-symbol-icon symbol-to-append)) (arrow-icon (lsp--headerline-breadcrumb-arrow-icon))) (format "%s %s %s" last-symbol-name arrow-icon (if symbol2-icon (concat symbol2-icon symbol2-name) symbol2-name)))) #<hash-table equal 4/4 0x1578ae42e959> "")
  lsp--headerline-build-string(#<hash-table equal 4/4 0x1578ae42e959>)
```

when I am in pyls buffer.

lsp-mode.el
ericdallo: I didn't get this problem with `lsp-dart`, could you provide more details? Maybe the dart or other server is sending different ranges?

yyoncho: I tested it, seems like the issue is lack of refresh mentioned bellow.

lsp-mode.el
ericdallo: Fixed it!

lsp-mode.el
ericdallo: I see, so I guess it's fixed now after adding `(force-modeline-update)`

yyoncho:
Can you merge it in plists branch as well?

Few things that should address in the future: 

1. Breadcrumb functionality - I like the way it works in vscode, clicking symbol shows the stuff on the same level.
2. It contains also the path to the file in the project.
3. Allowing clicking with avy, similar to what we do with lsp-avy-lens. I think this will be very powerfull way to navigate, with no equivalent in emacs world. That would allow jumping in anything containing the current file. As an alternative to avy we may have a shortcut open Nth symbol from the breadcrumb. 
4. Review the whole document symbol story. Those patches we did for imenu are really ugly.

ericdallo:
> Can you merge it in plists branch as well?
> 
> Few things that can be addressed in the future:
> 
> 1. Breadcrumb functionality - I like the way it works in vscode, clicking symbol shows the stuff on the same level.
> 2. It contains also the path to the file in the project.
> 3. Allowing clicking with avy, similar to what we do with lsp-avy-lens. I think this will be very powerfull way to navigate, with no equivalent in emacs world. That would allow jumping in anything containing the current file. As an alternative to avy we may have a shortcut open Nth symbol from the breadcrumb.
> 4. Review the whole document symbol story. Those patches we did for imenu are really ugly.

1. we can start using completing-read when user clicks on it.
2. I thought about that, but don't you think that can be too much stuff on headerline? what about just the file name?
3. Agree, we can think in implement that.

yyoncho:
> 2\. I thought about that, but don't you think that can be too much stuff on headerline? what about just the file name?

I think that it looks fine in vscode. We could limit the length and use some kind of strategy to shorten the path(e. g. replace with the first letter of the directory). Anyways, it won't be a feature for everyone. 

Just to make sure we are on the same page, the proposal is to you the file name relative to project root.

PR #1806, opened 2020-06-15
lsp-f-ancestor-of?: fix wrongly detection logic

### LINKED ISSUE #1804: directory ancestry checking doesn't detect distinct directories
**Describe the bug**
Using lsp-mode-20200615.1829 for editing Go. I have two directories involved, call them:

```
~/src/hello
~/src/hello-there
```

Each are distinct Go-wise, having their own `go.mod` files. When opening a file in `~/src/hello-there`, I was not prompted to import the directory as a project root. I was prompted about whether to watch all files in `~/src/hello` which was surprising.

I believe this is because the [directory ancestry checking](https://github.com/emacs-lsp/lsp-mode/blob/bf2afe653a6cffcdbf336065d2c7200853feab36/lsp-mode.el#L1228) considers `~/src/hello-there` to be an ancestor of `~/src/hello`, even though they are at the same level under `~/src`.

Once I hand-edited `~/.emacs.d/.lsp-session-v1` to have entries for each directory, things worked as expected.

**To Reproduce**
Create two Go projects/modules, one in `~/src/hello` and another in `~/src/hello-there`. Probably have to open something in `~/src/hello` first so it gets added to the session list. After that, trying to open something in `~/src/hello-there` will use a session for `~/src/hello`.

**Expected behavior**
I expect `~/src/hello-there` to be considered a distinct project/directory since it's under `~/src` alongside `~/src/hello`.

gopls

**OS**
macOS 10.15.5 (Catalina)

kiennq:
Fix #1804

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1229,7 +1229,7 @@ Symlinks are not followed."
   "Return t if PATH-A is ancestor of PATH-B.
 Symlinks are not followed."
   (unless (lsp-f-same? path-a path-b)
-    (s-prefix? (lsp-f-canonical path-a)
+    (s-prefix? (concat (lsp-f-canonical path-a) (f-path-separator))
                (lsp-f-canonical path-b))))
 
 (defun lsp--merge-results (results method)
@@ -1646,7 +1646,7 @@ This set of allowed chars is enough for hexifying local file paths.")
         (if (and (directory-name-p file)
                  (not (lsp--string-match-any (lsp-file-watch-ignored) (f-join dir (f-filename file)))))
             (let* ((leaf (substring file 0 (1- (length file))))
-                   (full-file (concat dir "/" leaf)))
+                   (full-file (f-join dir leaf)))
               ;; Don't follow symlinks to other directories.
               (unless (file-symlink-p full-file)
                 (setq result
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -1229,7 +1229,7 @@ Symlinks are not followed."
   "Return t if PATH-A is ancestor of PATH-B.
 Symlinks are not followed."
   (unless (lsp-f-same? path-a path-b)
-    (s-prefix? (lsp-f-canonical path-a)
+    (s-prefix? (concat (lsp-f-canonical path-a) "/")
```
yyoncho, line 1232: "/" is os specific.

kiennq: Is it? The `lsp-f-canonical` is already normalizing path to use `/` already.
I've tested on Windows and Ubuntu, it seems working fine. Of course I can try to use `f--path-separator` too, but it seems to be over kill?

yyoncho: can you then add tests illustrating how it works with windows paths?

kiennq: I've added tests and also change to use `f--path-separator` too

kiennq: Revert the new test cases, it will not pass since the CI machine is in Ubuntu. Letting CI running on Windows is a pain so ...

yyoncho: > Revert the new test cases, it will not pass since the CI machine is in Ubuntu. Letting CI running on Windows is a pain so ...

It is still fine to have them wrapped conditionally to run only on windows. It will be good for documentation purpose.

PR #1808, opened 2020-06-16
lsp-modeline-code-action-mode: don't set/remove global-mode-string too much

kiennq:
We can just update `lsp--modeline-code-actions-string` instead of whole adding/remove from `global-mode-string`

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1938,7 +1938,7 @@ WORKSPACE is the workspace that contains the progress token."
           (string (if single-action?
                       (format " %s %s " icon first-action-string)
                     (format " %s %s %s " icon first-action-string
-                            (propertize (format "(%d more)" (seq-length actions))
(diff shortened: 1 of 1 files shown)


ericdallo:
Could you rebase it with master to CI run?
I fixed it on master a few minutes ago

kiennq:
> Could you rebase it with master to CI run?
> I fixed it on master a few minutes ago

@ericdallo  Can you merge #1809? I don't have permission to push on master. And we need to trigger CI on pull request too

ericdallo:
Done @kiennq

yyoncho:
> @ericdallo Can you merge #1809? I don't have permission to push on master. And we need to trigger CI on pull request to

@ericdallo can we add push rights for every member? We have been like that for a long time and there were no issues with that so IMO ATM there is no reason to have a stricter policy.

PR #1809, opened 2020-06-16
ci: fix gh action to triggered on pull_request too

kiennq:
We need this on `pull_request` too. Else pull request from other repository will not trigger ci


ericdallo:
Thanks, I forgot that it needs to add `pull_request`

PR #1811, opened 2020-06-16
lsp-diagnostics-modeline: perf improvement

### LINKED ISSUE #1793: lsp-diagnostics-modeline-mode is extremely slow
... it runs pretty much on keypress and it should be rewritten to avoid creating intermediate objects.

kiennq:
Attempt to fix #1793.

- Update mode line only when new diagnostic information is available.
- Modify `lsp-diagnostics` and `lsp--diagnostics-modeline-statistics` to avoid intermediate object.
- Cache the mode line string and recalculate only once

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1894,7 +1894,7 @@ The `:global' workspace is global one.")
 
 (defun lsp--diagnostics-reset-modeline-cache ()
   ""
-  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) :global)
+  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) nil)
   (plist-put lsp--diagnostics-modeline-wks->strings :global nil)
   (setq lsp--diagnostics-modeline-string nil))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1835,22 +1835,15 @@ WORKSPACE is the workspace that contains the progress token."
                                 (->> workspace
                                      (lsp--workspace-diagnostics)
                                      (maphash (lambda (file-name diagnostics)
-                                                (puthash file-name
-                                                         (append (gethash file-name result) diagnostics)
-                                                         result)))))
+                                                (if-let ((cur (gethash file-name result)))
+                                                    (nconc cur diagnostics)
+                                                  (puthash file-name diagnostics result))))))
                               workspaces)
                         result)))
       (ht)))
 
 
 ;; diagnostic modeline
-(defun lsp--severity-code->severity (severity-code)
-  (cl-case severity-code
-    (1 'error)
-    (2 'warning)
-    (3 'info)
-    (4 'hint)))
-
 (defcustom lsp-diagnostics-modeline-scope :workspace
   "The scope "
   :group 'lsp-mode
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -1867,34 +1870,42 @@ WORKSPACE is the workspace that contains the progress token."
...
-                 (propertize (format "%s" count)
-                             'face (cl-case code
-                                     ('error 'error)
-                                     ('warning 'warning)
-                                     ('info 'success)
-                                     ('hint 'success)))))
-         (s-join "/")
-         (format "%s"))))
+                              (-flatten-n 1))))))
+    (--> diagnostics
```
yyoncho, line 1874: the whole block till s-join could be rewritten with one -reduce. Without that, we will create 2 intermediate copies of the initial collection(one coming from -group-by and one from -map). Even the flatten above can be removed.

kiennq: I've rewritten it and also remove the sort.

lsp-mode.el
```diff
@@ -1859,42 +1852,63 @@ WORKSPACE is the workspace that contains the progress token."
...
-    (setq-local global-mode-string
-                (cond ((and lsp-diagnostics-modeline-mode
-                            (not (-contains? global-mode-string status)))
-                       (cons status global-mode-string))
-                      (t (remove status global-mode-string))))))
+  (cond
+    (lsp-diagnostics-modeline-mode
+     (add-to-list 'global-mode-string '(t (:eval lsp--diagnostics-modeline-string)))
+     (add-hook 'lsp-after-open-hook 'lsp--diagnostics-update-modeline nil t)
+     (add-hook 'lsp-diagnostics-updated-hook 'lsp--diagnostics-update-modeline nil t))
```
yyoncho, line 1907: this would mean that for all of the open buffers in lsp-mode we will update the lsp--diagnostics-modeline-string, right? 

Can we change the code like that:

1. When there is diagnostics update lsp--diagnostics-modeline-string is set to nil.
2. When modeline is built it either uses lsp--diagnostics-modeline-string if present or builts it and caches it.

WDYT?

kiennq: This one is local hook so it will run once when buffer is opened and every time diagnostic is updated. `lsp--diagnostics-modeline-string` is also a buffer local variable too.

When `lsp-diagnostics-modeline-scope` is `:file`, it's pretty optimal and we need to calculate it anyway.
When `lsp-diagnostics-modeline-scope` is `:workspace`, we're wasting recalculating the modeline for every opened buffers in the workspace.
When `lsp-diagnostics-modeline-scope` is `:global`, we're wasting recalculating the modeline for all opened lsp buffers.

We can try using cache to reduce that. But I would expected the number of opened buffers will not be that much.

yyoncho:
Looks good. I will do manual testing tomorrow and I will proceed with the merge.

lsp-mode.el
```diff
@@ -1859,42 +1852,88 @@ WORKSPACE is the workspace that contains the progress token."
...
+                                  ((equal i lsp/diagnostic-severity-hint) 'success)))))))
+             (cl-incf i))
+           strs)
+         (s-join "/"))))
+
+(defun lsp--diagnostics-reset-modeline-cache ()
+  ""
+  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) nil)
+  (plist-put lsp--diagnostics-modeline-wks->strings :global nil)
+  (setq lsp--diagnostics-modeline-string nil))
```
yyoncho, line 1899: that would reset it only in the current buffer, right? 

(As a side note using only this variable for local caching should be enough, it will be fine if we recompute once per buffer).

lsp-mode.el
```diff
@@ -1859,42 +1852,88 @@ WORKSPACE is the workspace that contains the progress token."
...
+                                  ((equal i lsp/diagnostic-severity-warning) 'warning)
+                                  ((equal i lsp/diagnostic-severity-information) 'success)
+                                  ((equal i lsp/diagnostic-severity-hint) 'success)))))))
+             (cl-incf i))
+           strs)
+         (s-join "/"))))
+
+(defun lsp--diagnostics-reset-modeline-cache ()
+  ""
+  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) nil)
```
kiennq, line 1897:
```suggestion
  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) :global)
```

lsp-mode.el
kiennq: > that would reset it only in the current buffer, right?

Yes, but this is only need if the `modeline-scope` is `:file`. In that case only reset/update current buffer is totally fine.
For case of `:workspace` and `:global` the cached modeline is properly reset already.

lsp-mode.el
```diff
@@ -1859,42 +1852,88 @@ WORKSPACE is the workspace that contains the progress token."
...
+                                  ((equal i lsp/diagnostic-severity-warning) 'warning)
+                                  ((equal i lsp/diagnostic-severity-information) 'success)
+                                  ((equal i lsp/diagnostic-severity-hint) 'success)))))))
+             (cl-incf i))
+           strs)
+         (s-join "/"))))
+
+(defun lsp--diagnostics-reset-modeline-cache ()
+  ""
+  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) :global)
```
kiennq, line 1897:
```suggestion
  (plist-put lsp--diagnostics-modeline-wks->strings (car (lsp-workspaces)) nil)
```

lsp-mode.el
yyoncho: If you are in file A and update something that will result in a change of diagnostics in buffer B when you switch to B it will show stale diagnostics, right? I guess we might leave it as it is.

kiennq: > If you are in file A and update something that will result in a change of diagnostics in buffer B when you switch to B it will show stale diagnostics, right?

Ah, yes, it might be if the `modeline-scope` is `:file`. The default is `:workspace` so I guess it would be fine. We can try to reset all buffer strings but I doubt it might be a perf hit.

lsp-mode.el
```diff
@@ -1859,42 +1852,88 @@ WORKSPACE is the workspace that contains the progress token."
...
+               strs
+               (i 0))
+           (mapc (lambda (buf-diags)
+                   (mapc (lambda (diag)
+                           (-let [(&Diagnostic? :severity?) diag]
+                             (when severity?
+                               (cl-incf (aref stats severity?)))))
+                         buf-diags))
+                 diagnostics)
+           (while (< i lsp/diagnostic-severity-max)
```
yyoncho, line 1879: Seems like this can be rewritten using `seq-map-indexed`.

kiennq: I've tried that before, it either returns a list of results which we will need to filter out `0` value or we can let it `nconc` to `strs` and waste the returned list.

PR #1817, opened 2020-06-18
Check return value before access hashtable

conao3:
Hi!
Thanks for such a great package!

I found `lsp--parser-on-message` occuor error in some condition.

This error occurs when I using `lsp-mode` with `clangd` for cpp
files.

Here is the backtrace. So, I fix code to check return value
before access hashtable via `gethash`.
```
Debugger entered--Lisp error: (wrong-type-argument hash-table-p nil)                             
  gethash("range" nil)                                                                           
  (let* ((hover input0) (range\? (gethash "range" hover)) (contents (gethash "contents" hover))) 
  (closure (cl-struct-lsp--log-entry-tags cl-struct-lsp-session-tags cl-struct-lsp--workspace-tag
  apply((closure (cl-struct-lsp--log-entry-tags cl-struct-lsp-session-tags cl-struct-lsp--workspa
  (save-current-buffer (set-buffer buf) (apply callback args))                                   
  (closure ((cleanup-hooks closure ((buf . #<buffer main.cpp>) (hooks (kill-buffer-hook . t) (aft
  funcall((closure ((cleanup-hooks closure ((buf . #<buffer main.cpp>) (hooks (kill-buffer-hook .
  (progn (funcall callback (if no-merge results (lsp--merge-results (-map (function cl-rest) resu
  (if (and (not (eq (length errors) (length workspaces))) (eq (+ (length errors) (length results)
  (closure ((errors) (results (#s(lsp--workspace :ewoc nil :server-capabilities #<hash-table equa
  funcall((closure ((errors) (results (#s(lsp--workspace :ewoc nil :server-capabilities #<hash-ta
  (closure ((callback closure ((errors) (results (#s(lsp--workspace :ewoc nil :server-capabilitie
  funcall((closure ((callback closure ((errors) (results (#s(lsp--workspace :ewoc nil :server-cap
  (progn (funcall callback (lsp:json-response-result json-data)) (remhash id (progn (or (and (mem
  (if callback (progn (funcall callback (lsp:json-response-result json-data)) (remhash id (progn 
  (let* ((--dash-source-146-- (gethash id (progn (or (and (memq ... cl-struct-lsp--client-tags) t
  (cond ((eq val (quote response)) (progn (or id (cl--assertion-failed (quote id))) nil) (let* ((
  (let* ((val (lsp--get-message-type json-data))) (cond ((eq val (quote response)) (progn (or id 
  (let* ((client (progn (or (and (memq (type-of workspace) cl-struct-lsp--workspace-tags) t) (sig
  (let ((lsp--cur-workspace workspace)) (let* ((client (progn (or (and (memq ... cl-struct-lsp--w
  (condition-case err (let ((lsp--cur-workspace workspace)) (let* ((client (progn (or (and ... t)
  lsp--parser-on-message(#<hash-table equal 3/65 0x26751f9> #s(lsp--workspace :ewoc nil :server-c
  #f(compiled-function (proc input) #<bytecode
  0x2d82911>)(#<process clangd<1>> "Content-Length:
 ```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6508,30 +6508,32 @@ WORKSPACE is the active workspace."
         (pcase (lsp--get-message-type json-data)
           ('response
            (cl-assert id)
-           (-let [(callback _ method start-time before-send) (gethash id (lsp--client-response-handlers client))]
-             (when lsp-print-io
-               (lsp--log-entry-new
-                (lsp--make-log-entry method id data 'incoming-resp
-                                     (/ (nth 2 (time-since before-send)) 1000))
-                workspace))
-             (when callback
-               (funcall callback (lsp:json-response-result json-data))
-               (remhash id (lsp--client-response-handlers client))
-               (lsp--log-request-time server-id method id start-time before-send
-                                      received-time after-parsed-time (current-time)))))
+           (when-let ((obj (lsp--client-response-handlers client)))
+             (-let [(callback _ method start-time before-send) (gethash id obj)]
+               (when lsp-print-io
+                 (lsp--log-entry-new
+                  (lsp--make-log-entry method id data 'incoming-resp
+                                       (/ (nth 2 (time-since before-send)) 1000))
+                  workspace))
+               (when callback
+                 (funcall callback (lsp:json-response-result json-data))
+                 (remhash id (lsp--client-response-handlers client))
+                 (lsp--log-request-time server-id method id start-time before-send
+                                        received-time after-parsed-time (current-time))))))
           ('response-error
            (cl-assert id)
-           (-let [(_ callback method start-time before-send) (gethash id (lsp--client-response-handlers client))]
-             (when lsp-print-io
-               (lsp--log-entry-new
-                (lsp--make-log-entry method id data 'incoming-resp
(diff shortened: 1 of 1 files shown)


yyoncho:
Isn't that fixed by https://github.com/emacs-lsp/lsp-mode/commit/890db8cf0b208fd855c9490f80154ffd606a6637 ?

conao3:
Yes! There're no issues after update!
Thanks.

conao3:
Don't you need this commit?
It is true that we don't get any errors, but it doesn't check the return value.
I think merging this code would increase the robustness.

yyoncho:
> I think merging this code would increase the robustness.

I am not quite sure which particular condition we are handling with this change, can you elaborate?

conao3:
If `lsp--client-response-handlers` returns `nil`, then No error.
`lsp--client-response-handlers` will never return `nil`. If this assumption holds, then the commit is not necessary.

yyoncho:
lsp--client-response-handlers cannot return nil. Getting the value from it might return nil but I prefer having an error in this case because it will be an indication that something is wrong.

conao3:
OK, agree. thanks.

PR #1819, opened 2020-06-19
Cleanup diagnostics after the buffer has changed

### LINKED ISSUE #1792: We should forget diagnostics after file is modified
not doing that is causing stale diagnostics until the new one are received.

yyoncho:
Fixes #1792

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4443,8 +4443,20 @@ Added to `after-change-functions'."
         ;; force cleanup overlays after each change
         (lsp--remove-overlays 'lsp-highlight)
         (lsp--after-change  (current-buffer))
-        (setq lsp--signature-last-index nil)
-        (setq lsp--signature-last nil)))))
+        (setq lsp--signature-last-index nil
+              lsp--signature-last nil)
+        ;; cleanup diagnostics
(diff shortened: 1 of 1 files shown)


yyoncho:
@kiennq can you help with testing on windows.

lsp-mode.el
```diff
@@ -4443,8 +4443,20 @@ Added to `after-change-functions'."
...
+              lsp--signature-last nil)
+        ;; cleanup diagnostics
+        (lsp-foreach-workspace
+         (-let [diagnostics (lsp--workspace-diagnostics lsp--cur-workspace)]
+           (pcase system-type
+             (`windows-nt (let ((buffer-name (s-downcase (buffer-file-name))))
+                            (->> diagnostics
+                                 (ht-keys)
+                                 (mapc (lambda (file-name)
+                                         (when (string= (s-downcase file-name)
```
kiennq, line 4456: Can this be done in more simpler way with `lsp--fix-path-casing`?
I mean we can modify `lsp-diagnostics` to put hash with fixed casing path. And then here we can just simply remove the fixed casing path element.

yyoncho: I am doing exactly that right now.

PR #1820, opened 2020-06-19
Avoid downcasing paths in uri conversion code

### LINKED ISSUE #1792: We should forget diagnostics after file is modified

yyoncho:
... do that when storing the URI.

Fixes https://github.com/emacs-lsp/lsp-java/issues/190
Fixes #1792

- Cleanup diagnostics after the buffer has changed

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1570,7 +1570,6 @@ On other systems, returns path without change."
                           file))))
     (->> file-name
          (concat (-some #'lsp--workspace-host-root (lsp-workspaces)))
-         (lsp--fix-path-casing)
          (lsp-remap-path-if-needed))))
 
 (defun lsp--buffer-uri ()
@@ -1904,7 +1903,7 @@ The `:global' workspace is global one.")
                 (let ((str (lsp--diagnostics-modeline-statistics)))
                   (if (string-empty-p str) ""
                     (concat " " str)))))
-    (setq lsp--diagnostics-modeline-string 
+    (setq lsp--diagnostics-modeline-string
           (cl-case lsp-diagnostics-modeline-scope
             (:file (or lsp--diagnostics-modeline-string
                        (calc-modeline)))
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -4449,8 +4448,12 @@ Added to `after-change-functions'."
         ;; force cleanup overlays after each change
         (lsp--remove-overlays 'lsp-highlight)
         (lsp--after-change  (current-buffer))
-        (setq lsp--signature-last-index nil)
-        (setq lsp--signature-last nil)))))
+        (setq lsp--signature-last-index nil
+              lsp--signature-last nil)
+        ;; cleanup diagnostics
```
alanz, line 4453: This behaviour should be optional.  I used to be able to chase through a list of errors in a file, now I have to wait for an update inbetween, and it slows things down.

yyoncho: Can you elaborate? There is flycheck-idle-change-delay which can be used to make the update faster. Also, lsp-idle-delay could be made more aggressive.

alanz: Using the haskell-language-server, I often make a small change to a type that generates a lot of errors in a file. This is expected, and why there are types. It means I have to make trivial changes in a lot of locations.  Prior to the change to lsp-mode to have the diagnostics cleared every time I changed the file, I could chase through the file, knowing it would eventually stabilise and I could confirm I had fixed things properly.

Now I have to make one fix, wait for the errors to re-appear, and only then navigate to the next one. It is annoyingly slower.

yyoncho: I see. This should be fixing your issue: https://github.com/yyoncho/lsp-mode/commit/e69cd68ea0520b727190ef6cea486612562243e3 setting the variable to nil will prevent the diagnostics cleanup. (ATM I do not have the time to do detailed testing on it).

alanz: FYI I am now running your branch, toggled the setting, and am happy again.

Thanks.

yyoncho: I will provide a PR tomorrow.

PR #1822, opened 2020-06-19
fix go-dot-mod-mode language id

rski:
Currently gopls tries to parse go.mod files as go files which results
in "LSP :: Error from the Language Server [snip] expected 'package',
found module) (Unknown error)" errors

The correct language id from
github.com/golang/tools/internal/lsp/source/util.go is go.mod, so use
that.


yyoncho:
@rski thank you!

PR #1825, opened 2020-06-19
Modify declare-function calls for breadcrumb

leungbk:
* treemacs-get-icon-value is in treemacs-icons
* lsp-treemacs-symbol-kind->icon is a normal function, so we can set
file-only to nil


yyoncho:
@leungbk previous time you mentioned that there is a way to check this automatically, can we add that to CI?

ericdallo:
Yeah, would be a good idea.
I dont't know how to add that to CI byte compiler, do you know @leungbk ?

PR #1827, opened 2020-06-20
Fix starting perl server, change perl server port

klausweiss:
With current implementation perl is given one string parameter `-MPerl::LanguageServer -e Perl::LanguageServer::run -- --port %o --version %s` which results in an error on stderr and failing to start (the status is stuck at `starting`):
```
Bareword found where operator expected at - line 0, near "--port"
	(Missing operator before port?)
Number found where operator expected at - line 0, near "port 32443"
	(Do you need to predeclare port?)
Bareword found where operator expected at - line 0, near "--version"
	(Missing operator before version?)
Number found where operator expected at - line 0, near "version 2.1.0"
	(Do you need to predeclare version?)
```

I spit the cli arguments so that the server starts.

Changing the port to its octal representation was a bit confusing, so `%o` was changed to `%d` in string format. I couldn't figure out a reason for using `%o`. If there was one, please correct me.

Sorry for not opening an issue first, but the fix was pretty straightforward.

### DIFF
--- a/lsp-perl.el
+++ b/lsp-perl.el
@@ -55,7 +55,7 @@
                                    (lambda ()
                                      (list lsp-perl-language-server-path
                                            "-MPerl::LanguageServer" "-e" "Perl::LanguageServer::run" "--" 
-                                           (format "--port %o --version %s"
+                                           (format "--port %d --version %s"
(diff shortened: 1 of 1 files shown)


yyoncho:
@kubo39 willing to take a look?

yyoncho:
@klausweiss 

PR #1831, opened 2020-06-21
Run flycheck if diagnostics are received after the buffer is saved

yyoncho:
- if there is no `idle-change` and buffer is saved we should run flycheck


yyoncho:
@scturtle

scturtle:
It works for me!

PR #1837, opened 2020-06-24
lsp-configure: allow rerun buffer configure on capabilities changes

kiennq:
Since there can be optional capabilities that can be enabled by request from server, we should move some of capabilities check to `lsp-configure-buffer` and have it to be also called when `client/registerCapability`.
This fixes the issue of `lsp-pwsh` doesn't have mode line action enabled automatically.

I've also add `lsp--clean-company`, `lsp-unconfig-buffer` to allow to cleanly remove configure on `lsp-disconnect`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -72,6 +72,7 @@
 (defvar c-basic-offset)
 (defvar yas-inhibit-overlay-modification-protection)
 (defvar yas-indent-line)
+(defvar dap-auto-configure-mode)
 
 (defconst lsp--message-type-face
   `((1 . ,compilation-error-face)
@@ -3526,11 +3527,14 @@ To find out what capabilities support your server use `M-x lsp-describe-session'
   "Cleanup buffer state.
 When a workspace is shut down, by request or from just
 disappearing, unset all the variables related to it."
-  (let ((proc (lsp--workspace-cmd-proc lsp--cur-workspace)))
+  (let ((proc (lsp--workspace-cmd-proc lsp--cur-workspace))
+        (buffers (lsp--workspace-buffers lsp--cur-workspace)))
     (when (process-live-p proc)
       (kill-process proc))
-    (unless lsp--buffer-workspaces
-      (lsp-managed-mode -1))))
+    (mapc (lambda (buf)
+            (with-current-buffer buf
+              (lsp-managed-mode -1)))
+          buffers)))
 
 (defun lsp--client-capabilities (&optional custom-capabilities)
   "Return the client capabilities."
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3992,6 +3998,20 @@ in that particular folder."
            (lsp--on-change-debounce buffer)
            (lsp--on-idle buffer)))))))
 
+(defun lsp-unconfig-buffer ()
+  (when lsp-modeline-code-actions-mode
+    (lsp-modeline-code-actions-mode -1))
+  (when lsp-headerline-breadcrumb-mode
+    (lsp-headerline-breadcrumb-mode -1))
+  (when (bound-and-true-p dap-auto-configure-mode)
```
yyoncho, line 4006: remove the dap part - otherwise disconnecting from one buffer will turn off dap-mode globally.

lsp-mode.el
```diff
@@ -3992,6 +3998,20 @@ in that particular folder."
            (lsp--on-change-debounce buffer)
            (lsp--on-idle buffer)))))))
 
+(defun lsp-unconfig-buffer ()
+  (when lsp-modeline-code-actions-mode
```
yyoncho, line 4002: are those checks needed or we can directly try to turn it off? At least I have never seen a code that checks if the mode is on before disabling it.

kiennq: I think it's needed, it will run the `t` block in the minor mode initialization. So better guard for it do not run again.

lsp-mode.el
```diff
@@ -6441,13 +6461,17 @@ WORKSPACE is the active workspace."
...
+                      (lsp-unconfig-buffer)
+                      (lsp-configure-buffer)
                       nil)
                      ((equal method "window/showMessageRequest")
                       (let ((choice (lsp--window-log-message-request params)))
                         `(:title ,choice)))
                      ((equal method "client/unregisterCapability")
                       (mapc #'lsp--server-unregister-capability
                             (lsp:unregistration-params-unregisterations params))
+                      (lsp-unconfig-buffer)
```
yyoncho, line 6473: this should be called in all buffers of the workspace.

lsp-mode.el
```diff
@@ -8059,6 +8103,7 @@ such."
   (interactive)
   (with-no-warnings (when (functionp 'lsp-ui-mode) (lsp-ui-mode -1)))
   (lsp--text-document-did-close t)
+  (lsp-unconfig-buffer)
```
yyoncho, line 8106: can we call unconfig buffer when disabling lsp-managed-mode? It seems like it should be there. The same applies to setting workspaces to nil. Feel free to merge without applying this comment.

kiennq: Nice, I will include `lsp-configure-buffer` and `lsp-unconfig-buffer` in enable/disable `lsp-managed-mode` respectively.
For setting workspaces to nil, it would be better to set all `defvar-local` variables to nil when disable `lsp-managed-mode`.
That can be another PR.

PR #1839, opened 2020-06-24
Fix lens regression

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-metals/issues/11

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2480,7 +2480,7 @@ BUFFER-MODIFIED? determines whether the buffer is modified or not."
   (-doto (make-sparse-keymap)
     (define-key [mouse-1] (lsp--lens-create-interactive-command command))))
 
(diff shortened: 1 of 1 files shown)


yyoncho:
@kurnevsky I was unable to test with metals but most likely this will fix your issue.

lsp-mode.el
```diff
@@ -2480,7 +2480,7 @@ BUFFER-MODIFIED? determines whether the buffer is modified or not."
   (-doto (make-sparse-keymap)
     (define-key [mouse-1] (lsp--lens-create-interactive-command command))))
 
-(lsp-defun lsp--lens-create-interactive-command ((&CodeLens :command?))
+(lsp-defun lsp--lens-create-interactive-command (command?)
```
kurnevsky, line 2483: Should it be just a `defun`?

yyoncho: Yes.

kurnevsky:
Just tested - works fine.

PR #1841, opened 2020-06-25
Show message when no formatting changes are returned by LS.

### LINKED ISSUE #985: lsp-format-buffer and error feedback when there is a formatting error
**Describe the bug**
If there is a syntax error in the file and I call `lsp-format-buffer`, I do not get visual feedback that the formatting failed.

I can see the error the formatter generated in `*lsp-log*`.

**To Reproduce**
Create a syntax error in a file and then call  `lsp-format-buffer`.

**Expected behavior**
An error should be shown to the user. Most likely in the minibuffer.

[metals-emacs](https://scalameta.org/metals/docs/editors/emacs.html)

**OS**
Ubuntu 19.04

**Error callstack**
N/A

jakejx:
Just a simple change to show a message to the user when the language server does not return any formatting changes.

Closes #985.


yyoncho:

PR #1844, opened 2020-06-26
lsp--capf-guess-prefix: make sure the char before prefix is not word constitute

kiennq:
Add addition heuristic to make sure the character before prefix is not word constitute.
This is done so for example in case of 
``` js
if (conf<|>
```
`f` will not be mistaken as prefix.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4769,27 +4769,28 @@ When the completion is incomplete, cache contains value of `incomplete'.")
     (set-marker nil))
   (setq lsp--capf-cache nil))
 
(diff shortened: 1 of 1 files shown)


yyoncho:
This will most likely work but since we stole the idea of calculating prefix like that from lsp4e, have you checked how do they handle this situation?

kiennq:
Did lsp4e works with `ts-ls`?
I've checked their implementation again and it seems they're just doing prefix match check like what we are doing until now.
https://git.eclipse.org/c/lsp4e/lsp4e.git/tree/org.eclipse.lsp4e/src/org/eclipse/lsp4e/operations/completion/LSIncompleteCompletionProposal.java#n378

Do you have `lsp4e`? Can you do a quick check?

PR #1855, opened 2020-06-29
lsp-elm/css/bash/purescript: add auto download function

kiennq:


lsp-mode.el
```diff
@@ -7422,7 +7422,7 @@ JavaScript file, tsserver.js (the *.js is required for Windows)."
     (-map (-compose #'symbol-name #'lsp--client-server-id) it)
     (format "%s" it)
     (propertize it 'face 'success)
-    (format "Installing following servers: %s" it)
+    (format " Installing following servers: %s" it)
```
ericdallo, line 7425: Any reason why adding this empty space?

kiennq: This is in mode line, and without the space, it's mangled with previous term.

PR #1856, opened 2020-06-29
lsp--capf-guess-prefix: check for empty char-before

### LINKED ISSUE #1852: Regression in lsp-capf after 70ea28f56eeff62f229aca6b469b60c1fd94307d
After 70ea28f56eeff62f229aca6b469b60c1fd94307d completion does not work correctly, at least in elm files. More specifically, company does not show completions, and throws an error.

When completing with "mai" as part of "main", I get the following error:

    Company: backend company-capf error "Wrong type argument: characterp, nil" with args (candidates mai)

The simplest way to reproduce this is:
`git clone https://github.com/rtfeldman/elm-spa-example.git`

Then if `elm-language-server` isn't installed:
`npm i -g @elm-tooling/elm-language-server`

Then open any file and start typing.

With `toggle-debug-on-error`, `lsp--capf-guess-prefix` shows up as the culprit.

Also, when reverting to the commit before the mentioned commit, everything works again.

It works fine in Csharp though.

kiennq:
Fix #1852 
I forgot to check for empty char before when `start == 1`


yyoncho:
We really should start adding tests in the format:

```
Input file + position + completions => 

1. Check prefix 
2. Check list - verify sorting, presence of elements and so on. 
```

PR #1859, opened 2020-06-30
completion: support resolveAdditionalTextEditsSupport capability

### LINKED ISSUE #1713: additionalTextEdits will lost if lsp-eslint and typescript-language-server enabled on same time.
**Describe the bug**
When editing typescript, If eslint enabled (lsp-eslint-server-command will set), lsp-mode ignoring `additionalTextEdits` from `lsp--resolve-completion` responses.

**To Reproduce**
`Emacs ver 28.0.50`

## Open emacs with following command.

```shell
/Applications/Emacs.app/Contents/MacOS/Emacs -Q -l /tmp/lsp-start-plain.el -l ~/.emacs.d/elpa/company-20200510.1614/company.el -l ~/.emacs.d/elpa/company-lsp-20190612.1553/company-lsp.el -l ~/.emacs.d/elpa/typescript-mode-20200430.1232/typescript-mode.el
```

## typescript project looks like that.

```
root
 - tsconfig.json
 - src
     |
     +- a.ts
     |
     +- b.ts
```

## Contents of tsconfig.json
```json
{
  "compilerOptions": {
    "target": "ESNext",
    "emitDecoratorMetadata": true,
    "experimentalDecorators": true,
    "allowSyntheticDefaultImports": true,
    "removeComments": false,
    "moduleResolution": "node",
    "sourceMap": true,
    "jsx": "preserve",
    "module": "esnext",
    "baseUrl": "./",
    "importHelpers": false,
    "sourceMap": true,
    "strict": true,
    "lib": ["dom", "es5", "ES2015", "es2016", "es2017", "es2018", "es2019"],
    "esModuleInterop": true
  },
  "exclude": ["node_modules", "src/**/flycheck_*"]
}
```

## Contents of a.ts
```typescript
export const testFunction = () => {};
```

## Contents of b.ts
```typescript
// In this file, I will completion with company.
testFunction
```

**Expected behavior**
If lsp-eslint disabled, `lsp--resolve-completion` `testFunction` will return `additionalTextEdits` to auto import `testFunction` from a.ts.
But if enabled, `lsp--resolve-completion` lost that.

`typescript-language-server` (ts-ls)
`eslintServer.js`

**OS**
macOS 10.14.6

**Error callstack**

## Elisp backtraces

```
* lsp--resolve-completion(#<hash-table equal 5/5 0x1fe2a1e5c791>)
  (let ((resolved-item (lsp--resolve-completion item)) (len (length candidate))) (put-text-property 0 len 'lsp-completion-item resolved-item candidate) (put-text-property 0 len 'company-lsp-resolved t candidate))
  (progn (let ((resolved-item (lsp--resolve-completion item)) (len (length candidate))) (put-text-property 0 len 'lsp-completion-item resolved-item candidate) (put-text-property 0 len 'company-lsp-resolved t candidate)))
  (if (seq-some #'(lambda (prop) (null (gethash prop item))) props) (progn (let ((resolved-item (lsp--resolve-completion item)) (len (length candidate))) (put-text-property 0 len 'lsp-completion-item resolved-item candidate) (put-text-property 0 len 'company-lsp-resolved t candidate))))
  (let ((item (company-lsp--candidate-item candidate))) (if (seq-some #'(lambda (prop) (null (gethash prop item))) props) (progn (let ((resolved-item (lsp--resolve-completion item)) (len (length candidate))) (put-text-property 0 len 'lsp-completion-item resolved-item candidate) (put-text-property 0 len 'company-lsp-resolved t candidate)))))
  (if (plist-get (text-properties-at 0 candidate) 'company-lsp-resolved) nil (let ((item (company-lsp--candidate-item candidate))) (if (seq-some #'(lambda (prop) (null (gethash prop item))) props) (progn (let ((resolved-item (lsp--resolve-completion item)) (len (length candidate))) (put-text-property 0 len 'lsp-completion-item resolved-item candidate) (put-text-property 0 len 'company-lsp-resolved t candidate))))))
  company-lsp--resolve-candidate(#("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))) "insertText" "textEdit" "additionalTextEdits")
  (let* ((resolved-candidate (company-lsp--resolve-candidate candidate "insertText" "textEdit" "additionalTextEdits")) (item (company-lsp--candidate-item resolved-candidate)) (prefix (company-lsp--candidate-prefix candidate)) (label (gethash "label" item)) (start (- (point) (length label))) (insert-text (gethash "insertText" item)) (insert-text-format (gethash "insertTextFormat" item)) (text-edit (gethash "textEdit" item)) (additional-text-edits (gethash "additionalTextEdits" item))) (cond (text-edit (setq insert-text (gethash "newText" text-edit)) (delete-region (- (point) (length candidate)) (point)) (insert prefix) (let* ((range (gethash "range" text-edit)) (new-text-length (length insert-text))) (setq start (lsp--position-to-point (gethash "start" range))) (lsp--apply-text-edit text-edit) (goto-char (+ start new-text-length)))) ((and insert-text (not (eq insert-text-format 2))) (progn (or (string-equal (buffer-substring-no-properties start (point)) label) (cl--assertion-failed '(string-equal ... label))) nil) (goto-char start) (delete-char (length label)) (insert insert-text))) (message "%S" resolved-candidate) (sit-for 240) (let ((start-marker (set-marker (make-marker) start))) (if (and additional-text-edits company-lsp-enable-additional-text-edit) (progn (lsp--apply-text-edits additional-text-edits))) (if (and company-lsp-enable-snippet (fboundp 'yas-expand-snippet)) (progn (if (and insert-text (eq insert-text-format 2)) (yas-expand-snippet (company-lsp--to-yasnippet-snippet insert-text) (marker-position start-marker) (point)) (let ((fallback-snippet ...)) (if fallback-snippet (progn ...)))))) (set-marker start-marker nil)) (if (and company-lsp-enable-recompletion (company-lsp--looking-back-trigger-characters-p)) (progn (setq this-command 'self-insert-command))))
  company-lsp--post-completion(#("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))))
  (cond ((eql command 'interactive) (company-begin-backend #'company-lsp)) ((eql command 'prefix) (and (and (boundp 'lsp-mode) lsp-mode) (lsp--capability "completionProvider") (or (let (needle) (let ((list ...) (continue t) (it-index 0)) (while (and list continue) (let ... ...) (setq it-index ...) (setq list ...))) needle) (not (company-in-string-or-comment))) (or (company-lsp--completion-prefix) 'stop))) ((eql command 'candidates) (or (company-lsp--cache-item-candidates (company-lsp--cache-get arg)) (and company-lsp-async (cons :async #'(lambda (callback) (company-lsp--candidates-async arg callback)))) (company-lsp--candidates-sync arg))) ((eql command 'sorted) t) ((eql command 'no-cache) (not (eq company-lsp-cache-candidates t))) ((eql command 'annotation) (lsp--annotate arg)) ((eql command 'quickhelp-string) (company-lsp--documentation arg)) ((eql command 'doc-buffer) (company-doc-buffer (company-lsp--documentation arg))) ((eql command 'match) (cdr (company-lsp--compute-flex-match arg))) ((eql command 'post-completion) (company-lsp--post-completion arg)))
  company-lsp(post-completion #("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))))
  apply(company-lsp (post-completion #("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t))))))
  (if (functionp company-backend) (apply company-backend args) (apply #'company--multi-backend-adapter company-backend args))
  (condition-case err (if (functionp company-backend) (apply company-backend args) (apply #'company--multi-backend-adapter company-backend args)) ((debug user-error) (user-error "Company: backend %s user-error: %s" company-backend (error-message-string err))) ((debug error) (error "Company: backend %s error \"%s\" with args %s" company-backend (error-message-string err) args)))
  company-call-backend-raw(post-completion #("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))))
  apply(company-call-backend-raw (post-completion #("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t))))))
  (let ((value (apply fun args))) (if (not (eq (car-safe value) :async)) value (let ((res 'trash) (start (time-to-seconds))) (funcall (cdr value) #'(lambda (result) (setq res result))) (while (eq res 'trash) (if (> (- (time-to-seconds) start) company-async-timeout) (error "Company: backend %s async timeout with args %s" backend args) (sleep-for company-async-wait))) res)))
  company--force-sync(company-call-backend-raw (post-completion #("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t))))) company-lsp)
  company-call-backend(post-completion #("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))))
  (let ((company-backend backend)) (run-hook-with-args 'company-completion-finished-hook result) (company-call-backend 'post-completion result))
  (if (stringp result) (let ((company-backend backend)) (run-hook-with-args 'company-completion-finished-hook result) (company-call-backend 'post-completion result)) (run-hook-with-args 'company-completion-cancelled-hook result))
  (progn (if (stringp result) (let ((company-backend backend)) (run-hook-with-args 'company-completion-finished-hook result) (company-call-backend 'post-completion result)) (run-hook-with-args 'company-completion-cancelled-hook result)) (run-hook-with-args 'company-after-completion-hook result))
  (if prefix (progn (if (stringp result) (let ((company-backend backend)) (run-hook-with-args 'company-completion-finished-hook result) (company-call-backend 'post-completion result)) (run-hook-with-args 'company-completion-cancelled-hook result)) (run-hook-with-args 'company-after-completion-hook result)))
  (let ((prefix company-prefix) (backend company-backend)) (setq company-backend nil company-prefix nil company-candidates nil company-candidates-length nil company-candidates-cache nil company-candidates-predicate nil company-common nil company-selection 0 company-selection-changed nil company--manual-action nil company--manual-prefix nil company--point-max nil company-point nil) (if company-timer (progn (cancel-timer company-timer))) (company-echo-cancel t) (company-search-mode 0) (company-call-frontends 'hide) (company-enable-overriding-keymap nil) (if prefix (progn (if (stringp result) (let ((company-backend backend)) (run-hook-with-args 'company-completion-finished-hook result) (company-call-backend 'post-completion result)) (run-hook-with-args 'company-completion-cancelled-hook result)) (run-hook-with-args 'company-after-completion-hook result))))
  company-cancel(#("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))))
  company-finish(#("testFunction" 0 12 (lsp-completion-item #<hash-table equal 5/5 0x1fe2a1e5c791> lsp-completion-prefix #("testFunctio" 0 11 (fontified t)))))
  (let ((result (nth company-selection company-candidates))) (company-finish result))
  (progn (let ((result (nth company-selection company-candidates))) (company-finish result)))
  (if (company-manual-begin) (progn (let ((result (nth company-selection company-candidates))) (company-finish result))))
  company-complete-selection()
  (if (and (not (cdr company-candidates)) (equal company-common (car company-candidates))) (company-complete-selection) (company--insert-candidate company-common))
  (progn (if (and (not (cdr company-candidates)) (equal company-common (car company-candidates))) (company-complete-selection) (company--insert-candidate company-common)))
  (if (company-manual-begin) (progn (if (and (not (cdr company-candidates)) (equal company-common (car company-candidates))) (company-complete-selection) (company--insert-candidate company-common))))
  company-complete-common()
  funcall-interactively(company-complete-common)
  call-interactively(company-complete-common nil nil)
  command-execute(company-complete-common)
```

## lsp-log

```
[Trace - 03:29:36 PM] Sending request 'completionItem/resolve - (327)'.
Params: {
  "data": {
    "entryNames": [
      {
        "source": "/tmp/t/src/a",
        "name": "testFunction"
      }
    ],
    "offset": 6,
    "line": 8,
    "file": "/tmp/t/src/b.ts"
  },
  "commitCharacters": [
    ".",
    ",",
    "("
  ],
  "sortText": "￿5",
  "kind": 21,
  "label": "testFunction"
}


[Trace - 03:29:36 PM] Received response 'completionItem/resolve - (327)' in 18ms.
Result: {
  "additionalTextEdits": [
    {
      "newText": "\nimport { testFunction } from \"./a\";\n",
      "range": {
        "end": {
          "character": 0,
          "line": 6
        },
        "start": {
          "character": 0,
          "line": 6
        }
      }
    }
  ],
  "documentation": {
    "value": "\n\n*@fileOverview*  \n\n*@name* — a.ts  \n\n*@author* — Taketoshi Aono  \n\n*@license*",
    "kind": "markdown"
  },
  "detail": "Auto import from './a'\nconst testFunction: () => void",
  "data": {
    "entryNames": [
      {
        "source": "/tmp/t/src/a",
        "name": "testFunction"
      }
    ],
    "offset": 6,
    "line": 8,
    "file": "/tmp/t/src/b.ts"
  },
  "commitCharacters": [
    ".",
    ",",
    "("
  ],
  "sortText": "￿5",
  "kind": 21,
  "label": "testFunction"
}
```

**emacs-lisp**

```lisp
(setq lsp-log-io t)
(setq lsp-eslint-server-command 
      '("node" 
        <eslintServer.js path>
        "--stdio"))
(setq lsp-document-sync-method 2)
(setq lsp-response-timeout 5)
(setq read-process-output-max (* (* 1024 1024) 3))
```

kiennq:
Support `resolveAdditionalTextEditsSupport`
Issue: https://github.com/emacs-lsp/lsp-java/pull/242

Also resolve #1713


yyoncho:
I prefer having this disabled by default or have a client flag. The server should provide a capability for that otherwise we will do one sync call in one of the most critical parts of the application which in a lot of cases will be just NOP.

kiennq:
I'm thinking of change the  sync call to async, similar to what javascript (VSCode) is doing. The additional texts can be applied at later time.
It seems that in VsCode, there's no problem with that approach so far, do you think it's ok?

yyoncho:
I am not quite sure if  the vscode call is performed after you press enter. I think that they start the resolve process once the item is displayed(or selected?). IMO we should introduce this resolve call without affecting the clients that do not populate text edits as part.

kiennq:
> I am not quite sure if the vscode call is performed after you press enter. I think that they start the resolve process once the item is displayed(or selected?)

We've also doing that for `company-doc-buffer`, however, when we select and commit the first selected item quickly, it should be quite the same as doing async resolving in committing selected item.



> IMO we should introduce this resolve call without affecting the clients that do not populate text edits as part it.

I agree, since we don't have server capability yet, just use async resolving would also work too.
Of course another flag on client can be defined so we can have this feature support by certain clients only. WDYT?

yyoncho:
> Of course another flag on client can be defined so we can have this feature support by certain clients only. WDYT?

This is one of the options I suggested initially. This will most likely solve https://github.com/emacs-lsp/lsp-mode/issues/1713

yyoncho:
> > Of course another flag on client can be defined so we can have this feature support by certain clients only. WDYT?
> 
> This is one of the options I suggested initially. This will most likely solve #1713

It is ugly but better than unconditional sync call. I think that we should also have defcustom for the client capability as well. Some might prefer slower completion over sync call while inserting the record.

rrudakov:
Is there any chances it will be merged? Completion in lsp-java is incredibly slow at the moment :(

yyoncho:
@kiennq this: https://github.com/microsoft/language-server-protocol/issues/1003#issuecomment-653499069

kiennq:
@yyoncho It seems that there will be no server flag for this.
While we can make a `lsp-mode` flag for each client, it will not make scaling/maintaining easier.
How about using async request to apply additional text edits?

yyoncho:
> How about using async request to apply additional text edits?

I think that we should do that in the way described in microsoft/language-server-protocol#1003 - do the request and then apply the text edit even if the buffer is changed when the edits are compatible with the change. (e. g. you have edited line 10 but text edit is for line 1). IMO this is complex and unreliable but this is what MS is giving us...

kiennq:
> I think that we should do that in the way described in [microsoft/language-server-protocol#1003](https://github.com/microsoft/language-server-protocol/issues/1003) - do the request and then apply the text edit even if the buffer is changed when the edits are compatible with the change. (e. g. you have edited line 10 but text edit is for line 1). IMO this is complex and unreliable but this is what MS is giving us...

@yyoncho 
Well even with normal additional edits applying, we just applying the text edits without checking it compatible with the main text edit anyway.
So I think we can just do normally, but without blocking.

I've also push a new commit that merge the `completionItem/resolve` results and apply additional text edits asynchronously, can you help review?

yyoncho:
> we just applying the text edits without checking it compatible with the main text edit anyway.

Normally we either check the version or we are sync so we are sure that the text edits are for the current state of the file. 

I will accept this PR since from a practical standpoint of view it will  be hard to hit a bug in this case.

We may also store the source workspace here https://github.com/yyoncho/lsp-mode/blob/master/lsp-mode.el#L1332 and then retrieve it when resolving.

kiennq:
@rrudakov Can you check if this PR resolve your issue?

kiennq:
> > we just applying the text edits without checking it compatible with the main text edit anyway.
> 
> Normally we either check the version or we are sync so we are sure that the text edits are for the correct state of the file.
> 
> I will accept this PR since from a practical standpoint of view it will be hard to hit a bug in this case.
> 
> We may also store the source workspace here https://github.com/yyoncho/lsp-mode/blob/master/lsp-mode.el#L1332 and then retrieve it when resolving.

Oh, indeed, I haven't consider that.
Anyway, I've added a simple check to make sure the additional text edit is before and disconnected from the main edit.
That's pretty much the only time when we can apply additional text edits anyway.

lsp-mode.el
```diff
@@ -4984,27 +5038,37 @@ Others: TRIGGER-CHARS"
...
            (let (yas-indent-line)
              (yas-expand-snippet
               (lsp--to-yasnippet-snippet (buffer-substring start-point (point)))
               start-point
               (point))))
 
-         (when (and lsp-completion-enable-additional-text-edit additional-text-edits?)
-           (lsp--apply-text-edits additional-text-edits?))
+         (when lsp-completion-enable-additional-text-edit
+           (if (get-text-property 0 'lsp-completion-resolved candidate)
```
yyoncho, line 5064: don't we need to check if the edits are present and resolve only if they are nil?

lsp-mode.el
```diff
@@ -4230,6 +4273,17 @@ The method uses `replace-buffer-contents'."
             (with-no-warnings (undo-amalgamate-change-group change-group)))
           (progress-reporter-done reporter))))))
 
+(defun lsp--apply-additional-text-edits (edits &optional main-edit-region)
```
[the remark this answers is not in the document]
kiennq: Unless there's problem, I expect the server resolve for a completion item should be fairly fast, so user has little no chance of adding another edit that invalidate the additional text edit.
To properly track what you mentioned, we will need all the edits between when user select completion item and when the result is returned.
Did `onTypeFormating` already tracking the same thing? I can probably copy the copy code from that.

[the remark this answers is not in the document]
yyoncho: > To properly track what you mentioned, we will need all the edits between when user select completion item and when the result is returned.

We need only the first point that has been changed. The additionalTextEdits are valid as long as there is no change before the first user edit. 

> Did onTypeFormating already tracking the same thing? I can probably copy the copy code from that.

No, it just rejects the edits if the user has typed, which is not the best option.

As I mentioned here or in the bug report IMO it is acceptable to do the request with 'tick as we do in on type formatting but still, it won't be the best effort solution.

lsp-mode.el
kiennq: Fixed. I wonder if there's case of completion item need to be resolved even if additional edits are not null

lsp-mode.el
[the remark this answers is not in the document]
kiennq: I've implement another change that track the changes from when we complete completion and until the additional text edits are inserted. @yyoncho Can you help reviewing?

lsp-mode.el
yyoncho: Yes. When documentation is requested. You may refer to company-lsp code - it calls resolve only if particular field is missing.

lsp-mode.el
```diff
@@ -4511,16 +4580,34 @@ Added to `after-change-functions'."
...
-  (lsp--idle-reschedule (current-buffer)))
+                              buffer))
+
+  (when lsp--on-change-fns-timer
+    (cancel-timer lsp--on-change-fns-timer)
+    (setq lsp--changes
+          (plist-put lsp--changes
+                     buffer
+                     (nconc (plist-get lsp--changes buffer)
+                            `((,start ,end ,pre-change-len))))))
```
yyoncho, line 4604: don't we need to keep only the first start?

lsp-mode.el
```diff
@@ -4999,27 +5086,49 @@ Others: TRIGGER-CHARS"
...
-          (lsp--apply-text-edits additional-text-edits?))
+        (when lsp-completion-enable-additional-text-edit
+          (if (or (get-text-property 0 'lsp-completion-resolved candidate)
+                  additional-text-edits?)
+              (lsp--apply-additional-text-edits additional-text-edits? main-edit-region)
+            (add-hook 'lsp-on-change-functions change-tracker)
+            (lsp--resolve-completion-async
+             item
+             (lambda (resolved-item)
+               (remove-hook 'lsp-on-change-functions change-tracker)
```
yyoncho, line 5128: can we use directly before-change-functions? Seems like it should be buffer-local subscription. Can you generalize it so we can use it for on type formatting as well?

lsp-mode.el
```diff
@@ -4245,6 +4289,17 @@ it has to calculate identation based on SRC block position."
             (with-no-warnings (undo-amalgamate-change-group change-group)))
           (progress-reporter-done reporter))))))
 
+(defun lsp--apply-additional-text-edits (edits &optional main-edit-region)
+  "Apply the additional EDITS described in TextEdit[] object.
+This will also verify if the additional text edits are applicable
+regarding to the MAIN-EDIT-REGION."
+  (unless (and main-edit-region
```
yyoncho, line 4296: main-edit-region seems to be redundant - additional text edits are valid as long as there is no edit by the user.

lsp-mode.el
```diff
@@ -4482,6 +4546,11 @@ Added to `after-change-functions'."
   :group 'lsp-mode)
 
 
+(defvar lsp--on-change-fns-timer nil)
```
yyoncho, line 4549: What do we need that for?

lsp-mode.el
kiennq: Yes, but since I add new `lsp-on-change-functions`, it might be better to provide all information to its subscribers.
Thus all information will be passed.

lsp-mode.el
kiennq: This is a timer to debounce `lsp-on-change-functions`. So we will only notify those on change information when really in idle and after idle delay.

lsp-mode.el
kiennq: We might be able to use `before-change-functions` directly, but it's not debounced and not obeying to `lsp-inhibit-hook`. I don't think we need to apply additional text edits immediately, it already applied asynchronously, so wait a little bit more to a debounce check-point would also be fine.
The new hook `lsp-on-change-functions` can be used in type formatting too, we will need to make change tracker a separate function, I will do that in another PR.

lsp-mode.el
kiennq: The `main-edit-region` here is actually the edit region made by user. So as long as user doesn't edit overlappingly with additional texts, they will be applied

lsp-mode.el
yyoncho: We should not debounce for this reason - we should try to apply the additional text edits as soon as possible. The longer delay the be bigger the chance to fail is. Also, it will be hard to reason about why the text edits are not applied immediately.

lsp-mode.el
yyoncho: We don't need to debounce since the only thing that function does is to store information need for checking if a particular set of edits is invalidated, no?

lsp-mode.el
yyoncho: > it might be better to provide all information to its subscribers.

Why? Can we implement the simplest possible solution and extend it when a subscriber that needs more information is needed? More code = more trouble in the future.

yyoncho: Also, this will break if there are 2 calls that need the first edited point. The min position must be stored in the capture, not in buffer-local.

lsp-mode.el
yyoncho: > overlappingly

We are not looking for overlap but if there was a change after the first start point in the edits. Even if there is no overlap the edit will be invalid if there was change before it.

kiennq: Yes, the function is checking exactly for that.

lsp-mode.el
kiennq: It actually less code, we just store whatever `after-change-functions` gave us, without further modification.
There's no min position here yet, it's just a series of changes and the subscribers need to process more to get information they needed.
The main reason I'm not processing more than this is I don't know what exactly subscriber want (generally), they may want  only first edit point, or a range, put just raw data here provide the max flexibility.

lsp-mode.el
kiennq: If we eagerly apply the additional text edits, the document between client and server will be mismatched. That means the range returned by completion may completely un-applicable.
Thus I debounce this and apply it later on. At least make sure that while we're in middle of doing completion, the selected item can still be applied to correct range.

lsp-mode.el
yyoncho: We have one subscriber and we know what it needs. Either way, it should not interfere with the rest of the "subscribers" since they will have a scope. 

``` emacs-lisp
(let ((first-edited (point-max)))
  (add-hook 'before-change-functions 
            (lambda (start &rest _)
              (setq first-edited (min start first-edited)))
            nil t)

  (lsp-request-async 
   ...
   (lambda (edits)
     ;; check if edits is compatible with first-edited point
     )))
```

kiennq: Well, I've added a new hook `lsp-on-change-functions` and want to provide maximum flexibility for its potential subscribers.
It looks like that's an over engineering attempt, will revert that.

kiennq: @yyoncho

lsp-mode.el
```diff
@@ -5011,15 +5091,24 @@ Others: TRIGGER-CHARS"
...
+          (if (or (get-text-property 0 'lsp-completion-resolved candidate)
+                  additional-text-edits?)
+              (lsp--apply-text-edits additional-text-edits?)
+            (-let [(callback cancel-callback) (lsp--create-apply-text-edits-handlers)]
+              (lsp--resolve-completion-async
+               item
+               (lambda (resolved-item)
+                 (funcall callback
+                  (lsp:completion-item-additional-text-edits? resolved-item)))
+               cancel-callback))))
```
yyoncho, line 5111: maybe call cleanup on error handler as well?

lsp-mode.el
```diff
@@ -5138,15 +5227,31 @@ Others: TRIGGER-CHARS"
...
-                          (lsp--capability :completionProvider))
-                     (lsp-request "completionItem/resolve" item)))
-                (error)))
+  (or (ignore-errors
+        (when (lsp:completion-options-resolve-provider?
+               (lsp--capability :completionProvider))
+          (lsp-request "completionItem/resolve" item)))
       item))
 
+(defun lsp--resolve-completion-async (item callback &optional cleanup-fn)
```
yyoncho, line 5236: nit: is this really optional?

lsp-mode.el
```diff
@@ -4245,6 +4307,26 @@ it has to calculate identation based on SRC block position."
...
+Only works when mode is 'tick or 'alive."
+  (let* (first-edited
+         (func (lambda (start &rest _)
+                 (setq first-edited (if first-edited
+                                        (min start first-edited)
+                                      start)))))
+    (add-hook 'before-change-functions func nil t)
+    (list
+     (lambda (edits)
+       (unless (and first-edited
```
yyoncho, line 4321: can you add some kind of message that the additional edits will be skipped? Not 100% sure if it should be info/warn or lsp-log

leungbk:
Nit: you should be able to replace all the `(apply #'nconc` with `(cl-mapcan` instead.

lsp-mode.el
kiennq: Yes, the `cleanup-fn` can be optional, for example to resolve doc asynchronously (just assign properties to `item`)

lsp-mode.el
kiennq: `lsp--resolve-completion-async` will also do cleanup on error

lsp-mode.el
kiennq: added as warn

kiennq:
> Nit: you should be able to replace all the `(apply #'nconc` with `(cl-mapcan` instead.

Thanks, `cl-mapcan` will apply a function to list and `nconc` the result, since we don't need the function, just using `nconc` directly is simpler.

lsp-mode.el
```diff
@@ -1316,13 +1316,41 @@ depending on it."
...
                         #'lsp:completion-list-is-incomplete
                         results)
-         :items (apply 'append (--map (append (if (lsp-completion-list? it)
-                                                  (lsp:completion-list-items it)
-                                                it)
-                                              nil)
-                                      results))))
-       (_ (apply 'append (seq-map (lambda (it) (if (seqp it) it (list it)))
-                                  results)))))))
+         :items (mapcan (lambda (it) (append (if (lsp-completion-list? it)
```
leungbk, line 1319: I don't think `mapcan` is available on older Emacsen, so it should be `cl-mapcan` to accommodate.

yyoncho:
I am getting the following error when closing a buffer:

```
Debugger entered--Lisp error: (void-function closure)
  closure(((func closure ((first-edited . 134) cl-struct-lsp--log-entry-tags cl-struct-lsp-session-tags cl-struct-lsp--workspace-tags cl-struct-lsp--registered-capability-tags lsp-mode-menu cl-struct-lsp--folding-range-tags cl-struct-lsp-watch-tags cl-struct-lsp--client-tags lsp--log-lines company-backends t) (start &rest _) (setq first-edited (if first-edited (min start first-edited) start))) (first-edited . 134) cl-struct-lsp--log-entry-tags cl-struct-lsp-session-tags cl-struct-lsp--workspace-tags cl-struct-lsp--registered-capability-tags lsp-mode-menu cl-struct-lsp--folding-range-tags cl-struct-lsp-watch-tags cl-struct-lsp--client-tags lsp--log-lines company-backends t) nil (remove-hook 'before-change-functions func t))
  apply((closure ((func closure ((first-edited . 134) cl-struct-lsp--log-entry-tags cl-struct-lsp-session-tags cl-struct-lsp--workspace-tags cl-struct-lsp--registered-capability-tags lsp-mode-menu cl-struct-lsp--folding-range-tags cl-struct-lsp-watch-tags cl-struct-lsp--client-tags lsp--log-lines company-backends t) (start &rest _) (setq first-edited (if first-edited (min start first-edited) start))) (first-edited . 134) cl-struct-lsp--log-entry-tags cl-struct-lsp-session-tags cl-struct-lsp--workspace-tags cl-struct-lsp--registered-capability-tags lsp-mode-menu cl-struct-lsp--folding-range-tags cl-struct-lsp-watch-tags cl-struct-lsp--client-tags lsp--log-lines company-backends t) nil (remove-hook 'before-change-functions func t)))
  (save-current-buffer (set-buffer buf) (apply cancel-callback))
  (closure (... ... ... ... ... ... ... ... ... ... ... ... ... ... ...) nil (save-current-buffer ... ...))()
  funcall((closure ... nil ...))
```

rrudakov:
I've been testing it for a few hours, apart of the problem mentioned by @yyoncho completion works much faster now in lsp-java.

kiennq:
I found the problem now. Pushed a new commit.

rrudakov:
@kiennq confirm, the problem is gone.

Issue #1835, opened 2020-06-22
Warn about lsp related packages compiled when enabling plists


ericdallo:
To enable `lsp-use-plists` we can first add some kind of warning to the user when loading lsp to re-compile all lsp-related packages to avoid packages checking hts and other packages checking for plists.

PR #1861, opened 2020-07-01
Add check for inconsistency when use-plists

ericdallo:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/1835

This is an attempt to warn the user when `lsp-use-plists` is enabled but some function/variable of plists feature is not bounded.

Feel free to suggest some better place to move the check and the warning message

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8094,8 +8094,7 @@ such."
 (defun lsp--any-inconsistency? ()
   "Check for any inconsistency on LSP package"
   (and lsp-use-plists
-       (or (not (fboundp 'lsp-make-range))
-           (not (boundp 'lsp/symbol-kind-field)))))
+       (equal (symbol-function 'lsp-merge) 'ht-merge)))
 
 (defun lsp-shutdown-workspace ()
   "Shutdown language server."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8091,6 +8091,12 @@ such."
       (lsp--warn "No LSP server for %s(check *lsp-log*)." major-mode)
       nil)))
 
+(defun lsp--any-inconsistency? ()
+  "Check for any inconsistency on LSP package"
+  (and lsp-use-plists
+       (or (not (fboundp 'lsp-make-range))
+           (not (boundp 'lsp/symbol-kind-field)))))
+
 (defun lsp-shutdown-workspace ()
   "Shutdown language server."
   (interactive)
@@ -8216,7 +8222,10 @@ You may find the installation instructions at https://emacs-lsp.github.io/lsp-mo
        ;; no matches
        ((-> #'lsp--matching-clients? lsp--filter-clients not)
         (lsp--error "There are no language servers supporting current mode %s registered with `lsp-mode'."
-                    major-mode))))))
+                    major-mode)))
+      (unless (lsp--any-inconsistency?)
+        (lsp--warn "It's seems that one or more LSP packages are outdated or it may need to re-compile it to avoid errors.
+Make sure to update all LSP packages and byte-compile it again.")))))
 
 (defun lsp--init-if-visible ()
   "Run `lsp' for the current buffer if the buffer is visible.
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3091,7 +3091,9 @@ active `major-mode', or for all major modes when ALL-MODES is t."
   (folder->servers (make-hash-table :test 'equal))
   ;; ‘metadata’ is a generic storage for workspace specific data. It is
   ;; accessed via `lsp-workspace-set-metadata' and `lsp-workspace-set-metadata'
-  (metadata (make-hash-table :test 'equal)))
+  (metadata (make-hash-table :test 'equal))
+  ;; If we should warn user about a possible inconsistency between LSP packages.
+  (warn-possible-inconsistency? lsp-use-plists))
 
 (defun lsp-workspace-status (status-string &optional workspace)
   "Set current workspace status to STATUS-STRING.
@@ -7348,7 +7350,14 @@ SESSION is the active session."
 
          (with-lsp-workspace workspace
            (run-hooks 'lsp-after-initialize-hook))
-         (lsp--info "%s initialized successfully" (lsp--workspace-print workspace)))
+         (lsp--info "%s initialized successfully" (lsp--workspace-print workspace))
+
+         (condition-case ()
+             (when (lsp-session-warn-possible-inconsistency? session)
+               (lsp--warn "LSP packages may need to re-compile it to avoid errors.
+Make sure to update all LSP packages and byte-compile it again.")
+               (setf (lsp-session-warn-possible-inconsistency? session) nil))
+           (error t)))
        :mode 'detached))
     workspace))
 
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -8091,6 +8091,12 @@ such."
       (lsp--warn "No LSP server for %s(check *lsp-log*)." major-mode)
       nil)))
 
+(defun lsp--any-inconsistency? ()
+  "Check for any inconsistency on LSP package"
+  (and lsp-use-plists
+       (or (not (fboundp 'lsp-make-range))
```
kiennq, line 8097: Is this check correct? It just check for function that will be defined when `lsp-mode` is required (with `lsp-protocol`.)
I think the check should be something along
```
(xor lsp-use-plists
     (equal (symbol-function 'lsp-merge) 'append))
```

ericdallo: It checks if `lsp-make-*` that is a function that only works when lsp-mode is re-compiled correctly, right?
I don't know for sure the difference between using `(fboundp 'lsp-make-range)` and `(symbol-function 'lsp-make-range)` for checking if non-nil, I think both works for what we need here, am I wrong?

kiennq: > It checks if `lsp-make-*` that is a function that only works when lsp-mode is re-compiled correctly, right?

Not really, the `lsp-make-*` is defined regardless of state of `lsp-use-plists`.
So for example, you have been in recent `lsp-mode` with `lsp-use-plists` is `nil`, then there's update where `lsp-use-plists` become `t`, `(fboundp 'lsp-make*)` will still return `t` even if you don't recompile `lsp-mode` correctly.

On other hand, `lsp-merge`, which is aliased to `ht-merge` or `append` depends on `lsp-use-plists` will just return `ht-merge` if you don't recompile properly when switch to `lsp-use-plists`

ericdallo: I see, It make sense, thanks for the explanation, I'll test it locally then change it!

ericdallo: Fixed it!

lsp-mode.el
```diff
@@ -8216,7 +8221,10 @@ You may find the installation instructions at https://emacs-lsp.github.io/lsp-mo
        ;; no matches
        ((-> #'lsp--matching-clients? lsp--filter-clients not)
         (lsp--error "There are no language servers supporting current mode %s registered with `lsp-mode'."
-                    major-mode))))))
+                    major-mode)))
+      (unless (lsp--any-inconsistency?)
+        (lsp--warn "It's seems that one or more LSP packages are outdated or it may need to re-compile it to avoid errors.
```
ericdallo, line 8226: @yyoncho I'm not sure if this is the correct place to add the warning, could you please confirm that?

lsp-mode.el
```diff
@@ -8091,6 +8091,11 @@ such."
       (lsp--warn "No LSP server for %s(check *lsp-log*)." major-mode)
       nil)))
 
+(defun lsp--any-inconsistency? ()
+  "Check for any inconsistency on LSP package"
+  (and lsp-use-plists
+       (equal (symbol-function 'lsp-merge) 'ht-merge)))
```
yyoncho, line 8097: these are not the inconsistencies that we are going to hit. The issues will be when a package e. g. lsp-dart has version which has been compiled with lsp-plists off or if you update all at the same time but package.el decides first to download lsp-dart, compiles it against old lsp-mode and then updates lsp-mode. Unfortunately there is no easy way to fix that.

ericdallo: Oh, I see the problem...
Do you see a way to check that or should we close this PR?

yyoncho: one way to do that is to define in each package `(lsp-interface PACKAGE-check ...)` and then check if lsp-make-PACKAGE-check will return plist. IMO it is too much effort. It will be easier and less convenient for the users if we add a flag and ask users to delete all lsp-mode related packages and reinstall them and then set that flag to nil to skip that request.

ericdallo: Yeah, I agree, but do you think the delete would be manual by the user or some that we should do if the flag is true?
And the set of the flag to true would be the user that would set it or we somehow?

yyoncho: The flag will be something like: `lsp-skip-protocol-plist-warning` . IMO it is not a good idea to try to implement the delete automatically since there are different package managers with different APIs.

ericdallo: Yeah, it was what I thought about it, but yet I think it's too bad for the UE, I see users complaining about having to set a variable manually because a package migration and users that never had to set anything manually needing to do that.
Maybe let the problem happen and have a well-documented page explaining about it and how to fix it could be better, WDYT?

ericdallo: Maybe we can also explain this in the issue related to this PR and tag more maintainers to see their opinions

yyoncho: > Maybe we can also explain this in the issue related to this PR and tag more maintainers to see their opinions

That won't work - we will get dozens of issue reports for months. We were in that situation with dash in the past. Dash had introduced new destructuring which was used in the project and that was reported in gitter/reddit/dap-mode repo/lsp-mode repo maybe 30 times. And lsp-mode was not that popular.

We may keep the setting in lsp-session so we can store it in lsp-session-file so once answered it won't be asked again.

ericdallo: I see... Ok, I think saving in the lsp-session could be better, but yet we will get issues from users that ignored that, but I think that there is nothing that we can do more about it.
I'll implement this on `lsp-session` so

ericdallo: @yyoncho done

lsp-mode.el
```diff
@@ -3091,7 +3091,9 @@ active `major-mode', or for all major modes when ALL-MODES is t."
   (folder->servers (make-hash-table :test 'equal))
   ;; ‘metadata’ is a generic storage for workspace specific data. It is
   ;; accessed via `lsp-workspace-set-metadata' and `lsp-workspace-set-metadata'
-  (metadata (make-hash-table :test 'equal)))
+  (metadata (make-hash-table :test 'equal))
+  ;; If we should warn user about a possible inconsistency between LSP packages.
+  (warn-possible-inconsistency? lsp-use-plists))
```
yyoncho, line 3096: Adding this key will break session deserialization if you have serialized the old one without this field(not sure if the issue was present in all emacs versions).

Also, take a look at lsp--persist-session  - it does not serialize all keys. But lsp-session-metata might contain stuff that is not serializable like markers. So I guess we should serialize only certain keys from the metadata.

ericdallo: @yyoncho Done, moved the new boolean to metadata and on `lsp--persist-session`, save only this new field of metadata.

ericdallo:
@yyoncho do you think this PR still worth?

yyoncho:
I think that the instructions on how to byterecompile should be more detailed.

ericdallo:
What do you suggest @yyoncho

yyoncho:
> What do you suggest @yyoncho ?

Here it is my proposal:

create a macro that will generate `lsp-PACKAGE-check` function. The function will use lsp-protocol macro in it. Then lsp-mode when starting for the first time will check if the package has that function and it will call it. The function will return what the package was compiled with.

ericdallo:
LGTM @yyoncho I think we should close this PR and implement it as you said

PR #1864, opened 2020-07-01
Add lsp-serenata notification handler and init options

### LINKED ISSUE #1847: Serenata logs excessive warnings
**Describe the bug**
When using lsp-serenata in a buffer, it opens a window where it outputs a lot of:

```
Warning (lsp-mode): Unknown notification: serenata/didProgressIndexing
```
I was told to tag @Sasanidas 

**To Reproduce**

Clone https://github.com/xendk/dais/ and open the `dais` file. It should start spamming.

**Expected behavior**

It not opening a window with warnings

lsp-serenata

**OS**
Linux

Sasanidas:
Add some custom init options, also Fixes #1847 with the default Serenata message.

### DIFF
--- a/lsp-serenata.el
+++ b/lsp-serenata.el
@@ -76,7 +76,7 @@ on the next project initialization."
     :phpVersion ,lsp-serenata-php-version
     :fileExtensions ,lsp-serenata-file-extensions))
 
-(lsp-interface (serenata:didProgressIndexing (:info) nil))
+(lsp-interface (serenata:didProgressIndexing (:sequenceOfIndexedItem :totalItemsToIndex :progressPercentage :folderUri :fileUri :info) nil ))
 
 (lsp-register-client
  (make-lsp-client
--- a/lsp-serenata.el
+++ b/lsp-serenata.el
@@ -24,6 +24,7 @@
 
 ;;; Code:
 
+(require 'lsp-protocol)
 (require 'lsp-mode)
 
 (defgroup lsp-serenata nil
@@ -75,6 +76,8 @@ on the next project initialization."
     :phpVersion ,lsp-serenata-php-version
     :fileExtensions ,lsp-serenata-file-extensions))
 
+(lsp-interface (serenata:didProgressIndexing (:info) nil))
+
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-tcp-connection 'lsp-serenata-server-start-fun)
@@ -82,7 +85,7 @@ on the next project initialization."
   :priority -2
   :notification-handlers (ht ("serenata/didProgressIndexing"
 			      (lambda (server data)
-				(message "%s" (ht-get data "info")))))
+				(message "%s" (lsp:serenata-did-progress-indexing-info data)) )))
   :initialization-options #'lsp-serenata-init-options
   :initialized-fn (lambda (workspace)
                     (with-lsp-workspace workspace
--- a/lsp-serenata.el
+++ b/lsp-serenata.el
@@ -85,7 +85,7 @@ on the next project initialization."
   :priority -2
   :notification-handlers (ht ("serenata/didProgressIndexing"
 			      (lambda (_server data)
-				(message "%s" (lsp:serenata-did-progress-indexing-info data)) )))
+				(lsp--info "%s" (lsp:serenata-did-progress-indexing-info data)) )))
   :initialization-options #'lsp-serenata-init-options
   :initialized-fn (lambda (workspace)
                     (with-lsp-workspace workspace
--- a/lsp-serenata.el
+++ b/lsp-serenata.el
@@ -63,7 +63,7 @@ files having them from the index if they are already present.
 Adding new ones will cause the files having them to be picked up
 on the next project initialization."
   :group 'lsp-serenata
-  :type 'lsp-string-array)
+  :type 'lsp-string-vector)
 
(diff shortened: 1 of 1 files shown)


lsp-serenata.el
```diff
@@ -39,21 +39,56 @@ It can be downloaded from https://gitlab.com/Serenata/Serenata/-/releases."
...
+
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-tcp-connection 'lsp-serenata-server-start-fun)
   :major-modes '(php-mode)
   :priority -2
-  :server-id 'serenata
+  :notification-handlers (ht ("serenata/didProgressIndexing"
+			      (lambda (server data)
+				(message "%s" (ht-get data "info")))))
```
ericdallo, line 85: We currently support ht and plists, so I think it's better to define an interface for this notification like [here](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-protocol.el#L261) and use `(lsp-my-interface-info data)` here.

Sasanidas: Sorry, I dont understand the correct way to do it, 

(lsp-interface (serenata:ProcessStartNotification nil (:info))

  :notification-handlers 'lsp:serenata-process-start-notification-info 

That code give me some error.

I think I'm doing something wrong.

ericdallo: Basically, we created a way to define some protocol for elisp of the LSP protocol and other server specific protocols.
If you have a Notification that returns an object of `ProcessStartNotification` that contains some fields like `info` as required and `range` as optional, for example, you can define an interface like:
```elisp
(lsp-interface (serenata:ProcessStartNotification (:info) (:range)))
```
In your case maybe: 
```elisp
(lsp-interface (serenata:ProcessStartNotification (:info) nil))
```

After that, when you require `lsp-protocol` in the file you can use `(lsp:serenata-process-start-notification-info data)`

yyoncho: and one more thing: the `ht` after notification handlers stays as it is. The only thing that is changed is the way you access the structures coming from the language server. You may take a look at the unit tests for example usages: https://github.com/emacs-lsp/lsp-mode/blob/master/test/lsp-protocol-test.el

Sasanidas: Ok ,I think I figure out how to solve the thing :+1: 

Thanks for the tips!

lsp-serenata.el
```diff
@@ -39,21 +40,58 @@ It can be downloaded from https://gitlab.com/Serenata/Serenata/-/releases."
...
+(lsp-interface (serenata:didProgressIndexing (:info) nil))
+
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-tcp-connection 'lsp-serenata-server-start-fun)
   :major-modes '(php-mode)
   :priority -2
-  :server-id 'serenata
+  :notification-handlers (ht ("serenata/didProgressIndexing"
+			      (lambda (server data)
```
yyoncho, line 87: nit: prefix server with _ to fix the CI.

lsp-serenata.el
```diff
@@ -39,21 +40,58 @@ It can be downloaded from https://gitlab.com/Serenata/Serenata/-/releases."
...
+indexing.  Usually you'll want to set this to at least include
+php, as it is the most common PHP extension.  phpt is not
+included by default as it is often used to contain test code that
+is not directly part of the code.  Note that for existing
+projects, removing extensions will not not automatically prune
+files having them from the index if they are already present.
+Adding new ones will cause the files having them to be picked up
+on the next project initialization."
+  :group 'lsp-serenata
+  :type 'array)
```
yyoncho, line 66: array -> lsp-string-array.

lsp-serenata.el
```diff
@@ -39,21 +40,58 @@ It can be downloaded from https://gitlab.com/Serenata/Serenata/-/releases."
...
   `(,lsp-serenata-server-path
     "-u" ,(number-to-string port)))
 
+(defun lsp-serenata-init-options ()
+  "Init options for lsp-serenata."
+  `(
+    :phpVersion ,lsp-serenata-php-version
+    :fileExtensions ,lsp-serenata-file-extensions))
+
+(lsp-interface (serenata:didProgressIndexing (:info) nil))
```
kiennq, line 79: The type for `serenata/didProgressIndexing` is different from what you put here
https://gitlab.com/Serenata/Serenata/-/blob/master/src/Indexing/DirectoryIndexRequestDemuxer.php

Could you consider to properly add all the type as defined in the source?
It will save other time to modify the interface again

Sasanidas: Sure, no problem

lsp-serenata.el
```diff
@@ -39,21 +40,58 @@ It can be downloaded from https://gitlab.com/Serenata/Serenata/-/releases."
...
+
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-tcp-connection 'lsp-serenata-server-start-fun)
   :major-modes '(php-mode)
   :priority -2
-  :server-id 'serenata
+  :notification-handlers (ht ("serenata/didProgressIndexing"
+			      (lambda (_server data)
+				(message "%s" (lsp:serenata-did-progress-indexing-info data)) )))
```
kiennq, line 88: Using `lsp--info` instead?
```suggestion
				(lsp--info "%s" (lsp:serenata-did-progress-indexing-info data)) )))
```

PR #1865, opened 2020-07-02
[blog] Add blog section

ericdallo:
Add blog section with the first post about 7.0 release

image


MaskRay:
For the website, should we use a separate repo?

ericdallo:
@MaskRay ATM the arch of lsp mode website fits a new section very well, but if we see that we need more improvements, like comments/new UX and other things I agree in separating to a new repo

PR #1878, opened 2020-07-04
Clean modeline code actions on shutdown

ericdallo:


lsp-mode.el
```diff
@@ -3947,6 +3951,7 @@ in that particular folder."
 
       (remove-hook 'xref-backend-functions #'lsp--xref-backend t)
       (setq-local global-mode-string (remove status global-mode-string))
+      (lsp--modeline-clean-code-actions)
```
kiennq, line 3954: Why this is needed?
The `lsp-modeline-code-actions-mode` is already call to be disabled in `lsp-unconfig-buffer`, which will remove the mode line phrase from `global-mode-string` already.

kiennq: Also, I don't think just clear mode line only is good, you should call into `(lsp-modeline-code-actions-mode -1)`, which will do properly cleanup for `lsp-mode-line-code-actions-mode`

ericdallo: sorry, you are right, for some reason when lsp-dart server crashed, I thought I saw the code actions on modeline, but I tested it again and I think this is already working and disabling the mode when lsp server shutdown.
Thanks, I'll close the PR

PR #1886, opened 2020-07-07
lsp--capf-filtering: dont reordering the first time completed completions come

### LINKED ISSUE #1883: completionItem sortText not respected by lsp--capf-filter-candidates
**Describe the bug**
A change from 6.3 (working) and 7.0 (breaking), it appears that `lsp--capf-filter-candidates` scores completion items, then sorts by those scores. If a CompletionItem `sortText` field was specified, it seems to be completely overridden. 

**To Reproduce**
From a message traffic perspective, it is helpful to log a request to verify the ordering and sortText keys of the inbound completions. The displayed completion options to produce a different order. I am using a non-public language server so I cannot share a direct example. I'm unfamiliar with existing language servers that may include sortText results. 

**Expected behavior**
I expected that the server's `sortText` key would be used for sort order.

If I comment out this code in `lsp--capf-filter-candidates`, I get the expected behavior.
```
(-sort (lambda (o1 o2)
         (> (get-text-property 0 'sort-score o1)
            (get-text-property 0 'sort-score o2))))
```

Non-public

**OS**
Mac OS X 10.15.5

kiennq:
Fix #1883 by not reordering the first time before caching completion items.
After that, we will use the client filtering score when filter down from cache.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -805,6 +805,12 @@ ignored."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-completion-no-cache nil
+  "Whether or not caching the returned completions from server."
+  :type 'boolean
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.0.1"))
+
 (defcustom lsp-server-trace nil
   "Request tracing on the server side.
 The actual trace output at each level depends on the language server in use.
@@ -4814,6 +4820,9 @@ Return `nil' when fails to guess prefix."
                            'lsp-completion-score score?))
              it)))
 
+(defvar lsp--capf-no-reordering nil
+  "Dont do client-side reordering completion items when set.")
+
 (cl-defun lsp--capf-filter-candidates (items
                                        &rest plist
                                        &key lsp-items
@@ -4824,7 +4833,7 @@ Also, additional data to attached to each candidate can be passed via PLIST."
   (lsp--while-no-input
    (->>
     (if items
-        (->>
+        (-->
          (let (queries fuz-queries)
            (-keep (lambda (cand)
                     (let* ((start-point (get-text-property 0 'lsp-completion-start-point cand))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4970,7 +4982,8 @@ Also, additional data to attached to each candidate can be passed via PLIST."
                                                           bounds-start)))
                                            it))))
                         (markers (list bounds-start (copy-marker (point) t)))
-                        (prefix (buffer-substring-no-properties bounds-start (point))))
+                        (prefix (buffer-substring-no-properties bounds-start (point)))
+                        (lsp--capf-no-reordering t))
```
yyoncho, line 4986: based on @payneca comment I rather rever that and have it yes/no instead of have it only initial sort.

lsp-mode.el
```diff
@@ -4814,6 +4820,9 @@ Return `nil' when fails to guess prefix."
                            'lsp-completion-score score?))
              it)))
 
+(defvar lsp--capf-no-reordering nil
```
yyoncho, line 4823: seems like this should be public?

lsp-mode.el
kiennq: I still would like to keep this for initial sort only.
The argument is because of this is initial sort right after we receive results from server, so server should probably have better ordering than client. We still need filtering though.

Here is our full options https://github.com/emacs-lsp/lsp-mode/issues/1883#issuecomment-655186783

In this special case, I think we should change from server side, so it can work with more clients, instead of just relying on our client implementation behavior.

lsp-mode.el
kiennq: I can make this public, but can I keep this as `defvar`, since this option is dangerous (and quirky).
Alternatively, the option should be set is `lsp-completion-no-cache`.

kiennq:
@yyoncho WDYT about this PR? I think it's still beneficial to user to have additional options to tinker around. Even for debugging purpose.

yyoncho:
> @yyoncho WDYT about this PR? I think it's still beneficial to user to have additional options to tinker around. Even for debugging purpose.

I am fine with don't cache flag. The other stuff I will have to try it for some time. My concern is that it will lead to hard to explain behaviour. For example if we have `barfoo` and  `foobar` and prefix `fo`. Depending on `sortText` `barfoo` might be before `foobar` initially. But then when you type `foo` they will reorder which I would say is very unexpected behaviour.

kiennq:
> My concern is that it will lead to hard to explain behaviour. For example if we have `barfoo` and `foobar` and prefix `fo`. Depending on `sortText` `barfoo` might be before `foobar` initially.

Yes, but I would expect the candidates from server will be fairly accurate initially.
Even with current fuzzy logic, candidates can still change rank depends on what you type, so one key stroke and radically change the candidates ranking

Issue #1853, opened 2020-06-29
Company mode with lsp-mode stops working


benwainwright:
Before logging the bug, please make sure that:

1. You are using the latest version of `lsp-mode` related packages. ✅
2. You may also try reproduce the issue using clean environment using the following command ✅  (see below)
3. Check FAQ and Troubleshooting section (https://github.com/emacs-lsp/lsp-mode#faq and https://github.com/emacs-lsp/lsp-mode#troubleshooting) ❌  The problem is related to the way that `company` interacts with `lsp-mode`, and isn't covered there.

**Describe the bug**
I use `company-mode` in conjunction with `lsp-mode` and it seems that when I'm using it, after a period of time, it just... stops working. `lsp-mode` appears to be still active and working, but the company autosuggestion box seems to just stop appearing. At the present time I only seem to be able to resolve it by restarting emacs. Any suggestions on what I can do to debug this would be much appreciated.

I've already raised an issue with `company-mode` (https://github.com/company-mode/company-mode/issues/991) and they've advised me to post here

**To Reproduce**
I've managed to reproduce by taking `lsp-start-plain.el` and adding `company` to the package list and the specific parts of my config that relate to company and lsp (also `evil` because I can't really use raw emacs, so its hard to test without it):

```
(defvar company-lsp-cache-candidates)
(setq company-lsp-cache-candidates t)

(setq company-backends (remove 'company-gtags company-backends))
(setq company-backends (remove 'company-etags company-backends))
(setq company-minimum-prefix-length 0)
(setq company-dabbrev-downcase 0)

(setq company-idle-delay 0)
(setq lsp-prefer-capf t)
(setq read-process-output-max (* 1024 1024))
(setq gc-cons-threshold 100000000)
(add-hook 'after-init-hook 'global-company-mode)

(setenv "TSSERVER_LOG_FILE" "/tmp/tsserver.log")
(add-hook 'typescript-mode-hook #'lsp)

(evil-mode 1)
```

**Expected behavior**
Autosuggestions continue to appear until the end of time

**Which Language Server did you use**
`ts-ls`

**OS**
MacOS 10.14.6

kiennq:
Can you follow https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/ and gather `*lsp-log*` and workspace log (`lsp-toggle-trace-io` and `lsp-workspace-show-log`) after you can repro your issue?

benwainwright:
[...]

PR #1887, opened 2020-07-07
lsp--capf-cache: also clear cache on company completion start

kiennq:
Another potential fix for #1853

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7146,6 +7146,7 @@ returns the command to execute."
 (defun lsp--setup-company ()
   (add-hook 'company-completion-started-hook
             (lambda (&rest _)
+              (lsp--capf-clear-cache)
               (setq-local lsp-inhibit-lsp-hooks t))
             nil
             t)


kiennq:
@yyoncho Sorry for this but I think I need to revert this.
The `company-completion-started-hook` is running **after** `c-a-p-f`, thus it will always clear the cache when first getting from server.

yyoncho:
> @yyoncho Sorry for this but I think I need to revert this.
The company-completion-started-hook is running after c-a-p-f, thus it will always clear the cache when first getting from server.

Then the option is to use advice on appropriate company function? 

CC @dgutov

dgutov:
Have you tried `company-after-completion-hook`?

Speaking of using the cache, do you do that because some servers reply slowly? IIUC Eglot doesn't cache.

Another option would be to try and resurrect https://github.com/company-mode/company-mode/issues/975.

kiennq:
> Have you tried `company-after-completion-hook`?

We've already used that, but it seems it's not enough. I.e,: in this bug, it seems the hook is not called.



> Speaking of using the cache, do you do that because some servers reply slowly? IIUC Eglot doesn't cache.

The server has an option of specifying if the returned completion items are all it can do, via `isIncomplete`.
In that case, querying more to server is pointless, and actually take time. So we cache the completion items in that case.

yyoncho:
> Speaking of using the cache, do you do that because some servers reply slowly? IIUC Eglot doesn't cache.

Some servers return thousand of completion items and do not filter server side. Not using cache is not only slow but it is killing the garbage collector. And the protocol is designed to work this way.

yyoncho:
@kiennq did you had a chance to check if this will work for us? https://github.com/company-mode/company-mode/issues/975

@dgutov is company-completion-started-hook called before capf the expected behaviour? If yes, probably there is a suitable function to use in combination with add-advice? 

@kiennq Alternatively, we may keep the modified tick in the cache and do not clean it the first time...

dgutov:
> We've already used that, but it seems it's not enough. I.e,: in this bug, it seems the hook is not called.

Did you determine the exact reason why? It might be that the first time your completion function is called, it returns nil. And company doesn't enter completion. So simply avoiding caching the value nil could be one solution.

Another is in the linked issue. Where/how do you store the cache, BTW?

yyoncho:
> Where/how do you store the cache, BTW?

we keep the cache in `defvar` when the appropriate from lsp point of view.

> It might be that the first time your completion function is called, it returns nil. 

We might return `nil` if there is no time to calculate the result due to user input. I guess this is causing the finished hook not to be called.

dgutov:
> We might return nil if there is no time to calculate the result due to user input.

BTW, you should only do that if `non-essential` is `t`.

But it's a different issue: one if whether you save that `nil` value to the said defvar. You shouldn't.

But if you avoid caching any `nil` results at all, that aspect will be moot anyway.

kiennq:
> But it's a different issue: a problem if you save that `nil` value to the said defvar. You shouldn't.
> 
> But if you avoid caching any `nil` results at all, that aspect will be moot anyway.

We've already not caching `nil` result already.

I think that may cause problem, since the completion hook is not called, the cache will never be cleaned.

dgutov:
Not sure if I understand you here.

Are there situations where the after-completion-hook is not called after your completion function returns a non-empty table?

If you can show an example, I'll make sure to try and fix it.

yyoncho:
> Are there situations where the after-completion-hook is not called after your completion function returns a non-empty table?
> 
> If you can show an example, I'll make sure to try and fix it.

we don't have a reproducer but it is happening.

dgutov:
How about an execution trace? You can add some loud function `company-after-completion-hook`. E.g. one that calls `message`.

PR #1895, opened 2020-07-09
Added option to skip kind in completion

yyoncho:


lsp-mode.el
```diff
@@ -800,6 +800,12 @@ ignored."
   :group 'lsp-mode
   :package-version '(lsp-mode . "6.3.2"))
 
+(defcustom lsp-completion-show-kind t
+  "Whether or not to show kind of completion candidates."
+  :type 'boolean
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.0.1"))
```
ericdallo, line 807: As this is a new feature, shouldn't this be `7.1.0`?

ericdallo: Oh, forget it, it's a existent feature

yyoncho: @kiennq started using 7.0.1. Both with 7.1.0 work for me. We may add snapshot in the end, e. g. `7.1.0-SNAPSHOT`.

ericdallo: I like to follow semantic versioning on `lsp-dart`, with `x.1.x` being new features and `x.x.1` being a patch/fix

yyoncho: > I like to follow semantic versioning on `lsp-dart`, with `x.1.x` being new features and `x.x.1` being a patch/fix

I am fine with this approach.

PR #1896, opened 2020-07-09
lsp-completion-at-point: invalidate cache when the bounds-start is  different

kiennq:
This will have aggressively invalidate cache.
Potential fix for #1853


kiennq:
> 2\. Remove the advice for cleaning the cache the hooks and use only 1) as cache invalidation mechanism

We can but since the cache also contain markers, we should clean the cache ASAP (after completion) rather than wait for next key typing.

PR #1897, opened 2020-07-10
Activate rust-analyzer inlay hints mode in :after-open-fn 

jerrypnz:
Check whether inlay hints are enabled in the :after-open-fn rather than at load time. Currently it requires setting `lsp-rust-analyzer-server-display-inlay-hints` before loading lsp-rust to activate it.

This also fixes the issue where it attempts to activate inlay hints for other non-rust buffers when there are two active workspaces with one of them being RA


yyoncho:
IMO using `:after-open-fn` in clients registration is suitable for this custom initialization.

jerrypnz:
@yyoncho Thanks for the pointer, i'll take a look at `:after-open-fn`

jerrypnz:
Changed it to an `:after-open-fn` instead. It indeed looks nicer.

yyoncho:
Thank you!

PR #1898, opened 2020-07-11
Fix and improve rust-analyzer runnables support

### LINKED ISSUE #1833: rust-analyzer: update runnables request
The name has changed: https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/rust-analyzer/src/lsp_ext.rs#L129

jerrypnz:
Update the command based on the new Runnables protocol, see [rust-analyzer lsp-extensions doc](https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#runnables).

Also stores the last runnable label (instead of the full runnable which may no longer be valid if there is a code change) in a buffer local var.

This fixes #1833.

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -670,7 +670,7 @@ The command should include `--message=format=json` or similar option."
 
 (defun lsp-rust-analyzer--runnables ()
   (lsp-send-request (lsp-make-request
-                     "rust-analyzer/runnables"
+                     "experimental/runnables"
                      (lsp-make-rust-analyzer-runnables-params
                       :text-document (lsp--text-document-identifier)
                       :position? (lsp--cur-position)))))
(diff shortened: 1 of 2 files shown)


lsp-rust.el
```diff
@@ -666,38 +666,48 @@ The command should include `--message=format=json` or similar option."
...
-     (if (functionp 'cargo-process-mode) 'cargo-process-mode nil)
-     (lambda (_) (concat "*" label "*")))
-    (setq lsp-rust-analyzer--last-runnable runnable)))
+  "Select and run a runnable action."
+  (interactive (list (lsp-rust-analyzer--select-runnable nil)))
+  (-let* (((&rust-analyzer:Runnable :kind :label :args) runnable)
+          ((&rust-analyzer:RunnableArgs :cargo-args :executable-args :workspace-root?) args)
+          (default-directory (or workspace-root? default-directory)))
+    (if (not (string-equal kind "cargo"))
+        (message "'%s' runnable is not supported" kind)
```
yyoncho, line 699: I would rather use `lsp--error` or `user-error` to make that more visible.

lsp-rust.el
```diff
@@ -666,38 +666,48 @@ The command should include `--message=format=json` or similar option."
     (display-buffer buf)))
 
 ;; runnables
-(defvar lsp-rust-analyzer--last-runnable nil)
+(defvar-local lsp-rust-analyzer--last-runnable-label nil)
```
flodiebold, line 669: Not sure whether making this buffer-local is a good idea -- it's nice to be able to run a test and then rerun it while working on the main code

jerrypnz: That's a good point I haven't thought about. In that case I probably shouldn't change the design here, i.e. it should store the full `Runnable` record rather than just a label.

yyoncho: By the way if we use `compile` instead of `compilation-start` (I am not sure if this is possible) then one could use `recompile` to run the last command.

jerrypnz:
@yyoncho Re your question

> Can you elaborate on this?

So there are a couple of things I was trying to improve:

1. Dedupe the list so that the last runnable does not show up twice
2. If a runnable is no longer valid (e.g. the test name has changed), does not show it in the list
3. Make each buffer has its own last runnable, so that when the user switch buffers

After @flodiebold pointed out the use case "run a test and then rerun it while working on the main code", I decided making it a buffer local might not be the best idea so I changed it back and implemented 1) differently.

> As a side note, this code can be reused for lenses and eventually it should be wired up with dap-mode, let me know if you want to implement that as well.

Are you talking about the execution part (i.e. run the command via `compile` or whatever)? The protocol itself is an extension and I don't see how it's related to lenses / dap-mode. That said, I'd be happy to help if you can point me to some existing code.

lsp-rust.el
jerrypnz: Had a quick look at `compilation-start` and `compile`. I guess `compilation-start` was used because it needs to customize the buffer name and mode?

yyoncho:
> > As a side note, this code can be reused for lenses and eventually it should be wired up with dap-mode, let me know if you want to implement that as well.
> 
> Are you talking about the execution part (i.e. run the command via `compile` or whatever)? The protocol itself is an extension and I don't see how it's related to lenses / dap-mode. That said, I'd be happy to help if you can point me to some existing code

So, AFACS the lenses(which are enabled via M-x lsp-lens-mode) use the same data structure to provide run/debug functionality which could be wired to dap-mode. Here it is an example how this is implemented in lsp-metals - https://github.com/emacs-lsp/lsp-metals/blob/master/lsp-metals.el#L406 .

Also, I believe that the current core (the run functionality) could be also wired to `dap-debug`.

And finally, IMO we should use vscode-cpp debugger as the most capable for rust debugger for which we do not have integration inside dap-mode but it is something with relatively high priority. 

```
(defun dap-cppdbg--debug-test-at-point ()
  (interactive)
  (dap-debug (dap--template "Rust :: Debug test (at point)")))


(defun dap-cppdbg--populate-test-at-point (conf)
  "Populate CONF with the required arguments."
  (let ((cwd (lsp-workspace-root)))
    (plist-put conf :dap-server-path
           '("/Users/adamkowalski/.vscode-insiders/extensions/ms-vscode.cpptools-0.28.0-insiders/debugAdapters/OpenDebugAD7"))
    (plist-put conf :cwd cwd)
    (plist-put conf :request "launch")
    (plist-put conf :program (concat cwd "/target/debug/deps/test_parser-3aef394e7183e621"))
    (plist-put conf :externalTerminal :json-false)
    (plist-put conf :MIMode "lldb")
    (plist-put conf :args '("--test" "parse_local_define"))
    conf))


(dap-register-debug-provider "cppdbg" 'dap-cppdbg--populate-test-at-point)
(dap-register-debug-template "Rust :: Debug test (at point)"
                 (list :type "cppdbg"
                   :name "Rust :: Debug test (at point)"))
```

jerrypnz:
> So, AFACS the lenses(which are enabled via M-x lsp-lens-mode) use the same data structure to provide run/debug functionality which could be wired to dap-mode. 

Did some quick test and looked at the request/response payloads and I see what you mean. Yes that sounds like a better idea. I'm keen to look into it but can't promise anything for now. There is a lot to learn about DAP and rust-analyzer debugging support. Maybe that can be tracked in a separate issue.

yyoncho:
> I'm keen to look into it but can't promise anything for now. There is a lot to learn about DAP and rust-analyzer debugging support. Maybe that can be tracked in a separate issue.

No worries - there is no pressure or any expectations at all. I just pointed out about functionality that you might be interested in implementing. 

One more thing - please give us a heads up when you start working on something to avoid collisions.

Issue #1823, opened 2020-06-19
[Feature] make our breadcrumb better than vscode's one


yyoncho:
Few ideas discussed offline with @ericdallo 

1. ~Include the project path (like vscode)~ https://github.com/emacs-lsp/lsp-mode/pull/1901
2. Allow clicking on each node and browsing the current level (like vscode).
3. ~Allow going to each segment with a prefix e. g. `C-3 M-x lsp-breadcrumb-open` will go to the third element in the breadcrumb (unique)~ https://github.com/emacs-lsp/lsp-mode/pull/1901
4. Allow going to each element via `avy` (unique)
5. We might also mark the elements with errors like vscode does. The same goes for folders as well. If we decide to implement that we should be careful with the performance.
6. Implement some kind of strategy to make the paths shorter when it is more than a particular length. The priority should be symbol data, file name, project, path. (we start truncating from the path).
7. Middle-click on file name should reset narrowing.

wpchou:
Maybe We could make use of [header-lines](https://www.gnu.org/software/emacs/manual/html_node/elisp/Header-Lines.html)

ericdallo:
@wpchou we currently use header-line-format to build the LSP headerline already :)

danielmartin:
Some additional ideas we could implement from old Emacs Semantic: https://www.gnu.org/software/emacs/manual/html_node/semantic/Sticky-Func-Mode.html

ericdallo:
Yeah, this mode seems really nice for our use case, can we use it or is it deprecated?

danielmartin:
It's not deprecated, but it may be a bit coupled to Semantic. We can get ideas from it. For example, I like the idea of copying or narrowing to function. Narrowing from the breadcrumb may be more useful than what VSCode offers. You could focus on only one function, perform search and replace locally, or even undo only in that particular function.

yyoncho:
I think that middle-click on the file name should reset the narrowing.

ericdallo:
I updated the issue description updating the current status on master

yyoncho:
5. We might also mark the elements with errors like vscode does. The same goes for folders as well. If we decide to implement that we should be careful with the performance.

yyoncho:
6. Implement some kind of strategy to make the paths shorter when it is more than a particular length. The priority should be symbol data, file name, project, path. (we start truncating from the path).

PR #1899, opened 2020-07-12
Improve breadcrumb

ericdallo:
Improves on our current breadcrumb following some ideas from https://github.com/emacs-lsp/lsp-mode/pull/1899:

* Add project path on breadcrumb (like vscode), currently without any mouse support/action 
* Add mouse support for document symbols
  * mouse-1: go to symbol 
  * mouse-2: narrow to region (following @danielmartin [suggestion](https://github.com/emacs-lsp/lsp-mode/issues/1823#issuecomment-651365097))
* Add function to go to symbol and function to narrow via prefix args: `C-3 M-x lsp-breadcrumb-go-to-symbol`

image

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2097,15 +2097,14 @@ The `:global' workspace is global one.")
                                       (lsp-treemacs-symbol-icon kind))))
                            :background (face-attribute 'header-line :background))))))
 
-(lsp-defun lsp--headerline-breadcrumb-go-to-symbol ((&DocumentSymbol :selection-range (&Range :start)))
+(lsp-defun lsp--headerline-breadcrumb-go-to-symbol ((&DocumentSymbol :selection-range (&RangeToPoint :start)))
   "Go to breadcrumb symbol."
   (->> start
-       lsp--position-to-point
        goto-char))
 
-(lsp-defun lsp--headerline-breadcrumb-narrow-to-symbol ((&DocumentSymbol :range (&Range :start :end)))
+(lsp-defun lsp--headerline-breadcrumb-narrow-to-symbol ((&DocumentSymbol :range (&RangeToPoint :start :end)))
   "Narrow to breadcrumb symbol range."
-  (narrow-to-region (lsp--position-to-point start) (lsp--position-to-point end)))
+  (narrow-to-region start end))
 
 (lsp-defun lsp--headerline-with-action ((symbol &as &DocumentSymbol :name) symbol-string)
   "Build action for SYMBOL and SYMBOL-STRING."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2144,7 +2144,7 @@ The `:global' workspace is global one.")
 (defun lsp--headerline-dirs-until-root (root-path path)
   "Find recursively the folders until the project ROOT-PATH.
 PATH is the current folder to be checked."
-  (let ((cur-path (list (f-base path))))
+  (let ((cur-path (list (f-filename path))))
     (if (lsp-f-same? root-path (lsp-f-parent path))
         cur-path
       (append (lsp--headerline-dirs-until-root root-path (lsp-f-parent path)) cur-path))))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -2087,24 +2087,73 @@ The `:global' workspace is global one.")
...
-                              (concat symbol2-icon symbol2-name)
-                            symbol2-name))))
+                          (lsp--headerline-with-action symbol-to-append full-symbol-2))))
               symbols-hierarchy ""))
 
+(defun lsp--headerline-dirs-until-root (root-path path)
+  "Find recursively the folders until the project ROOT-PATH.
+PATH is the current folder to be checked."
+  (let ((cur-path (list (f-base path))))
+    (if (f-same? root-path (f-parent path))
```
kiennq, line 2139: `lsp-f-same?`, unless you want to follow the file symlink

kiennq: Btw, will this function to be called frequently? Following the symlink is quite slow. So you may want to rewrite `f-parent` so it's not using `f-same?` too.

ericdallo: Good catch, implemented `lsp-f-parent`

lsp-mode.el
```diff
@@ -2087,24 +2087,73 @@ The `:global' workspace is global one.")
...
 
-(defun lsp--headerline-build-string (symbols-hierarchy)
-  "Build the header-line from SYMBOLS-HIERARCHY."
+(lsp-defun lsp--headerline-breadcrumb-go-to-symbol ((&DocumentSymbol :selection-range (&Range :start)))
+  "Go to breadcrumb symbol."
+  (->> start
+       lsp--position-to-point
+       goto-char))
+
+(lsp-defun lsp--headerline-breadcrumb-narrow-to-symbol ((&DocumentSymbol :range (&Range :start :end)))
```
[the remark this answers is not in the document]
ericdallo: I didn't find any `&RangeToRegion`, it is on master?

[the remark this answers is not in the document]
yyoncho: Sorry, it is &RangeToPoint

yyoncho:
Seems like the file name is missing on the screenshot? Is that the case? Can you add the filename icons as well? You may use helm-icons package as inspiration.

ericdallo:
@yyoncho it is displaying the file name but without the extension, I'll fix it

ericdallo:
I don't know for sure how to get the icon from the filename, I'll take a look into helm-icons

lsp-mode.el
```diff
@@ -2087,24 +2097,72 @@ The `:global' workspace is global one.")
...
+      (append (lsp--headerline-dirs-until-root root-path (lsp-f-parent path)) cur-path))))
+
+(defun lsp--headerline-breadcrumb-build-prefix-string ()
+  "Build the prefix for breadcrumb."
+  (seq-reduce (lambda (last-dirs next-dir)
+                (format "%s %s %s"
+                        last-dirs
+                        (lsp--headerline-breadcrumb-arrow-icon)
+                        (propertize next-dir 'font-lock-face lsp-headerline-breadcrumb-face)))
+              (lsp--headerline-dirs-until-root (lsp--suggest-project-root) (buffer-file-name)) ""))
```
yyoncho, line 2159: it should be lsp-workspace-root.

yyoncho:
Looks good! IMO we might also introduce an option to narrow deeply nested paths(e. g. use only first letters of the directories) when the path is longer.

I am not 100% how jump should work - e. g. do we need to add a command to jump into directories, e. g. M-2 M-x lsp-breadcrump-jump-dir to jump in the root or we should have one command for jumping and threat files + symbols as one three.

ericdallo:
I see... it looks to be a good idea to implement on next, we can keep discussing on the linked issue

ericdallo:
I'll merge this and implement the icons on a next PR too.

yyoncho:
And one more thing - we should have a flag to turn off the path and keep only symbols.

PR #1900, opened 2020-07-12
Proper spacing in breadcrumb

yyoncho:


lsp-mode.el
```diff
@@ -2072,14 +2072,15 @@ The `:global' workspace is global one.")
...
-                           :background (face-attribute 'header-line :background))))))
+    (concat (propertize " " 'display
+                        (cl-list* 'image
+                                  (plist-put
+                                   (cl-copy-list
+                                    (cl-rest (get-text-property
+                                              0 'display
+                                              (lsp-treemacs-symbol-icon kind))))
+                                   :background (face-attribute 'header-line :background))))
+            " ")))
```
kiennq, line 2083: Do we really need this space?
I think in modeline/header line generally we only need prepend space for a component?

yyoncho: The original treemacs icon is two spaces. When we recreate the icon, we should preserve that, e. g.

```
(length (treemacs-get-icon-value "c" nil lsp-treemacs-theme))
```

PR #1906, opened 2020-07-13
Make semantic tokens fetching conform to spec

jamesnvc:
My reading of the [LSP specification for semantic tokens](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_semanticTokens) seems to indicate that:

  1. The method to get tokens for the entire file is `textDocument/semanticTokens/full`, not just `textDocument/semanticTokens`.
  2. The server capability to indicate support for semantic token range requests is just `range`, not `rangeProvider`.

I assume there is a reason for the current implementation, but maybe this can start a discussion if some existing language servers require some sort of backwards compatibility.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -948,6 +948,11 @@ directory")
     ("textDocument/rename" :capability :renameProvider)
     ("textDocument/selectionRange" :capability :selectionRangeProvider)
     ("textDocument/semanticTokens" :capability :semanticTokensProvider)
+    ("textDocument/semanticTokensLegacy"
+     :check-command (lambda (workspace)
+                      (with-lsp-workspace workspace
+                        (lsp-get (lsp--capability :semanticTokensProvider)
+                                 :rangeProvider))))
     ("textDocument/semanticTokensRangeProvider"
      :check-command (lambda (workspace)
                       (with-lsp-workspace workspace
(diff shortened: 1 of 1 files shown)


sebastiansturm:
thanks for the pointer! I didn't find that version of the spec when I implemented semanticTokens support; maybe this document was added or revised recently (there's a new clangd commit implementing similar changes, cf. https://github.com/llvm/llvm-project/commit/5fea54bc05a71e81e843fd9284d258cd38d7fe23). I won't be able to look into this today or tomorrow, but should be able to do it on Wednesday

jamesnvc:
> I didn't find that version of the spec when I implemented semanticTokens support; maybe this document was added or revised recently

The perils of being an early adopter! :laughing: Glad to know I wasn't just losing my mind.

sebastiansturm:
looks correct to me, and seems to work well with (very) recent builds of clangd. Rust-analyzer, OTOH, doesn't seem to support the new spec as of commit 2c67ca0. 
While having to support deprecated interfaces is annoying, the required effort is minimal in this case, and I'd rather not break semantic highlighting for almost every current user of the feature. Would you like to amend your PR such as to check for legacy-style semanticTokens support upfront and make sure to call the appropriate method?

jamesnvc:
Sounds good, I’ll have a look tomorrow.`range` vs `rangeProvider` should be easy enough to handle, but I’ll have to figure out how to distinguish whether the sever expects `textDocument/semtanticTokens` vs `textDocument/semtanticTokens/full`

jamesnvc:
Pushed another small change that attempts to determine if the server is using the "legacy" version of semanticTokens. It feels a bit funny to refer to something using a still-unreleased version of the spec as "legacy", but ¯\\\_(ツ)\_/¯

sebastiansturm:
thank you! Was on holiday the last few days, but will have a look at it this evening after work

sebastiansturm:
seems to me that change won't work if the language server conforms to the latest spec but doesn't support ranged requests (as is the case for recent clangd builds). I'd suggest to check for both `full` and `rangeProvider` for now. When we later get rid of legacy support, we should probably also replace calls to `semanticTokens/full` by appropriate ranged requests in the unlikely case that a server only supports ranged requests

jamesnvc:
Ah yes, that's a good point. I was checking `range` vs `rangeProvider` just for a way to distinguish "new" versus "old" implementations, because I'm not sure how else to distinguish. I'll have a look at how, e.g. clangd reports its capabilities.

jamesnvc:
Pushed a change that explicitly checks for the `semanticTokens/full` capability although, as you say, it would be possible for the server to be using the new API and also not have the `full` capability, in which case it would presumably give an error about the unknown method.

yyoncho:
@jamesnvc thank you for contributing to `lsp-mode`!

PR #1909, opened 2020-07-13
[breadcrumb] Fix unknown projects

ericdallo:
If user opens a file from an unknown project, we should not fail.

image

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2130,9 +2130,11 @@ The `:global' workspace is global one.")
         (file-ext (f-ext file-path)))
     (if file-ext
         (when (require 'treemacs nil t)
-          (format "%s %s"
-                  (lsp--fix-image-background (treemacs-get-icon-value file-ext nil (treemacs-theme->name (treemacs-current-theme))))
-                  filename))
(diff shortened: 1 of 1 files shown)


ericdallo:
Force merging this as it is wrong on master

lsp-mode.el
```diff
@@ -2203,17 +2208,18 @@ PATH is the current folder to be checked."
...
-                  (format "%s%s %s"
-                          (if last-dirs
-                              (concat last-dirs " ")
-                            (if (eq (cl-first lsp-headerline-breadcrumb-segments) 'path-up-to-project)
-                                ""
-                              " "))
-                          (lsp--headerline-breadcrumb-arrow-icon)
-                          (propertize next-dir 'font-lock-face lsp-headerline-breadcrumb-prefix-face)))
-                (lsp--headerline-path-up-to-project-root (lsp-workspace-root) (f-parent (buffer-file-name))) nil)))
+  (when (and (member 'path-up-to-project lsp-headerline-breadcrumb-segments))
```
leungbk, line 2211: The `and` shouldn't be required?

ericdallo: Yeah, I'll remove this, thanks!

ericdallo: Done here: https://github.com/emacs-lsp/lsp-mode/pull/1913

PR #1910, opened 2020-07-14
breadcrumb: handle files in project root

leungbk:
If `lsp-headerline-breadcrumb-segments` contains `'path-up-to-project`, the breadcrumb does not currently display for files in the project-root directory. This patch tries to address that.


ericdallo:
Tested and it works nice, I'll add some cache for that later, as path-up-to-project seems a little expensive, thank you

PR #1914, opened 2020-07-14
[breadcrumb] Move headerline to another file

ericdallo:
This is a attempt to start moving some features from lsp-mode.el to another file

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -310,16 +310,18 @@ PATH is the current folder to be checked."
   :global nil
   (cond
    (lsp-headerline-breadcrumb-mode
+    (lsp-headerline--breadcrumb-cache-path-up-to-project)
     (add-to-list 'header-line-format '(t (:eval lsp--headerline-breadcrumb-string)))
-    (add-hook 'lsp-managed-mode-hook #'lsp-headerline--breadcrumb-cache-path-up-to-project nil t)
+
     (add-hook 'lsp-on-idle-hook #'lsp-headerline--check-breadcrumb nil t)
     (add-hook 'lsp-configure-hook #'lsp--headerline-enable-breadcrumb nil t)
     (add-hook 'lsp-unconfigure-hook #'lsp--headerline-disable-breadcrumb nil t))
    (t
-    (remove-hook 'lsp-managed-mode-hook #'lsp-headerline--breadcrumb-cache-path-up-to-project t)
     (remove-hook 'lsp-on-idle-hook #'lsp-headerline--check-breadcrumb t)
     (remove-hook 'lsp-configure-hook #'lsp--headerline-enable-breadcrumb t)
     (remove-hook 'lsp-unconfigure-hook #'lsp--headerline-disable-breadcrumb t)
+
+    (setq lsp--headerline-breadcrumb-path-up-to-project-string nil)
     (setq header-line-format (remove '(t (:eval lsp--headerline-breadcrumb-string)) header-line-format)))))
 
 ;;;###autoload
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -1,0 +1,354 @@
+;;; lsp-headerline.el --- LSP headerline features -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2020 emacs-lsp maintainers
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;;  LSP headerline features
+;;
+;;; Code:
+
+(require 'lsp-mode)
+
+(defcustom lsp-headerline-breadcrumb-segments '(path-up-to-project file symbols)
+  "Face used on breadcrumb text on modeline."
+  :type '(repeat
+          (choice (const :tag "Include the project name." project)
+                  (const :tag "Include the open file name." file)
+                  (const :tag "Include the directories up to project." path-up-to-project)
+                  (const :tag "Include document symbols if server supports it." symbols)))
+  :group 'lsp-mode)
+
+(defcustom lsp-headerline-breadcrumb-enable-symbol-numbers nil
+  "Whether to label symbols with numbers on the breadcrumb."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defface lsp-headerline-breadcrumb-separator-face '((t :inherit shadow :height 0.8))
+  "Face used for breadcrumb separator on headerline."
+  :group 'lsp-faces)
+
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -7423,6 +7118,9 @@ returns the command to execute."
   (when (functionp 'lsp-ui-mode)
     (lsp-ui-mode))
 
+  (when lsp-headerline-breadcrumb-enable
+    (add-hook 'lsp-managed-mode-hook 'lsp-headerline-breadcrumb-mode))
+
```
kiennq, line 7123: Is this needed? The mode should have been enabled with `lsp-configure-hook` already

lsp-mode.el
```diff
@@ -4184,14 +3883,11 @@ in that particular folder."
 
 (defun lsp-configure-buffer ()
   (when lsp-auto-configure
+    (run-hooks 'lsp-configure-hook)
```
kiennq, line 3886: Is this new hook? You forgot to define it?

lsp-mode.el
kiennq: Or you should add it into `lsp-configure-hook` instead of `lsp-managed-mode-hook`

lsp-headerline.el
```diff
@@ -0,0 +1,345 @@
...
+
+;;;###autoload
+(define-minor-mode lsp-headerline-breadcrumb-mode
+  "Toggle breadcrumb on headerline."
+  :group 'lsp-mode
+  :global nil
+  (cond
+   (lsp-headerline-breadcrumb-mode
+    (add-to-list 'header-line-format '(t (:eval lsp--headerline-breadcrumb-string)))
+    (add-hook 'lsp-managed-mode-hook #'lsp-headerline--breadcrumb-cache-path-up-to-project nil t)
```
kiennq, line 305: Shouldn't `lsp-headerline--breadcrumb-cache-path-up-to-project` be called per `lsp-headerline-breadcumb-mode` instead?

lsp-mode.el
```diff
@@ -7423,6 +7118,9 @@ returns the command to execute."
   (when (functionp 'lsp-ui-mode)
     (lsp-ui-mode))
 
+  (when lsp-headerline-breadcrumb-enable
+    (add-hook 'lsp-managed-mode-hook 'lsp-headerline-breadcrumb-mode))
```
kiennq, line 7122: After rethinking, maybe this is better
```suggestion
    (lsp-headerline-breadcrumb-mode 1))
```
Since this is auto configure function and you don't want a hook left behind when cleaning `lsp-mode`

lsp-mode.el
ericdallo: Yes a new hook, I Forgot to define it

lsp-headerline.el
ericdallo: Yes, I think so, I can change to run when this mode is enabled so

lsp-mode.el
ericdallo: This gives byte-compile errors because `lsp-mode.el` doesn't know `lsp-headerline.el`

lsp-mode.el
ericdallo: We need it because we need to auto-enable the breadcrumb mode if `lsp-headerline-breadcrumb-enable` is `t`.
This is actually not working and I don't know why yet, For some reason the `lsp` function does not run for me.

ericdallo:
@kiennq the whole idea discussed with @yyoncho was to have `lsp-headerline.el` in a separated file requiring `lsp-mode.el`, `lsp-mode.el` should just run the hooks and `lsp-headerline.el` and any other future feature in another file should enable/disable the respective modes

lsp-mode.el
ericdallo: Maybe we should run another hook when cleaning `lsp-mode`

lsp-headerline.el
ericdallo: rethinking, I'm not sure if we can just call `lsp-headerline--breadcrumb-cache-path-up-to-project` per this mode initialization as lsp can shutdown then reconnect and we may want to recalculate that, maybe depending on managed-hook is safer.

kiennq: When you shutdown lsp and reconnect, shouldn't this mode initialization run again?
That's the whole point of `lsp--configure-buffer` isn't it?

lsp-mode.el
kiennq: Will it? Shouldn't `lsp-headerline-breadcrumb-mode` is autoloaded?
Or you can check for `functionp`, but I don't think it's necessary.

lsp-mode.el
kiennq: `lsp` will run only once per buffer (unless you disable and reenable it).

lsp-mode.el
kiennq: Resolved this one, let's discuss in another comment thread

lsp-mode.el
kiennq: Also, I think you should really use `lsp-configure-hook` here instead.
Else this `lsp-headerline-breadcrumb-mode` will also be enabled again when `(lsp-managed-mode -1)`

ericdallo: I see, I think it makes sense, I moved it to `lsp-configure-buffer`, I don't know if it is correct, but I added the hook and then later on the same function fire the lsp-configure-hook.

Now it works nice :)

lsp-headerline.el
ericdallo: Yeah, I think it's ok too, change to run when the mode is enabled

lsp-mode.el
kiennq: I think this part should stay in `lsp--auto-configure` function. Do you still have it not working even when toggle `lsp-mode` on/off again?

ericdallo: In `lsp--auto-configure` I can't make trigger, that's was my previous problem.
now in `lsp-configure-buffer` it works even toggling `lsp-mode` on/off

kiennq: I mean when in `lsp--auto-configure`, even with `lsp-mode` off/on you still cannot making it triggered?
That seems weird.
Can you try to close the buffer, and open the buffer again?
It will trigger both `lsp` and `lsp-managed-mode`, which will add the hook and also trigger `lsp-configure-buffer`.

ericdallo: Yeah, this keeps not working, even adding a `(message "-->")` [here](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L8420) and closing emacs and opening a lsp client does not print...
I wonder if it's something with my emacs, I use doom-emacs with [no-byte-compile enabled](https://github.com/ericdallo/dotfiles/blob/master/.doom.d/packages.el#L18)

ericdallo: Forget it, I compiled lsp-mode and it seems to work, I'll try to move to `lsp--auto-configure` again

ericdallo: @kiennq Well, after some many attempts, I could not reproduce it again, for some reason my `lsp` function is always compiled even using `no-byte-compile` in straight.el, or recompiling doom, but I agree that `lsp--auto-configure` is the right place so I moved to there.

lsp-mode.el
```diff
@@ -4236,6 +3932,8 @@ in that particular folder."
       (lsp--semantic-tokens-initialize-buffer
        (lsp-feature? "textDocument/semanticTokensRangeProvider"))))
 
+  (run-hooks 'lsp-configure-hook)
```
kiennq, line 3935: Shouldn't you move this inside `(when lsp-auto-configure)` block? Else when new dynamic registration changes come, breadcrumb will be enabled even without `lsp-auto-configure`.

ericdallo: Yes, done!

ericdallo:
Resolved conflicts

PR #1915, opened 2020-07-14
Optionally number breadcrumb entries

leungbk:
This obviates the need for the user to manually count symbol positions
when using lsp-breadcrumb-narrow-to-symbol and lsp-breadcrumb-go-to-symbol.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -670,7 +670,7 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :group 'lsp-faces)
 
 (defcustom lsp-headerline-breadcrumb-enable nil
-  "Wheter to enable breadcrumb on headerline."
+  "Whether to enable breadcrumb on headerline."
   :type 'boolean
   :group 'lsp-mode)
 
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -60,6 +60,20 @@
...
+    (defun seq-map-indexed (fn sequence)
+      "Return the result of applying FN to each element of SEQUENCE.
+Unlike `seq-map', FN takes two arguments: the element of
+the sequence, and its index within the sequence."
+      (let ((index 0))
+        (seq-map (lambda (elt)
+                   (prog1
+                       (funcall fn elt index)
+                     (setq index (1+ index))))
+                 sequence)))))
```
ericdallo, line 75: Can't we just use [-map-indexed](https://github.com/magnars/dash.el#-map-indexed-fn-list)?

lsp-mode.el
```diff
@@ -683,6 +697,11 @@ If this is set to nil, `eldoc' will show only the symbol information."
                   (const :tag "Include document symbols if server supports it." symbols)))
   :group 'lsp-mode)
 
+(defcustom lsp-headerline-breadcrumb-enable-symbol-numbers t
```
ericdallo, line 700: I'm not sure if this should be `t` by default, but it seems a simple UI change so I'm ok with this
image

yyoncho: I vote this to be off by default - the central feature of the headerline is the visual function, not the navigation one. Thus IMO it should be optimized for better visuals, not for navigation.

yyoncho:
Thank you!

PR #1916, opened 2020-07-14
[breadcrumb] Permit visiting out-of-bounds parent symbols within a narrowed buffer

leungbk:


kiennq:
Sorry for asking, but what's expected behavior on this?
From the code, it seems it will drop user current narrowed region and narrow to another region, is that expected?
For example if the buffer is narrowed to include the symbol `range` start but not include the end, it will not be widen to include the symbol end?

I think the behavior of hiding user current narrowed region is quite disruptive, how about change it something like **if the buffer is narrowed, narrow the buffer from symbol `range` start (not `selectionRange`) to either `range` end or `(point-max)`, depends on which one is wider.

leungbk:
@kiennq Could you please provide an example where this would produce unexpected behavior?

> From the code, it seems it will drop user current narrowed region and narrow to another region, is that expected?

My intent was to narrow to a wider region if necessary. I assumed the symbols obeyed a parenthesis property, so that "a is the parent of b, b is the parent of c" implies `a_start_pos < b_start_pos < c_start_pos < c_end_pos < b_end_pos < a_end_pos` (perhaps not with strict inequality). Are there cases where this doesn't make sense?

> how about change it something like **if the buffer is narrowed, narrow the buffer from symbol range start (not selectionRange) to either range end or (point-max), depends on which one is wider.

I narrowed to `range` start since that's what `lsp--headerline-breadcrumb-narrow-to-symbol` currently does.

kiennq:
@leungbk, from your code
``` elisp
  (when (< selection-range-start (point-min))
      (narrow-to-region
       narrowing-range-start
       narrowing-range-end))
  (goto-char selection-range-start))
```
So in case the user is narrowing to some other region, you will only check the visibility of `point-min` of symbol selection's start, then narrow to symbol region only.
The side effect of that is user current narrowed region may become invisible, which is undesirable, IMO.

What's my propose is
``` elisp
(when (buffer-narrowed-p)
  (narrow-to-region
   (min (point-min) narrowing-range-start)
   (max (point-max) narrowing-range-end)))
(goto-char selection-range-start)
```
Which will **not** hide current user visible region, but just widen it.

leungbk:
@kiennq I see, good catch. I hadn't considered the possibility that user-defined narrowings could span unbalanced parenthesis zones. OK, go ahead and commit your proposed patch.

leungbk:
@kiennq Have you submitted your fix? I don't see it in the repo.

kiennq:
No, sorry I kinda forgot about this. Will push a PR for this soon.

Issue #1863, opened 2020-07-01
Pyright support


zeronone:
Is it possible to add Pyright support? https://github.com/microsoft/pyright

yyoncho:
I moved the issue in lsp-mode repo since current repo provides support for the C# server and both doesn't seem to be related(correct me if I am wrong)

ericdallo:
Any problem supporting both `mspyls` and this new one? And we keep the best as the default? I think this gives more options to LSP users

zeronone:
> @zeronone we can use this even if it is not open-source as long as the license allows that. Do you know more about that?

Sorry, I don't know much about it either. 

Regarding `pylance`, I gives much better completions than `mypyls` thanks to type-inference. Although quite rough, I could make it work with `lsp-mode`.

```bash
# install pylance extension
code --install-extension ms-python.vscode-pylance

# add a pylance executable
cat  <<EOF > ~/bin/pylance.sh
#!/usr/bin/env bash
set -euo pipefail

node $HOME/.vscode/extensions/ms-python.vscode-pylance-2020.6.1/server/server.bundle.js --stdio
EOF
```

```elisp
(defun +register|pylance ()
  (setq lsp-pylance-ms-executable "~/bin/pylance.sh")

  (lsp-register-client
   (make-lsp-client
    :new-connection (lsp-stdio-connection (lambda () lsp-pylance-ms-executable)
                                          (lambda () (f-exists? lsp-pylance-ms-executable)))
    :major-modes '(python-mode)
    :server-id 'mspylance
    :priority 3
    :initialized-fn (lambda (workspace)
                      (with-lsp-workspace workspace
                        (lsp--set-configuration (lsp-configuration-section "python")))))))
```

jfcherng:
The [LICENSE](https://marketplace.visualstudio.com/items/ms-python.vscode-pylance/license) says

```
a)    General. You may install and use any number of copies of the software only with Microsoft Visual Studio, Visual Studio for Mac, Visual Studio Code, Azure DevOps, Team Foundation Server, and successor Microsoft products and services (collectively, the “Visual Studio Products and Services”) to develop and test your applications.
```

So it's not legit to use it outside of MS products?

yyoncho:
@jfcherng just like vscode-cpp - they are saying "it is not easy to go open source" but at the same time the license forbids anything but vscode... (this is for pylance).

seagle0128:
[...]

PR #1921, opened 2020-07-16
Add pyright language server for Python

zeronone:
Add support for Pyright language server (Python) as discussed #1863

### DIFF
--- a/lsp-pyright.el
+++ b/lsp-pyright.el
@@ -35,19 +35,19 @@
   "Command to start pyright-langserver."
   :type '(repeat string)
   :group 'lsp-pyright
-  :package-version '(lsp-mode . "7.0"))
+  :package-version '(lsp-mode . "7.0.1"))
 
 (defcustom lsp-pyright-disable-language-services nil
   "Disables all language services except for \"hover\"."
   :type 'boolean
   :group 'lsp-pyright
-  :package-version '(lsp-mode . "7.0"))
+  :package-version '(lsp-mode . "7.0.1"))
 
 (defcustom lsp-pyright-disable-organize-imports nil
   "Disables the \"Organize Imports\" command."
   :type 'boolean
   :group 'lsp-pyright
-  :package-version '(lsp-mode . "7.0"))
+  :package-version '(lsp-mode . "7.0.1"))
 
 (defcustom lsp-pyright-use-library-code-for-types nil
   "Determines whether pyright reads, parses and analyzes library code to
@@ -56,7 +56,7 @@ This can add significant overhead and may result in poor-quality type informatio
 The default value for this option is false."
   :type 'boolean
   :group 'lsp-pyright
-  :package-version '(lsp-mode . "7.0"))
+  :package-version '(lsp-mode . "7.0.1"))
 
 (defcustom lsp-pyright-diagnostic-mode "openFilesOnly"
   "Determines whether pyright analyzes (and reports errors for) all files
@@ -66,7 +66,7 @@ If this option is set to \"openFilesOnly\", pyright analyzes only open files."
           (const "openFilesOnly")
           (const "workspace"))
   :group 'lsp-pyright
-  :package-version '(lsp-mode . "7.0"))
+  :package-version '(lsp-mode . "7.0.1"))
 
 (defcustom lsp-pyright-typechecking-mode "basic"
   "Determines the default type-checking level used by pyright.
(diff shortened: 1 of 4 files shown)


leungbk:
@andrew-christianson created a PR that he asked to revert over at the lsp-python-ms repo: https://github.com/emacs-lsp/lsp-python-ms/pull/110

Maybe you guys could collaborate on this. He had been planning to offer the option to disable definition-jumping with `pyright` and use the C# server for that instead since `pyright`'s definition-jumping behavior is a bit different.

zeronone:
> @andrew-christianson created a PR that he asked to revert over at the lsp-python-ms repo: emacs-lsp/lsp-python-ms#110

Oh sorry, I didn't know about that PR. Let me know what you guys think and which approach is more viable.

andrew-christianson:
@leungbk thanks for the tag. I was lagging on re-submitting emacs-lsp/lsp-python-ms#110, but it's there now as emacs-lsp/lsp-python-ms#114.  

@zeronone thanks for this, and for microsoft/pyright#810 ! This looks good, and addresses one of my TODOs on the other PR.  To confirm: are you assuming that users will set other configuration (e.g. venvPath, pythonPath) with a pyrightconfig.json?

lsp-pyright.el
```diff
@@ -0,0 +1,182 @@
...
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection (lambda ()
+                                          (cons (lsp-package-path 'pyright)
+                                                lsp-pyright-langserver-command-args)))
+  :major-modes '(python-mode)
+  :server-id 'pyright
+  :multi-root t
+  :priority -1
```
yyoncho, line 167: this and pyls priority should be different.

lsp-pyright.el
```diff
@@ -0,0 +1,182 @@
...
+(defgroup lsp-pyright nil
+  "LSP support for python using the Pyright Language Server."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/microsoft/pyright"))
+
+(defcustom lsp-pyright-langserver-command-args '("--stdio")
+  "Command to start pyright-langserver."
+  :type '(repeat string)
+  :group 'lsp-pyright
+  :package-version '(lsp-mode . "7.0"))
```
yyoncho, line 38: nit: 7.0 -> 7.0.1

lsp-pyright.el
zeronone: As it is a new lsp-client, I set priority to lower `-2`.

zeronone:
> 2. standalone file in lsp-mode project

+1 for this option. As of now I think, it is quite simple and does not warrant a separate repo. 

----------

@andrew-christianson I added `python.pythonPath` configuration (which is basically `locate-python` function from `lsp-python-ms`). I am a little bit hesitant to add `python.venvPath` as I think it is only useful in multi-root workspaces where each project has different python version, in which case the user needs additional configuration in `pyrightconfig.json` anyway (`executionEnvironment`) - please correct me if I am wrong here!

The `locate-python` function can be improved, but for now it seems sufficient (i.e: detect `VIRTUAL_ENV` env var set by `pyvenv`)

Regarding: https://github.com/emacs-lsp/lsp-python-ms/pull/114 I actually don't how the integration would work, and how we can run two simultaneous language servers at the same time. If you think that is a better approach, I can close this PR.

leungbk:
@seagle0128 What's the preferred option here?

seagle0128:
> @seagle0128 What's the preferred option here?

I am fine with option 2 and 3. Two servers in `lsp-go`.

andrew-christianson:
I'm in favor of option 2/3 here.  I like the side-by-side version in emacs-lsp/lsp-python-ms#114, but it's easy enough to make :add-on a configurable value here vs there, and it'd probably be better to wait for pyright to stabilize a bit more.  

I'll plan to close out the other PR and contribute any changes I have for pyright here.

That said, @yyoncho I still think our discussion from #1923 would be nice for this application, so we can mask out features of any language server when running as an add-on

yyoncho:
Ok, seems like everyone is fine with 2). I will create the repo and I will send invites to @zeronone @andrew-christianson and @seagle0128 . If someone else is interested to work on that he/she could ping me.


@andrew-christianson 

> That said, @yyoncho I still think our discussion from #1923 would be nice for this application, so we can mask out features of any language server when running as an add-on

We should define what and how we want to achieve it. ATM I believe that you can alter the records in method-requirements, e. g.
changing 
``` elisp
    ("textDocument/signatureHelp" :capability :signatureHelpProvider)
```
to 
``` elisp
("textDocument/signatureHelp" 
 (lambda (workspace)
   (with-lsp-workspace workspace
     (and (lsp--capability :signatureHelpProvider)
          (not (eq (lsp--client-server-id (lsp--workspace-client workspace))
                   'mspyls))))))
```
will disable signature help for mspyls server. We can create some helper functions to make that easier but IMO, either way, the configuration should come from lsp-mode contributors and it won't be something that average user will configure.

yyoncho:
Please move the PR here - https://github.com/emacs-lsp/lsp-pyright

Issue #1922, opened 2020-07-16
lsp-make-completion-list does not accept :isIncomplete


andrew-christianson:
**Describe the bug**

`lsp-make-completion-list` does not include `isIncomplete` when given key `:isIncomplete`, but does when given key `is-incomplete`.  

When running under my normal configuration, it returns a hashtable with _just_ the `:items` key.  When running under `lsp-start-plain` this throws an error.  When running under `debug-on-error` in my configuration, it appears to launch the debugger, but the backtrace window is automatically closed within milliseconds.

This breaks [`lsp--merge-results`](https://github.com/emacs-lsp/lsp-mode/blob/9571998e1cd92d26827e770ffe6495bd08d5415e/lsp-mode.el#L1338) at L1338 while merging multiple responses to a `textDocument/completion` request, which ultimately breaks (only capf?) completion when both the primary workspace and an add-on workspace respond to that request.

**To Reproduce**

```elisp
(json-encode (lsp-make-completion-list :is-incomplete t :items '(a s d f)))
;; Emacs 27, my config and lsp-start-plain: "{\"isIncomplete\":true,\"items\":[\"a\",\"s\",\"d\",\"f\"]}"

(json-encode (lsp-make-completion-list :isIncomplete t :items '(a s d f)))
;; Emacs 27, my config:  "{\"items\":[\"a\",\"s\",\"d\",\"f\"]}"
;; Emacs 27, lsp-start-plain: Lisp error: (error "Keyword argument :isIncomplete not one of (:items :is-incomplete)")
```

Emacs Version:
```
GNU Emacs 27.0.91 (build 12, x86_64-apple-darwin19.5.0, NS appkit-1894.50 Version 10.15.5 (Build 19F101))
 of 2020-07-07
```

**Expected behavior**
`lsp-make-completion-list` should include the expected keys

**Which Language Server did you use**
I hit this while running emacs-lsp/lsp-python-ms#114, which runs MS C# python language server alongside their `pyright` language server.  But the bug does not appear to be dependent on specific data, just a pair of servers responding to a completion request

**OS**
MacOS 10.15.5

**Error callstack**
When hitting this bug when in `lsp-completion-at-point`, it barfs the entire response to `textDocument/completion` into the lsp log and the echo area.

PR #1923, opened 2020-07-16
Use :is-incomplete in lsp-make-completion-list

andrew-christianson:
This PR is the simple resolution to #1922.  I suspect there may be a more comprehensive solution involving `lsp-interface` but I have less insight into that area of this project

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1335,7 +1335,7 @@ depending on it."
                                         hovers))))
        ("textDocument/completion"
         (lsp-make-completion-list
-         :isIncomplete (seq-some
+         :is-incomplete (seq-some
                         #'lsp:completion-list-is-incomplete
                         results)
          :items (apply 'append (--map (append (if (lsp-completion-list? it)


yyoncho:
Thank you - it is typo when doing the lsp-mode lsp-protocol.el migration.

yyoncho:
@andrew-christianson 

Do you think that we should provide an option(or code snippet) to allow disabling the completion results from one of the servers for pyright and ms-pyls case?

andrew-christianson:
@yyoncho thanks for the quick merge.  I'd definitely like to have an option (or a snippet/example of how) to remove a specific capability from a server.  It'd be helpful for my msplys / mspyright case, but I think there could be other applications in areas where there are multiple competing/complementary language servers users might want to chose between.

I think `lsp--server-unregister-capability` is in the right direction, but I'm not sure if it's the best way to accomplish that

yyoncho:
I think that the easiest the most straight forward way is to alter lsp-method-requirements and lsp--find-workspaces-for - we may introduce a key :disabled-if (or better name?) in lsp-method-requirements which is a list of functions. Unregister will work too but it should be combined with altering server capabilities. This is harder to implement becase sometimes a method is enabled when a particular provider is present and sometimes a method is enabled depending on a value in server capabilities so it will require strong protocol knowledge or ability to read elisp to know what to touch to disable the method.

yyoncho:
> :disabled-if (or better name?) in lsp-method-requirements which is a list of functions

In could be different variable as well.

PR #1926, opened 2020-07-16
[Rust-Analyzer] Support SnippetTextEdit

### LINKED ISSUE #1907: "Add `#[derive]`" missing from code actions with rust-analyzer
**Bug description**
Code actions for structs don't have the "Add \`#[derive]\`" item, which used to be available when using older versions of rust-analyzer. The "Add \`#[derive]\`" item is available in vscode, though, but I can't seem to get it to appear in Emacs even when using the same version of rust-analyzer as with vscode.

**To reproduce**

1. `cargo new lsp-ra-bug-repro`
1. Use the following code for lsp-ra-bug-repro/src/main.rs:
    ```rust
    struct WantToPrint {
        x: i32,
        y: i32,
    }
    
    fn main() {
        let wtp = WantToPrint { x: 1234, y: 5678, };
        println!("{:?}", wtp);
    }
    ```
1. Move cursor/caret to the beginning of the first line
1. Invoke lsp-mode's execution of code actions (I pressed `C-c l a a` having configured it thus)
1. An option to "Add \`#[derive]\`" is missing. Here's how it appears in a very bare-bones Emacs with only rust-mode and lsp-mode manually installed, after pressing TAB a couple of times after the previous step:
emacs_no_derive
1. Additionally, pressing RET here gives the error "Wrong type argument: hash-table-p, nil". I don't remember this happening when using an older version of rust-analyzer, but I suppose it's somehow linked to how I seem to get an empty(?) item at the beginning of the minibuffer when using Ido mode on my default configuration. Here's how the empty item looks in that case – the `{ |` at the beginning without any text between those characters, none of the actually usable code actions being currently selected:
emacs_empty_item
To be clear, pressing RET on that empty item in Ido mode will also yield the same wrong type argument error message.

**Expected behavior**
An option to "Add \`#[derive]\`" should be available. Here's how it appears in vscode (the item is highlighted because I had my mouse cursor on it, which isn't visible in the screenshot):
vscode_derive_exists

**Environment**
OS: Ubuntu 20.04
Emacs: GNU Emacs 26.3 (build 2, x86_64-pc-linux-gnu, GTK+ Version 3.24.14) of 2020-03-26, modified by Debian
lsp-mode: 20200713.506 from MELPA (I've also tested on several other versions during the past month)
rust-analyzer: 5ca7cd9 (included with vscode), 4b1c372 (manually built today, 2020-07-13), also several others during the past month

**LSP Mode workspace log**
I captured this [workspace log](https://github.com/emacs-lsp/lsp-mode/files/4913739/workspace.log) from pressing `C-c l a a` with the above file open and the cursor where stated in the above reproduction instructions.

yyoncho:
Fixes #1907


lsp-protocol.el
```diff
@@ -475,6 +475,7 @@ See `-let' for a description of the destructuring mechanism."
  (TextDocumentItem (:languageId :text :uri :version) nil)
  (TextDocumentSyncOptions nil (:change :openClose :save :willSave :willSaveWaitUntil))
  (TextEdit (:newText :range) nil)
+ (SnippetTextEdit (:newText :range) (:insertTextFormat))
```
ericdallo, line 478: is this part of LSP spec? or is rust specific? if so should this go [here](https://github.com/emacs-lsp/lsp-mode/pull/1926/files#diff-459ac98aed37663772c354a97a7006a1R270)?

yyoncho: it is implemented in vscode experimental features, at least I was unable to find it in rust analyzer client. My guess is that it will make to the spec at some point.

lsp-mode.el
```diff
@@ -4519,7 +4519,12 @@ The method uses `replace-buffer-contents'."
                  (seq-sort #'lsp--text-edit-sort-predicate)
                  (mapc (lambda (edit)
                          (progress-reporter-update reporter (cl-incf done))
-                         (funcall apply-edit edit))))
+                         (funcall apply-edit edit)
+                         (when (lsp-snippet-text-edit? edit)
+                           (-when-let ((&SnippetTextEdit :range (&RangeToPoint :start)
+                                                         :insert-text-format? :new-text) edit)
+                             (when (eq insert-text-format? lsp/insert-text-format-snippet)
+                               (yas-expand-snippet (lsp--to-yasnippet-snippet new-text) start (+ start (length new-text)))))))))
```
kiennq, line 4527: You may need to wrap this one with `let (yas-indent-line)` too. To prevent snippet indentation.

yyoncho: Fixed.

PR #1927, opened 2020-07-16
Provide lsp-diagnostic-clean-after-change setting

yyoncho:
see https://github.com/emacs-lsp/lsp-mode/pull/1820#discussion_r456030583


yyoncho:
cc @alanz

PR #1928, opened 2020-07-16
Replace .editorconfig with .dir-locals and .gitattributes

### LINKED ISSUE #1924: Prefer committing .dir-locals.el over .editorconfig?
Everyone here uses Emacs, so anything we would use a `.editorconfig` for can be done with a `.dir-locals.el` and a `.gitattributes` file.

One concrete benefit of committing a project-wide `.dir-locals.el` is that we should be able to modify `lisp-imenu-generic-expression` to detect `lsp-defun` and `lsp-interface`, allowing `M-x imenu` to be more useful. I use `imenu` all the time, but in this project I often find myself using it to find a function and then realizing "oh right it can't detect that".

leungbk:
Closes #1924.

I wanted to add the `lsp-defun` stuff to `lisp-imenu-generic-expression` instead of `imenu-generic-expression`, but I noticed that if I did that and left the `(eval...` in the `emacs-lisp-mode` section of `.dir-locals.el`, then the `imenu-generic-expression` would not have gotten updated with `lsp-defun` the first time opening a file in the directory, assuming that that directory-local setting had been previously set.

The `Normalize files` commit has a long line, so that shouldn't be viewed within Emacs.


yyoncho:
Thank you for performing this cleanup. Please fix the CI which breaks due to trying to byte-compile the dir-locals file.

leungbk:
OK, it seems to be fixed.

lsp-mode.el
```diff
@@ -8758,7 +8758,3 @@ See https://github.com/emacs-lsp/lsp-mode."
 
 (provide 'lsp-mode)
 ;;; lsp-mode.el ends here
-
-;; Local Variables:
-;; flycheck-disabled-checkers: (emacs-lisp-checkdoc)
-;; End:
```
ericdallo, line 8764: nice!

PR #1934, opened 2020-07-18
Fix bug in `lsp-headerline--filename-with-icon`

willthefrog:
returns `nil` if `lsp-treemacs` is not available

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -89,15 +89,14 @@ for caching purposes.")
 
 (defun lsp-headerline--filename-with-icon (file-path)
   "Return the filename from FILE-PATH with the extension related icon."
-  (let ((filename (f-filename file-path))
-        (file-ext (f-ext file-path)))
-    (if file-ext
-        (when (require 'lsp-treemacs nil t)
-          (if-let (icon (lsp-treemacs-get-icon file-ext))
-              (format "%s %s"
-                      (lsp-headerline--fix-image-background icon)
-                      filename)
-            filename))
+  (let ((filename (f-filename file-path)))
+    (if-let ((file-ext (f-ext file-path))
+             (icon (and file-ext
+                        (require 'lsp-treemacs nil t)
+                        (lsp-treemacs-get-icon file-ext))))
+        (format "%s %s"
+                (lsp-headerline--fix-image-background icon)
+                filename)
       filename)))
 
 (defun lsp-headerline--arrow-icon ()
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -90,8 +90,8 @@ for caching purposes.")
 (defun lsp-headerline--filename-with-icon (file-path)
   "Return the filename from FILE-PATH with the extension related icon."
   (let ((filename (f-filename file-path)))
-    (if-let ((file-ext (f-ext file-path))
-             (icon (and file-ext
+    (-if-let* ((file-ext (f-ext file-path))
+               (icon (and file-ext
                         (require 'lsp-treemacs nil t)
                         (lsp-treemacs-get-icon file-ext))))
         (format "%s %s"


lsp-headerline.el
```diff
@@ -92,12 +92,12 @@ for caching purposes.")
   (let ((filename (f-filename file-path))
         (file-ext (f-ext file-path)))
     (if file-ext
-        (when (require 'lsp-treemacs nil t)
-          (if-let (icon (lsp-treemacs-get-icon file-ext))
-              (format "%s %s"
-                      (lsp-headerline--fix-image-background icon)
-                      filename)
-            filename))
+        (if (require 'lsp-treemacs nil t)
```
leungbk, line 95: Since the `else` clauses here are all the same, you could try combining some of the `if`s to cut down on the nesting. One way of doing it might be

``` emacs_lisp
(let ((filename (f-filename file-path)))
    (if (require 'lsp-treemacs nil t)
        (if-let* ((file-ext (f-ext file-path))
                  (icon (lsp-treemacs-get-icon file-ext)))
            (format "%s %s"
                    (lsp-headerline--fix-image-background icon)
                    filename)
          filename)
      filename))
```

There might be something better though.

lsp-headerline.el
```diff
@@ -92,12 +92,12 @@ for caching purposes.")
...
+            (if-let (icon (lsp-treemacs-get-icon file-ext))
```
leungbk, line 96: The `(if-let` needs a matching `else` clause to avoid having the function return nil.

lsp-headerline.el
willthefrog: done

lsp-headerline.el
willthefrog: done

willthefrog:
> Thanks for the PR!

pleased to help, thanks for this great package! 😆

lsp-headerline.el
```diff
@@ -89,15 +89,14 @@ for caching purposes.")
...
-        (file-ext (f-ext file-path)))
-    (if file-ext
-        (when (require 'lsp-treemacs nil t)
-          (if-let (icon (lsp-treemacs-get-icon file-ext))
-              (format "%s %s"
-                      (lsp-headerline--fix-image-background icon)
-                      filename)
-            filename))
+  (let ((filename (f-filename file-path)))
+    (if-let* ((file-ext (f-ext file-path))
```
yyoncho, line 93: can you change to -if-let* since it is breaking the CI for emacs 25.x. Most likely if-let* is not present in emacs 25.

> lsp-headerline.el:90:44:Error: ‘(file-ext (f-ext file-path))’ is a malformed function
make: *** [compile] Error 1

willthefrog: done, just noticed the failing CI, works on my setup though (emacs 28.0)

lsp-headerline.el
```diff
@@ -89,15 +89,14 @@ for caching purposes.")
...
-        (file-ext (f-ext file-path)))
-    (if file-ext
-        (when (require 'lsp-treemacs nil t)
-          (if-let (icon (lsp-treemacs-get-icon file-ext))
-              (format "%s %s"
-                      (lsp-headerline--fix-image-background icon)
-                      filename)
-            filename))
+  (let ((filename (f-filename file-path)))
+    (if-let ((file-ext (f-ext file-path))
```
leungbk, line 93: At least in the version of Emacs I'm using, this appears to expand to `(if-let* ...`, which is unexpected to me but makes some sense.

I think it's less confusing to write `(if-let*` instead. But since the CI requires old Emacsen, we should use `dash`'s `(-if-let*`.

yyoncho:
@leungbk I edited the PR to fix if-let usage, do you have more comments on that?

yyoncho:
@willthefrog thank you for contributing to `lsp-mode` and congrats for your first `lsp-mode` PR! :tada:

PR #1935, opened 2020-07-18
Indent snippets everywhere but in org-mode buffers

### LINKED ISSUE #1805: Discrepancies between VSCode/Sublime and lsp-mode when using terraform-ls: incorrect indentation of nested blocks, extra schema stuff showing up during autocompletion
See [the issue I previously opened on terraform-ls](https://github.com/hashicorp/terraform-ls/issues/162).

I asked about this in #lsp-mode gitter and @yyoncho suggested that I open an issue for tracking purposes. I'm not sure if it's necessarily clear yet whether this should be fixed on the client or the server. Yyoncho suggested that:

> server authors would assume [this] to be correct - what vscode does, not what the spec says

Also seems that whatever way VSCode is handling this, Sublime Text seems to handle it in the same way.

**Describe the bug**

Essentially, (nested) autocompleted nested blocks do not get indented. The block itself has some indent information being returned from the language server, but when autocompleting a nested block, it is placed in the editor as if it were not nested. Example:

```hcl
resource "aws_appmesh_mesh" "test" {
  spec {
  egress_filter {
  0
}
}
}
```

The expected behavior would be to indent something like this:

```hcl
resource "aws_appmesh_mesh" "test" {
  spec {
    egress_filter {
      0
    }
  }
}
```

Also, this may be related. I'm just noticing this now that I've fired up VSCode for testing, but there's another discrepancy in the way that lsp-mode handles things. When I start typing `resource` in VSCode and autocomplete it, the resource block looks something like this:

```hcl
resource "" "name" {
  
}
```

And the cursor is placed inside the "" for you to type the resource name. In lsp-mode, it's autocompleted like this:

```hcl
resource "1" "name" {
  3
}
```

I believe the `1` and the `3` are coming from the number of required items in that part of the schema. You can see some of that above in `egress_filter`.

**To Reproduce**
Minimal project that I've been using for testing:

https://github.com/gruntwork-io/intro-to-terraform/tree/master/database

You will need to have terraform binary installed and on your path. You will also need to do a `terraform init` inside of the `database` folder.

Note: you may need to open the workspace at the level of `database` rather than top level `intro-to-terraform` (especially in VSCode, I believe in lsp-mode it will prompt you for which workspace to import). This is because currently the language server assumes that the top level of your workspace is the base level of the module. However, `intro-to-terraform` includes multiple modules.

This is in the new [terraform-ls](https://github.com/hashicorp/terraform-ls) server, which [hashicorp has started developing](https://www.hashicorp.com/blog/supporting-the-hashicorp-terraform-extension-for-visual-studio-code/) rather than Julio Sueiras's [terraform-lsp](https://github.com/juliosueiras/terraform-lsp) which lsp-mode uses by default.

**OS**
Arch Linux

**Error callstack**
Log output from the language server running both in VSCode and lsp-mode.

VSCode log:
https://gist.github.com/amasover/507b165beee1639f3c922492a8dfb6fd

lsp-mode log:
https://gist.github.com/amasover/ed2ba0c211bb84cc612f2acc21c79933

yyoncho:
- Fixes #1805

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4200,6 +4200,14 @@ The method uses `replace-buffer-contents'."
                                     beg (+ beg (length new-text))
                                     length)))))))))
 
+(defun lsp--indent-snippets? ()
+  "Enable indenting of snippets for everything but `org-mode'.
+
+Indending snippets is extremely slow in `org-mode' buffers since
+it has to calculate identation based on SRC block position."
+  (unless (derived-mode-p 'org-mode)
+    'auto))
+
 (defun lsp--apply-text-edits (edits)
   "Apply the edits described in the TextEdit[] object."
   (unless (seq-empty-p edits)
@@ -4230,7 +4238,7 @@ The method uses `replace-buffer-contents'."
                            (-when-let ((&SnippetTextEdit :range (&RangeToPoint :start)
                                                          :insert-text-format? :new-text) edit)
                              (when (eq insert-text-format? lsp/insert-text-format-snippet)
-                               (let (yas-indent-line)
+                               (let ((yas-indent-line (lsp--indent-snippets?)))
                                  (yas-expand-snippet (lsp--to-yasnippet-snippet new-text)
                                                      start (+ start (length new-text))))))))))
           (when (fboundp 'undo-amalgamate-change-group)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4230,7 +4238,7 @@ The method uses `replace-buffer-contents'."
                            (-when-let ((&SnippetTextEdit :range (&RangeToPoint :start)
                                                          :insert-text-format? :new-text) edit)
                              (when (eq insert-text-format? lsp/insert-text-format-snippet)
-                               (let (yas-indent-line)
+                               (let ((yas-indent-line (lsp--indent-snippets?)))
```
kiennq, line 4241: Instead of that, why don't we use something like 
```suggestion
                               (let ((yas-indent-line 'auto)
                                     (indent-line-function 'indent-relative-first-indent-point)))
```

Since I believe the server should return the snippet with some indentation already (according to server format), we shouldn't apply client indentation logic again.
However, we should insert server snippet relative to current indent point

yyoncho: > Since I believe the server should return the snippet with some indentation already (according to server format)

This, unfortunately, is not true - the indentation should be performed client side.

> (indent-line-function 'indent-relative-first-indent-point)

I tested it and it does not yield good results since everything is aligned to the first point which is incorrect.  Check the example at https://github.com/emacs-lsp/lsp-mode/issues/1805

kiennq: > > Since I believe the server should return the snippet with some indentation already (according to server format)
> 
> This, unfortunately, is not true - the indentation should be performed client side.

Given that the server side format and client side format can be different and `lsp-format-buffer` will format the current buffer according to server side format, I think that client side should not perform indentation according to major mode.
Even with #1805, the server is returning snippet with correct indentation, just that the insert point in client side has some offset, and those offsets are not respected when we expand the snippet

> I tested it and it does not yield good results since everything is aligned to the first point which is incorrect.

Then we may need to find another function so that the indentation is simply shift everything according to where the first line is expanded.


> Check the example at #1805

The problem in #1805 is because the snippet is not indented, you can see that the first line in the snippet expansion is inserted with some offset from line begin, whereas the rest is inserted without offset.
If those following line in snippet is inserted with offset, then we will have working snippet with correct indentation.

yyoncho: Consider the following case:
```
    /** 123 */
    public static void main(final String[] args) 
```

1/2/3 are the positions in which you invoke the completion. In the 3 cases the indentation returned by the server is `"\n * ${0}\n * @param args\n"` and it the 3 cases it is expected to end up with the same result so the indentation is not relative to the point in which the completion has been requested.

kiennq: I found this LSP thread
https://github.com/microsoft/language-server-protocol/issues/83
And also some proposed from VsCode lsp extension
https://github.com/microsoft/vscode/issues/57093
https://github.com/Microsoft/vscode/blob/25589f6cc4aa3450450fa0f49979fcf7835f52d4/src/vs/vscode.proposed.d.ts#L16-L39

Without looking at the real implementation, from the proposed extension, it leaves an impression on me that VsCode is actually (proposing to other editor) to just adjust leading whitespace of new lines to match with the indentation of the line where snippet is inserted.
``` ts
/**
 * Keep whitespace as is. By default, the editor adjusts leading
 * whitespace of new lines so that they match the indentation of
 * the line for which the item is accepted.
 */
```

Also, if we turn on magic indentation by client, I doubt `yaml-mode` may have been broken by that, since spaces has special meaning for it.

Come back to your case, since the snippet is inserted in the first line, the consequence new lines should be insert at the offset of the first line indentation, which works perfectly here.
And also that will not slow down `org-mode`.
We will need special handling for `yaml-mode` though, may be put that as a field (opt-out) in `lsp-client`?

yyoncho: > Also, if we turn on magic indentation by client, I doubt `yaml-mode` may have been broken by that, since spaces has special meaning for it.

We were running years with indenting being on until I stopped it when I was doing `org-mode` integration a few months ago so most likely there won't be any issues with having that on.

Is there a case that is unhandled by this PR?

kiennq: Ok, I didn't notice that. Nevertheless I think doing indent based on the line where snippet is inserted is better and not burden Emacs side. Thus it can even apply to `org-mode` and will ensure same kind of formatting between server and client

kiennq: Those can be address in another PR though

PR #1937, opened 2020-07-18
Merge completion results so :isIncomplete is preserved

yyoncho:
- without this change lsp-completion-list? may return false when isIncomplete is nil.


test/lsp-mode-test.el
```diff
@@ -0,0 +1,43 @@
...
+
+;;; Code:
+
+(require 'lsp-mode)
+(require 'ert)
+
+(ert-deftest lsp-mode-test-merging-completion-results ()
+  (let ((merged-completions (lsp--merge-results
+                             `([,(lsp-make-completion-item :label "$any"
+                                                           :kind? 2
```
kiennq, line 33: shouldn't there's constant for this `2`?

kiennq:
This is really tricky, especially for new user of `lsp-protocol` to be aware of this subtle different.
Can we modify the code in `lsp-make-%s` to only check for `when` of symbol end with `?`

lsp-protocol.el
```diff
@@ -136,9 +136,10 @@ Example usage with `dash`.
                                                        (s-dashed-words (symbol-name interface))
                                                        (substring (symbol-name label) 1)))
                                    (object value)
-                                 ,(if lsp-use-plists
-                                      `(plist-put object ,name value)
-                                    `(puthash ,(lsp-keyword->string name) value object)))))
+                                 ,@(if lsp-use-plists
+                                       `((plist-put object ,name value))
```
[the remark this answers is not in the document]
yyoncho: This operation seems redundant, can you elaborate?

yyoncho:
> This is really tricky, especially for new user of `lsp-protocol` to be aware of this subtle different.
> Can we modify the code in `lsp-make-%s` to only check for `when` of symbol end with `?`

It is tricky and inconsistent and it will stay like that due to the fact that elisp does not have false and it does not map to json seamlessly. The proposed change will lead to filling nils for all fields which will make instance-of? methods to succeed although they should not because of nil being invalid value for objects that are not boolean. If we do the check only for boolean properties then it will be still wrong/inconsistent because depending on the structure we should use either `:json-false` or `nil` for those properties. To make the things consistent we might remote the boolean properties from the constructor and force users to use setters but given the lower impact IMO it will be simpler to go as it is and eventually document this tricky handling.

test/lsp-mode-test.el
yyoncho: I will change it.

lsp-protocol.el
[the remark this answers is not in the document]
kiennq: The document for `plist-put` suggest it. It will work when `object` is `nil`.
In case of `hash-table`, it will throw error when `object` is `nil`, but for `plist` it will innocently pass, seems to return correct value but actually not changing `object`.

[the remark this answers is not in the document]
yyoncho: This generates setter, so the value is returned to the caller of the function

yyoncho:
Actually, we can do different fix - change the way parameters are passed to the constructor function, u will update the PR later today.

kiennq:
> > This is really tricky, especially for new user of `lsp-protocol` to be aware of this subtle different.
> > Can we modify the code in `lsp-make-%s` to only check for `when` of symbol end with `?`
> 
> It is tricky and inconsistent and it will stay like that due to the fact that elisp does not have false and it does not map to json seamlessly. The proposed change will lead to filling nils for all fields which will make instance-of? methods to succeed although they should not because of nil being invalid value for objects that are not boolean. If we do the check only for boolean properties then it will be still wrong/inconsistent because depending on the structure we should use either `:json-false` or `nil` for those properties. To make the things consistent we might remote the boolean properties from the constructor and force users to use setters but given the lower impact IMO it will be simpler to go as it is and eventually document this tricky handling.

I would expect `lsp-make-xyz` pass the check of `lsp-xyz?` which whatever passed in the former function though.
Even currently we cannot validate all passed parameters are in the correct type though, for example someone can pass an arbitrary not null `plist` to parameter and still `instance-of?` method will be satisficed since it's not null.

yyoncho:
> I would expect `lsp-make-xyz` pass the check of `lsp-xyz?`

I will update PR later today to address that.

lsp-protocol.el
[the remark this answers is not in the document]
kiennq: Sorry I missed this comment. Exactly because it's setter, we should expect it to set the value into object.
The returns are actually irrelevant here, since setter, as its name, is mostly used for side effect only.
However, since we're passing the object via parameter directly, I agree that `setq` is not useful here.

test/lsp-mode-test.el
```diff
@@ -0,0 +1,43 @@
...
+;;; Commentary:
+
+;;
+
+;;; Code:
+
+(require 'lsp-mode)
+(require 'ert)
+
+(ert-deftest lsp-mode-test-merging-completion-results ()
```
kiennq, line 30: We have another test for merging result under `lsp-common-test`, can you move that here so we have all merging tests under same file?

lsp-protocol.el
```diff
@@ -106,22 +106,20 @@ Example usage with `dash`.
                       ((listp object) (-all? (lambda (prop)
                                                (plist-member object prop))
                                              ',required))))
-                  `(cl-defun ,(intern (format "lsp-make-%s" (s-dashed-words (symbol-name interface))))
-                       (&key ,@(-map (-lambda ((key))
-                                       (intern (substring (symbol-name key) 1))) params))
+                  `(defun ,(intern (format "lsp-make-%s" (s-dashed-words (symbol-name interface))))
+                       (&rest plist)
```
kiennq, line 110: We're losing parameter quick view via `eldoc` with this. Do you want to do this to passing extra keys?
`&allow-other-keys` may help

yyoncho: We cant have both: if they are named we should process them one by one thus we will lose the ability to distinguish the case when the parameter is present or not.

test/lsp-mode-test.el
yyoncho: moved the other one here.

lsp-protocol.el
yyoncho: I pushed a version which generates a doc which lists what are the expected keys.

lsp-protocol.el
```diff
@@ -106,22 +106,21 @@ Example usage with `dash`.
                       ((listp object) (-all? (lambda (prop)
                                                (plist-member object prop))
                                              ',required))))
-                  `(cl-defun ,(intern (format "lsp-make-%s" (s-dashed-words (symbol-name interface))))
-                       (&key ,@(-map (-lambda ((key))
-                                       (intern (substring (symbol-name key) 1))) params))
-
+                  `(defun ,(intern (format "lsp-make-%s" (s-dashed-words (symbol-name interface))))
+                       (&rest plist)
+                     ,(format "Constructs %s from `plist.'
```
kiennq, line 111:
```suggestion
                     ,(format "Constructs %s from `plist'.
```

lsp-protocol.el
kiennq: Can't we define something like this?
``` elisp
(cl-defun lsp--capf-filter-candidates (items
                                       &rest plist
                                       &key lsp-items
                                         &allow-other-keys)
)
```

yyoncho: If we agree on that this will be breaking change and should be addressed separately, not in issue which is a blocker for using multiple language servers in one file. This approach also has disadvantages like the fact that it is mandatory to list all non-optional params and with a lot of required elements it will make the code less clear which is much bigger problem than the lack of eldoc for the params.

kiennq: Maybe I'm misunderstanding something here.
Without your new function signature, it turned into `(defun lsp-make-xyz (&rest plist))`, so you can access `plist` directly. However, the eldoc for params will be lost.

What I'm proposed is `(cl-defun lsp-make-xyz (&rest plist &key x y z))`, which will still gave you the same `plist` for using, and the eldoc for params is preserved.
What's the breaking change here?

yyoncho: I thought that you are proposing to have initial list of mandatory parameters(at least this is what your sample have). The latest example would work.

kiennq: Sorry for bad example. The first example is something from `lsp-mode.el` so it put it as it.

PR #1943, opened 2020-07-21
lsp-headerline: refactor and allow browsing clicked directories in Dired upon clicking

leungbk:


ericdallo:
I agree with @yyoncho, maybe we should add a card to lsp-mode github project https://github.com/emacs-lsp/lsp-mode/projects specifying what we are working on

ericdallo:
I created a board where we can test it: https://github.com/emacs-lsp/lsp-mode/projects/4#card-42271333

lsp-headerline.el
```diff
@@ -130,128 +130,135 @@ narrow to the outer symbol."
...
-                              ('project (lsp-headerline--build-project-string))
-                              ('file (lsp-headerline--build-file-string))
-                              ('path-up-to-project (or lsp-headerline--path-up-to-project-string
-                                                       (lsp-headerline--build-path-up-to-project-string)))
-                              ('symbols (lsp-headerline--build-symbols-string)))
-                            "")))
-              lsp-headerline-breadcrumb-segments ""))
+  (concat
+   (format "%s " (lsp-headerline--arrow-icon))
+   (s-join (format " %s "
```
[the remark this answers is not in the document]
leungbk: I've tried making it work. But I can't figure out anything cleaner than what I have now. The problem is that I would have to either use `-flatten` or use some kind of `mapconcat`-like operation across two levels of nesting.

leungbk:
> One thing - to avoid collisions, please drop a note that you start working or a particular thing.

OK, I'll make sure to do that in the future.

lsp-headerline.el
```diff
@@ -130,128 +130,140 @@ narrow to the outer symbol."
...
-                                                       (lsp-headerline--build-path-up-to-project-string)))
-                              ('symbols (lsp-headerline--build-symbols-string)))
-                            "")))
-              lsp-headerline-breadcrumb-segments ""))
+  (concat
+   (format "%s " (lsp-headerline--arrow-icon))
+   (s-join
+    (format " %s "
+            (lsp-headerline--arrow-icon))
+    (-flatten
```
yyoncho, line 256: I am a bit concerned about the memory usage of this component, since more, and more the garbage collector performance will be hitting us. This code is on the hot path so we should try to optimize it as much as possible.

ATM we create intermittent lists as a result of the component call, -map and -flatten. Can we change each of the components result to "string" and then use mapconcat? Generally, headerline elements wont be that much but this operation will be called thousand times for a single session.

leungbk: I hadn't realized that this would gum up the GC.

OK, I can change the component results back to string, which is how it was before. There are some other simplifications from this commit that could stay, though. Should I make a commit to revert this one first, or should I continue at this commit?

ericdallo: I don't see any problem in be implemented in a new PR

yyoncho: no need to revert, it is not something urgent. We should also see how the code will look like when we implement https://github.com/emacs-lsp/lsp-mode/issues/1823#issuecomment-661312426 . IMO this will be very useful unique for emacs ecosystem feature.

PR #1946, opened 2020-07-21
lsp-headerline: go back to returning strings in the components

leungbk:


lsp-headerline.el
```diff
@@ -182,88 +182,95 @@ PATH is the current folder to be checked."
...
+    (lambda (segment)
+      (let ((segment-string
+             (pcase segment
+               ('project (lsp-headerline--build-project-string))
+               ('file (lsp-headerline--build-file-string))
+               ('path-up-to-project
+                (or lsp-headerline--path-up-to-project-string
+                    (lsp-headerline--build-path-up-to-project-string)))
+               ('symbols (lsp-headerline--build-symbol-string))
+               (_ ""))))
```
yyoncho, line 266: maybe user-error here?

leungbk: How about `message`? The only thing I don't like about `user-error` is that the headerline won't update until the user fixes the error.

ericdallo: or `lsp-log`

PR #1954, opened 2020-07-23
Drop 25.x support

### LINKED ISSUE #1668: [Discussion] drop 25.x support
Opening the discussion. I am now investigating whether threads could be used to improve lsp-mode responsiveness in certain cases(e. g. file watches).

yyoncho:
Fixes #1668

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4,7 +4,7 @@
 
 ;; Author: Vibhav Pant, Fangrui Song, Ivan Yonchovski
 ;; Keywords: languages
-;; Package-Requires: ((emacs "25.1") (dash "2.14.1") (dash-functional "2.14.1") (f "0.20.0") (ht "2.0") (spinner "1.7.3") (markdown-mode "2.3") (lv "0"))
+;; Package-Requires: ((emacs "26.1") (dash "2.14.1") (dash-functional "2.14.1") (f "0.20.0") (ht "2.0") (spinner "1.7.3") (markdown-mode "2.3") (lv "0"))
 ;; Version: 7.0
 
 ;; URL: https://github.com/emacs-lsp/lsp-mode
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -4062,8 +4070,7 @@ The method uses `replace-buffer-contents'."
                     (length (- end beg)))
                 (run-hook-with-args 'before-change-functions
                                     beg end)
-                (when (fboundp 'replace-buffer-contents)
-                  (with-no-warnings (replace-buffer-contents temp)))
+                (replace-buffer-contents temp)
```
kiennq, line 4073: Just asking but what's the advantage of this `replace-buffer-contents` over `delete` and `insert`?
I've done some simple benchmark and it seems the the later one is much faster.
``` elisp
(with-current-buffer (get-buffer-create "xyz")
  (insert "1234\n123\n123333"))
(with-current-buffer (get-buffer-create "aa")
  (insert "1234\n123\n123333\n1234\n123\n123333"))

(benchmark-run 100000 '(with-current-buffer (get-buffer-create "xyz")
                       (replace-buffer-contents (get-buffer-create "aa"))))
(0.017054 0 0.0)

(benchmark-run 100000 '(with-current-buffer (get-buffer-create "xyz")
                         (delete-region (point-min) (point-max))
                         (insert (with-current-buffer (get-buffer-create "aa") (buffer-string)))))
(0.0014909999999999993 0 0.0)
```

yyoncho: replace-buffer-contents does not break lenses and overlays.

yyoncho: And also it is faster in real-world scenarios due to the fact that it takes advantage from the internal buffer representation. E. g. apply 100+ edits in 1mb file.

PR #1955, opened 2020-07-23
Remove some Emacs < 26, and fix some whitespace/indentation

leungbk:
I'm fine with not merging the whitespace/indentation commit if it'll mess up the git history too much.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1210,13 +1210,7 @@ FORMAT and ARGS i the same as for `message'."
 
 (defalias 'lsp-ht 'ht)
 
-;; `file-local-name' was added in Emacs 26.1.
-(defalias 'lsp-file-local-name
-  (if (fboundp 'file-local-name)
-      'file-local-name
-    (lambda (file)
-      "Return the local name component of FILE."
-      (or (file-remote-p file 'localname) file))))
+(defalias 'lsp-file-local-name 'file-local-name)
 
 (defun lsp-f-canonical (file-name)
   "Return the canonical, without trailing slash FILE-NAME."
@@ -1959,25 +1953,23 @@ WORKSPACE is the workspace that contains the diagnostics."
                                                         :end (&Position :line end-line))) it)
                             ((start . end) (lsp--range-to-region range)))
                       (when (= start end)
-                        (if-let ((region (and (fboundp 'flymake-diag-region)
-                                              (flymake-diag-region (current-buffer)
-                                                                   (1+ start-line)
-                                                                   character))))
+                        (if-let ((region (flymake-diag-region (current-buffer)
+                                                              (1+ start-line)
+                                                              character)))
                             (setq start (car region)
                                   end (cdr region))
                           (lsp-save-restriction-and-excursion
                             (goto-char (point-min))
                             (setq start (point-at-bol (1+ start-line))
                                   end (point-at-eol (1+ end-line))))))
-                      (and (fboundp 'flymake-make-diagnostic)
-                           (flymake-make-diagnostic (current-buffer)
-                                                    start
-                                                    end
-                                                    (cl-case severity?
-                                                      (1 :error)
-                                                      (2 :warning)
-                                                      (t :note))
-                                                    message)))))
+                      (flymake-make-diagnostic (current-buffer)
+                                               start
+                                               end
+                                               (cl-case severity?
+                                                 (1 :error)
+                                                 (2 :warning)
+                                                 (t :note))
+                                               message))))
            ;; This :region keyword forces flymake to delete old diagnostics in
            ;; case the buffer hasn't changed since the last call to the report
            ;; function. See https://github.com/joaotavora/eglot/issues/159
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -5666,7 +5657,7 @@ A reference is highlighted only if it is visible in a window."
        "textDocument/semanticTokens/full")
       (t "textDocument/semanticTokens"))
      `(:textDocument ,(lsp--text-document-identifier)
-       ,@(if region (list :range (lsp--region-to-range (car region) (cdr region))) '()))
+                     ,@(if region (list :range (lsp--region-to-range (car region) (cdr region))) '()))
```
kiennq, line 5660: This line should be changed back, emacs indentation is wrong here

lsp-mode.el
```diff
@@ -5853,12 +5844,11 @@ perform the request synchronously."
...
-  (let ((edits (lsp-request "textDocument/rename"
-                            `(:textDocument ,(lsp--text-document-identifier)
-                                            :position ,(lsp--cur-position)
-                                            :newName ,newname))))
-    (when edits
-      (lsp--apply-workspace-edit edits))))
+  (when-let ((edits (lsp-request "textDocument/rename"
+                                 `(:textDocument ,(lsp--text-document-identifier)
+                                                 :position ,(lsp--cur-position)
+                                                 :newName ,newname))))
```
kiennq, line 5850: these lines should be reverted

lsp-mode.el
leungbk: I edited the `.dir-locals.el` to use different plist indentation in the `lsp-mode` repo. If it seems fine to you then I'll make the plist changes elsewhere also.

lsp-mode.el
leungbk: Would you mind explaining? I can't spot any obviously-important difference in the expanded forms.

kiennq: Sorry I didn't notice the before change diff.
These should be change so the `:position` is aligned with `:textDocument` and so on

leungbk:
Something like this has happened several times today for me: https://github.com/emacs-lsp/lsp-mode/runs/905172961#step:5:112

I never had this problem before, which is weird.

lsp-mode.el
```diff
@@ -3882,28 +3873,29 @@ Only works when mode is 'tick or 'alive."
   (if (zerop length)
       ;; Adding something only, work from start only
       `(:range ,(lsp--range (lsp--point-to-position start)
-                            (lsp--point-to-position start))
-               :rangeLength 0
-               :text ,(buffer-substring-no-properties start end))
+                 (lsp--point-to-position start))
```
kiennq, line 3876: This seems to be wrongly formatted

lsp-mode.el
```diff
@@ -3882,28 +3873,29 @@ Only works when mode is 'tick or 'alive."
...
 
     (if (eq start end)
         ;; Deleting something only
         (if (lsp--bracketed-change-p start length)
             ;; The before-change value is bracketed, use it
             `(:range ,(lsp--range (lsp--point-to-position start)
-                                  (plist-get lsp--before-change-vals :end-pos))
-                     :rangeLength ,length
-                     :text "")
+                       (plist-get lsp--before-change-vals :end-pos))
```
kiennq, line 3885: same here

lsp-mode.el
```diff
@@ -5332,9 +5324,9 @@ It will show up only if current point has signature help."
                          (propertize
                           lsp-overlay-document-color-char
                           'face `((:foreground ,(format "#%s%s%s"
-                                                        (lsp--number->color red)
-                                                        (lsp--number->color green)
-                                                        (lsp--number->color blue))))
+                                                 (lsp--number->color red)
```
kiennq, line 5327: same here

yyoncho:
Fixing the indentation of plists is really nice idea. FTR we use alists in some methods (e. g. lsp--client-capabilities) due to the fact that plists cannot be formatted nicely.

I checked out the source, and then I applied M-x indent-region over the buffer I saw a lot of changes in it, and some of them seem to be unwanted. Do you see the same thing?

leungbk:
> I checked out the source, and then I applied M-x indent-region over the buffer I saw a lot of changes in it, and some of them seem to be unwanted. Do you see the same thing?

Yeah, this isn't a great solution. It might be easier in the long run to have `lisp-indent-function` as `common-lisp-indent-function` in the `.dir-locals.el` and just deal with the different indentation of CL (or patch the indentation we don't like); at least that will work with `indent-region`.

lsp-mode.el
leungbk: Should be fixed.

lsp-mode.el
leungbk: Should be fixed.

lsp-mode.el
leungbk: Should be fixed.

kiennq:
> Yeah, this isn't a great solution. It might be easier in the long run to have `lisp-indent-function` as `common-lisp-indent-function` in the `.dir-locals.el` and just deal with the different indentation of CL (or patch the indentation we don't like); at least that will work with `indent-region`.

Oh, I've done that (that's why `&allow-other-keys` is indent slightly different with other `&`) and it will become really troublesome soon, we need to patch nearly all macro, especially `dash.el`

yyoncho:
ATM if the user keeps customizing section in his init.el the function evaluation will be inserted in it. Apart from moving some of the definitions that are currently in .dir-locals.el to lsp-mode.el do you know of any other solution to avoid that?

leungbk:
> ATM if the user keeps customizing section in his init.el the function evaluation will be inserted in it. Apart from moving some of the definitions that are currently in .dir-locals.el to lsp-mode.el do you know of any other solution to avoid that?

Sorry, I don't know how to avoid this.

PR #1960, opened 2020-07-24
formatting-params: using indent as tab size

kiennq:
In Emacs we have indent width which different from major mode to major mode and `tab-width`, which is mostly unchanged.
In LSP, the concept of indent width and tab width is merged and considered as `tabSize`.
Until now we have a mitigation for `c-related-mode` to using `c-basic-offset` instead of `tab-width`. However, that should be reconsidered as indent width for each major mode.
After all, it's indent width that's used to indent line (and need formatting) instead of tab width.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -69,7 +69,6 @@
 (declare-function dap-auto-configure-mode "ext:dap-mode")
 
 (defvar company-backends)
-(defvar c-basic-offset)
 (defvar yas-inhibit-overlay-modification-protection)
 (defvar yas-indent-line)
 (defvar dap-auto-configure-mode)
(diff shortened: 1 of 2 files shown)


yyoncho:
Do you know why the build does not run?

I like the idea - we may try also do the reverse - have `lsp-indent-size` and when it is set lsp-mode will set it in all modes to the proper variable. This will introduce a single entry point for configuring emacs ident size. I have seen people in reddit asking for that functionality.

lsp-mode.el
```diff
@@ -5431,13 +5430,54 @@ If ACTION is not set it will be selected from `lsp-code-actions-at-point'."
    ((stringp command?) (lsp--execute-command action))
    ((lsp-command? command?) (lsp--execute-command command?))))
 
+(defvar lsp--formatting-indent-alist
```
yyoncho, line 5433: is there a source for that or you did it by yourself?

kiennq: I took it from `dtrt-indent-mode`.

kiennq:
> Do you know why the build does not run?

I have no idea too, maybe our Github minutes are running out?
At least the build passed as in here https://github.com/kiennq/lsp-mode/tree/bug/formatting-params


> I like the idea - we may try also do the reverse - have `lsp-indent-size` and when it is set lsp-mode will set it in all modes to the proper variable. This will introduce a single entry point for configuring emacs ident size. I have seen people in reddit asking for that functionality.

That would be nice. However, I'm using `dtrt-indent-mode` to automatically figure out correct indentation based on buffer histogram.
It's not working with `lsp-mode` before because of we're setting `tab-width`, now it works perfectly.

kiennq:
Oh, force another push somehow fixes this

Issue #1842, opened 2020-06-25
Support for source.fixAll code action


muirdm:
There apparently is a kind of code action called `source.fixAll` meant to be automatically run in save hooks. I learned of this via gopls, but it is mentioned [here](https://code.visualstudio.com/api/references/vscode-api#CodeActionKind). Should we add support in lsp-mode to make it easier to set up a hook, perhaps by adding a function to invoke this kind of code action similar to `lsp-organize-imports`?

Have we considered automatically configuring save hooks for organizeImports/format-buffer?

ericdallo:
I didn't know about that kind, I think we could check the source kind and run a custom lsp hook

@yyoncho what do you think?

yyoncho:
I think it most likely narrows down to what @muirdm mentioned: 

> perhaps by adding a function to invoke this kind of code action similar to lsp-organize-imports


@muirdm 

> Have we considered automatically configuring save hooks for organizeImports/format-buffer?

IMO this should be a user option. A lot of users does like intrusive code changes because it introduces unrelated changes to what they are doing now.

samhedin:
Alright, I have no idea how LSP or elisp works, but over summer I have time to spend some time to learn. Would this only be a matter of copy-pasting everything related `lsp-organize-imports` and changing some key areas?

``` emacs-lisp
(defun lsp-organize-imports ()
  "Perform the source.organizeImports code action, if available."
  (interactive)
  (condition-case nil
      (lsp-execute-code-action-by-kind "source.organizeImports")
    (lsp-no-code-actions
     (when (called-interactively-p 'any)
       (lsp--info "source.organizeImports action not available")))))
```
to

``` emacs-lisp
(defun lsp-fix-all ()
  "Perform the source.fixAll code action, if available."
  (interactive)
  (condition-case nil
      (lsp-execute-code-action-by-kind "source.fixAll")
    (lsp-no-code-actions
     (when (called-interactively-p 'any)
       (lsp--info "source.fixAll action not available")))))
```

`"ro" lsp-organize-imports (lsp-feature? "textDocument/codeAction")` 
to
`"ra" lsp-fix-all (lsp-feature? "textDocument/codeAction")`?

I also added it to the list in `lsp--client-capabilities`.

Testing on this file:
``` go
package main

import "fmt"

func main() {
	var pow = []int{1, 2, 3, 4}
	for _ = range pow {
		fmt.Println("yes")
	}
}
```
[...]

Issue #1634, opened 2020-05-01
lsp-eslint-auto-fix-on-save implemented?


jkdufair:
I have

`(setq lsp-eslint-auto-fix-on-save t)`

and it appears to have no effect. Looking at the source, perhaps this is not implemented yet? If not, I can attempt a PR.

yyoncho:
definitely something that is not implemented, PR would be great. Feel free to ask questions here or in gitter chat(generally, the first step is to find out what vscode does).

jkdufair:
Since we already have `lsp-eslint-apply-all-fixes` would it be sufficient to simply run that when we save? I've just added it as a `before-save-hook` in my Emacs startup for the time being and it works perfectly. Or is it more involved than that?

yyoncho:
It depends on what you are looking for: if you just want to apply the fixes - I think we are good with that solution. We are looking for feature parity with vscode - in this case, vscode might be doing more.

leungbk:
The `eslint` server supports the `source.fixAll` code action, so this should be the same issue as https://github.com/emacs-lsp/lsp-mode/issues/1842.

yyoncho:
Closing as duplicate of #1842

PR #1966, opened 2020-07-25
[lsp-eslint] Handle source.fixAll code action

leungbk:
* lsp-mode: define the macro lsp-make-interactive-code-action

Related to #1842 and #1634.

I tested this with VSCode ESLint at commit d89696237a74e16fd6c09c95ae1c0830e6774ef4, prior to the weird `confirmLocalESLint` request. This seems to work from anywhere in the buffer as long as there is something that can be autofixed.

I'm not sure how best to put this on `before-save-hook` since doing it that way will make the fix-all function run in non-ESLint buffers.

### DIFF
--- a/lsp-eslint.el
+++ b/lsp-eslint.el
@@ -88,6 +88,16 @@
   :type 'boolean
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-eslint-fix-all-problem-type
+  "all"
+  "Determines which problems are fixed when running the
+source.fixAll code action."
+  :type '(choice
+          (const "all")
+          (const "problems")
+          string)
+  :package-version '(lsp-mode . "7.0"))
+
 (defcustom lsp-eslint-quiet nil
   "Turns on quiet mode, which ignores warnings."
   :type 'boolean
@@ -178,7 +188,8 @@
                     (with-current-buffer buffer
                       (list :validate "probe"
                             :packageManager lsp-eslint-package-manager
-                            :codeActionOnSave t
+                            :codeActionOnSave (list :enable t
+                                                    :mode lsp-eslint-fix-all-problem-type)
                             :format (lsp-json-bool lsp-eslint-format)
                             :options (or lsp-eslint-options (ht))
                             :run (or lsp-eslint-run "onType")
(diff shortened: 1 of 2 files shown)


yyoncho:
Looks good to me. You may add the current lsp-mode version tag to the new property.

> I'm not sure how best to put this on `before-save-hook` since doing it that way
> will make the fix-all function run in non-ESLint buffers.

Check this one: https://github.com/emacs-lsp/lsp-mode/pull/1939#issuecomment-663561975

lsp-mode.el
```diff
@@ -5509,14 +5509,19 @@ If ACTION is not set it will be selected from `lsp-code-actions-at-point'."
...
-  (interactive)
-  (condition-case nil
-      (lsp-execute-code-action-by-kind "source.organizeImports")
-    (lsp-no-code-actions
-     (when (called-interactively-p 'any)
-       (lsp--info "source.organizeImports action not available")))))
+(defmacro lsp-make-interactive-code-action (func-name code-action-kind)
+  "Define an interactive function FUNC-NAME that attempts to
+execute a CODE-ACTION-KIND action."
+  `(defun ,func-name ()
```
kiennq, line 5515: I would suggest to automatically append prefix `lsp-` into `func-name`, since its cleaner than looking at two lsp name in same line.
So it will become
`(lsp-make-interactive-code-action organize-imports...)

lsp-eslint.el
```diff
@@ -88,6 +88,16 @@
...
 
+(defcustom lsp-eslint-fix-all-problem-type
+  "all"
+  "Determines which problems are fixed when running the
+source.fixAll code action."
+  :type '(choice
+          (const "all")
+          (const "problems")
+          string)
+  :package-version '(lsp-mode . "7.0"))
```
kiennq, line 99: package should be `7.0.1`

lsp-eslint.el
```diff
@@ -178,7 +188,9 @@
                     (with-current-buffer buffer
                       (list :validate "probe"
                             :packageManager lsp-eslint-package-manager
-                            :codeActionOnSave t
+                            :codeActionOnSave
+                            (list :enable t
```
kiennq, line 192: nit: can you move this `(list ...)` to same line with `:codeActionOnSave`?
That would make it easier to distinguish (and the line is not that long)

yyoncho:
Thank you!

PR #1967, opened 2020-07-26
[migration] Move completion to file

ericdallo:
Sorry for the huge PR, but I think it's way better organized now.

- Move all completion/company/capf functions to `lsp-completion.el`
- Create `lsp-completion-mode` to be able to enable and disable the feature as the other LSP features files.
- Define obsolete `lsp-prefer-capf` -> `lsp-completion-prefer-capf`
- Define obsolete `lsp-enable-completion-at-point` -> `lsp-completion-enable`
- Fix: Only auto enable completion if `lsp-auto-enable` is enabled like the other features, (before it was ignoring `lsp-auto-enable`, now only enable the mode if this variable is `t`).

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -24,13 +24,16 @@
 (require 'lsp-mode)
 
 (define-obsolete-variable-alias 'lsp-prefer-capf
-  'lsp-completion-prefer-capf  "lsp-mode 7.0.1")
+  'lsp-completion-provider  "lsp-mode 7.0.1")
 
-(defcustom lsp-completion-prefer-capf nil
-  "Prefer capf."
-  :type 'boolean
+(defcustom lsp-completion-provider :capf
+  "The completion backend provider."
+  :type '(choice
+          (const :tag "Prefer company-capf" :capf)
+          (const :tag "Prefer company-capf" t)
+          (const :tag "Deprecated - Prefer company-lsp" nil))
   :group 'lsp-mode
-  :package-version '(lsp-mode . "6.3"))
+  :package-version '(lsp-mode . "7.0.1"))
 
 (defcustom lsp-completion-enable-additional-text-edit t
   "Whether or not to apply additional text edit when performing completion.
@@ -489,18 +492,19 @@ Others: TRIGGER-CHARS"
     (setq-local completion-category-defaults
                 (add-to-list 'completion-category-defaults '(lsp-capf (styles basic))))
 
-    (when (lsp--capability :completionProvider)
-      (cond
-       ((and (functionp 'company-lsp)
-             (not lsp-completion-prefer-capf))
-        (progn
-          (company-mode 1)
-          (add-to-list 'company-backends 'company-lsp)
-          (setq-local company-backends (remove 'company-capf company-backends))))
-
-       ((and (fboundp 'company-mode) lsp-completion-enable)
+    (cond
+     ((and (null lsp-completion-provider)
+           (functionp 'company-lsp))
+      (progn
         (company-mode 1)
-        (add-to-list 'company-backends 'company-capf))))
+        (add-to-list 'company-backends 'company-lsp)
+        (setq-local company-backends (remove 'company-capf company-backends))))
+
+     ((and (or (eq lsp-completion-provider :capf)
+               lsp-completion-provider)
+           (fboundp 'company-mode))
+      (company-mode 1)
+      (add-to-list 'company-backends 'company-capf)))
 
     (when (bound-and-true-p company-mode)
       (add-hook 'company-completion-started-hook
@@ -514,13 +518,11 @@ Others: TRIGGER-CHARS"
                   (setq-local lsp-inhibit-lsp-hooks nil))
                 nil
                 t))
-    (add-hook 'lsp-configure-hook #'lsp-completion--enable nil t)
     (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t))
    (t
     (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
     (setq-local completion-category-defaults
                 (cl-remove 'lsp-capf completion-category-defaults :key #'car))
-    (remove-hook 'lsp-configure-hook #'lsp-completion--enable t)
     (remove-hook 'lsp-unconfigure-hook #'lsp-completion--disable t)
     (when (featurep 'company)
       (remove-hook 'company-completion-started-hook
@@ -535,7 +537,8 @@ Others: TRIGGER-CHARS"
 
 ;;;###autoload
 (add-hook 'lsp-configure-hook (lambda ()
-                                (when lsp-completion-enable
+                                (when (and lsp-auto-configure
+                                           lsp-completion-enable)
                                   (lsp-completion--enable))))
 
 (provide 'lsp-completion)
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -473,32 +473,7 @@ Others: TRIGGER-CHARS"
           (setq this-command 'self-insert-command)))
     (lsp-completion--capf-clear-cache)))
 
-(defun lsp-completion--setup-company ()
-  "Setup company-mode."
-  (add-hook 'company-completion-started-hook
-            (lambda (&rest _)
-              (setq-local lsp-inhibit-lsp-hooks t))
-            nil
-            t)
-  (add-hook 'company-after-completion-hook
-            (lambda (&rest _)
-              (lsp-completion--capf-clear-cache)
-              (setq-local lsp-inhibit-lsp-hooks nil))
-            nil
-            t))
-
-(defun lsp-completion--clean-company ()
-  "Clean company-mode."
-  (remove-hook 'company-completion-started-hook
-               (lambda (&rest _)
-                 (setq-local lsp-inhibit-lsp-hooks t))
-               t)
-  (remove-hook 'company-after-completion-hook
-               (lambda (&rest _)
-                 (lsp-completion--capf-clear-cache)
-                 (setq-local lsp-inhibit-lsp-hooks nil))
-               t))
-
+;;;###autoload
 (defun lsp-completion--auto-configure ()
   "Auto configure completion."
   (cond
@@ -516,20 +491,11 @@ Others: TRIGGER-CHARS"
 
 (defun lsp-completion--enable ()
   "Enable LSP completion support."
-  (when (and lsp-completion-enable
-             (lsp-feature? "textDocument/completion"))
-    (setq-local completion-at-point-functions nil)
-    (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
-    (setq-local completion-category-defaults
-                (add-to-list 'completion-category-defaults '(lsp-capf (styles basic))))
-    (lsp-completion-mode 1)))
+    (lsp-completion-mode 1))
 
 (defun lsp-completion--disable ()
   "Disable LSP completion support."
-  (lsp-completion-mode -1)
-  (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
-  (setq-local completion-category-defaults
-              (cl-remove 'lsp-capf completion-category-defaults :key #'car)))
+  (lsp-completion-mode -1))
 
 ;;;###autoload
 (define-minor-mode lsp-completion-mode
@@ -539,15 +505,42 @@ Others: TRIGGER-CHARS"
   :lighter ""
   (cond
    (lsp-completion-mode
+    (when (and lsp-completion-enable
+               (lsp-feature? "textDocument/completion"))
+      (setq-local completion-at-point-functions nil)
+      (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
+      (setq-local completion-category-defaults
+                  (add-to-list 'completion-category-defaults '(lsp-capf (styles basic)))))
     (add-hook 'lsp-configure-hook #'lsp-completion--enable nil t)
     (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t)
     (when (bound-and-true-p company-mode)
-      (lsp-completion--setup-company)))
+      (add-hook 'company-completion-started-hook
+                (lambda (&rest _)
+                  (setq-local lsp-inhibit-lsp-hooks t))
+                nil
+                t)
+      (add-hook 'company-after-completion-hook
+                (lambda (&rest _)
+                  (lsp-completion--capf-clear-cache)
+                  (setq-local lsp-inhibit-lsp-hooks nil))
+                nil
+                t)))
    (t
+    (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
+    (setq-local completion-category-defaults
+                (cl-remove 'lsp-capf completion-category-defaults :key #'car))
     (remove-hook 'lsp-configure-hook #'lsp-completion--enable t)
     (remove-hook 'lsp-unconfigure-hook #'lsp-completion--disable t)
     (when (featurep 'company)
-      (lsp-completion--clean-company)))))
+      (remove-hook 'company-completion-started-hook
+                   (lambda (&rest _)
+                     (setq-local lsp-inhibit-lsp-hooks t))
+                   t)
+      (remove-hook 'company-after-completion-hook
+                   (lambda (&rest _)
+                     (lsp-completion--capf-clear-cache)
+                     (setq-local lsp-inhibit-lsp-hooks nil))
+                   t)))))
 
 ;;;###autoload
 (add-hook 'lsp--auto-configure-hook #'lsp-completion--auto-configure)
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -474,21 +474,6 @@ Others: TRIGGER-CHARS"
     (lsp-completion--capf-clear-cache)))
 
 ;;;###autoload
-(defun lsp-completion--auto-configure ()
-  "Auto configure completion."
-  (cond
-   ((and (functionp 'company-lsp)
-         (not lsp-completion-prefer-capf))
-    (progn
-      (company-mode 1)
-      (add-to-list 'company-backends 'company-lsp)
-      (setq-local company-backends (remove 'company-capf company-backends))))
-
-   ((and (fboundp 'company-mode) lsp-completion-enable)
-    (company-mode 1)
-    (add-to-list 'company-backends 'company-capf)))
-  (lsp-completion-mode 1))
-
 (defun lsp-completion--enable ()
   "Enable LSP completion support."
     (lsp-completion-mode 1))
@@ -511,8 +496,20 @@ Others: TRIGGER-CHARS"
       (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
       (setq-local completion-category-defaults
                   (add-to-list 'completion-category-defaults '(lsp-capf (styles basic)))))
-    (add-hook 'lsp-configure-hook #'lsp-completion--enable nil t)
-    (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t)
+
+    (when (lsp--capability :completionProvider)
+      (cond
+       ((and (functionp 'company-lsp)
+             (not lsp-completion-prefer-capf))
+        (progn
+          (company-mode 1)
+          (add-to-list 'company-backends 'company-lsp)
+          (setq-local company-backends (remove 'company-capf company-backends))))
+
+       ((and (fboundp 'company-mode) lsp-completion-enable)
+        (company-mode 1)
+        (add-to-list 'company-backends 'company-capf))))
+
     (when (bound-and-true-p company-mode)
       (add-hook 'company-completion-started-hook
                 (lambda (&rest _)
@@ -524,7 +521,9 @@ Others: TRIGGER-CHARS"
                   (lsp-completion--capf-clear-cache)
                   (setq-local lsp-inhibit-lsp-hooks nil))
                 nil
-                t)))
+                t))
+    (add-hook 'lsp-configure-hook #'lsp-completion--enable nil t)
+    (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t))
    (t
     (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
     (setq-local completion-category-defaults
@@ -543,7 +542,7 @@ Others: TRIGGER-CHARS"
                    t)))))
 
 ;;;###autoload
-(add-hook 'lsp--auto-configure-hook #'lsp-completion--auto-configure)
+(add-hook 'lsp--auto-configure-hook #'lsp-completion--enable)
 
 (provide 'lsp-completion)
 ;;; lsp-completion.el ends here
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -62,7 +62,7 @@ ignored."
   :group 'lsp-mode
   :package-version '(lsp-mode . "7.0.1"))
 
-(defconst lsp--completion-item-kind
+(defconst lsp-completion--item-kind
   [nil
    "Text"
    "Method"
@@ -93,11 +93,14 @@ ignored."
 (defvar yas-indent-line)
 (defvar company-backends)
 
+(defvar lsp-completion--capf-no-reordering nil
+  "Dont do client-side reordering completion items when set.")
+
 (declare-function company-mode "ext:company")
 (declare-function company-doc-buffer "ext:company")
 (declare-function yas-expand-snippet "ext:yasnippet")
 
-(defun lsp--make-completion-item (item &rest plist)
+(defun lsp-completion--make-item (item &rest plist)
   "Make completion item from lsp ITEM and PLIST."
   (-let (((&CompletionItem :label
                            :insert-text?
@@ -112,17 +115,17 @@ ignored."
                 'lsp-completion-markers markers
                 'lsp-completion-prefix prefix)))
 
-(defun lsp--annotate (item)
+(defun lsp-completion--annotate (item)
   "Annotate ITEM detail."
   (-let (((&CompletionItem :detail? :kind?) (plist-get (text-properties-at 0 item)
                                                        'lsp-completion-item)))
     (concat (when (and lsp-completion-show-detail detail?)
               (concat " " (s-replace "\r" "" detail?)))
             (when lsp-completion-show-kind
-              (when-let (kind-name (and kind? (aref lsp--completion-item-kind kind?)))
+              (when-let (kind-name (and kind? (aref lsp-completion--item-kind kind?)))
                 (format " (%s)" kind-name))))))
 
-(defun lsp--looking-back-trigger-characterp (trigger-characters)
+(defun lsp-completion--looking-back-trigger-characterp (trigger-characters)
   "Return trigger character if text before point match any of the TRIGGER-CHARACTERS."
   (unless (= (point) (point-at-bol))
     (seq-some
@@ -132,22 +135,22 @@ ignored."
             trigger-char))
      trigger-characters)))
 
-(defvar lsp--capf-cache nil
+(defvar lsp-completion--capf-cache nil
   "Cached candidates for completion at point function.
 In the form of list (prefix prefix-pos items :lsp-items ...).
 When the completion is incomplete, cache contains value of `incomplete'.")
 
-(defun lsp--capf-clear-cache (&rest _)
+(defun lsp-completion--capf-clear-cache (&rest _)
   "Clear completion caches."
-  (-some-> (and (listp lsp--capf-cache) lsp--capf-cache)
+  (-some-> (and (listp lsp-completion--capf-cache) lsp-completion--capf-cache)
     (cddr)
     (cdr)
     (plist-get :markers)
     (cl-second)
     (set-marker nil))
-  (setq lsp--capf-cache nil))
+  (setq lsp-completion--capf-cache nil))
 
-(lsp-defun lsp--capf-guess-prefix ((item &as &CompletionItem :text-edit?))
+(lsp-defun lsp-completion--capf-guess-prefix ((item &as &CompletionItem :text-edit?))
   "Guess ITEM's prefix start point according to following heuristics:
 - If `textEdit' exists, use insertion range start as prefix start point.
 - Else, find the point before current point is longest prefix match of
@@ -173,7 +176,7 @@ Return `nil' when fails to guess prefix."
         (setq char-before (char-before start)))
       start-point))))
 
-(defun lsp--capf-client-items (items)
+(defun lsp-completion--capf-client-items (items)
   "Convert ITEMS into client items form."
   (--> items
        (-map (-lambda ((item &as &CompletionItem
@@ -187,10 +190,7 @@ Return `nil' when fails to guess prefix."
                            'lsp-completion-score score?))
              it)))
 
-(defvar lsp--capf-no-reordering nil
-  "Dont do client-side reordering completion items when set.")
-
-(cl-defun lsp--capf-filter-candidates (items
+(cl-defun lsp-completion--capf-filter-candidates (items
                                        &rest plist
                                        &key lsp-items
                                        &allow-other-keys)
@@ -223,7 +223,7 @@ Also, additional data to attached to each candidate can be passed via PLIST."
                                            cand)
                         cand)))
                   items))
-         (if lsp--capf-no-reordering
+         (if lsp-completion--capf-no-reordering
              it
            (sort it (lambda (o1 o2)
                       (> (get-text-property 0 'sort-score o1)
@@ -231,9 +231,9 @@ Also, additional data to attached to each candidate can be passed via PLIST."
          ;; TODO: pass additional function to sort the candidates
          (-map (-partial #'get-text-property 0 'lsp-completion-item) it))
       lsp-items)
-    (-map (lambda (item) (apply #'lsp--make-completion-item item plist))))))
+    (-map (lambda (item) (apply #'lsp-completion--make-item item plist))))))
 
-(defun lsp--capf-company-match (candidate)
+(defun lsp-completion--capf-company-match (candidate)
   "Return highlight of typed prefix inside CANDIDATE."
   (let* ((prefix (downcase
                   (buffer-substring-no-properties
@@ -267,7 +267,7 @@ Also, additional data to attached to each candidate can be passed via PLIST."
         (setq label-pos 0)))
     matches))
 
-(defun lsp--capf-get-documentation (item)
+(defun lsp-completion--capf-get-documentation (item)
   "Get doc comment for completion ITEM."
   (unless (get-text-property 0 'lsp-completion-resolved item)
     (let ((resolved-item
@@ -282,14 +282,14 @@ Also, additional data to attached to each candidate can be passed via PLIST."
     (lsp:completion-item-documentation?)
     (lsp--render-element)))
 
-(defun lsp--capf-get-context (trigger-characters)
+(defun lsp-completion--capf-get-context (trigger-characters)
   "Get completion context with provided TRIGGER-CHARACTERS."
   (let* (trigger-char
          (trigger-kind (cond
-                        ((setq trigger-char (lsp--looking-back-trigger-characterp
+                        ((setq trigger-char (lsp-completion--looking-back-trigger-characterp
                                              trigger-characters))
                          lsp/completion-trigger-kind-trigger-character)
-                        ((equal lsp--capf-cache 'incomplete)
+                        ((equal lsp-completion--capf-cache 'incomplete)
                          lsp/completion-trigger-kind-trigger-for-incomplete-completions)
                         (t lsp/completion-trigger-kind-invoked))))
     (apply #'lsp-make-completion-context
@@ -298,7 +298,7 @@ Also, additional data to attached to each candidate can be passed via PLIST."
             (when trigger-char
               `(:trigger-character? ,trigger-char))))))
 
-(defun lsp--sort-completions (completions)
+(defun lsp-completion--sort-completions (completions)
   "Sort COMPLETIONS."
   (sort
    completions
@@ -321,7 +321,7 @@ Also, additional data to attached to each candidate can be passed via PLIST."
                                (save-excursion
                                  (ignore-errors
                                    (goto-char (+ it 1))
-                                   (while (lsp--looking-back-trigger-characterp trigger-chars)
+                                   (while (lsp-completion--looking-back-trigger-characterp trigger-chars)
                                      (cl-incf it)
                                      (forward-char))
                                    it)))
@@ -332,17 +332,17 @@ Also, additional data to attached to each candidate can be passed via PLIST."
               (cond
                (done? result)
                ((and (not lsp-completion-no-cache)
-                     lsp--capf-cache
-                     (listp lsp--capf-cache)
-                     (equal (cl-second lsp--capf-cache) bounds-start)
-                     (s-prefix? (car lsp--capf-cache)
+                     lsp-completion--capf-cache
+                     (listp lsp-completion--capf-cache)
+                     (equal (cl-second lsp-completion--capf-cache) bounds-start)
+                     (s-prefix? (car lsp-completion--capf-cache)
                                 (buffer-substring-no-properties bounds-start (point))))
-                (apply #'lsp--capf-filter-candidates (cddr lsp--capf-cache)))
+                (apply #'lsp-completion--capf-filter-candidates (cddr lsp-completion--capf-cache)))
                (t
                 (-let* ((resp (lsp-request-while-no-input
                                "textDocument/completion"
                                (plist-put (lsp--text-document-position-params)
-                                          :context (lsp--capf-get-context trigger-chars))))
+                                          :context (lsp-completion--capf-get-context trigger-chars))))
                         (completed (or (and resp (not (lsp-completion-list? resp)))
                                        (not (lsp:completion-list-is-incomplete resp))))
                         (items (lsp--while-no-input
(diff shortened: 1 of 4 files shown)


lsp-completion.el
```diff
@@ -0,0 +1,542 @@
...
+
+;;;###autoload
+(define-minor-mode lsp-completion-mode
+  "Toggle LSP completion support."
+  :group 'lsp-mode
+  :global nil
+  :lighter ""
+  (cond
+   (lsp-completion-mode
+    (when (bound-and-true-p company-mode)
```
kiennq, line 531: This one is not correct, since `company-mode` and completion are loosely coupled.
Before even without `company-mode` user can still get completion via `completion-at-point`, now since this check, the completion at point functions will not be added

kiennq:
> * Fix: Only auto enable completion if `lsp-auto-enable` is enabled like the other features, (before it was ignoring `lsp-auto-enable`, now only enable the mode if this variable is `t`).

I'm worry about this, since completion is quite a core functionality to LSP as I can tell.
Before, completion is always enabled regardless of `lsp-auto-enable`, but now if old/advanced user ever set `lsp-auto-enable` to `nil`, they will loose the completion functionality, and it's not easy (no official hook) to re-enable it, given the hook in `lsp-configure-buffer` will only be run when `lsp-auto-enable`.

What I can suggest is that we should probably scrap the dual check of `lsp-auto-enable` and `lsp-xyz-enable` everywhere.
Instead we can set the `lsp-xyz-enable` to `nil` for feature that need auto enable to be used, and `t` otherwise.

The `lsp-auto-enable` will only being used in `lsp--auto-configure` and will set all the `lsp-xyz-enable` to `t`. Or even better, we can expose `lsp--auto-configure-hook` (internal hook) so each feature can add its own auto configure.
Just put a line like bellow and you will have the `add-hook` for auto configure appeared on autoloads and no need to add new code in `lsp-mode.el`.
``` elisp
;;;###autoload
(add-hook 'lsp--auto-configure-hook (lambda () (......)))
```

With these change, it will be straight forward for user to configure `lsp-mode`, just need to enable `lsp-xyz-enable` for feature they want while disable auto configure. Or enable auto configure and have everything ready.

ericdallo:
@kiennq it really seems a good idea, I'll try to do that, thank you!

lsp-completion.el
ericdallo: Good to know, I thought company-mode was required to use completion.
I moved out the 2 hooks of this check.

ericdallo:
Done @kiennq, I really liked your suggestion, with that we can move the `lsp-xyz-enable` to its own feature, after this PR I'll do that to it remaining feature

lsp-mode.el
```diff
@@ -3474,46 +3388,45 @@ in that particular folder."
       (lsp--remove-overlays 'lsp-links)
 
       (remove-hook 'xref-backend-functions #'lsp--xref-backend t)
-      (setq-local global-mode-string (remove status global-mode-string))
-      (when (bound-and-true-p company-mode)
-        (lsp--clean-company))))))
+      (setq-local global-mode-string (remove status global-mode-string))))))
 
 (defun lsp-configure-buffer ()
-  (when lsp-auto-configure
```
yyoncho, line 3482: why do we remove lsp-auto-configure check?

lsp-completion.el
```diff
@@ -0,0 +1,553 @@
...
+      (lsp-completion--setup-company)))
+   (t
+    (remove-hook 'lsp-configure-hook #'lsp-completion--enable t)
+    (remove-hook 'lsp-unconfigure-hook #'lsp-completion--disable t)
+    (when (bound-and-true-p company-mode)
+      (lsp-completion--clean-company)))))
+
+;;;###autoload
+(add-hook 'lsp--auto-configure-hook #'lsp-completion--enable)
```
yyoncho, line 550: why do we have to go like this?

ericdallo: This way we can handle the auto configure function to it feature and its own file

ericdallo: Do you see any problem with this way?

yyoncho: lsp-completion--enable is not autoloaded so this most likely will cause trouble. I don't like having executable code on top level in a file. Why do we need two hooks `lsp--auto-configure-hook` and `lsp-configure-hook`? What is different here compared to the rest of the stuff that we moved to separate files?

lsp-completion.el
```diff
@@ -0,0 +1,553 @@
...
+  "Setup company-mode."
+  (cond
+   ((and (functionp 'company-lsp)
+         (not lsp-completion-prefer-capf))
+    (progn
+      (company-mode 1)
+      (add-to-list 'company-backends 'company-lsp)
+      (setq-local company-backends (remove 'company-capf company-backends))))
+
+   ((and (fboundp 'company-mode) lsp-completion-enable)
```
kiennq, line 486: These should belong to auto configure

lsp-completion.el
```diff
@@ -0,0 +1,553 @@
...
+  (cond
+   (lsp-completion-mode
+    (add-hook 'lsp-configure-hook #'lsp-completion--enable nil t)
+    (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t)
+    (when (bound-and-true-p company-mode)
+      (lsp-completion--setup-company)))
+   (t
+    (remove-hook 'lsp-configure-hook #'lsp-completion--enable t)
+    (remove-hook 'lsp-unconfigure-hook #'lsp-completion--disable t)
+    (when (bound-and-true-p company-mode)
```
kiennq, line 546: This should probably replaced with
```suggestion
    (when (featurep 'company)
```
Since want to clean `company-mode` even when it's being disabled at time of we disabling `lsp-completion-mode`

lsp-completion.el
kiennq: Except the `add-hook` stuff, I think the part of configuring `company-lsp` or `company-capf` should be moved to auto configure

lsp-completion.el
```diff
@@ -0,0 +1,553 @@
...
+               (lambda (&rest _)
+                 (lsp-completion--capf-clear-cache)
+                 (setq-local lsp-inhibit-lsp-hooks nil))
+               t))
+
+(defun lsp-completion--enable ()
+  "Enable LSP completion support."
+  (when (and lsp-completion-enable
+             (lsp-feature? "textDocument/completion"))
+    (setq-local completion-at-point-functions nil)
```
kiennq, line 521: Why don't this 3 lines belong to `lsp-completion-mode` define block?

lsp-completion.el
```diff
@@ -0,0 +1,553 @@
...
+    (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
+    (setq-local completion-category-defaults
+                (add-to-list 'completion-category-defaults '(lsp-capf (styles basic))))
+    (lsp-completion-mode 1)))
+
+(defun lsp-completion--disable ()
+  "Disable LSP completion support."
+  (lsp-completion-mode -1)
+  (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
```
kiennq, line 530: Why don't this two lines belong to `lsp-completion-mode` define block?

lsp-completion.el
ericdallo: It's just a common function to enable/disable but I can move to there

lsp-completion.el
kiennq: This will not work since `lsp-completion--enable` is not autoloaded.
You should probably replace this with something like
``` elisp
;;;###autoload
(defun lsp-completion--auto-configure ()
  (setq lsp-completion-enable t)
  (cond
   ((and (functionp 'company-lsp)
         (not lsp-completion-prefer-capf))
    (progn
      (company-mode 1)
      (add-to-list 'company-backends 'company-lsp)
      (setq-local company-backends (remove 'company-capf company-backends))))
   ((and (fboundp 'company-mode) lsp-completion-enable)
    (company-mode 1)
    (add-to-list 'company-backends 'company-capf)))
  (lsp-completion-mode 1))

;;;###autoload
(add-hook 'lsp--auto-configure-hook #'lsp-completion--auto-configure)
```

ericdallo: I just did this, but without the `(setq lsp-completion-enable t)`, why do we need it? I thought user would enable that

lsp-mode.el
kiennq: I think that we should probably remove the dual check of `lsp-auto-enable` and `lsp-xyz-enable` everywhere.
Instead we can set the `lsp-xyz-enable` to `nil` for feature that need auto enable to be used, and `t` otherwise.

The `lsp-auto-enable` will only being used in `lsp--auto-configure` and will set all the `lsp-xyz-enable` to t. Or even better, we can expose `lsp--auto-configure-hook` (internal hook) so each feature can add its own auto configure.

lsp-completion.el
ericdallo: Yeah, I changed to use another autoloaded function which will be called only on autoconfigure.
AFAIK auto-configure is different from configure, they are triggered in different places, one on `lsp` function and another on `lsp-managed-mode`.I think keep this hook of auto-configure for each feature could help organize better the features for it file

lsp-mode.el
ericdallo: you mean then that `lsp-auto-enable` will always override user definition of `lsp-xyz-enable` setting it to `t`?

yyoncho: > The `lsp-auto-enable` will only being used in `lsp--auto-configure` and will set all the `lsp-xyz-enable` to t. Or even better, we can expose `lsp--auto-configure-hook` (internal hook) so each feature can add its own auto configure.

So, if someone wants to turn off one feature he should turn off auto-enable and then go and enable the rest one? Is that what you are suggesting?

ericdallo: Yeah, this sounds bad to me, I think we should keep the `lsp-auto-enable` just enabling the features if `lsp-xyz-enable` is `t`

lsp-completion.el
yyoncho: > AFAIK auto-configure is different from configure,

I think that this is incidental - the code that was in lsp--auto-configure in the old source could be easily moved to managed-mode hook.

yyoncho: > the code that was in lsp--auto-configure

By "code" I mean - "the completion related code"

ericdallo: So, can I move that code to be enabled when `lsp-configure-hook` run?

yyoncho: I think yes. Moreover, this code is incorrect - it should check if there is completionProvider and enable capf only if it is present. And the correct place to check if xxxProvider is present is in managed mode hooks.

ericdallo: Done, could you check it again?

yyoncho: lsp--auto-configure-hook is no longer needed?

ericdallo: Well, without it, we still need to couple `lsp-xyz-enable` with `lsp-mode.el` and add:
```elisp
(when lsp-completion-enable
    (add-hook 'lsp-configure-hook 'lsp-completion-mode))
``` 
in `lsp--auto-configure` function

lsp-mode.el
kiennq: We can look at that from different angle, until now if we want to enable just an optional feature like `lens`, we have to enable `lsp-auto-enable` and disable the rest. Which is not natural at all.

If we really want to archive that:
- User can easily turn off auto configure and enable one feature
- User can easily leave auto configure on and disable one feature
We should have new flags like `lsp-auto-enable-for-xyz`, which will really be over-killed.

How about we do some thing like checking `(get 'lsp-xyz-enable 'not-autoconfig)` in auto configure function and set `lsp-xyz-enable` to `t` if that property is not set?
With that we can still promote `lsp-xyz-enable` feature when auto config is on. And user can disable feature with `(set 'lsp-xyz-enable 'not-autoconfig)`

lsp-completion.el
yyoncho: The idea behind lsp-xyz-enable was to prevent loading lsp-xyz.el . If we change all submodules to attach to that private hook we will eagerly load all features even if they are disabled.

lsp-completion.el
```diff
@@ -0,0 +1,548 @@
...
+      (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
+      (setq-local completion-category-defaults
+                  (add-to-list 'completion-category-defaults '(lsp-capf (styles basic)))))
+
+    (when (lsp--capability :completionProvider)
+      (cond
+       ((and (functionp 'company-lsp)
+             (not lsp-completion-prefer-capf))
+        (progn
+          (company-mode 1)
```
kiennq, line 505: I would say this is auto configure code, since it messes directly with user configuration.
For example, I have `company-mode`  installed and not planning to use it, instead using completion via `ivy-posframe`, this will override that configuration.

lsp-completion.el
ericdallo: well, if that is the problem we can wrap checking if `lsp-xyz-enable` is `t` before enabling the submodule, what do you think?
Something like:
```elisp

(defun lsp-completion--enable ()
  "Enable LSP completion support."
    (lsp-completion-mode 1))

;;;###autoload
(defun lsp-completion--auto-enable ()
  (when lsp-completion-enable
    (lsp-completion--enable)))

;;;###autoload
(add-hook 'lsp--auto-configure-hook #'lsp-completion--auto-enable)
```

lsp-completion.el
```diff
@@ -0,0 +1,548 @@
...
+;;;###autoload
+(define-minor-mode lsp-completion-mode
+  "Toggle LSP completion support."
+  :group 'lsp-mode
+  :global nil
+  :lighter ""
+  (cond
+   (lsp-completion-mode
+    (when (and lsp-completion-enable
+               (lsp-feature? "textDocument/completion"))
```
yyoncho, line 494: I think that this should go in lsp-completion--enable and we should not call the mode at all if this is not t.

Alternatively, you should slurp the sexp bellow in this `when`.

lsp-completion.el
kiennq: > I think that this is incidental - the code that was in lsp--auto-configure in the old source could be easily moved to managed-mode hook.

Is this correct?
I saw code that change user configuration in `lsp-auto-configure` like setting `company-backends`. So auto configure is definitely different from configuring feature.

lsp-completion.el
ericdallo: Done

lsp-completion.el
ericdallo: sorry, I don't get it, what's your suggestion?

lsp-completion.el
yyoncho: > I saw code that change user configuration in `lsp-auto-configure` like setting `company-backends`. So auto configure is definitely different from configuring feature.

@kiennq 
Changing company-backends even before the server has sent initialized response doesn't seem the right thing.

yyoncho: > well, if that is the problem we can wrap checking if `lsp-xyz-enable` is `t` before enabling the submodule, what do you think?

@ericdallo that won't help - once you call an autoloaded function from a module, the whole module is loaded.

ericdallo: Hum, I see, well, I think it's a tradeoff, right? 
1. Load the submodule even if the user doesn't use it.
2. coupling submodule with `lsp-mode.el` without loading the submodule if not enabled.

I still prefer the first one as the most lsp features are default `t` and it looks a cleaner code for me.

lsp-completion.el
kiennq: I mean the part of 
``` elisp
(cond
       ((and (functionp 'company-lsp)
             (not lsp-completion-prefer-capf))
        (progn
          (company-mode 1)
...
```
That's user configuration since it modifies `company-backends`. I have a special configuration for `capf` (like multi backend for `company-capf` and `dabbrev-code`) and by this, I will have no way of keeping my `company-backends` unchanged. It will also break my configuration.
I think this should move to `lsp-completion--auto-configuration` function like before so user can disable auto configuration for `lsp-completion` (which is set up `company-mode` and enable `lsp-completion-mode`?) and use their own initialization code for it.

ericdallo: It makes sense, done!
I just don't understand how you should be able to have other features auto enabled as if you disable `lsp-auto-configure` because of your custom company-backends it will disable LSP auto configuration for all LSP features as well.

lsp-completion.el
yyoncho: > * Load the submodule even if the user doesn't use it.

Not only loading module: loading module + all submodule dependencies. E. g. if someone does not use flymake it will still have it loaded because we will load lsp-flymake unconditionally. 

> 2\. coupling submodule with `lsp-mode.el` without loading the submodule if not enabled.

If we aim to what lsp-dart.el is - this is not a concern. Also, we have so much stuff related to integration with external components like evil, projectile, which-key, avy, etc to be concerned with this issue.

lsp-completion.el
kiennq: I disable auto configure and only enable it for some features.

lsp-completion.el
ericdallo: I see your point, I'm just not sure what is better, so should I remove this `lsp--auto-configure-hook` and go back to what we have on master for the features, what do you think @yyoncho @kiennq?

For completion, I think I'll need to move the content of `lsp-completion--auto-configure` to `lsp-mode.el`? :pensive:

kiennq: > Changing company-backends even before the server has sent initialized response doesn't seem the right thing.

@yyoncho 
That means the location of `auto-configure` should be moved after server initialized response, doesn't it?
What I mean is that the functionality of `auto-configure` and `configure` is different. While `auto-configure` can override user configuration (like `company-backends`), `configure` should not do that.

kiennq: > that won't help - once you call an autoloaded function from a module, the whole module is loaded.

That can be done in easier way while doesn't load the whole module
``` elisp
;;;###autoload
(add-hook 'lsp--auto-configure-hook (lambda () (when lsp-completion-enable (lsp-completion--enable))))
```

No autoloaded function from module is called when `lsp-completion-enable` is `nil`

yyoncho: > That means the location of `auto-configure` should be moved after server initialized response, doesn't it?

@kiennq 

Yes. This is what I am saying. `auto-configure` will just set the stuff to run in the `configure` method. I think that you did one pass with that.

> If yes, for completion, I think I'll need to move the content of lsp-completion--auto-configure to lsp-mode.el

@ericdallo why?

ericdallo: Oh, @kiennq I thought it was the same of:
```elisp
;;;###autoload
(add-hook 'lsp--auto-configure-hook #'lsp-completion--auto-configure)
```

When we use a lambda for that, emacs won't load the module if is not `t`?

ericdallo: > @ericdallo why?
@yyoncho
Where should it go the content of ` lsp-completion--auto-configure` so?

yyoncho: In `configure` hook.

ericdallo: But then I think we'll have this problem, right?
https://github.com/emacs-lsp/lsp-mode/pull/1967#discussion_r460557714

kiennq: > When we use a lambda for that, emacs won't load the module if is not `t`?

No, since it will not call any autoloaded function.

yyoncho: > But then I think we'll have this problem, right?
> [#1967 (comment)](https://github.com/emacs-lsp/lsp-mode/pull/1967#discussion_r460557714)

I don't understand that comment. @kiennq  can you elaborate?

> That's user configuration since it modifies company-backends. I have a special configuration for capf (like multi backend for company-capf and dabbrev-code) and by this, I will have no way of keeping my company-backends unchanged. It will also break my configuration.

You can disable lsp-completion-enable or use lsp-managed-mode-hook to set whatever backends you want.

kiennq: > You can disable lsp-completion-enable or use lsp-managed-mode-hook to set whatever backends you want.

@yyoncho Disable `lsp-completion-enable` is no good, since I want to use `capf`.
Using `lsp-managed-hook` to reset the backends is not good either. That means I have to copy my whole big `company-backends` in two places, one is configuration for `company-mode`, and another one is configuration for `lsp-mode`, that's redundant and doesn't isolate `lsp-mode` from `company-mode`.

My point is `company-backends` is user configuration, and I don't think it's good for other package to mess with it.
`lsp-mode` provide `auto-configure` to sweeten user configuration so it can do that while auto configure is enabled, else it shouldn't do that.

yyoncho: > @yyoncho Disable lsp-completion-enable is no good, since I want to use capf.

I don't understand - If you use capf, enabling it is one line.

> My point is `company-backends` is user configuration, and I don't think it's good for other package to mess with it.
> `lsp-mode` provide `auto-configure` to sweeten user configuration so it can do that while auto configure is enabled, else it shouldn't do that.

And that is what will happen - if you disable auto-configuration it won't touch company-backends and it won't perform any configuration so you can configure everything by yourself.

yyoncho: To elaborate a bit more, our configuration story is:

1. We provide some autoconfiguration OOTB which targets the majority of the users.
2. If the user does not want autoconfiguration - they can turn everything off via lsp-auto-configure 
3. If the user does not like particular setting - they set lsp-FEATURE-enable to nil
4. If the user does not like how a particular feature is configured - they turn it off, copy the parts that they like, remove the parts that they dont like. 


Regarding 4: most of the features are enabled within 2-5 lines. If there is more code, then it will make sense to introduce an additional variable like lsp-FEATURE-SUBFEATURE-enable to control that. But we cannot introduce a flag to turn off/on every single line of code in the autoconfiguration section.

ericdallo: I agree with all these points

kiennq: > And that is what will happen - if you disable auto-configuration it won't touch company-backends and it won't perform any configuration so you can configure everything by yourself.

I think this is too extreme.
Let's say I disable auto-configuration, do you mean that I need to configure `completion-at-point-functions`  ... by myself? Which is require knowledge both about `lsp-mode` and `completion` itself, so I don't think that's easy for user to do.
User can indeed copy those configuration lines, but it calls internal functions and there's no warrantee it will not be broken by next update.
That's why I want to separate `auto-config` from normal configuration. For example, setting `completion-at-point-functions`, `completion-category-defaults` are normal configuration, even `major-mode` already do that and should be enabled when `lsp-mode` supports it.
On other hand, `company-backends` configuration are `auto-config`, when user use `company-mode`, they probably already customize that value already.
If you want to ask where to draw the line, I think it's depend, when we want to configure optional packages that doesn't have hard dependency of `lsp-mode`, that probably should go to `auto-config`?
Does that make sense?

> 4\. If the user does not like how a particular feature is configured - they turn it off, copy the parts that they like, remove the parts that they dont like.

While I agree about configuration stories, the previous implementation that using dual check of `lsp-auto-enable` inside `lsp-configure-buffer` prevent user to run `lsp-configure-hook` when `lsp-auto-enable` is not turned on. Which is quite bad since user cannot turn on feature now.
However, if we remove `lsp-auto-enable` in `lsp-configure-hook`, all of the features `lsp-xyz-enable` which is set to `t` by default will be loaded immediately when buffer is configured.
Is there solution for that without introducing new configuration flags?

I proposed https://github.com/emacs-lsp/lsp-mode/pull/1967#discussion_r460555441 which allows user to disable a feature by setting `lsp-xyz-enable` to `nil` and put a property of `not-autoconfig` on it.

Alternatively, how about setting default value for `lsp-xyz-enable` as `init`, so we know that user haven't customize that value yet, and can do something like enable that feature if `lsp-auto-enable` or set that feature to default enablement state when `lsp-auto-enable` is not enabled?

yyoncho:
> Let's say I disable auto-configuration, do you mean that I need to
> configure `completion-at-point-functions` ... by myself? Which is
> require knowledge both about `lsp-mode` and `completion` itself, so I
> don't think that's easy for user to do.
> User can indeed copy those configuration lines, but it calls internal
> functions and there's no warrantee it will not be broken by next
> update.

This is not true: the configuration code is 3 sexps, all of them use
public interface.

```
(setq-local completion-at-point-functions nil)
(add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
(setq-local completion-category-defaults
    (add-to-list 'completion-category-defaults '(lsp-capf (styles basic))))
```

> That's why I want to separate `auto-config` from normal configuration.
> For example, setting `completion-at-point-functions`,
> `completion-category-defaults` are normal configuration, even
> `major-mode` already do that and should be enabled when `lsp-mode`
> supports it.

All of the code related to completion in lsp--auto-configure was there
by mistake and this should be fixed.

> While I agree about configuration stories, the previous implementation
> that using dual check of `lsp-auto-enable` inside
> `lsp-configure-buffer` prevent user to run `lsp-configure-hook` when
> `lsp-auto-enable` is not turned on. Which is quite bad since user
> cannot turn on feature now.

Why is quite bad? What cannot be confiured? Can you give concrete
example what cannot be configured right now? What is the issue with the
4 points listed above? It is really hard to understand what the issue is
without examples.

> However, if we remove `lsp-auto-enable` in `lsp-configure-hook`, all
> of the features `lsp-xyz-enable` which is set to `t` by default will
> be loaded immediately when buffer is configured. Is there solution for
> that without introducing new configuration flags?

I guess by lsp-auto-enable you mean lsp-auto-configure? If yes, we
should not remove that check. See point 2).

>
> I proposed
> https://github.com/emacs-lsp/lsp-mode/pull/1967#discussion_r460555441
> which allows user to disable a feature by setting `lsp-xyz-enable` to
> `nil` and put a property of `not-autoconfig` on it.

This is nice and we can use it for lsp-configure-hook, but this does not
change the fact that the code running in autoconfigure function should
not be there. We should not touch company-backends if there is no
completionProvider registered and if lsp-completion-enable is not set.

>
> Alternatively, how about setting default value for `lsp-xyz-enable` as
> `init`, so we know that user haven't customize that value yet, and can
> do something like enable that feature if `lsp-auto-enable` or set that
> feature to default enablement state when `lsp-auto-enable` is not
> enabled?

Before any further discussion, we should have concrete example of
something that is hard to be configured - otherwise we don't know what
we are trying to solve.

kiennq:
> This is not true: the configuration code is 3 sexps, all of them use public interface. ``` (setq-local completion-at-point-functions nil) (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t) (setq-local completion-category-defaults (add-to-list 'completion-category-defaults '(lsp-capf (styles basic)))) ```

@yyoncho The stuff to clearing `capf` cache too, and it's tricky. Also the function to clear cache are obviously internal.
Now about the 3 sexps, I don't think those sexps are easy to understand without good knowledge both about completion at point function and `lsp-mode`.
For example with category `lsp-capf`, that come out of no where and there's no mention for it in document.

Also the setting `completion-at-point-functions`, I've rarely seen anyone configuring that in their init file. I've quick checked `doom-emacs` and `spacemacs` and there's no code setting that thing too.
On the hand, `company-backends` are widely being configured in both configurations.

yyoncho:
I can't answer any of the questions without an example what you are
doing.

kiennq:
> I can't answer any of the questions without an example what you are doing.

Sorry for late reply.
I think I've given example (in words) in https://github.com/emacs-lsp/lsp-mode/pull/1967#issuecomment-664157166

Here is my current config
``` elisp
(use-package lsp-mode
  :quelpa (lsp-mode :repo "kiennq/lsp-mode" :fetcher github)
  :defer t
  :functions (lsp--filter-cr-compact
              lsp--client-priority
              lsp--client-server-id
              lsp--client-uri->path-fn
              lsp--client-path->uri-fn
              lsp--client-new-connection
              lsp--uri-to-path-1
              lsp--path-to-uri-1
              lsp--capability)
  :init
  (setq lsp-keep-workspace-alive nil
        lsp-signature-doc-lines 0
        lsp-idle-delay 1
        lsp-prefer-capf t
        lsp-client-packages nil)
  (setq lsp-enable-symbol-highlighting nil
        lsp-enable-links nil)
  :config
  (advice-add #'lsp--auto-configure :override
              (lambda ()
                (lsp-flycheck-enable)
                ;; (lsp--flymake-setup)
                ;; (flymake-start t)
                (let ((lsp-keymap-prefix "SPC l"))
                  (lsp-enable-which-key-integration))
                (add-hook 'lsp-configure-hook 'lsp-modeline-code-actions-mode)
                (add-hook 'lsp-configure-hook 'lsp-modeline-diagnostics-mode))
              '((name . ---custom-configure))))
```

As you can see, I have to keep `lsp-auto-configure` else `lsp-modeline-code-actions-mode` and stuff will not be run in `lsp-configure-hook`.
But to prevent my company settings to be overwritten, I have to override `lsp--auto-configure` and put manual stuff there.

Now with this PR, `lsp-configure-hook` can be run regardless of `lsp-auto-configure`, so I can turn it off. However, since all `lsp-enable-*` are by default enabled, I have to manually turn off them now. Which is annoying, especially since new features can be added with enable by default state, and I have to hunt them and (maybe) turn them off.

yyoncho:
> But to prevent my company settings to be overwritten, I have to override `lsp--auto-configure` and put manual stuff there.

What are these company settings? ATM the official way to join lsp-mode completion with other backends like yasnippets is like this:

```
(add-hook 'lsp-managed-mode-hook
          (lambda ()
            (setq-local company-backends '((company-yasnippet :with company-capf)))))
```

Is there anything more than this that you need?

kiennq:
> Is there anything more than this that you need?

I have to reconfigure `company-backends` when using `lsp-mode`, isn't that a problem?
I mean I already configured `company-backends` in `(use-package company-mode)`, why I have to reconfigure it again with `lsp-mode`? That's something unexpected, especially since `lsp-mode` have no hard dependency on `company-mode`.

yyoncho:
> We now have a opportunity to separate lsp-auto-configure from lsp-enabled-* stuff, why don't we do it?

* It is wrong to configure company-mode in auto-configure hook because the server might not support completion at all. 
* It makes the configuration code trickier, more complex and harder to explain. Why introducing 2 hooks and one of them private when we have a solution with one hook is acceptable and we are running it for like 2 years with no issues? 
* It splits company related configuration in multiple places which again makes the code harder to reason about.
* There is no valid reason to do so - in the beginning, we decided to do a few things that traditional modes don't do for the sake of working OOTB. There is the same issue with flycheck-backends and the solution is exactly the same - if you want to run multiple flycheck backends you have to setq-local flycheck-backends in managed-mode hook.

lsp-mode.el
```diff
@@ -3474,46 +3388,45 @@ in that particular folder."
...
-    (when (and lsp-enable-text-document-color
-               (lsp-feature? "textDocument/documentColor"))
-      (add-hook 'lsp-on-change-hook #'lsp--document-color nil t))
+  "Configure LSP features for current buffer."
```
yyoncho, line 3394: can you restore the auto-configure check?

lsp-completion.el
```diff
@@ -0,0 +1,551 @@
...
+                     (setq-local lsp-inhibit-lsp-hooks t))
+                   t)
+      (remove-hook 'company-after-completion-hook
+                   (lambda (&rest _)
+                     (lsp-completion--capf-clear-cache)
+                     (setq-local lsp-inhibit-lsp-hooks nil))
+                   t)))))
+
+;;;###autoload
+(add-hook 'lsp--auto-configure-hook #'lsp-completion--auto-configure)
```
yyoncho, line 548: please drop that.

As a side note, we may write the code like that:

```
;;;###autoload
(add-hook 'lsp-configure-hook (lambda ()
                                (when lsp-enable-completion
                                  (lsp-completion--auto-configure))))
```

and drop the whole auto-configure method in lsp-mode. But still, the flag should live in lsp-mode.el. 

At some point, we might switch to what you did in dap-mode - have a list of features to enable. I like this approach a bit more that single enable flags because it is one centralized place.

lsp-completion.el
```diff
@@ -0,0 +1,551 @@
...
+     ((and (fboundp 'company-mode) lsp-completion-enable)
+      (company-mode 1)
+      (add-to-list 'company-backends 'company-capf))))
+  (lsp-completion--enable))
+
+(defun lsp-completion--enable ()
+  "Enable LSP completion support."
+  (when (and lsp-completion-enable
+             (lsp-feature? "textDocument/completion")))
+  (lsp-completion-mode 1))
```
yyoncho, line 497: looks like this is not in the when.

lsp-completion.el
```diff
@@ -0,0 +1,551 @@
...
+        (setq-local lsp-inhibit-lsp-hooks nil)
+
+        (when (lsp-completion--looking-back-trigger-characterp trigger-chars)
+          (setq this-command 'self-insert-command)))
+    (lsp-completion--capf-clear-cache)))
+
+;;;###autoload
+(defun lsp-completion--auto-configure ()
+  "Auto configure LSP completion."
+  (when (lsp--capability :completionProvider)
```
yyoncho, line 479: please move that in lsp-completion-mode.

kiennq:
> * It is wrong to configure company-mode in auto-configure hook because the server might not support completion at all.

Is this being resolved if we move the auto-configure hook to right place? After the initial request?

> * It makes the configuration code trickier, more complex and harder to explain. Why introducing 2 hooks and one of them private when we have a solution with one hook is acceptable and we are running it for like 2 years with no issues?

By two hooks, do you mean the `lsp-configure-hook` and `lsp--auto-configure-hook`? Which one hook is acceptable that's you're talking about? `lsp-manage-mode-hook`?


> * It splits company related configuration in multiple places which again makes the code harder to reason about.
> * There is no valid reason to do so - in the beginning, we decided to do a few things that traditional modes don't do for the sake of working OOTB. There is the same issue with flycheck-backends and the solution is exactly the same - if you want to run multiple flycheck backends you have to setq-local flycheck-backends in managed-mode hook.

The `flycheck` setting codes is also inside `lsp--auto-configure`, just saying.
Also, before `company-mode` configuration is separated with one is setting `company-backends` and the other is settings hooks  which is needed (and set) like a company compatible backend would do. The second one is using internal functional call.

Now, if we combine both of them, how can someone configure `company-mode` with auto configure disabled? The choice of using internal function seems not the bright to me.
Also, the code for setting completion (which is emacs built-in hooks) is normally set by major mode and completion assist packages, so it should not be in optional auto configure.

I don't understand why we cannot archive both OOTB experience working and do not violate user configuration if they request?
With my [proposed ](https://github.com/emacs-lsp/lsp-mode/pull/1967#discussion_r460617258), user can still keep their configuration working and OOTB experience can still also work.
If we doing something like merging both optional auto configure and required configure, at least mine (and maybe someone with similar configuration of disable auto configure) will be broken and needed to be updated.

yyoncho:
> By two hooks, do you mean the `lsp-configure-hook` and `lsp--auto-configure-hook`?

Yes.

> Now, if we combine both of them, how can someone configure company-mode with auto configure disabled? 

`M-x lsp-completion-mode` 

Either way, the "issue" with backends which is solvable with one line could be controlled by a changing `lsp-prefer-capf` to work like `lsp-diagnostics-package`. Which again won't change a thing because if the users that for example to combine company-capf with other backends will have to use `lsp-configure-hook`/`lsp-managed-mode/lsp-SERVER-ID-hook`. The benefit will be pretty much that it will change a line in your config from 

```
:hooks (lsp-managed-mode . (lambda () (setq-local company-backends '(foo bar))))
```
to 
```
(setq lsp-prefer-capf :none)
```

... and we are discussing that several days already.

kiennq:
> > Now, if we combine both of them, how can someone configure company-mode with auto configure disabled?
> 
> `M-x lsp-completion-mode`

And it will not pull in the auto configure stuff for `company-mode`?

Another note, if we still keep the `lsp-auto-configure` check in `lsp-configure-buffer`, then there will be no easy way to configure feature for dynamic registration feature, since the `lsp-configure-hook` will not be run in that case.

> Either way, the "issue" with backends which is solvable with one line could be controlled by a changing `lsp-prefer-capf` to work like `lsp-diagnostics-package`. Which again won't change a thing because if the users that for example to combine company-capf with other backends will have to use `lsp-configure-hook`/`lsp-managed-mode/lsp-SERVER-ID-hook`. The benefit will be pretty much that it will change a line in your config from
> 
> ```
> :hooks (lsp-managed-mode . (lambda () (setq-local company-backends '(foo bar))))
> ```
> 
> to
> 
> ```
> (setq lsp-prefer-capf :none)
> ```

Is that also require code change and it looks more complicated than what I propose, doesn't it?

Again, my proposal is:
- Set the initial value for `lsp-enable-*` to be `init`, user is advised to change it to `t` or `nil` depends on their likes (if user doesn't set that, we will set it based on auto configure or our default enablement states).
- Have the `lsp--auto-configure-hook` to run and check for `lsp-auto-configure`, it will set the value of `lsp-enable-*` (if user doesn't set it) to enable and user recommended options if auto configured, or set `lsp-enable-*` to our default states (`t` for `completion`, `nil` for diagnostic ....) if not.
- Have `lsp-configure-buffer` to check for `lsp-enable-*` and server capabilities only, not checking for `lsp-auto-configure`.

It will reduce the magic `lsp-auto-configure` check, and making `lsp-enable-*` more visible, and also it's straight forward.
Also, no user change is required. It actually works more closely with your initial 4 goals.

> ... and we are discussing that several days already.

It has been long :sad:
So, how about I will provide another PR to implement my proposal so we can continue with this PR? (or @ericdallo will probably implement already since this PR is quite close to what I propose already)

yyoncho:
@kiennq, the whole discussion is because you, one of the maintainers of lsp-mode does not want put one line in your config which has been our configuration story for 2 years and it is used by dozens of lsp-mode users. Please, in the future, indicate your intentions in the first comments.

> And it will not pull in the auto configure stuff for `company-mode`?

It will, and you will have to put that **one line** that in your configuration. 

> Another note, if we still keep the `lsp-auto-configure` check in `lsp-configure-buffer`, then there will be no easy way to configure feature for dynamic registration feature, since the `lsp-configure-hook` will not be run in that case.

Configure hook should run even if auto-configure is off.

> Is that also require code change and it looks more complicated than what I propose, doesn't it?

No, this is one change that we should do either way.

> So, how about I will provide another PR to implement my proposal so we can continue with this PR? (or @ericdallo will probably implement already since this PR is quite close to what I propose already)

I do not see benefits in that compared with what we have now or compared to what we have in dap-mode. Dap-mode approach is the simplest solution to implement and manage so I prefer sticking with it.

kiennq:
> @kiennq, the whole discussion is because you, one of the maintainers of lsp-mode does not want put one line in your config which has been our configuration story for 2 years and it is used by dozens of lsp-mode users. Please, in the future, indicate your intentions in the first comments.

Sorry for stubborn about this. I thought one of benefit of being the maintainer is that we can openly discuss about this, and also having `lsp-mode` doesn't break my config if it's reasonable. I'm an user of `lsp-mode`, after all.



> It will, and you will have to put that **one line** that in your configuration.

That's one line is required another change isn't it?
And I will have to wait until that implemented before I can easily change my config?
I **agree** that having that one line is nice too, but it's for another PR.


> Configure hook should run even if auto-configure is off.

With current `master` branch, `lsp-configure-buffer` is checking `lsp-auto-configure` before running `lsp-configure-hook`.

If we remove that check, then all of other optional features (like `lsp-enable-links`, `lsp-enable-imenu`, `lsp-enable-semantic-higlighting` ...) will be run since they're set to `t` by default, and user has to set them to off manually, which is quite a handful of configuration.

That leads to my proposal change.


> I do not see benefits in that compared with what we have now or compared to what we have in dap-mode. Dap-mode approach is the simplest solution to implement and manage so I prefer sticking with it.

I think my approach is not that different from `dap-mode`, both enable everything by default when auto configure mode is kicked in.
The different is that in `dap-mode` we don't have granular `lsp-enable-*` settings, so we can just simply turn on everything or not touching it.
With `lsp-mode`, we have granular settings and required changes to make both OOTB working while still respect user settings.

yyoncho:
kiennq writes:

> That's one line is required another change isn't it?

If we want to implement :none for lsp-prefer-capf it is a single liner.

> And I will have to wait until that implemented before I can easily change my config?

You could remove the reference to the private function using the
suggested code line with add-hook even without. This is much better than
using advice and achieves the same results.

> If we remove that check, then all of other optional features (like
> `lsp-enable-links`, `lsp-enable-imenu`,
> `lsp-enable-semantic-higlighting` ...) will be run since they're set
> to `t` by default, and user has to set them to off manually, which is
> quite a handful of configuration.

Check my upcomming review.

> I think my approach is not that different from `dap-mode`, both enable
> everything by default when auto configure mode is kicked in. The
> different is that in `dap-mode` we don't have granular `lsp-enable-*`
> settings, so we can just simply turn on everything or not touching it.
> With `lsp-mode`, we have granular settings and required changes to
> make both OOTB working while still respect user settings.

I want to narrow down to:

(setq lsp-enabled-features '(completion headerline modeline eldoc))

and remove all of the other flags.

ericdallo:
Nice, I like the idea of having one variable only like `(setq lsp-enabled-features '(completion headerline modeline eldoc))`, we can implement that on a next PR

yyoncho:
> Nice, I like the idea of having one variable only like `(setq lsp-enabled-features '(completion headerline modeline eldoc))`, we can implement that on a next PR

For that to happen we should unify all features to have a mode or some kind of unified structure.

lsp-completion.el
ericdallo: Good idea indeed

ericdallo:
@done @yyoncho and @kiennq, rolled back to use only `lsp-configure-hook` like the other features.

lsp-completion.el
```diff
@@ -0,0 +1,542 @@
...
+  :global nil
+  :lighter ""
+  (cond
+   (lsp-completion-mode
+    (setq-local completion-at-point-functions nil)
+    (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
+    (setq-local completion-category-defaults
+                (add-to-list 'completion-category-defaults '(lsp-capf (styles basic))))
+
+    (when (lsp--capability :completionProvider)
```
yyoncho, line 492: this is already checked in `--enable` -  `(lsp-feature? "textDocument/completion")` is equivalent to that check.

lsp-completion.el
```diff
@@ -0,0 +1,542 @@
...
+;;  LSP completion
+;;
+;;; Code:
+
+(require 'lsp-mode)
+
+(define-obsolete-variable-alias 'lsp-prefer-capf
+  'lsp-completion-prefer-capf  "lsp-mode 7.0.1")
+
+(defcustom lsp-completion-prefer-capf nil
```
yyoncho, line 29: can you rename that to lsp-completion-provider (or something like that)? We did similar transformation for lsp-prefer-flymake - it was renamed to lsp-diagnostics-package with :auto/:flymake/:flycheck/:none (t and nil are kept for backward compatibility). as values.

lsp-completion.el
```diff
@@ -0,0 +1,542 @@
...
+                  (setq-local lsp-inhibit-lsp-hooks t))
+                nil
+                t)
+      (add-hook 'company-after-completion-hook
+                (lambda (&rest _)
+                  (lsp-completion--capf-clear-cache)
+                  (setq-local lsp-inhibit-lsp-hooks nil))
+                nil
+                t))
+    (add-hook 'lsp-configure-hook #'lsp-completion--enable nil t)
```
yyoncho, line 517: no longer needed?

lsp-completion.el
```diff
@@ -0,0 +1,542 @@
...
+    (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t))
+   (t
+    (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
+    (setq-local completion-category-defaults
+                (cl-remove 'lsp-capf completion-category-defaults :key #'car))
+    (remove-hook 'lsp-configure-hook #'lsp-completion--enable t)
```
yyoncho, line 523: the same - if we are adding it via autoload

lsp-completion.el
```diff
@@ -0,0 +1,542 @@
...
+                   t)
+      (remove-hook 'company-after-completion-hook
+                   (lambda (&rest _)
+                     (lsp-completion--capf-clear-cache)
+                     (setq-local lsp-inhibit-lsp-hooks nil))
+                   t)))))
+
+;;;###autoload
+(add-hook 'lsp-configure-hook (lambda ()
+                                (when lsp-completion-enable
```
yyoncho, line 538: seems like it should be `(when (and lsp-auto-configure lsp-completion-enable) ...` if we want to have one place to turn everything off.

kiennq: Currently `lsp-configure-hook` is only run when `lsp-auto-configure` so it's not needed now. I think we can defer this until we fix how `lsp-auto-configure` affect `lsp-configure-hook`, that will be one swoop change

lsp-completion.el
ericdallo: Good idea

lsp-completion.el
ericdallo: I think if we remove the `remove-hook`, the user won't be able to disable this even turning off the mode, right?

yyoncho: This operation atm is NOP, because this is not removing the hook that is added in the end of the file.

ericdallo: Ok, it makes sense, done!

Issue #1209, opened 2019-12-01
when add to program language mode hoook, lsp-deferred conflicted with set company-backends function


phenix3443:
**Describe the bug**
i have some code to  config lsp-mode,company-lsp,company-lua to work together,  like follow:

(use-package lsp-mode
  :ensure t
  :commands (lsp lsp-deferred)
  :config

(use-package company-lua
  :ensure t
  :after (company)
  :config
  )

(defun set-company-backends-for-lua()
  "Set lua company backend."
  (setq-local company-backends '(
                                 (
                                  company-lua
                                  company-lsp
                                  company-keywords
                                  )
                                 company-gtags
                                 company-capf
                                 company-dabbrev-code
                                 company-files
                                 )))

(use-package lua-mode
  :ensure t
  :mode "\\.lua$"
  :interpreter "lua"
  :hook (
         (lua-mode . set-company-backends-for-lua)
         (lua-mode . lsp-deferred)  
         )
  :config
  )
)

when i open a lua file, the company-backends should be( i expecte):
`
((company-lua company-lsp company-keywords)
 company-gtags company-capf company-dabbrev-code company-files)
`
but in face it is 
` lisp
(company-lsp
 (company-lua company-lsp company-keywords)
 company-gtags company-dabbrev-code company-files)
`
in the case, the group backends i configured do not works. now my solution Is  that add lsp to lua-mode-hook intead of lsp-deferred, just as 

(use-package lua-mode
  :ensure t
  :mode "\\.lua$"
  :interpreter "lua"
  :hook (
         (lua-mode . set-company-backends-for-lua)
         (lua-mode . lsp)      ; keep this cofigure order, and do not use lsp-deferred
         )

  :config
  )

suggestion：delete the following code in lsp--auto-configure function in lsp-mode.el at line 5439:
`
      (add-to-list 'company-backends 'company-lsp)
 `

yyoncho:
@phenix3443 

What you have described is the desired behaviour. 

> suggestion：delete the following code in lsp--auto-configure function in lsp-mode.el at line 5439:
(add-to-list 'company-backends 'company-lsp)

This is what autoconfiguration do - it provides default configuration - deleting this line will break thousand of configs so this is no go. If the auto-configure logic does not work for you - you could disable it via setting `lsp-auto-configure` to `nil`.

PR #1983, opened 2020-07-31
Drop company-lsp support

ericdallo:


lsp-completion.el
```diff
@@ -495,19 +495,18 @@ Others: TRIGGER-CHARS"
...
-           (functionp 'company-lsp))
-      (progn
-        (company-mode 1)
-        (add-to-list 'company-backends 'company-lsp)
-        (setq-local company-backends (remove 'company-capf company-backends))))
+
+     ((and lsp-completion-provider
+           (fboundp 'company-mode)
+           (member 'company-lsp company-backends))
+      (message "Warning: company-lsp is not supported anymore, make sure `lsp-completion-provider` is configured correctly."))
```
yyoncho, line 502: can you use push company-capf and change the wording that we are going to use company-capf even if there was company-capf

lsp-completion.el
```diff
@@ -495,19 +495,20 @@ Others: TRIGGER-CHARS"
...
-           (functionp 'company-lsp))
-      (progn
-        (company-mode 1)
-        (add-to-list 'company-backends 'company-lsp)
-        (setq-local company-backends (remove 'company-capf company-backends))))
+
+     ((and lsp-completion-provider
+           (fboundp 'company-mode)
+           (member 'company-lsp company-backends))
+      (message "Warning: `company-lsp` is not supported anymore, using `company-capf` as the `lsp-completion-provider`.")
```
kiennq, line 502: `lsp-warn` ?

ericdallo: Nice, I was searching for that!

ericdallo: Done!

lsp-completion.el
```diff
@@ -495,19 +495,20 @@ Others: TRIGGER-CHARS"
...
     (cond
      ((equal lsp-completion-provider :none))
-     ((and (null lsp-completion-provider)
-           (functionp 'company-lsp))
-      (progn
-        (company-mode 1)
-        (add-to-list 'company-backends 'company-lsp)
-        (setq-local company-backends (remove 'company-capf company-backends))))
+
+     ((and lsp-completion-provider
```
kiennq, line 499: This will pass even with `lsp-completion-provider` equal to `:capf` and `(member 'company-lsp company-backends)`.
Shouldn't you combine two checks into one and check showing the warning when `(null lsp-completion-provider)`?

lsp-completion.el
```diff
@@ -495,19 +495,20 @@ Others: TRIGGER-CHARS"
...
+           (fboundp 'company-mode)
+           (member 'company-lsp company-backends))
+      (message "Warning: `company-lsp` is not supported anymore, using `company-capf` as the `lsp-completion-provider`.")
+      (company-mode 1)
+      (add-to-list 'company-backends 'company-capf))
```
kiennq, line 504: This will not likely work for default `company-backends` and with `clangd`.
Here is the default `company-backends`
```
(company-bbdb company-eclim company-semantic company-clang company-xcode company-cmake company-capf company-files
              (company-dabbrev-code company-gtags company-etags company-keywords)
              company-oddmuse company-dabbrev)
```

Since `company-clang` is before `company-capf`, the `add-to-list` will still keep the position of `company-capf` the same after `company-clang`. Thus `company-clang` will be used for `cpp` file instead.
Moreover, `add-to-list` modify the global `company-backends`, which is affecting user configuration for something not in `lsp-mode`.

Maybe you should just do `(setq company-backends '(company-capf))`

ericdallo: Ok, but this can remove any other `company-backend` right like `company-yasnippet`?

yyoncho: We have a separate issue for that, and most likely it will be fixed in company-mode. Lets get that in faster because ATM company-lsp is broken.

kiennq: At least I think we should make `company-backends` become local variable here instead of global `add-to-list`. But that's just nit pick, so feel free to go ahead.

kiennq: > Ok, but this can remove any other `company-backend` right like `company-yasnippet`?

Also, unless `company-yasnippet` is merged into `company-capf` as multi backends, it will be something like either snippet or  `capf`, which prevents the other backend from displayed.

yyoncho: > At least I think we should make `company-backends` become local variable here instead of global `add-to-list`. But that's just nit pick, so feel free to go ahead.

Why are you insisting on breaking users' configs?  We are running this for several months with no complaints. Furthermore, there were a number of complaints regarding the decision to use buffer-local for company-backends and company-lsp.

kiennq: > Why are you insisting on breaking users' configs? We are running this for several months with no complaints. Furthermore, there were a number of complaints regarding the decision to use buffer-local for company-backends and company-lsp.

Which is the complain about making `company-backends` buffer local?

I tried to search for `company-backends local` in this repo and found no related issue so far. Whereas I did find a complain about `company-backends` is changed. https://github.com/emacs-lsp/lsp-mode/issues/1209. That issue can be regarded as should just disable auto configure.

Let me point out that setting `(add-to-list 'company-backends 'company-capf)` is mostly wasteful, since by default `company-capf` is included in `company-backends`, but if user ever remove `company-capf` (I don't think anyone does this, thus no complain so far), then the `add-to-list` will modify the global `company-backends` and affect other files out of `lsp-mode`.
Purely from logic POV, we've tried to set local variable for most of configuration, why it's not good to make `company-backends` local variable here?

yyoncho: > 
> I tried to search for `company-backends local` in this repo and found no related issue so far. Whereas I did find a complain about `company-backends` is changed. #1209. That issue can be regarded as should just disable auto configure.

Here: https://github.com/emacs-lsp/lsp-mode/commit/7c55ba55af5c5e061285993c952074cbc52bfa67#commitcomment-31759080 . There were also in reddit, gitter, spacemacs channel, etc. So, when I implemented the initial capf support I decided not to create a local variable. Since then, there wasn't a single complaint in lsp-mode repo for that. 

> Purely from logic POV, we've tried to set local variable for most of configuration, why it's not good to make `company-backends` local variable here?

because it doesn't bring any value for that particular case and it will break users configs for no reason.

kiennq: I think you misunderstand my point.
Here is my point out:

> Let me point out that setting `(add-to-list 'company-backends 'company-capf)` is mostly wasteful, since by default `company-capf` is included in `company-backends`, but if user ever remove `company-capf` (I don't think anyone does this, thus no complain so far), then the `add-to-list` will modify the global `company-backends` and affect other files out of `lsp-mode`.

And the example you're giving is about setting `company-backends` to `company-lsp` only, and user is complain (maybe just asking why).
Here in this issue https://github.com/emacs-lsp/lsp-mode/issues/1209, it actually complain about **adding** into `company-backends`.

Just to make it clear, I said it would be nice to just set `(setq-local company-backends '(company-capf))`, and it should set `(make-local-variable 'company-backends)`.

Let me point out another thing again, currently `(add-to-list 'company-backends 'company-capf)` is no different from `(setq 'company-backends 'company-capf)` if user has no stand-alone `company-capf` in `company-backends`.
The reason why is because `lsp-completion-at-point-function` will try to complete everything and leave no later backend to run.

kiennq: Here is another example, if your `company-backends` is configured as `(company-nxml company-capf)`, and you visit and `xml` file with autoconfigure, you will get completions from `company-nxml` instead of `capf`.

yyoncho: > Here is another example, if your `company-backends` is configured as `(company-nxml company-capf)`, and you visit and `xml` file with autoconfigure, you will get completions from `company-nxml` instead of `capf`.

Indeed  - in this case, the user prefers company-nxml and most likely uses company-capf with M-x company-other-backend. What you are proposing will break user's config for no reason. I am locking the conversation.

kiennq: **Ok, while I'm hoping you can reconsider, I respect your choice and direction to go forward.**

The example I gave out before can be just user configures `company-mode` long time before to work with XML file, and now discover `lsp-mode`, however, the completion is not working OOTB. Same thing for `company-clang`.

Please also consider this as well
> Let me point out another thing again, currently (add-to-list 'company-backends 'company-capf) is no different from (setq 'company-backends 'company-capf) if user has no stand-alone company-capf in company-backends.
The reason why is because lsp-completion-at-point-function will try to complete everything and leave no later backend to run.

Well, again, thanks for your patient and opinions.

PR #1984, opened 2020-07-31
lsp-completion: move the rest of completion related code and change provider configure to align with diagnostic

kiennq:


lsp-completion.el
```diff
@@ -495,16 +610,16 @@ Others: TRIGGER-CHARS"
...
+           (fboundp 'company-mode))
       (when (or (null lsp-completion-provider)
                 (member 'company-lsp company-backends))
-        (lsp-warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
-
+        (lsp--warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
       (company-mode 1)
-      (add-to-list 'company-backends 'company-capf)))
-   
+      (setq-local company-backends '(company-capf)))
```
kiennq, line 619: I sneaky put this line here, hoping for @yyoncho to reconsider.
If it's still no good, I will revert this line.

ericdallo: I don't know if this works but can't we just remove company-capf from company-backends then add it again to make sure it will be in the last of the list and be used by lsp?

ericdallo: But this can be a user option to keep compny-capf before other backend like @yyoncho said it here: https://github.com/emacs-lsp/lsp-mode/pull/1983#discussion_r463753702

kiennq: With complex backends, I think the option is to use `lsp-completion-provider :none` and configure it manually.
For autoconfigure, it might be better to narrow down to `company-capf` only, to make sure that no other backend has the control and make the `capf` completion items not displaying.
For reference Eglot  also set to `company-capf` only https://github.com/joaotavora/eglot/blob/master/eglot.el#L1352

yyoncho: Please revert that change.

kiennq:
```suggestion
      (add-to-list 'company-backends 'company-capf)))
```

lsp-completion.el
```diff
@@ -495,16 +610,16 @@ Others: TRIGGER-CHARS"
...
+           (fboundp 'company-mode))
       (when (or (null lsp-completion-provider)
                 (member 'company-lsp company-backends))
-        (lsp-warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
-
+        (lsp--warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
       (company-mode 1)
       (add-to-list 'company-backends 'company-capf)))
-   
+     (t
```
ericdallo, line 620: I think this `t` is out of the `cond`?

lsp-completion.el
```diff
@@ -495,16 +610,16 @@ Others: TRIGGER-CHARS"
...
-
+     ((and (member lsp-completion-provider '(:capf nil t))
+           (fboundp 'company-mode))
       (when (or (null lsp-completion-provider)
                 (member 'company-lsp company-backends))
-        (lsp-warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
-
+        (lsp--warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
       (company-mode 1)
       (add-to-list 'company-backends 'company-capf)))
```
kiennq, line 619:
```suggestion
      (add-to-list 'company-backends 'company-capf))
```

lsp-completion.el
```diff
@@ -495,16 +610,16 @@ Others: TRIGGER-CHARS"
 
     (cond
      ((equal lsp-completion-provider :none))
-
-     ((fboundp 'company-mode)
-
+     ((and (member lsp-completion-provider '(:capf nil t))
+           (fboundp 'company-mode))
       (when (or (null lsp-completion-provider)
```
kiennq, line 615:
```suggestion
      (company-mode 1)
      (when (or (null lsp-completion-provider)
```

lsp-completion.el
```diff
@@ -495,16 +610,16 @@ Others: TRIGGER-CHARS"
...
                 (member 'company-lsp company-backends))
-        (lsp-warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
-
+        (lsp--warn "`company-lsp` is not supported anymore. Using `company-capf` as the `lsp-completion-provider`."))
       (company-mode 1)
```
kiennq, line 618:
```suggestion
```

PR #1990, opened 2020-08-02
Move Flycheck/Flymake to lsp-diagnostics.el

ericdallo:
* Move Flycheck integration to `lsp-checker.el`
* Move Flymake integration to `lsp-checker.el`
* Create `lsp-checker-mode` and `lsp-checker-enable`
* Define obsolete variable `lsp-diagnostic-package` -> `lsp-checker-provider`
* Define obsolete variable `lsp-flycheck-default-level` -> `lsp-checker-flycheck-default-level`
* Define obsolete variable `lsp-diagnostics-attributes` -> `lsp-checker-diagnostics-attributes`

### DIFF
--- a/lsp-checker.el
+++ b/lsp-checker.el
@@ -203,6 +203,74 @@ See https://github.com/emacs-lsp/lsp-mode."
                               (lsp-cpp-flycheck-clang-tidy-error-explainer e))
                              (t (flycheck-error-message e))))))
 
+
+;; Flymake integration
+
+(declare-function flymake-mode "ext:flymake")
+(declare-function flymake-make-diagnostic "ext:flymake")
+(declare-function flymake-diag-region "ext:flymake")
+
+(defvar flymake-diagnostic-functions)
+(defvar flymake-mode)
+(defvar-local lsp-checker--flymake-report-fn nil)
+
+(defun lsp-checker--flymake-setup ()
+  "Setup flymake."
+  (setq lsp-checker--flymake-report-fn nil)
+  (flymake-mode 1)
+  (add-hook 'flymake-diagnostic-functions 'lsp-checker--flymake-backend nil t)
+  (add-hook 'lsp-diagnostics-updated-hook 'lsp-checker--flymake-after-diagnostics nil t))
+
+(defun lsp-checker--flymake-after-diagnostics ()
+  "Handler for `lsp-diagnostics-updated-hook'."
+  (cond
+   ((and lsp-checker--flymake-report-fn flymake-mode)
+    (lsp-checker--flymake-update-diagnostics))
+   ((not flymake-mode)
+    (setq lsp-checker--flymake-report-fn nil))))
+
+(defun lsp-checker--flymake-backend (report-fn &rest _args)
+  "Flymake backend using REPORT-FN."
+  (let ((first-run (null lsp-checker--flymake-report-fn)))
+    (setq lsp-checker--flymake-report-fn report-fn)
+    (when first-run
+      (lsp-checker--flymake-update-diagnostics))))
+
+(defun lsp-checker--flymake-update-diagnostics ()
+  "Report new diagnostics to flymake."
+  (funcall lsp-checker--flymake-report-fn
+           (-some->> (lsp-diagnostics t)
+             (gethash (lsp--fix-path-casing buffer-file-name))
+             (--map (-let* (((&Diagnostic :message :severity?
+                                          :range (range &as &Range
+                                                        :start (&Position :line start-line :character)
+                                                        :end (&Position :line end-line))) it)
+                            ((start . end) (lsp--range-to-region range)))
+                      (when (= start end)
+                        (if-let ((region (flymake-diag-region (current-buffer)
+                                                              (1+ start-line)
+                                                              character)))
+                            (setq start (car region)
+                                  end (cdr region))
+                          (lsp-save-restriction-and-excursion
+                            (goto-char (point-min))
+                            (setq start (point-at-bol (1+ start-line))
+                                  end (point-at-eol (1+ end-line))))))
+                      (flymake-make-diagnostic (current-buffer)
+                                               start
+                                               end
+                                               (cl-case severity?
+                                                 (1 :error)
+                                                 (2 :warning)
+                                                 (t :note))
+                                               message))))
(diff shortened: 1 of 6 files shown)


lsp-checker.el
```diff
@@ -0,0 +1,319 @@
...
+      (lsp--warn "Unable to autoconfigure flycheck/flymake. The diagnostics won't be rendered.")))
+
+    (add-hook 'lsp-unconfigure-hook #'lsp-checker--disable nil t))
+   (t
+    (remove-hook 'lsp-unconfigure-hook #'lsp-checker--disable t))))
+
+;;;###autoload
+(add-hook 'lsp-configure-hook (lambda ()
+                                (when (and lsp-auto-configure
+                                           lsp-checker-enable)
```
kiennq, line 315: this has been checked in `lsp-checker--enable`

ericdallo:
I think segregating lsp-mode by features, like modeline, headerline, completion instead of spec names, could it be better, imagine breadcrumb use diagnostics and flycheck too, so breadcrumb and flycheck would be in the same lsp-diagnostics? I think it would be more confusing.

yyoncho:
ATM the majority of the features/functions/packages are named after either the protocol or emacs term/package (lens, completion, modeline, headerline, workspace-symbol, signature, eldoc, imenu). I might be wrong but this seems like the first time we are naming something on a generic term. For me considering that we have `lsp-modeline-diagnostics-scope`  `lsp-diagnostics-package` make more sense than `lsp-checker-package`.

ericdallo:
@yyoncho I understand your point, I renamed to `lsp-diagnostics.el`, in the future after most features are renamed/moved, we can rethink the names, segregating to files I think it's already a great improvement.

lsp-diagnostics.el
```diff
@@ -0,0 +1,315 @@
...
+     ((or
+       (and (eq lsp-diagnostics-provider :auto)
+            (functionp 'flycheck-mode))
+       (and (eq lsp-diagnostics-provider :flycheck)
+            (or (functionp 'flycheck-mode)
+                (user-error "The lsp-diagnostics-provider is set to :flycheck but flycheck is not installed?")))
+       ;; legacy
+       (null lsp-diagnostics-provider))
+      (lsp-diagnostics--flycheck-enable))
+     ((and (not (version< emacs-version "26.1"))
```
kiennq, line 296: Can this version check be removed? I think we all upgrade to `26.x` now for `lsp-mode`

ericdallo: Done!

lsp-diagnostics.el
```diff
@@ -0,0 +1,314 @@
...
+           ;; function. See https://github.com/joaotavora/eglot/issues/159
+           :region (cons (point-min) (point-max))))
+
+

+
+;;;###autoload
+(defun lsp-diagnostics--enable ()
+  "Enable LSP checker support."
+  (unless (eq lsp-diagnostics-provider :none)
```
kiennq, line 271: nit: since the `:none` configuration is acceptable (user provide their own configuration), do you need to check this?
Shouldn't this check changed to something like `(when (member lsp-dianostics-provider '(:auto :none ....)))`?

Note that the mode is not turn on is also fine, but I can imagine that if user bind a key map for this mode and not have this turn on may cause problem.

ericdallo: You mean only enable the mode if it was provided a known provider like `:auto`, `:none`, `:flycheck` or `:flymake`?

kiennq: Yes, `t` and `nil` too. Since those are configurations that don't trigger `lsp--warn`

yyoncho:
> in the future after most features are renamed/moved, we can rethink the names, segregating to files I think it's already a great improvement.

Yes. Let's also have a table with what are we going to extract upfront and with the target names/structure.

lsp-diagnostics.el
```diff
@@ -0,0 +1,314 @@
...
+;;; Commentary:
+;;
+;;  LSP diagnostics integration
+;;
+;;; Code:
+
+(require 'lsp-mode)
+
+(define-obsolete-variable-alias 'lsp-prefer-flymake
+  'lsp-diagnostics-provider  "lsp-mode 6.2")
```
yyoncho, line 27: IMO this could be already removed.

lsp-diagnostics.el
```diff
@@ -0,0 +1,314 @@
...
+
+(define-obsolete-variable-alias 'lsp-diagnostic-package
+  'lsp-diagnostics-provider  "lsp-mode 7.0.1")
```
yyoncho, line 30: This one I am not 100% it will work:

e. g. 
```
(use-package lsp-mode
   :config (setq lsp-diagnostics-package 'foobar)
```

The setq will be executed after lsp-mode.el is loaded but at this point lsp-diagnostics.el won't be loaded and the alias won't be loaded. I think the result will be that the `provider` wont be set. I see several options to avoid that:

1. Move the defalias temporary(until we drop the alias) in lsp-mode.el
2. Keep the original name and rename after several months/releases.
?

ericdallo: yeah, but I think the problem is bigger than that, even moving to `lsp-mode.el`, `lsp-diagnostics.el` won't be loaded when this `(setq)` run, right?

yyoncho: > yeah, but I think the problem is bigger than that, even moving to `lsp-mode.el`, `lsp-diagnostics.el` won't be loaded when this `(setq)` run, right?

Yes, but setting symbol before the defining package is loaded is fine. The issue when you set an alias of a symbol that is not loaded.

kiennq: Another options is to use autoload so this will be loaded even before package is loaded.

kiennq: Some thing like
``` elisp
;;;###autoload
(define-obsolete-variable-alias 'a 'b "now")
```

yyoncho: This will most likely work.

ericdallo: Good idea, I'll do that

kiennq:
I approved the changed. Another nit pick is that we currently have quite a few `define-obsolete-variable-alias` defined in sub module like `lsp-modeline`, `lsp-completion`, can you also fix them?

ericdallo:
Yes, it makes sense to do that, I'll fix in another PR

PR #1991, opened 2020-08-03
Migrate to new form of ‘if-let’ and ‘when-let’.

phst:
The (SYM VAL) form exists only for backward compatibility, and the ((SYM VAL))
form is more generic and consistent with ‘let’.


yyoncho:
Thank you!

PR #1992, opened 2020-08-03
Add a few ‘eval-and-compile’ forms.

phst:
Depending whether and how these files are byte-compiled, the definitions need
to be available during compile time as well.


yyoncho:
Thank you.

How do you reproduce the issue? I mean, we do not catch that with the CI were we byte-compile the files. The macro has to access those during macro expansion time, but still, it somehow works without the eval-and-compile forms.

PR #1997, opened 2020-08-05
Custom modeline code actions

### LINKED ISSUE #1996: make lsp-modeline-code-actions less noisy
It's somewhat interesting to know that code actions are available at the current point,
so the lightbulb icon is nice,
but I do not care what the first of a list of probably more than one is
and find this text quite noisy in the very limited modeline space

I haven't found a way to customize what is shown.

A way to disable the text would be great.

ericdallo:
Closes #1996

- Add `lsp-modeline-code-actions-segments`

image


lsp-modeline.el
```diff
@@ -60,41 +69,50 @@
...
-          (string (if single-action?
-                      (concat " " icon " " first-action-string " ")
-                    (concat " " icon " " first-action-string " "
-                            (propertize (concat "(" (number-to-string (1- (seq-length actions))) " more)")
-                                        'display `((height 0.9))
-                                        'face 'lsp-modeline-code-actions-face)
-                            " "))))
-    (propertize string
+          (built-string (concat " "
+                                (lsp-modeline--build-code-actions-segments actions)
```
kiennq, line 110: If this returns empty, we will always display two whitespaces on mode line in place of code actions mode line, which will make it looking not nice.
I think we should change to check to string empty before adding prefix and suffix spaces

lsp-modeline.el
```diff
@@ -60,41 +69,50 @@
...
+                                " ")))
```
kiennq, line 111: I always wonder why the spaces of code action mode line seems to be a little bit weird, turn out it's because of suffix white space.
image
Does it make sense to append the prefix space only? The next mode line components should add their prefix space already (like in this case, diagnostic)

lsp-modeline.el
```diff
@@ -60,41 +69,50 @@
...
+    (propertize built-string
```
kiennq, line 112: `propertize` copy the whole string, since this code is in hot path, I think you can change this propertize into `set-text-properties`.

lsp-modeline.el
ericdallo: It makes sense, each modeline thing should add only the prefix, otherwise, this will happen, thanks for noticing it!

ericdallo: `set-text-properties` replaces all the properties, `add-text-properties` works

lsp-modeline.el
```diff
@@ -61,49 +70,60 @@
...
+                                         "select from multiple code actions"))
+                               'mouse-face 'mode-line-highlight
+                               'local-map (make-mode-line-mouse-map
+                                           'mouse-1 (lambda ()
+                                                      (interactive)
+                                                      (if single-action?
+                                                          (lsp-execute-code-action (lsp-seq-first actions))
+                                                        (lsp-execute-code-action (lsp--select-action actions))))))
+                         built-string)
+    (unless (string= "" built-string)
```
kiennq, line 125: It seems that the prefix space is not prepend? Or did I miss something?

ericdallo: Yeah, it was missing, done @kiennq !

PR #1998, opened 2020-08-05
Rework signature

yyoncho:
- use post-command-hook instead of lsp-idle-hook
- change the rendering so it can be displayed in posframe

----

#


kiennq:
nit: Why using `post-command-hook` instead of on idle hook?

yyoncho:
> it: Why using `post-command-hook` instead of on idle hook?

Because it provides a better experience. The signature help is closer to completion. E. g. "you type and it updates", not "you type, it stays the same, you stop for 0.5s - it updates"

PR #1999, opened 2020-08-05
Signature posframe support

yyoncho:
----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4873,6 +4873,30 @@ RENDER-ALL - nil if only the signature should be rendered."
           (lv-message "%s" message)))
     (lv-delete-window)))
 
+(declare-function posframe-show "ext:posframe")
+(declare-function posframe-hide "ext:posframe")
+(declare-function posframe-poshandler-point-bottom-left-corner-upward "ext:posframe")
+
+(defvar lsp-signature-posframe-params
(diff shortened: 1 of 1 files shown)


muirdm:
Looks cool. The posframe seemed to bounce around with point - can it stay anchored to the start of the function call? I also had trouble customizing the colors on the posframe. I customized `tooltip` face, but it didn't really work.

yyoncho:
> - can it stay anchored to the start of the function call?

The server does not provide that data. VScode also moves the popup.

> I also had trouble customizing the colors on the posframe. I customized tooltip face, but it didn't really work.

I will expose the posframe params before merging.

kiennq:
@yyoncho  Is there any other problem for this?

yyoncho:
> @yyoncho Is there any other problem for this?

I think that I was planning to add some customization which is not possible ATM.

FelipeLema:
is this posframe displayed on request? on timeout? both? either?

yyoncho:
> is this posframe displayed on request? on timeout? both? either?

When you press signatureHelp triggerChar (e. g. `(`)/

yyoncho:
I tested the latest version and it seems like it is working fine, except the fact that you cannot scroll the signature window screen(which will be addressed).

yyoncho:
> > except the fact that you cannot scroll the signature window screen
> 
> @yyoncho Is there anywhere tracking that work?

@benbot afaik no

PR #2003, opened 2020-08-06
Add configuration example

kentbull:
To help new users learn how to configure lsp-mode a simple example like this goes a long way.

### DIFF
--- a/docs/page/installation.md
+++ b/docs/page/installation.md
@@ -105,6 +105,21 @@ To defer LSP server startup (and DidOpen notifications) until the buffer is visi
     :commands (lsp lsp-deferred))
 ```
 
+#### use-package configuration
+
+Setting configuration options is easy:
+
+```
+(use-package lsp-mode
+             ...other options...
+             :config
+             (setq lsp-enable-on-type-formatting nil
+                   lsp-enable-file-watchers nil
+                   lsp-enable-xref t
+                   lsp-enable-completion-at-point t
+                   lsp-prefer-flymake nil))
+```
+
 ## Install a language server
 
 For instructions on how to install a server for your language, check the [available supported servers](./languages.md).


docs/page/installation.md
```diff
@@ -105,6 +105,21 @@ To defer LSP server startup (and DidOpen notifications) until the buffer is visi
     :commands (lsp lsp-deferred))
 ```
 
+#### use-package configuration
```
ericdallo, line 108: We already have a section called `use-package` above, what do you think about:
```suggestion
### Example configuration
```

docs/page/installation.md
```diff
@@ -105,6 +105,21 @@ To defer LSP server startup (and DidOpen notifications) until the buffer is visi
...
+Setting configuration options is easy:
+
+```
+(use-package lsp-mode
+             ...other options...
+             :config
+             (setq lsp-enable-on-type-formatting nil
+                   lsp-enable-file-watchers nil
+                   lsp-enable-xref t
+                   lsp-enable-completion-at-point t
```
kiennq, line 119: These should be set before `lsp-mode`, so you may want to move these into `:init` of `use-package`

ericdallo:
@kentbull could you take a look, please?

docs/page/installation.md
raxod502: Really? If so, that seems bad. My understanding was that the variables only take effect once `lsp-mode` is enabled, not when `lsp-mode.el` is _loaded_. Putting the configuration in `:config` will set the variables in between when `lsp-mode.el` has finished loading and when `lsp-mode` is enabled, even if it's triggered from an autoload.

kentbull:
Yes, I'll take a look

docs/page/installation.md
nbfalcon: @raxod502: What do you mean by enabled? I can imagine that some parts of lsp check for the truthyness of variables and enable feature-xyz (as is the case of `lsp-pyright`'s `lsp-pyright-multi-root` variable) if they are t. They are still defined using defcustom, which doesn't change the value if already set for some other reason. How would you implement such a thing in a way that reflects the variable's state live, given that for things like company hooks have to set-up? (Aside from variable watchers, but that would be really hacky)

raxod502: > which doesn't change the value if already set for some other reason

This is so that you can set the value of a variable before loading the library. It's the same for `defvar`. There is support in Custom for running code when a user option is changed, which is intended to handle cases like this, but it kind of doesn't work because it assumes you change the user option through Custom rather than directly.

> How would you implement such a thing

It's true sometimes it cannot be avoided, so maybe my comment was misplaced. But I think reducing the number of cases where this happens is good, because it makes the behavior of the package less complex to understand. It's often possible to work around the problem by reframing what value the variable contains. For example, I did a grep in this repository and the first match for a dynamically initialized variable was:

```elisp
(defcustom lsp-fsharp-server-install-dir (f-join lsp-server-install-dir "fsautocomplete/")
  "Install directory for fsautocomplete server.
The slash is expected at the end."
  :group 'lsp-fsharp
  :risky t
  :type 'directory
```

I might change it to:

```elisp
(defcustom lsp-fsharp-server-install-dir "fsautocomplete/"
  "Install directory for fsautocomplete server.
If this is a relative path, it's expanded relative to
`lsp-server-install-dir'.
The slash is expected at the end."
  :group 'lsp-fsharp
  :risky t
  :type 'directory
```

and then update the usage to add the `f-join` there. This has the advantage of retaining all the same functionality, but there is no longer a surprising behavior if you happen to accidentally load `lsp-fsharp` before you set `lsp-server-install-dir`.

Anyway, like I said this can't always be done, and it is true that sometimes this kind of workaround will make things less clear, so your initial suggestion to move the variable settings is definitely a good one. Looking at the referenced variables, it does seem to me like none of them could be adjusted like I'm suggesting (without some nontrivial inconvenience), so sorry for presuming.

ericdallo:
Ping @kentbull

kentbull:
Thanks for the ping, I lost track of this. I will re-open a PR if this still adds value.

PR #2005, opened 2020-08-06
Faces for rust-analyzer inlay hints

fredmorcos:
What the title says, make the faces stand out and are now reasonably customizable by the user.

### DIFF
--- a/lsp-rust.el
+++ b/lsp-rust.el
@@ -587,6 +587,60 @@ The command should include `--message=format=json` or similar option."
   :group 'lsp-rust
   :package-version '(lsp-mode . "7.0"))
 
+(defface lsp-rust-analyzer-inlay-type-face
+  '((t :inherit lsp-rust-analyzer-inlay-face))
+  "Face for inlay type hints (e.g. inferred variable types)."
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-rust-analyzer-inlay-type-space-format "%s"
+  "Format string for spacing around variable inlays (not part of the inlay face)."
+  :type '(string :tag "String")
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-rust-analyzer-inlay-type-format ": %s"
+  "Format string for variable inlays (part of the inlay face)."
+  :type '(string :tag "String")
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defface lsp-rust-analyzer-inlay-param-face
+  '((t :inherit lsp-rust-analyzer-inlay-face))
+  "Face for inlay parameter hints (e.g. function parameter names at call-site)."
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-rust-analyzer-inlay-param-space-format "%s "
+  "Format string for spacing around parameter inlays (not part of the inlay face)."
+  :type '(string :tag "String")
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-rust-analyzer-inlay-param-format "%s:"
+  "Format string for parameter inlays (part of the inlay face)."
+  :type '(string :tag "String")
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defface lsp-rust-analyzer-inlay-chain-face
+  '((t :inherit lsp-rust-analyzer-inlay-face))
+  "Face for inlay chaining hints (e.g. inferred chain intermediate types)."
+  :group 'lsp-rust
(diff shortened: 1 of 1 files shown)


yyoncho:
Looks good to me except for the fact that `:` are missing (it might be my theme
though). Is that desired effect?

yyoncho:
Screenshot at 2020-08-06 19-58-09

fredmorcos:
Yes the `:` are missing on purpose. I thought having the hint highlighted and being a space away from the variable/chain/etc meant it didn't need the `:` anymore. Would you prefer if I add them back?

yyoncho:
I am just sharing an observation. It is up to the rust users/maintainers to make a call. To me it looks fine and I guess that `:` is just a noise in that case.

yyoncho:
cc @flodiebold

flodiebold:
I would prefer keeping the `:`. (And I actually like the old face.) That's how we display them in VSCode as well.

fredmorcos:
Alright, tomorrow I'll add `-text-prefix`, `-text-postfix`, `-space-prefix` and `-space-postfix` options for each.

lsp-rust.el
```diff
@@ -587,6 +587,21 @@ The command should include `--message=format=json` or similar option."
 
 (defvar-local lsp-rust-analyzer-inlay-hints-timer nil)
 
+(defface lsp-rust-analyzer-inlay-type-face
```
kiennq, line 590: Shouldn't these inherit `lsp-rust-analyzer-inlay-face` instead of defining a completely new one?
So user will not stumped upon breaking visual change when update `lsp-mode`

kiennq:
I also in favor of keeping `:`, it looks more like rust syntax

lsp-rust.el
brotzeit: Maybe we should also document how to change faces. This is a common problem for new emacs users.

fredmorcos:
I pushed the changes. I think this PR should now be satisfactory to everybody by adding enough options for the user to customize what they would like. The defaults keep the current settings.

fredmorcos:
Example:
image

lsp-rust.el
```diff
@@ -587,6 +581,111 @@ The command should include `--message=format=json` or similar option."
...
+  "The face to use for the Rust Analyzer inlays."
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.0"))
+
+(defface lsp-rust-analyzer-inlay-type-face
+  '((t :inherit lsp-rust-analyzer-inlay-face))
+  "Face for inlay type hints (e.g. inferred types)."
+  :group 'lsp-rust)
+
+(defcustom lsp-rust-analyzer-inlay-type-space-prefix nil
```
kiennq, line 595: Can't `-space-prefix` and `-prefix` be merged together? Same thing for `-posfix` too.
Merging them still provide same functionality though.

And it avoids providing multiple customization that do the same thing.

kiennq: Alternatively, to save on number of customize options, we can even provide something like `..-inlay-type-format`, that's just a string to pass to `(format lsp-rust-analyzer-inlay-type-format type)`.
For example, set the format to `": %s"` and we will have what it looks like right now.
Just an alternative suggestion though.

fredmorcos: `-space-prefix` and `-prefix` don't do the same thing, as explained in their documentation strings, one is part of the hint (and uses the same face as the hint) and the other uses the default face. In the screenshot, the chain inlay hints have a space before them, then an arrow and a space inside of them.

I like the idea of the format string, but I want to have two of them for each kind of inlay, one for the inlay text and one for the spacing around the inlay text.

Would that be satisfactory to everybody?

kiennq: > I like the idea of the format string, but I want to have two of them for each kind of inlay, one for the inlay text and one for the spacing around the inlay text.

Great. If you want to use that, then the format can be changed to something like `" %s %s"`, with the first phrase is `-prefix`, and the second phrase is actual hint. If you propertize the string before pass it to `format` it will retain format for each phrase

fredmorcos: I'm not following, how would that work for having both prefixes and postfixes? Am I missing something? Could you please provide an example that handles all 4 cases (space-prefix, prefix, postfix, space-postfix) as shown in the screenshot [here](https://github.com/emacs-lsp/lsp-mode/pull/2005#issuecomment-670585154)?

kiennq: Ah, sorry, I kinda forgot my previous suggestion :smile:.
You're right, if you want to maintain different face for `-space-prefix` and `-prefix`, it will require two formats instead of one.

fredmorcos:
Inlays now have only two customizations each, one format string for spacing and one format string for the inlay text.

yyoncho:
@fredmorcos thank you for contributing to `lsp-mode` and congrats for your first `lsp-mode` PR! :tada:

PR #2006, opened 2020-08-06
[infra] Add a .git-blame-ignore-revs file

leungbk:
Git 2.23 added the blame.ignoreRevsFile option. After running the
configuration command once, `magit-blame` should no longer display the
ignored commits.


yyoncho:
Nice, I was unaware of that feature.

Issue #2008, opened 2020-08-06
json-serialize does not encode list like json-encode


wyuenho:
**Describe the bug**
In [`lsp--json-serialize`](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L2501-L2509), lsp-mode naively replaces `json-encode` with `json-serialize` and ended up giving surprising results when using `lsp-configuration-section` to configure client `:initialization-options`. According to the docstring, `json-serialize` and `json-parse-*` only process vectors, so if some of the initialization options are lists, an error like the following will result:

```
Debugger entered--Lisp error: (wrong-type-argument symbolp "foo")
   json-serialize(("foo" "bar")) 
```

**To Reproduce**
```
(defvar some-lsp-server-test '("foo" "bar))

(lsp-register-custom-settings
  '(("some-lsp-server.test" some-lsp-server-test)))

...
;; Client config

:initialization-options (lambda () (lsp-configuration-section "some-lsp-server"))
...
```

**Expected behavior**
Options that are lists should be encoded without trouble just like `json-encode`.

kiennq:
This is currently **by design** for `json-serialize`.

We can try to automatically detect and convert list to vector but, I think we want to squeeze every possible perf for json serialization in `lsp-mode` too, thus there's no such automatic conversion/checking yet.

wyuenho:
Are you telling me lsp-mode expects client configs to use vectors instead of lists on purpose?

kiennq:
Yes. Is there anything wrong with that?
Unless `json-serialize` can support list encoding upstream too.

yyoncho:
There were multiple issues related to old json.el implementations and one of them was list serialization. They are fixed/removed in json.c - one of the things is that `list` is always json map and the only structure that represents json array is elisp array. It does not make sense to reintroduce the heuristic applied by json.el in lsp-mode code to make json.c and json.el comparable.

wyuenho:
> Yes. Is there anything wrong with that?

[...]

PR #2012, opened 2020-08-08
Use widget for lsp-string-vector

wyuenho:
This PR solves a pain point identified in #2008. Currently lsp-mode expects client `defcustom`s to use vectors instead of lists due to limitations in `json-serialize`, but there's no a customization type similar to `repeat` that will return a vector. This forces all `lsp-string-vector` `defcustom`s to simply provide a sexp field. This defeats the purpose of `defcustom`. This PR rectifies this problem by implementing 2 new customization types - `lsp-editable-vector` and `lsp-repeatable-vector`, and change the definition of `lsp-string-vector` to use `(lsp-repeatable-vector string)` instead. This change is completely backward compatible. No change to current clients required.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1059,15 +1059,15 @@ every element of it is of type string, else nil."
    (vectorp candidate)
    (seq-every-p #'stringp candidate)))
 
-(make-obsolete 'lsp--string-vector-p nil "7.0.2")
+(make-obsolete 'lsp--string-vector-p nil "7.1")
 
-(defun lsp-editable-vector-match (widget value)
+(defun lsp--editable-vector-match (widget value)
   "Function for `lsp-editable-vector' :match."
   ;; Value must be a list or a vector and all the members must match the type.
   (and (or (listp value) (vectorp value))
-       (length (cdr (lsp-editable-vector-match-inline widget value)))))
+       (length (cdr (lsp--editable-vector-match-inline widget value)))))
 
-(defun lsp-editable-vector-match-inline (widget value)
+(defun lsp--editable-vector-match-inline (widget value)
   "Value for `lsp-editable-vector' :match-inline."
   (let ((type (nth 0 (widget-get widget :args)))
         (ok t)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1059,14 +1059,61 @@ every element of it is of type string, else nil."
    (vectorp candidate)
    (seq-every-p #'stringp candidate)))
 
+(make-obsolete 'lsp--string-vector-p nil "7.0.2")
```
kiennq, line 1062: The version should be `7.1` now

lsp-mode.el
```diff
@@ -1059,14 +1059,61 @@ every element of it is of type string, else nil."
...
+      (apply 'vector internal-value)
+    internal-value))
+
+(defun lsp-editable-vector-value-to-internal (_widget external-value)
+  "Convert the external vector value to a list."
+  (if (vectorp external-value)
+      (append external-value nil)
+    external-value))
+
+(define-widget 'lsp-repeatable-vector 'lsp-editable-vector
```
kiennq, line 1105: Except `lsp-repeatable-vector`, I think other related functions, widget can be private, WDYT? Can you add double dash to their name?

lsp-mode.el
yyoncho: Why? 

@ericdallo as a side note, I think that we should bump the version in lsp-mode.el after the release, not before it to avoid that kind of discussions.

wyuenho: Why as in?

lsp-mode.el
wyuenho: No, this type is very much meant for consumption by the clients. Conceivably there will be configurations that look like `[1, 3, 5]`, if that's the case client authors can just use `(lsp-repeatable-vector integer)` as the defcustom type instead of asking us to make a `lsp-integer-vector` or some weird thing. `lsp-string-vector` is only preserved for backward compatibility. I actually think we should deprecate that.

lsp-mode.el
yyoncho: (my why was for @kiennq )

kiennq: I saw 7.0.1 has been release, and `lsp-modeline` has adopt 7.1 already

lsp-mode.el
kiennq: That why I said except `lsp-repeatable-vector`. Now everything you added has public scope naming.

lsp-mode.el
wyuenho: So I should change to 7.1?

kiennq: Alternatively, I think you can define a `alpha` version for anything that's in development, then change all of them to exact version number when release. Contributor can know exact version (`alpha`) then

wyuenho: Who would change the alpha to a concrete version?

kiennq: The one who release new version? Mostly @yyoncho and @ericdallo. Using `alpha` is just my proposal to they though

wyuenho: I've changed to 7.1. Please update it if necessary before release.

lsp-mode.el
```diff
@@ -1059,14 +1059,65 @@ every element of it is of type string, else nil."
    (vectorp candidate)
    (seq-every-p #'stringp candidate)))
 
+(make-obsolete 'lsp--string-vector-p nil "7.1")
```
kiennq, line 1062:
```suggestion
(make-obsolete 'lsp--string-vector-p nil "lsp-mode 7.1")
```

lsp-mode.el
```diff
@@ -1059,14 +1059,65 @@ every element of it is of type string, else nil."
...
+serialize as a JSON array of strings.
+
+Deprecated. Use `lsp-repeatable-vector' instead. "
   :offset 4
   :tag "Vector"
-  :type '(restricted-sexp
-          :match-alternatives (lsp--string-vector-p)))
+  :type '(lsp-repeatable-vector string))
+
+(make-obsolete 'lsp-string-vector nil "7.1")
```
kiennq, line 1120:
```suggestion
(make-obsolete 'lsp-string-vector nil "lsp-mode 7.1")
```

kiennq:
I approved with a few nit comment about the prefix package name for `make-obsolete`

wyuenho:
Done

yyoncho:
thank you!

PR #2013, opened 2020-08-08
Filter symbols by kind before being used to create imenu index

wyuenho:
Currently `lsp--imenu-create-index` simply dumps all the symbols the language server returns to imenu, which is insane because often the symbols return useless things like all the variables, imports or even strings in the response. There is absolutely no reason to display the location of every string in imenu. This PR implements a user configurable filter to reduce noise in imenu, and make the lsp imenu behave a little closer to the default imenu-create-index functions in the various major modes. Tested with lsp-jedi.


lsp-mode.el
```diff
@@ -5810,16 +5840,18 @@ an alist
...
-    (not (eq (->> location
-                  (lsp:location-uri)
-                  (lsp--uri-to-path)
-                  (find-buffer-visiting))
-             (current-buffer)))))
+(lsp-defun lsp--symbol-filter ((&SymbolInformation :kind :location))
+  "Determine if SYM is for the current document and is to be shown."
+  ;; It's a SymbolInformation or DocumentSymbol, which is always in the
+  ;; current buffer file.
+  (or (and (not (null lsp-imenu-index-symbol-kinds))
```
kiennq, line 5847: Shouldn't this be `and` instead of `or`?
If this is `or` how can it filter more than the old function?

kiennq: Also you can replace `(not (null a)) ` with just `a`

wyuenho: > Shouldn't this be and instead of or?

No :)

lsp-mode.el
```diff
@@ -715,6 +715,36 @@ are determined by the index of the element."
                          (const kind)))
   :group 'lsp-imenu)
 
+(defcustom lsp-imenu-index-symbol-kinds '(5 6 7 8 9 10 11 12 23)
```
yyoncho, line 718: I don't like the fact that the users that don't use customize interface will have to use numbers instead of symbols here. I will change the code that we use for generating stubs to generate vector lookups for enum properties to avoid the need to manually define them and then you could use that to convert number -> symbol or (even generate the defcustom type?).

lsp-mode.el
```diff
@@ -715,6 +715,36 @@ are determined by the index of the element."
...
+  "Which symbol kinds to show in imenu."
+  :type '(repeat (choice (const :tag  "File" 1)
```
kiennq, line 720: Please check `lsp-protocol`, we have enum value for those number now.
Ex: `lsp/symbol-kind-file`.
Please use them instead of number like this

lsp-mode.el
yyoncho: Here it is the PR - https://github.com/emacs-lsp/lsp-mode/pull/2014/

wyuenho:
The problem with not using numbers is now when matching with `memql`, I'll now have to pre-generate a set before I use them for constant time comparison. There are a number of ways we can do this, one is to provide a custom `:set` in the defcustom, another is to pre-generate a reverse lookup table in #2014, non is as elegant as this. If somebody who doesn't use customize to configure lsp-mode is to configure this, it's as easy as doing something like `(add-to-list 'lsp-imenu-index-symbol-kinds lsp/symbol-kind-file)`. This way lsp-mode doesn't have to do anything.

yyoncho:
> If somebody who doesn't use customize to configure lsp-mode is to configure this, it's as easy as doing something like `(add-to-list 'lsp-imenu-index-symbol-kinds lsp/symbol-kind-file)`.

This looks acceptable too and will make https://github.com/emacs-lsp/lsp-mode/pull/2014 redundant.

lsp-mode.el
```diff
@@ -715,6 +715,37 @@ are determined by the index of the element."
                          (const kind)))
   :group 'lsp-imenu)
 
+(defcustom lsp-imenu-index-symbol-kinds
+  '(Class Method Property Field Constructor Enum Interface Function Struct)
```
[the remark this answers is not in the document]
wyuenho: 3 is the whole point of this PR - eliminate junk from the index. 2 is the first I tried, but they typically result in all the imports being included, and the first page is typically all imports. Nobody uses the index to find out where the imports are.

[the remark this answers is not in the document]
yyoncho: The PR will still be helpful to allow users customizing the visible elements even if we show everything by default. My concern is that we will my break imenu for someone.

> but they typically result in all the imports being included, and the first page is typically all imports. 

Can we then skip String, Variable, and whatever symbol stands for imports?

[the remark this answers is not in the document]
yyoncho: Also, if you want, we may merge it with everything displayed by default to get that working for you and then discuss what to hide by default in separate issue.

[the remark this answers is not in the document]
wyuenho: File/Module/Namespace/Package are most of the imports. There are language servers returning other kinds for imports, but those are probably wrong. The clients should be able to handle these cases but most of them aren't.

PR #2014, opened 2020-08-08
Create lookups for all enums

yyoncho:
----

#


scripts/lsp-generate-bindings.el
```diff
@@ -31,34 +31,43 @@
...
-             (-map (-lambda ((name type-data))
-                     (let* ((key (intern (substring (symbol-name name) 1)))
-                            (required (-map (lambda (field-name)
-                                              (intern (concat ":" field-name))) (plist-get type-data :required)))
-                            (params (->> (plist-get type-data :properties)
-                                         (-partition 2)
-                                         (-map #'cl-first)
-                                         (-filter (-not (-partial #'-contains? required))))))
-                       (list key required params))))
+         (-partition 2)
```
kiennq, line 57: The indent look weird?

lsp-protocol.el
```diff
@@ -290,6 +290,8 @@ See `-let' for a description of the destructuring mechanism."
 
 ;; begin autogenerated code
 
+(defvar lsp/completion-item-kind-lookup
```
kiennq, line 293: How should we use this look up?

wyuenho:
I see where you are going with this, but this doesn't really help. Ideally, we should use something that only requires a constant time lookup. Generating an array still requires me to linearly scan the vector with `seq-position` or something similar. Maybe try generating records or defstructs?

yyoncho:
> Ideally, we should use something that only requires a constant time lookup.

Most of the time it will end up to `Int -> Symbol` which will use `aget` which has constant time.

scripts/lsp-generate-bindings.el
yyoncho: Yes, but this is what we get with the latest format settings.

yyoncho:
Alternatively, instead of using aget everywhere, we might generate convertion functions to wrap that.

wyuenho:
Why? The enums generated currently already gives me the number.

yyoncho:
> Why? The enums generated currently already gives me the number.

If your question is about the wrapper functions it may help with error handling(e. g. going out of bounds).

wyuenho:
Generate a reverse lookup table then? Given a symbol kind integer, returns the name of the kind. This vector won't  help me at all since I don't need the index, I need the enum integer value.

yyoncho:
> Generate a reverse lookup table then? Given a symbol kind integer, returns the name of the kind. This vector won't help me at all since I don't need the index, I need the enum integer value.

This is lookup table, e. g. 

``` (memql (aref lsp/symbol-kind-lookup kind) lsp-imenu-index-symbol-kinds) ```

**Comment by wyuenho** [comment]
But how would I generate the set in `lsp-imenu-index-symbol-kinds` ? I still have to copy and paste the symbols in `lsp/symbol-kind-lookup`. You are also assuming the enum integers will always be a serial list of numbers with no gaps. I wouldn't make this assumption.

**Comment by yyoncho** [comment]
> But how would I generate the set in `lsp-imenu-index-symbol-kinds` ?

You do:
```
`(defcustom lsp-imenu-index-symbol-kinds '(Function Class Interface...))`
```

**Comment by wyuenho** [comment]
That's makes the defcustom so much longer, and the users not using customize still has to look up what symbols there are in the lookup table in addition to just finding what variables there are in lsp-protocols. If you insist, sure, I don't feel too strongly about this.

File: scripts/lsp-generate-bindings.el (hunk 1 above)
Comment (leungbk) [reply scripts/lsp-generate-bindings.el:57]: One of the `FIXME`s in the `.dir-locals.el` mentions how to work around this: we'll have to put the immediately-preceding sexp on its own line.

**Comment by yyoncho** [comment]
IMO keeping the numbers is no-go. The question here is whether to replace them with constants from lsp-protocol.el or with actual enum values as defined in the spec(e. g. Class, Interface). To me, using the enum values and then the lookup introduced in this PR seems like the most natural solution. Using the constants is probably acceptable too if we assume that beginners will use customize interface, while more advanced users will be table to find the constants from lsp-protocol.el. Let others share their opinion as well because we should establish that as a pattern and follow it through the project.

File: lsp-protocol.el (hunk 2 above)
Comment (ericdallo) [reply lsp-protocol.el:293]: I'm not sure too how we should use it.
@yyoncho can you give an example using these enums?

Comment (yyoncho) [reply lsp-protocol.el:293]:
```
(defcustom lsp-imenu-index-symbol-kinds '(Function Class Interface...))
```

Then, when you receive kind from the language server it can be converted to `Funtion`, `Class`, etc using the lookups autogenerated in this PR.

PR #2015, opened 2020-08-08
Support nested paths for workspace/configuration

zeronone:
- Language servers such as `pyright` request for nested paths in `workspace/configuration` (such as `python.analysis`)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5498,7 +5498,8 @@ textDocument/didOpen for the new file."
 PARAMS are the `workspace/configuration' request params"
   (->> items
        (-map (-lambda ((&ConfigurationItem :section?))
-               (gethash section? (lsp-configuration-section section?))))
+               (apply 'ht-get* (append (list (lsp-configuration-section section?))
+                                       (split-string section? "\\.")))))
        (apply #'vector)))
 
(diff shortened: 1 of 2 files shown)


yyoncho:
Can you add test here - https://github.com/emacs-lsp/lsp-mode/blob/master/test/lsp-common-test.el#L112 ? It is very easy to do so and it will help a lot.

lsp-mode.el
```diff
@@ -5498,7 +5498,8 @@ textDocument/didOpen for the new file."
 PARAMS are the `workspace/configuration' request params"
   (->> items
        (-map (-lambda ((&ConfigurationItem :section?))
-               (gethash section? (lsp-configuration-section section?))))
+               (apply 'ht-get* (append (list (lsp-configuration-section section?))
```
kiennq, line 5501: This can be 
```suggestion
               (apply 'ht-get* (nconc (list (lsp-configuration-section section?))
```
since you've already create a new list for the first item. `append` create new list.

Or you can write that like `` `(,lsp-configuration-section section) ,@(split-string section? "\\."))`` which doesn't require writing `append` or `nconc`

yyoncho:
Thank you!

PR #2017, opened 2020-08-08
Fix quotation

Vtec234:
see [here](https://github.com/emacs-lsp/lsp-mode/blob/c7199f6bd95d174b20bb5f6fb4e7b24c6ce1784a/lsp-clients.el#L118)


yyoncho:
Thank you!

Issue #2022, opened 2020-08-09
Lsp-diagnose fails if company is not installed


jech:
Running lsp-mode 20200608.122 from Melpa under Emacs 26.3.  Company is not installed.

`M-x lsp-diagnose` fails with the following message:
```
funcall-interactively: Symbol’s value as variable is void: company-backends
```
The backtrace doesn't provide any useful information:
```
Debugger entered--Lisp error: (void-variable company-backends)
  lsp-diagnose()
  funcall-interactively(lsp-diagnose)
  call-interactively(lsp-diagnose record nil)
  command-execute(lsp-diagnose record)
  execute-extended-command(nil "lsp-diagnose" "lsp-diagnose")
  funcall-interactively(execute-extended-command nil "lsp-diagnose" "lsp-diagnose")
  call-interactively(execute-extended-command nil nil)
  command-execute(execute-extended-command)
```

ericdallo:
Thanks for the report, I'll take a look

PR #2023, opened 2020-08-09
Fix company-backends validation on lsp-diagnose and safe rename to lsp-doctor

ericdallo:
Fix https://github.com/emacs-lsp/lsp-mode/issues/2022

During the migration of completion stuff to `lsp-completion.el` we moved this `company-backend` variable.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7368,7 +7368,7 @@ This avoids overloading the server with many files when starting Emacs."
 
 ;; lsp internal validation.
 
-(defmacro lsp--validate (&rest checks)
+(defmacro lsp--doctor (&rest checks)
   `(-let [buf (current-buffer)]
      (with-current-buffer (get-buffer-create "*lsp-performance*")
        (with-help-window (current-buffer)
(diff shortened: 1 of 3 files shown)


yyoncho:
You should update the docs referencing to lsp-diagnose.

lsp-mode.el
```diff
@@ -7380,10 +7380,15 @@ This avoids overloading the server with many files when starting Emacs."
                                       (propertize "ERROR" 'face 'error)))))
                  (-partition 2 checks))))))
 
-(defun lsp-diagnose ()
+(defvar company-backends)
```
kiennq, line 7383: How can just this fixing the mentioned issue?
From what I saw, the issue is when `company-mode` is **not** installed, thus forwarding declaration here will not help.
The forwarding declaration is only for byte compiling, for this issue, you may want to check for `(boundp 'company-backends)` instead.

ericdallo: How before it used to work without `boundp`?

kiennq: Does it work before without `company-mode`?
Maybe user set `company-backends`, but when I test with `emacs -Q`, it failed.
For example:
``` elisp
(setq lsp-completion-provider :none)    ; so company-mode will not be activated
(package-initialize)
< open any file lsp-mode support >
M-x lsp
M-x lsp-doctor
```
it failed.

ericdallo: I see, I'll open another PR changing to use `boundp`, thanks!

PR #2027, opened 2020-08-10
Add Nix support

sei40kr:


lsp-clients.el
```diff
@@ -1029,5 +1029,17 @@ responsiveness at the cost of possible stability issues."
...
+

+;; Nix
+(defgroup lsp-nix nil
+  "LSP support for Nix, using rnix-lsp."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/nix-community/rnix-lsp"))
+
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-stdio-connection "rnix-lsp")
```
kiennq, line 1040: You should probably defcustom this one so user can customize where the binary is.
Also using function instead of constant, so that when user change the binary path after `nix-ls` is loaded, it still works

ericdallo:
@sei40kr could you rebase and if you agree, follow the @kiennq suggestion?

leungbk:
@sei40kr Thanks for the PR. We closed this using your name in https://github.com/emacs-lsp/lsp-mode/commit/87033228302d8319e1cb68ce7a704f18b6e1beb0.

PR #2034, opened 2020-08-10
[docs] Add file watcher to performance section

ericdallo:


docs/page/performance.md
```diff
@@ -41,6 +41,16 @@ Benchmarks show that Emacs 27 is `~15 times` faster than Emacs when using Elisp
 (setq lsp-idle-delay 0.500)
 ```
 
+## Ignore watch folders/files
+
+By default `lsp-mode` tries to watch all files and folders of the project ignoring the regexp from `lsp-file-watch-ignored`. If you don't want some file or folder
```
yyoncho, line 46: This will happen only if the server wants to watch files in the project.

ericdallo: If the server has the capability you mean? [this](https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeWatchedFiles) capability?

yyoncho: Not exactly. The server might have the capability but we start watching the folder after server requests that.

ericdallo: I see, I'll update it

ericdallo: BTW I found nothing about server requesting that to the client in the [spec](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/)

yyoncho: What I meant is that not about server having the capability but about server sending `DidChangeWatchedFilesRegistrationOptions`

ericdallo: Oh got it!

PR #1832, opened 2020-06-21
Add LSP configuration to use sqls for SQL

v2okimochi:
I added LSP configuration for [sqls](https://github.com/lighttiger2505/sqls).

To use sqls, it need following steps:
- Get sqls through golang ( `go get`) according to sqls document
- Put workspace configuration to connect to DB

[There are some configuration methods](https://github.com/lighttiger2505/sqls#configuration-methods) to use sqls features, and this PR has one of method that `workspace/configuration`.

We can use this method through creating `config.json` under `.sqls/` at workspace root directory.

Sample `config.json` is following (cf. [workspace configuration sample by sqls official](https://github.com/lighttiger2505/sqls#workspace-configuration-sample))

```json:workspace-root-dir/.sqls/config.json
{
  "sqls": {
    "connections": [
      {
        "driver": "mysql",
        "dataSourceName": "user1:password1@tcp(localhost:3306)/sample_db"
      }
    ]
  }
}
```

If anyone have more great way than this PR, I would like to agree it.

PR #2035, opened 2020-08-10
Add Svelte support

endenwer:
Basic support for svelte with no configuration.

### DIFF
--- a/lsp-svelte.el
+++ b/lsp-svelte.el
@@ -1,0 +1,53 @@
+;;; lsp-svelte.el --- LSP Svelte integration -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 Stepan Lusnikov
+
+;; Author: Stepan Lusnikov <endenwer@gmail.com>
+;; Keywords: lsp svelte
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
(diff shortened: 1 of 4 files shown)


lsp-svelte.el
```diff
@@ -0,0 +1,53 @@
+;;; lsp-svelte.el --- LSP Svelte integration -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 Stepan Lusnikov
+
+;; Author: Stepan Lusnikov <endenwer@gmail.com>
```
ericdallo, line 5: I'm not sure if this should be `lsp maintainers` or your name.
Maybe @yyoncho can confirm that

lsp-svelte.el
```diff
@@ -0,0 +1,53 @@
...
+                '(:system "svelteserver")
+                '(:npm :package "svelte-language-server"
+                       :path "svelteserver"))
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection
+                   (lambda ()
+                     `(,(lsp-package-path 'svelte-language-server)
+                       "--stdio")))
```
ericdallo, line 45: What do you think about moving `"--stdio"` to a `defcustom`, this way user can pass any more args as he wants

endenwer: `"--stdio"` is required for the server to work. I can add `defcustom` with default value `'("--stdio")`, but if someone will try to change it and forget to add that value to the list the server will stop working.

lsp-svelte.el
endenwer: I checked different files and they have different authors. For example, one of the last pr #1832

yyoncho: It is fine to stay as it is.

lsp-svelte.el
ericdallo: then you can use a hardcoded `--stdio` concating with a defcustom for extra args, WDYT?

endenwer: I can do that, but there are no any other arguments. This is the only one. And I don't think there will be any, because everything else is passed thought the connection.

ericdallo:
Thank you for your contribution and for your first `lsp-mode` PR :tada:

PR #2038, opened 2020-08-10
Search for clangd executable in the path with names variations.

Ergus:
* lsp-clients.el (lsp-clients--clangd-command) : Use some logic to
search for the clangd executable in the path with different names.

(lsp-clients-clangd-executable) : Change default value to nil to search
in the path automatically. If the user sets a different name that one
will be preferred in the search.

(lsp-clients-clangd-executable-found) : Internal variable to set the
full path for the clangd executable when found.

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -430,13 +430,18 @@ particular FILE-NAME and MODE."
   :group 'lsp-mode
   :link '(url-link "https://clang.llvm.org/extra/clangd/"))
 
-(defcustom lsp-clients-clangd-executable "clangd"
+(defcustom lsp-clients-clangd-executable nil
   "The clangd executable to use.
-Leave as just the executable name to use the default behavior of
-finding the executable with `exec-path'."
+When `'non-nil' use the name of the clangd executable file
+available in your path to use. Otherwise the system will try to
+find a suitable one. Set this variable before loading lsp."
   :group 'lsp-clangd
   :risky t
   :type 'file)
+
+(defvar lsp-clients-clangd-executable-found nil
+  "Clang executable full path when found.
+This must be set only once after loading the clang client.")
 
 (defcustom lsp-clients-clangd-args '()
   "Extra arguments for the clangd executable."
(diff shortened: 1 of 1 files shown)


yyoncho:

kiennq:
This change broke default `clangd` setup in Windows. The problem is windows file name has `.exe` suffix and locate file will not be able to find those files.

ericdallo:
We should add a test on CI checking if `locate-file` is being used in any file on the repo and fail the test

Ergus:
Hi:

Sorry for my error. 

 > We should add a test on CI checking if `locate-file` is being used in any file on the repo and fail the test

But AFAIK `executable-find` calls `locate-file` internally; so why shouldn't be used?

The problem with the initial solution was that it didn't consider `exec-suffixes` to pass to `locate-file`. I am not sure if calling `executable-find` multiple times is ok when there is a suffix function for a reason. But I suppose it is ok.

Maybe I should have used something like:

```
(flatten-list (mapcar (lambda (x)
			(mapcar (lambda (y)
				  (concat x y))
				exec-suffixes))
		      '("" "-10" "-9" "-8" "-7")))
```

kiennq:
> I am not sure if calling `executable-find` multiple times is ok when there is a suffix function for a reason.

Yes, it may not be ok if `locate-file` with suffix is implemented natively in C and `executable-find` is implemented in Elisp. However since both `executable-find` and `locate-file` are implemented in Elisp, there should be no different in perf between them and we can use that.

Ergus:
`locate-file` is a wrapper around `locate-file-internal` just to parse the `predicate` argument when not numeric. `locate-file-internal` is the one who processes the suffixes and it does in native C.

kiennq:
Ah, I didn't notice that. Thanks, however the benchmark result point out that the code using `locate-file` with manual product of suffixes is much more slower than just using `executable-file`.

``` elisp
(benchmark-run 100 (catch 'path
                     (mapc (lambda (suffix)
                             (let ((path (executable-find (concat "clangd" suffix))))
                               (when path (throw 'path path))))
                           '("" "-10" "-9" "-8" "-7" "-6"))))
(1.5425339999999998 0 0.0)


(benchmark-run 100
  (locate-file "clangd"
               exec-path
               (flatten-list (mapcar (lambda (x)
                                       (mapcar (lambda (y)
                                                 (concat x y))
                                               exec-suffixes))
                                     '("" "-10" "-9" "-8" "-7")))
               1))
(7.373785 0 0.0)
```

Ergus:
Yes, you are right. 

This is a triple loop in lisp (2 mapcar + flatten) to build the full list in advance + an internal loop in C doing a system call. While the other one can potentially do an early exit if it finds the executable in the first cycles. 

I changed the order of the entries in your benchmark and as expected the numbers change. But lets assume we put the most probable alternatives first. So your solution is better.

PR #2043, opened 2020-08-11
Change to project-root from obsolete project-roots

theothornhill:
I got a bytecode compiling warning warning about an obsolete command. Not sure if it is safe to assume we can use this freely, but nevertheless, here is a patch doing the requested change. :)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3346,6 +3346,9 @@ in that particular folder."
        (lsp:text-document-sync-options-save?)
        (lsp:text-document-save-registration-options-include-text?)))
 
+(declare-function project-roots "ext:project" (project) t)
+(declare-function project-root "ext:project" (project) t)
+
 (defun lsp--suggest-project-root ()
   "Get project root."
   (or
@@ -3354,8 +3357,10 @@ in that particular folder."
                                   (error nil)))
    (when (featurep 'project)
      (when-let ((project (project-current)))
-       (car (with-no-warnings
-              (project-roots project)))))))
+       (if (fboundp 'project-root)
+           (project-root project)
+         (car (with-no-warnings
+                (project-roots project))))))))
 
 (defun lsp--read-from-file (file)
   "Read FILE content."


ericdallo:
@theothornhill the CI is not finding this variable, is this available only on emacs recent versions?

leungbk:
> @theothornhill the CI is not finding this variable, is this available only on emacs recent versions?

It's in the latest `project.el` in MELPA, as well as in Emacs 28. The `project.el` bundled with Emacs <= 27 won't necessarily have it.

yyoncho:
I think that we should do the following:

1. If project-roots is defined call it
2. If project-roots is not defined call project-root in ignore-warning block. 

Having hard dependency to latest project.el in order to use unconditionally project-roots does not make sense to me since this is not a critical component.

theothornhill:
> 
> Having hard dependency to latest project.el in order to use unconditionally project-roots does not make sense to me since this is not a critical component.

Yeah, I agree. Or we could just skip this issue, and mend things later down the road?

I'll look into expanding the patch anyways.

leungbk:
> Or we could just skip this issue, and mend things later down the road?
> 
> I'll look into expanding the patch anyways.

Your submission won't step on anyone's toes as long as you do what Ivan suggested. Also make sure to add

``` elisp
(declare-function project-roots "ext:project" nil t)
(declare-function project-root "ext:project" nil t)
```

kiennq:
The `fileonly` argument in `declare-function` is intended to use only for function definitions that `check-declare` does not recognize, like `defstruct`. Using it for this case seems overkill.
And if you check function with `fboundp` before use, the `declare-function` is not needed.

leungbk:
> The `fileonly` argument in `declare-function` is intended to use only for function definitions that `check-declare` does not recognize, like `defstruct`. Using it for this case seems overkill.

Hmm, I had assumed that `cl-defgeneric` was in that class of unrecognized functions. But one of the guys who contributes to `project.el` recently submitted this to Magit:

https://github.com/magit/magit/blob/4033285e04403178b2cb9191d88749c3b672e9d0/lisp/magit-extras.el#L34

So you might be right here.

> And if you check function with `fboundp` before use, the `declare-function` is not needed.

`fboundp` gives no hint to the byte-compiler where to find a function:

``` c
DEFUN ("fboundp", Ffboundp, Sfboundp, 1, 1, 0,
       doc: /* Return t if SYMBOL's function definition is not void.  */)
  (Lisp_Object symbol)
{
  CHECK_SYMBOL (symbol);
  return NILP (XSYMBOL (symbol)->u.s.function) ? Qnil : Qt;
}
```

It seems to help us only when we're using it at runtime.

When I used Guix to manage my packages, the absence of `declare-function` in packages would, in some cases, prevent certain functions in those packages from working at all, whether or not `fboundp` was being used. Including `declare-function` made the problems go away in those (rare) cases.

In almost all cases, everything worked fine for me without the expected `declare-function` invocations. But I could never develop a general rule for when it actually mattered.

kiennq:
> `fboundp` gives no hint to the byte-compiler where to find a function:
> 
> ```c
> DEFUN ("fboundp", Ffboundp, Sfboundp, 1, 1, 0,
>        doc: /* Return t if SYMBOL's function definition is not void.  */)
>   (Lisp_Object symbol)
> {
>   CHECK_SYMBOL (symbol);
>   return NILP (XSYMBOL (symbol)->u.s.function) ? Qnil : Qt;
> }
> ```
> 
> It seems to help us only when we're using it at runtime.

Well, I have little to no knowledge about elisp byte-compiler hint finding so I cannot really comment on this.

However from other language compiler, unless the function needed to be also byte-compiled, the byte compiler should not need to look to where it's defined.

Also from `declare-function` docs, the FILE argument is not used for `byte-compiler`, but by the `check-declare` package. So I don't think `declare-function` and check for `fboundp` is that different.

leungbk:
> However from other language compiler, unless the function needed to be also byte-compiled, the byte compiler should not need to look to where it's defined.
> 
> Also from `declare-function` docs, the FILE argument is not used for `byte-compiler`, but by the `check-declare` package. So I don't think `declare-function` and check for `fboundp` is that different.

Hmm, you may be right about `fboundp` having a role in byte-compilation. https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/bytecomp.el#n4048 and https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/bytecomp.el#n4085 seem to handle `fboundp` stuff.

But `declare-function` seems to be of interest to the byte-compiler: https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/bytecomp.el#n470 puts `declare-function` in the initial environment, and https://git.savannah.gnu.org/cgit/emacs.git/tree/lisp/emacs-lisp/bytecomp.el#n3101 makes clear that even though `FILE` is not of immediate interest to the byte-compilation process, the function still appears in the `byte-compilation-environment`.

If neither of us is particularly knowledgeable about Emacs' byte-compilation process, it seems safer to err on the side of caution and simply use both `fboundp` and `declare-function`.

theothornhill:
Now things are building cleanly, trying to follow what @kiennq and @leungbk discussed. 

What do you think?

lsp-mode.el
```diff
@@ -3354,8 +3357,10 @@ in that particular folder."
                                   (error nil)))
    (when (featurep 'project)
      (when-let ((project (project-current)))
-       (car (with-no-warnings
-              (project-roots project)))))))
+       (if (fboundp 'project-roots)
+           (car (with-no-warnings
```
kiennq, line 3361: Can you remove this `with-no-warning` and see if it works?

lsp-mode.el
```diff
@@ -3354,8 +3357,10 @@ in that particular folder."
                                   (error nil)))
    (when (featurep 'project)
      (when-let ((project (project-current)))
-       (car (with-no-warnings
-              (project-roots project)))))))
+       (if (fboundp 'project-roots)
```
kiennq, line 3360: You should check for `(fboundp 'project-root)` since that's newer function

lsp-mode.el
theothornhill: yeah

lsp-mode.el
theothornhill: sure

theothornhill:
> Can you remove this with-no-warning and see if it works?

@kiennq It didn't, actually.

lsp-mode.el
```diff
@@ -3354,8 +3357,10 @@ in that particular folder."
                                   (error nil)))
    (when (featurep 'project)
      (when-let ((project (project-current)))
-       (car (with-no-warnings
-              (project-roots project)))))))
+       (if (fboundp 'project-root)
+           (car (project-root project))
```
yyoncho, line 3361: on my system `(project-root (project-current))` returns a string?

theothornhill: Yes, it looks like it. Better remove that "car" then?

yyoncho: Yep.

theothornhill: removed the car

yyoncho:

PR #2028, opened 2020-08-10
Add Lua Language Server, Lua-Lsp and update EmmyLua

elxbarbosa:
Hi, 

In an early [discussion](https://github.com/emacs-lsp/lsp-mode/issues/831) the [Lua Language Server](https://github.com/sumneko/lua-language-server) had not enough instructions as on how clients should call it.

Changes:
- add lsp-lua.el
- add lua-lsp and lua-language-server to lsp-clients.json
- update mkdocs.yml
- update changelog
- add to lsp-client-packages

TODO:
- add instruction to install servers
- automatically install/update emmylua and lua-language-server
- test servers are running

---- 
more

[Buildind](https://github.com/sumneko/lua-language-server/wiki/Build-and-Run-(Standalone)) Lua Language Server is simple as it requires only Ninja as an external dependency. It takes no more than a few seconds(20s) to build.

Once built, its binary will be in `bin/Linux/lua-language-server`. 

[Calling](https://github.com/sumneko/lua-language-server/wiki/Setting-without-VSCode) the server goes as **binary** + **flags options** `-E` and its `main.lua` file under the root dir of the project.

As for the **EmmyLua server**, it provides a single jar to be ready, maybe the way that most servers here on lsp-mode "install" can be re-used. 👍🏿 

https://github.com/EmmyLua/EmmyLua-LanguageServer/releases

EmmyLua location aint anymore hard-coded to Emacs `<= 26` location, as Emacs 27 will look for XDG CONFIG HOME too :)

PR #2048, opened 2020-08-11
Ada client to lsp-ada.el file

elxbarbosa:
Hi,

Moving Ada client to its own file ls-ada.el.

No modification done other than from `:package-version '(lsp-mode . "6.1"))` to `:version "7.1"`

I did try to test the server but even the `ada-mode` aint loading in my emacs 27.1, lots of errors.

more: https://comp.lang.ada.narkive.com/NOrFHxka/installing-ada-mode-6-2-1-on-debian-buster

### DIFF
--- a/lsp-lua.el
+++ b/lsp-lua.el
@@ -33,7 +33,7 @@
   :version "7.1"
   :link '(url-link "https://github.com/EmmyLua/EmmyLua-LanguageServer"))
 
-(defcustom lsp-clients-emmy-lua-java-path (executable-find "java")
+(defcustom lsp-clients-emmy-lua-java-path "java"
   "Java Runtime binary location."
   :group 'lsp-emmy-lua
   :version "7.1"


lsp-ada.el
```diff
@@ -0,0 +1,73 @@
...
+(require 'lsp-mode)
+
+;;; Ada
+(defgroup lsp-ada nil
+  "LSP support for the Ada Programming Language, using the Ada Language Server."
+  :group 'lsp-mode
+  :tag "Language Server"
+  :version "7.1")
+
+(defcustom lsp-clients-ada-project-file "default.gpr"
```
ericdallo, line 36: Like @yyoncho said it [here](https://github.com/emacs-lsp/lsp-mode/pull/2028#issuecomment-671714613), you should keep the same variable names.
I think you can keep this name but 'define-obsolete-variable-alias' like [here](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L96-L101) to deprecate those old names.

ericdallo: And you can keep the "6.1" variable package-version, since that already exists, it was only renamed

lsp-ada.el
```diff
@@ -0,0 +1,73 @@
+;;; lsp-ada.el --- description -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 E. Alexander Barbosa
```
ericdallo, line 3: I still think this should be [lsp maintainers](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-headerline.el#L3) since everybody will maintain this

elxbarbosa: yep, I forgot to ask about that :)

lsp-ada.el
elxbarbosa: done and done.

oh, my bad. I did lsp-ada before reading that, forgot to undo it before pushing :)

lsp-ada.el
elxbarbosa: done

PR #2052, opened 2020-08-12
JavaScript and TypeScript clients to lsp-javascript.el

elxbarbosa:
Hi,

I may be wrong here but just to be safe I have add all JS/TS clients as in one file. 

I hope front-end people do not curse me for that  :)


kiennq:
You will need to fix the test too. But anyway, nice.
Btw, I would go with the name `lsp-js` :smile:, which javascript deserved as most popular one right now

elxbarbosa:
I will need help as for tests, I can even install cask on linux :/

kiennq:
The test are failing at ` (void-function lsp-clients-flow-tag-string-present-p)` so probably you will need to `(require 'lsp-javascript)` in the test file.
Normally I read the source code and let Github run CI. Don't have Cask too.

elxbarbosa:
Done resolving conflict. Kindly do this one first :)

elxbarbosa:
hey, @ericdallo @yyoncho 

lsp-clients-test too are extract to its own file(clangd,javascript). I would not dare to remove lsp-clients as lsp-rust seems to depend on it! 
https://github.com/emacs-lsp/lsp-mode/blob/70551a71b5d3f98c0b4d4b113107db9029617caa/lsp-rust.el#L571

ericdallo:
@elxbarbosa, `lsp-rust.el` does not depends on `lsp-clients.el`, this is just a variable from `lsp-mode`. You can confirm that too, checking that `lsp-rust.el` does not `(require 'lsp-clients)`.

elxbarbosa:
here goes nothing

test/lsp-javascript-test.el
```diff
@@ -105,42 +105,4 @@
   (should (not (lsp-typescript-javascript-tsx-jsx-activate-p "abc.tsxx")))
   (should (not (lsp-typescript-javascript-tsx-jsx-activate-p "abc.jss"))))
 
-(ert-deftest lsp-clients-extract-signature-from-clangd-on-hover ()
```
ericdallo, line 108: I believe these tests should not be removed, right?

elxbarbosa: Good catch. lsp-clients-test.el has to still hold clangd tests!

PR #2053, opened 2020-08-12
Dockerfile client to lsp-dockerfile.el

elxbarbosa:
Hi,

changes:
- add defgroup


kiennq:
The test are failing at ` (void-function lsp-clients-flow-tag-string-present-p)` so probably you will need to `(require 'lsp-javascript)` in the test file.
Normally I read the source code and let Github run CI. Don't have Cask too.

PR #2057, opened 2020-08-12
Nim clients to lsp-nim.el

elxbarbosa:
Hi,

Changes:
- none

Issues:
- none

### DIFF
--- a/lsp-clients.el
+++ b/lsp-clients.el
@@ -790,20 +790,8 @@ responsiveness at the cost of possible stability issues."
  (make-lsp-client :new-connection (lsp-stdio-connection lsp-clients-crystal-executable)
                   :major-modes '(crystal-mode)
                   :server-id 'scry))
+
 
-
-;; Nim
-(defgroup lsp-nim nil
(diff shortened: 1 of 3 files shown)


lsp-clients.el
```diff
@@ -609,17 +609,22 @@ responsiveness at the cost of possible stability issues."
...
+  "LSP support for Crystal via scry."
   :group 'lsp-mode
-  :link '(url-link "https://github.com/PMunch/nimlsp"))
+  :link '(url-link "https://github.com/crystal-lang-tools/scry"))
+
+(defcustom lsp-clients-crystal-executable '("scry" "--stdio")
+  "Command to start the scry language server."
+  :group 'lsp-crystal
+  :risky t
+  :type 'file)
```
ericdallo, line 622: I think this is wrong, right?
crystal already has its own server on `lsp-crystal.el`

elxbarbosa: fixed!

PS: It seems github do not notify every conversation...

PR #2062, opened 2020-08-12
Lsp groovy.el

elxbarbosa:
Let's groove tonigh :)


lsp-groovy.el
```diff
@@ -0,0 +1,53 @@
...
+;;; Commentary:
+
+;; LSP Clients for the Groovy Programming Language.
+
+;;; Code:
+
+(require 'lsp-mode)
+(require 'f)
+
+;;; Groovy
```
ericdallo, line 30:
```suggestion
```

elxbarbosa: done!

PR #2064, opened 2020-08-12
R clients to lsp-r.el

elxbarbosa:
up


elxbarbosa:
good catch! 👍🏿

ericdallo:
Thank you!

PR #2070, opened 2020-08-13
Add lsp-elixir.el

elxbarbosa:
up


tarsius:
What about the `lsp-elixir` package from https://github.com/elixir-lsp/lsp-elixir.el, which is distributed on Melpa?

That obviously conflicts. Should it be removed from Melpa?

/cc @Trevoke

Trevoke:
Thanks for bringing this up. As the elixir-ls community, we need to make a decision on this. We'll get back to you.

Trevoke:
@tarsius yeah, it can be removed from melpa, thanks for pinging us and making sure we're involved and kept updated.

tarsius:
Thanks!

PR #2073, opened 2020-08-13
Add lsp-tex.el

elxbarbosa:
up


ericdallo:
@ROCKTAKEY could you confirm that this is ok?

ROCKTAKEY:
@ericdallo Yeah, of course OK. I have no right to prevent :) .

elxbarbosa:
next  ones should be faster, as I had to prepare my mom's birthday party early this morning. Its done!

PR #2076, opened 2020-08-13
[infra] Use Emacs 28's calculate-lisp-indent

leungbk:
----

#

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -15,41 +15,29 @@
 
             (defvar lsp--override-calculate-lisp-indent?
               nil
-              "Whether to override the default
-              `calculate-lisp-indent'.")
+              "Whether to override `lisp-indent-function' with
+              the updated `calculate-lisp-indent' definition from
+              Emacs 28.")
 
-            (setq-local lsp--override-calculate-lisp-indent? t)
+            ;; from Emacs 28
 
-            ;; FIXME: This doesn't appear to correct the behavior of
-            ;; `evil-indent' or `indent-region'. But it does seem to
-            ;; work with `newline-and-indent'.
+            (defun wrap-calculate-lisp-indent (func &optional parse-start)
+              "Return appropriate indentation for current line as Lisp code.
+In usual case returns an integer: the column to indent to.
+If the value is nil, that means don't change the indentation
+because the line starts inside a string.
 
-            ;; FIXME: This also doesn't work particularly well with
-            ;; comma-evaluated forms inside backticks if any arguments
-            ;; are on the same line as the function. But there is a workaround.
+PARSE-START may be a buffer position to start parsing from, or a
+parse state as returned by calling `parse-partial-sexp' up to the
+beginning of the current line.
 
-            ;; For example, the incorrectly-indented
-
-            ;; `(:range ,(lsp--range (lsp--point-to-position start)
-            ;;            (plist-get lsp--before-change-vals :end-pos)))
-
-            ;; can be changed to
-
-            ;; `(:range ,(lsp--range
-            ;;            (lsp--point-to-position start)
-            ;;            (plist-get lsp--before-change-vals :end-pos)))
-
-            ;; by executing `newline-and-indent'.
-
-            ;; https://www.reddit.com/r/emacs/comments/d7x7x8/finally_fixing_indentation_of_quoted_lists/
-            (defun wrap~calculate-lisp-indent (fn &optional parse-start)
-              "Add better indentation for quoted and backquoted lists."
+The value can also be a list of the form (COLUMN CONTAINING-SEXP-START).
+This means that following lines at the same level of indentation
+should not necessarily be indented the same as this line.
+Then COLUMN is the column to indent to, and CONTAINING-SEXP-START
+is the buffer position of the start of the containing expression."
               (if (not lsp--override-calculate-lisp-indent?)
-                  (funcall fn parse-start)
-                ;; This line because `calculate-lisp-indent-last-sexp` was defined with `defvar`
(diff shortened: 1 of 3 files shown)


kiennq:
If this file is not really related to `lsp-mode` functionality, I recommended to put it under `scripts` folder so user will not receive it.

That a side, I think we want the indentation to become default for everyone who works on `lsp-mode`, why move it to a file and make it optional then?

leungbk:
> I think we want the indentation to become default for everyone who works on `lsp-mode`, why move it to a file and make it optional then?

I think it's because @yyoncho didn't want people to get the safe-local-variable (I forget the name) warning.

Emacs 28 provides an alternate means of indenting plists and such. Ideally we'd use that and get rid of the wrapper entirely, but it's too much to expect everybody who contributes to the repo to use 28.

Another solution is to make the wrapper part of our own minor mode and ask people to do

``` elisp
(defun maybe-init-lsp-minor-mode ()
    (when (and buffer-file-name
               (string-match "\\<lsp-\\>" buffer-file-name))
      (lsp-dev-minor-mode 1)))

(add-hook 'emacs-lisp-mode-hook #'maybe-init-lsp-minor-mode)
```

What do you think?

kiennq:
> Emacs 28 provides an alternate means of indenting plists and such. Ideally we'd use that and get rid of the wrapper entirely, but it's too much to expect everybody who contributes to the repo to use 28.

Shouldn't the wrap function check `emacs 28` and use emacs 28 plist indentation instead?

You will still get the `safe-local-variable` warning for `flycheck-disabled-checkers` though, so moving just the indentation to other files seems not helping much.
Instead of modify config file (and keeps that config file in sync with several other machines), I think it's easier for contributor to just press `!` to accept local dir variables in `lsp-mode`. WDYT @yyoncho

leungbk:
> Shouldn't the wrap function check `emacs 28` and use emacs 28 plist indentation instead?

Yeah, we can do that. 

```
*** A space between an open paren and a symbol changes the indentation rule.
The presence of a space between an open paren and a symbol now is
taken as a statement by the programmer that this should be indented
as a data list rather than as a piece of code.
```

We can either a) continue using `.dir-locals.el` in some form for this, b) make a `lsp-dev-minor-mode`, or c) re-define `calculate-lisp-indent` globally since it's not defined in way that's going to step on anyone's toes. Which do you think is best?

kiennq:
I would recommend to keep using `.dir-locals.el` since it's standard way in Emacs to keep enforcing coding guidelines.
Just need to modify its indent calculation to keep up with newest Emacs version (nit: can you just copy Emacs 28 indent function and put it there?)

yyoncho:
The issue on my side is that the progn that is evaluated as part of dir-locals has to be accepted each time I start emacs. I was unable to make the accept work across sessions. If we just keep one variable and it narrows down to accepting this one time I am fine.

yyoncho:
Apparently there is `(setq enable-local-variables :all)` which avoids the query.

leungbk:
> The issue on my side is that the progn that is evaluated as part of dir-locals has to be accepted each time I start emacs.

That's weird, I can't reproduce this with either a clean `init.el` or my main `init.el`. Does this still happen when you remove the indentation function from `.dir-locals.el` but keep the `imenu` stuff in the `eval`?

leungbk:
> Apparently there is `(setq enable-local-variables :all)` which avoids the query.

Oh, I didn't see this message at first. So you're OK with keeping the function in `.dir-locals.el`?

yyoncho:
> Oh, I didn't see this message at first. So you're OK with keeping the function in `.dir-locals.el`?

Yes.

> That's weird, I can't reproduce this with either a clean init.el or my main init.el. Does this still happen when you remove the indentation function from .dir-locals.el but keep the imenu stuff in the eval?

I think it might be related to spacemacs and the way it keeps the custom data.

lsp-mode.el
```diff
@@ -3816,8 +3816,10 @@ interface TextDocumentEdit {
     (lsp:set-position-line (max 0 line))
     (lsp:set-position-character (max 0 character))))
 
-(lsp-defun lsp--apply-text-edit-replace-buffer-contents ((edit &as &TextEdit :range (&Range :start :end)
-                                                                             :new-text))
+(lsp-defun lsp--apply-text-edit-replace-buffer-contents (( edit &as
+                                                           &TextEdit
+                                                           :range (&Range :start :end)
+                                                           :new-text))
```
leungbk, line 3822: This is a bit ugly. But without the space, the indentation is arguably worse.

.dir-locals.el
```diff
@@ -219,6 +177,7 @@
                             (t
                              normal-indent)))))))
 
-            (advice-add #'calculate-lisp-indent :around #'wrap~calculate-lisp-indent)))
+            (advice-add #'calculate-lisp-indent :around #'wrap-calculate-lisp-indent)))
```
kiennq, line 180: nit: I think we can check Emacs version here and don't need to advice it if the version is >= "28.0"

lsp-mode.el
kiennq: Did this look worse
image

I think it looks fine, especially since it indent `:range` and `:new-text` as property of `a`

leungbk:
OK, I've updated to use 28's indentation in `lsp-mode.el`. If this gets approved, the commits should ideally not be squashed since the commit updating the `.git-blame-ignore-revs` file assumes the commits are separate.

yyoncho:
Looks good. I will push it using command line later today to preserve the commits.

yyoncho:
Merged manually

PR #2085, opened 2020-08-15
Filter textDocument/didChangeWatchedFiles events according to kind

### LINKED ISSUE #938: didChangeWatchedFiles "kind" registration option not respected
**Describe the bug**
I registered for only "change" events, but my server gets sent "create" and "delete" events as well. 

**To Reproduce**
When registering for watched file events, set `DidChangeWatchedFilesRegistrationOptions.Kind` to "2" (which is `Change` only). Notice Emacs still sends "create" and "delete" notifications.

**Expected behavior**
Emacs only sends "change" notifications.

gopls (doesn't support watching files yet)

**OS**
Mac

leungbk:
Closes #938.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -930,6 +930,11 @@ must be used for handling a particular message.")
     (changed . 2)
     (deleted . 3)))
 
+(defconst lsp--watch-kind
+  `((create . 1)
+    (change . 2)
+    (delete . 4)))
+
 (defvar lsp-window-body-width 40
   "Window body width when rendering doc.")
 
@@ -3237,7 +3242,10 @@ disappearing, unset all the variables related to it."
 
 (defun lsp--file-process-event (session root-folder event)
   "Process file event."
-  (let ((changed-file (cl-third event)))
+  (let* ((changed-file (cl-third event))
+         (event-numeric-kind (alist-get (cl-second event) lsp--file-change-type))
+         (bit-position (1- event-numeric-kind))
+         (watch-bit (ash 1 bit-position)))
     (->>
      session
      lsp-session-folder->servers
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -928,7 +928,7 @@ must be used for handling a particular message.")
 (defconst lsp--file-change-type
   `((created . 1)
     (changed . 2)
-    (deleted . 3)))
+    (deleted . 4)))
```
muirdm, line 931: This should be 3 according to https://microsoft.github.io/language-server-protocol/specification#workspace_didChangeWatchedFiles, so I don't think you can change it to 4 to allow bitwise operations.

leungbk: Oh, I didn't see that part of the spec. I mistook `WatchFile` events (which proceed in 1, 2, 4 order) for these events.

lsp-mode.el
```diff
@@ -3256,14 +3264,16 @@ disappearing, unset all the variables related to it."
                            lsp--registered-capability-options
                            (lsp:did-change-watched-files-registration-options-watchers)
                            (seq-find
-                            (-lambda ((&FileSystemWatcher :glob-pattern))
-                              (-let [glob-regex (lsp-glob-to-regexp glob-pattern)]
-                                (or (string-match glob-regex changed-file)
-                                    (string-match glob-regex (f-relative changed-file root-folder)))))))))))
+                            (-lambda ((&FileSystemWatcher :glob-pattern :kind?))
+                              (and (or (null kind?)
```
yyoncho, line 3268: this `and` is `when`?

lsp-mode.el
```diff
@@ -3256,14 +3264,16 @@ disappearing, unset all the variables related to it."
...
+                                       (> (logand kind? watch-bit) 0))
```
yyoncho, line 3269: nit: you may use natnump instead of `> 0`

lsp-mode.el
leungbk: Done.

lsp-mode.el
leungbk: I don't think `natnump` works here since we want to reject 0 as well, and 0 is a natural number.

PR #2089, opened 2020-08-16
Set `lsp--symbol-bounds-of-last-highlight-invocation' when...

leungbk:
...point isn't initially on highlighted text.

Related to #1239.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4512,21 +4512,19 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
   was most recently invoked.")
 
 (defun lsp--document-highlight ()
-  (unless (or (looking-at "[[:space:]\n]")
-              (not lsp-enable-symbol-highlighting)
-              (and (lsp--point-on-highlight?)
-                   (let ((curr-sym-bounds (bounds-of-thing-at-point 'symbol)))
-                     (prog1
-                         (and curr-sym-bounds
-                              (equal curr-sym-bounds
-                                     lsp--symbol-bounds-of-last-highlight-invocation))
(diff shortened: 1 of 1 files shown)


muirdm:
Seems to do the trick, thanks!

lsp-mode.el
```diff
@@ -4512,21 +4512,20 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
...
-                             curr-sym-bounds)))))
-    (lsp-request-async "textDocument/documentHighlight"
-                       (lsp--text-document-position-params)
-                       #'lsp--document-highlight-callback
-                       :mode 'tick
-                       :cancel-token :highlights)))
+  (let ((curr-sym-bounds (bounds-of-thing-at-point 'symbol)))
+    (unless (or (looking-at "[[:space:]\n]")
+                (not lsp-enable-symbol-highlighting)
+                (and (lsp--point-on-highlight?)
```
yyoncho, line 4518: lsp--point-on-highlight? is redundant now because practically it will be always overridden by current symbol check.

leungbk: OK, I removed that invocation.

yyoncho: (and the function itself).

yyoncho: > (and the function itself).

I am not quite sure about that though. there is one more usage.

leungbk: I'd prefer to keep the function. I think it would be messier if we replaced its invocation with a `bounds-of-thing-at-point` check of some kind.

yyoncho: Yes. Actually I like the current implementation very much. It removes the overlays when you go out of the overlay and it refreshes the overlay when you move out of the current symbol.

PR #2090, opened 2020-08-16
Fix `lsp--document-highlight'

leungbk:
If point moves off a highlighted symbol to some whitespace, then moves back to the symbol, then that symbol will not be highlighted without the additional check.


yyoncho:
I guess the alternative is to clear lsp--symbol-bounds-of-last-highlight-invocation when we clear the highlights?

leungbk:
> I guess the alternative is to clear lsp--symbol-bounds-of-last-highlight-invocation when we clear the highlights?

I changed the check to look at a boolean instead.

PR #2093, opened 2020-08-17
Turn on `async-bytecomp-package-mode' when available

leungbk:
This reduces the likelihood of a bad reinstallation of `lsp-mode'.

----

#


kiennq:
> This reduces the likelihood of a bad reinstallation of `lsp-mode'.

Why? This should probably goes into user configuration file instead of something like `lsp-mode`.


Moreover currently there's no official way to throttling the number of process `async-start` spawns, in Windows if there's a lot of packages, mostly user will get `too much files opened` error and fail the package installation.

Another problem is that the dependencies of other packages on `lsp-mode` (for example, `lsp-treemacs`  ...). 
If both of them got update asynchronously, since the `load-path` is only injected into sub emacs process (which run the async command) at the begining, once `lsp-mode` finished updating and the old folder is deleted, the process which compiling `lsp-treemacs` will mostly get the cannot find `lsp-mode` error and cannot continue to byte compiling.

yyoncho:
> Another problem is that the dependencies of other packages on `lsp-mode` (for example, `lsp-treemacs` ...).
> If both of them got update asynchronously, since the `load-path` is only injected into sub emacs process (which run the async command) at the begining, once `lsp-mode` finished updating and the old folder is deleted, the process which compiling `lsp-treemacs` will mostly get the cannot find `lsp-mode` error and cannot continue to byte compiling.

Is this how async-byte-compile works? Because if it works like that it will be unusable.

kiennq:
> Is this how async-byte-compile works? Because if it works like that it will be unusable.

This is timing issue, since `async-byte-compile-mode` advise `package--compile` and will run the asynchronous compiling in package unpack phase.
Since the `load-patch` is injected once when the compiling starting (packages should be downloaded and now need unpacked), if two dependent packages got into that overlapping and one of them finish deleting old packages before the other get into that part to compile, it will happen.
Perhaps the package author @jwiegley can help more.

That's what I spectacular after receiving failing to installing a lot.
A quick `(async-byte-recompile-directory package-user-dir 'quiet)` at the end will restore everything back to normal though.

leungbk:
> This is timing issue, since `async-byte-compile-mode` advise `package--compile` and will run the asynchronous compiling in package unpack phase.
> Since the `load-patch` is injected once when the compiling starting (packages should be downloaded and now need unpacked), if two dependent packages got into that overlapping and one of them finish deleting old packages before the other get into that part to compile, it will happen.
> Perhaps the package author @jwiegley can help more.

@thierryvolpiatto might have some input too. I was hoping that this would eliminate the problem of byte-compiling something in an environment containing outdated macro definitions and such.

If this turns out to be a completely wrong approach, I suppose we should instead recommend in the README alternative package managers (Straight, Nix, Guix) that don't encounter this problem.

yyoncho:
> If this turns out to be a completely wrong approach, I suppose we should instead recommend in the README alternative package managers (Straight, Nix, Guix) that don't encounter this problem.

We could recommend a 3-stage update procedure.

1. Delete all lsp-mode related stuff.
2. Restart emacs 
3. Install the stuff again.

This is what spacemacs is doing to avoid the issue.

jwiegley:
Please consult @thierryvolpiatto, as I did not write the async compilation logic.

kiennq:
> If this turns out to be a completely wrong approach, I suppose we should instead recommend in the README alternative package managers (Straight, Nix, Guix) that don't encounter this problem.

I wonder why other package managers don't encounter this problem? Did they do not byte compile or have a better package dependencies updating path?

yyoncho:
> I wonder why other package managers don't encounter this problem? Did they do not byte compile or have a better package dependencies updating path?

It depends on:

1. whether the package manager recompiles FOO's downstream dependencies when you update package FOO 
2. the way FOO's files are compiled. E. g. when updating lsp-mode the package manager should start byte-compiling from lsp-mode.el, not from lsp-eslint.el.

Most likely these package managers fail too but I might be wrong.

thierryvolpiatto:
Ivan Yonchovski <notifications@github.com> writes:

> It depends on:
>
>  1. whether the package manager recompiles FOO's downstream
>  dependencies when you update package FOO

package.el is not taking care of this, packages and dependencies are
treated in a sorted order, so most of the time, you will have FOO
compiled/loaded first and then FOO-DEP1 FOO-DEP2 etc... compiled loaded
after.
This lead to errors in the best cases at compile time with error
messages, worst error occur later when using package's commands
and users think it is a bug in package.

See https://github.com/emacs-helm/helm/wiki#melpa

>  
>  2. the way FOO's files are compiled. E. g. when updating lsp-mode the
>  package manager should start byte-compiling from lsp-mode.el, not
>  from lsp-eslint.el.

So it seems what you are expecting have nothing to do with async
compilation, but the order in which the packages are compiled/loaded,
but anyway compiling async is always safer for other reasons even if it
doesn't fix your problem.

> Most likely these package managers fail too but I might be wrong.

It seems straight is taking care of compiling/loading dependencies first
but better ask its maintainer to be sure.
For compilation, straight build packages when restarting emacs, so it
don't need async compilation as it builds always in a clean environment.


-- 
Thierry

leungbk:
> So it seems what you are expecting have nothing to do with async compilation, but the order in which the packages are compiled/loaded, but anyway compiling async is always safer for other reasons even if it doesn't fix your problem.

Thanks for your response.

OK, I updated the troubleshooting file to recommend Ivan's procedure for people using `package.el`.

PR #2094, opened 2020-08-18
[lsp-lua] fix invalid executable testing for emmy-lua client

sunlin7:
Hi, 
The lsp-emmy-lua is breaked for the invalid execuable testing function.
Here is the fix.


yyoncho:

sunlin7:
@yyoncho thank you so fast response!

Issue #1759, opened 2020-06-07
[Task] investigate removing lsp--folding-range 


yyoncho:
After https://github.com/emacs-lsp/lsp-mode/issues/1751 is fixed we may check if lsp--folding-range can be removed in favour of using directly the struct coming from the server. We did that for lsp-diagnostics and it resulted in better/less code overall.

yyoncho:
cc @leungbk

leungbk:
The `FoldingRange`s returned by LSP servers don't have a `children` property of any kind. `origami.el` treats folding ranges as trees with children. I think we should keep the `lsp--folding-range` struct for this reason.

We should be able to remove the `orig-folding-range` property on that struct, though. It doesn't seem like `lsp-focus` or `lsp-origami` makes use of it.

PR #2096, opened 2020-08-18
[folding] Remove orig-folding-range property from `lsp--folding-range'

leungbk:
re #1759.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1937,7 +1937,7 @@ WORKSPACE is the workspace that contains the diagnostics."
 
 ;; textDocument/foldingRange support
 
-(cl-defstruct lsp--folding-range beg end kind children orig-folding-range)
+(cl-defstruct lsp--folding-range beg end kind children)
 
 (defvar-local lsp--cached-folding-ranges nil)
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
Thank you!

yyoncho:
Assuming that you are lsp-origami user you might be interested in writing auto configuration.

leungbk:
> Assuming that you are lsp-origami user you might be interested in writing auto configuration.

I think `orig-folding-range` refers to the original folding range returned by the server, not anything to do with `origami.el`.

The `origami.el` set up has been mentioned already in the `lsp-origami` README. Should we mention it in this repo also?

yyoncho:
> The `origami.el` set up has been mentioned already in the `lsp-origami` README. Should we mention it in this repo also?

It is fine to mention it in the readme. What I am saying is that we can change lsp-origami.el to auto-configure itself like we auto-configure completion, etc. E. g. have a variable lsp-origami-enable which when it is true will result in lsp-mode autoconfiguring origami when the server has support for folding range.

PR #2098, opened 2020-08-19
auto-install: support dockerfile, intelephense, vetur, vim, yaml

kiennq:


lsp-dockerfile.el
```diff
@@ -36,20 +36,29 @@
   :link '(url-link "https://github.com/rcjsuen/dockerfile-language-server-nodejs"))
 
 (defcustom lsp-dockerfile-language-server-command
-  '("docker-langserver" "--stdio")
+  '(nil "--stdio")
```
yyoncho, line 39: this being nil is hard to explain. Can you keep the previous value and change `(cl-first lsp-dockerfile-language-server-command)` to `(executable-find (cl-first lsp-dockerfile-language-server-command))` in the connection function?

kiennq: Done

lsp-vimscript.el
```diff
@@ -31,7 +31,7 @@
   :group 'lsp-mode
   :link '(url-link "https://github.com/iamcco/vim-language-server"))
 
-(defcustom lsp-clients-vim-executable '("vim-language-server" "--stdio")
+(defcustom lsp-clients-vim-executable '(nil "--stdio")
```
yyoncho, line 34: You have missed that place.

yyoncho (approved):

lsp-vimscript.el
kiennq: Thanks

Issue #2039, opened 2020-08-10
Missing space in mode line


muirdm:
After updating recently there is no longer a space after the code action mode line (missing space after the light bulb):
![Screen Shot 2020-08-10 at 11 36 15 AM](https://user-images.githubusercontent.com/11356572/89817970-d9919800-dafd-11ea-90f1-df3fac96d113.png)

Also I am having a heck of a time trying to disable the mode line diagnostic info. I set `(setq lsp-modeline-diagnostics-enable nil)` but as you see in the screenshot the diagnostic numbers still appear. I restarted Emacs and one buffer no longer shows diagnostics, but this buffer still does.

ericdallo:
That was proposital thinking that each node on the modeling should add a prefix string. But I don't know if each node follow this pattern...

what do you think @kiennq ?

muirdm:
Currently I'm using doom modeline with a config like this:

```
(doom-modeline-def-modeline 'main
  '(bar matches buffer-info remote-host buffer-position selection-info)
  '(bar space misc-info debug lsp minor-modes major-mode process vcs checker))
```

I have a `space` between the purple `bar` and the `misc-info`, but maybe instead that space needs to be in the clock segment?

This is my current global-mode-string:
```
((t
  (:eval lsp-modeline--code-actions-string))
 (t
  (:eval
   (-keep #'lsp--workspace-status-string
          (lsp-workspaces))))
 (t
  (:eval
   (lsp-modeline--diagnostics-update-modeline)))
 ((t lsp-java-progress-string))
 display-time-string)
```

kiennq:
@muirdm Yes, I think you will need to add a space in the clock segment instead.
The reason we adding a prefix space only is to follow with pattern that's already inside `lsp-mode`, so that the item on mode line is showing up predictably.

Also, it seems that mode line diagnostic info is indeed disabled, since from `global-mode-string`, that info should be **behind**  code action.
What's the value of `global-mode-string` in the buffer that mode line diagnostic still showing?

muirdm:
When I turn off my fancy modeline and use the default, I still see no space between the light bulb and the clock:
![Screen Shot 2020-08-10 at 6 57 02 PM](https://user-images.githubusercontent.com/11356572/89848565-48411680-db3b-11ea-825a-bc8cde31b351.png)

> What's the value of global-mode-string in the buffer that mode line diagnostic still showing?
[...]

PR #2100, opened 2020-08-19
Add trailing space to separate modeline diagnostics

danielmartin:
Our mode-line strings are added via `add-to-list` with `APPEND` set to nil,
so that means they are always placed at the beginning of `global-mode-string`.
It's a good idea to put things at the beginning to increase the chance that
users see the diagnostics count even in small frames or with a lot of minor modes.

However, if we want to play nice with the rest of things that can be part of `global-mode-string`
(like `display-time-mode`), then we need to append, and not prepend, a space after
our mode-line string.

This fixes a problem where the diagnostics string may overlap with things like
`display-time-mode'.

Screenshot 2020-08-19 at 14 38 48

### DIFF
--- a/lsp-modeline.el
+++ b/lsp-modeline.el
@@ -139,7 +139,7 @@
                                                         (lsp-execute-code-action (lsp--select-action actions))))))
                          built-string)
     (unless (string= "" built-string)
-      (concat " " built-string))))
+      (concat built-string " "))))
 
 (defun lsp--modeline-update-code-actions (actions)
   "Update modeline with new code ACTIONS."
@@ -252,7 +252,7 @@ The `:global' workspace is global one.")
   (cl-labels ((calc-modeline ()
                              (let ((str (lsp-modeline-diagnostics-statistics)))
                                (if (string-empty-p str) ""
-                                 (concat " " str)))))
+                                 (concat str " ")))))
     (setq lsp-modeline--diagnostics-string
           (cl-case lsp-modeline-diagnostics-scope
             (:file (or lsp-modeline--diagnostics-string


ericdallo:
I think this is the same issue from [here](https://github.com/emacs-lsp/lsp-mode/issues/2039). 
We are adding spaces only on the prefix of the segments as @kiennq said it [here](https://github.com/emacs-lsp/lsp-mode/issues/2039#issuecomment-671657044)

danielmartin:
> I think this is the same issue from here.
> We are adding spaces only on the prefix of the segments as @kiennq said it here

The problem is that `display-time-mode` does not follow that convention, so the overlap happens in some scenarios.

Unfortunately I don't see an elegant solution to this problem. I see that `display-battery-mode` also overlaps the display time. However, it's a little bit better because the battery percentage is inside square brackets:

Screenshot 2020-08-19 at 22 54 02

Maybe we can write the number of diagnostics inside square brackets as well. Another idea: `display-battery-mode` exposes a customisable `battery-mode-line-format`; if we do something similar for diagnostics, we may let users decide whether to display errors, or errors and warnings, and add the delimiters they want, using a format string.

ericdallo:
I see... I like the diagnostics-modeline-format idea instead of the brackets one

danielmartin:
> I see... I like the diagnostics-modeline-format idea instead of the brackets one

I can rework the PR to add that customisable modeline format instead.

yyoncho:
I wonder what is the standard solution here. Apparently none of the packages that add something to global-mode-string appends any separators.

jcs090218:
Maybe we can add a custom variable that holds all the minor modes that needed to add a space? And we could just simply check if one of the minor mode is active then we add a space to it. With this solution, even the user encountered this issue; they are able to fix it themselves in the first place.

danielmartin:
I've taken a look at this issue again and I saw two things:

- We always add our mode line string to the beginning of `global-mode-string`.
- We _prepend_ a space before the text.

If I'm not mistaken, we should _append_ a space instead, to ensure that lsp-mode plays 
well with the rest of the mode line strings.

This fixes the problem with `display-time-mode` (and possibly other modes as well).

ericdallo (approved):

PR #2101, opened 2020-08-20
Fix not a number or marker error

wyuenho:
There's a bug in `lsp--symbol-to-hierarchical-imenu-elem` where symbol children don't get filtered out but `lsp--collect-lines-and-cols` only returns the lines and cols of filtered children, so when `lsp--symbol-to-hierarchical-imenu-elem` traverses down to the leaves not in the `lsp--line-col-to-point-hash-table`, a `nil` is recorded in some imenu items and results in an error when selecting those.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6072,12 +6072,13 @@ an alist
 
   (\"symbol-name\" . ((\"(symbol-kind)\" . start-point)
                     cons-cells-from-children))"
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
I will keep it open for some time in case @leungbk wants to review as well.

Issue #2099, opened 2020-08-19
Sort xref results


muirdm:
Currently gopls returns find-references, find-implementations, go-to-definition, etc. results in a random/useless order. I opened https://github.com/golang/go/issues/40904 to discuss, but previously they have not been receptive to ordering on the server.

It would be useful if lsp-mode would sort the results for me. I propose we sort by file name and then by position within file. For file name, I propose we prefer the origin file, then we prefer files within the same directory (sorted lexicographically), then we just sort remaining files lexicographically.

yyoncho:
Make sense. As a side note, you may wanna try `M-x lsp-treemacs-references` and related.

muirdm:
> lsp-treemacs-references

Interesting. Does it support `(next-error)` functionality to cycle through results? How does it order results?

Also, I got an error trying lsp-treemacs-references on a larger project:
```
Debugger entered--Lisp error: (wrong-type-argument stringp nil)
  directory-file-name(nil)
  (file-name-nondirectory (directory-file-name path))
  f-filename(nil)
...
```

yyoncho:
> Interesting. Does it support (next-error) functionality to cycle through results? 

Do you mean next reference? I think that treemacs might have next-node or something. 

> How does it order results?

ATM it is grouping the results, sorting them is a single liner.

> Also, I got an error trying lsp-treemacs-references on a larger project (it worked on a small project):

Is this the full callstack? Can you provide recipe to reproduce that? Most likely filename is missing for some of the results.

muirdm:
@leungbk sorry, I was playing around with this today and didn't assign myself. I didn't do any heuristics beyond basic file/line/column sorting.

leungbk:
> @leungbk sorry, I was playing around with this today and didn't assign myself. I didn't do any heuristics beyond basic file/line/column sorting. It was a learning experience for me so no hard feelings if you do/did something better sweat_smile

It's fine, go ahead and PR when you're ready.

PR #2102, opened 2020-08-20
Sort xref items.

muirdm:
Sort xref items by file name, then by line number, then by column.
Some servers such as gopls return items in a random order so this will
at least make things consistent.

There are other heuristics we could try like preferring the current
file/directory, disfavoring library files, etc., but let's just start
with this.

Updates #2099.


lsp-mode.el
```diff
@@ -4462,14 +4470,24 @@ and the position respectively."
...
+         (seq-map #'get-xrefs-in-file)
+         (apply #'nconc))))
+
+(defun lsp--location-before-p (a b)
+  "Sort first by file, then by line, then by column."
+  (let ((a-uri (lsp--location-uri a))
+        (b-uri (lsp--location-uri b)))
+    (if (not (string= a-uri b-uri))
+        (string< a-uri b-uri)
+      (-let (((&Range :start (&Position :line a-line :character a-char)) (lsp--location-range a))
```
yyoncho, line 4486: there is `lsp--position-compare`

lsp-mode.el
```diff
@@ -4462,14 +4470,24 @@ and the position respectively."
...
-               (->> locations
-                    (seq-group-by (-compose #'lsp--uri-to-path #'lsp:location-uri))
-                    (seq-map (-rpartial #'get-xrefs-in-file nil)))
-             (->> locations
-                  (seq-group-by (-compose #'lsp--uri-to-path #'lsp:location-link-target-uri))
-                  (seq-map (-rpartial #'get-xrefs-in-file t)))))))
+
+    (->> locations
+         (seq-sort #'lsp--location-before-p)
+         (seq-group-by (-compose #'lsp--uri-to-path #'lsp--location-uri))
```
yyoncho, line 4476: not a big deal, but generally, group-by then sort should be faster and does not depend on group-by keeping the order of the items.

yyoncho (approved):
Looks nice and clean. Some nitpicking:

lsp-mode.el
yyoncho: I guess it is fine to stay as it is.

lsp-mode.el
muirdm: Done. I tweaked a few related things to be clearer.

lsp-mode.el
```diff
@@ -3785,28 +3785,28 @@ interface TextDocumentEdit {
            (lsp-with-filename file-name
              (lsp--apply-text-edits (lsp:text-document-edit-edits edit))))))))
 
-(lsp-defun lsp--position-compare ((&Position :line left-line
```
yyoncho, line 3788: why did you change the convention from left/right to a/b? It is used here, and in 4+ packages in emacs-lsp organization. Also, is reversing the result of compare function neccessary?

muirdm: I mentioned it in my commit message:
```
I reused but tweaked lsp--position-compare:
 - Invert comparison so t is returned if the first param is less (I
   think this is more conventional).
 - Rename to lsp--position-before-p and make doc string clearer.
 - Rename args from "left" and "right" to "a" and "b" (left/right is
   confusing since it suggests a positional relationship).
```

"compare" gives no indication of how it compares the params. I read the doc string but it also did not indicate, so I had to examine the function. It turned out to be the opposite of what I would have guessed. To make it clearer, I renamed it `...less-p` indicating it tests if the first param is less than the second (and I flipped the logic to match). I think "less" is more conventional in programming than "more" (e.g. when you call "sort" by default it arranges in ascending order in most languages).

I found "left" and "right" confusing because it seemed like it was saying one of the positions was on the left (in the buffer), and one on the right. I switched to "a" and "b" since those do not suggest any sort of relationship between the two parameters.

yyoncho: I understand that but when you do these changes, with no tests to prove that there is no change in the behaviour I have to go there and do manual testing for all of the changed positions. In this case, lsp--text-edit-sort-predicate should perform the stable sort and if it doesn't this will break applying changes. This is one of the critical sections in the application that has caused a lot of trouble in the past.

> I found "left" and "right" confusing because it seemed like it was saying one of the positions was on the left (in the buffer), and one on the right. I switched to "a" and "b" since those do not suggest any sort of relationship between the two parameters.

Left and right are pretty standard for sort related functions, please revert that.

muirdm: So there are no existing tests for the critical behavior? I figured if the tests passed I didn't break anything important. I can undo all the last changes.

kiennq: If the doc string for the function is not clear, you're welcome to change it. That would break nothing and you still get the desired behavior.

muirdm: Undid changes and clarified lsp--position-compare doc string.

PR #2106, opened 2020-08-22
lsp-completion-filter: allow to customize initial sort or not

kiennq:


lsp-completion.el
```diff
@@ -402,7 +408,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
                                            it))))
                         (markers (list bounds-start (copy-marker (point) t)))
                         (prefix (buffer-substring-no-properties bounds-start (point)))
-                        (lsp-completion--no-reordering t))
+                        (lsp-completion--no-reordering (not lsp-completion-sort-initial-results)))
```
muirdm, line 411: Did you mean to change the default behavior here? It wasn't clear from your commit message what the use case is and whether you meant to make it sort by default.

kiennq: Yes, the default behavior is changed so the everything will be sorted and filtered by client.
There's discussion on gitter and this PR purpose is:
- Change the default behavior to align with VsCode
- Add option so user can customize their experiences as they like.

muirdm: Thanks for explaining. I'm not on gitter because I don't want to give it access to all my GH organizations. Anyway, I don't mean to single you out, but I think this project as a whole needs to do a better job including the "why" in commit messages. Maybe I will start a discussion to see what others think.

yyoncho: > I'm not on gitter because I don't want to give it access to all my GH organizations.

You may join gitter in a lot of different ways.

In a similar note, we should start including a message in changelog each time we change the default behavior.

PR #2109, opened 2020-08-22
Move clients to folder

ericdallo:
After [this](https://github.com/melpa/melpa/pull/7084) we can now reorganize the clients structure.


ericdallo:
@kiennq I'm kind of stuck on those tests, I already changed to load the clients folder but for some reason CI doesn't find it

leungbk:
> @kiennq I'm kind of stuck on those tests, I already changed to load the clients folder but for some reason CI doesn't find it

I tried running CI on my fork just now with your branch. Everything passes when you `(require 'lsp-pyls)` in `lsp-integration-test.el`

kiennq:
> I tried running CI on my fork just now with your branch. Everything passes when you `(require 'lsp-pyls)` in `lsp-integration-test.el`

Is there a real break on that. I mean `lsp-pyls` should be required with `lsp-client-packages` already.
Else we will have some user breaking change

yyoncho:
> Is there a real break on that. I mean `lsp-pyls` should be required with `lsp-client-packages` already.
> Else we will have some user breaking change

We will see when melpa kicks in. But generally the fix is odd, because M-x lsp will also do require lsp-pyls.

ericdallo:
Thanks for the help @leungbk

iitalics:
This seems like it breaks lsp-mode for me. None of the clients can be found, so i get errors like `File mode specification error: (file-missing Cannot open load file No such file or directory lsp-rust)`.

ericdallo:
@iitalics  make sure you have byte-compiled all lsp files, for doom-emacs you probably need to `rm -rf ~/.emacs.d/.local/straight/build/lsp-mode`

iitalics:
This is a fresh lsp-mode install through straight + use-package on otherwise vanilla emacs. Am I supposed to run additional commands to compile stuff?

ericdallo:
if you are using straight recipe manually, make sure the `clients` folder are included on your recipe, check here for the recipe: https://github.com/melpa/melpa/blob/master/recipes/lsp-mode#L1-L4

iitalics:
I see, I will do that.  Is there any way to get this simple invocation to work in the future?
```
(use-package lsp-mode)
```
Or at least could you document more clearly how to setup lsp-mode through use-package?

iitalics:
This has worked in the past and it is odd to see it suddenly break on a new install.

ericdallo:
We have [this documentation](https://emacs-lsp.github.io/lsp-mode/page/installation/) and [this PR waiting to be merged](https://github.com/emacs-lsp/lsp-mode/pull/2003)

iitalics:
Ah, thanks.

kiennq:
Unless you specify the recipe yourselves when using `straight.el` or you have a stale MELPA recipe repository, `straight.el` should also pull recipe from MELPA.
https://github.com/raxod502/straight.el/blob/master/straight.el#L2662
@raxod502 for more advices.

raxod502:
Yes. I will not claim that `straight.el` has perfect MELPA compatibility, but MELPA compatibility is far from the most likely explanation for issues like this. As @kiennq says, you should make sure you have pulled the latest version of MELPA so that you have the updated recipe.

PR #2112, opened 2020-08-25
Fallback default-directory when project and projectile return nil

KaratasFurkan:
When using `lsp-auto-guess-root`, this helps a lot when jumping source code of third party python packages and also when working on a small, dummy project with no git repository in it.

AFAIK, eglot use the same approach, too.


yyoncho (approved):

yyoncho:
Thank you. I think that we had that in the past, seems like it was deleted by mistake.

PR #2116, opened 2020-08-26
fix(eslint): only check for location when first command is node

danielpza:
If the command is not in the form of `node path/to/eslint/server` it shouldn't check for the second argument to exist. eg
I'm using a fork of the vscode eslint server which is just an executable https://github.com/danielpza/vscode-eslint-server.

### DIFF
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -236,7 +236,7 @@ source.fixAll code action."
     (let* ((command-name (f-base (f-filename (cl-first lsp-eslint-server-command))))
            (first-argument (cl-second lsp-eslint-server-command))
            (first-argument-exist (and first-argument (file-exists-p first-argument))))
-     (if (equal command-name "node") first-argument-exist (not (not (executable-find (cl-first lsp-eslint-server-command))))))))
+     (if (equal command-name "node") first-argument-exist (executable-find (cl-first lsp-eslint-server-command))))))
   :activation-fn (lambda (filename &optional _)
                    (or (string-match-p (rx (one-or-more anything) "."
                                            (or "ts" "js" "jsx" "tsx" "html" "vue"))
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -234,7 +234,7 @@ source.fixAll code action."
    (lambda () lsp-eslint-server-command)
    (lambda ()
      (or
-      (not (string-equal (cl-first lsp-eslint-server-command) "node"))
+      (not (equal (f-base (f-filename (cl-first lsp-eslint-server-command))) "node")) ;; if first command is not node then assume it's good
       (and (cl-second lsp-eslint-server-command)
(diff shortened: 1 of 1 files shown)


clients/lsp-eslint.el
```diff
@@ -233,8 +233,10 @@ source.fixAll code action."
   (lsp-stdio-connection
    (lambda () lsp-eslint-server-command)
    (lambda ()
-     (and (cl-second lsp-eslint-server-command)
-          (file-exists-p (cl-second lsp-eslint-server-command)))))
+     (or
+      (not (string-equal (cl-first lsp-eslint-server-command) "node"))
```
yyoncho, line 237: can you use `(f-base (f-filename (cl-first lsp-eslint-server-command)))`  to handle the case when the user has provided full path to node? Also, maybe string-equal has to be changed to contains? IDK how common is to have node binary in format node12.exe or something like that.

danielpza: @yyoncho updated!

clients/lsp-eslint.el
```diff
@@ -233,8 +233,10 @@ source.fixAll code action."
   (lsp-stdio-connection
    (lambda () lsp-eslint-server-command)
    (lambda ()
-     (and (cl-second lsp-eslint-server-command)
-          (file-exists-p (cl-second lsp-eslint-server-command)))))
+     (or
+      (not (equal (f-base (f-filename (cl-first lsp-eslint-server-command))) "node")) ;; if first command is not node then assume it's good
```
kiennq, line 237: This line is too long, you should split the comment into different line

danielpza: done!

clients/lsp-eslint.el
```diff
@@ -233,8 +233,10 @@ source.fixAll code action."
   (lsp-stdio-connection
    (lambda () lsp-eslint-server-command)
    (lambda ()
-     (and (cl-second lsp-eslint-server-command)
-          (file-exists-p (cl-second lsp-eslint-server-command)))))
+    (let* ((command-name (f-base (f-filename (cl-first lsp-eslint-server-command))))
+          (first-argument (cl-second lsp-eslint-server-command))
+          (first-argument-exist (and first-argument (file-exists-p first-argument))))
+     (if (equal command-name "node") first-argument-exist t))))
```
yyoncho, line 239: if the first element is not `node`, then we should check if the first element exists in the `else` block, right?

danielpza: how can I check that? the command I'm expecting is something like `eslint-server` instead of the path to the command.

yyoncho: I think that you should do something like `(executable-find (cl-first (lsp-eslint-server-command)))` in the else block.

danielpza: done!

yyoncho (approved):

yyoncho:

PR #2121, opened 2020-08-27
Add lsp-jedi to language index

wyuenho:


mkdocs.yml
```diff
@@ -70,6 +70,7 @@ nav:
     - PHP (felixbecker): page/lsp-php.md
     - Powershell: page/lsp-pwsh.md
     - PureScript: page/lsp-purescript.md
+    - Python: page/lsp-jedi.md
```
ericdallo, line 73: Actually @wyuenho, could you add some sufix to the key? something like: 
```suggestion
    - Python (jedi): page/lsp-jedi.md
```

wyuenho: I feel like we should add (Palantir) to pyls instead, and steer people to start with lsp-jedi first. Pyls is so slow that almost no one uses it anymore. What do you think?

ericdallo: I see no problem with that, but I'm not a heavy python user, maybe @mpanarin could confirm that?

mpanarin: Are we really gonna argue about what should go first? :\
As well as saying `almost no one uses it anymore` nothing else but a stretch...

Anyway, I am fine with adding `(Palantir)` to pyls and `(jedi)` to this ls.

ericdallo: @wyuenho, Could you rename Pyls and jedi so?

wyuenho: At risk of being argumentative, I must say @mpanarin you are the only one being argumentative here.

ericdallo:
Thank you!

PR #2123, opened 2020-08-30
lsp-completion: Running command after insert selected candidate

### LINKED ISSUE #2033: [Completion] Running command after insert selected completion item
We should also run `command?` in completion item after inserting it.

kiennq:
Fix #2033

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -469,7 +469,8 @@ Others: TRIGGER-CHARS"
                        'lsp-completion-prefix prefix)
                (text-properties-at 0 candidate))
               ((&CompletionItem :label :insert-text? :text-edit? :insert-text-format?
-                                :additional-text-edits? :keep-whitespace?)
+                                :additional-text-edits? :keep-whitespace?
+                                :command?)
                item))
         (cond
          (text-edit?
@@ -500,6 +501,14 @@ Others: TRIGGER-CHARS"
                  (funcall callback
                           (lsp:completion-item-additional-text-edits? resolved-item)))
                cleanup-fn))))
+
+        (when command?
+          (condition-case-unless-debug err
+              (lsp--execute-command command?)
+            (error
+             (lsp--error "Please open an issue in lsp-mode for implementing `%s'.\n\n%S"
+                         (lsp:command-command command?)
+                         err))))
 
         (when (and lsp-signature-auto-activate
                    (lsp-feature? "textDocument/signatureHelp"))


lsp-completion.el
```diff
@@ -501,6 +502,14 @@ Others: TRIGGER-CHARS"
                           (lsp:completion-item-additional-text-edits? resolved-item)))
                cleanup-fn))))
 
+        (when command?
+          (lsp-execute-command
```
yyoncho, line 506: let's use `lsp-execute-code-action`. We should drop `lsp-execute-command` since it does not work fine when there are multiple workspaces.

yyoncho: I mean `lsp--execute-command`

kiennq: Okay.
I think `lsp--execute-command` will also suffered from multiple workspaces.
Since it will only getting command from first action handler in `lsp--find-action-handler`.
Just asking but we will not care about `lsp-execute-command` anymore (thus making generic `lsp-executable-command` become obsoleted) is that correct? Since I saw that `rls` still registering its action handlers via `lsp-execute-command` and the `:action-handlers` seems to point to non-existed function.

yyoncho: > I think lsp--execute-command will also suffered from multiple workspaces.

It won't, once we do it right.

1. Local command execution: The command ids are unique(mirroring the VScode internal API). Furthermore, in tramp/docker scenarios you can register one server with different server id's

2. Remote command execution: we should check executeCommandProvider.commands when performing requests to the server. So by default `lsp--find-workspaces-for` should use that to determine which server to pick for `lsp--execute-command` in case there is no local handler.

lsp-mode.el
```diff
@@ -5093,8 +5101,8 @@ It will show up only if current point has signature help."
 
 (lsp-defun lsp--execute-command ((action &as &Command :command :arguments?))
   "Parse and execute a code ACTION represented as a Command LSP type."
-  (-if-let* ((action-handler (lsp--find-action-handler command)))
-      (funcall action-handler action)
+  (-if-let* ((action-handlers (lsp--find-action-handlers command)))
```
yyoncho, line 5104: the IDs are unique, see https://github.com/microsoft/vscode-eslint/blob/d1a6ff2b44c0504820e44825b2e31fd6489ebbc2/client/src/extension.ts#L1484

and here are vscode docs on that:

```
		 * Registers a command that can be invoked via a keyboard shortcut,
		 * a menu item, an action, or directly.
		 *
		 * Registering a command with an existing command identifier twice
		 * will cause an error.
```

lsp-completion.el
```diff
@@ -501,6 +502,8 @@ Others: TRIGGER-CHARS"
                           (lsp:completion-item-additional-text-edits? resolved-item)))
                cleanup-fn))))
 
+        (when command? (lsp--execute-command command?))
+
         (when (and lsp-signature-auto-activate
                    (lsp-feature? "textDocument/signatureHelp"))
```
yyoncho, line 508: this is also related to the PR. Should we remove it and let the server ask us to do that?

lsp-mode.el
kiennq: Nice, I didn't know that

lsp-completion.el
kiennq: Since we have user options on this, I would say just leave this as it and maybe changing the default to `nil`.
WDYT?

lsp-completion.el
kiennq: Okay

lsp-completion.el
yyoncho: Make sense. We should also look for the editor specific command for triggering completion/signature help(I am not in front of my PC) and implement them in lsp-mode. I think that dart server sends such commands.

kiennq: @ericdallo

ericdallo: I'm not sure dart has that, [this is the dart server lsp documentation](https://github.com/dart-lang/sdk/blob/master/pkg/analysis_server/tool/lsp_spec/README.md).

yyoncho: according to https://github.com/microsoft/language-server-protocol/issues/905 the commands are: editor.action.triggerSuggest and editor.action.triggerParameterHints . I think that in order to handle this we should introduce something like  `lsp--default-notification-handlers` but for actions.  I am not sure if there is something that can trigger those.

kiennq: I would put that in another PR since it will be another scope not covered in this. To make sure not braking user, we can turn `lsp-signature-auto-activate` back to `t` and disable it once the default notification handlers is implemented.
WDYT?

lsp-mode.el
```diff
@@ -990,7 +990,7 @@ They are added to `markdown-code-lang-modes'")
   :group 'lsp-mode
   :package-version '(lsp-mode . "6.2"))
 
-(defcustom lsp-signature-auto-activate t
+(defcustom lsp-signature-auto-activate nil
```
kiennq, line 993:
```suggestion
(defcustom lsp-signature-auto-activate t
```

lsp-completion.el
yyoncho: I am fine if you are going to work on that after this PR. Otherwise, we risk getting errors on after completion because we will call the server and it will show an error.

yyoncho (approved):

lsp-completion.el
yyoncho: Also, I am it will be nice if we find a server that actually sends these commands. I am not familiar with that part of the spec so we might need declaring client capability to enable particular editor command.

kiennq: I found this https://github.com/scalameta/metals/pull/552, but we need to disguised as VsCode for it actually works

yyoncho: I suggest opening scala metals bug report and eventually testing with rust analyzer - https://github.com/rust-analyzer/rust-analyzer/blob/master/crates/rust-analyzer/src/to_proto.rs#L212

kiennq: Checked it. We should wrap `lsp--execute-command` with demoted error to prevent blocking user

yyoncho: Yes, we can have message like "When you see an error here please open an issue in lsp-mode for implementing the method "XX"

kiennq: Done

PR #2124, opened 2020-08-30
Add windows tests CI

ericdallo:


Makefile
```diff
@@ -60,9 +77,16 @@ lint:
 		--eval $(LINT) \
 		*.el
 
-test:
+unix-test:
 	$(CASK) exec ert-runner -L . -L clients  -t '!no-win' -t '!org'
 
+windows-test:
+	@$(EMACS) -Q --batch \
```
kiennq, line 84: I would make a `boostrap.el` file for test like this
``` elisp
(require 'package)

(let* ((package-archives '(("melpa" . "https://melpa.org/packages/")
                           ("gnu" . "https://elpa.gnu.org/packages/")))
       (no-byte-compile t)
       (user-emacs-directory (expand-file-name (make-temp-name ".emacs.d")
                                               "~"))
       (package-user-dir (expand-file-name (make-temp-name "tmp-elpa")
                                           user-emacs-directory))
       (custom-file (expand-file-name "custom.el" package-user-dir))
       (deps '(aio helm dash dash-functional f
               (jsonrpc :fetcher url :url "http://git.savannah.gnu.org/cgit/emacs.git/plain/lisp/jsonrpc.el"))))
  (package-initialize)

  ;; bootstrap quelpa
  (unless (package-installed-p 'quelpa)
    (with-temp-buffer
      (url-insert-file-contents "https://github.com/quelpa/quelpa/raw/master/quelpa.el")
      (eval-buffer)
      (quelpa-self-upgrade)))

  (mapc (lambda (pkg)
          (quelpa pkg)
          (require (if (listp pkg) (car pkg) pkg)))
        deps)
  (add-hook 'kill-emacs-hook `(lambda ()
                                (delete-directory ,user-emacs-directory t))))

(provide 'bootstrap)
```

Then run the test with 
``` makefile
test:
	$(EMACS) -Q -batch $(LOADPATH) \
		-l test/bootstrap.el \
		-l test/test-utils.el \
		-f ert-run-tests-batch-and-exit
```

ericdallo: It looks a good idea, I'll just name as `windows-bootstrap` as this happens only for windows, windows doesn't support cask.

kiennq: yeah, it can potentially support both `unix` and `windows` though. Bypassing cask completely.

ericdallo: Wel, cask is our dependency manager, I'd like to keep it for unix, we need this bootstrap just for windows

yyoncho (approved):

test/windows-bootstrap.el
```diff
@@ -0,0 +1,33 @@
+;;; windows-bootstrap.el --- Windows test bootstrap -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2020
+;;
+;; Author:  <http://github/greg>
```
kiennq, line 5: These information needs to be updated

Makefile
```diff
@@ -3,6 +3,8 @@ SHELL := /usr/bin/env bash
 EMACS ?= emacs
 CASK ?= cask
 
+EXTRA ?= "(progn)"
```
kiennq, line 6: This is not used?

Makefile
```diff
@@ -3,6 +3,8 @@ SHELL := /usr/bin/env bash
...
+
 INIT="(progn \
```
kiennq, line 8: This is also not used?

kiennq (approved):

test/windows-bootstrap.el
ericdallo: My bad

Makefile
ericdallo: Yes, removed.

Makefile
ericdallo: This is still used for the `make lint`

PR #2131, opened 2020-09-02
Convert symbol-informations -> document-symbol

ericdallo:
Convert the deprecated SymbolInformation -> DocumentSymbol


lsp-mode.el
```diff
@@ -5506,17 +5506,26 @@ perform the request synchronously."
...
   "Convert SYMBOLS-INFORMATIONS to symbols hierarchy on CURRENT-POSITION."
   (seq-filter (-lambda ((symbol &as &SymbolInformation :location (&Location :range (&Range :start start-position
                                                                                            :end end-position))))
                 (when (and (lsp--position-compare current-position start-position)
                            (lsp--position-compare end-position current-position))
-                  symbol))
+                  (lsp--symbol-information->document-symbol symbol)))
               symbols-informations))
 
 (defun lsp--symbols->symbols-hierarchy (symbols)
```
leungbk, line 5531:
```suggestion
(defun lsp--symbols->symbols-hierarchy (symbols)
```
```suggestion
(defun lsp--symbols->document-symbols-hierarchy (symbols)
```

ericdallo: Thanks :sweat_smile:

lsp-mode.el
```diff
@@ -5506,27 +5506,36 @@ perform the request synchronously."
...
     (if children?
-        (cons symbol (lsp--document-symbols->symbols-hierarchy children? current-position))
+        (cons symbol (lsp--document-symbols->document-symbols-hierarchy children? current-position))
       (when symbol
         (list symbol)))))
 
-(defun lsp--symbols-informations->symbols-hierarchy (symbols-informations current-position)
+(lsp-defun lsp--symbol-information->document-symbol ((&SymbolInformation :name :kind :location :container-name? :deprecated?))
+  "Convert a SymbolInformation to a DocumentInformation"
+  (lsp-make-document-symbol :name name
```
yyoncho, line 5515: you have to set the selection-range as well.

ericdallo: Done @yyoncho

lsp-mode.el
```diff
@@ -5506,27 +5506,37 @@ perform the request synchronously."
...
+                            :kind kind
+                            :range (lsp:location-range location)
+                            :children? nil
+                            :deprecated? deprecated?
+                            :selection-range (lsp:location-range location)
+                            :detail? container-name?))
+
+(defun lsp--symbols-informations->document-symbols-hierarchy (symbols-informations current-position)
   "Convert SYMBOLS-INFORMATIONS to symbols hierarchy on CURRENT-POSITION."
   (seq-filter (-lambda ((symbol &as &SymbolInformation :location (&Location :range (&Range :start start-position
```
leungbk, line 5525: This is insufficient since it doesn't actually perform the expected conversion. You need something like

``` elisp
(mapcan (-lambda ((symbol &as &SymbolInformation :location (&Location :range (&Range :start start-position
                                                                                       :end end-position))))
            (when (and (lsp--position-compare current-position start-position)
                       (lsp--position-compare end-position current-position))
              (list (lsp--symbol-information->document-symbol symbol))))
          symbols-informations)
```

or the equivalent `cl-loop` list comprehension.

leungbk:
I tested this with `flow-lsp`. In addition to the comment above, I also noticed that the headerline symbols did not generally appear in a reasonable order, so we need to sort them here.

yyoncho:
@ericdallo if you are looking for a server to test with `pyls` is an option - it is easy to install, and works with standalone python files.

ericdallo:
Thanks @yyoncho and @leungbk, I'll test with pyls then try to reproduce the bug

leungbk:
Repro: https://github.com/KaTeX/KaTeX//blob/37990cc5b392776cbb7780c9a7ab5e0d16ca1e58/src/parseNode.js#L30

First you need to run `yarn install` in the project root, and, on NixOS, something like

``` shell
  patchelf \
    --set-interpreter "$(cat $(echo $(nix-build --no-out-link -E 'with import <nixpkgs> {}; pkgs.stdenv.cc')/nix-support/dynamic-linker))" \
    ./.yarn/unplugged/flow-bin-npm-0.132.0-941e001cc0/node_modules/flow-bin/flow-linux64-v0.132.0/flow
```

to patch the `flow` executable.

At the exact line above, the breadcrumb components appear out of order. I tested `lsp--symbols-informations->document-symbols-hierarchy` by calling `lsp-breadcrumb-go-to-symbol`.

ericdallo:
I fixed both the bug and the sort. 
Thank you very much for the explanation how to setup flow-ls @leungbk , especially on NixOS :)
I made a sort via ranges, though IMO we just `(reverse)` the list and I think everything would work.

leungbk:
> I fixed both the bug and the sort.

Thanks! I think you forgot to revert the changes you made in `lsp-javascript.el` for testing this.

> Thank you very much for the explanation how to setup flow-ls @leungbk , especially on NixOS :)

NP.

> I made a sort via ranges, though IMO we just `(reverse)` the list and I think everything would work.

If there's no sort order specified by the protocol, wouldn't sorting be the better choice? Does `pyls` also return everything in reverse order?

ericdallo:
> If there's no sort order specified by the protocol, wouldn't sorting be the better choice? Does pyls also return everything in reverse order?

I didn't test with `pyls` but that's what motivated me to implement the sort via ranges instead of just using reverse... 

The documentation doesn't say anything about sorting, but IMO relying on range sort it's the better choice ATM.

leungbk (approved):

lsp-mode.el
```diff
@@ -5506,27 +5506,41 @@ perform the request synchronously."
...
-              symbols-informations))
-
-(defun lsp--symbols->symbols-hierarchy (symbols)
+  (->> symbols-informations
+       (mapcan (-lambda ((symbol &as &SymbolInformation :location (&Location :range (&Range :start start-position
+                                                                                            :end end-position))))
+                 (when (and (lsp--position-compare current-position start-position)
+                            (lsp--position-compare end-position current-position))
+                   (list (lsp--symbol-information->document-symbol symbol)))))
+       (-sort (-lambda ((&DocumentSymbol :range (&Range :start a-start-position :end a-end-position))
```
kiennq, line 5531: Using `sort` will save more memory since `-sort` is duplicating the list

ericdallo:
Ping @kiennq.

kiennq (approved):

PR #2132, opened 2020-09-02
Clean up for lsp-csharp

jcs090218:
Another clean up patch for `lsp-csharp.el`! Basically removes unnecessary syntax like `format`, `if` to `when`, etc.


yyoncho (approved):

yyoncho:
Thank you!

PR #2134, opened 2020-09-02
Clean up some docstrings and fix typos

leungbk:
Ideally, this shouldn't be squashed since the commit updating `.git-blame-ignore-revs` must refer to the correct commit hash to have the desired effect. And if the commits get rebased and wind up with new hashes, the commit updating `.git-blame-ignore-revs` should be amended to use the new commit hash of the typo-fix commit.


yyoncho:
I merged it manually, please let me know if everything is ok.

PR #2135, opened 2020-09-02
Clean up for lsp-haxe.

jcs090218:
Another clean up patch, mainly for indentation and better readability to `lsp-haxe.el` file. There are also documentation missing for many `defcustom` variable, but I didn't bother to add the documentation yet.

### DIFF
--- a/clients/lsp-haxe.el
+++ b/clients/lsp-haxe.el
@@ -66,6 +66,7 @@
 (defcustom lsp-haxe-hxml "build.hxml"
   "The compile file for the haxe project."
   :type 'file
+  :group 'lsp-haxe
   :package-version '(lsp-mode . "7.0"))
 
 ;; https://github.com/emacs-lsp/lsp-mode/blob/150a933694349df960dc8fd7a15e04f5727e6433/lsp-rust.el#L251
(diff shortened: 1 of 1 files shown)


leungbk:
In the future, please try to register style/formatting commits in the `.git-blame-ignore-revs` file. I have a PR updating that file, so I'll register this commit myself.

Thanks for the PR!

jcs090218:
No problem! Just one question, I wonder how you do that? 😕

leungbk:
You can make a series of “Clean up lsp-foo” commits as normal, and then
make one commit in which you paste in those earlier commit hashes.

jcs090218:
Just paste it in manually? I will try with this PR here https://github.com/emacs-lsp/lsp-mode/pull/2136.

jcs090218:
I don't think I am able to get the hashes unless I made the changes? Right? In other word I could not do it? 😕 

I have compared my own branch and master branch (this) and their hashes are different.

leungbk:
> I have compared my own branch and master branch (this) and their hashes are different.

Their hashes are different because the commit appearing on the upstream branch is a squashed version of the commit you made.

> I don't think I am able to get the hashes unless I made the changes? Right? In other word I could not do it?

It's possible, see https://github.com/emacs-lsp/lsp-mode/pull/2134.

I guess it's not really a big deal though, since an alternative way to handling style commits is to have someone periodically add them to the `.git-blame-ignore-revs` file in batches.

PR #2136, opened 2020-09-02
Clean up for lsp-json.

jcs090218:
Another clean up patch for `lsp-json`. Also added one minor docstring.

### DIFF
--- a/clients/lsp-json.el
+++ b/clients/lsp-json.el
@@ -62,7 +62,8 @@
   `(:provideFormatter t
     :handledSchemaProtocols ["file" "http" "https"]))
 
-(defvar lsp-json--major-modes '(json-mode jsonc-mode))
+(defvar lsp-json--major-modes '(json-mode jsonc-mode)
+  "List of supported JSON major mode.")
 
 (defvar lsp-json--schema-associations
   `(:/*.css-data.json ["https://raw.githubusercontent.com/Microsoft/vscode-css-languageservice/master/docs/customData.schema.json"]
@@ -87,13 +88,14 @@
 (defun lsp-json--get-content (_workspace uri callback)
   "Get content from URI."
   (ignore-errors
-    (url-retrieve uri (lambda (_status callback)
-                        (goto-char (point-min))
-                        (re-search-forward "\n\n" nil 'noerror)
-                        (funcall
-                         callback
-                         (decode-coding-string (buffer-substring (point) (point-max))
-                                               'utf-8-unix)))
+    (url-retrieve uri
+                  (lambda (_status callback)
+                    (goto-char (point-min))
+                    (re-search-forward "\n\n" nil 'noerror)
+                    (funcall
+                     callback
+                     (decode-coding-string (buffer-substring (point) (point-max))
+                                           'utf-8-unix)))
                   (list callback))))
 
 (lsp-dependency 'vscode-json-languageserver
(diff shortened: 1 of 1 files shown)


clients/lsp-json.el
```diff
@@ -103,24 +105,26 @@
...
-  :initialized-fn (lambda (w)
-                    (with-lsp-workspace w
-                      (lsp--set-configuration
-                       (ht-merge (lsp-configuration-section "json")
-                                 (lsp-configuration-section "http")))
-                      (lsp-notify "json/schemaAssociations" lsp-json--schema-associations)))
-  :download-server-fn (lambda (_client callback error-callback _update?)
-                        (lsp-package-ensure 'vscode-json-languageserver callback error-callback))))
+  :initialized-fn
+  (lambda (w)
```
kiennq, line 119: Can you keep this `lambda` on the same line with `:initialized-fn`?
I intentionally put that so every new line start with `:` in the same indent. Which objectively easier to see from glance

clients/lsp-json.el
```diff
@@ -103,24 +105,26 @@
...
+    (with-lsp-workspace w
+      (lsp--set-configuration
+       (ht-merge (lsp-configuration-section "json")
+                 (lsp-configuration-section "http")))
+      (lsp-notify "json/schemaAssociations" lsp-json--schema-associations)))
+  :download-server-fn
+  (lambda (_client callback error-callback _update?)
```
kiennq, line 126: Same here

clients/lsp-json.el
```diff
@@ -62,7 +62,8 @@
   `(:provideFormatter t
     :handledSchemaProtocols ["file" "http" "https"]))
 
-(defvar lsp-json--major-modes '(json-mode jsonc-mode))
+(defvar lsp-json--major-modes '(json-mode jsonc-mode)
+  "List of supported JSON major mode.")
```
kiennq, line 66:
```suggestion
  "List of supported JSON major modes.")
```

clients/lsp-json.el
jcs090218: Yeah, I can see that. I personally like to have it on the next line because this section of code will mostly likely runs over columns 80. And I don't prefer that kind of scenario. If you don't agree with me, it's fine and feel free to close it! 😃  I guess this is againsting coding convention to this repo here. 👍

kiennq: Well, I don't think column 80 is a really hard requirement. The monitor now is big, and even with screen split into twos, I still can easily fit a line over 100 col.
If you wan to uphold that, it's fine, but we can always make the function body inside lambda to do that (split second parameter to new line ...).
I appreciate your other changes of adding doc string for `defvar` ..., so just a nit pick here

jcs090218: Not a big deal. I do split screen. I know many people uses one screen and think this isn't a big deal.

This stackoverflow question basically answers what I was trying to say. See https://stackoverflow.com/questions/110928/is-there-a-valid-reason-for-enforcing-a-maximum-width-of-80-characters-in-a-code

Mostly because of the readability and human eyes issue. :)

yyoncho: ATM we don't particular code style defined - up to now it has been more of personal preference and we haven't enforced any coding style rules except maybe closing all brackets on a single line. Also, the requirements for code consistency and style, in general, are a bit relaxed for `lsp-SERVER.el` files.

With lsp-mode being written in a functional point-free style having 80 symbols restriction seems to be hard to implement. I am fine with merging as it is or after applying @kiennq's comments. I wish we don't spend a lot of time discussing formatting issues.

jcs090218: Yeah, I agree with @yyoncho. I don't want to waste the precious time to the whole lsp-team here. If this isn't what you want, feel free to close it! Just let me know the reason! ;)

yyoncho (approved):

kiennq (approved):

PR #2143, opened 2020-09-05
Add optional blacklist for lsp-diagnostics-mode

wyuenho:
Currently lsp-diagnostics is an all-or-nothing setting. There are some LSP servers that advertise support for diagnostics but don't really do much except checking for syntax errors, in those cases it's probably better to turn lsp-diagnostics off just for those modes and let the user set up flycheck or flymake with the appropriate linters themselves.

### DIFF
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -62,6 +62,11 @@ on top the flycheck face for that error level."
   :type '(repeat list)
   :group 'lsp-mode)
 
+(defcustom lsp-diagnostics-disabled-modes nil
+  "A list of major models for which `lsp-diagnostics-mode' should be disabled."
+  :type '(repeat symbol)
+  :group 'lsp-mode)
+
 ;; Flycheck integration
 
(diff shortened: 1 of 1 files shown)


lsp-diagnostics.el
```diff
@@ -62,6 +62,11 @@ on top the flycheck face for that error level."
   :type '(repeat list)
   :group 'lsp-mode)
 
+(defcustom lsp-diagnostics-disabled-modes nil
+  "A list of major models for which `lsp-diagnostics-mode' should be disabled."
+  :type '(repeat symbol)
+  :group 'lsp-mode)
```
yyoncho, line 68: can you add `:version "7.1.0"`?

yyoncho: (`:package-version '(lsp-mode . "7.1.0")` in the correct annotation)

yyoncho (approved):

yyoncho:
Thank you!

PR #2144, opened 2020-09-05
Change Makefile

leungbk:
* Fix `windows-test` target
* Also explicitly declare what to byte-compile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,15 @@ LINT="(progn \
 		(require 'package-lint) \
 		(package-lint-batch-and-exit))"
 
+LSP-FILES := lsp-protocol.el lsp-mode.el lsp.el lsp-completion.el \
+		lsp-diagnostics.el lsp-lens.el lsp-modeline.el \
+		$(wildcard clients/*.el)
+
+TEST-FILES := test/windows-bootstrap.el test/test-helper.el \
+		$(shell ls test/lsp-*.el)
+LOAD-FILE = -l $(test-file)
+LOAD-TEST-FILES := $(foreach test-file, $(TEST-FILES), $(LOAD-FILE))
+
 all:
 	$(CASK) build
 
@@ -32,17 +41,15 @@ unix-compile:
 	@$(CASK) $(EMACS) -Q --batch \
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
-		-f batch-byte-compile \
-		*.el clients/*.el
+		-f batch-byte-compile $(LSP-FILES)
 
 windows-compile:
 	@echo "Compiling..."
 	@$(CASK) $(EMACS) -Q --batch \
-	    -l test/windows-bootstrap.el \
+		-l test/windows-bootstrap.el \
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
-		-f batch-byte-compile \
-		*.el clients/*.el
+		-f batch-byte-compile $(LSP-FILES)
 
 checkdoc:
 	$(eval LOG := $(shell mktemp -d)/checklog.log)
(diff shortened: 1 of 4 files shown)


ericdallo (approved):
The only drawback I see is remember to change all these places when we create a new lsp-*.el file

yyoncho (approved):

leungbk:
@ericdallo I changed the `windows-test` build target. It wasn't running before, not even in your [original PR](https://github.com/emacs-lsp/lsp-mode/runs/1048445273), because of a missing line-continuation backslash:

```
Done (Total of 5 files compiled, 2 skipped in 2 directories)
f ert-runner/run
process_begin: CreateProcess(NULL, f ert-runner/run, ...) failed.
make (e=2): The system cannot find the file specified.

mingw32-make: [Makefile:80: windows-test] Error 2 (ignored)
```

So for now I'm using `ert` instead of `ert-runner` since I don't know how to make `ert-runner` work in the Windows environment.

ericdallo:
Good job @leungbk!
I think now it's working and running the windows tests successfully, right?

lsp-mode.el
```diff
@@ -1631,10 +1631,10 @@ On other systems, returns path without change."
                           uri)
                       ;; `url-generic-parse-url' is buggy on windows:
                       ;; https://github.com/emacs-lsp/lsp-mode/pull/265
-                      (or (and (eq system-type 'windows-nt)
-                               (eq (elt file 0) ?\/)
-                               (substring file 1))
-                          file))))
+                      (if (and (eq system-type 'windows-nt)
+                               (s-prefix? "/c:" file))
```
kiennq, line 1635: This is not really correct, since the path can start from other drive (`/d:`, `/e:` ...), so you probably need to pattern matching it.
What wrong with the original code?

Makefile
```diff
@@ -32,29 +41,26 @@ unix-compile:
...
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
-		-f batch-byte-compile \
-		*.el clients/*.el
+		-f batch-byte-compile $(LSP-FILES)
 
 windows-compile:
 	@echo "Compiling..."
 	@$(CASK) $(EMACS) -Q --batch \
 	    -l test/windows-bootstrap.el \
```
kiennq, line 49: this line should be indented so in the same level with following lines?

kiennq (approved):

lsp-mode.el
leungbk: https://github.com/emacs-lsp/lsp-mode/blob/1805c9caa146f60a5d9109f2a14d3596a5d1228f/test/lsp-common-test.el#L38 and https://github.com/emacs-lsp/lsp-mode/blob/1805c9caa146f60a5d9109f2a14d3596a5d1228f/test/lsp-common-test.el#L54 failed on Windows before making this change, because the leading forward slash in something like "/root/usr/" got trimmed off.

kiennq: I see.
I think we should also fix the test instead of production code. The `"/root/%5E/%60"` is not suppose to be a path in Windows.
Also in https://github.com/emacs-lsp/lsp-mode/pull/265, the reason the first slash is removed in Windows is because of `file:///` which is prefix in Windows will be parse as only `file://` and left the final `/` in the path by `url-generic-parse-url`.
The original check would be enough

kiennq (changes requested):
Can you please check the last comment on `lsp--uri-to-path-1` and address it?

lsp-mode.el
leungbk: OK, I went back to the original check and used `'gnu-linux` as `system-type` in the offending tests.

kiennq (approved):

ericdallo:
Feel free to merge @leungbk

Issue #1894, opened 2020-07-09
company-mode should not hide the dialog when lsp-mode does not have the time handle the call


yyoncho:
At this point `lsp-mode` returns `nil` to `company-mode` when the server was not
able to respond and the user has started typing. This is a workaround of capf
being synchronous. This triggers hiding of completion dialog which later it
displayed again if there is completion result for the longer prefix. We should
eliminate that.

Potential fixed:

1. Throwing something that company-mode expects and it will be treated like "keep
   the dialog open"
2. Implementing traditional(not capf) async company backend. I suspect that the
   described issue does not exist in that case.
3. We may try to return the previous available result. Not sure if this will
   work but we do that on other places like eldoc-mode to avoid blinking.

cc @dgutov @kiennq

kiennq:
We can indeed returns previous value in this case.
@dgutov Do you think that will fix the issue?

yyoncho:
I just want to mention that this issue is hurting the usability of posframe solutions because they are a bit slower in showing/hiding the completion items. At the same time, the solutions with overlays that company offers are not compatible with overlays. So if lsp-mode wants to support lsp protocol flawlessly with we should have working posframe solution.

kiennq:
> I just want to mention that this issue is hurting the usability of posframe solutions because they are a bit slower in showing/hiding the completion items.

I'm aware that create new frame can be slow but posframe showing/hiding is fairly fast since we don't have to recreate the posframe everytime?

dgutov:
> We can indeed returns previous value in this case.

This should work. It's an interesting hack, actually.

Option 2 probably isn't going to help (I never optimized for this scenario together with based frontends, and the overlay-based one hides/shows very quickly), but it's an area that could be improved. And without a major redesign of the loop, internally the change is probably going to behave at lot like your suggestion. ;-)

API-wise, option 1 would be ideal, but I think it needs some support in the core as well (else it would break `M-x completion-at-point`, and you probably care about that, right?). You might as well wait for an async convention in completion tables instead (so that `company-capf` turns it into option 2). Maybe file an Emacs feature request to remind the maintainers about this need.

yyoncho:
> (else it would break `M-x completion-at-point`, and you probably care about that, right?).

Yes. Generally, we should be blocking in completion-at-point and not aborting the computation on user input.

> Maybe file an Emacs feature request to remind the maintainers about this need.

I will do that. Are you aware of any plans to move completion stuff in elpa package?

[...]

PR #2146, opened 2020-09-06
lsp-completion--exit-fn: works with non props candidate

kiennq:
In case of just using default `completion-at-point` (in that case `completion-in-region-function` will be `completion--in-region`), the `:exit-function` will be called with selected item as string without any properties that we built up during filtering phase.
We need a way to re-associated the striped string with the full one.

The propose here is:
- Remember the last completion results in `lsp-completion--last-result`, use that to trace back to full string in `:exit-function`.
- Make `try-completion` will always return `nil` (no exact completion item) so that the selection will always shown and prevent early clean up of completion cache.

This will also server as first step for #1894

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -153,6 +153,9 @@ ignored."
 In the form of plist (prefix prefix-pos items :lsp-items ...).
 When the completion is incomplete, cache contains value of `incomplete'.")
 
+(defvar lsp-completion--last-result nil
+  "Last completion result.")
+
 (defun lsp-completion--clear-cache (&rest _)
   "Clear completion caches."
   (-some-> (and (listp lsp-completion--cache) lsp-completion--cache)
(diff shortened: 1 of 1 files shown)


yyoncho:
> will be called with selected item as string without any properties that we built up during filtering phase.

Isn't that emacs bug?

kiennq:
> > will be called with selected item as string without any properties that we built up during filtering phase.
> 
> Isn't that emacs bug?

Is it? I think there's no guarantee that the string's properties will be preserved when passed to `:exit-function`.

For that's by design since the completion items will be displayed in other buffer in case of default `completion-in-region-function`.
image

Thus it will mostly cannot retain the property when selecting one of them.

Even if this's a bug (not likely), we still need to work-around in `lsp-mode` since it will not be fixed in pre Emacs 28.

yyoncho:
Hm, some of the previous capf implementations used to preserve the properties.

yyoncho:
The text properties used to work. I even filed a helm issue - https://github.com/emacs-helm/helm/issues/2265#issuecomment-576582659 . Do you know what broke them?

kiennq:
AFAIK, `ivy`, `company`, `emacs-helm` are preserving text properties for `:exit-function`.

However, the default function for `completion-in-region-function` (which is `completion--in-region`) doesn't do that.
I've checked Emacs25 and it seems that behavior is there too, so it's pretty much default behavior.
https://github.com/emacs-mirror/emacs/blob/c3ff6712ad24fcf45874dc0665a8606e9b2208a4/lisp/minibuffer.el#L1005

I think you can test this by disable `helm-mode` and manually trigger `completion-at-point` (bind to `C-M-i`), then select item with mouse

yyoncho (approved):

PR #2148, opened 2020-09-07
lsp-completion: use last results in case of interrupted by keyboard

### LINKED ISSUE #1894: company-mode should not hide the dialog when lsp-mode does not have the time handle the call

kiennq:
Fix #1894 
It's done by:
- Make `lsp-*-while-no-input` respects `non-essential`
- Return or throw when `body` of `lsp-*-while-no-input` is interrupted by keyboard.
- Use `lsp-completion--last-result` when completion is interrupted by keyboard.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -76,6 +76,13 @@ ignored."
   :group 'lsp-mode
   :package-version '(lsp-mode . "7.1"))
 
+(defcustom lsp-completion-use-last-result t
+  "Temporarily use last server result when interrupted by keyboard.
+This will help minimize popup flickering issue in `company-mode'."
+  :type 'boolean
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
 (defconst lsp-completion--item-kind
   [nil
    "Text"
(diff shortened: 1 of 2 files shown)


kiennq:
@dgutov

kiennq:
@yyoncho Can you check if the issue of `company-mode` window flickering is resolved?

dgutov:
Looks good. :+1: 

Let me know if you find any issues with this approach.

kiennq:
@yyoncho

yyoncho:
Sorry for the delay. I tested it and it feels like a huge usability improvement!

Due to the fact that this is one of the most critical components we might consider adding a kill switch for this behavior. How hard will be to do so?

yyoncho (approved):

kiennq:
> Due to the fact that this is one of the most critical components we might consider adding a kill switch for this behavior. How hard will be to do so?

What do you mean by kill switch?

yyoncho:
A setting that will turn off newly introduced behaviour.

kiennq:
Added `lsp-completion-use-last-result` for that.

PR #2149, opened 2020-09-07
When cleaning non-local hooks we should not check if buf is alive

yyoncho:
- fixes 2141

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3057,17 +3057,20 @@ CANCEL-TOKEN is the token that can be used to cancel request."
              ;; that all of the captured arguments are the same - in our case
              ;; `lsp--create-request-cancel' will return the same lambda when
              ;; called with the same params.
-             (cleanup-hooks (lambda ()
-                              (when (buffer-live-p buf)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3057,17 +3057,21 @@ CANCEL-TOKEN is the token that can be used to cancel request."
...
+                            (if local
+                                (when (buffer-live-p buf)
+                                  (with-current-buffer buf
+                                    (remove-hook hook
+                                                 (lsp--create-request-cancel
+                                                  id target-workspaces hook buf method cancel-callback)
+                                                 local)))
+                              (remove-hook hook (lsp--create-request-cancel
+                                                 id target-workspaces hook buf method cancel-callback)
+                                           local)))
```
kiennq, line 3072: I think we can replace `local` here and the previous `remove-hook` with its exact value (`nil` and `t`)?

yyoncho: replaced with t and nothing.

kiennq (approved):

Issue #2156, opened 2020-09-07
Add Prolog LSP to list of implementations of semantic highlighting


jamesnvc:
Following up on comment on the release notes here: http://disq.us/p/2bphekw

[My Prolog LSP](https://github.com/jamesnvc/lsp_server) implementation also supports semantic highlighting (related to #1906) & could be added to the list of language servers that implement the new highlighting method.

ericdallo:
Nice @jamesnvc, we can wait for someone to open a PR for that but as you are the maintainer and understands how the server behave, you can follow the documentation on [how to setup a server on lsp-mode](https://emacs-lsp.github.io/lsp-mode/page/adding-new-language/) too :)

yyoncho:
> [My Prolog LSP](https://github.com/jamesnvc/lsp_server) implementation also supports semantic highlighting (related to #1906) & could be added to the list of language servers that implement the new highlighting method.

You mean to the blog entry? 

As a side note feel free to add lsp-client registration for your server to lsp-mode and link your server.

ericdallo:
@yyoncho no, he meant that he would like to see his server on lsp-clients via the blog comment section [here](https://emacs-lsp.github.io/lsp-mode/blog/2020/7.0-release/#comment-5061495056)

PR #2157, opened 2020-09-07
Add Prolog LSP implementation to list of supported languages

jamesnvc:
Following up on #2156

### DIFF
--- a/clients/lsp-prolog.el
+++ b/clients/lsp-prolog.el
@@ -44,8 +44,7 @@
 
 (lsp-register-client
  (make-lsp-client
-  :new-connection
-  (lsp-stdio-connection lsp-prolog-server-command)
+  :new-connection (lsp-stdio-connection (lambda () lsp-prolog-server-command))
   :major-modes '(prolog-mode)
   :multi-root t
   :server-id 'prolog-lsp))
--- a/clients/lsp-prolog.el
+++ b/clients/lsp-prolog.el
@@ -1,0 +1,54 @@
+;;; lsp-prolog.el --- Prolog Client settings -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020  James Cash
+
+;; Author: James Cash <james.nvc@gmail.com>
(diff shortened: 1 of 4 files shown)


ericdallo:
@jamesnvc Just making it clear that if you want, you can migrate the `lsp-register-client` from your readme to here, with that users just need to install lsp-mode and when opening a prolog major mode everything should work :)
Take a look on an already existing client like [lsp-clojure](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-clojure.el) for example, LMK you need more info/help.

jamesnvc:
> @jamesnvc Just making it clear that if you want, you can migrate the `lsp-register-client` from your readme to here, with that users just need to install lsp-mode and when opening a prolog major mode everything should work :)

Oh cool, will do, thank you!

leungbk:
Thanks for your PR! Please make sure to update `lsp-language-id-configuration`.

ericdallo (approved):

clients/lsp-prolog.el
```diff
@@ -0,0 +1,54 @@
...
+                                       "--" "stdio")
+  "The prolog-lsp server command."
+  :group 'lsp-prolog
+  :risky t
+  :type 'list)
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection
+  (lsp-stdio-connection lsp-prolog-server-command)
```
kiennq, line 48: You should make this a lambda that returns `lsp-prolog-server-command`. So user can configure it **after** `lsp-prolog` is loaded.

jamesnvc:
> @kiennq commented on this pull request.
> You should make this a lambda that returns `lsp-prolog-server-command`. So user can configure it **after** `lsp-prolog` is loaded.

Ah, good point! Thank-you

jcs090218:
New language? Cool! Thanks for the PR! 😄

yyoncho (approved):

PR #2158, opened 2020-09-08
Fix lsp-clangd.el to return 'cpp' language when content is oneline signature

zhaozhixu:


clients/lsp-clangd.el
```diff
@@ -226,7 +226,9 @@ returned to avoid that the echo area grows uncomfortably."
                                        "\n")
                                       "```")) nil t nil)
           (progn (narrow-to-region (match-beginning 2) (match-end 2))
-                 (lsp--render-element (lsp-clangd-join-region (point-min) (point-max))))
+                 (lsp--render-element
+                  (ht ("language" "cpp")
```
kiennq, line 230: You should use `lsp-make-marked-string` instead.

zhaozhixu: Fixed it.

kiennq (approved):

danielmartin (approved):
Thanks for the fix! I've tested it and works well. I think I didn't noticed the problem because I don't work with raw pointers in C++. 😅

PR #2159, opened 2020-09-09
lsp-signature-auto-active: add option for activate after completion

kiennq:
Change `lsp-signature-auto-activate` to list of `(:after-completion :on-trigger-char)` so we can granularly control when to trigger signature auto activation

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -521,7 +521,8 @@ Others: TRIGGER-CHARS CANDIDATES"
                          (lsp:command-command command?)
                          err))))
 
-        (when (and lsp-signature-auto-activate
+        (when (and (or (equal lsp-signature-auto-activate t)
+                       (memq :after-completion lsp-signature-auto-activate))
                    (lsp-feature? "textDocument/signatureHelp"))
           (lsp-signature-activate))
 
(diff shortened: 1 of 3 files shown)


ericdallo (approved):

lsp-completion.el
```diff
@@ -521,7 +521,7 @@ Others: TRIGGER-CHARS CANDIDATES"
                          (lsp:command-command command?)
                          err))))
 
-        (when (and lsp-signature-auto-activate
+        (when (and (memq :after-completion lsp-signature-auto-activate)
```
yyoncho, line 524: We should handle `lsp-signagure-auto-active` being `t` and/or document this as a breaking change.

lsp-mode.el
```diff
@@ -2131,7 +2132,8 @@ WORKSPACE is the workspace that contains the diagnostics."
 (defun lsp-toggle-signature-auto-activate ()
   "Toggle signature auto activate."
   (interactive)
-  (setq lsp-signature-auto-activate (not lsp-signature-auto-activate))
+  (setq lsp-signature-auto-activate
+        (if lsp-signature-auto-activate nil '(:on-trigger-char)))
```
yyoncho, line 2136: nit: use unless

yyoncho (commented):
Looks good.

lsp-completion.el
kiennq: Done. Handle `lsp-signature-auto-active` being `t`

yyoncho (approved):

PR #2160, opened 2020-09-09
Trailing space depends on minor-modes activation.

jcs090218:
Proposal for #2100.

### DIFF
--- a/lsp-modeline.el
+++ b/lsp-modeline.el
@@ -57,6 +57,11 @@
 (declare-function all-the-icons-octicon "ext:all-the-icons" t t)
 (declare-function lsp-treemacs-errors-list "ext:lsp-treemacs" t)
 
+(defcustom lsp-modeline-overlap-minor-modes '(display-time-mode)
+  "Minor modes that need to add a space to prevent overlap."
+  :type 'list
+  :group 'lsp-mode)
+
 ;; code actions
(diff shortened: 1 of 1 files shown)


lsp-modeline.el
```diff
@@ -236,7 +252,9 @@ The `:global' workspace is global one.")
   (cl-labels ((calc-modeline ()
                              (let ((str (lsp-modeline-diagnostics-statistics)))
                                (if (string-empty-p str) ""
-                                 (concat " " str)))))
+                                 (concat " " str
```
kiennq, line 255: You will need to check for code action mode line too, and workspace status, download status mode line ...
User can customize and change order of those in `global-mode-string`.
Speaking of that, wouldn't it better to add an option for trailing front/behind space (maybe via a format string) instead of manually adding minor modes to `lsp-modeline--overlap-minor-modes`

jcs090218: Got it! I have thought of that too. I guess that would be another solution? :)

ericdallo:
I think the proposed chages at https://github.com/emacs-lsp/lsp-mode/pull/2100#issuecomment-703317779, should solve this issue, right?

jcs090218:
I guess this PR is no longer needed? Close this now! Thanks!

PR #2161, opened 2020-09-10
Auto detect macos clangd

wyuenho:
Allow LSP to pick up the clangd that comes with Xcode or Developer Command Line Tools automatically without having the user to install LLVM with Homebrew or MacPorts.

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -178,14 +178,6 @@ This must be set only once after loading the clang client.")
   :risky t
   :type '(repeat string))
 
-(defun lsp-clients--clangd-darwin-executable-find (executable &rest args)
-  (when-let* ((executable-path (executable-find executable))
-              (clangd-path
-               (with-temp-buffer
-                 (when (zerop (apply 'call-process executable-path nil t nil args))
-                   (buffer-substring-no-properties (point-min) (point-max))))))
-    (string-trim clangd-path)))
-
 (defun lsp-clients--clangd-command ()
   "Generate the language server startup command."
   (unless lsp-clients--clangd-default-executable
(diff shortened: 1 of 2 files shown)


clients/lsp-clangd.el
```diff
@@ -177,14 +178,24 @@ This must be set only once after loading the clang client.")
   :risky t
   :type '(repeat string))
 
+(defun lsp-clients--clangd-darwin-executable-find (executable &rest args)
```
kiennq, line 181: This is not only applied for darwin but actually anywhere. So we should name this as something more generally.
```suggestion
(defun lsp-clients--clangd-execute (executable &rest args)
```

clients/lsp-clangd.el
```diff
@@ -177,14 +178,24 @@ This must be set only once after loading the clang client.")
...
+  (when-let* ((executable-path (executable-find executable))
```
kiennq, line 182: Add document for this.

lsp-mode.el
```diff
@@ -1519,6 +1519,25 @@ BODY should never return `t' value."
   download-in-progress?
   buffers)
 
+(defun lsp-clients-executable-find (find-command &rest args)
```
kiennq, line 1522: nit: Can't this be named something more generic? This doesn't even reassemble the `execute-find`, both in parameters and implementation.

I would imagine that this should be named (since it's already inside `lsp-mode.el`) like `lsp-call-procecss` and be documented as a wrapper of `call-process` which returns output instead of inserting it in a buffer.
Same thing like how `lsp-async-start-process` exists.

With helper function like this, more generic name covers more use cases and keeps us from re-inventing the wheel.

yyoncho: FTR there is `shell-command-to-string`

jcs090218: > With helper function like this, more generic name covers more use cases and keeps us from re-inventing the wheel.

I think @kiennq have a good point here. Sometime I implemented a helper function that may work across all modules yet is hard for me to decide where to put it; especially, the function itself is for in very generic use. I wonder if there is a file that maybe we can place all our helper functions? (Sorry, I am quite new here. :P)

wyuenho: > I would imagine that this should be named (since it's already inside lsp-mode.el) like lsp-call-procecss and be documented as a wrapper of call-process which returns output instead of inserting it in a buffer.
Same thing like how lsp-async-start-process exists.

Scope creep.

> FTR there is shell-command-to-string

You can't get the exit status with `shell-command-to-string`. Command errors still returns output from that function.

> if there is a file that maybe we can place all our helper functions?

I was wondering about the same thing. Every large project should have a utils file/module somewhere, but it's spread out all over the place in our case.

yyoncho (approved):

jcs090218 (approved):

yyoncho:
Thank you!

@kiennq  I hope you are fine with the commit as it is.

PR #2164, opened 2020-09-10
Use number equal.

jcs090218:
In my understanding, `i` and `lsp/diagnostic-severity-` related variables are integer comparison? Maybe use `=` instead of `equal` for better understanding.

### DIFF
--- a/lsp-modeline.el
+++ b/lsp-modeline.el
@@ -213,10 +213,10 @@ The `:global' workspace is global one.")
                          (format "%s" (aref stats i))
                          'face
                          (cond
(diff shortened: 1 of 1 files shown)


ericdallo (approved):

kiennq:
Since `equal` works everywhere, not like `=` which only works for integers, I tended to use `equal`.
If we're sure this will be integer only, using `=` is also fine too

jcs090218:
I feel like `=` will be the better choice because I don't need to look into the variable's type. :)

kiennq:
Yeah, the down side is that it's checked in runtime, so if something happens, we cannot catch it in byte-compiled phase

PR #2171, opened 2020-09-13
ci: fix python-language-server installation failure

kiennq:
As in https://github.com/palantir/python-language-server/issues/863, it seems that python 3 will be required for `python-language-server`.


kiennq:
This will help unblock merging of other PR that cannot be done due to CI failures

yyoncho (approved):

PR #2177, opened 2020-09-14
watch all targets by default with rust-analyzer

brotzeit:
I think people expect this as a default configuration (https://github.com/brotzeit/rustic/issues/135)

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -378,7 +378,7 @@ The command should include `--message=format=json` or similar option."
   :group 'lsp-rust
   :package-version '(lsp-mode . "6.2.2"))
 
-(defcustom lsp-rust-analyzer-cargo-all-targets nil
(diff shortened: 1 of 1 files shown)


yyoncho:
@flodiebold

flodiebold:
Isn't that what `lsp-rust-analyzer-cargo-all-targets` is for? This would be turned on by default on the server, so it makes sense to change the default to `t`, I think.

brotzeit:
> Isn't that what lsp-rust-analyzer-cargo-all-targets is for? This would be turned on by default on the server, so it makes sense to change the default to t, I think.

I forgot about this one. Updated.

flodiebold (approved):
LGTM

brotzeit:
Need a member's approval.

kiennq (approved):

jcs090218 (approved):

PR #2180, opened 2020-09-15
`lsp--symbol-filter': ensure that kind is bounded

leungbk:
Servers like ccls use 'kind' values outside the normal range when
describing symbols that lie outside the protocol. We set those values
to zero and allow nil to represent the corresponding symbols in
`lsp-imenu-index-symbol-kinds'.

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6186,7 +6186,7 @@ an alist
       (cons name
             (lsp--imenu-create-hierarchical-index filtered-children)))))
 
-(lsp-defun lsp--symbol-filter ((&SymbolInformation :kind :location))
+(lsp-defun lsp--symbol-ignore ((&SymbolInformation :kind :location))
   "Determine if SYM is for the current document and is to be shown."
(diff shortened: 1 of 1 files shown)


ericdallo (approved):

lsp-mode.el
```diff
@@ -744,7 +744,8 @@ are determined by the index of the element."
                          (const :tag "Struct" Struct)
                          (const :tag "Event" Event)
                          (const :tag "Operator" Operator)
-                         (const :tag "Type Parameter" TypeParameter)))
+                         (const :tag "Type Parameter" TypeParameter)
+                         (const :tag "Other" nil)))
```
leungbk, line 748: I was considering changing the 0th entry of `lsp/symbol-kind-lookup` to `Other` or something similar and updating this value to match.

kiennq, line 748: Perhaps you should move this `Other` to the first row, so it will be on the same order of `lsp/symbol-kind-lookup`

kiennq (approved):

PR #2183, opened 2020-09-15
lsp-send-execute-command: better error message when fail to execute

kiennq:
As the title, we need better error message  in case server do not provide `workspace/executeCommand` or if there's error message when executing `lsp--send-execute-command`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5767,7 +5767,12 @@ REFERENCES? t when METHOD returns references."
   (let ((params (if args
                     (list :command command :arguments args)
                   (list :command command))))
-    (lsp-request "workspace/executeCommand" params)))
+    (condition-case-unless-debug err
+        (lsp-request "workspace/executeCommand" params)
+      (error
+       (lsp--error "Please open an issue in lsp-mode for implementing `%s'.\n\n%S"
+                   command
+                   err)))))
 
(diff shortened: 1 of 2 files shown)


yyoncho (approved):

lsp-mode.el
```diff
@@ -5767,7 +5767,12 @@ REFERENCES? t when METHOD returns references."
   (let ((params (if args
                     (list :command command :arguments args)
                   (list :command command))))
-    (lsp-request "workspace/executeCommand" params)))
+    (condition-case-unless-debug err
+        (lsp-request "workspace/executeCommand" params)
+      (error
+       (lsp--error "Please open an issue in lsp-mode for implementing `%s'.\n\n%S"
```
ericdallo, line 5773: What if server return an error on `workspace/executeCommand`?

kiennq: We will throw other error message for that. But yes, it should be handle differently.

kiennq: On other thought, when server fail to handle `workspace/executeCommand`, we should still ask user open a bug for client handle that command.
Server may choose to only handle some command with `workspace/executeCommand`

ericdallo: I see, thanks

ericdallo (approved):

PR #2187, opened 2020-09-16
[modeline] - Use a different face for code action when it is preferred

ericdallo:
Preferred code action:
image

Normal code actions:
image


yyoncho (approved):
Looks good. We should add function to apply prefered code action.

Issue #2188, opened 2020-09-16
Emacs + ls-mode fail on C# file in Linux Ubuntu 20.04


snowgoon88:
**Describe the bug**
Using emacs + lsp-mode with Linux (Ubuntu 20.04) on C# files, it seems I can connect to the omnisharp server but then nothing work and I have "LSP :: Error from the Language Server: Internal Error - System.NullReferenceException".
Seems related to issue #1125.

**To Reproduce**
Emacs with lsp-mode package.
Open .cs file
M-x lsp

**Expected behavior**
Code analysis and flycheck errors (as working for my c++ files)

**Which Language Server did you use**
omnisharp-roslyn as downloaded by emacs in
`~/.emacs.d/.cache/lsp/omnisharp-roslyn/v1.37.1`
(I tried to git clone the omnisharp-roselyn and built but failed because of gitVersion)

**OS**
Linux Ubuntu 20.04
`=> mono --version
Mono JIT compiler version 6.8.0.105 (Debian 6.8.0.105+dfsg-2 Wed Feb 26 23:23:50 UTC 2020)
Copyright (C) 2002-2014 Novell, Inc, Xamarin Inc and Contributors. www.mono-project.com
	TLS:           __thread
	SIGSEGV:       altstack
	Notifications: epoll
	Architecture:  amd64
	Disabled:      none
	Misc:          softdebug 
	Interpreter:   yes
	LLVM:          supported, not enabled.
	Suspend:       hybrid
	GC:            sgen (concurrent by default)`

**Error callstack**
lsp-workspace-log in https://gist.github.com/snowgoon88/6e0f1a0cf3eadff11c64076c8dd4fa31#file-lsp-workspace-log
messages.log in https://gist.github.com/snowgoon88/6e0f1a0cf3eadff11c64076c8dd4fa31#file-messages-log

lsp-session
```
[-] /home/alain/Projets/Unity/ImPacman
 -[-] csharp:429402
    |-[-] Buffers
    |  -[+] MoveFromAbove.cs
    -[+] Capabilities
[X] /home/alain/Projets/ChuchuGL
```

The `Assembly-CSharp.csproj` and `ImPacman.sln` (generated with Unity) are in `/home/alain/Projets/Unity/ImPacman`

It might be a difficulty to read/analyse the .csproj or .sln. But I'm not expert enough in C# or emacs to tell. I can share these files if needed.

yyoncho:
[...]

PR #2196, opened 2020-09-20
lsp-csharp: handle interop notifications

razzmatazz:
This adds handlers for '#o/XXX' notifications that have been added recently to facilitate omnisharp clients transitioning from custom to LSP protocol (see https://github.com/OmniSharp/omnisharp-roslyn/pull/1911). 

The events are ignored by default, except for 'o#/error' which may help resolving issues when projects are not being properly loaded (this may have helped to diagnose #2188)

### DIFF
--- a/clients/lsp-csharp.el
+++ b/clients/lsp-csharp.el
@@ -247,7 +247,7 @@ using the `textDocument/references' request."
     (message "No references found")))
 
 (lsp-defun lsp-csharp--handle-os-error (_workspace (&omnisharp:ErrorMessage :file-name :text))
-  "Handle the 'o#/error' (interop) notification with PARAMS by displaying a message with lsp-warn."
+  "Handle the 'o#/error' (interop) notification by displaying a message with lsp-warn."
   (lsp-warn "%s: %s" file-name text))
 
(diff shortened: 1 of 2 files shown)


clients/lsp-csharp.el
```diff
@@ -246,6 +246,11 @@ using the `textDocument/references' request."
       (lsp-show-xrefs (lsp--locations-to-xref-items locations-found) nil t)
     (message "No references found")))
 
+(defun lsp-csharp--handle-os-error (_workspace params)
+  "Handle the 'o#/error' (interop) notification with PARAMS by displaying a message with lsp-warn."
+  (-let (((&hash "FileName" filename "Text" text) params))
```
yyoncho, line 251: we should not use &hash because soon we will switch to plist for internal representation of the jsonrpc objects.

You have 2 options:

1. Use lsp-get
2. Define data structures and then use the getters/destructorings generated for the interface.

razzmatazz: done (with opt 2)

ericdallo (approved):
I faced this issue today :sweat_smile: 
Thanks for your contribution!

PR #2197, opened 2020-09-20
Automatic download related work

yyoncho:
Simple and straightforward implementation.

- implemented :download provider
- implemented download for rust-analyzer and for xmlls

----

#

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -555,7 +555,7 @@ The command should include `--message=format=json` or similar option."
     (lsp--apply-text-edits result)))
 
 (defcustom lsp-rust-analyzer-download-url
-  (format "https://github.com/rust-analyzer/rust-analyzer/releases/download/nightly/%s"
+  (format "https://github.com/rust-analyzer/rust-analyzer/releases/latest/download/%s"
           (pcase system-type
             ('gnu/linux "rust-analyzer-linux")
             ('darwin "rust-analyzer-macos")
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -316,12 +316,17 @@ PARAMS progress report notification data."
 
 
 ;; rust-analyzer
-
 (defcustom lsp-rust-analyzer-server-command '("rust-analyzer")
   "Command to start rust-analyzer."
   :type '(repeat string)
   :group 'lsp-rust
   :package-version '(lsp-mode . "6.2"))
+
+(defcustom lsp-rust-analyzer-server-args '()
+  "Args to start rust-analyzer with when rust-analyzer is downloaded automatically."
+  :type '(repeat string)
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
 
 (defcustom lsp-rust-analyzer-server-display-inlay-hints nil
   "Show inlay hints."
@@ -549,9 +554,43 @@ The command should include `--message=format=json` or similar option."
          (result (lsp-send-request (lsp-make-request "experimental/joinLines" params))))
     (lsp--apply-text-edits result)))
 
+(defcustom lsp-rust-analyzer-download-url
+  (format "https://github.com/rust-analyzer/rust-analyzer/releases/download/nightly/%s"
+          (pcase system-type
+            ('gnu/linux "rust-analyzer-linux")
+            ('darwin "rust-analyzer-macos")
+            ('windows-nt "rust-analyzer-windows.exe")))
+  "Automatic download url for Rust Analyzer"
+  :type 'string
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-rust-analyzer-store-path (f-join lsp-server-install-dir
+                                                "rust"
+                                                (if (eq system-type 'windows-nt)
+                                                    "rust-analyzer.exe"
+                                                  "rust-analyzer"))
+  "The path to the file in which `rust-analyzer' will be stored."
+  :type 'file
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(lsp-dependency
+ 'rust-analyzer
+ '(:system "rust-analyzer")
+ `(:download :url lsp-rust-analyzer-download-url
+             :store-path lsp-rust-analyzer-store-path
+             :set-executable? t))
+
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection (lambda () lsp-rust-analyzer-server-command))
+  :new-connection (lsp-stdio-connection
+                   (lambda ()
+                     `(,(or (executable-find
+                             (cl-first lsp-rust-analyzer-server-command))
+                            (lsp-package-path 'rust-analyzer)
+                            "rust-analyzer")
+                       ,@(cl-rest lsp-rust-analyzer-server-args))))
   :major-modes '(rust-mode rustic-mode)
   :priority (if (eq lsp-rust-server 'rust-analyzer) 1 -1)
   :initialization-options 'lsp-rust-analyzer--make-init-options
@@ -563,7 +602,9 @@ The command should include `--message=format=json` or similar option."
                      (lsp-rust-analyzer-inlay-hints-mode)))
   :ignore-messages nil
   :server-id 'rust-analyzer
-  :custom-capabilities `((experimental . ((snippetTextEdit . ,lsp-enable-snippet ))))))
+  :custom-capabilities `((experimental . ((snippetTextEdit . ,lsp-enable-snippet ))))
+  :download-server-fn (lambda (_client callback error-callback _update?)
+                        (lsp-package-ensure 'rust-analyzer callback error-callback))))
 
 (defun lsp-rust-switch-server (&optional lsp-server)
   "Switch priorities of lsp servers, unless LSP-SERVER is already active."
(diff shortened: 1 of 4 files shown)


ericdallo (approved):
Nice!

clients/lsp-rust.el
```diff
@@ -549,9 +554,43 @@ The command should include `--message=format=json` or similar option."
          (result (lsp-send-request (lsp-make-request "experimental/joinLines" params))))
     (lsp--apply-text-edits result)))
 
+(defcustom lsp-rust-analyzer-download-url
+  (format "https://github.com/rust-analyzer/rust-analyzer/releases/download/nightly/%s"
```
kiennq, line 558: Shouldn't this be `latest` instead of `nightly`? At least for the default.

kiennq:
```suggestion
  (format "https://github.com/rust-analyzer/rust-analyzer/releases/latest/download/%s"
```

clients/lsp-rust.el
```diff
@@ -549,9 +554,43 @@ The command should include `--message=format=json` or similar option."
...
+            ('windows-nt "rust-analyzer-windows.exe")))
+  "Automatic download url for Rust Analyzer"
+  :type 'string
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-rust-analyzer-store-path (f-join lsp-server-install-dir
+                                                "rust"
+                                                (if (eq system-type 'windows-nt)
+                                                    "rust-analyzer.exe"
```
kiennq, line 571: No need to add `.exe`, Windows can automatically detect that.

lsp-mode.el
```diff
@@ -6986,6 +6998,38 @@ nil."
...
+         (error (funcall error-callback err)))))))
+
+(cl-defun lsp-download-path (&key store-path set-executable? &allow-other-keys)
+  "Download URL and store it into STORE-PATH.
+
+SET-EXECUTABLE? when non-nil change the executable flags of
+STORE-PATH to make it executable."
+  (let ((store-path (lsp-resolve-value store-path)))
+    (cond
+     ((executable-find store-path) store-path)
```
kiennq, line 7025: In Windows, given a `store-path` without extension for `executable-find`, it can be extended to full path with extension (`.exe`, `.cmd`, ...).
I think you should put this one out of `cond` as an optional resolvers  like `(let ((store-path (or (executable-find store-path) store-path))))` and do the `cond` with the rest conditional switch.

clients/lsp-rust.el
yyoncho: @flodiebold any opinion on that? Is nighly stable enough for daily use or better to use the latest stable?

lsp-mode.el
yyoncho: I am not sure I understand the comment 100%. Are you suggesting to perform executable-find two times in this function?

clients/lsp-rust.el
flodiebold: I agree the weekly release is a better default.

yyoncho: Thank you, applied the comment.

lsp-mode.el
kiennq: No, you will only need to perform executable-find once.

``` elisp
(let* ((store-path (lsp-resolve-value store-path))
       (store-path (or (executable-find store-path) store-path)))
  (cond
   ((and set-executable? (f-exists? store-path))
    (set-file-modes store-path #o0700)
    store-path)
   ((f-exists? store-path) store-path)))
```

yyoncho: Structuring the code this way will result it calling set-file-modes everytime.

kiennq: Does setting file mode every time cause problem?
If that's not what you want, I think we can always temporarily cache the value from `executable-path` to prevent it form calling twice.

In Windows, if you have file `test`, `test.cmd` in the same folder and, the `executable-find` will return `test.cmd` and this function will return `test`. Which is not executable by Windows.

yyoncho: It is syscall so it is slow and it will be better if we can avoid it. I am looking at your suggestion and in the code in the PR and I fail to see a case in which both will act differently. Can you give an example?

kiennq: For example, if you have `a.cmd` under `~/tmp/`.
In Windows, with your original function:
``` elisp
(lsp-download-path :store-path "~/tmp/a")
"~/tmp/a"
```
and with the modified function
``` elisp
(lsp-download-path :store-path "~/tmp/a")
"c:/Users/xyz/tmp/a.cmd"
```

Since `~/tmp/a` is not existed, shouldn't the function return `nil` or correct path with suffix to executable file?

My previous suggestion is under assumption that you want to return the correct path with suffix.
But if that's not correct and you actually want to return `nil` here, maybe you should use the combination of `file-executable-p` and `file-directory-p` instead of `executable-find`, which will work on exact `store-path`.

yyoncho: > My previous suggestion is under assumption that you want to return the correct path with suffix.

There is no such need - this will be handled later by the code that will run the server.

kiennq: > There is no such need - this will be handled later by the code that will run the server.

Do you mean the code in `:new-connection` will do that?
Another extreme example, you're suppose to download `a.exe` to `bin/` folder, and there's already a `a.exe.cmd` file under that folder.
Now if your download failed somehow, the `lsp-download-path` will still return `bin/a.exe` (since `a.exe.cmd` is executable) and the server will fail to start. Isn't that a bug?

yyoncho: > Do you mean the code in :new-connection will do that?

The code in this PR does the same thing that we do in  `lsp-server-present?`.

> Now if your download failed somehow, the `lsp-download-path` will still return `bin/a.exe` (since `a.exe.cmd` is executable) and the server will fail to start. Isn't that a bug?

That same thing will happen if you set the server to `C:\server.exe` and it does not exist but there is `C:\server.exe.cmd`. I don't think that we should spend effort on handling cases like that.

kiennq: > The code in this PR does the same thing that we do in `lsp-server-present?`.

Is that the same? `lsp-server-present?` is returning `(executable-find path)` result instead `path` though.

> That same thing will happen if you set the server to `C:\server.exe` and it does not exist but there is `C:\server.exe.cmd`. I don't think that we should spend effort on handling cases like that.

But we normally just set the server to `server` so it works in all platform, don't we?

I think the check is kind of a bug and user may be affected by it. It can be easily resolved with one line change, why shouldn't we fix it?

yyoncho: > Is that the same? `lsp-server-present?` is returning `(executable-find path)` result instead `path` though.

You are missing the point. `lsp-server-present?` returns `nil`/`non-nil` as an indication that the server is present. We do the same in this function. 

> I think the check is kind of a bug and user may be affected by it. It can be easily resolved with one line change, why shouldn't we fix it?

The chances that anyone will create `rust-analyzer.exe.cmd` and we will mistakenly start the server. I don't understand what is the bug that your suggestion will fix.

kiennq: > You are missing the point. `lsp-server-present?` returns `nil`/`non-nil` as an indication that the server is present. We do the same in this function.

Except that in this function we even return `non-nil` for non existed server?
Since currently `lsp-server-present?` takes output from command lambda, which is using `lsp-download-path`, if `lsp-download-path` is returning `nil` for non existed server, that would make it fallback to final condition clause and allow `lsp-server-present?` takes correct parameters.



> The chances that anyone will create `rust-analyzer.exe.cmd` and we will mistakenly start the server. I don't understand what is the bug that your suggestion will fix.

With my suggestion in https://github.com/emacs-lsp/lsp-mode/pull/2197#discussion_r493687753, we will not mistakenly start the server even when `rust-analyzer.exe` is not existed.


The discussion has gotten long. If you think my point is invalid then it's fine, I'm ok with that.

yyoncho: Sorry I got lost. If the whole discussion is to guard us against user creating `~/.emacs.d/.cache/lsp-rust/rust-analyzer.exe.cmd` which then we will mistakenly assume as the rust-analyzer then the discussion shouldn't have started at all because this is not something that ever going to happen and even if it happens it will be entirely users fault.

kiennq (approved):

lsp-mode.el
kiennq: > it will be entirely users fault

Even if it's user fault, it will be a special case since user's `lsp-mode` is not working.
Further more, the error message in `*lsp-log*` doesn't mention anything like that, so it will be hard for user to know why their configuration is not working.

My main concern in this function is we check for one thing (`(executable-find path)`) but returns another thing `path`, which seems weird.

I've approved the PR anyway.

clients/lsp-rust.el
```diff
@@ -549,9 +554,43 @@ The command should include `--message=format=json` or similar option."
...
+  "The path to the file in which `rust-analyzer' will be stored."
+  :type 'file
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1"))
+
+(lsp-dependency
+ 'rust-analyzer
+ '(:system "rust-analyzer")
+ `(:download :url lsp-rust-analyzer-download-url
+             :store-path lsp-rust-analyzer-store-path
```
ericdallo, line 582: @yyoncho, is that right? Isn't that missing commas?
```elisp
`(:download :url ,lsp-rust-analyzer-download-url
             :store-path ,lsp-rust-analyzer-store-path
```

PR #2199, opened 2020-09-21
fix a bug that prevents client packages to be loaded

zhengyangfeng00:
This fixes a bug introduced in #2152. I think the intention for the mentioned PR is to load client packages only for the first time but the logic was inverted.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1029,7 +1029,7 @@ called with nil the signature info must be cleared."
 
 (defvar lsp--tcp-port 10000)
 
-(defvar lsp--client-packages-required nil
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
Look right for me
c/c @yyoncho

ericdallo:
Actually I think both are right, is just the names/doc-strings are confused

kiennq:
I think just the doc-string is incorrected.
The name is actually fine `lsp--client-packages-required` means packages is already `require`-d doesn't it?

zhengyangfeng00:
ah got it, I mis-read the variable name. I'll close the PR

PR #2200, opened 2020-09-21
Add lsp-nix to the list of lsp-client-packages

hongchangwu:
#2115 added Nix support, but it is not on the list of clients that are automatically required. This PR fixes it. Also made cosmetic changes to make `lsp-nix.el` more consistent with other clients.


ericdallo (approved):
Thanks for the fix!

PR #2201, opened 2020-09-22
[lsp-json] Change executable path for system-installed server

leungbk:
When installing the server via Nix, this seems to be the correct path. I'm not sure if other package managers have this as their path, though.

### DIFF
--- a/clients/lsp-json.el
+++ b/clients/lsp-json.el
@@ -99,7 +99,7 @@
                   (list callback))))
 
 (lsp-dependency 'vscode-json-languageserver
-                '(:system "vscode-json-languageserver")
+                '(:system "json-languageserver")
                 '(:npm :package "vscode-json-languageserver"
                        :path "vscode-json-languageserver"))
 


ericdallo (approved):

clients/lsp-json.el
```diff
@@ -99,7 +99,7 @@
                   (list callback))))
 
 (lsp-dependency 'vscode-json-languageserver
-                '(:system "vscode-json-languageserver")
+                '(:system "json-languageserver")
```
kiennq, line 102: Is this the correct one?
When I installed this via `npm` as global, the binary name is `vscode-json-languageserver`.
Did Nix change the binary name?

kiennq: The old one seems to be named `json-languageserver`, but that one is obsoleted.
https://www.npmjs.com/package/vscode-json-languageserver

ericdallo: I confirmed Nix use the `json-languageserver` name, and I didn't find anything changing the name: https://search.nixos.org/packages?show=nodePackages.vscode-json-languageserver-bin&query=json-language&from=0&size=30&sort=relevance&channel=unstable#disabled

ericdallo: Hum, probably have a outdated json-laguageserver, is that right @leungbk ?

leungbk: As it turns out, Nix's version is quite old. I'll have to revert that commit.

PR #2205, opened 2020-09-23
lsp-completion: cleanup lsp-completion--last-result at each completion session

kiennq:
Properly cleanup `lsp-completion--last-result` at each completion session.
I kind of forgot this

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -37,6 +37,15 @@
   :group 'lsp-mode
   :package-version '(lsp-mode . "7.0.1"))
 
+;;;###autoload
+(define-obsolete-variable-alias 'lsp-enable-completion-at-point
+  'lsp-completion-enable "lsp-mode 7.0.1")
+
+(defcustom lsp-completion-enable t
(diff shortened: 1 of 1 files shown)


lsp-completion.el
```diff
@@ -671,53 +681,54 @@ The CLEANUP-FN will be called to cleanup."
...
+        (add-to-list 'company-backends 'company-capf))
+       (t
+        (lsp--warn "Unable to autoconfigure company-mode.")))
+
+      (when (bound-and-true-p company-mode)
+        (add-hook 'company-completion-started-hook
+                  completion-started-fn
+                  nil
+                  t)
+        (add-hook 'company-after-completion-hook
```
yyoncho, line 715: We had issues with these hooks in the past, right? Dont we have to us the other mechanism that crosschecks if the completion cache is relevant to the current position?

kiennq: Yes, I've updated that part too. Thanks

yyoncho (approved):
Looks good

PR #2206, opened 2020-09-23
Add default lsp-file-watch-ignored entries for Haskell/cabal projects

michaelpj:
We have `.stack-work` for the Stack build tool. Cabal is the other main
Haskell build tool. It has two typical build directories: `dist` and
`dist-newstyle` (for more recent versions).


leungbk (approved):

leungbk:
Thanks!

PR #2208, opened 2020-09-23
`lsp-file-watch-ignored': use proper regexes

nbfalcon:
Emacs regexes specify the end of string using \', not $ as was used in that
variable (the latter matching *either* the end of the line or the string, which
could cause issues with files that have newlines in them (Linux allows this)).
Fix that by using \' for the end of string character instead.

Also properly handle node_modules. It is not necessarily always at the root of
the project, as is the case with .pyi files needed by pyright, which are in a
node_modules folder somewhere in the user's home directory. This causes
annoyances after Emacs-restarts, with `lsp' prompting the user to watch 11+k
immutable files. Fix that by removing the `eos' matcher and matching a slash
after "node_modules".


kiennq:
> Also properly handle node_modules. It is not necessarily always at the root of
> the project, as is the case with .pyi files needed by pyright, which are in a
> node_modules folder somewhere in the user's home directory. This causes
> annoyances after Emacs-restarts, with `lsp' prompting the user to watch 11+k immutable files. Fix that by removing the `eos' matcher and matching a slash
> after "node_modules".

Why adding a matching slash in `node_modules` solve this problem? I think in `lsp--directory-files-recursively`, the regex list is a only used to match directory (minus final slash) only

nbfalcon:
@kiennq it doesn't, it just restricts it further; what does solve the problem though is removing the final $. The final slash only forces node_modules to be a directory somewhere, so a node_modules file at the project root wouldn't match. Sorry, I was a bit unclear in this commit message.

kiennq:
> it doesn't, it just restricts it further; what does solve the problem though is removing the final $. The final slash only forces node_modules to be a directory somewhere, so a node_modules file at the project root wouldn't match

Does that mean we will not ignore `node_modules` at project root?
I think in lsp--directory-files-recursively, the regex list is a only used to match directory (minus final slash) only, so shouldn't regex with last slash will never be matched?

nbfalcon:
@kieenq We still ignore node_modules at the project root, unless it is an ordinary file. Now it will be ignored anywhere in the file path.

nbfalcon:
Updated message.

nbfalcon:
Is there something blocking this?

lsp-mode.el
```diff
@@ -7563,13 +7563,13 @@ When ARG is t the lsp mode will start new language server even if
...
 ARG is nil current file will be opened in multi folder language
 server if there is such. When `lsp' is called with prefix
-argument ask the user to select which language server to start. "
+argument ask the user to select which language server to start."
   (interactive "P")
 
   (when (and lsp-auto-configure (not lsp--client-packages-required))
     (seq-do (lambda (package)
               ;; loading client is slow and `lsp' can be called repeatedly
-              (unless (featurep package) (require package nil t)))
```
kiennq, line 7572: Please refer to https://github.com/emacs-lsp/lsp-mode/pull/1498, this change is undid that.
Is the original commit is not necessary anymore?

kiennq: > It is only slow the first time it is run, because require isn't fast.
> 
> Here is how require works in Emacs:
> There is a set of symbols, called the "features". To check whether a feature is provided (a feature just being a symbol), the `feature` function, implemented in C, checks whether that symbol is in that. A feature can be added to that set using the `provide` function, which is why all modules should end in (provide 'name).
> 
> Require first checks if some feature is provided, and, if not, tries to find the module that it needs to load. This is slow, since it has to check lots of files (the entire `load-path`, 327 directories on my system), but only has to be done once. So (require 'foo) might be somewhat slow the first time, but 'foo will then be provided, so subsequent requires are fast. The issue with lsp-mode was (is?) that some clients which aren't installed are required. This always fails, since those clients aren't in the `load-path`, meaning that those clients never get their features provided, so the `load-path` has to be traversed all the time, which is slow.
> 
> I hope that clears things up.

The slow that #1498 fixed is not from `require` have to look up load path but because of `lsp-register-custom-settings` is called multiple times.
Your change undid that so I would like to ask for more clarification.

kiennq:
> Is there something blocking this?

Sorry I'm still not fully understand this, mainly due to what's different from other ignore regex compare to `node_modules`?

Is the reason to remove final `$` from `node_modules` due to the current `lsp-mode` only scan and match the regex against the top level folders? And the subsequences recursive scanned folder will not be matched against the regexes, that's why the final `$` needed to be removed from `node_modules`.

The change from `$` -> `\\'` looks fine though.

Since I'm not fully understand this, I would left it for other maintainers to take decision here.

lsp-mode.el
kiennq: Okay, `require` seems only load package once, https://www.gnu.org/software/emacs/manual/html_node/elisp/Named-Features.html.
The change in #1498 probably needed to be reworked

nbfalcon:
@kiennq you correctly understood :) pyright stores .pyi in its installation directory, which, if it is globally installed, is a node_modules folder somewhere outside the project and not directly below node_modules. In my case, this would be the "/home/nikita/.local/share/npm/lib/node_modules/pyright/dist/typeshed-fallback/stdlib/", folder, which is *not* at the top level, causing it to watch that immutable, huge directory. I'm guessing that the project it found was "/home/nikita/.local/share/npm/lib/node_modules/pyright/", which has no node_modules subfolder.

I hope this might be somewhat clearer.

lsp-mode.el
nbfalcon: Actually, since we already guard that, that won't make sense; the features shouldn't be loaded already, meaning optimizing for that case somewhat odd. Should I still revert that?

kiennq (approved):

Issue #2195, opened 2020-09-19
Bad gopls completion reordering


muirdm:
lsp-mode seems to be rearranging the completion candidate order from gopls. In below screenshot, `foo.str` should be the first candidate. It is of type `string`, so it matches the assignment well.
![Screen Shot 2020-09-19 at 10 24 41 AM](https://user-images.githubusercontent.com/11356572/93680171-7f6fe500-fa62-11ea-88c5-afaed68cbb1c.png)

<details>
<summary>
Candidates
</summary>

    [Trace - 10:24:27 AM] Received response 'textDocument/completion - (10408)' in 6ms.
    Result: {
      "items": [
        {
          "textEdit": {
            "newText": "foo.str",
            "range": {
              "end": {
                "character": 17,
                "line": 4
              },
              "start": {
                "character": 16,
                "line": 4
              }
            }
          },
          "insertTextFormat": 2,
          "filterText": "foo.str",
          "sortText": "00000",
          "preselect": true,
          "detail": "string",
          "kind": 5,
          "label": "foo.str"
        },
        {
          "additionalTextEdits": [
            {
              "newText": "\nimport \"path/filepath\"\n",
              "range": {
                "end": {
                  "character": 0,
                  "line": 1
                },
                "start": {
                  "character": 0,
                  "line": 1
                }
              }
            }
          ],
          "textEdit": {
            "newText": "filepath.Base(${1:path string})",
            "range": {
              "end": {
                "character": 17,
                "line": 4
              },
              "start": {
                "character": 16,
                "line": 4
              }
            }
          },
          "insertTextFormat": 2,
          "filterText": "filepath.Base",
          "sortText": "00001",
          "detail": "func(path string) string (from \"path/filepath\")",
          "kind": 3,
          "label": "filepath.Base"
        },
        {
          "additionalTextEdits": [
            {
              "newText": "\nimport \"path/filepath\"\n",
              "range": {
                "end": {
                  "character": 0,
                  "line": 1
                },
                "start": {
                  "character": 0,
                  "line": 1
                }
              }
            }
          ],
          "textEdit": {
            "newText": "filepath.Clean(${1:path string})",
            "range": {
              "end": {
                "character": 17,
                "line": 4
              },
              "start": {
                "character": 16,
                "line": 4
              }
            }
          },
          "insertTextFormat": 2,
          "filterText": "filepath.Clean",
          "sortText": "00002",
[...]

PR #2210, opened 2020-09-25
fuz-score: rework fuz-score algo to boost prefix match and not consider length

kiennq:
Slightly modify the current heuristic of calculating fuzzy score.
Before: `fuz_score(str) = len(query) / ((sum(hole_score(i)) + 1) * len(str))`
After: `fuz_score(str) = len(query) / (sum(hole_score(i)) + 1)`

With the `hole_score` of the first hole is boosted to prefer the prefix match.
And also we do not consider string length anymore, thus no longer boost the score for shorter matched string.
As proposed in https://github.com/emacs-lsp/lsp-mode/issues/2195#issuecomment-698879445

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -572,10 +572,9 @@ Others: TRIGGER-CHARS CANDIDATES"
                                  (match-data))))))
                (start (pop md))
                (len (length str))
-               ;; To understand how this works, consider these bad
-               ;; ascii(tm) diagrams showing how the pattern "foo"
-               ;; flex-matches "fabrobazo", "fbarbazoo" and
-               ;; "barfoobaz":
+               ;; To understand how this works, consider these bad ascii(tm)
+               ;; diagrams showing how the pattern "foo" flex-matches
+               ;; "fabrobazo", "fbarbazoo" and "barfoobaz":
 
                ;;      f abr o baz o
                ;;      + --- + --- +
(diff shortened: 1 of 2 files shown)


yyoncho:
Looks good to me. Can you add a bit more tests in order to know where we are? I am interested in caps lock matches as well, e. g. prefix `F` candidates: `Foo`, `foo`

kiennq:
Do you have some test cases in mind?

yyoncho:
Here are(the first entry is the prefix)

F foo Foo
Fo daFo safo
F daFo safo
foo foo afoo aafoo aaafoo

kiennq:
Added

yyoncho (approved):
Thank you.

yyoncho:
After this can we remove the gopls hack?

kiennq:
I'm afraid no.

> The issue here is because in the former one , the `filterText` is the same for every completion items, and that make our re-sorting for fuzzy useless, since all the items have the same score. The sort is stable so the correct order is showing as in [#2195 (comment)](https://github.com/emacs-lsp/lsp-mode/issues/2195#issuecomment-695783305).
> 
> The latest version (0.5.0) however put different `filterText` for each item, and that make we reordering the candidates.
> And here is the commit which's doing that [golang/tools@b2d8b03#diff-4f76f1e205dd09c758df98edb416d93b](https://github.com/golang/tools/commit/b2d8b03366329f6cb0ff15138a7732def14679a7#diff-4f76f1e205dd09c758df98edb416d93b).
> 
> We can follow their step and put the same logic inside `lsp-go` or we can doing simpler with `(setq-local lsp-completion-filter-on-incomplete nil)` inside `:after-open-fn` of `lsp-go` client registration, which will benefit all user of `lsp-go`

So now VsCode Go plugin will do the `filterText` normalization hack instead of `gopls`.
Thus, we either need to replicate the same hack on `lsp-go` or we can just use own option of not filtering on incomplete.
Note that `gopls` will surely return incomplete (unless it support `score` like `clangd`) so that client will request to server everytime.

yyoncho:
> So now VsCode Go plugin will do the filterText 

What is that?

muirdm:
lsp-mode should never change the order of gopls completion results. gopls orders the results based on type information and other things not available to lsp-mode.

kiennq:
> > So now VsCode Go plugin will do the filterText
> 
> What is that?


Here is the commit in VsCode Go plugin that doing the hack (and prevent VsCode from reordering) https://github.com/golang/vscode-go/commit/423d7f568485f1b0dea137a50bdac363c9540c49.

yyoncho:
> > > So now VsCode Go plugin will do the filterText
> > 
> > 
> > What is that?
> 
> Here is the commit in VsCode Go plugin that doing the hack (and prevent VsCode from reordering) [golang/vscode-go@423d7f5](https://github.com/golang/vscode-go/commit/423d7f568485f1b0dea137a50bdac363c9540c49).

Ok. The presence of that hack in VScode justifies the equivalent hack on our side.

> (item.kind === CompletionItemKind.Method || item.kind === CompletionItemKind.Function) 

This is interesting too.

kiennq:
> This is interesting too.

Yes, we probably should change the condition of automatically trigger signature after completion to check for completion item kind too.

PR #2216, opened 2020-09-29
Added custom configuration options for lua-language-server

folke:
As the title says, I've added all the configuration options for [lua-language-server](https://github.com/sumneko/lua-language-server) as described on [Settings without VSCode](https://github.com/sumneko/lua-language-server/wiki/Setting-without-VSCode)

This also fixes an issue with the server, since after initialisation, it requests the `workspace/configuration` from the client. It will only start scanning your workspace once it received the configuration. Requested sections at startup are **Lua**, **files.associations** and **files.exclude**. With the new default options, the server will work as expected.

An example of configuring some options (in doom emacs):
```elisp
(after! lsp-mode
  (setq lsp-lua-workspace-max-preload 300)
  (setq lsp-lua-diagnostics-globals ["hs" "spoon"]))
```

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -127,6 +127,214 @@
   (and (f-exists? lsp-clients-lua-language-server-main-location)
        (f-exists? lsp-clients-lua-language-server-bin)))
 
+(defcustom lsp-lua-color-mode "Semantic" "Color mode."
+  :type '(choice (:tag "Grammar" "Semantic"))
+  :package-version '(lsp-mode . "7.0")
+  :group 'lsp-lua-language-server)
+
+(defcustom lsp-lua-completion-call-snippet "Disable" "Shows function call snippets."
+  :type '(choice (:tag "Disable" "Both" "Replace"))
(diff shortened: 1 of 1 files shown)


clients/lsp-lua.el
```diff
@@ -127,6 +127,214 @@
   (and (f-exists? lsp-clients-lua-language-server-main-location)
        (f-exists? lsp-clients-lua-language-server-bin)))
 
+(defcustom lsp-lua-color-mode "Semantic" "Color mode."
+  :type '(choice (:tag "Grammar" "Semantic"))
+  :package-version '(lsp-mode . "7.0")
```
yyoncho, line 132: please change these to 7.1. Can you also place the doc string on a new line, here and in the whole review?

folke: I've made the requested changes.
Thank you for the quick reply!

yyoncho (approved):

yyoncho:

folke:
Awesome, thanks!

PR #2220, opened 2020-10-03
[breadcrumb] Set evil jumps when invoking lsp-breadcrumb-go-to-symbol

leungbk:
----

#

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -161,11 +161,15 @@ Switch to current mouse interacting window before doing BODY."
                                        directory-display-string)
                                directory-display-string))
 
+(declare-function evil-set-jump "ext:evil-jumps")
+
 (lsp-defun lsp-headerline--symbol-with-action ((symbol &as &DocumentSymbol :name) symbol-display-string)
   "Build action for SYMBOL and SYMBOL-STRING."
   (lsp-headerline--with-action (let ((map (make-sparse-keymap)))
                                  (define-key map [header-line mouse-1]
                                    (lsp-headerline--make-mouse-handler
+                                     (when (bound-and-true-p evil-mode)
+                                       (evil-set-jump))
                                      (lsp-headerline--go-to-symbol symbol)))
                                  (define-key map [header-line mouse-2]
                                    (lsp-headerline--make-mouse-handler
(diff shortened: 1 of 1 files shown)


lsp-headerline.el
```diff
@@ -335,6 +339,11 @@ PATH is the current folder to be checked."
         (lsp--info "Server does not support breadcrumb."))
     (lsp--info "Call this function with a number representing the symbol position on breadcrumb")))
 
+(declare-function evil-set-command-property "ext:evil-common")
+
+(with-eval-after-load 'evil
+  (evil-set-command-property 'lsp-breadcrumb-go-to-symbol :jump t))
```
ericdallo, line 345: Shouldn't we move this to the `lsp-headerline-breadcrumb-mode`?

When enabled set that and when disabled, remove the evil-command-property?

leungbk: I hadn't considered that. Is there any difference in behavior? I cannot think of any context where an `evil` user would want `:jump nil` for `#'lsp-breadcrumb-go-to-symbol`.

ericdallo: Just keeping the pattern of only activating things when modes are enabled and remove the changes when the mode is disabled.
This pattern helps debugging bugs and keeps a clean state IMO

leungbk: `evil-set-command-property` modifies `evil-command-properties`, an alist mapping functions to properties, but does not modify the `lsp-breadcrumb-go-to-symbol` function; so I don't see how moving the invocation to the mode definition would make any `lsp`-managed state any simpler, or make the `lsp-breadcrumb-go-to-symbol` function easier to debug. It also wouldn't influence how functions defined by other packages behave.

Moving the invocation also incurs more overhead since `evil-command-properties` has several hundred entries, and since `evil-set-command-property` will in general take linear time.

leungbk: Actually, there's another benefit of leaving the `evil-set-command-property` invocation outside of `lsp-headerline-breadcrumb-mode`: if it is left outside and I call `lsp-breadcrumb-go-to-symbol` when `lsp-headerline-breadcrumb-mode` is off, then the `jump` marker gets set anyway. Which is desirable behavior, since a user would want to have those markers in the jump list whether or not `lsp-headerline-breadcrumb-mode` is on.

ericdallo: OK, no problem, I see your point :)

ericdallo (approved):

lsp-headerline.el
```diff
@@ -161,11 +161,16 @@ Switch to current mouse interacting window before doing BODY."
...
                                directory-display-string))
 
+(declare-function evil-set-jump "ext:evil-jumps")
+
 (lsp-defun lsp-headerline--symbol-with-action ((symbol &as &DocumentSymbol :name) symbol-display-string)
   "Build action for SYMBOL and SYMBOL-STRING."
   (lsp-headerline--with-action (let ((map (make-sparse-keymap)))
                                  (define-key map [header-line mouse-1]
                                    (lsp-headerline--make-mouse-handler
+                                     (when (and (require 'evil nil t)
```
yyoncho, line 171: seems like the `require` part is not needed here.

leungbk: Fixed.

yyoncho (approved):

PR #2222, opened 2020-10-03
ignore variables affecting serialization in `lsp-persist`

Lompik:
the `prin1*` (used by `lsp--persist`) family of function are affected by `print-level`, `print-length` and others (https://www.gnu.org/software/emacs/manual/html_node/elisp/Output-Variables.html). 

Without this patch, setting `print-level` to less that the length of `lsp-session` will silently corrupt `lsp-session-file` when saved. 
(in my case, I did not manually set `print-level` but my `lsp-sessions-file` ended with bare "..." as if truncated and lsp warned `Failed to parse the session Hash table data` on next emacs startup)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3443,7 +3443,9 @@ in that particular folder."
 
 (defun lsp--persist (file-name to-persist)
   "Persist TO-PERSIST in FILE-NAME."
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:

PR #2227, opened 2020-10-07
lsp-headerline: make mouse-2 action for symbol to toggle narrowing

tsuu32:
Currently consecutive [header-line mouse-2] for symbol does nothing.
So I think toggling narrowing is better experience.

Please review my suggestion and code.

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -173,9 +173,17 @@ Switch to current mouse interacting window before doing BODY."
                                      (lsp-headerline--go-to-symbol symbol)))
                                  (define-key map [header-line mouse-2]
                                    (lsp-headerline--make-mouse-handler
-                                     (lsp-headerline--narrow-to-symbol symbol)))
+                                     (-let (((&DocumentSymbol :range (&RangeToPoint :start :end)) symbol))
+                                       (if (and (eq (point-min) start) (eq (point-max) end))
+                                           (widen)
+                                         (lsp-headerline--narrow-to-symbol symbol)))))
                                  map)
-                               (format "mouse-1: go to '%s' symbol\nmouse-2: narrow to '%s' range" name name)
+                               (format "mouse-1: go to '%s' symbol\nmouse-2: %s"
+                                       name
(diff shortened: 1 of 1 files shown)


lsp-headerline.el
```diff
@@ -173,9 +173,15 @@ Switch to current mouse interacting window before doing BODY."
                                      (lsp-headerline--go-to-symbol symbol)))
                                  (define-key map [header-line mouse-2]
                                    (lsp-headerline--make-mouse-handler
-                                     (lsp-headerline--narrow-to-symbol symbol)))
+                                     (if (buffer-narrowed-p)
```
yyoncho, line 176: Doesn't it make sense to widen only if you are trying to narrow to the exact same region? In case you want to narrow to another item(e. g. parent), it seems like it makes more sense to narrow to that region.

tsuu32: Oh, You are right. I've never seen multiple (parent and child) symbols in headerline. 
I try to fix my code.

tsuu32:
force-pushed.
I think 3d74358 works as you said but I don't know this is good condition and I can't test this.

tsuu32:
I tested this. When I narrow parent and then narrow child,  I feel weird that whole buffer widen by pushing mouse-2 on child...

lsp-headerline.el
```diff
@@ -173,9 +173,17 @@ Switch to current mouse interacting window before doing BODY."
                                      (lsp-headerline--go-to-symbol symbol)))
                                  (define-key map [header-line mouse-2]
                                    (lsp-headerline--make-mouse-handler
-                                     (lsp-headerline--narrow-to-symbol symbol)))
+                                     (-let (((&DocumentSymbol :range (&RangeToPoint :start :end)) symbol))
+                                       (if (= (- (point-max) (point-min)) (- end start))
```
yyoncho, line 177: Shouldn't you compare point-min with start and point-max with end? If you compare the difference then it will have wrong behaviour if 2 elements have the same size, no?

tsuu32: fixed

yyoncho (approved):

yyoncho:

Issue #2229, opened 2020-10-07
[completion] completion sorting issues with ts-ls


yyoncho:
1. `React` placed infront of `classname`

![image](https://user-images.githubusercontent.com/13259670/95370607-b27afb00-08e1-11eb-9ac7-e5563da81495.png)

2. When there is a mismatch in the case the entry is with very low flex score.

![image](https://user-images.githubusercontent.com/13259670/95370657-c6266180-08e1-11eb-9371-f32c12b67101.png)

yyoncho:
cc @kiennq

kiennq:
The first one looks odd, do you have log for that?

The second one is kind of expected, since the case mismatch will be ranked low, depends on `lsp-completion--fuz-case-sensitiveness`.
I can employed a smart case matching strategy: if there's upper case character, use sensitive matching, else ignore case sensitive. WDYT?

kiennq:
Also the second order looks odd too, `MediaStreamTrack` should have lower order than `onreadystatechange`.
Do you have `company-statistics-mode` on?

yyoncho:
> Also the second order looks odd too, `MediaStreamTrack` should have lower order than `onreadystatechange`.
> Do you have `company-statistics-mode` on?

No, this is tested with a vanilla config. I will post the project.

yyoncho:
> > Also the second order looks odd too, `MediaStreamTrack` should have lower order than `onreadystatechange`.
> > Do you have `company-statistics-mode` on?
> 
> No, this is tested with a vanilla config. I will post the project.

Hm, I am unable to reproduce that today. I suspect that I might have run some old version of lsp-mode.

PR #2234, opened 2020-10-08
lsp-completion: push case mismatch items to near next rank

kiennq:
Implement what's proposed in https://github.com/emacs-lsp/lsp-mode/issues/2229#issuecomment-705264022.
Case mismatch will be pushed to **near** next rank, i.e., lower score than the item has the same case match but higher than item has same case match but further from start.
Refer to test cases to see how's the ranking now.

### DIFF
--- a/test/lsp-completion-test.el
+++ b/test/lsp-completion-test.el
@@ -49,32 +49,35 @@
              '("hash-map"
                "as-definition"
                "as-def"
-               "aS-selection"
+               "As-selection"
                "To-as-expected"
                "amused"
                "subclass-1"
                "superand-sort")
              '("as-definition"    ; Prefix match
                "as-def"           ; Also prefix match (stable)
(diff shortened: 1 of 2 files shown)


test/lsp-completion-test.el
```diff
@@ -73,8 +73,8 @@
              '("F" "daFo" "safo")
              '("F" "daFo" "safo"))
     (do-test "foo"
```
yyoncho, line 75: can you add a single letter prefix test like:

> f foo Foo aFoo afoo

yyoncho: and maybe 
> F foo Foo aFoo afoo

kiennq: Done

yyoncho (approved):

yyoncho:
Thank you!

PR #2237, opened 2020-10-08
[WIP] add pyls tutorial

mpanarin:


docs/tutorials/PYLS-guide.md
```diff
@@ -0,0 +1,86 @@
...
+2. You need to make emacs know about virtualenv, and you need to do it **before** a call to `(lsp)`
+    * If you use Spacemacs \
+    You are in luck. Just create a file `.venv` with a path to your virtualenv in the root of your project.
+    * If you use Doom Emacs \
+    TBA (try the vanilla emacs approach, as well as check this [issue](https://github.com/hlissner/doom-emacs/issues/1666) in doom repo)
+    * If you use vanilla Emacs \
+    Check how spacemacs [does it](https://github.com/syl20bnr/spacemacs/blob/a9fc0d5fcc46aaef95aef99de3097fd79f0e3d26/layers/%2Blang/python/funcs.el#L225) with the [pyvenv](https://github.com/jorgenschaefer/pyvenv) emacs package \
+    It should look something like this (ignoring other options except `:hook`):
+    ```elisp
+    (use-package lsp-mode
```
yyoncho, line 68: Can we have some self-contained configuration like the one in cpp guide or this is not applicable for python? I wish there is a script that "just-works" and that can be used as a base.

mpanarin: I'll look into it more. But python is a bit too dynamic for that with different virtualenv tools emerging as fast as new js frameworks. 
If you are about the per-project approach being a bit too cumbersome to set up - it is. So we can specify the global installation as a recommended way, while leaving the per-project approach on the mercy of the user.
With global install we can try to make an educated guess for the virtualenv location, kinda like spacemacs does, and set the `pyls.jedi.environment` ourselves in `(lsp)` call somewhere during configuration of the client. 

The only issue I see with this is if we guess wrong, which we easily can, how do we allow the user to specify a proper virtualenv? Because, afaiu, dir-locals won't work, we'd just overwrite them...

yyoncho: Dir locals work afaik.

I want the tutorial to be in the form "if you follow each of the steps it will just work" with all of the steps. E. g. you may use the recommended tool for creating a project skeleton(or the one you like) then describe a flow. I wish venvs usecase to be covered because this question often pops up but no need to have a detailed description, just one way to make it work. E. g. like this - https://code.visualstudio.com/docs/nodejs/reactjs-tutorial even if you have zero JS experience you will be able to follow it.

mpanarin: Okay, I will only leave a global install as it is more straightforward

docs/tutorials/PYLS-guide.md
```diff
@@ -0,0 +1,180 @@
...
+(when (cl-find-if-not #'package-installed-p package-selected-packages)
+  (package-refresh-contents)
+  (mapc #'package-install package-selected-packages))
+
+;; sample `helm' configuration use https://github.com/emacs-helm/helm/ for details
+(helm-mode)
+(require 'helm-xref)
+(define-key global-map [remap find-file] #'helm-find-files)
+(define-key global-map [remap execute-extended-command] #'helm-M-x)
+(define-key global-map [remap switch-to-buffer] #'helm-mini)
```
petr-tik, line 62: Are these helm settings relevant for lsp-pyls?

yyoncho:
Cleanup - feel free to resubmit if you want to work on that.

Issue #2230, opened 2020-10-07
Typing `-` in json property removes all candidates


yyoncho:
![image](https://user-images.githubusercontent.com/13259670/95374745-66cb5000-08e7-11eb-9263-7db15b46df2e.png)

I have added the following elements:
```
         :/.eslintrc.json ["http://json.schemastore.org/eslintrc"]
        :/.eslintrc ["http://json.schemastore.org/eslintrc"]
```
in  `lsp-json--schema-associations`.
        

```
{
    "env": {
        "browser": true,
        "es6": true
    },
    "extends": [
        "eslint:recommended",
        "plugin:react/recommended"
    ],
    "globals": {
        "Atomics": "readonly",
        "SharedArrayBuffer": "readonly"
    },
    "parserOptions": {
        "ecmaFeatures": {
            "jsx": true
        },
        "ecmaVersion": 2018,
        "sourceType": "module"
    },
    "plugins": [
        "react"
    ],
    "rules": {
        "no-": "error"
    }
}
```

yyoncho:
cc @kiennq

kiennq:
I see no problem though?
![image](https://user-images.githubusercontent.com/2631472/95401232-a6298900-0947-11eb-9db3-24dd1964d03d.png)

yyoncho:
I will test again with vanilla, it might be an issue with my config.

yyoncho:
I confirm that it reproduces with lsp-start-plain.el

yyoncho:
In addition to that 
```
(when (lsp-completion--looking-back-trigger-characterp trigger-chars)
          (setq this-command 'self-insert-command))
```
in the exit function triggers completion in an inappropriate place. I suggest either deleting it or disabling it by default.

kiennq:
> In addition to that
> 
> ```
> (when (lsp-completion--looking-back-trigger-characterp trigger-chars)
>           (setq this-command 'self-insert-command))
> ```
> 
> in the exit function triggers completion in an inappropriate place. I suggest either deleting it or disabling it by default.

Oh, that has been there for a long time. Do you know why it's added there?

kiennq:
> I confirm that it reproduces with lsp-start-plain.el

Do you have log for that? it seems working with mine in `lsp-start-plain.el` though
![image](https://user-images.githubusercontent.com/2631472/95474581-3c999100-09c0-11eb-82a8-253a0b6be71a.png)

yyoncho:
[...]

PR #2238, opened 2020-10-08
completion-exit-fn: remove unnecessary trigger self-insert-command

kiennq:
We no longer need that part of code as it can wrongly trigger completion unneeded and as suggested in https://github.com/emacs-lsp/lsp-mode/issues/2230#issuecomment-705588534
Language server should trigger completion via `command` field of completion item with `editor.action.triggerSuggest`.


yyoncho (approved):

kiennq:
@yyoncho Can you merge this? It seems `cask` installing has some problems and I cannot merge this without all CIs are green.

PR #2240, opened 2020-10-08
Make sure that {} is presented when using plists

### LINKED ISSUE #1807: lsp-rust diagnostic is not updated when using lsp-use-plists
With `lsp-use-plists` enabled, the diagnostic information for `lsp-rust` is not updated.
Sample rust file:
``` rust
fn main() {
    let foobar = 1;
    let Foobar: i32 = 2;
    // foobar = 2;
}
```

After removing the comment (`//`), the diagnostic is not updated

yyoncho:
- fixes #1807

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6108,6 +6108,12 @@ WORKSPACE is the active workspace."
   (let ((json-encoding-pretty-print t))
     (json-encode (json-read-from-string msg))))
 
+(defvar lsp-parsed-message nil
+  "This will store the string representation of the json message.
+
+In some cases like #1807 we lose information during json
+deserialization.")
+
 (defun lsp--create-filter-function (workspace)
   "Make filter for the workspace."
   (let ((body-received 0)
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
:clap:

kiennq (approved):

kiennq:
I will merge this, is there anything else blocking?

ericdallo:
I think this is what was missing to enable plists on lsp-mode by default, so I think we are good and we can enable lsp-use-plists, right @yyoncho

yyoncho:
> I think this is what was missing to enable plists on lsp-mode by default, so I think we are good and we can enable lsp-use-plists, right @yyoncho ?

We can. But IMO we should put a message that the user should recompile or delete and install again in a new session all lsp-mode related packages. It should be possible to turn that message off.

ericdallo:
This PR doesn't do that @yyoncho? https://github.com/emacs-lsp/lsp-mode/pull/1861

kiennq:
@yyoncho @ericdallo I've merged this PR. Looking forward to `lsp-use-plists` to be enabled by default.

Issue #1845, opened 2020-06-26
lsp starting stuck when open tramp file.


tshu-w:
**Describe the bug**
lsp starting stuck when open tramp file..

**To Reproduce**
1. `emacs -q -l lsp-start-plain.el`
2. `C-x C-f /ssh:host_name:~/test/test.py` to open an empty project with an empty python file.

my lsp-start-plain.el
```emacs-lisp
(require 'package)

(setq debug-on-error t)

(let* ((package-archives '(("melpa" . "https://melpa.org/packages/")
                           ("gnu" . "https://elpa.gnu.org/packages/")))
       (no-byte-compile t)
       (package-user-dir (expand-file-name (make-temp-name "lsp-tmp-elpa")
                                           user-emacs-directory))
       (custom-file (expand-file-name "custom.el" package-user-dir))
       (pkg-list '(lsp-mode lsp-ui yasnippet lsp-java lsp-python-ms lsp-haskell helm-lsp lsp-treemacs dap-mode lsp-origami lsp-dart company flycheck)))

  (package-initialize)
  (package-refresh-contents)

  (mapcar (lambda (pkg)
            (unless (package-installed-p pkg)
              (package-install pkg))
            (require pkg))
          pkg-list)

  (add-hook 'prog-mode-hook 'lsp)
  (add-hook 'kill-emacs-hook `(lambda ()
                                (delete-directory ,package-user-dir t))))

(provide 'lsp-start-plain)
;;; lsp-start-plain.el ends here

(require 'tramp)
(add-to-list 'tramp-remote-path 'tramp-own-remote-path) ;; make execute-find pyls on remote

(with-eval-after-load 'lsp-mode
  (setq lsp-print-io t)

  (lsp-register-client
   (make-lsp-client :new-connection (lsp-tramp-connection "pyls")
                    :major-modes '(python-mode)
                    :remote? t
                    :server-id 'pyls-remote)))
```

**Expected behavior**
lsp start server correcting.

**Which Language Server did you use**
lsp-python

**OS**
Local: macOS GNU Emacs 27.0.91 (build 1, x86_64-apple-darwin19.5.0, NS appkit-1894.50 Version 10.15.5 (Build 19F101)) of 2020-06-20
remote: ubuntu 

**Error callstack**
there is no error.
maybe related:

\*lsp-log\*:
```
Command "pyls" is present on the path.
Command "pyls" is present on the path.
[...]

PR #2253, opened 2020-10-14
Using full remote path for tramp (fix #2251)

### LINKED ISSUE #2251: Cannot use tramp connections with full path
**Describe the bug**
Using a full path for a tramp connection is not supported.

**To Reproduce**
```elisp
(defun my/full-path-for-remote-pyls ()
  (if-let* ((remote-prefix (file-remote-p default-directory))
    ;; only activate on remote buffers
    (concat
         ;; with this prefix below, command is correctly detected and a command is executed in remote host
         ;; however, the command will be incorrect: it will include the prefix
         ;; something like: `ssh remotehost "ssh://remotehost:/non/standard/path/bin/pyls"`
         ;; if we remove this prefix, the command will be correct: `ssh remotehost /non/standard/path/bin/pyls`
         ;; but it will not be properly tested (see https://github.com/emacs-lsp/lsp-mode/issues/1688 )
         remote-prefix
         "/non/standard/path/bin/pyls"))
(lsp-register-client
   (make-lsp-client
    :new-connection (lsp-tramp-connection
                     'my/full-path-for-remote-pyls)
    :major-modes '(python-mode)
    :remote? t
    :priority -1
    :server-id 'pyls-remote))
```
**Expected behavior**
Remote Language Server manages remote/tramp buffer.

Applies to all Language Servers

**OS**
GNU/Linux on local and remote machine.

**Error callstack**
No error callstack (errors are `condition-case` -ed)

FelipeLema:
Make "using a full path on remote hosts" consistent (by always providing tramp prefix).

I didn't find any tramp-related unit tests, so I didn't add any. Let me know if they should be added.

### DIFF
--- a/docs/page/remote.md
+++ b/docs/page/remote.md
@@ -25,6 +25,23 @@ Here it is example how you can configure python language server to work when usi
 
 _Note:_ when you do not have root privileges on the remote machine to put the language server on the path you may alter the remote path by changing `tramp-remote-path`.
 
+When providing a full path for a remote Language Server, it should be with the proper Tramp prefix. We recommend that you provide the command as a separate function so you can control different settings in a single place.
+
+```elisp
+(defun my/remote-lsp-command ()
+  "Use pyls, but only on remoteA and on a custom location"
+  (when (and (file-remote-p default-directory)
+             (s-matches? "remoteA" default-directory))
+    (concat (file-remote-p default-directory)
+            "/home/me/bin/pyls")))
+(lsp-register-client
+    (make-lsp-client :new-connection 'my/remote-lsp-command)
+                     :major-modes '(python-mode)
+                     :remote? t
+                     :server-id 'remoteA-pyls-remote))
+```
+
+
 ### Dealing with stderr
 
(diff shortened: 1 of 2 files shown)


FelipeLema:
I'm not getting this to work consistently, so I'm currently checking out what's missing

FelipeLema:
I don't know why, but `lsp` won't connect unless `lsp-log-io` has been set to non-nil value *once*. Like
1- boot up `lsp`
2- `lsp` will keep waiting for start
3- `clangd` reports `JSON parse error: [3:2077, byte=2079]: Expected , or } after object property`

I'll repeat 1-3 indefinitely. Now, if I set `lsp-log-io` to `t`, then
1- boot up `lsp`
2- connects correctly to `clangd`
And will keep doing it so correctly ever after... 

I don't know if this is related to my change... will be researching.

FelipeLema:
May be related: https://github.com/emacs-lsp/lsp-mode/issues/1845

FelipeLema:
closing because is out-of-place

see https://github.com/emacs-lsp/lsp-mode/issues/2251#issuecomment-712535886

PR #2257, opened 2020-10-16
Add `lsp-headerline--check-breadcrumb' to local `xref-after-jump-hook'

leungbk:
This allows the breadcrumb to be updated while mashing
xref-next/prev-line in an xref-find-references session. Without this
change, it is necessary to move back to the lsp-managed buffer to see
the breadcrumb update.

The drawback to this is that `lsp-headerline--check-breadcrumb` will generally get called twice if we use `xref-find-definitions`; once in the `xref-after-jump-hook` and once in the `lsp-on-idle-hook`. Not sure if there's an easy fix for this.

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -323,6 +323,8 @@ PATH is the current folder to be checked."
     (lsp-headerline--breadcrumb-cache-path-up-to-project)
     (add-to-list 'header-line-format '(t (:eval lsp-headerline--string)))
 
+    (add-hook 'xref-after-jump-hook #'lsp-headerline--check-breadcrumb nil t)
+
     (add-hook 'lsp-on-idle-hook #'lsp-headerline--check-breadcrumb nil t)
     (add-hook 'lsp-configure-hook #'lsp-headerline--enable-breadcrumb nil t)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3611,6 +3613,9 @@ in that particular folder."
     (when lsp-enable-xref
       (add-hook 'xref-backend-functions #'lsp--xref-backend nil t))
 
+    (when lsp-headerline-breadcrumb-enable
```
kiennq, line 3616: These should be added in `lsp-headerline-breadcrumb-mode`, to `lsp-configure-hook` and `lsp-unconfigure-hook`

kiennq:
> Without this change, it is necessary to move back to the lsp-managed buffer to see the breadcrumb update.

I'm not fully understand the problem, is that happens because we change position inside not focused buffer?

leungbk:
> I'm not fully understand the problem, is that happens because we change position inside not focused buffer?

Yeah, I think so.

yyoncho:
@kiennq do you have more input on that? To me, it seems like it is good to go?

kiennq (approved):

yyoncho (approved):

PR #2271, opened 2020-10-18
Bugfix/xref syntax highlighting

nbfalcon:
When using xref to show all references/definitions/... of some symbol, all entries were gray, even though they had syntax highlighting in the buffer where they were defined. This was because `lsp--extract-line-from-buffer', used by lsp's xref routines, used
`buffer-substring-no-properties' to extract lines. Use `buffer-substring' instead.

If needed, I can rebase this against master.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4481,8 +4481,7 @@ and the position respectively."
          (inhibit-field-text-motion t))
     (save-excursion
       (goto-char point)
-      (buffer-substring-no-properties (line-beginning-position)
-                                      (line-end-position)))))
+      (buffer-substring (line-beginning-position) (line-end-position)))))
 
 (lsp-defun lsp--xref-make-item (filename (&Range :start (start &as &Position :character start-char :line start-line)
                                                  :end (end &as &Position :character end-char)))


lsp-mode.el
```diff
@@ -480,32 +480,32 @@ the server has requested that."
 ;;;###autoload(put 'lsp-enable-file-watchers 'safe-local-variable #'booleanp)
 
 (defcustom lsp-file-watch-ignored '(; SCM tools
```
kiennq, line 482: this changes is not related.

nbfalcon: Yup, layered commits. All of my current lsp-mode PRs have this issue, except the first one.

nbfalcon:
And that one is rebased.

lsp-mode.el
nbfalcon: Fixed it.

nbfalcon:
I just removed the proper-regexes commits, so this now only contains a single trivial change, and I rebased against master.

kiennq:
This is still not correctly rebased :(.
Your changes are consist of 2 commits which include unnecessary change.

leungbk:
@nbfalcon The easiest way to handle this is to, on your current `xref` branch, go to `magit-status` and press `A d RET <select your master branch> RET` to donate this last commit to the master branch. Then, still on your `xref` branch, `O h` to reset your `xref` branch to your new master branch.

lsp-mode.el
```diff
@@ -480,12 +480,12 @@ the server has requested that."
...
-                                    "[/\\\\]\\.git$"
```
seagle0128, line 483: Question: why changed these regexp?

nbfalcon: I resolved conflicts incorrectly.

nbfalcon:
Now it is only a single SLOC commit - sorry for the trouble.

kiennq (approved):

PR #2272, opened 2020-10-18
Bugfix/php ls works

nbfalcon:
Make php-ls actually work now, leveraging reverse-tcp.

Refactor `lsp-tcp-connection' and `lsp-tcp-server' to support a :test-fn argument, which should fix an issue that I think I saw by @leungbk.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6237,14 +6237,20 @@ Return a nested alist keyed by symbol names. e.g.
   (when menu-bar-mode
     (lsp--imenu-refresh)))
 
+(defun lsp--check-command-list (command)
+  (cl-assert (seq-every-p (apply-partially #'stringp) command) nil
+             "Invalid command list"))
+
 (defun lsp-resolve-final-function (command)
   "Resolve final function COMMAND."
   (-let [command (if (functionp command) (funcall command) command)]
     (cl-etypecase command
       (list
-       (cl-assert (seq-every-p (apply-partially #'stringp) command) nil
-                  "Invalid command list")
+       (lsp--check-command-list command)
        command)
+      (vector
+       (lsp--check-command-list command)
+       (cl-coerce command 'list))
       (string (list command)))))
 
 (defun lsp-server-present? (final-command)
@@ -6364,34 +6370,25 @@ returned by COMMAND is available via `executable-find'"
         (file-error (setq success t))))
     port))
 
-(defun lsp--command-present? (command)
-  "Check if COMMAND can be started.
-COMMAND shall be a command as returned by the COMMAND-FN argument
-of the various server starting functions.
+(defun lsp-resolve-tcp-function (command-fn port)
+  "Get a command from COMMAND-FN.
+COMMAND-FN shall be a function that takes a single integer
+argument, PORT, and yields a command (see
+`lsp-resolve-final-function')."
+  (lsp-resolve-final-function (funcall command-fn port)))
 
-It may be an argv-vector (list or vector) or a string, specifying
-a shell command."
-  (when-let ((program
-              ;; `nth' to support vectors
-              (or (and (or (listp command) (vectorp command)) (nth 0 command))
-                  (car (split-string-and-unquote command)))))
-    (if (file-remote-p default-directory)
-        (executable-find program t)
-      (executable-find program))))
-
-(defun lsp--command-fn-present? (command-fn)
+(defun lsp--tcp-command-present (command-fn)
   "Make a checker-function for COMMAND-FN.
 The result is a function that, if called, checks wether
 COMMAND-FN (as defined in `lsp-tcp-connection') yields a command
 that can be executed."
-  (lambda () (lsp--command-present? (funcall command-fn 0))))
+  (lambda () (lsp-server-present? (lsp-resolve-tcp-function command-fn 0))))
 
 (defun lsp-tcp-connection (command-fn &optional test-fn)
   "Return an lsp connection object for TCP servers.
 COMMAND-FN will be called to start the server and shall take a
 single argument, the port, which is an int. It should return a
-command to launch the server in a way that listens on that port.
-The command should be an argv-vector (as a list).
+command to launch the server (see `lsp-resolve-final-function').
 
 TEST-FN is a function that should yield a truthy value if and
 only if the server can be started. If this is unspecified,
@@ -6403,7 +6400,7 @@ COMMAND-FN will be called with PORT=0 and its output examined."
      (let* ((host "localhost")
             (port (lsp--find-available-port host (cl-incf lsp--tcp-port)))
             (command (funcall command-fn port))
-            (final-command (if (consp command) command (list command)))
+            (final-command (lsp-resolve-tcp-function command port))
             (_ (unless (executable-find (cl-first final-command))
                  (user-error (format "Couldn't find executable %s" (cl-first final-command)))))
             (process-environment
@@ -6420,7 +6417,7 @@ COMMAND-FN will be called with PORT=0 and its output examined."
        (set-process-query-on-exit-flag tcp-proc nil)
        (set-process-filter tcp-proc filter)
        (cons tcp-proc proc)))
-   :test? (or test-fn (lsp--command-fn-present? command-fn))))
+   :test? (or test-fn (lsp--tcp-command-present command-fn))))
 
 (defalias 'lsp-tcp-server 'lsp-tcp-server-command)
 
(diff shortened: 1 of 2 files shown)


nbfalcon:
I think php-ls is better than serenata, but php-intellephense sadly still beats it (modern features like syntax highlighting, highlighting occurrences). Perhaps with some love it could be made into a powerful language server, and I might possibly maintain it a bit.

nbfalcon:
I am sorry for the awkward PR-layering and I will rebase any of them if needed.

nbfalcon:
The last CI failure seems related to installation?

lsp-mode.el
```diff
@@ -7563,13 +7596,13 @@ When ARG is t the lsp mode will start new language server even if
...
 ARG is nil current file will be opened in multi folder language
 server if there is such. When `lsp' is called with prefix
-argument ask the user to select which language server to start. "
+argument ask the user to select which language server to start."
   (interactive "P")
 
   (when (and lsp-auto-configure (not lsp--client-packages-required))
     (seq-do (lambda (package)
               ;; loading client is slow and `lsp' can be called repeatedly
-              (unless (featurep package) (require package nil t)))
```
kiennq, line 7572: This is intended. why removing `unless` stuff?

lsp-mode.el
```diff
@@ -480,32 +480,32 @@ the server has requested that."
 ;;;###autoload(put 'lsp-enable-file-watchers 'safe-local-variable #'booleanp)
 
 (defcustom lsp-file-watch-ignored '(; SCM tools
```
kiennq, line 482: This change is not related.

lsp-mode.el
nbfalcon: `require' already checks whether the feature is provided, meaning that we are doing redundant work. The `require' becomes a no-op if package is `featurep'.

lsp-mode.el
```diff
@@ -4480,8 +4480,7 @@ and the position respectively."
          (inhibit-field-text-motion t))
     (save-excursion
       (goto-char point)
-      (buffer-substring-no-properties (line-beginning-position)
```
kiennq, line 4483: The buffer substring no properties is smaller and potential faster

lsp-mode.el
nbfalcon: True. My latest PRs are layered on top of each other, because I needed all of them applied to my local lsp-mode. Do you want me to rebase this one?

lsp-mode.el
nbfalcon: It might be, but it removes syntax highighting from xref.

nbfalcon: What I mean is this: `ivy-xref' uses that string, with its properties, to show references (that string becomes `xref-item-summary`). If the properties are removed, the text becomes gray, which is arguably worse than a pretty, syntax-highlighted usage line.

lsp-mode.el
kiennq: Yes, please separate them.

lsp-mode.el
```diff
@@ -6361,14 +6360,20 @@ Return a nested alist keyed by symbol names. e.g.
   (when menu-bar-mode
     (lsp--imenu-refresh)))
 
+(defun lsp--check-command-list (command)
```
kiennq, line 6363: docs?

lsp-mode.el
kiennq: I see. nice

lsp-mode.el
nbfalcon: Okay, I'll do that tomorrow, since I don't have much time right now.

lsp-mode.el
kiennq: I think you should rebase and create separate branch for each PR. For now, it will mix thing too much if you make PR right from your `master` branch.

lsp-mode.el
kiennq: Well, there's a bug that reloading clients is slow, and the solution is to check `featurep` before requiring again.
Is there docs/notes saying that `require` will not load the package again?

If the file only contains function definition, it would fast to be reloaded since the function's not required to be reparse again, however, client packages contains `lsp-register-client`, which is run every time file is loaded, thus slow.
See https://github.com/emacs-lsp/lsp-mode/pull/1498

nbfalcon: "If feature FEATURE is not loaded, load it from FILENAME."

Also, try the following: (benchmark-run 100000 (require 'lsp-mode)). We don't even have 100k servers, and that completes in under a second.

lsp-mode.el
```diff
@@ -6361,14 +6360,20 @@ Return a nested alist keyed by symbol names. e.g.
...
 (defun lsp-resolve-final-function (command)
   "Resolve final function COMMAND."
   (-let [command (if (functionp command) (funcall command) command)]
     (cl-etypecase command
       (list
-       (cl-assert (seq-every-p (apply-partially #'stringp) command) nil
-                  "Invalid command list")
+       (lsp--check-command-list command)
        command)
+      (vector
```
kiennq, line 6374: Is there case of command is vector?

lsp-mode.el
nbfalcon: `featurep' is slightly faster: when called 100k times, it is in total 60ms faster. But for the max 100 servers we have, any speedups are negligible.

lsp-mode.el
nbfalcon: Not really, I just thought that it might be a call feature.

lsp-mode.el
```diff
@@ -6426,31 +6431,33 @@ returned by COMMAND is available via `executable-find'"
                                                (seq-every-p (lambda (el)
                                                               (stringp el))
                                                             l))))))
-  (list :connect (lambda (filter sentinel name environment-fn)
```
kiennq, line 6429: If there's no real change in this, I think we should not touch this, it makes review code harder.

nbfalcon: Sorry, I thought it would be better to have everything be below < 80 columns.

lsp-mode.el

lsp-mode.el
kiennq: > If feature FEATURE is not loaded, load it from FILENAME.

I'm not sure if that mean `if feature is loaded, not load it again`??
From the profiler in #1498, it said otherwise though.
If there's new behavior of `require`, did it happens in `Emacs 26`?

I think the slowness comes from code that's executed when file is loaded. Requiring `lsp-mode`, which doesn't execute `lsp` on top level, is not much different.

kiennq: Let's continue discussion here https://github.com/emacs-lsp/lsp-mode/pull/2208/files#r507208878.
I think we can get that PR merge first so it will be easier for you to separate your work

nbfalcon: It is only slow the first time it is run, because require isn't fast.

Here is how require works in Emacs:
There is a set of symbols, called the "features". To check whether a feature is provided (a feature just being a symbol), the `feature` function, implemented in C, checks whether that symbol is in that. A feature can be added to that set using the `provide` function, which is why all modules should end in (provide 'name).

Require first checks if some feature is provided, and, if not, tries to find the module that it needs to load. This is slow, since it has to check lots of files (the entire `load-path`, 327 directories on my system), but only has to be done once. So (require 'foo) might be somewhat slow the first time, but 'foo will then be provided, so subsequent requires are fast. The issue with lsp-mode was (is?) that some clients which aren't installed are required. This always fails, since those clients aren't in the `load-path`, meaning that those clients never get their features provided, so the `load-path` has to be traversed all the time, which is slow.

I hope that clears things up.

nbfalcon:
Okay, I just rebased this against; I will do that for my XREF commit, which will make my PRs not-layered. I think im slowly getting the hang of git :)

lsp-mode.el
```diff
@@ -6364,83 +6376,109 @@ returned by COMMAND is available via `executable-find'"
...
-                (set-process-query-on-exit-flag tcp-proc nil)
-                (set-process-filter tcp-proc filter)
-                (cons tcp-proc proc)))
-   :test? (lambda () (executable-find (cl-first (funcall command-fn 0))))))
+   :connect
+   (lambda (filter sentinel name environment-fn)
+     (let* ((host "localhost")
+            (port (lsp--find-available-port host (cl-incf lsp--tcp-port)))
+            (command (funcall command-fn port))
+            (final-command (lsp-resolve-tcp-function command port))
```
kiennq, line 6409: Is this correct? The `command` is result of `(funcall command-fn port)` already, and you're going to treat it as function again in `lsp-resolve-tcp-function`?

Issue #2260, opened 2020-10-16
zig goto definition not working: wrong-type-argument sequencep


travisstaloch:
Forgive me for not following the template.  After posting this to [lsp-ui](https://github.com/emacs-lsp/lsp-ui/issues/525), it was suggested that i raise this issue here.  I'm using emacs 27.1 and lsp-mode/lsp-ui are freshly installed in the past two days. This issue happens when using <https://github.com/zigtools/zls> and suggested configs shown at bottom.

When selecting 'goto definition' from the right click context menu, i get the following:

```
Error running timer: (wrong-type-argument sequencep #s(hash-table size 65
 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("uri" "file:///.../file.zig" 
"range" #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 
data ("start" #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 
0.8125 data ("line" 471 "character" 15)) "end" #s(hash-table size 65 test equal 
rehash-size 1.5 rehash-threshold 0.8125 data ("line" 471 "character" 15)))))))
```

Also, i'm seeing the following on startup even though hover works fine.  This has also been reported as a [zls (zig language server) issue](https://github.com/zigtools/zls/issues/110):
```
Error (use-package): lsp-ui/:config: 
  The connected server(s) does not support method textDocument/hover.
```

In zig's irc, they suggest this is likely some type of emacs error.  Goto definition with zls is working pretty well in vscode, vim, kakoune.  Any ideas what could be happening or do I need to provide more information?  I'm mainly hoping to be able to use the goto definition feature with zig. 

```lisp
(require 'lsp)
(add-to-list 'lsp-language-id-configuration '(zig-mode . "zig"))
(lsp-register-client
  (make-lsp-client
    :new-connection (lsp-stdio-connection "/path/to/zls/here")
    :major-modes '(zig-mode)
    :server-id 'zls))

(use-package lsp-mode
  :commands lsp
  :hook ((zig-mode) . lsp)
  :bind (:map lsp-mode-map
              ;; ("TAB" . completion-at-point)
	      ))

(use-package lsp-ui
  :hook (lsp-mode . lsp-ui-mode)
  :config
  (setq lsp-ui-sideline-enable t)
  (setq lsp-ui-sideline-show-hover nil)
  (setq lsp-ui-doc-position 'bottom)
[...]

PR #2276, opened 2020-10-19
Quick fix for invalid seq.

jcs090218:
This patch is a quick fix for #2260 from @shajra.

I think this is a good temporary fix. Can anyone review it?

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5687,14 +5687,17 @@ perform the request synchronously."
                    (display-action . ,display-action))))
     (xref--show-xrefs xrefs display-action)))
 
+(cl-defmethod seq-empty-p ((ht hash-table))
+  "Function `seq-empty-p' for hash-table."
+  (hash-table-empty-p ht))
+
 (cl-defun lsp-find-locations (method &optional extra &key display-action references?)
   "Send request named METHOD and get cross references of the symbol under point.
 EXTRA is a plist of extra parameters.
 REFERENCES? t when METHOD returns references."
   (let ((loc (lsp-request method
                           (append (lsp--text-document-position-params) extra))))
-    (if (or (seq-empty-p loc)
-            (and (hash-table-p loc) (hash-table-empty-p loc)))
+    (if (seq-empty-p loc)
         (message "Not found for: %s" (thing-at-point 'symbol t))
       (lsp-show-xrefs (lsp--locations-to-xref-items loc) display-action references?))))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5693,7 +5693,7 @@ EXTRA is a plist of extra parameters.
 REFERENCES? t when METHOD returns references."
   (let ((loc (lsp-request method
                           (append (lsp--text-document-position-params) extra))))
-    (if (seq-empty-p loc)
+    (if (or (null loc) (hash-table-empty-p loc))
         (message "Not found for: %s" (thing-at-point 'symbol t))
       (lsp-show-xrefs (lsp--locations-to-xref-items loc) display-action references?))))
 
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -5693,7 +5693,7 @@ EXTRA is a plist of extra parameters.
 REFERENCES? t when METHOD returns references."
   (let ((loc (lsp-request method
                           (append (lsp--text-document-position-params) extra))))
-    (if (seq-empty-p loc)
+    (if (or (null loc) (hash-table-empty-p loc))
```
kiennq, line 5696: The check of `seq-empty-p` and `(null)` is not equivalent, for example when `loc` is vector.

The problem here is `seq-empty-p` don't know how to deal with hash table, how about define an `cl-defmethod` for `seq-empty-p` with hash table specializer? Or you can define new `lsp-seq-empty?` that works.

Btw, the `loc` here can be either `list`, `vector` or `object` of `plist` or `hash table` depends on if `plist` feature is enabled or not

shajra: Okay, my elisp is not so great, but I understand if `loc` can be all of these object types.  I'll see if I can follow advice regarding `cl-defmethod` and specializers.  Maybe someone will have a fix before I get to it.

lsp-mode.el
```diff
@@ -5693,7 +5693,8 @@ EXTRA is a plist of extra parameters.
 REFERENCES? t when METHOD returns references."
   (let ((loc (lsp-request method
                           (append (lsp--text-document-position-params) extra))))
-    (if (or (null loc) (hash-table-empty-p loc))
+    (if (or (seq-empty-p loc)
+            (and (hash-table-p loc) (hash-table-empty-p loc)))
```
jcs090218, line 5697: How about this? I think this is simple enough for now.

> The problem here is seq-empty-p don't know how to deal with hash table, how about define an cl-defmethod for seq-empty-p with hash table specializer? Or you can define new lsp-seq-empty? that works.

I don't know if we need to specify a `cl-defmethod` here. Do we use it somewhere else besides this place? 😕 

> Btw, the loc here can be either list, vector or object of plist or hash table depends on if plist feature is enabled or not

Got it. I have reverted back to `seq-empty-p`; and by pass in `nil` to function `seq-empty-p` also returns true, which means it handle the same way as the function `null`.

Ping. @kiennq and @shajra

lsp-mode.el
jcs090218: See https://github.com/emacs-lsp/lsp-mode/pull/2276#discussion_r508252375

lsp-mode.el
kiennq: > I don't know if we need to specify a `cl-defmethod` here. Do we use it somewhere else besides this place?

I think we tend to have wrapper for that (like `lsp-seq-first`), but I think a simple `cl-defmethod` would also solve the problem cleanly
Some thing like
``` elisp
(cl-defmethod seq-empty-p ((ht hash-table))
  (ht-empty? ht))
```
Based on the spec https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_definition, the response can be

> Response:
> - result: Location | Location[] | LocationLink[] | null
> - partial result: Location[] | LocationLink[]
> - error: code and message set in case an exception happens during the definition request

thus, the hash table here is not relevant and depends on `lsp-use-plists` is enabled or not. For better abstraction, I think we shouldn't check the hash table type here, but instead using either wrapper or an `cl-defmethod` for `seq-empty-p`

jcs090218: Got it! I see want you mean here. 

Everything should be working now! Thanks!

kiennq (approved):

jcs090218:
Merged this quick fix. Any further improvement, please make another commitment! Thank you!

Issue #2280, opened 2020-10-22
Error when selecting 'fill match arms' code action: (void-function yas-expand-snippet)


jimblandy:
In lsp-mode-20201021.1658, clicking on the light bulb in the mode line to fill in a `match` statement raises an error saying that `yas-expand-snippet` is not defined, at the backtrace below.

I do not have `yasnippet` installed.

```
Debugger entered--Lisp error: (void-function yas-expand-snippet)
  yas-expand-snippet("\\{\n            Some(${0:_}) => \\{}\n            Non..." 4829 4894 nil)
  lsp--expand-snippet("{\n            Some(${0:_}) => {}\n            None ..." 4829 4894)
  #f(compiled-function (edit) #<bytecode 0xdcf815>)(#<hash-table equal 3/3 0xe649e1>)
  mapc(#f(compiled-function (edit) #<bytecode 0xdcf815>) [#<hash-table equal 3/3 0xe649e1>])
  lsp--apply-text-edits([#<hash-table equal 3/3 0xe649e1>])
  lsp--apply-text-document-edit(#<hash-table equal 2/2 0xe64989>)
  mapc(lsp--apply-text-document-edit (#<hash-table equal 2/2 0xe64989>))
  seq-do(lsp--apply-text-document-edit (#<hash-table equal 2/2 0xe64989>))
  lsp--apply-workspace-edit(#<hash-table equal 1/1 0xe64951>)
  lsp-execute-code-action(#<hash-table equal 3/3 0xe648f9>)
  #f(compiled-function () (interactive nil) #<bytecode 0xe6b869>)()
  funcall-interactively(#f(compiled-function () (interactive nil) #<bytecode 0xe6b869>))
  call-interactively(#f(compiled-function () (interactive nil) #<bytecode 0xe6b869>) nil nil)
  command-execute(#f(compiled-function () (interactive nil) #<bytecode 0xe6b869>))
```

jcs090218:
I think we are calling external function `yas-expand-snippet` without ensuring the function existence. I think the error is cause by the line here.

https://github.com/emacs-lsp/lsp-mode/blob/9116389c96b2f7ddd537d2842550d2f3bc140119/lsp-mode.el#L3964

I will mark this as bug for now! Thank you for reporting this issue to us!

yyoncho:
The issue here is server-side - Rust Analyzer does not support running without snippets which we try to enforce. The only option ATM to fix that client-side is the user to install yasnippet package.

[...]

PR #2281, opened 2020-10-22
Fixed void function yas-expand-snippet.

jcs090218:
Fix for #2280.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3961,9 +3961,10 @@ LSP server result."
                                                 (forward-line 0)
                                                 (insert offset)))
                                  indent-line-function)))
-    (yas-expand-snippet
-     (lsp--to-yasnippet-snippet snippet)
-     start end expand-env)))
+    (when (featurep 'yasnippet)
+      (yas-expand-snippet
+       (lsp--to-yasnippet-snippet snippet)
+       start end expand-env))))
 
 (defun lsp--apply-text-edits (edits)
   "Apply the EDITS described in the TextEdit[] object."


lsp-mode.el
```diff
@@ -3961,9 +3961,10 @@ LSP server result."
                                                 (forward-line 0)
                                                 (insert offset)))
                                  indent-line-function)))
-    (yas-expand-snippet
-     (lsp--to-yasnippet-snippet snippet)
-     start end expand-env)))
+    (when (featurep 'yasnippet)
```
yyoncho, line 3964: here it is too late. If yasnippet is not present the completion candidate will stay with placeholders. I think that the bug here is that rust-analyzer supports only running with snippet support = true.

jcs090218: > here it is too late.

Any suggestion for the fix? I can try implement it.

> I think that the bug here is that rust-analyzer supports only running with snippet support = true.

Not sure what you mean by this. Can you elaborate a bit more? Thank you!

yyoncho: Check this section - https://github.com/yyoncho/lsp-mode/blob/rust-analyzer-plists/lsp-mode.el#L3263

If we don't have yasnippet, snippetSupport should be false and the server should not send snippets. So the bug is either in this section or in the server implementation.

nbfalcon: Also, wouldn't that break if yasnippet were lazy-loaded, and just not loaded yet? Something like fbound 'yas-minor-mode should be better.

kiennq: Yasnippet is required when loading `lsp-mode` with no error option. So unless Yasnippet is not install, there should be no issue here. We purposely to load Yasnippet before hand because it's slow and having it load here should cause significant lag when user types.

jcs090218: > If we don't have yasnippet, snippetSupport should be false and the server should not send snippets. So the bug is either in this section or in the server implementation.

Ahh... I see what you mean here. So you prefer PR that prevents sending snippet request?

If this is the case, what could be wrong here? I wonder why this is happening? How come it leads to this operation. 😕 

> Yasnippet is required when loading lsp-mode with no error option. So unless Yasnippet is not install, there should be no issue here.

Got it! I have now much understand the code base. ;)

nbfalcon: I don't see a snippetSupport for code actions, so how would not sending snippetSupport for it fix #2281?

kiennq: > I don't see a snippetSupport for code actions, so how would not sending snippetSupport for it fix #2281?

Sending snippet or not is depended on server side. The client side (`lsp-mode`) will tell the server that it cannot support snippet if `yas-snippet` is not presented, now it up to the server side to uphold the LSP spec and do not send anything that contains snippet back

yyoncho:
see https://github.com/emacs-lsp/lsp-mode/issues/2280#issuecomment-723780044

PR #2284, opened 2020-10-26
Handle single-location xref results when `lsp-use-plists' is t

leungbk:
Otherwise, we get a type error when we compile with `lsp-use-plists` set to `t` since the previous `sequencep` check did not distinguish between a plist and a list of plists.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4517,11 +4517,15 @@ and the position respectively."
     (lsp:location-link-target-selection-range loc)))
 
 (defun lsp--locations-to-xref-items (locations)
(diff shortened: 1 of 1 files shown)


kiennq:
Since the type is clear, shouldn't we check if:
- The param is sequence, check if it's first item is type of `lsp-location?`, `lsp-location-link?`
- The param is not sequence, is it `location` or `location-link`.

Using that there's will be no explicitly dealing with hash table and plist in this code

ericdallo (approved):

kiennq (approved):

PR #2290, opened 2020-10-28
lsp-fsharp--project-list: correctly parse response

### LINKED ISSUE #2288: Regression after 0203136efa104a132bd46e1dacdc79ddeca62d8b
I have removed elpa folder completely several times debugging this, and during bisecting, 0203136efa104a132bd46e1dacdc79ddeca62d8b seems to be the first bad commit.

I can no longer initialize with F#, it just hangs for an eternity reporting this (with lsp-use-plists):
```
Error processing message (wrong-type-argument hash-table-p (:jsonrpc "2.0" :id 1 :result (:capabilities (:textDocumentSync (:openClose t :change 1 :save (:includeText t)) :hoverProvider t :completionProvider (:resolveProvider t :triggerCharacters ["." "'"]) :signatureHelpProvider (:triggerCharacters ["(" ","]) :definitionProvider t :implementationProvider t :typeDefinitionProvider t :referencesProvider t :documentHighlightProvider t :documentSymbolProvider t :workspaceSymbolProvider t :codeActionProvider t :codeLensProvider (:resolveProvider t) :documentFormattingProvider t :documentRangeFormattingProvider nil :renameProvider t :foldingRangeProvider t))))
```

Without `lsp-use-plists` it reports something like this:
```
(wrong-type-argument
 listp
 #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 data
               ("Kind" "workspacePeek" "Data" #s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 data
                                                            ("Found" [#s(hash-table size 65 test equal rehash-size 1.5 rehash-threshold 0.8125 data
                                                                                    ("Type" "solution" "Data" ... removed paths etc here))])))))
```

It crashes on `workspacePeek`, it seems.

Checking out the commit before, everything works fine again.

I tried enabling and disabling `lsp-use-plists`, but couldn't figure this one out. What am I missing here?

kiennq:
The response is parsed using old json function (slow) and expecting an `alist`. However a new change in `lsp--start-workspace` let-bind that to `hash-table`. We should fix `lsp-fsharp` to use native json parser when possible and using `lsp-get` primitive.

This also root fix #2288

### DIFF
--- a/clients/lsp-fsharp.el
+++ b/clients/lsp-fsharp.el
@@ -233,11 +233,10 @@ disable if `--backgorund-service-enabled' is not used"
                                 `(:directory ,(lsp-workspace-root)
                                              :deep 10
                                              :excludedDirs ["paket-files" ".git" "packages" "node_modules"])))
-         (data (json-read-from-string (lsp-get response :content)))
-         (found (cdr (assq 'Found (cdr (assq 'Data data)))))
-         (directory (car (seq-filter (lambda (d) (equal "directory" (cdr (assq 'Type d)))) found))))
-    (cdr (assq 'Fsprojs (cdr (assq 'Data directory))))))
-
+         (data (lsp--read-json (lsp-get response :content)))
+         (found (-> data (lsp-get :Data) (lsp-get :Found)))
+         (directory (seq-find (lambda (d) (equal "directory" (lsp-get d :Type))) found)))
+    (-> directory (lsp-get :Data) (lsp-get :Fsprojs))))
 
 ;;;###autoload
 (defun lsp-fsharp--workspace-load (projects)


lsp-mode.el
```diff
@@ -6015,20 +6015,24 @@ WORKSPACE is the active workspace."
 
 (defmacro lsp--read-json (str)
   "Read json string STR."
+  `(let ((json-object-type (if lsp-use-plists
```
yyoncho, line 6018: can we go without these changes? Using the global variables will result in a slow down. It is not big deal I guess but it will still be better to avoid that considering this is on the hot path.

yyoncho (approved):
Looks good! Consider binding json-object-type inside of fsharp method in order to avoid regression in case we touch the core code.

kiennq:
> Looks good! Consider binding json-object-type inside of fsharp method in order to avoid regression in case we touch the core code.

Yeah, maybe, for now `lsp-get` is enough.

PR #2297, opened 2020-11-01
Remove lsp-mode-line, implement modeline using built in capabilities

sirikid:


sirikid:
> What do you mean by "to not multiply entities without necessity"?

There is a special markup for modeline, it seems strange to me not to use it when it fits.

> Does this really have any benefits, given that it isn't even less lines of code?

It's simpler, with fewer function calls. I can format it differently and there will be two lines less also.

sirikid:
It is simpler, with fewer function calls. I can also format it differently and there will be two lines less.

sirikid:
> What do you mean by "to not multiply entities without necessity"?

There is a markup for modeline, it seems strange to me not to use it when it fits.

> Does this really have any benefits, given that it isn't even less lines of code?

It is simpler, with fewer function calls. I can also format it differently and there will be two lines less.

sirikid:
> What do you mean by "to not multiply entities without necessity"?

There is a special markup for modeline, it seems strange to me not to use it when it fits.

> Does this really have any benefits, given that it isn't even less lines of code?

It's simpler, with fewer function calls. I can format it differently and there will be two lines less also.

sirikid:
Idk what is happening, but for me the order of comments is wrong

nbfalcon:
What does this gain us? Sorry if I misunderstood, but didn't you just inline `lsp-mode-line`?

sirikid:
> didn't you just inline `lsp-mode-line`?

Kind of. I remade it using mode line format, unfortunately I could not get rid of `:eval` completely. What for? In order not to multiply entities without necessity.

nbfalcon:
What do you mean by "to not multiply entities without necessity"? Does this really have any benefits, given that it isn't even less lines of code? I think I prefer the `lsp-modeline-format` version we had previously.

lsp-mode.el
```diff
@@ -2580,7 +2580,14 @@ an Elisp regexp."
 (define-minor-mode lsp-mode ""
   nil nil nil
   :keymap lsp-mode-map
-  :lighter (:eval (lsp-mode-line))
+  :lighter
+  '(" LSP["
+    (lsp--cur-workspace
+     (:eval (lsp--workspace-print lsp--cur-workspace))
```
yyoncho, line 2586: there is no need to handle `lsp--cur-workspace`, that leg can be removed.

yyoncho (commented):
Looks good.

lsp-mode.el
sirikid: Done

kiennq (approved):

yyoncho:

Sorry for the delay, I forgot about this PR.

yyoncho:
@sirikid hm, do you know which version of emacs introduced this feature? On my side it works on emacs 28 but does not work on emacs 27.1 release.

sirikid:
> do you know which version of emacs introduced this feature?

Which one? `mode-line-format` was introduced at or before Emacs 18.

> On my side it works on emacs 28 but does not work on emacs 27.1 release.

I tested it on Emacs 27.1 and everything is fine.

kiennq:
It doesn't work in Emacs27.1 Windows due to quote is inserted in lighter. See #2323

blahgeek:
I was using an advice around lsp-mode-line to delete the pid number of the lsp modeline to keep my modeline clean (e.g. modify "clangd:23123" to just "clangd". Sadly after this PR merged, there's no easy way to do this.

Personally I think that keeping a separated function is a better approach than "inlining" it, since it would give more freedom to customization. (Or it would be even better to have supported way to customize the mode line string.)  (I know this may look like https://xkcd.com/1172/ , but it's my workflow :P

sirikid:
@blahgeek you can use [delight][1] or [diminish][2], both also supported as [use-package keywords][3]. My setup:
```elisp
(delight #'lsp-mode
         '(" Λ"
           (lsp--buffer-workspaces
            ("[" (:eval
                  (mapconcat (lambda (w)
                               (symbol-name
                                (lsp--workspace-server-id w)))
                             lsp--buffer-workspaces
                             "/"))
             "]"))))
```

[1]: https://elpa.gnu.org/packages/delight.html
[2]: https://melpa.org/#/diminish
[3]: https://github.com/jwiegley/use-package/#diminishing-and-delighting-minor-modes

nbfalcon:
@blahgeek you can also achieve the same without `delight` and `diminish` (based on @sirikid's answer, not tested though):
```emacs-lisp
(eval-after-load 'lsp-mode
  (setf (nth 1 (assq 'lsp-mode minor-mode-alist))
        '(" Λ"
          (lsp--buffer-workspaces
           ("[" (:eval
                 (mapconcat (lambda (w)
                              (symbol-name
                               (lsp--workspace-server-id w)))
                            lsp--buffer-workspaces
                            "/"))
            "]")))))
```

nbfalcon:
`minor-mode-alist` has the form `((MODE MODELINE-FORM)...)`.

sirikid:
@nbfalcon either quote the form or use `with-eval-after-load`

nbfalcon:
@sirikid I thought that the FORM was quoted by a compiler macro for some reason... Thanks! (Fixed)

PR #2299, opened 2020-11-03
lsp-rust: support rust-analyzer.showReference lens

nbfalcon:
Add a rust-analyzer specific :action-handler for the showReference lens,
which leverages `lsp-show-xrefs'. Refactor: eliminate
`lsp-execute-command', because it caused the :title to be lost, which is
needed to distinguish between the "references" or "implementations"
variants of that lens.

Defining a new `lsp-interface' to destructure
"rust-analyzer.showReference" was impossible, as the former doesn't
support destructuring lists (the :arguments? paramter is a special,
3-element list.).

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -635,8 +635,6 @@ them with `crate` or the crate name they refer to."
 (lsp-defun lsp-rust--analyzer-run-single ((&Command :arguments?))
   (lsp-rust-analyzer-run (lsp-seq-first arguments?)))
 
-;; `lsp-interface' is unsuitable for ShowReference destructuring, as that code
-;; action yields an array of three elements, not a map.
 (lsp-defun lsp-rust--analyzer-show-references
   ((&Command :title :arguments? [_uri _filepos references]))
   (lsp-show-xrefs (lsp--locations-to-xref-items references) nil
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -295,7 +295,7 @@ PARAMS progress report notification data."
         (lsp-workspace-status nil workspace)
       (lsp-workspace-status (format "%s - %s" title (or message? "")) workspace))))
 
-(cl-defmethod lsp-execute-command (_server (_command (eql rls.run)) params)
+(lsp-defun lsp-rust--rls-run ((&Command :arguments? params))
   (-let* (((&rls:Cmd :env :binary :args :cwd) (lsp-seq-first params))
           (default-directory (or cwd (lsp-workspace-root) default-directory) ))
     (compile
@@ -635,6 +635,13 @@ them with `crate` or the crate name they refer to."
 (lsp-defun lsp-rust--analyzer-run-single ((&Command :arguments?))
   (lsp-rust-analyzer-run (lsp-seq-first arguments?)))
 
+;; `lsp-interface' is unsuitable for ShowReference destructuring, as that code
+;; action yields an array of three elements, not a map.
+(lsp-defun lsp-rust--analyzer-show-references
+  ((&Command :title :arguments? [_uri _filepos references]))
+  (lsp-show-xrefs (lsp--locations-to-xref-items references) nil
+                  (s-contains-p "reference" title)))
+
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-stdio-connection
@@ -648,7 +655,8 @@ them with `crate` or the crate name they refer to."
   :priority (if (eq lsp-rust-server 'rust-analyzer) 1 -1)
   :initialization-options 'lsp-rust-analyzer--make-init-options
   :notification-handlers (ht<-alist lsp-rust-notification-handlers)
-  :action-handlers (ht ("rust-analyzer.runSingle" #'lsp-rust--analyzer-run-single))
+  :action-handlers (ht ("rust-analyzer.runSingle" #'lsp-rust--analyzer-run-single)
+                       ("rust-analyzer.showReferences" #'lsp-rust--analyzer-show-references))
   :library-folders-fn (lambda (_workspace) lsp-rust-library-directories)
   :after-open-fn (lambda ()
                    (when lsp-rust-analyzer-server-display-inlay-hints
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -144,7 +144,9 @@ See `lsp-lens--schedule-refresh' for details."
     (define-key [mouse-1] (lsp-lens--create-interactive-command command))))
 
 (defun lsp-lens--create-interactive-command (command?)
-  "Create an interactive COMMAND? for the lens."
+  "Create an interactive COMMAND? for the lens.
+COMMAND? shall be an `&Command' (e.g. `&CodeLens' :command?) and
+mustn't be nil."
   (if (functionp (lsp:command-command command?))
       (lsp:command-command command?)
     (lambda ()
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -145,16 +145,11 @@ See `lsp-lens--schedule-refresh' for details."
 
 (defun lsp-lens--create-interactive-command (command?)
   "Create an interactive COMMAND? for the lens."
-  (let ((server-id (->> (lsp-workspaces)
-                        (cl-first)
-                        (or lsp--cur-workspace)
-                        (lsp--workspace-client)
-                        (lsp--client-server-id))))
-    (if (functionp (lsp:command-command command?))
-        (lsp:command-command command?)
-      (lambda ()
-        (interactive)
-        (lsp--execute-command command?)))))
+  (if (functionp (lsp:command-command command?))
+      (lsp:command-command command?)
+    (lambda ()
+      (interactive)
+      (lsp--execute-command command?))))
 
 (defun lsp-lens--display (lenses)
   "Show LENSES."
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -154,9 +154,7 @@ See `lsp-lens--schedule-refresh' for details."
         (lsp:command-command command?)
       (lambda ()
         (interactive)
-        (lsp-execute-command server-id
-                             (intern (lsp:command-command command?))
-                             (lsp:command-arguments? command?))))))
+        (lsp--execute-command command?)))))
 
 (defun lsp-lens--display (lenses)
   "Show LENSES."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5209,17 +5209,6 @@ It will filter by KIND if non nil."
       (funcall action-handler action)
     (lsp--send-execute-command command arguments?)))
 
-(defun lsp-execute-command (_server command arguments)
-  "Execute code action COMMAND with ARGUMENTS?.
-Note that this function can now longer be overloaded using
-`cl-defmethod', because that proved to be a liability. Use the
-:action-handlers argument of `make-lsp-client' instead."
-  (lsp--execute-command
-   (lsp-make-command :command (symbol-name command)
-                     :arguments? arguments)))
-(make-obsolete 'lsp-execute-command "use `lsp--execute-command' instead."
-               "7.1.0")
-
 (lsp-defun lsp-execute-code-action ((action &as &CodeAction :command? :edit?))
   "Execute code action ACTION.
 If ACTION is not set it will be selected from `lsp-code-actions-at-point'.
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -968,6 +968,9 @@ calling `remove-overlays'.")
 
 (defvar-local lsp--virtual-buffer-point-max nil)
 
+(cl-defgeneric lsp-execute-command (server command arguments)
+  "Ask SERVER to execute COMMAND with ARGUMENTS.")
+
 (defun lsp-elt (sequence n)
   "Return Nth element of SEQUENCE or nil if N is out of range."
   (cond
@@ -5205,9 +5208,18 @@ It will filter by KIND if non nil."
 
 (lsp-defun lsp--execute-command ((action &as &Command :command :arguments?))
   "Parse and execute a code ACTION represented as a Command LSP type."
-  (-if-let* ((action-handler (lsp--find-action-handler command)))
-      (funcall action-handler action)
-    (lsp--send-execute-command command arguments?)))
+  (let ((server-id (->> (lsp-workspaces)
+                        (cl-first)
+                        (or lsp--cur-workspace)
+                        (lsp--workspace-client)
+                        (lsp--client-server-id))))
+    (condition-case nil
+        (prog1 (lsp-execute-command server-id (intern command) arguments?)
+          (lsp--warn "`lsp-execute-command' is deprecated"))
+      (cl-no-applicable-method
+       (if-let ((action-handler (lsp--find-action-handler command)))
+           (funcall action-handler action)
+         (lsp--send-execute-command command arguments?))))))
 
 (lsp-defun lsp-execute-code-action ((action &as &CodeAction :command? :edit?))
   "Execute code action ACTION.
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5209,7 +5209,7 @@ It will filter by KIND if non nil."
       (funcall action-handler action)
     (lsp--send-execute-command command arguments?)))
 
-(defun lsp-execute-command (_ command arguments)
+(defun lsp-execute-command (_server command arguments)
   "Execute code action COMMAND with ARGUMENTS?.
 Note that this function can now longer be overloaded using
 `cl-defmethod', because that proved to be a liability. Use the
@@ -5217,7 +5217,8 @@ Note that this function can now longer be overloaded using
   (lsp--execute-command
    (lsp-make-command :command (symbol-name command)
                      :arguments? arguments)))
-(make-obsolete #'lsp-execute-command #'lsp--execute-command "7.1.0")
+(make-obsolete 'lsp-execute-command "use `lsp--execute-command' instead."
+               "7.1.0")
 
 (lsp-defun lsp-execute-code-action ((action &as &CodeAction :command? :edit?))
   "Execute code action ACTION.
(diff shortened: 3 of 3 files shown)


yyoncho:
Please revert the lsp-execute-command elimination because it is used multiple times in different projects (e. g. lsp-java, ccls, etc).

clients/lsp-rust.el
```diff
@@ -585,6 +585,13 @@ The command should include `--message=format=json` or similar option."
 (lsp-defun lsp-rust--analyzer-run-single ((&Command :arguments?))
   (lsp-rust-analyzer-run (lsp-seq-first arguments?)))
 
+;; `lsp-interface' is unsuitable for ShowReference destructuring, as that code
+;; action yields an array of three elements, not a map.
+(lsp-defun lsp-rust--analyzer-show-references
+  ((&Command :title :arguments? [_uri _filepos references]))
```
kiennq, line 591: This should have different indent or should be on the same line with the function name

clients/lsp-rust.el
```diff
@@ -585,6 +585,13 @@ The command should include `--message=format=json` or similar option."
 (lsp-defun lsp-rust--analyzer-run-single ((&Command :arguments?))
   (lsp-rust-analyzer-run (lsp-seq-first arguments?)))
 
+;; `lsp-interface' is unsuitable for ShowReference destructuring, as that code
+;; action yields an array of three elements, not a map.
```
yyoncho, line 589: lsp-interface is for defining destructuring for objects, not for arrays. dash has destructuring for arrays/lists thus there is no need for lsp-interface to handle that. Thus, this comment can be deleted.

clients/lsp-rust.el
nbfalcon: Then the line would be > 80 columns. Is that preferred?

clients/lsp-rust.el
nbfalcon: Will do that.

nbfalcon:
@yyoncho can I perhaps instead refactor the other projects to not use it? That function needs to be at least be refactored, as it removes the :title field (without it, it is impossible to know whether to use `xref--show-defs' or `xref--show-references'), and it would become a liability in the future, if more fields would be added to the Lens lsp structure, yielding confusing errors because fields become nil out of nowhere.

yyoncho:
> @yyoncho can I perhaps instead refactor the other projects to not use it? That function needs to be at least be refactored, as it removes the :title field (without it, it is impossible to know whether to use `xref--show-defs' or `xref--show-references'), and it would become a liability in the future, if more fields would be added to the Lens lsp structure, yielding confusing errors because fields become nil out of nowhere.

I am fine with this approach.

nbfalcon:
@yyoncho can you give me a list of projects that use it?

nbfalcon:
lsp-java is done.

nbfalcon:
ccls.el is done.

yyoncho:
> @yyoncho can you give me a list of projects that use it?

IDK - you have to check all packages that depend on lsp-mode.

kiennq:
Instead of just deleting code, I think we should probably make that function become obsoleted, and promote `lsp--execute-command` to public function, that ways we will not break any package that depends on `lsp-mode`.
The parameter for the old function `lsp-execute-command` is clear, it will just passing `command` and `arguments`, and not react to anything else, so the user of that API should aware that already.

nbfalcon:
@kiennq actually, `lsp-execute-command` isn't a function that the user calls, it is the other way around: it is a `cl-defgeneric`, which some language server clients provide, instead of using :action-handlers. The problem here is that lens resolution goes trough that function, causing the :title? field to be dropped. This makes it impossible to implement this lens properly, but much worse than that might become a huge blocker in the future. I could theoretically extend `lsp--execute-command` to try to see if a `cl-defmethod` exists, but doing so would involve horrible hacks, and I'd much rather fix all downstream packages instead.

nbfalcon:
Also, from the packages I checked, there are no more uses of it. We'll need to wait for ccls and lsp-java, and then this can be merged.

clients/lsp-rust.el
kiennq: You can put multi-line for the parameters right?
That would make it easier to distinguished between params and body of a function

kiennq:
> actually, `lsp-execute-command` isn't a function that the user calls, it's the other way around: that function is a `cl-defgeneric`, provided by some language server clients. The problem here is that lens resolution goes trough that function, causing the :title? field to be dropped.

Lens resolution is not using `:title` information until now, and for this feature we better switch to another function that also passing the `:title` information, I'm totally agree with that.
But, that's not the reason to delete this function and its functionality.

The `cl-defgeneric` of this function provides an easy way to add new handlers for `:action-handlers` without modifying the `lsp-mode` source code that user can easily extend.

A [quick search](https://github.com/search?q=lsp-execute-command&type=code) of `lsp-execute-command` shows over 1K results, I assume all will be broken (not including private repo) when this change is merged.

nbfalcon:
We cannot use a new function for this. Also, my problem isn't only the :title field, even though it hints the problems that will follow; `lsp-execute-command` was misdesigned: it should pass the entire action trough, not only specific fields of it. Not doing that will become a problem when new fields are added, perhaps even server-specific ones. This API is a liability and needs to be eliminated.

Regarding your github search, most of them are elpa directories containing `lsp-java-boot`. I already made a PR to fix that there. I'll investigate the interesting ones. Your search will be a good starting point, thank you.

nbfalcon:
Nevermind, it's still too many. I propose an alternative: `lsp-execute-command` will still be available as a function, but `cl-defmethod` overrides of it will no longer be possible. This means nothing should be broken.

nbfalcon:
I'll re-add it as a normal defun. It is not overriden using `cl-defmethod` anywhere I didn't provide a PR to fix that though, see [this query]
(https://github.com/search?p=3&q=cl-defmethod+lsp-execute-command+-filename%3Alsp-java-boot.el+-filename%3Accls.el+-filename%3Alsp-java.el+-filename%3Alsp-java.elc+-filename%3Alsp-java-boot.elc+-filename%3Accls.elc+-filename%3A*.elc+-filename%3Alsp-rust.el+-filename%3Alsp-mode.el&type=Code)

Note the exclusions: I have provided PRs for `lsp-java-*` and `ccls.el`. All the results are just copies of `lsp-mode`, ccls.el and lsp-java.el, also in compiled form.

The following PRs need to be merged before this can be, though:
- [lsp-java](https://github.com/emacs-lsp/lsp-java/pull/291)
- [ccls.el](https://github.com/MaskRay/emacs-ccls/pull/104)

lsp-mode.el
```diff
@@ -5167,6 +5164,17 @@ It will show up only if current point has signature help."
       (funcall action-handler action)
     (lsp--send-execute-command command arguments?)))
 
+(defun lsp-execute-command (_server command arguments)
```
yyoncho, line 5167: It seems to be that this does not improve backward compatibility

nbfalcon: It does. Many places call it as a function. I didn't actually find any other unfixed place that uses `cl-defmethod` on this across the entirety of github, only lots of elpa directories containing outdated ccls.elc? and lsp-java-*.elc? files.

yyoncho: ok, this will work, but this is not the usecase covered by this method. It was an extension point, not an entry point supposed to be called.

nbfalcon: That is true, however people abused it like that anyway. Do you consider it acceptable to break all that code? (I actually like breaking changes, but we need to be careful for a project this large) There is no way I can go trough 800 projects to make this slightly more elegant.

yyoncho: I think we should either clean it up or make sure it is fully backward compatible.

nbfalcon: Now I'm unsure: I can't seem to find anyone really using it: [github search](https://github.com/search?q=lsp-execute-command+-filename%3Accls.el+-filename%3Accls.elc+-filename%3Alsp-*&type=Code)

This kak-lsp stuff seems to be unrelated to lsp-mode, and the one use as a keybinding is wrong anyway, since that function wasn't ever interactive.

nbfalcon: @yyoncho by clean up, do you mean remove it? From my latest search, it seems that no one actually uses it.

nbfalcon:
We may want to make a new release for this though, as it would break ccls.el and lsp-java-* for anyone who decides to update `lsp-mode` independently of these other packages.

lsp-mode.el
kiennq: Can we do this in backward compatible way?

Maybe something like
``` elisp
(defvar lsp--from-internal-call nil)
(make-obsolete 'lsp-execute-command 'new-one "7.2")
(cl-defmethod lsp-execute-command (_server command arguments)
  "Dispatch COMMAND execution."
  (if lsp--from-internal-call
      :obsoleted
    (lsp--execute-command (lsp-make-command :command (symbol-name command)
                                            :arguments? arguments)))

(defun lsp-lens--create-interactive-command (command?)
  ;; ...
  (let ((lsp--from-internal-call t))
    (when (eq (lsp-execute-command server-id
                                     (intern (lsp:command-command command?))
                                     (lsp:command-arguments? command?))
                :obsoleted)
      (lsp--execute-command ...))))
```

yyoncho: Either this or binding the full command in `lsp-lens--...` and checking if it is bound in lsp--execute-command. Not sure which one is less ugly.

nbfalcon: @yyoncho so you mean something like `fboundp` `lsp-execute-command` and leaving the latter undefined in `lsp-mode.el`? This would almost work, since `cl-defmethod` introduces the function if it doesn't yet exist. However, it would break code that calls `lsp-execute-command` as a function.

yyoncho: No. The idea of @kiennq  is to make lsp-execute-command work as it is and still pass the title. The @kiennq's proposal can be achieved with some dynamic stope tricks. I am not 100% it worths the effort.

nbfalcon: I'll return it with an obsoletion warning, since I just had the following idea: we can leverage `cl-no-applicable-method`, limiting the amount of ugly hacks. It should still be removed at some point, perhaps in the next major release.

nbfalcon:
`lsp-execute-command` now works as it did previously. I tested this with `ccls` before my PR; however, to get it working I had to do two things: use `lsp-show-xrefs` instead of `xref--show-xrefs` (because the latter takes a function, not a list of xrefs) and to handle symbols in `lsp--send-execute-command` because I was getting errors related to `json-serialize` (wrong type argument `json-value-p` for a symbol). These seem unrelated to the changes though (I'll investigate with master).

nbfalcon:
I tested with master, `ccls`' xref lenses are broken independently of this PR (I have made a PR there which fixes these issues, along with not using `lsp-execute-command`), so I believe that this can be merged.

lsp-mode.el
```diff
@@ -5171,9 +5171,18 @@ It will show up only if current point has signature help."
...
-      (funcall action-handler action)
-    (lsp--send-execute-command command arguments?)))
+  (let ((server-id (->> (lsp-workspaces)
+                        (cl-first)
+                        (or lsp--cur-workspace)
+                        (lsp--workspace-client)
+                        (lsp--client-server-id))))
+    (condition-case nil
+        (prog1 (lsp-execute-command server-id (intern command) arguments?)
+          (lsp--warn "`lsp-execute-command' is deprecated"))
```
kiennq, line 5181: Why don't you just make `lsp-execute-command` obsolete instead? We will have compile warning about that, and that would be standard way to alert package maintainer about obsolete function.

clients/lsp-rust.el
nbfalcon: The options are:
```elisp
(lsp-defun lsp-rust--analyzer-show-references ((&Command
                                                :title :arguments?
                                                [_uri _filepos references]))
```
or
```elisp
(lsp-defun lsp-rust--analyzer-show-references ((&Command :title :arguments?
                                                         [_uri _filepos
                                                               references]))
```
, both of which require two more lines and are IMHO uglier. What do you think?

lsp-mode.el
nbfalcon: That somehow slipped my mind; concluded from testing, calling that function doesn't cause warnings, but overriding it using `cl-defmethod` does. I couldn't find anything in the Emacs docs about the obsoletion of generics, though.

clients/lsp-rust.el
nbfalcon: Fixed.

clients/lsp-rust.el
```diff
@@ -585,6 +585,11 @@ The command should include `--message=format=json` or similar option."
 (lsp-defun lsp-rust--analyzer-run-single ((&Command :arguments?))
   (lsp-rust-analyzer-run (lsp-seq-first arguments?)))
 
+(lsp-defun lsp-rust--analyzer-show-references
+  ((&Command :title :arguments? [_uri _filepos references]))
+  (lsp-show-xrefs (lsp--locations-to-xref-items references) nil
+                  (s-contains-p "reference" title)))
```
yyoncho, line 591: out of curiosity what are the expected `title` values here?

yyoncho (commented):
The code looks good to me. I will proceed with the merge after doing manual testing.

clients/lsp-rust.el
nbfalcon: <any number> reference(s)
<any number> definition(s)
E.g. 2 references, 2 definitions. I'm unsure whether reference can actually be encountered in the wild.

nbfalcon:
There is one issue: it throws warnings because `lsp-execute-command` is called even though it is deprecated.

clients/lsp-rust.el
nbfalcon: As a side note, do we prefer scheme-style `?` or common-lisp-style `p` predicate functions? I.e. `s-contains-p` or `s-contains?`.

nbfalcon:
@yyoncho you may want to hold off merging until [my ccls.el PR](https://github.com/MaskRay/emacs-ccls/pull/104) gets merged, as otherwise users will get inevitable byte-compiler warnings.

clients/lsp-rust.el
yyoncho: > As a side note, do we prefer scheme-style `?` or common-lisp-style `p` predicate functions? I.e. `s-contains-p` or `s-contains?`.

Most of the code is written using scheme-style although this is not enforced 100% during reviews.

yyoncho:
> @yyoncho you may want to hold off merging until [my ccls.el PR](https://github.com/MaskRay/emacs-ccls/pull/104) gets merged, as otherwise users will get inevitable byte-compiler warnings.

Is the ccls pr mandatory?

clients/lsp-rust.el
kiennq: Sorry for late reply, both look fine to me.
The current solution has same indent between args and body and that make it quite hard to distinguish them.

lsp-lens.el
```diff
@@ -145,18 +145,11 @@ BUFFER-MODIFIED? determines whether the buffer is modified or not."
...
-                        (lsp--workspace-client)
-                        (lsp--client-server-id))))
-    (if (functionp (lsp:command-command command?))
-        (lsp:command-command command?)
-      (lambda ()
-        (interactive)
-        (lsp-execute-command server-id
-                             (intern (lsp:command-command command?))
-                             (lsp:command-arguments? command?))))))
+  (if (functionp (lsp:command-command command?))
```
kiennq, line 148: Can `command?` be `nil` here?
If so, don't you need to check before passing it to `lsp--execute-command`?

nbfalcon:
@yyoncho I'd argue yes, because `ccls` users would get annoying depreciation warnings if they chose to upgrade `lsp-mode`. We could also not deprecate that interface, but it kind-of is a second class citizen here, so it should probably be removed at some point.

lsp-lens.el
nbfalcon: I'd guess no, because it wasn't possible previously (this wouldn't be a regression caused by this PR):
```emacs-lisp
    (if (functionp (lsp:command-command command?))
        (lsp:command-command command?)
      (lambda ()
        (interactive)
        (lsp-execute-command server-id
                             (intern (lsp:command-command command?))
                             (lsp:command-arguments? command?))))
; (if (functionp (lsp:command-command nil)) ...) -> (functionp nil) -> nil
```
This means that the control flow passed to the other branch, creating a lambda that would try to `intern` nil, which errors.

```emacs-lisp
(lsp:command-command nil) ; => nil
(intern nil) ; => error
```

nbfalcon: This function is called on line 190 in master (without this PR), and only lenses that have a command are passed to it (`-filter #'lsp:command-command`), so I guess command? cannot be nil. I'll note that in the docstring.

nbfalcon: Fixed.

kiennq: The reason I'm asking is because of parameter name `command?` which indicates it can be `nil`, we should change the parameter name too.

nbfalcon: I wouldn't do that, since it is an Elisp-level change, meaning that this entire thing has to be manually tested again, for something this trivial. See [this comment by @yyoncho](https://github.com/emacs-lsp/lsp-mode/pull/2317#discussion_r520169506). Noting that in the docstring is good enough, IMHO.

Also, the name isn't that bad: `command?` also corresponds to the naming of the `CodeLens` and `CodeAction` `:command?` fields, indicating that this function expects them to be passed in.

nbfalcon:
I would like to propose not deprecating `lsp-execute-command`, given that my `ccls` PR apparently won't be merged too soon. @yyoncho what do you think?

clients/lsp-rust.el
nbfalcon: @yyoncho which do you prefer?

yyoncho: For me all three options are acceptable... Just want to mention that we do not enforce 80 symbols or at least there is no agreement on that ATM.

nbfalcon: Thank you, I'll keep your latter point in mind. Not changing the call here is the best option then, because it means no new changes have to made and reviewed.

yyoncho:
Can you rebase?

yyoncho (approved):

PR #2303, opened 2020-11-05
Auto-configure compile-commands-dir for clangd

### LINKED ISSUE #2278: Try auto-finding the directory with compile_commands.json for lsp-clangd
Running current version of lsp-clangd with a cmake based out-of-source builds, I set `set(CMAKE_EXPORT_COMPILE_COMMANDS ON)`, which results in a compile_commands.json file in my build directory. 

After that I need to copy or symlink it to project root every time I re-configure top-level CML (add a new file, new dependency etc) for clangd to pick up info about my includes and successfully resolve goto-def queries. Symlinking doesn't work when I use Docker for Desktop with a WSL volume mounted to Docker, so I have to manually copy it every time. 

I noticed from `clangd-10 --help`  that it takes the following a flag to specify a directory for clangd to find compile_commands.json. 

```bash
clangd compilation flags options:

  --compile-commands-dir=<string> - Specify a path to look for compile_commands.json. If path is invalid, clangd will look in the current directory and parent paths of each source file
```

The heuristic used by clangd 
> If path is invalid, clangd will look in the current directory and parent paths of each source file

seems quite useless in out-of-source builds. 

I wanted to suggest running a 1-level deep search for a compile_commands.json file from the project root and setting this flag with the directory name for this flag eg. "--compile-commands-dir=build/". 

I don't know if lsp currently has hooks for lsp-server specific project initialisation routines, but I think it would help. Happy to take a crack at it, if it's possible with the current API and is in line with what you have in mind about easy to use default options. 

Amazing work - i appreciate your efforts.

petr-tik:
Fix #2278

Depends on fixing lsp-workspace-root to be available when opening a new workspace

Tested locally on projects with and without a compile_commands.json file.

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -210,14 +210,13 @@ Then return either the default args or the modified args"
   (if
        ;; lsp-client-clangd-args might already lists a
        ;; "--compile-commands-dir=build/" and will fail if we pass another one
-       ;; trust the user - if it's already set, don't mess around
+       ;; trust the user - if compile-commands-dir is already set, don't mess around
       (cl-find "--compile-commands-dir" lsp-clients-clangd-args)
       (progn
-        (lsp--info "compilation_commands_dir set in clangd-args")
+        (lsp--info "compile-commands-dir set in clangd-args")
         lsp-clients-clangd-args)
     (if-let (found-dir (lsp-clangd-find-compile-commands-dir))
-        ;; TODO expensive to create a list from 1 element - how to append atom to list
-        (append lsp-clients-clangd-args (list (concat "--compile-commands-dir=" found-dir)))
+        (cons (concat "--compile-commands-dir=" found-dir) lsp-clients-clangd-args)
       lsp-clients-clangd-args)))
 
 
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -184,20 +184,20 @@ This must be set only once after loading the clang client.")
 If multiple compile_commands.json files are
 found in the project directory tree - ask the user to choose."
   ;; TODO lsp-workspace-root won't find anything at the start of the workspace
-  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "compile_commands.json")))
-    (pcase (length candidate-dirs)
+  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "\\`compile_commands.json\\'")))
+    (pcase candidate-dirs
           ;; TODO can I lsp--warn-user-in-message-and-log at the same time
-      (`0 (lsp--warn "Cannot find compile_commands.json - intellisense might not work")
+      (`nil (lsp--warn "Starting without compile_commands.json - clangd might not be able to handle complex projects")
           nil)
       ;; if you find 1 only - assume it's correct and strip the dirname from the full path
       ;; happy path
-      (`1 (file-name-directory (car candidate-dirs)))
-      ;; if you have several - ask the user to choose with lsp--completing-read
+      (`(,unique-compile-commands-dir) (file-name-directory unique-compile-commands-dir))
+      ;; if you have several - ask the user to choose
       ;; TODO if Ctrl-G from inside this completing read still falls to default lsp-clients-clangd-args
-      (_ (file-name-directory (lsp--completing-read
-           "Found compile_commands.json in several directories, which one do you want to use?"
-           candidate-dirs
-           (lambda (candidate) candidate)))))))
+      (multiple-dirs (file-name-directory (completing-read
+           "Found compile_commands.json in several directories, which one do you want to use?
+Consider setting lsp-clients-clangd-args to avoid this lookup in the future"
+           multiple-dirs))))))
 
 
 (defun lsp-clangd-update-args-with-compile-commands ()
@@ -207,14 +207,14 @@ Else run a potentially interactive search for the directory
 that contains compile_commands.json.
 
 Then return either the default args or the modified args"
-  (if (-first
+  (if
        ;; lsp-client-clangd-args might already lists a
        ;; "--compile-commands-dir=build/" and will fail if we pass another one
        ;; trust the user - if it's already set, don't mess around
-       (lambda (arg) (string-match-p "--compile-commands-dir" arg))
-       lsp-clients-clangd-args)
-      ((lsp--info "compilation_commands_dir set in clangd-args")
-       lsp-clients-clangd-args)
+      (cl-find "--compile-commands-dir" lsp-clients-clangd-args)
+      (progn
+        (lsp--info "compilation_commands_dir set in clangd-args")
+        lsp-clients-clangd-args)
     (if-let (found-dir (lsp-clangd-find-compile-commands-dir))
         ;; TODO expensive to create a list from 1 element - how to append atom to list
         (append lsp-clients-clangd-args (list (concat "--compile-commands-dir=" found-dir)))
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -178,6 +178,48 @@ This must be set only once after loading the clang client.")
   :risky t
   :type '(repeat string))
 
+(defun lsp-clangd-find-compile-commands-dir ()
+  "Return the full directory path (or nil) that contain a compile_commands.json.
+
+If multiple compile_commands.json files are
+found in the project directory tree - ask the user to choose."
+  ;; TODO lsp-workspace-root won't find anything at the start of the workspace
+  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "compile_commands.json")))
+    (pcase (length candidate-dirs)
+          ;; TODO can I lsp--warn-user-in-message-and-log at the same time
+      (`0 (lsp--warn "Cannot find compile_commands.json - intellisense might not work")
+          nil)
+      ;; if you find 1 only - assume it's correct and strip the dirname from the full path
+      ;; happy path
+      (`1 (file-name-directory (car candidate-dirs)))
+      ;; if you have several - ask the user to choose with lsp--completing-read
+      ;; TODO if Ctrl-G from inside this completing read still falls to default lsp-clients-clangd-args
+      (_ (file-name-directory (lsp--completing-read
+           "Found compile_commands.json in several directories, which one do you want to use?"
+           candidate-dirs
+           (lambda (candidate) candidate)))))))
+
+
+(defun lsp-clangd-update-args-with-compile-commands ()
+  "Return the default or updated clangd-args for the start command.
+If --compile-commands-dir is already set - return it.
+Else run a potentially interactive search for the directory
+that contains compile_commands.json.
+
+Then return either the default args or the modified args"
+  (if (-first
+       ;; lsp-client-clangd-args might already lists a "--compile-commands-dir=build/" and will fail if we pass another one
+       ;; trust the user - if it's already set, don't mess around
+       (lambda (arg) (string-prefix-p "--compile-commands-dir" arg))
+       lsp-clients-clangd-args)
+      ((lsp--info "compilation_commands_dir set in clangd-args")
+       lsp-clients-clangd-args)
+    (if-let (found-dir (lsp-clangd-find-compile-commands-dir))
+        ;; TODO expensive to create a list from 1 element - how to append atom to list
+        (append lsp-clients-clangd-args (list (concat "--compile-commands-dir=" found-dir)))
+      lsp-clients-clangd-args)))
+
+
 (defun lsp-clients--clangd-command ()
   "Generate the language server startup command."
   (unless lsp-clients--clangd-default-executable
@@ -190,7 +232,7 @@ This must be set only once after loading the clang client.")
               (lsp-clients-executable-find "xcrun" "--find" "clangd"))))
 
   `(,(or lsp-clients-clangd-executable lsp-clients--clangd-default-executable "clangd")
-    ,@lsp-clients-clangd-args))
+    ,@(lsp-clangd-update-args-with-compile-commands)))
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -208,9 +208,10 @@ that contains compile_commands.json.
 
 Then return either the default args or the modified args"
   (if (-first
-       ;; lsp-client-clangd-args might already lists a "--compile-commands-dir=build/" and will fail if we pass another one
+       ;; lsp-client-clangd-args might already lists a
+       ;; "--compile-commands-dir=build/" and will fail if we pass another one
        ;; trust the user - if it's already set, don't mess around
-       (lambda (arg) (string-prefix-p "--compile-commands-dir" arg))
+       (lambda (arg) (string-match-p "--compile-commands-dir" arg))
        lsp-clients-clangd-args)
       ((lsp--info "compilation_commands_dir set in clangd-args")
        lsp-clients-clangd-args)
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -10,6 +10,7 @@
   * Added ~lsp-modeline-workspace-status-mode~ and option to disable it via ~lsp-modeline-workspace-status-enable~.
   * Automatically download [[https://github.com/eclipse/lemminx][XML language server Lemminx]]
   * Add Vala support.
+  * Enable clangd to find compile_commands.json out of the box before starting
 ** Release 7.0.1
   * Introduced ~lsp-diagnostics-mode~.
   * Safe renamed ~lsp-flycheck-default-level~ -> ~lsp-diagnostics-flycheck-default-level~


petr-tik:
@yyoncho as discussed [here](https://github.com/emacs-lsp/lsp-mode/issues/2278#issuecomment-714723749), feel free to mark this PR as WIP or wait for `lsp-workspace-root` to support this workflow, wanted to get your thoughts as it's my first real elisp contribution.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
   :type '(repeat string))
 
+(defun lsp-clangd-find-compile-commands-dir ()
+  "Return the full directory path (or nil) that contain a compile_commands.json.
+
+If multiple compile_commands.json files are
+found in the project directory tree - ask the user to choose."
+  ;; TODO lsp-workspace-root won't find anything at the start of the workspace
+  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "compile_commands.json")))
+    (pcase (length candidate-dirs)
```
yyoncho, line 188: nit: Refer to lsp-shutdown-workspace on how to pcase a list.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+          nil)
+      ;; if you find 1 only - assume it's correct and strip the dirname from the full path
+      ;; happy path
+      (`1 (file-name-directory (car candidate-dirs)))
+      ;; if you have several - ask the user to choose with lsp--completing-read
+      ;; TODO if Ctrl-G from inside this completing read still falls to default lsp-clients-clangd-args
+      (_ (file-name-directory (lsp--completing-read
+           "Found compile_commands.json in several directories, which one do you want to use?"
+           candidate-dirs
+           (lambda (candidate) candidate)))))))
```
yyoncho, line 200: nit: identity

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+          ;; TODO can I lsp--warn-user-in-message-and-log at the same time
+      (`0 (lsp--warn "Cannot find compile_commands.json - intellisense might not work")
+          nil)
+      ;; if you find 1 only - assume it's correct and strip the dirname from the full path
+      ;; happy path
+      (`1 (file-name-directory (car candidate-dirs)))
+      ;; if you have several - ask the user to choose with lsp--completing-read
+      ;; TODO if Ctrl-G from inside this completing read still falls to default lsp-clients-clangd-args
+      (_ (file-name-directory (lsp--completing-read
+           "Found compile_commands.json in several directories, which one do you want to use?"
```
yyoncho, line 198: Nit: I suggest mentioning that you may use lsp-clients-clangd-args to avoid being asked each time you open that project.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+(defun lsp-clangd-find-compile-commands-dir ()
+  "Return the full directory path (or nil) that contain a compile_commands.json.
+
+If multiple compile_commands.json files are
+found in the project directory tree - ask the user to choose."
+  ;; TODO lsp-workspace-root won't find anything at the start of the workspace
+  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "compile_commands.json")))
+    (pcase (length candidate-dirs)
+          ;; TODO can I lsp--warn-user-in-message-and-log at the same time
+      (`0 (lsp--warn "Cannot find compile_commands.json - intellisense might not work")
```
yyoncho, line 190: using lsp--warn is a bit more for me since there are projects that may work without compile_commands. I suggest using lsp-log which won't be visible by the user.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+that contains compile_commands.json.
+
+Then return either the default args or the modified args"
+  (if (-first
+       ;; lsp-client-clangd-args might already lists a
+       ;; "--compile-commands-dir=build/" and will fail if we pass another one
+       ;; trust the user - if it's already set, don't mess around
+       (lambda (arg) (string-match-p "--compile-commands-dir" arg))
+       lsp-clients-clangd-args)
+      ((lsp--info "compilation_commands_dir set in clangd-args")
```
yyoncho, line 216: missing progn?

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+       lsp-clients-clangd-args)
+    (if-let (found-dir (lsp-clangd-find-compile-commands-dir))
+        ;; TODO expensive to create a list from 1 element - how to append atom to list
```
yyoncho, line 219: the alternative is to use `cons` and append in the beginning. Not sure this matter.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
   :risky t
   :type '(repeat string))
 
+(defun lsp-clangd-find-compile-commands-dir ()
+  "Return the full directory path (or nil) that contain a compile_commands.json.
+
+If multiple compile_commands.json files are
+found in the project directory tree - ask the user to choose."
+  ;; TODO lsp-workspace-root won't find anything at the start of the workspace
+  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "compile_commands.json")))
```
yyoncho, line 187: Does it make sense to look at only one or 2 levels? If not, what will be the perf for huge code-bases?

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+    (pcase (length candidate-dirs)
+          ;; TODO can I lsp--warn-user-in-message-and-log at the same time
+      (`0 (lsp--warn "Cannot find compile_commands.json - intellisense might not work")
+          nil)
+      ;; if you find 1 only - assume it's correct and strip the dirname from the full path
+      ;; happy path
+      (`1 (file-name-directory (car candidate-dirs)))
+      ;; if you have several - ask the user to choose with lsp--completing-read
+      ;; TODO if Ctrl-G from inside this completing read still falls to default lsp-clients-clangd-args
+      (_ (file-name-directory (lsp--completing-read
```
yyoncho, line 197: does it make sense to prefer compile commands in the root of the project?

yyoncho (commented):
Looks good, some nits, suggestions and questions:

clients/lsp-clangd.el
yyoncho: Actually, you can use directly `completing-read`

clients/lsp-clangd.el
petr-tik: Yes, I haven't measured perf yet (don't have access to such large codebases), but I thought paying the cost as a one-off should be fine? After the server is instantiated we don't need to run it again, right?

clients/lsp-clangd.el
petr-tik: Is completing-read preferred over its lsp counterpart?

yyoncho: lsp version is when you want to apply a function over the item to be displayed. In this case, you do not apply such a function(the function is identity).

clients/lsp-clangd.el
petr-tik: Is explicit UX not better? I have yet to see a big cmake-based cpp project that works out of the box.

clients/lsp-clangd.el
yyoncho: If emacs is stuck for 10 seconds this is not fine. I would say anything more than 1 second will result in an issue report, especially considering that this is a regression compared to the case in which you have compile_commands in the root. If we decide to keep it as it is, then I would prefer to have that feature off by default.

clients/lsp-clangd.el
yyoncho: My concern is that a beginner opening a standalone cpp file will see the error and he/she will be confused/misled.

yyoncho: The alternative is to change it to info and change the wording to something like: `Starting without compile_commands.json` without implying that there might be an error.

clients/lsp-clangd.el
nbfalcon:
```suggestion
  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "\\`compile_commands.json\\'")))
```
we don't want to match foo_compile_commands.json5. The \\` forces matching at the beginning (the regex is for basenames) and the \\' at the end (regex anchors).

clients/lsp-clangd.el
```diff
@@ -208,9 +208,10 @@ that contains compile_commands.json.
 
 Then return either the default args or the modified args"
   (if (-first
-       ;; lsp-client-clangd-args might already lists a "--compile-commands-dir=build/" and will fail if we pass another one
+       ;; lsp-client-clangd-args might already lists a
+       ;; "--compile-commands-dir=build/" and will fail if we pass another one
        ;; trust the user - if it's already set, don't mess around
-       (lambda (arg) (string-prefix-p "--compile-commands-dir" arg))
+       (lambda (arg) (string-match-p "--compile-commands-dir" arg))
```
nbfalcon, line 214: We could just use `cl-find` for that entire thing. (if `cl-find` "--compile-commands-dir"). We won't even lose functionality - `cl-find` works for vectors, too, unlike `-first`.

clients/lsp-clangd.el
petr-tik: > My concern is that a beginner opening a standalone cpp file will see the error and he/she will be confused/misled.

Fair point. What about something like

> Starting without compile_commands.json - some features might not work

clients/lsp-clangd.el
petr-tik: > If emacs is stuck for 10 seconds this is not fine. I would say anything more than 1 second will result in an issue report.

you raise a point that only someone with your maintainer experience would know - not my intention that people flood you with bug reports, so I will happily comply. What's the best way to make this an optional feature that is off by default?

petr-tik: alternatively, is there a way to limit directory-files-recursively to go only X levels down from root - i would set it to 2 or 3?

clients/lsp-clangd.el
yyoncho: Thinking a bit more about that seems like it will be better to have that message visible to the user. There are a lot of users who were confused by clangd not being able to handle their project due to missing compile_commands.json. Lets have it like that:

> Starting without compile_commands.json(without this clagnd might not be able to handle complex project).

(or some alternative wording with the same structure).

petr-tik:
@nbfalcon @yyoncho thanks for such a timely and thorough code review! I will try to address it properly over the weekend - you guys rock. I think i might be able to dedicate some time to clangd and rust-analyzer (work and play respectively) clients, so if you think my contributions are useful - i would like to continue

clients/lsp-clangd.el
yyoncho: So, as per your report, typically we are interested in the root and in the build dir, right? If this is true, I would change this method to:

1. Check if compile_commands.json is in root dir -> return it
2. Get dirs in the root dir and check if any of them contain compile_commands and collect them -> handle matches the way you do now.

I like the fact that we will show info when we cannot find the compile_commands.json

yyoncho: There is one more option - not sure if practical(I do not have CPP experience) - can we have a list of predefined locations, e. g. {root}/compile_commands.json, {root}/build/compile_commands.json? Does this make any sense?

petr-tik: i don't think we need to set a flag if it's in the root dir - that's where clangd looks by default. 

I have seen build/ directories with subdirectories (release, debug) that have their own compile_commands.json, hence my suggestion for max-depth 3

> I like the fact that we will show info when we cannot find the compile_commands.json

That's my primary UX objective with this, therefore I wanted to make it a warning not an info message

yyoncho:
> I think i might be able to dedicate some time to clangd and rust-analyzer (work and play respectively) clients, so if you think my contributions are useful

Feel free to pick whatever task you want and ask for help/clarification/etc. We try to make contributions to the project as pleasant as possible.

petr-tik:
> We try to make contributions to the project as pleasant as possible.

You *succeed* in making contributions as pleasant as possible. 

Is it worth tagging issues with relevant client tags? eg.
clangd
rust-analyzer
pyls
microsoft-python-server


etc

clients/lsp-clangd.el
nbfalcon: Don't forget that clangd also supports [compile_flags.txt](https://releases.llvm.org/8.0.0/tools/clang/tools/extra/docs/clangd/Installation.html#compile-flags-txt).

petr-tik: > Don't forget that clangd also supports [compile_flags.txt](https://releases.llvm.org/8.0.0/tools/clang/tools/extra/docs/clangd/Installation.html#compile-flags-txt).

Thanks for the tip-off, however, it doesn't look like --compile-commands-dir will recognise compile_flags.txt.

petr-tik:
I rebased onto upstream/master, so my CI should run now. 

I was thinking about adding a small cpp project with a CMakeLists file to run integration tests similar to what python has. 

I don't see how useful it will be here but might be useful in other features.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
...
+If --compile-commands-dir is already set - return it.
+Else run a potentially interactive search for the directory
+that contains compile_commands.json.
+
+Then return either the default args or the modified args"
+  (if
+       ;; lsp-client-clangd-args might already lists a
+       ;; "--compile-commands-dir=build/" and will fail if we pass another one
+       ;; trust the user - if it's already set, don't mess around
+      (cl-find "--compile-commands-dir" lsp-clients-clangd-args)
```
kiennq, line 214: nit: I would recommend using `-first` or `seq-find` then `cl-find` in this order of availability as it appear in this project.

clients/lsp-clangd.el
```diff
@@ -178,6 +178,49 @@ This must be set only once after loading the clang client.")
   :risky t
   :type '(repeat string))
 
+(defun lsp-clangd-find-compile-commands-dir ()
+  "Return the full directory path (or nil) that contain a compile_commands.json.
+
+If multiple compile_commands.json files are
+found in the project directory tree - ask the user to choose."
+  ;; TODO lsp-workspace-root won't find anything at the start of the workspace
+  (let ((candidate-dirs (directory-files-recursively (lsp-workspace-root) "\\`compile_commands.json\\'")))
```
kiennq, line 187: This is expensive looking up, especially in large project and HDD, which can potentially freeze Emacs.
Can we change the heuristic here?
- Looking at the root workspace only
- Or looking using `locate-dominating-file` from current open file?

Alternatively, can we run these command in background by calling it from shell (`start-file-process-shell-command`) and parse the returns or using `[async.el](https://github.com/jwiegley/emacs-async)`?
So that user will not be blocked when open a `.cpp` file

clients/lsp-clangd.el
petr-tik: nbfalcon suggested I replace -first with cl-find. I don't mind either - please let me know what works best

clients/lsp-clangd.el
petr-tik: > Looking at the root workspace only

clangd looks up compile_commands.json in the project root by default. This auto-configure is precisely to improve on this naive heuristic for out of source cmake builds, where compile_commands.json will be in another directory. 

You and yyoncho have raised very valid points and thanks for the suggestions - I will try to do my best with minimising the runtime cost.

kiennq:
Btw, did you take a look at `initializationOptions.compilationDatabasePath` extension from clangd?
https://clangd.llvm.org/extensions.html

petr-tik:
Coming back to this PR after a bit of a delay. 

> Btw, did you take a look at `initializationOptions.compilationDatabasePath` extension from clangd?
> https://clangd.llvm.org/extensions.html

Yes, I didn't write all my reasons not to include it down, but from memory some of them were:
keeping compilation database path in the lsp-clients-clangd-args enables setting it through .dir-locals.el
cmdline args will show up in lsp-log and make it easier for people to file informative bug reports about clangd misbehaving
The client-args variable already existed, whereas nothing like that existed for init options for clangd

clients/lsp-clangd.el
petr-tik: my heuristic would be to start by most recently mtime. If people build out of source, compile_commands.json will be generated in the cmake `build/` directory and some may copy it to project-root. However, if compile_commands.json changes and the user forgets to copy the new file to project-root again - the UX will be suboptimal. Symlinks work sometimes, but not in docker on WSL

kalj:
@petr-tik How is this going? I too am looking forward to this :+1:

petr-tik:
> @petr-tik How is this going? I too am looking forward to this +1

what a trip down memory lane! I should have done something explicit with this PR though.

@kiennq had a good point about this PR performing a potentially expensive check.  
https://github.com/emacs-lsp/lsp-mode/pull/2303/files#r521761801

since then clangd added the logic to automatically look up $PROJECT_ROOT/build/ 
https://clangd.llvm.org/installation.html#compile_commandsjson

and enabled configuring clangd (per project or system-wide) with the name of build directory
https://clangd.llvm.org/config#compilationdatabase

I would love lsp-mode or clangd to work better out of the box, but I feel that imposing a performance penalty on startup for all users is too high a cost for this. 

Maybe lsp-clangd can send a warning/info message to the user if clangd fails to find compile_commands.json and refer to documentation to help the user configure their workspace. 

Thoughts?

yyoncho:
> Maybe lsp-clangd can send a warning/info message to the user if clangd fails to find compile_commands.json and refer to documentation to help the user configure their workspace.
> 
> Thoughts?

That warning should be implemented on the server because generally lsp-mode is not aware of that file. But also, note that for hello world apps it will be confusing and in the end, it will be confusing for the beginners...

sambrightman:
I feel this PR still has some legs. Whilst clangd now checks build/compile_commands.json there remain quite a few variations (e.g. build_xyz/, build/releasemode/, build/generatortype, out/). It feels like restricting the search to a few prefixes and 2 or 3 levels of depth (both configurable?) would cover a very high percentage of cases not currently covered whilst also greatly reducing the initialisation cost.

yyoncho:
> I feel this PR still has some legs. Whilst clangd now checks `build/compile_commands.json` there remain quite a few variations (e.g. `build_xyz/`, `build/releasemode/`, `build/generatortype`, `out/`). It feels like restricting the search to a few prefixes and 2 or 3 levels of depth (both configurable?) would cover a very high percentage of cases not currently covered whilst also greatly reducing the initialisation cost.

@sambrightman we can go in this direction. I am closing this PR for now. If someone is interested can go with @sambrightman suggestion.

PR #2306, opened 2020-11-06
`lsp-defun' is the same as `-defun'

nbfalcon:
Depends on either magnars/dash.el#347 or magnars/dash.el#336.

# TODO

- [ ] Bump dash.el dependency version


yyoncho:
I am performing cleanup, PR will be reopened/resubmitted when it is ready.

PR #2301, opened 2020-11-05
`lsp-imenu': show signatures

nbfalcon:
What I already did for lsp-treemacs, now for lsp-imenu. Is enabling imenu details by default sensible?
```
(defcustom lsp-imenu-detailed-outline t
```

PR #2307, opened 2020-11-06
Refactor/centralize document symbol rendering

nbfalcon:
Abstract away `&DocumentSymbol` rendering

With microsoft/language-server-protocol#1130, the :detail? of
DocumentSymbols might become more complex, causing lots of breakage. As
such, abstracting away `&DocumentSymbol` rendering is needed.

This needs to be a public API, as `lsp-treemacs` also renders
DocumentSymbols. Bump the version because of that.

I'll do the corresponding `lsp-treemacs` PR soon.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,4 +1,8 @@
 * Changelog
+** Release 7.2
+  * Add a new ~lsp-render-symbol~ function, which can be used to render a
+    ~DocumentSymbol~. It will take any future improvements to ~DocumentSymbol~
+    into account.
 ** Release 7.1
   * Safe renamed ~lsp-diagnose~ to ~lsp-doctor~.
   * Add ~lsp-modeline-code-actions-segments~ for a better customization.
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,8 +1,4 @@
 * Changelog
-** Release 7.2
-  * Add a new ~lsp-render-symbol~ function, which can be used to render a
-    ~DocumentSymbol~. It will take any future improvements to ~DocumentSymbol~
-    into account.
 ** Release 7.1
   * Safe renamed ~lsp-diagnose~ to ~lsp-doctor~.
   * Add ~lsp-modeline-code-actions-segments~ for a better customization.
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6212,7 +6212,7 @@ Return a cons cell (full-name . start-point)."
             name)
           start-point)))
 
-(lsp-defun lsp--symbol-to-hierarchical-imenu-elem ((sym &as &DocumentSymbol :name :detail? :children?))
+(lsp-defun lsp--symbol-to-hierarchical-imenu-elem ((sym &as &DocumentSymbol :children?))
   "Convert SYM to hierarchical imenu elements.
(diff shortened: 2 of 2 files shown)


yyoncho:
Can you rebase properly?

nbfalcon:
@yyoncho I will do that when my proposal gets accepted. However, this patch doesn't work against master, due to making use of `lsp-imenu-detailed-outline`, which my other patch adds.

nbfalcon:
Or were you referring to the merge conflicts? I can fix them.

yyoncho:
Which other PR?

nbfalcon:
#2301.

nbfalcon:
Oh, it's merged. Yeah, I'll rebase soon. Right now I am busy with getting `-defun` done.

nbfalcon:
I have rebased it against master.

lsp-mode.el
```diff
@@ -5,7 +5,7 @@
 ;; Author: Vibhav Pant, Fangrui Song, Ivan Yonchovski
 ;; Keywords: languages
 ;; Package-Requires: ((emacs "26.1") (dash "2.14.1") (dash-functional "2.14.1") (f "0.20.0") (ht "2.0") (spinner "1.7.3") (markdown-mode "2.3") (lv "0"))
-;; Version: 7.1.0
```
yyoncho, line 8: 2 options:

1. Revert this as melpa is pretty much disregarding this version and you can skip that step.
2. Do proper release and update the changelog and release the other components in emacs-lsp as well.

nbfalcon: 1. Is changing the Version here too bad though? I think it is good style to be consistent.
2. I'll make a changelog entry. What other components are you referring to, though? I have made a PR for `lsp-treemacs`.

yyoncho: > 1. Is changing the Version here too bad though? I think it is good style to be consistent.

Indeed - when you release one component you have to release the other. If you release 7.1 some of the other packages(lsp-java, lsp-metals, etc) has to be release as well.

> I'll make a changelog entry.

You have to make changelog entry for all changes since last release.

nbfalcon: But why release all others? `lsp-java` has absolutely nothing to do with it (I haven't touched it for this PR). Also, don't you think a release for this small function is overkill? Perhaps we can release this together with the one to remove `lsp-execute-command`.

yyoncho: When you release lsp-mode it will be published in melpa stable where we have an old version of other components and there is a potential for unmatching components because during 7.1 release we have performed breaking changes(potentially).

Also, this is not the way to do release - first, you have to tag a version and then bump the version otherwise melpa stable won't have 7.1 version.


> Also, don't you think a release for this small function is overkill? 

I am not saying that you should do a release.

CHANGELOG.org
```diff
@@ -1,4 +1,8 @@
 * Changelog
+** Release 7.2
```
yyoncho, line 2: that should be under 7.1

lsp-mode.el
nbfalcon: I'll undo the release. I think its overkill here.

yyoncho (approved):

PR #2308, opened 2020-11-06
Refactor: decouple `lsp--get-document-symbols'

nbfalcon:
That function should not depend on imenu, and not just call
`lsp--imenu-refresh` out of nowhere. Take a callback argument instead,
which implies async. This way, the `imenu` advice could be removed.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5515,39 +5515,33 @@ A reference is highlighted only if it is visible in a window."
                                       line
                                       character)))
 
-(defun lsp--get-document-symbols ()
+(defun lsp--get-document-symbols (&optional callback)
   "Get document symbols.
 
 If the buffer has not been modified since symbols were last
 retrieved, simply return the latest result.
 
-Else, if the request was initiated by Imenu updating its menu-bar
-entry, perform it asynchronously; i.e., give Imenu the latest
-result and then force a refresh when a new one is available.
+If CALLBACK is specified, call it when the document's symbols
+have been fetched. It implies
+`lsp--document-symbols-request-async'.
 
-Else (e.g., due to interactive use of `imenu' or `xref'),
-perform the request synchronously."
+Else (e.g., due to interactive use of `imenu' or `xref'), perform
+the request synchronously."
   (if (= (buffer-chars-modified-tick) lsp--document-symbols-tick)
       lsp--document-symbols
     (let ((method "textDocument/documentSymbol")
           (params `(:textDocument ,(lsp--text-document-identifier)))
           (tick (buffer-chars-modified-tick)))
-      (if (not lsp--document-symbols-request-async)
-          (prog1
-              (setq lsp--document-symbols (lsp-request method params))
-            (setq lsp--document-symbols-tick tick))
-        (lsp-request-async method params
-                           (lambda (document-symbols)
-                             (setq lsp--document-symbols document-symbols
-                                   lsp--document-symbols-tick tick)
-                             (lsp--imenu-refresh))
-                           :mode 'alive)
+      (if (or callback lsp--document-symbols-request-async)
+          (lsp-request-async method params
+                             (lambda (document-symbols)
+                               (setq lsp--document-symbols document-symbols
+                                     lsp--document-symbols-tick tick)
+                               (when callback (funcall callback)))
+                             :mode 'alive)
+        (setq lsp--document-symbols (lsp-request method params)
+              lsp--document-symbols-tick tick)
         lsp--document-symbols))))
-
-(advice-add 'imenu-update-menubar :around
-            (lambda (oldfun &rest r)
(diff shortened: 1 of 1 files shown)


yyoncho:
For some reason the commens are misplaced. I will put them in the right place.

lsp-mode.el
```diff
@@ -5515,40 +5515,34 @@ A reference is highlighted only if it is visible in a window."
...
+                             (lambda (document-symbols)
+                               (setq lsp--document-symbols document-symbols
+                                     lsp--document-symbols-tick tick)
+                               (when callback (funcall callback)))
+                             :mode 'alive)
+        (setq lsp--document-symbols (lsp-request method params)
+              lsp--document-symbols-tick tick)
         lsp--document-symbols))))
 
-(advice-add 'imenu-update-menubar :around
```
yyoncho, line 5547: Please avoid doing changes like that - you cannot go and just delete a code like that. This will cause the imenu to be loaded synchronously and emacs will block until the server has started which may be minutes.

nbfalcon:
I think I might have some logic-issue in there.

lsp-mode.el
```diff
@@ -5515,40 +5515,34 @@ A reference is highlighted only if it is visible in a window."
...
-                             (lsp--imenu-refresh))
-                           :mode 'alive)
+      (if (or callback lsp--document-symbols-request-async)
+          (lsp-request-async method params
+                             (lambda (document-symbols)
+                               (setq lsp--document-symbols document-symbols
+                                     lsp--document-symbols-tick tick)
+                               (when callback (funcall callback)))
+                             :mode 'alive)
+        (setq lsp--document-symbols (lsp-request method params)
```
yyoncho, line 5542: this is regression - the previous code was returning symbols, the new one is returning tick.

lsp-mode.el
nbfalcon: This won't cause issues. Note that there now is a callback parameter, which is specified for `lsp--imenu-create-index`. It implies asynchronous operation. The callback is `lsp--imenu-refresh` for `lsp--imenu-create-index`.

lsp-mode.el
nbfalcon: You are right - for some reason I wasn't using lsp-imenu when testing. I fixed that. `lsp-imenu` works for me.

lsp-mode.el
yyoncho: No - you are introducing a regression. lsp--imenu-create-index function is called in 2 contexts - one is async and one is sync.  It is not calling lsp--imenu-refresh only if it is called from the menubar, so from the callers perspective there is no difference.

yyoncho: > . This will cause the imenu to be loaded synchronously and emacs will block until the server has started which may be minutes.

(The issue is actually you will break the sync path, not the async one).

nbfalcon:
CALLBACK no longer implies async, meaning that `imenu` should work again.

lsp-mode.el
```diff
@@ -5515,39 +5515,33 @@ A reference is highlighted only if it is visible in a window."
...
-              (let ((lsp--document-symbols-request-async t))
-                (apply oldfun r))))
+      (if lsp--document-symbols-request-async
+          (lsp-request-async method params
+                             (lambda (document-symbols)
+                               (setq lsp--document-symbols document-symbols
+                                     lsp--document-symbols-tick tick)
+                               (when callback (funcall callback)))
+                             :mode 'alive)
+        (setq lsp--document-symbols (lsp-request method params)
```
yyoncho, line 5542: Please revert irrelevant changes in this block.

lsp-mode.el
```diff
@@ -5515,39 +5515,33 @@ A reference is highlighted only if it is visible in a window."
...
-            (setq lsp--document-symbols-tick tick))
-        (lsp-request-async method params
-                           (lambda (document-symbols)
-                             (setq lsp--document-symbols document-symbols
-                                   lsp--document-symbols-tick tick)
-                             (lsp--imenu-refresh))
-                           :mode 'alive)
-        lsp--document-symbols))))
-
-(advice-add 'imenu-update-menubar :around
```
yyoncho, line 5547: ditto, this block was moved for no reason.

yyoncho (commented):
Please explain what are the benefits from this PR - moving `lsp--imenu-refresh` hack from `lsp--get-document-symbols` to `lsp--imenu-create-index`?

nbfalcon:
IMHO it makes the code a bit cleaner, since `lsp--get-document-symbols` is called from various places, and as such shouldn't have any imenu-specific code in it. This way, all imenu-related code is together.

nbfalcon:
Actually, I do agree that this patch might be more effort than it is worth. Feel free to close this.

yyoncho:
ok, closing.

Issue #2305, opened 2020-11-06
`onTypeFormatting` causing `$/cancelRequest`


AndreasMatthias:
There seems to be an issue with `onTypeFormatting`. The communication goes like this:

1. Sending notification `textDocument/didChange`.
2. Sending request `textDocument/onTypeFormatting - (2)`.
3. Received response `textDocument/onTypeFormatting - (2)`.
4. Sending notification `$/cancelRequest`.
5. Sending notification `textDocument/didChange`.

The modification proposed by the server in (3) is actually applied to the buffer and confirmed in (5), which is correct. Nevertheless lsp-mode tries to cancel the request in (4), which is wrong. From here on lsp-mode is sending a `$/cancalRequest` for `id:2` after each and every keystroke.

To reproduce this issue I've written a small client `client.el` and a small server `server.py`. See below.

The server is written in python and requires the `pygls` package, which can be installed with `pip`.

To run the test do the following:

* emacs -l client.el test.txt
* M-x foobar-mode
* type some text and then `)`

'`)`' which will trigger `onTypeFormatting`, and the server will respond by inserting `xxx` at the beginning of the line.

Here is the source from `client.el` and `server.py`:

**client.el**
```
(require 'lsp-mode)

(define-derived-mode foobar-mode prog-mode "Foobar")
(add-hook 'foobar-mode-hook #'lsp)
(add-to-list 'lsp-language-id-configuration '(foobar-mode . "foobar"))

(setq lsp-log-io t)
(setq lsp-log-max t)

(eval-when-compile
  (setq server-name (concat (file-name-directory load-file-name) "server.py")))
  
(lsp-register-client
 (make-lsp-client
  :new-connection (lsp-stdio-connection `("python3" ,server-name))
  :major-modes '(foobar-mode)
  :server-id 'foobar-ls
  :language-id "foobar"))

(provide 'foobar)
```

**server.py**
```
#!/usr/bin/env python3

from pygls.features import (FORMATTING, ON_TYPE_FORMATTING)

from pygls.server import LanguageServer

from pygls.types import (Position, Range, TextEdit,
                         DocumentFormattingParams,
                         DocumentOnTypeFormattingParams)

modls = LanguageServer()

@modls.feature(FORMATTING)
def formatting (params: DocumentFormattingParams):
    return []


@modls.feature(ON_TYPE_FORMATTING,
               first_trigger_character = r')')
def onTypeFormatting (params: DocumentOnTypeFormattingParams):
[...]

PR #2309, opened 2020-11-06
Do not use :cancel-handler

yyoncho:
- Fixes https://github.com/emacs-lsp/lsp-mode/issues/2305

The issue isthat when using :cancel-handler we leak cleanup handlers. Putting a
log in lsp--cancel-request proves that. Currently the cancel-handler does not
kick due to tick mode so it is fine to switch to that code

----

#


jcs090218 (approved):

yyoncho:
Fixed issue reported in gitter by @AndreasMatthias .

PR #2310, opened 2020-11-06
Centralize SymbolInformation rendering

nbfalcon:
In various different parts of `lsp`, where SymbolInformation has to be
rendered (lsp-ivy, lsp-treemacs, lsp-imenu), only the :name is shown as
a string. This is insufficient, as other information is then ignore
displayed (currently only :deprecated?).

Add a function, `lsp-render-symbol-information`, which behaves similar
to `lsp-render-symbol`. Use it in `lsp--imenu-create-non-hierarchical`
index.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6205,7 +6205,7 @@ the signature)."
 Handle :deprecated?. If SEPARATOR is non-nil, the
 symbol's (optional) parent, SEPARATOR and the symbol itself are
 concatenated."
-  (when (and separator container-name?)
+  (when (and separator container-name? (not (string-empty-p container-name?)))
     (setq name (concat name separator container-name?)))
   (if deprecated? (propertize name 'face 'lsp-face-semhl-deprecated) name))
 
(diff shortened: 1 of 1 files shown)


yyoncho (approved):
Not part of this review ATM the issue is that you cannot navigate to element if it has children which is a very big limitation when we are rendering DocumentSymbol. The proposal is to build a list(not tree) with the following structure 

> Container1 / Container2 / Node

Willing to address this limitation?

nbfalcon:
I'm sorry, but I don't understand your request. SymbolInformation has no :children field. It's :container-name? field is also probably hacky to rely on, as the LSP Protocol spec says "[...] It [`containerName`] cannot be used to re-infer a hierarchy". Also, hierarchical imenu works like a charm for me (both with `imenu-list` and normal `imenu`).

yyoncho:
I am talking about DocumentSymbol

Here it is - https://emacs.stackexchange.com/questions/61141/why-is-my-imenu-not-showing-function-and-class-definitions/61159?noredirect=1#comment96231_61159

nbfalcon:
That is coincidentally exactly what I am working on right now.

nbfalcon:
The CI failure isn't my fault - "spinner" isn't installable.

PR #2312, opened 2020-11-06
`lsp-render-symbol': check if :detail? is empty

### LINKED ISSUE #2311: Recent changes in imenu have broken imenu for Java:
image

@nbfalcon

nbfalcon:
Some language servers apparently send an empty detail? field, causing
strange output.

Fixes #2311.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6196,7 +6196,8 @@ information, for example if it doesn't support DocumentSymbols."
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
   (let ((detail (and show-detail? (s-present? detail?)
-                     (propertize (concat " " detail?) 'face 'font-lock-type-face)))
+                     (propertize (concat " " (s-trim-left detail?))
+                                 'face 'lsp-lens-face)))
         (name (if deprecated?
                   (propertize name 'face 'lsp-face-semhl-deprecated) name)))
     (concat name detail)))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6195,7 +6195,8 @@ information, for example if it doesn't support DocumentSymbols."
 Things like line numbers, signatures, ... are considered
 additional information. Often, additional faces are defined that
 inherit from this face by default, like `lsp-signature-face', and
-they may be customized for finer control.")
+they may be customized for finer control."
+  :group 'lsp-faces)
 
 (defface lsp-signature-face '((t :inherit lsp-details-face))
   "Used to display signatures in `imenu', ...."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6190,7 +6190,14 @@ information, for example if it doesn't support DocumentSymbols."
   :group 'lsp-imenu
   :type 'boolean)
 
-(defface lsp-signature-face '((t :height 0.8 :inherit shadow))
+(defface lsp-details-face '((t :height 0.8 :inherit shadow))
+  "Used to display additional information troughout `lsp'.
+Things like line numbers, signatures, ... are considered
+additional information. Often, additional faces are defined that
+inherit from this face by default, like `lsp-signature-face', and
+they may be customized for finer control.")
+
+(defface lsp-signature-face '((t :inherit lsp-details-face))
   "Used to display signatures in `imenu', ...."
   :group 'lsp-faces)
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6196,7 +6196,7 @@ information, for example if it doesn't support DocumentSymbols."
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
   (let ((detail (and show-detail? (s-present? detail?)
-                     (propertize (concat ": " detail?) 'face 'font-lock-type-face)))
+                     (propertize (concat " " detail?) 'face 'font-lock-type-face)))
         (name (if deprecated?
                   (propertize name 'face 'lsp-face-semhl-deprecated) name)))
     (concat name detail)))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6195,11 +6195,11 @@ information, for example if it doesn't support DocumentSymbols."
   "Render INPUT0, an `&DocumentSymbol', to a string.
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
-  (let ((base (or (and show-detail? detail?
-                       (not (string-empty-p detail?)) detail?)
-                  name)))
-    (if deprecated? (propertize base 'face 'lsp-face-semhl-deprecated)
-      base)))
+  (let ((detail (and show-detail? detail? (not (string-empty-p detail?))
+                     (propertize (concat ": " detail?) 'face 'font-lock-type-face)))
+        (name (if deprecated?
+                  (propertize name 'face 'lsp-face-semhl-deprecated) name)))
+    (concat name detail)))
 
 (lsp-defun lsp--symbol-to-imenu-elem ((sym &as &SymbolInformation :name :container-name?))
   "Convert SYM to imenu element.
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6195,7 +6195,7 @@ information, for example if it doesn't support DocumentSymbols."
   "Render INPUT0, an `&DocumentSymbol', to a string.
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
-  (let ((detail (and show-detail? detail? (not (string-empty-p detail?))
+  (let ((detail (and show-detail? (s-present? detail?)
                      (propertize (concat ": " detail?) 'face 'font-lock-type-face)))
         (name (if deprecated?
                   (propertize name 'face 'lsp-face-semhl-deprecated) name)))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6190,6 +6190,10 @@ information, for example if it doesn't support DocumentSymbols."
   :group 'lsp-imenu
   :type 'boolean)
 
+(defface lsp-signature-face '((t :height 0.8 :inherit shadow))
+  "Used to display signatures in `imenu', ...."
+  :group 'lsp-faces)
+
 (lsp-defun lsp-render-symbol ((&DocumentSymbol :name :detail? :deprecated?)
                               show-detail?)
   "Render INPUT0, an `&DocumentSymbol', to a string.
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -6195,7 +6195,9 @@ information, for example if it doesn't support DocumentSymbols."
   "Render INPUT0, an `&DocumentSymbol', to a string.
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
-  (let ((base (or (and show-detail? detail?) detail? name)))
+  (let ((base (or (and show-detail? detail?
```
yyoncho, line 6198: nit: s-present? 

I am yet to test if it works.

yyoncho:
This does not work. The issue is that the whole idea with using the details does not work(or at least it should not be on by default)

```
[Trace - 12:09:01 AM] Received response 'textDocument/documentSymbol - (20)' in 4ms.
Result: [
  {
    "deprecated": null,
    "detail": "",
    "selectionRange": {
      "end": {
        "character": 13,
        "line": 0
      },
      "start": {
        "character": 0,
        "line": 0
      }
    },
    "range": {
      "end": {
        "character": 13,
        "line": 0
      },
      "start": {
        "character": 0,
        "line": 0
      }
    },
    "kind": 4,
    "name": "temp"
  },
  {
    "children": [
      {
        "children": [],
        "deprecated": null,
        "detail": " : void",
        "selectionRange": {
          "end": {
            "character": 19,
            "line": 6
          },
          "start": {
            "character": 14,
            "line": 6
          }
        },
        "range": {
          "end": {
            "character": 3,
            "line": 9
          },
          "start": {
            "character": 2,
            "line": 5
          }
        },
        "kind": 6,
        "name": "testA()"
      },
      {
        "children": [],
        "deprecated": null,
        "detail": " : void",
        "selectionRange": {
          "end": {
            "character": 19,
            "line": 14
          },
          "start": {
            "character": 14,
            "line": 14
          }
        },
        "range": {
          "end": {
            "character": 3,
            "line": 17
          },
          "start": {
            "character": 2,
            "line": 13
          }
        },
        "kind": 6,
        "name": "testB()"
      },
      {
        "children": [],
        "deprecated": null,
        "detail": " : void",
        "selectionRange": {
          "end": {
            "character": 18,
            "line": 19
          },
          "start": {
            "character": 15,
            "line": 19
          }
        },
        "range": {
          "end": {
            "character": 3,
            "line": 21
          },
          "start": {
            "character": 2,
            "line": 19
          }
        },
        "kind": 6,
        "name": "foo()"
      },
      {
        "children": [],
        "deprecated": null,
        "detail": " : void",
        "selectionRange": {
          "end": {
            "character": 19,
            "line": 23
          },
          "start": {
            "character": 15,
            "line": 23
          }
        },
        "range": {
          "end": {
            "character": 3,
            "line": 25
          },
          "start": {
            "character": 2,
            "line": 23
          }
        },
....
```

yyoncho:
image

yyoncho:
Here it is how VScode works in this case:

image

nbfalcon:
Don't merge yet - I'll need to rethink rendering. So apparently JDTLS' :detail? is a signature without the name, while ccls' :detail? is a signature with name...

lsp-mode.el
```diff
@@ -6195,9 +6195,11 @@ information, for example if it doesn't support DocumentSymbols."
   "Render INPUT0, an `&DocumentSymbol', to a string.
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
-  (let ((base (or (and show-detail? detail?) detail? name)))
-    (if deprecated? (propertize base 'face 'lsp-face-semhl-deprecated)
-      base)))
+  (let ((detail (and show-detail? detail? (not (string-empty-p detail?))
+                     (propertize (concat ": " detail?) 'face 'font-lock-type-face)))
```
yyoncho, line 6199: I would use face like `lsp-lens-face` - height 0.8 and  :inherit shadow

nbfalcon: But isn't `font-lock-type-face` more appropriate here, given that :detail? is almost always a type?

lsp-mode.el
```diff
@@ -6195,9 +6195,11 @@ information, for example if it doesn't support DocumentSymbols."
   "Render INPUT0, an `&DocumentSymbol', to a string.
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
-  (let ((base (or (and show-detail? detail?) detail? name)))
-    (if deprecated? (propertize base 'face 'lsp-face-semhl-deprecated)
-      base)))
+  (let ((detail (and show-detail? (s-present? detail?)
+                     (propertize (concat ": " detail?) 'face 'font-lock-type-face)))
```
yyoncho, line 6199: ": " is not needed

nbfalcon: Compare without and with colon:



Are you sure? I think I prefer the colon version, because it more clearly separates signature from name (and VSCode does it too), but I am of course open to different opinions.

lsp-mode.el
yyoncho: Yes, but:

1. Check how it looks like in vscode, the focus is on the name, not on the detail. Thus they are using a smaller font .
2. We already use this on multiple places in the project(you may refer on how many places we use lsp-lens-face in lsp-treemacs). Using that will lead to consistent look.

nbfalcon: Ok, without the colon it also still looks good.

nbfalcon: `lsp-lens-face` also looks cool, but it is a bit too dark I think:

lsp-mode.el
yyoncho: Here it is the JDTLS detail:

>  "detail": " : void"

so adding ":" is not on the table.

nbfalcon: True, that would look awkward. However, we cannot get rid of the double space, as everything else would look strange

 then:

nbfalcon: Perhaps we should fix this upstream?

yyoncho: I guess we have to trim the detail

yyoncho: > Perhaps we should fix this upstream?

upstream does not care as long as it works with vscode.

nbfalcon: Does VSCode also do the trimming?

lsp-mode.el
yyoncho: > `lsp-lens-face` also looks cool, but it is a bit too dark I think:

If you use doom-dark+ which is the emacs port of the vscode theme it will look exactly like the vscode counterpart. In general, treemacs controls are port of vscode one.

nbfalcon: We could also just make the face customizable.

lsp-mode.el
nbfalcon: From looking at the VSCode outline, it doesn't seem to trim, either? But I guess we'll have to fold, as fixing this upstream might be a pain.

lsp-mode.el
```diff
@@ -6195,9 +6195,12 @@ information, for example if it doesn't support DocumentSymbols."
   "Render INPUT0, an `&DocumentSymbol', to a string.
 If SHOW-DETAIL? is set, make use of its `:detail?' field (often
 the signature)."
-  (let ((base (or (and show-detail? detail?) detail? name)))
-    (if deprecated? (propertize base 'face 'lsp-face-semhl-deprecated)
-      base)))
+  (let ((detail (and show-detail? (s-present? detail?)
+                     (propertize (concat " " (s-trim-left detail?))
+                                 'face 'lsp-lens-face)))
```
yyoncho, line 6200: lsp-lens-face is defined in lsp-lens.el which might not be present since lsp-lens.el is loaded only if lenses are loaded. I suggest creating a duplicate of that face, lets call it lsp-details-face in lsp-mode.el. Then, it makes sense to replace lsp-treemacs.el references to lsp-lens-face with lsp-details-face.

lsp-mode.el
yyoncho: Seems like they are not trimming. Feel free to trim/not to trim. IMO both are acceptable.

Looking on imenu-list on your screenshot seems like the detail does not look very well(at least it looks very noisy/unreadable due to using the same font). What do you think about disabling detail for imenu by default?

lsp-mode.el
nbfalcon: We use a new `lsp-signature-face` instead now. This even makes it customizable.

lsp-mode.el
nbfalcon: Actually, I think it looks fine. Here is another screenshot with `lsp-treemacs-symbols`, `imenu` and `imenu-list`:

For some reason, imenu entries get cut off, but that isn't an issue with lsp-mode.

lsp-mode.el
```diff
@@ -6190,14 +6190,21 @@ information, for example if it doesn't support DocumentSymbols."
   :group 'lsp-imenu
   :type 'boolean)
 
+(defface lsp-signature-face '((t :height 0.8 :inherit shadow))
```
yyoncho, line 6193: If you name it lsp-signature-face we won't be able to use it in lsp-treemacs. Why not using the more generic name lsp-details-face as suggested?

lsp-mode.el
nbfalcon: `imenu` might be too dark though.

lsp-mode.el
nbfalcon: Is that really a good idea though? Details might be many different things. A user might want to show signatures with `font-lock-type-face` for instance, and have `lsp-lens-face` for various details. I do agree though that having a details face is a good idea, and that we should add one, making `lsp-signature-face` inherit from it.

yyoncho: base details face sounds good.

nbfalcon: Great! It is now also used in `lsp-lens`. Lots of inheritance for styling is usually a good idea.

lsp-mode.el
```diff
@@ -6190,14 +6190,28 @@ information, for example if it doesn't support DocumentSymbols."
   :group 'lsp-imenu
   :type 'boolean)
 
+(defface lsp-details-face '((t :height 0.8 :inherit shadow))
+  "Used to display additional information troughout `lsp'.
+Things like line numbers, signatures, ... are considered
+additional information. Often, additional faces are defined that
+inherit from this face by default, like `lsp-signature-face', and
+they may be customized for finer control.")
```
yyoncho, line 6198: group is missing.

yyoncho (approved):

lsp-mode.el
nbfalcon: Also, should we perhaps make `lsp-signature-face` inherit from `font-lock-type-face`, with a height of :0.8?

yyoncho: VScode has a huge team of user experience experts and designers and for things like color selection, we follow their decisions.

nbfalcon:
Great! Only a few more PRs to go :)

Issue #2246, opened 2020-10-12
LSP mode with imenu does not show Python classes/functions properly


dmalyuta:
(The original question was posted on stack exchange: https://emacs.stackexchange.com/q/61141/13661)

I am using Emacs 27.1 with the latest versions of all packages. My LSP-related setup is:

```
(use-package lsp-pyright
  :ensure t
  :hook (python-mode . (lambda ()
                         (require 'lsp-pyright)
                         (lsp))))  ; or lsp-deferred

(use-package lsp-mode
  :ensure t
  :init
  (setq lsp-keymap-prefix "C-c l")
  :config
  (require 'lsp-mode)
  (add-hook 'python-mode-hook #'lsp)
  (add-hook 'c-mode-common-hook #'lsp)
  (require 'company-capf)
  (setq lsp-prefer-capf t)
  (push 'company-capf company-backends)
  (push 'company-c-headers company-backends)
  (setq lsp-idle-delay 0.500)
  (setq lsp-diagnostic-package :none)
  (add-hook 'lsp-mode-hook (lambda ()
			     (setq company-minimum-prefix-length 1
				   company-idle-delay 0.0)))
  (setq lsp-enable-semantic-highlighting t)
  (setq lsp-enable-snippet nil)  ;; Enable arguments completion
  (setq lsp-signature-auto-activate nil)
  (setq lsp-progress-via-spinner nil)
  )

(use-package lsp-ui
  :ensure t
  :config
  (add-hook 'lsp-mode-hook
	    (lambda ()
	      (setq lsp-ui-doc-enable nil ;; disable docs
		    ;; lsp-ui-doc-delay 1
   		    ;; lsp-ui-doc-use-childframe t
   		    lsp-ui-doc-position 'bottom
		    lsp-ui-doc-max-height 20
   		    lsp-ui-doc-include-signature t

   		    lsp-ui-sideline-enable t
		    lsp-ui-sideline-delay 0.2
		    lsp-ui-sideline-show-code-actions nil
		    lsp-ui-sideline-show-hover nil

   		    lsp-ui-flycheck-enable t
   		    lsp-ui-flycheck-list-position 'right
   		    lsp-ui-flycheck-live-reporting t

		    lsp-ui-peek-enable nil
   		    ;; lsp-ui-peek-list-width 60
   		    ;; lsp-ui-peek-peek-height 25

		    lsp-ui-imenu-enable nil
		    )
[...]

PR #2313, opened 2020-11-07
`lsp-imenu`: support categorizing by :kind?

nbfalcon:


lsp-mode.el
```diff
@@ -6303,15 +6303,81 @@ representation to point representation."
...
                                                 lsp--collect-lines-and-cols
                                                 lsp--convert-line-col-to-points-batch)))
-    (if (lsp--imenu-hierarchical-p filtered-symbols)
-        (lsp--imenu-create-hierarchical-index filtered-symbols)
-      (lsp--imenu-create-non-hierarchical-index filtered-symbols))))
+    (if (lsp--imenu-hierarchical-p symbols)
+        (lsp--imenu-create-hierarchical-index symbols)
+      (lsp--imenu-create-non-hierarchical-index symbols))))
+
+(defcustom lsp-imenu-symbol-categories
```
nbfalcon, line 6315: Should I `lsp--symbol-kind` here? I wanted to make this at least somewhat customizable (I had in mind that a user could  replace the strings with icons, for example).

nbfalcon:
This time, I didn't change any defaults.

lsp-mode.el
nbfalcon: Also, should we make `lsp--symbol-kind` defcustom?

lsp-mode.el
```diff
@@ -6303,15 +6303,81 @@ representation to point representation."
...
+   "Key"
+   "Null"
+   "Enumerator"
+   "Struct"
+   "Event"
+   "Operator"
+   "Type Parameter"]
+  "Vector mapping LSP SymbolKinds to `imenu' categories."
+  :group 'lsp-imenu
+  :type '(vector string))
```
yyoncho, line 6344: there is no type vector.

lsp-mode.el
yyoncho: Seems both have the same info and it will be better to use the same variable. It can be changed to defcustom.

lsp-mode.el
```diff
@@ -6303,15 +6303,81 @@ representation to point representation."
...
+SymbolInformation."
+  (mapcan
+   (-lambda ((type . symbols))
+     (let ((cat (and type (< type (length lsp-imenu-symbol-categories))
+                     (elt lsp-imenu-symbol-categories type)))
+           (symbols (lsp-imenu-create-uncategorized-index symbols)))
+       ;; If there is no :kind (this is being defensive), or we couldn't look it
+       ;; up, just display the symbols inline, without categories.
+       (if cat (list (cons cat symbols)) symbols)))
+   (sort (seq-group-by (-lambda ((&DocumentSymbol :kind)) kind) symbols)
```
yyoncho, line 6359: nit: use lsp:document-symbol-kind

lsp-mode.el
nbfalcon: I made that one defcustom. However, I believe having a separate alist is the proper way to do it, as it allows pluralized categories (as is imenu convention, apparently).

lsp-mode.el
```diff
@@ -5499,18 +5500,26 @@ A reference is highlighted only if it is visible in a window."
     (19 . "Object")
     (20 . "Key")
     (21 . "Null")
-    (22 . "Enum Member")
+    (22 . "Enumerator")
```
kiennq, line 5503: Since I'm not native speaker, I'm not sure about this change. But I think the meaning of `enumerator` and `enum member` is different.
The [LSP specs](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md) is using `EnumMember` so we should probably stick with that.

nbfalcon: This probably depends on the programming language and context, but in C++ and "Enumerator" is an "Enum member": https://en.cppreference.com/w/cpp/language/enum, Unscoped declaration (1): `enum` [name] { *enumerator* = [...] }

yyoncho: If it is language-specific we should stick with the LSP spec term.

nbfalcon: I thought it was a standard term, but the Java spec doesn't use it, preferring "Enum Constant". I'm reverting.

nbfalcon:
Fixes #2246.

lsp-mode.el

lsp-mode.el
nbfalcon: Fixed, because we use `alist` now.

nbfalcon:
I would consider this PR ready.

yyoncho:
When I customize lsp-imenu-index-function from the UI I get an error.

nbfalcon:
Fixed.

yyoncho:
When grouped by kind I would expect all methods to be grouped in one place - in your case the methods are annotated with classes. Is this expected? 

image

nbfalcon:
Yes, I only group top-level symbols. I don't think doing anything else makes sense, as e.g. there would be no way to infer what class a method belongs to otherwise. Imagine a huge Java/C++ file with lots of classes with many common methods, and each of them would show up under "methods". It would be difficult to select one of the methods from some specific class.

yyoncho:
I find the name `lsp-imenu-create-uncategorized-index` confusing, isn't this the server order/grouping?

nbfalcon:
Although it would probably make sense to have a list of symbols that should always be at the top-level: some languages (e.g. `python`) support scoped imports within function definitions. `imenu` would be much more useful if it could display an overview of all imports, no matter where they are defined.

nbfalcon:
@yyoncho yes, `lsp-create-uncategorized-index` returns the symbols as acquired from the server, not grouping them by `type`.

yyoncho:
> @yyoncho yes, `lsp-create-uncategorized-index` returns the symbols as acquired from the server, not grouping them by `type`.

Can you reflect that in the doc of the entry?

nbfalcon:
Another issue is that a top-level symbol isn't shown if it has child-variables: only a classes' members are shown, or a function's arguments, not the class or function itself. This could be fixed, but at the cost of duplicate entries in `imenu-list`.

As to your last request, that is fixed.

yyoncho:
> Another issue is that a top-level symbol isn't shown if it has child-variables: only a classes' members are shown, or a function's arguments, not the class or function itself. This could be fixed, but at the cost of duplicate entries in `imenu-list`.

we should fix that by flattening the hierarchy. AFAICT this is what the default imenu implementation is doing.

image

This is what I see with pyls, is this expected? At least we should have a space between the name/container name.

nbfalcon:
Flattening the hierarchy would make `imenu-list` completely useless, because the whole point of it is to show a hierarchy.

nbfalcon:
This seems like a bug in `pyls`, because it doesn't send the hierarchies correctly. Can you check `lsp-treemacs-symbols`?

yyoncho:
> This seems like a bug in `pyls`, because it doesn't send the hierarchies correctly. Can you check `lsp-treemacs-symbols`?

It works fine in lsp-treemacs-symbols. pyls does not return hierarchical elements.

nbfalcon:
In `lsp-treemacs-symbols`, do you see function names like `test_2/test_foo`, or do you see a foldable `test_2` with `test_foo` as a child? For reference, here is how it looks for me with pyright:


I agree that this PR is not 100% ready yet. I'll work on consolidating my contributions soon.

yyoncho:
Here it is how it imenu looks like when using vanilla python-mode. 

image

yyoncho:
> In `lsp-treemacs-symbols`, do you see function names like `test_2/test_foo`, or do you see a foldable `test_2` with `test_foo` as a child? For reference, here is how it looks for me with pyright:
> 
> 
> I agree that this PR is not 100% ready yet. I'll work on consolidating my contributions soon.


It is foldable, lsp-treemacs is building a tree.

yyoncho:
FTR I don't think that using imenu to squeeze everything that lsp protocol provides. IMHO having some rudimentary imenu integration for the sake of completeness is fine and we should focus on creating fully featured components using help/ivy/treemacs, etc.

nbfalcon:
@yyoncho I have addressed your latest request: all functions/variables etc. can now be grouped, albeit currently with no sorting (do you think that is even necessary?). Here are some screenshots with the new `lsp-imenu-create-categorized-index` (the old is now `lsp-imenu-create-top-level-categorized-index`):



It also works with `SymbolInformation` (Iph, PHP):

nbfalcon:
In addition, each symbol with children now also gets an entry, so we don't have to solve that problem with special `this` children (see screenshot 2): `pyright` adds locals as children, and `do_ko`'s locals and the function itself can be jumped to.

yyoncho:
I will retest.

yyoncho:
There is no function  `lsp-imenu-create-top-levelcategorized-index`.

yyoncho:
Other than that, seems like we are good to be merged. 

In separate PR, are you willing to fix the inability to go to the parent element in hierarchical symbols? I guess we have 2 options: 1. Flatten out
2. Include the parent as a child. I think that python-mode(?) was doing that by default.

nbfalcon:
> There is no function `lsp-imenu-create-top-levelcategorized-index`.

Fixed.

As to you second point, are you referring to non-`lsp-imenu-create-categorized-index` functions, because the former already does that? I am willing to implement that, and I think including the parent as a child would be the best way to go about this.

yyoncho:
> As to you second point, are you referring to non-`lsp-imenu-create-categorized-index` functions, because the former already does that? I

I am referring `lsp-imenu-create-uncategorized-index`. In general, for all 3 options, we should support going to each of the elements.

nbfalcon:
One last question: what would be the best way to add sorting? Simply sort by kind id, or perhaps allow a user-customizable kinds list, so that users could, depending on preference, have functions first, then classes, then methods and then everything else?

Also, do you think `lsp-imenu-create-categorized-index` should sort using `lsp--imenu-symbol-lessp`, with regard to performance?

Aside from these questions, this PR is ready.

yyoncho:
> One last question: what would be the best way to add sorting? Simply sort by kind id, or perhaps allow a user-customizable kinds list, so that users could, depending on preference, have functions first, then classes, then methods and then everything else?

Don't we have lsp-imenu-sort-methods ?

> Also, do you think lsp-imenu-create-categorized-index should sort using lsp--imenu-symbol-lessp, with regard to performance?

Do we have issues with performance?

yyoncho (approved):

nbfalcon:
@yyoncho `lsp-imenu-sort-methods` is for `imenu` symbols, not for categories.

yyoncho:
> @yyoncho `lsp-imenu-sort-methods` is for `imenu` symbols, not for categories.

Got it. Feel free to add sorting on categories if you need that. Otherwise, wait for a feature request.

PR #2317, opened 2020-11-09
`lsp-rename`: improve UI

### LINKED ISSUE #2314: `lsp-rename`: check if supported before prompting for a name
When I start using a new language sever, I don't know what it supports. If because of that I try to rename something, even though the language server doesn't support it, I just end up wasting my time coming up with a new name.

`lsp-rename` should do a capability check + prepareRename before asking the user for a new name. This will improve discoverability, because users aren't punished for not knowing that rename doesn't work.

This will be an easy fix.

nbfalcon:
`lsp-rename` doesn't check whether renaming is supported first. Do that,
throwing an error otherwise.

Fixes #2314.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1191,6 +1191,13 @@ Deprecated. Use `lsp-repeatable-vector' instead. "
   "Display lsp error message with FORMAT with ARGS."
   (message "%s :: %s" (propertize "LSP" 'face 'error) (apply #'format format args)))
 
+(defun lsp--fatal-error (format &rest args)
+  "Show an lsp error message using `error'.
+FORMAT is the `format' format to use, while ARGS provides the
+fields to it."
+  (error "%s :: %s" (propertize "LSP" 'face 'error)
+         (apply #'format format args)))
+
 (defun lsp--eldoc-message (&optional msg)
   "Show MSG in eldoc."
   (setq lsp--eldoc-saved-message msg)
@@ -5654,22 +5661,30 @@ perform the request synchronously."
            (lsp-request "workspace/symbol" `(:query ,pattern))))
 
 (defun lsp--get-symbol-to-rename ()
-  "Get symbol to rename and placeholder at point."
-  (if (let ((rename-provider (or (lsp--capability :renameProvider)
-                                 (-some-> (lsp--registered-capability "textDocument/rename")
-                                   (lsp--registered-capability-options)))))
-        (lsp:rename-options-prepare-provider? rename-provider))
-      (-when-let (response (lsp-request "textDocument/prepareRename"
+  "Get symbol to rename and placeholder at point.
+Returns a cons (SYMBOL-TO-RENAME . PLACEHOLDER)."
+  ;; Is there some connected server that supports renaming?
+  (if-let ((rename-provider
+            (or (lsp--capability :renameProvider)
+                (-some-> (lsp--registered-capability "textDocument/rename")
+                  (lsp--registered-capability-options)))))
+      ;; Do we have prepare rename? Do it in that case. If we fail, don't do a
+      ;; rename, since it's invalid.
+      (if (lsp:rename-options-prepare-provider? rename-provider)
+          (-let* ((prepare (lsp-request "textDocument/prepareRename"
                                         (lsp--text-document-position-params)))
-        (-let* (((start . end) (lsp--range-to-region
-                                (if (lsp-range? response)
-                                    response
-                                  (lsp:prepare-rename-result-range response))))
-                (symbol (buffer-substring-no-properties start end))
-                (placeholder (lsp:prepare-rename-result-placeholder response)))
-          (cons symbol (or placeholder symbol))))
-    (let ((symbol (thing-at-point 'symbol t)))
-      (cons symbol symbol))))
+                  ((start . end) (lsp--range-to-region
+                                  (if (lsp-range? prepare)
+                                      prepare
+                                    (lsp:prepare-rename-result-range prepare))))
+                  (symbol (buffer-substring-no-properties start end))
+                  (placeholder (lsp:prepare-rename-result-placeholder prepare)))
+            (cons symbol (or placeholder symbol)))
+        ;; We don't support prepare-rename, so just yield the symbol at point.
+        (let ((sym (thing-at-point 'symbol)))
+          (cons sym sym)))
+    ;; No connected server supports renaming.
+    (lsp--fatal-error "The connected server(s) doesn't support renaming")))
 
 (defun lsp-rename (newname)
   "Rename the symbol (and all references to it) under point to NEWNAME."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5689,6 +5689,12 @@ language server as the initial input of a new-name prompt."
 Renaming can be done using `lsp-rename'."
   :group 'lsp-faces)
 
+(defface lsp-rename-placeholder-face '((t :inherit font-lock-variable-name-face))
+  "Face used to display the rename placeholder in.
+When calling `lsp-rename' interactively, this will be the face of
+the new name."
+  :group 'lsp-faces)
+
 (defvar lsp--rename-history '()
   "History for `lsp--read-rename'.")
 
@@ -5709,6 +5715,7 @@ relied upon."
           ;; Do the `buffer-substring' first to not include `lsp-face-rename'
           (rename-me (buffer-substring start end))
           (placeholder (or placeholder? rename-me))
+          (placeholder (propertize placeholder 'face 'lsp-rename-placeholder-face))
 
           overlay)
     ;; We need unwind protect, as the user might cancel here, causing the
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5675,9 +5675,11 @@ renaming is supported but cannot be done at point."
 
 (defun lsp-rename (newname)
   "Rename the symbol (and all references to it) under point to NEWNAME."
-  (interactive (list (-when-let ((symbol . placeholder) (lsp--get-symbol-to-rename))
-                       (read-string (format "Rename %s to: " symbol) placeholder nil symbol))))
-  (unless newname (user-error "`lsp-rename' is invalid here"))
+  (interactive
+   (-if-let ((symbol . placeholder) (lsp--get-symbol-to-rename))
+       (list (read-string (format "Rename %s to: " symbol)
+                          placeholder nil symbol))
+     (user-error "`lsp-rename' is invalid here")))
   (when-let ((edits (lsp-request "textDocument/rename"
                                  `( :textDocument ,(lsp--text-document-identifier)
                                     :position ,(lsp--cur-position)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5715,8 +5715,9 @@ relied upon."
     ;; overlay to linger.
     (unwind-protect
         (progn
-          (setq overlay (-doto (make-overlay start end)
-                          (overlay-put 'face 'lsp-face-rename)))
+          (setq overlay (make-overlay start end))
+          (overlay-put overlay 'face 'lsp-face-rename)
+
           (read-string (format "Rename %s to: " rename-me) placeholder
                        'lsp--rename-history))
       (and overlay (delete-overlay overlay)))))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5654,31 +5654,65 @@ perform the request synchronously."
            (lsp-request "workspace/symbol" `(:query ,pattern))))
 
 (defun lsp--get-symbol-to-rename ()
-  "Get symbol to rename and placeholder at point.
-Returns a cons (SYMBOL-TO-RENAME . PLACEHOLDER). Returns nil if
-renaming is supported but cannot be done at point."
+  "Get a symbol to rename and placeholder at point.
+Returns a cons ((START . END) . PLACEHOLDER?), and nil if
+renaming is generally supported but cannot be done at point.
+START and END are the bounds of the identifiers being renamed,
+while PLACEHOLDER?, is either nil or a string suggested by the
+language server as the initial input of a new-name prompt."
   (unless (lsp-feature? "textDocument/rename")
     (error "The connected server(s) doesn't support renaming"))
   (if (lsp-feature? "textDocument/prepareRename")
       (when-let ((response
                   (lsp-request "textDocument/prepareRename"
                                (lsp--text-document-position-params))))
-        (-let* (((start . end) (lsp--range-to-region
-                                (if (lsp-range? response)
-                                    response
-                                  (lsp:prepare-rename-result-range response))))
-                (symbol (buffer-substring start end))
-                (placeholder (lsp:prepare-rename-result-placeholder response)))
-          (cons symbol (or placeholder symbol))))
-    (when-let ((symbol (thing-at-point 'symbol)))
-      (cons symbol symbol))))
+        (let* ((bounds (lsp--range-to-region
+                        (if (lsp-range? response)
+                            response
+                          (lsp:prepare-rename-result-range response))))
+               (placeholder (lsp:prepare-rename-result-placeholder response)))
+          (cons bounds placeholder)))
+    (when-let ((bounds (bounds-of-thing-at-point 'symbol)))
+      (cons bounds nil))))
+
+(defface lsp-face-rename '((t :inherit highlight))
+  "Face used to highlight the identifier being renamed.
+Renaming can be done using `lsp-rename'.")
+
+(defvar lsp--rename-history '()
+  "History for `lsp--read-rename'.")
+
+(defun lsp--read-rename (at-point)
+  "Read a new name for a `lsp-rename' at `point' from the user.
+AT-POINT shall be a structure as returned by
+`lsp--get-symbol-to-rename'.
+
+Returns a string, which should be the new name for the identifier
+at point. If renaming cannot be done at point (as determined from
+AT-POINT), throw a `user-error'.
+
+This function is for use in `lsp-rename' only, and shall not be
+relied upon."
+  (unless at-point
+    (user-error "`lsp-rename' is invalid here"))
+  (-let* ((((start . end) . placeholder?) at-point)
+          ;; Do the `buffer-substring' first to not include `lsp-face-rename'
+          (rename-me (buffer-substring start end))
+          (placeholder (or placeholder? rename-me)))
+    ;; We need unwind protect, as the user might cancel here, causing the
+    ;; overlay to linger.
+    (unwind-protect
+        (progn
+          (-doto (make-overlay start end)
+            (overlay-put 'face 'lsp-face-rename)
+            (overlay-put 'lsp-rename t))
+          (read-string (format "Rename %s to: " rename-me) placeholder
+                       'lsp--rename-history))
+      (lsp--remove-overlays 'lsp-rename))))
 
 (defun lsp-rename (newname)
   "Rename the symbol (and all references to it) under point to NEWNAME."
-  (interactive (list (-when-let ((symbol . placeholder) (lsp--get-symbol-to-rename))
-                       (read-string (format "Rename %s to: " symbol) placeholder nil symbol))))
-  (unless newname
-    (user-error "`lsp-rename' is invalid here"))
+  (interactive (list (lsp--read-rename (lsp--get-symbol-to-rename))))
   (when-let ((edits (lsp-request "textDocument/rename"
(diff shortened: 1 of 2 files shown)


yyoncho:
I am not sure we need anything more than...

```
(unless (lsp-feature? "textDocument/rename")
  (user-error "Rename not..." ))
```

in the `lsp-rename` interactive form.

nbfalcon:
Don't merge yet - there is a bug that I need to fix. As to your comment, if we use that new function, there will be a nice LSP :: message instead. Do you consider that undesirable? If so, I'll revert.

nbfalcon:
Also, the prompt is now a bit nicer, because I don't use buffer-substring-no-properties.

nbfalcon:
- The bug is fixed; rename now works
- I use `lsp-feature?` now, thanks for the tip. It is really much simpler, and probably less fragile.

lsp-mode.el
```diff
@@ -5646,29 +5659,35 @@ perform the request synchronously."
...
-                                  (lsp:prepare-rename-result-range response))))
-                (symbol (buffer-substring-no-properties start end))
-                (placeholder (lsp:prepare-rename-result-placeholder response)))
-          (cons symbol (or placeholder symbol))))
-    (let ((symbol (thing-at-point 'symbol t)))
-      (cons symbol symbol))))
+  "Get symbol to rename and placeholder at point.
+Returns a cons (SYMBOL-TO-RENAME . PLACEHOLDER)."
+  ;; Is there some connected server that supports renaming?
+  (if-let (lsp-feature? "textDocument/rename")
```
yyoncho, line 5665:
```
(-if-let (lsp-feature? "xxx")
    lsp-feature?
  ...)
```

returns "xxx".

Can you use the unless form as suggested?

lsp-mode.el
```diff
@@ -5646,29 +5659,35 @@ perform the request synchronously."
...
+                                    (if (lsp-range? prepare)
+                                        prepare
+                                      (lsp:prepare-rename-result-range prepare))))
+                    (symbol (buffer-substring start end))
+                    (placeholder (lsp:prepare-rename-result-placeholder prepare)))
+              (cons symbol (or placeholder symbol))))
+        ;; We don't support prepare-rename, so just yield the symbol at point.
+        (let ((sym (thing-at-point 'symbol)))
+          (cons sym sym)))
+    ;; No connected server supports renaming.
```
yyoncho, line 5682: can you avoid putting comments like that? This should be evident from the content of the next line.

nbfalcon: I usually don't do that - this one was an accident. Fixed it.

nbfalcon:
Note that the error messages aren't propertized, which may require further thought.

lsp-mode.el
```diff
@@ -5646,29 +5659,32 @@ perform the request synchronously."
...
+                                    prepare
+                                  (lsp:prepare-rename-result-range prepare))))
+                (symbol (buffer-substring start end))
+                (placeholder (lsp:prepare-rename-result-placeholder prepare)))
           (cons symbol (or placeholder symbol))))
-    (let ((symbol (thing-at-point 'symbol t)))
-      (cons symbol symbol))))
+    ;; We don't support prepare-rename, so just yield the symbol at point.
+    (let ((sym (thing-at-point 'symbol)))
+      (cons sym sym))))
```
yyoncho, line 5680: 1. prepareRename does not work with `lsp-feature?`
2. Can you revert the renamed symbols?
3. Can you remove irrelevant comments?

lsp-mode.el
```diff
@@ -1188,6 +1188,19 @@ Deprecated. Use `lsp-repeatable-vector' instead. "
   "Display lsp error message with FORMAT with ARGS."
   (message "%s :: %s" (propertize "LSP" 'face 'error) (apply #'format format args)))
 
+(defun lsp--fatal-error (format &rest args)
```
yyoncho, line 1191: seems like this is redundant: there is lsp--error.

nbfalcon:
Because of the unpropertized error messages, I'd advise you to wait before merging.

lsp-mode.el
```diff
@@ -1188,6 +1188,19 @@ Deprecated. Use `lsp-repeatable-vector' instead. "
...
+  "Show an lsp error message using `error'.
+FORMAT is the `format' format to use, while ARGS provides the
+fields to it."
+  (error "%s :: %s" (propertize "LSP" 'face 'error)
+         (apply #'format format args)))
+
+(defun lsp--user-error (format &rest args)
```
yyoncho, line 1198: I don't see any effect from this propertize - the message is displayed with default font.

lsp-mode.el
yyoncho: Ignore... I missed the fact that it is using error not message. But still, is this function used in this PR? Are we going to replace all error calls with that?

lsp-mode.el
nbfalcon: Yes, that is the issue currently.

lsp-mode.el
nbfalcon: `lsp--error` logs a `message`, which doesn't cause evaluation to abort.

nbfalcon: This function is used by this PR, which is why I added it. However, replacing all error calls might be a bit much. From looking at the code, the `(error (lsp-error ...))` idiom seems common, however it doesn't solve the text-properties issue.

lsp-mode.el
nbfalcon: 1. What should I use instead? The old code?
2, 3. Will do.

yyoncho: The old code should be fine.

nbfalcon: Are you sure that it doesn't though? From stepping trough that with edebug, `lsp-feature?` returned a complex result for "textDocument/prepareRename", while consistently returning nil if I evaluate it somewhere without `lsp-mode` enabled (a *scratch* buffer, for instance).

yyoncho: I checked and  lsp-method-requirements are actually implemented so lsp-feature? will work. It is odd that we haven't replaced the `lsp--get-symbol-to-rename`

nbfalcon:
I cleaned up the history a bit. Aside from that, I eliminated the new error functions, since they don't work they way I wanted anyway, and use `user-error' and normal `error' instead. This is probably also more idiomatic.

nbfalcon:
This PR is now ready.

lsp-mode.el
```diff
@@ -5646,29 +5646,32 @@ perform the request synchronously."
...
   "Rename the symbol (and all references to it) under point to NEWNAME."
-  (interactive (list (-when-let ((symbol . placeholder) (lsp--get-symbol-to-rename))
-                       (read-string (format "Rename %s to: " symbol) placeholder nil symbol))))
-  (unless newname
-    (user-error "A rename is not valid at this position"))
+  (interactive
+   (-if-let ((symbol . placeholder) (lsp--get-symbol-to-rename))
+       (list (read-string (format "Rename %s to: " symbol)
+                          placeholder nil symbol))
+     (user-error "`lsp-rename' is invalid here")))
```
yyoncho, line 5674: please revert that.

lsp-mode.el
```diff
@@ -5646,29 +5646,32 @@ perform the request synchronously."
...
                                 (if (lsp-range? response)
                                     response
                                   (lsp:prepare-rename-result-range response))))
-                (symbol (buffer-substring-no-properties start end))
+                (symbol (buffer-substring start end))
                 (placeholder (lsp:prepare-rename-result-placeholder response)))
           (cons symbol (or placeholder symbol))))
-    (let ((symbol (thing-at-point 'symbol t)))
-      (cons symbol symbol))))
+    (when-let ((sym (thing-at-point 'symbol)))
```
yyoncho, line 5665: why do you rename symbol? Why do you introduce when-let? 

Let's clarify that for future reference - do not introduce random renames, restructuring, etc - nobody is doing this PR in this repo. You are making the life of the reviewers harder than it should be. I have to go and test each that you have changed to verify that each line of all lines that you have changed does not introduce a regression. I cannot spend that much time reviewing trivial changes chasing for regressions.

lsp-mode.el
```diff
@@ -5646,29 +5646,32 @@ perform the request synchronously."
...
+  (if (lsp-feature? "textDocument/prepareRename")
+      (when-let ((response
+                  (lsp-request "textDocument/prepareRename"
+                               (lsp--text-document-position-params))))
         (-let* (((start . end) (lsp--range-to-region
                                 (if (lsp-range? response)
                                     response
                                   (lsp:prepare-rename-result-range response))))
-                (symbol (buffer-substring-no-properties start end))
+                (symbol (buffer-substring start end))
```
kiennq, line 5662: Why we need to use `buffer-substring` instead of `-no-properties` version?
Do we need the highlighting when prompting?

lsp-mode.el
```diff
@@ -5646,29 +5646,32 @@ perform the request synchronously."
...
                 (placeholder (lsp:prepare-rename-result-placeholder response)))
           (cons symbol (or placeholder symbol))))
-    (let ((symbol (thing-at-point 'symbol t)))
-      (cons symbol symbol))))
+    (when-let ((sym (thing-at-point 'symbol)))
+      (cons sym sym))))
 
 (defun lsp-rename (newname)
```
kiennq, line 5668: Why don't we just check the capability in this command (public) instead of changing internal one?
I don't think guarding every internal one is desirable/good from perf POV. Just guard once in public one should be enough, shouldn't it?

nbfalcon: If we guard only in `lsp-rename`, the user would be prompted to enter a new name even though renaming is unsupported.

lsp-mode.el
nbfalcon: Compare:

VS with highlighting:


Do you prefer the former?

lsp-mode.el
nbfalcon: Also, it's fast enough:
```emacs-lisp
(with-current-buffer "main.c"
  (benchmark-run 1000 (ignore (lsp-feature? "textDocument/rename"))))
```
yields 0.037 seconds.

Calling it only twice gives me 2ms timings, which is less than a 120FPS monitor's refresh latency.

lsp-mode.el
nbfalcon: Ok, I want to do that in the future. The `when-let` is because that function should return nil if there is nothing to rename at point. Without that `when-let`, if `(thing-at-point symbol)` were to yield nil, the function would return `(cons nil nil)`, instead of nil as it should. This, unlike I had assumed, didn't cause any real problems, because -if-let checks the bindings, not the input against nil.

```emacs-lisp
(eq nil nil) => true
(eq nil (cons nil nil)) => false
(equal nil (cons nil nil)) => false
```

lsp-mode.el
yyoncho: > If we guard only in `lsp-rename`, the user would be prompted to enter a new name even though renaming is unsupported.

see https://github.com/emacs-lsp/lsp-mode/pull/2317#issuecomment-724147973
https://github.com/yyoncho/lsp-mode/commit/443f801a4117361d2247c408b02dcc0d8c95a441 .

nbfalcon: Is it better to move the check outside of `lsp--get-symbol-to-rename` though?

lsp-mode.el
yyoncho: You will still rename all local variables, introduce random changes without documenting/testing them based on assumptions and you will expect the rest of the team to do that for you?!

nbfalcon: Sorry, I missspelled won't...

lsp-mode.el
nbfalcon: It's actually called only once though, in `lsp--get-symbol-to-rename`. We needn't add checks in every function, because the server will respond with a proper "not implemented" error if it isn't supported. Here, we cannot rely on that though, since we mustn't prompt the user for a name if the rename wouldn't work anyway.

nbfalcon:
I'll extend this to highlight the symbol being renamed; with the refactoring necessary do that, I can add tests ensuring that it will work in all cases.

nbfalcon:
The various `lsp-rename' helper functions are now thoroughly tested, and the symbol being renamed is highlighted.

nbfalcon:
In summary:
- `lsp-rename` now checks whether it is supported in the interactive form, trough a new function
- `lsp-rename` no longer user errors if the input is nil
- Renamed identifiers are marked in the buffer using `lsp-face-rename`
- The symbol to be renamed is propertized in the prompt, as is the initial input
- Lots of tests for the two functions. I'll need to investigate the CI failures; the tests work in isolation on my machine just fine.

nbfalcon:
Here is a screenshot:

lsp-mode.el
kiennq: I think you don't get my point. Let say if the `lsp--get-symbol-to-rename` is gone/refactored, which can happen since it's internal function, now we will need to find a new place to put the capacity checking.
On other hand, `lsp-rename` is public function, and is unlikely to change without a breaking change, put the capacity guard there is the best, since user has to use that function to actually go through rename.
Is there any specific concern why the capacity guard has to be put in `lsp--get-symbol-to-rename`?

lsp-mode.el
```diff
@@ -5654,29 +5654,69 @@ perform the request synchronously."
...
+Renaming can be done using `lsp-rename'."
+  :group 'lsp-faces)
+
+(defvar lsp--rename-history '()
+  "History for `lsp--read-rename'.")
+
+(defun lsp--read-rename (at-point)
+  "Read a new name for a `lsp-rename' at `point' from the user.
+AT-POINT shall be a structure as returned by
+`lsp--get-symbol-to-rename'.
```
kiennq, line 5691: It would be better to specify what the structure look like instead of just saying it will be the return value from `lsp--get-symbol-to-rename`

lsp-mode.el
```diff
@@ -5654,29 +5654,69 @@ perform the request synchronously."
...
+          ;; Do the `buffer-substring' first to not include `lsp-face-rename'
+          (rename-me (buffer-substring start end))
+          (placeholder (or placeholder? rename-me))
+
+          overlay)
+    ;; We need unwind protect, as the user might cancel here, causing the
+    ;; overlay to linger.
+    (unwind-protect
+        (progn
+          (setq overlay (-doto (make-overlay start end)
```
kiennq, line 5711: nit: is there a need for `-doto` in this case, since there's only on form?

lsp-mode.el
nbfalcon: The structure is defined in `lsp--get-symbol-to-rename`'s docstring.

lsp-mode.el
nbfalcon: We could add an additional `lsp-feature?` check to `lsp-rename`. However, the check in `lsp--get-symbol-to-rename` must stay: the interactive block is executed before the body, meaning the user would get prompted for the symbol to rename even though rename might not be possible at all or at point.

lsp-mode.el
nbfalcon: Fixed, thanks.

nbfalcon:
There are two issues with this as of now (hence the "Draft" status):
- [ ] `lsp-rename` sometimes override symbol-highlight overlays, and sometimes don't, which leads to incosistency.
- [ ] For some language servers, `prepareRename` can take too long or time out (`rust-analyzer`). Maybe somehow waiting until C-g would be better here, or perhaps setting a lower timeout, notifying the user somehow. Not using `prepareRename` is also a possibility, but is certainly sub-optimal.

And perhaps some ToDos:
- [ ] Support rename preview, like VSCode

danielmartin:
> And perhaps some ToDos:
> 
>   Support rename preview, like VSCode

I'm working on this feature at the moment.  It'll be orthogonal to the changes proposed here (if accepted).  You'll enter the rename preview mode if you invoke `lsp-rename` with `C-u`.

lsp-mode.el
kiennq: You can put the check in `(interactive)` form too. Like what @yyoncho suggested.
If you hate that, you can even make `(interactive '(nil))` and check for `newName` outside of `interactive` block and call to `lsp--get-symbol-to-rename`.

lsp-mode.el
```diff
@@ -5653,30 +5653,85 @@ perform the request synchronously."
...
+Renaming can be done using `lsp-rename'."
+  :group 'lsp-faces)
+
+(defface lsp-rename-placeholder-face '((t :inherit font-lock-variable-name-face))
+  "Face used to display the rename placeholder in.
+When calling `lsp-rename' interactively, this will be the face of
+the new name."
+  :group 'lsp-faces)
+
+(defvar lsp--rename-history '()
```
yyoncho, line 5698: Shouldn't this variable be public to avoid persisting it?

lsp-mode.el
```diff
@@ -5653,30 +5653,85 @@ perform the request synchronously."
...
+
+This function is for use in `lsp-rename' only, and shall not be
+relied upon."
+  (unless at-point
+    (user-error "`lsp-rename' is invalid here"))
+  (-let* ((((start . end) . placeholder?) at-point)
+          ;; Do the `buffer-substring' first to not include `lsp-face-rename'
+          (rename-me (buffer-substring start end))
+          (placeholder (or placeholder? rename-me))
+          (placeholder (propertize placeholder 'face 'lsp-rename-placeholder-face))
```
yyoncho, line 5718: I would vote for using different faces for placeholder and for the thing to rename. Using `bold` face seems to be a good candidate for one of them (to avoid being too colorful).

lsp-mode.el
```diff
@@ -5653,30 +5653,85 @@ perform the request synchronously."
...
+
+          overlay)
+    ;; We need unwind protect, as the user might cancel here, causing the
+    ;; overlay to linger.
+    (unwind-protect
```
yyoncho, line 5723: here we get an error if you kill the buffer which contains the thing to rename and then press ret in the read-string.

yyoncho (commented):
Looks good. Please drop a note which servers you have tested with. 

Few minor nits suggestions:

lsp-mode.el
nbfalcon: You make a good point, and from looking at other packages (`magit`, `evil (evil-ex)`), they also seem to be using public history variables. I'm making this public.

lsp-mode.el
nbfalcon: Do you propose overriding the face of the thing to be renamed the same way as is done for the placeholder? What we might also want to explore is adding a face to it, using `add-face-text-property`. I'll post some screenshots of various approaches soon.

nbfalcon:
I tested with `jdtls`, `rust-analyzer`, `ccls` and `pyright`.

lsp-mode.el
nbfalcon: That is unavoidable, because `read-string` errors if the old buffer was killed. However, I don't see how that would cause issues? Overlays can be deleted if the old buffer is dead; the error would just get propagated. I think this should be fixed in Emacs core, because all packages can't be expected to guard against a buffer-live-p error just to be able to use `read-string`.

`delete-overlay` + dead-buffer:
```elisp
(delete-overlay
 (with-temp-buffer
   (insert "foo")
   (make-overlay 0 1)))
```

Bind `M-k` to kill the old buffer in the minibuffer map:
```elisp
(define-key minibuffer-local-map (kbd "M-k") (lambda () (interactive) (kill-buffer (window-buffer (minibuffer-selected-window)))))
```

nbfalcon:
`rust-analyzer` seems to send incorrect ranges in certain cases (sometimes with an extra :uri), which causes the rename overlays and placeholder to be borked (not a regression, master also has this). This is a spec violation though, as the result shall be a Range, not something containing a :uri.

nbfalcon:
Digging trough RA's code, it seems that the strange results are fixed, at least in master, because there is no code returning a uri in `handle_prepare_rename`.

lsp-mode.el
yyoncho: Ok - I thought the error can be prevented.

yyoncho: .. easily.

lsp-mode.el
yyoncho: I am proposing using 2 different faces like this:

``` emacs-lisp
(let ((placeholder "foo")
      (rename-me "foo"))
  (read-string (format "Rename %s to: " (propertize rename-me 'face 'bold)) 
               (propertize placeholder 'face 'font-lock-variable-name-face)
               'lsp-rename-history))
```

yyoncho (approved):

yyoncho:
Thank you.

I merged it as it is - the cosmetics about the faces can be discussed separatelly.

PR #2319, opened 2020-11-09
Run `lsp-configure-buffer' in an idle-timer

### LINKED ISSUE #2138: lsp-mode with gopls be slow when using real-auto-save in Emacs27
I am using Emacs27 now

I use the `https://github.com/ChillarAnand/real-auto-save`, and set the `real-auto-save-interval` to 1

in go file, the completion be very slow, but the other function like go-to-definition is ok

I also write python, real-auto-save is ok with lsp-python-ms.

On the other hand, emacs26 works fine with lsp-mode and gopls.

sorry, I try the lsp-start-plain.el, but there's no useful log.

If you need more info, please tell me.

nbfalcon:
Fixes #2138.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3610,13 +3610,13 @@ in that particular folder."
     (when lsp-enable-xref
       (add-hook 'xref-backend-functions #'lsp--xref-backend nil t))
 
-    (lsp-configure-buffer)
     (let ((buffer (lsp-current-buffer)))
       (run-with-idle-timer
        0.0 nil
        (lambda ()
          (when (lsp-buffer-live-p buffer)
            (lsp-with-current-buffer buffer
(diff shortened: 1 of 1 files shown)


jcs090218:
Base on the information what @zy9306 gave us; I think the issue is more of `Emacs 26` vs `Emacs 27`.

@nbfalcon Can you elaborate the fix? 😕 Thanks! ;)

nbfalcon:
This might be related to 27.1 vs 26 - I am running Emacs 27.1 built from source. I am myself acutally unsure why this works. It is not because of `lsp-with-current-buffer`, since I tried the fix outside of the idle-timer within such a block, but that didn't work.

nbfalcon:
This is fixed by #2327, as I could confirm.

PR #2327, opened 2020-11-11
Capture lsp--cur-workspace in lsp-configure-buffer

### LINKED ISSUE #2318: `lsp-configure-buffer`: doesn't configure anything
**Describe the bug**
When opening a new file, after `lsp-mode` is enabled for it due to call to `lsp`, none of the features in `lsp-configure-buffer` are enabled, as I concluded from debugging that function using `edebug` (`lsp-feature?` always returned nil). I noticed this because symbol occurrences weren't highlighted.

**To Reproduce**
1. Open a file
2. `M-x lsp`
3. Note that symbols aren't highlighted

**Expected behavior**

rust-analyzer, ccls, ...

**OS**
Ubuntu 20.04 LTS

### LINKED ISSUE #2326: [Bug] Need config lsp twice in go-mode-hook to take effect
**Describe the bug**
Need config `lsp` twice in go-mode-hook to take effect

After upgrade the lsp-mode to 202011xx(I didn't remember the exactly version from which it works wrong), I have to config `lsp` twice in my go-mode-hook to make it take effect. Before that, I only write one `lsp` word in the my .emacs file
```
 
'(go-mode-hook
   '(lsp lsp company-mode yas-minor-mode
...... (other hooks)
```

The first day I upgrade the lsp-mode to 202011xx, when I opened golang source file, no complete popup, no jump, it seems that the lsp-mode not enabled, so I typed `M-x lsp`, it works. When I checked the go-mode-hook, `lsp` was there as what it was, so I tried add another `lsp`, it works.

**To Reproduce**

**Expected behavior**
Only one `lsp` in config file should works well

Golang and gopls

**OS**
MacOS 10.15.6
Emacs: Version 27.1, download from https://emacsformacosx.com/

**Error callstack**
If there is an error please reproduce the issue with `toggle-debug-on-error` and after setting `lsp-print-io` to t, and then include the callstack and attach the content of `*lsp-log*` buffer and the content of the server log(use `lsp-workspace-show-log`)

If `emacs` hangs please do `M-x` `toggle-debug-on-quit` and then do `C-g` when emacs hangs and include the callstack as part of the issue report.

``` emacs-lisp
(setq lsp-print-io t)
```

yyoncho:
- fixes #2326
- fixes #2318

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3652,7 +3652,12 @@ in that particular folder."
 
 (defun lsp-configure-buffer ()
   "Configure LSP features for current buffer."
-  (let (lsp--cur-workspace)
+  ;; make sure the core is running in the context of all available workspaces
+  ;; to avoid misconfiguration in case we are running in `with-lsp-workspace' context
+  (let ((lsp--buffer-workspaces (cond
+                                 (lsp--buffer-workspaces)
+                                 (lsp--cur-workspace (list lsp--cur-workspace))))
+        lsp--cur-workspace)
     (when lsp-auto-configure
       (when (and lsp-enable-text-document-color
                  (lsp-feature? "textDocument/documentColor"))


kiennq (approved):

trivialfis (approved):
Tested it works with clangd.  Thanks for the fix!

Issue #2279, opened 2020-10-21
Extend lsp-clangd to include protocol extension


petr-tik:
Clangd has several extensions introduced at different versions. After the clangd executable is found, we can check what version is running and enable everything <= found version. 

Personally, I am most interested in these extensions (descending order of preference):
- [ ] Switch between source/header
- [ ] Compilation commands
- [ ] Code completion scores and sorting auto-complete candidates accordingly
- [ ] Inline fixes for diagnostics since clangd "computes these eagerly" we can provide them straightaway. 
- [ ] Compilation commands

https://clangd.llvm.org/extensions.html

kiennq:
The code completion scores stuff has been supported.

petr-tik:
I wanted to add a SampleCppProject with CMakeLists to test lsp-clangd. 

Is that ok or do you think this is too heavy a dependency (cmake + clangd) to install in our CI?

yyoncho:
Depends on how much will slow down the CI. My guess is that it will be fine

petr-tik:
sounds great - i will be honest I have never written elisp tests, so they might be naive, but I look forward to your feedback

petr-tik:
Since the code completion scores is already done, I am going to tackle the "Switch between source/header" function and use it to write some tests too

nbfalcon:
Why are there two "Compilation commands" checkboxes? Aside from that, what do you mean by inline fixes for diagnostics? Flycheck doesn't support fixes.

nbfalcon:
FYI, there is also `projectile-find-other-file` to switch between source/header.

nbfalcon:
Compilation commands also kind of doesn't make sense, since Emacs is not an Editor with a strong sense of project configuration like say Eclipse. There are no project settings, meaning that clangd would end up relying on the compile_commands.json anyway.

yyoncho:
> FYI, there is also `projectile-find-other-file` to switch between source/header.

the clangd version will be faster and more precise. 

> Aside from that, what do you mean by inline fixes for diagnostics? Flycheck doesn't support fixes.

We do have quick fixes in lsp-treemacs-errors-list. Also, I think that the overall idea is to save one call to the server. But then all diagnostics will receive the code actions as well. Which will increase the traffic.

petr-tik:
> > FYI, there is also `projectile-find-other-file` to switch between source/header.
> 
> the clangd version will be faster and more precise.

[...]

PR #2331, opened 2020-11-15
Address #2279 - add switch-to-other for clangd

petr-tik:
Made an interactive function, use the lowest cmake version that supports compile_commands.json. 

Add a cmake-based repo for intergration tests of clangd. 

Install cmake and clang(d) for unix-ci and add a makefile step for unix-ci. Experimenting with unix-ci with github actions, please let me know, if I am doing something wrong.

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -235,18 +235,28 @@ returned to avoid that the echo area grows uncomfortably."
         (car (s-lines (lsp--render-element contents)))))))
 
 
+(defun lsp-clangd-to-other ()
+  "Open the corresponding header/source file."
+  (if-let ((other-fname
+            (with-lsp-workspace (lsp-find-workspace 'clangd)
+              ;; TODO define an extension with lsp-interace
+              ;; similar to rust-analyzer on lines 278-292 in
+              ;; lsp-protocol.el
+              (lsp-send-request (lsp-make-request
+                                 "textDocument/switchSourceHeader"
+                                 (lsp--text-document-identifier))))
+            ))
+      other-fname
+    (lsp--info "This file doesn't have a corresponding file")
+    nil))
 ;; Running this in "/home/petr_tik/Coding/rr/src/AutoRemoteSyscalls.cc"
-(with-lsp-workspace (lsp-find-workspace 'clangd)
-                             (lsp-send-request (lsp-make-request
-                                                "textDocument/switchSourceHeader"
-                                                (lsp--text-document-identifier))))
 ;; returns this a URI
 ;; "file:///home/petr_tik/Coding/rr/src/AutoRemoteSyscalls.h"
-;; running in a file with a companion - returns nil
+;; running in a file without a companion - returns nil
 
-;; (url-generic-parse-url "file:///home/petr_tik/Coding/rr/src/AutoRemoteSyscalls.h")
+;;
 ;; returns a structure that we can access
-
+;; (url-generic-parse-url "file:///home/petr_tik/Coding/rr/src/AutoRemoteSyscalls.h")
 
 (provide 'lsp-clangd)
 ;;; lsp-clangd.el ends here
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -237,7 +237,10 @@ returned to avoid that the echo area grows uncomfortably."
 
 (defun lsp-clangd-to-other ()
   "Open the corresponding header/source file."
+  (interactive)
   (if-let ((other-fname
+            ;; find-workspace might be too brittle if users make their own lsp IDs
+            ;; eg. clangd-remote
             (with-lsp-workspace (lsp-find-workspace 'clangd)
               ;; TODO define an extension with lsp-interace
               ;; similar to rust-analyzer on lines 278-292 in
@@ -246,7 +249,7 @@ returned to avoid that the echo area grows uncomfortably."
                                  "textDocument/switchSourceHeader"
                                  (lsp--text-document-identifier))))
             ))
-      ;; TODO use lsp-goto-location instead
+      ;; TODO maybe worth using lsp-goto-location instead
       (find-file (lsp--uri-to-path other-fname))
     (lsp--info "This file doesn't have a corresponding file")
     nil))
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -236,25 +236,17 @@ returned to avoid that the echo area grows uncomfortably."
 
 
 (defun lsp-clangd-to-other (&optional arg)
-  "Open the corresponding header/source file or return nil.
+  "Open the corresponding header/source file.
 
 If called with ARG universal argument, the file will open in the other window."
   (interactive "P")
-  (if-let ((other-fname
-            ;; find-workspace might be too brittle if users make their own lsp IDs
-            ;; eg. clangd-remote
-              ;; TODO define an extension with lsp-interace
-              ;; similar to rust-analyzer on lines 278-292 in
-              ;; lsp-protocol.el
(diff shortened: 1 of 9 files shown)


petr-tik:
windows-ci currently failing - i didn't have a way to test it locally and would appreciate any help with github actions.

clients/lsp-clangd.el
```diff
@@ -235,5 +235,24 @@ returned to avoid that the echo area grows uncomfortably."
         (car (s-lines (lsp--render-element contents)))))))
 
 
+(defun lsp-clangd-to-other ()
+  "Open the corresponding header/source file."
+  (interactive)
+  (if-let ((other-fname
+            ;; find-workspace might be too brittle if users make their own lsp IDs
+            ;; eg. clangd-remote
+            (with-lsp-workspace (lsp-find-workspace 'clangd)
```
kiennq, line 244: I don't think you need to do this, `lsp-send-request` can find the workspace that support the command. Or if the capability is not registered, it will send to all workspace, which is actually fine.

clients/lsp-clangd.el
```diff
@@ -235,5 +235,24 @@ returned to avoid that the echo area grows uncomfortably."
...
+              ;; similar to rust-analyzer on lines 278-292 in
+              ;; lsp-protocol.el
+              (lsp-send-request (lsp-make-request
+                                 "textDocument/switchSourceHeader"
+                                 (lsp--text-document-identifier))))
+            ))
+      ;; TODO maybe worth using lsp-goto-location instead
+      (find-file (lsp--uri-to-path other-fname))
+    (lsp--info "This file doesn't have a corresponding file")
+    nil))
```
kiennq, line 255: Is this needed?

test/lsp-clangd-test.el
```diff
@@ -24,6 +24,24 @@
 (require 'ert)
 (require 'lsp-clangd)
 
+;; TODO move to a common/shared test-utils.el
```
kiennq, line 27: Isn't this in `lsp-integration-test` already? Can't you just `require` or `load` it?

petr-tik: Yes it is - is it ok, if I move these helpers out to a `test-utils.el` file? I am not knowledgeable enough to know about macro evaluation if I decide to load lsp-integration-test.el in lsp-clangd-test.el

clients/lsp-clangd.el
petr-tik: By "this" are you referring you to the `nil` return value or lsp--info message?

petr-tik: also, lsp-goto-location doesn't show up when I grep for it in the whole repo, which suggests I can get rid of the TODO?

test/lsp-clangd-test.el
yyoncho: > I am not knowledgeable enough to know about macro evaluation if I decide to load lsp-integration-test.el in lsp-clangd-test.el

That should be fine.

clients/lsp-clangd.el
```diff
@@ -235,5 +235,26 @@ returned to avoid that the echo area grows uncomfortably."
...
+  "Open the corresponding header/source file or return nil.
+
+If called with ARG universal argument, the file will open in the other window."
+  (interactive "P")
+  (if-let ((other-fname
+            ;; find-workspace might be too brittle if users make their own lsp IDs
+            ;; eg. clangd-remote
+              ;; TODO define an extension with lsp-interace
+              ;; similar to rust-analyzer on lines 278-292 in
+              ;; lsp-protocol.el
```
yyoncho, line 248: I think you can delete this comment. You need your interfaces only if there are no lsp-mode types already covering this functionality. And lsp-find-workspace is supposed to be used very rarely. In this case, lsp-mode should be able to handle the request directly using lsp-request.

clients/lsp-clangd.el
```diff
@@ -235,5 +235,26 @@ returned to avoid that the echo area grows uncomfortably."
...
+              (lsp-send-request (lsp-make-request
```
yyoncho, line 249: better to use lsp-request (just remove lsp-make-request call).

test/lsp-clangd-test.el
```diff
@@ -62,4 +80,65 @@
...
+   ;; switch to other again - check it's main.cpp again
+   (lsp-clangd-to-other)
+   (should (string= (buffer-name) "main.cpp"))))
+
+(ert-deftest lsp-clangd-switch-to-nonexistent-other ()
+  ;; TODO add a cpp file without a corresponding header file and add it to CML
+  (lsp-in-sample-cpp-project
+   (find-file (f-join lsp-test-location "fixtures/SampleCppProject/src/individual_file.cpp"))
+   (should (string= (buffer-name) "individual_file.cpp"))
+   (should (= (lsp-clangd-to-other) nil))
```
yyoncho, line 141: Remove this line after you fix the @kiennq comment about removing nil. Asserting that the buffer hasn't changed should be good enough.

yyoncho (commented):
Looks good to me, few minor nits. Good to go after getting the tests green.

clients/lsp-clangd.el
kiennq: Yes, the "this"  points to `nil`.
Also since the server is returning URI instead of location, I think you can remove the TODO.

petr-tik:
> Looks good to me, few minor nits. Good to go after getting the tests green.

Thanks for the review. I would appreciate any help with getting tests to pass. "It works on my machine" (TM), but I don't seem to understand why it fails in CI. 

Is there a way to programmatically print lsp-describe-session to the CI logs and see what capabilities this LSP offers? My slight suspicion is we are accidentally connecting to the wrong clangd of some kind. 

Maybe this should be another feature, but it would be great if we logged/presented the LSP server version somewhere. It's usually available in stdout of clangd like this

> I[23:10:41.820] Ubuntu clangd version 11.0.0-++20201008094836+176249bd673-1~exp1~20201008080315.114

kiennq:
For the time being, I think you can just fake the return from server, so no need to connect it to real `clangd` for test to work

petr-tik:
> For the time being, I think you can just fake the return from server, so no need to connect it to real `clangd` for test to work

I would like to start this test integration harness so I and other can add features to it in the future.

yyoncho:
> Is there a way to programmatically print lsp-describe-session to the CI logs and see what capabilities this LSP offers? My slight suspicion is we are accidentally connecting to the wrong clangd of some kind.

You may try using lsp--find-workspaces-for and add debug markers in it.

petr-tik:
> You may try using lsp--find-workspaces-for and add debug markers in it.

What kind of debug markers? This is v puzzling. Clangd-11 is the most recent version and it is certainly supported when I call it in my editor with clangd-11

yyoncho:
> What kind of debug markers? This is v puzzling. Clangd-11 is the most recent version and it is certainly supported when I call it in my editor with clangd-11

like calls to `message` and so on. We need to figure out why this function returns nil. I think that somehow the call runs in a buffer that is not managed by lsp-mode but all this has to be traced.

petr-tik:
> like calls to `message` and so on. We need to figure out why this function returns nil. I think that somehow the call runs in a buffer that is not managed by lsp-mode but all this has to be traced.

The fact that this test line seems to pass suggests that we are running in the right file. It used to fail with the message capability not supported. 

```elisp
   (should (string= (buffer-name) "main.cpp"))
```

petr-tik:
Now it's failing because github actions cancels the step "setup cmake". Please excuse the email spam - i didn't know github actions is so chatty.

petr-tik:
Will close the PR until I can get my CI to pass without spamming the main repo.

PR #2332, opened 2020-11-16
Mark lsp-solargraph-use-bundler as safe if booleanp

amake:
Wanting to use bundler or not is very likely to be project-specific and is something you would want to set in a .dir-locals.el.


jcs090218 (approved):

yyoncho:
Thank you!

PR #2333, opened 2020-11-16
[WIP] GalaxyProject XML

mtekman:
Add the the GalaxyProject's language server for XML files

https://github.com/galaxyproject/galaxy-language-server

### DIFF
--- a/clients/lsp-galaxy.el
+++ b/clients/lsp-galaxy.el
@@ -1,0 +1,86 @@
+;;; lsp-galaxy.el --- description -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 emacs-lsp maintainers
+
+;; Author: emacs-lsp maintainers
+;; Keywords: lsp, php
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; LSP Clients for the GalaxyProject
+
+;;; Code:
+(require 'lsp-mode)
+
+(defgroup lsp-galaxy nil
+  "LSP support for Galaxy"
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/galaxyproject/galaxy-language-server/")
+  :package-version '(lsp-mode . "6.3"))
+
+(defcustom lsp-galaxy-galaxy-cmd '["python" "-m" "galaxyls"]
+  "Path to binary used in Galaxy Language Server.  Defaults to `python -m galaxyls' if nil."
+  :type 'lsp-string-vector
+  :group 'lsp-galaxy
+  :package-version '(lsp-mode . "7.0.1"))
+
+(defcustom lsp-galaxy-language-server-host "127.0.0.1"
+  "Choose host address."
+  :type 'string
+  :group 'lsp-galaxy
+  :package-version '(lsp-mode . "6.3"))
+
+(defcustom lsp-galaxy-language-server-port 2087
+  "Choose listen port."
+  :type 'integer
+  :group 'lsp-galaxy
+  :package-version '(lsp-mode . "6.3"))
+
+(defcustom lsp-galaxy-language-server-client-version "2.1.0"
+  "Choose client version."
+  :type 'string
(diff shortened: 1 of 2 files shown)


clients/lsp-galaxy.el
```diff
@@ -0,0 +1,86 @@
+;;; lsp-galaxy.el --- description -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020 emacs-lsp maintainers
+
+;; Author: emacs-lsp maintainers
+;; Keywords: lsp, php
```
yyoncho, line 6: nit: fix keywords

clients/lsp-galaxy.el
```diff
@@ -0,0 +1,86 @@
...
+;; LSP Clients for the GalaxyProject
+
+;;; Code:
+(require 'lsp-mode)
+
+(defgroup lsp-galaxy nil
+  "LSP support for Galaxy"
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/galaxyproject/galaxy-language-server/")
+  :package-version '(lsp-mode . "6.3"))
```
yyoncho, line 32: nit: please change all versions to 7.1.0

lsp-mode.el
```diff
@@ -355,6 +355,7 @@ unless overridden by a more specific face association."
          lsp-kotlin lsp-lua lsp-nim lsp-nix lsp-metals lsp-ocaml lsp-perl lsp-php lsp-pwsh
          lsp-pyls lsp-python-ms lsp-purescript lsp-r lsp-rf lsp-rust lsp-solargraph
          lsp-tex lsp-terraform lsp-vala lsp-verilog lsp-vetur lsp-vhdl lsp-vimscript lsp-xml
+         lsp-galaxy
```
yyoncho, line 358: nit: packages are sorted alphabetically.

lsp-mode.el
```diff
@@ -809,7 +810,7 @@ Changes take effect only when a new session is started."
                                         (".*\\.tsx$" . "typescriptreact")
                                         (".*\\.ts$" . "typescript")
                                         (".*\\.jsx$" . "javascriptreact")
-                                        (".*\\.xml$" . "xml")
+                                        (".*\\.xml$" . "galaxy")
```
yyoncho, line 813: This change is not fine - it will break the xml language servers. Are you sure it should go like this?

yyoncho (commented):
Looks good, you should add the server in docs/lsp-clients.json as well otherwise it won't be displayed in the docs.

lsp-mode.el
mtekman: whoops sorry, I committed this part without thinking. The language server does however process exclusively ".XML" files, which will definitely conflict with the default behaviour.

Is there a way to get around this? Or is this something that a galaxyproject user should override in their config?

clients/lsp-galaxy.el
mtekman: should I change the authors here too, or are all contributors maintainers? (I'm also happy to leave it as it is, I just feel I should not burden you with the task of maintaining this extension if it falls out of date)

yyoncho: You can leave it as it is or put your name.

lsp-mode.el
yyoncho: Either this or we can provide a hook in the language server so it can change the mappings in a buffer-local variable.

clients/lsp-galaxy.el
```diff
@@ -0,0 +1,86 @@
...
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection
+                   (lambda ()
+                     (list lsp-galaxy-galaxy-cmd "--tcp"
+                           (format "--host %s --port %d"
+                                   lsp-galaxy-language-server-host
+                                   lsp-galaxy-language-server-port))))
+  :major-modes '(galaxy-mode cgalaxy-mode)
```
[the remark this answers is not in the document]
yyoncho: ?

as a side note, we may use these modes to map the files. Then the user will fire up this major mode when they want to start galaxty ls.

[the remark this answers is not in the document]
kiennq: Can `galaxy-ts` support arbitrary `xml` file? If not this shouldn't contains `nxml-mode`

mtekman:
@yyoncho I'm actually unable to get the mode to start, despite registering it correctly (AFAIK).

When I open an xml file, the warning message tells me:

```
LSP :: There are no language servers supporting current mode `planemo-mode' registered with `lsp-mode'.
2. The language server that you expect to run is not configured to run for major mode `planemo-mode'. You may check that by checking the `:major-modes' that are passed to `lsp-register-client'.
3. `lsp-mode' doesn't have any integration for the language behind `planemo-mode'. Refer to https://emacs-lsp.github.io/lsp-mode/page/languages and https://langserver.org/ .
```

yyoncho:
Can you make sure that the `lsp-galaxy.el` is loaded?

mtekman:
I think it is. If I perform an `eval-buffer` on the  "lsp-galaxy.el" buffer, and then in my XML buffer I run `(require 'lsp-galaxy)` I still get the same message

PR #2338, opened 2020-11-18
Update projectLoadingStart message in angular ls

systemctl603:
`angular-language-service` new sends these messages prefixed with just `angular` instead of `angular-language-service`,
so warnings pop up about this not being a known message.

Here I just replaced that prefix, and now the warnings disappear.


yyoncho (approved):

yyoncho:

PR #2339, opened 2020-11-18
`lsp-avy-lens': don't `error' for no lenses

nbfalcon:
If there are no elements, `avy-process` returns t, causing
`funcall-interactively` to error. Before continuing to that function,
check if there are lenses, only then calling `avy`. If there are none,
`user-error` out.

Replacing the guard from the previous commit with `consp' was also an
option, but that would yield a less intuitive "zero candidates" message
instead.

Don't worry, this bug wasn't introduced by my previous PR; `avy-process` can also return t if there are no candidates, so we should guard against zero lenses, too.

### DIFF
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -376,7 +376,8 @@ CALLBACK - callback for the lenses."
   (unless lsp-lens--overlays
     (user-error "No lenses in current buffer"))
   (let* ((avy-action 'identity)
-         (actions (avy-process
+         (action (cl-third
+                  (avy-process
                    (-mapcat
                     (lambda (overlay)
                       (-map-indexed
@@ -406,8 +407,8 @@ CALLBACK - callback for the lenses."
                    (lambda ()
                      (--map (overlay-put it 'before-string
                                          (overlay-get it 'lsp-original))
-                            lsp-lens--overlays)))))
-    (when actions (funcall-interactively (cl-third actions)))))
+                            lsp-lens--overlays))))))
+    (when action (funcall-interactively action))))
 
 (provide 'lsp-lens)
 ;;; lsp-lens.el ends here
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -391,7 +391,7 @@ CALLBACK - callback for the lenses."
                             (str (propertize (string (car (last path)))
                                              'face 'avy-lead-face))
                             (old-str (overlay-get ov 'before-string))
-                            (old-str-tokens (s-split "\|" old-str))
+                            (old-str-tokens (s-split "|" old-str))
                             (old-token (seq-elt old-str-tokens index))
                             (tokens `(,@(-take index old-str-tokens)
                                       ,(-if-let ((_ prefix suffix)
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -375,9 +375,10 @@ CALLBACK - callback for the lenses."
   (interactive)
   (if (not lsp-lens-mode)
       (user-error "`lsp-lens-mode' not active")
+    (unless lsp-lens--overlays
+      (user-error "No lenses in current buffer"))
     (let* ((avy-action 'identity)
-           (action (cl-third
-                    (avy-process
+           (actions (avy-process
                      (-mapcat
                       (lambda (overlay)
                         (-map-indexed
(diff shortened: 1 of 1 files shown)


lsp-lens.el
```diff
@@ -375,9 +375,10 @@ CALLBACK - callback for the lenses."
   (interactive)
   (if (not lsp-lens-mode)
       (user-error "`lsp-lens-mode' not active")
+    (unless lsp-lens--overlays
+      (user-error "No lenses in current buffer"))
     (let* ((avy-action 'identity)
-           (action (cl-third
-                    (avy-process
+           (actions (avy-process
                      (-mapcat
```
nbfalcon, line 382: We can use `mapcan` here, since only calls to `-map-indexed`, which always returns a new list, are mapped.

lsp-lens.el
```diff
@@ -375,9 +375,10 @@ CALLBACK - callback for the lenses."
   (interactive)
   (if (not lsp-lens-mode)
       (user-error "`lsp-lens-mode' not active")
```
nbfalcon, line 377: Those two `user-error`s can be linearized and put at the start of the function.

nbfalcon:
I have proposed some additional refactoring changes in comments, which I decided to not implementing for the sake of limiting this PR's scope. Should I not implement them, make a separate refactoring PR, or implement them in this PR?

lsp-lens.el
yyoncho: Sounds good. You could the lsp-lens-mode at all. Your check will be sufficient.

lsp-lens.el
yyoncho: seems like it is fine to stay as it is. If we are looking for optimization here(which IMO is not the case since this is not a critical section) then we should drop at all the higher-order functions in favour of using `while`.

lsp-lens.el
```diff
@@ -407,8 +408,8 @@ CALLBACK - callback for the lenses."
                      (lambda ()
                        (--map (overlay-put it 'before-string
                                            (overlay-get it 'lsp-original))
-                              lsp-lens--overlays))))))
-      (when action (funcall-interactively action)))))
+                              lsp-lens--overlays)))))
+      (when actions (funcall-interactively (cl-third actions))))))
```
yyoncho, line 412: I don't understand how does this fixes avy returning `t`? We will still call cl-third over `t`?

nbfalcon: This is an unrelated change, sorry. `avy-process` returns t only if there are no input lenses. This is handled by the `unless lsp-lens--overlays` `user-error` at the top. If there are lens overlays, `avy-process` cannot ever return t here. Should I revert?

lsp-lens.el
nbfalcon: It isn't about the optimization here; IMHO, using `mapcan` improves the code readability here, because it makes obvious that the returned lists of the inner lambda are newly-constructed, not taken from some global variable, because mutating them is Ok. This isn't the case for `dap-launch-configuration-providers`, for example, where we have to use `-mapcat`. Do you think I should leave it anyway?

lsp-lens.el
nbfalcon: Fixed.

lsp-lens.el
yyoncho: If this is unrelated it seems like it is better to revert or use better naming since it does not return `actions`.

lsp-lens.el
yyoncho: That is the whole benefit of using the pure function - you don't have to worry about the implementation and it will still work even if the inner code has changed. Using the non-pure function introduces one more thing to check when working with this code.

nbfalcon: Ok, given that this isn't performance critical, I'll leave this as-is.

lsp-lens.el
nbfalcon: I decided to revert, because micro-optimizing `cl-third` isn't worth-it and because coming up with a name that is as good as `action` here turned out to be a challenge (`result`?, `data`?). In addition, this won't error if the selected lens has no action (maybe we'll want to support selecting actionless lenses in the future for some reason?).

nbfalcon:
Unrelated, but shouldn't [the regexp here](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-lens.el#L394) be "\\|"? `s-replace` takes a regexp, and "\|" = "|". This works because a lone "|" is interpreted literally, but it is confusing from the perspective of someone looking at the code.

yyoncho:
> Unrelated, but shouldn't [the regexp here](https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-lens.el#L394) be "|"? "\|" = "|", so this makes no difference in terms of bugs. However, that way it seems like less of a typo.

If you wish you can change it. Let me know if you want to do it in this PR.

nbfalcon:
I'd prefer to do it in this PR. Can I go ahead?

yyoncho:
sure

nbfalcon:
Fixed. You may merge now.

Issue #2336, opened 2020-11-17
[Documentation] Document lsp-install-server on the website


jarivm:
After having trouble installing a language server (ESLint) manually, I discovered `lsp-install-server`. This made installing ESLint trivial, so I wish I found out about it sooner. 

**Suggested changes**
- Document `lsp-install-server` on the [languages page](https://emacs-lsp.github.io/lsp-mode/page/languages/), and mention not every language server is currently supported.
- Document installing ESLint via `lsp-install-server` on the [ESLint page](https://emacs-lsp.github.io/lsp-mode/page/lsp-eslint/) and remove the link to the (now) outdated Wiki page. 

I am going to start working on this myself!

yyoncho:
We used to document that the lsp-mode supports automatic installation - like this https://github.com/emacs-lsp/lsp-mode/blob/bc25dfb16dbfea4804c45a46504bc0fe7d9d44b9/docs/lsp-clients.json#L139 . This property is not updated for all servers that support automatic installation. It will be nice if we list the automatic and manual installation where appropriate in addition to covering lsp-install-server in the getting started section.

ericdallo:
I think we can close this, right?

PR #2340, opened 2020-11-18
Document lsp-install-server for languages that support it

jarivm:
This change documents lsp-install-server for languages that support it. The corresponding issue is #2336.

If this PR is merged, it should be investigated if the deprecated [ESLint wiki page](https://github.com/emacs-lsp/lsp-mode/wiki/LSP-ESlint-integration) can be removed, as the changes in this PR remove the link to this page, thus making the page redundant.

### DIFF
--- a/docs/lsp-clients.json
+++ b/docs/lsp-clients.json
@@ -22,6 +22,7 @@
     "server-name": "bash-language-server",
     "server-url": "https://github.com/mads-hartmann/bash-language-server",
     "installation": "npm i -g bash-language-server",
+    "lsp-install-server": "bash",
     "debugger": "Not available"
   },
   {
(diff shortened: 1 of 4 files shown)


yyoncho:
Thank you.

> If this PR is merged, it should be investigated if the deprecated [ESLint wiki page](https://github.com/emacs-lsp/lsp-mode/wiki/LSP-ESlint-integration) can be removed, as the changes in this PR remove the link to this page, thus making the page redundant

Yes, we can do that if the automatic install works unconditionally, which unfortunately is not the case. E. g. user might not unzip the tool on the machine, might have marketplace behind firewall, etc. 

I wonder whether it makes sense to list the automatic installation dependencies(e. g. `npm`) in the docs, WDYT?

ericdallo (approved):
Thank you!

jarivm:
> I wonder whether it makes sense to list the automatic installation dependencies(e. g. `npm`) in the docs, WDYT?

Yes, I think installation dependencies should be listed, because it would make installation more accessible.

yyoncho (approved):

PR #2348, opened 2020-11-20
Add support for Sorbet, Ruby typechecker

twopoint718:
This is a first crack at a Sorbet LSP client. Let me know if there are things I need to change/update. Thanks!

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,4 +1,6 @@
 * Changelog
+** Release 7.2
+  * Add [[https://github.com/sorbet/sorbet][Sorbet Language Server]] for typechecking Ruby code.
 ** Release 7.1
   * Safe renamed ~lsp-diagnose~ to ~lsp-doctor~.
   * Add ~lsp-modeline-code-actions-segments~ for a better customization.
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,6 +1,4 @@
 * Changelog
-** Release 7.2
-  * Add [[https://github.com/sorbet/sorbet][Sorbet Language Server]] for typechecking Ruby code.
 ** Release 7.1
   * Safe renamed ~lsp-diagnose~ to ~lsp-doctor~.
   * Add ~lsp-modeline-code-actions-segments~ for a better customization.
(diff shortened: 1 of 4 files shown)


clients/lsp-sorbet.el
```diff
@@ -0,0 +1,56 @@
...
+    (if lsp-sorbet-use-bundler
+              (append '("bundle" "exec") lsp-command)
+            lsp-command)))
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection
+                   #'lsp-sorbet--build-command)
+  :major-modes '(ruby-mode enh-ruby-mode)
+  :server-id 'ruby-ls))
```
yyoncho, line 53: you should use a different server-id.

clients/lsp-sorbet.el
```diff
@@ -0,0 +1,56 @@
...
+(defun lsp-sorbet--build-command ()
+  "Build sorbet command"
+  (let ((lsp-command '("srb" "typecheck" "--lsp" "--disable-watchman")))
+    (if lsp-sorbet-use-bundler
+              (append '("bundle" "exec") lsp-command)
+            lsp-command)))
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection
```
yyoncho, line 50: you should pass :priority as well. The priority will determine which server will kick in in case there are 2 servers present on the path.

clients/lsp-sorbet.el
```diff
@@ -0,0 +1,56 @@
...
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(defgroup lsp-sorbet nil
+  "LSP support for Ruby, using the Sorbet language server."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/sorbet/sorbet")
+  :package-version '(lsp-mode . "6.1"))
```
yyoncho, line 33: please change the version to 7.1.0 here and bellow.

yyoncho (commented):
Looks good, few nits:

Can you also add the new client to `docs/lsp-clients.json` ?

clients/lsp-sorbet.el
yyoncho: I guess the priority should be -2 I guess to make this with lower priority than the solargraph one.

CHANGELOG.org
```diff
@@ -1,4 +1,6 @@
 * Changelog
+** Release 7.2
```
yyoncho, line 2: 7.1 is still unreleased, can you move that there?

yyoncho (approved):

yyoncho:

PR #2350, opened 2020-11-22
Fix `lsp-rename' tests

nbfalcon:
Use `save-current-buffer` in `lsp-test--simulated-input`, fixing the `lsp--read-rename` test somehow.


nbfalcon:
`lsp-test-current-buffer-mode` still doesn't pass, but that seems unrelated.

yyoncho (approved):

PR #2351, opened 2020-11-22
Fix `lsp-test-current-buffer-mode'

nbfalcon:
`lsp-test-current-buffer-mode` was defined thrice for some reason; only
the first of those definitions actually passed. Fix the test, and with
it the entire test-suite, by removing the latter two ("make unix-test"
now passes).


nbfalcon:
The tests now pass, but the CI doesn't, because some dependencies ("spinner") can't be installed for some reason.

yyoncho (approved):

PR #2353, opened 2020-11-22
`lsp-rename': support defaultBehavior

### LINKED ISSUE #2352: `lsp-rename`: prepareRename: support LSP 3.16's { defaultBehaviour: true }
As of version 3.16, "textDocument/prepareRename" may return a [special object](https://microsoft.github.io/language-server-protocol/specifications/specification-3-16/#textDocument_prepareRename), `{defaultBehaviour: true}`. We should support that in `lsp--get-symbol-to-rename`.

@yyoncho how should I do the check here? Is `lsp-get` sufficient, or can we somehow generate protocol bindings?

nbfalcon:
- Add a rename capability noting that defaultBehavior is supported.
- Handle defaultBehavior in `lsp--get-symbol-to-rename` using `lsp-get`

Fixes #2352.

Note that I haven't tested this with a server that returns defaultBehaviour: true, but my previous `lsp--get-symbol-to-rename` tests still pass.

As stated in the issue, is `lsp-get` good enough here?


lsp-mode.el
```diff
@@ -5671,20 +5672,24 @@ while PLACEHOLDER?, is either nil or a string suggested by the
...
+                              (if (lsp-range? response)
+                                  response
+                                (lsp:prepare-rename-result-range response))))
+                     (placeholder
+                      (and (not (lsp-range? response))
+                           (lsp:prepare-rename-result-placeholder response))))
+                (cons bounds placeholder)))
+          ;; If prepareRename returned null, fail now and don't fall back to the
+          ;; symbol at point.
+          (cl-return nil)))
```
yyoncho, line 5690: after you have deleted the check in lsp-rename this should be `user-error` - unable to rename symbol or something like that. This is a regression from the previous PR.

lsp-mode.el
```diff
@@ -5671,20 +5672,24 @@ while PLACEHOLDER?, is either nil or a string suggested by the
...
+   (when-let ((bounds (bounds-of-thing-at-point 'symbol)))
```
yyoncho, line 5691: ditto, here it should be if-let with user error in the then clause.

yyoncho (commented):
> As stated in the issue, is lsp-get good enough here?

Better to add it to the lsp-interface and use the generated getter.

lsp-mode.el
nbfalcon: The `user-error` happens in `lsp--read-rename`, to which the result of this function is passed unmodified.

lsp-mode.el
nbfalcon: See above.

nbfalcon:
Should I add it manually, or use some sort of generation script?

yyoncho:
> Should I add an `lsp-interface` manually, or use some sort of generation script?

add it manually.

nbfalcon:
Implemented. The naming is quite awkward though - `PrepareRenameDefaultBehaviorResult`. Also, does the generated getter do type-checking, or will `lsp-interface` getters do so in the future? I'd like to support `{ defaultBehavior: false }` just in case, and as such decided against using `lsp:prepare-rename-default-behavior-result-p`.

yyoncho:
> Implemented. The naming is quite awkward though - `PrepareRenameDefaultBehaviorResult`.

It happens for some of the types, seems like the code can be refactored to use destructuring to avoid this naming 

> Also, does the generated getter do type-checking, or will `lsp-interface` getters do so in the future? I'd like to support `{ defaultBehavior: false }` just in case, and as such decided against using `lsp:prepare-rename-default-behavior-result-p`.

There is no type checking here, not sure even if it can be added without causing trouble.

lsp-protocol.el
```diff
@@ -290,6 +290,9 @@ See `-let' for a description of the destructuring mechanism."
                (rust-analyzer:CommandLink (:title :command) (:arguments :tooltip))
                (rust-analyzer:CommandLinkGroup (:commands) (:title)))
 
+(lsp-interface
+ (PrepareRenameDefaultBehaviorResult (:defaultBehavior) nil))
+
```
yyoncho, line 295: You should add it on line 596, not here.

nbfalcon:
@yyoncho it could be done really elegantly if `-unless-let` was a thing, but that sadly isn't the case. I don't see what else could be done.

lsp-protocol.el
nbfalcon: That is after ";;; begin automatically generated code". Wouldn't this be overwritten by some script?

lsp-mode.el
```diff
@@ -5662,7 +5663,7 @@ the symbol to rename at point."
   :group 'lsp-mode
   :type 'boolean)
 
-(defun lsp--get-symbol-to-rename ()
+(cl-defun lsp--get-symbol-to-rename ()
```
kiennq, line 5666: Why convert this to `cl-defun`? You're not adding named parameter here

nbfalcon: Because I use `cl-return`, and that needs an implicit `nil` block. I could also wrap this in `cl-block`, but that would cause extra-indentation.

nbfalcon:
I might have gotten a bit confused about `cl-return`.

lsp-mode.el
nbfalcon: Good catch, it didn't do what I thought.

lsp-protocol.el
yyoncho: No, this is not going to be regenerated anymore. We are supposed to add the new fields manually.

yyoncho:
> @yyoncho it could be done really elegantly if `-unless-let` was a thing, but that sadly isn't the case. I don't see what else could be done.

That should work too (or something like that). 

```
(if (and lsp-rename-use-prepare (lsp-feature? "textDocument/prepareRename"))
      (when-let ((response (lsp-request "textDocument/prepareRename"
                                        (lsp--text-document-position-params))))
        (if (lsp-prepare-rename-result? result)
            (-let [(&PrepareRenameResult :range :placeholder :default-behaviour) result]
              (if default-behaviour
                  (lsp--get-symbol-to-rename-default)
                (cons (lsp--range-to-region range) placeholder)))
          (cons (lsp--range-to-region result) nil)))
    (lsp--get-symbol-to-rename-default))
```

lsp-mode.el
kiennq: The `cl-block` for `cl-defun` has the name of the function while `cl-return` is returning from block `nil`.

nbfalcon: I just realized this too, which is why I made this PR draft.

nbfalcon: The control flow here will require further thought.

lsp-protocol.el
yyoncho: seems like I have misled you here. We should generate a new type, not alter the existing one. I thought that they are adding a new field but they are actually creating a new result type.

nbfalcon:
@yyoncho I also considered such an approach, but decided against it because it would add yet another function. I ended up going for `cl-return-from` with a correct block, since I couldn't come up with a good `cond`-based solution (I didn't try that hard though).

However, a different question is whether this feature is worth it at all, and whether we can always implement it reliably. `defaultBehavior` isn't very well specified: "the client should use its default behavior". language servers could probably always return this if renaming is possible, but that would cause trouble with more complex renames, e.g. lua/C labels, which include colons (I don't know of a language server that can rename labels in either of them, though). We'd have to implement some kind of `defaultBehavior` `lsp--get-symbol-to-rename` for every language, just to save a few bytes of traffic.

Feel free to close this PR if you don't consider it worth it.

# C label example
```c
int main() {
  label: goto label;
}
```

# Lua label example
```lua
function foo()
  ::label::
  goto label
end

**Comment by yyoncho** [comment]
> for every language, just to save a few bytes of traffic.

I haven't dug into the spec but is this all what :defaultBehaviour is for?

**Comment by nbfalcon** [comment]
@yyoncho I don't know, I'll need to look at the corresponding PR.

**Comment by nbfalcon** [comment]
The commit that introduced that feature in LSP is microsoft/language-server-protocol@bbcc5d088f572f312060c81d1bd1a39395368194. However, it's message doesn't clear things up. I'll need to make an issue.

File: lsp-protocol.el [hunk 5]
```diff
@@ -290,6 +290,9 @@ See `-let' for a description of the destructuring mechanism."
                (rust-analyzer:CommandLink (:title :command) (:arguments :tooltip))
                (rust-analyzer:CommandLinkGroup (:commands) (:title)))
 
+(lsp-interface
+ (PrepareRenameDefaultBehaviorResult (:defaultBehavior) nil))
```
kiennq, line 294: nit: can you group this with the `lsp-interface` block that contains `Edit` or group this with the auto-generated block instead?
Since we will not doing auto generating anymore.

lsp-mode.el
```diff
@@ -5671,20 +5672,25 @@ while PLACEHOLDER?, is either nil or a string suggested by the
...
-                (and (not (lsp-range? response))
-                     (lsp:prepare-rename-result-placeholder response))))
-          (cons bounds placeholder)))
-    (when-let ((bounds (bounds-of-thing-at-point 'symbol)))
-      (cons bounds nil))))
+  (or
+   (and lsp-rename-use-prepare (lsp-feature? "textDocument/prepareRename")
+        (if-let ((response (lsp-request "textDocument/prepareRename"
+                                        (lsp--text-document-position-params))))
+            (unless (lsp:prepare-rename-default-behavior-result-default-behavior
```
kiennq, line 5679: Why don't you just check type `lsp-prepare-rename-default-behavior-result?` instead?

kiennq:
Is it worth to support this feature? Or in another word, is that okay for Emacs to support this feature.
From what I see in the specs
>  If `{ defaultBehavior: boolean }` is returned (since 3.16) the rename position is valid and the client should use its default behavior to compute the rename range.

It seems that the server will not need to calculate the range and let the editor to determine the range instead.
That would save the server some process cycles (which is very small, given server has to check the range already), but the information from Emacs may not correct and largely depends on `major-mode`.
For example, in `java`, `@Override` is interpreted as symbol instead of just `Override`, `jdtls` may have different thinking about that bound.

We have similar problem with using just symbol for completion before and resorted in just using server information exclusively.

yyoncho:
I second @kiennq - we had a lot of trouble with the default behaviour in completion prefix calculation if there is a way to avoid that here I would prefer to go in that direction...

kiennq:
Just realize @nbfalcon has stated the similar thing already [here](https://github.com/emacs-lsp/lsp-mode/pull/2353#issuecomment-732788423).
@nbfalcon Do you think it's still worth the effort to provide support for this feature?

nbfalcon:
@kiennq we'll see: I have made an issue on microsoft/language-server-protocol asking for clarification around `defaultBehavior`. If there is indeed some well-defined obvious case, my PR might work or be made to work. In that case, I believe implementing as much as possible of the protocol is a good thing. (Of course not if that comes with lots of effort for no gain).

nbfalcon:
@yyoncho in the end I decided to go with your approach, of adding a new `defun` since after some thought it made more sense to me. However, if you don't want to duplicate review effort, I can revert (it is a single atomic commit).

Don't close this just yet, we should wait for my upstream issue to be answered.

nbfalcon:
@yyoncho well, VSCode's `defaultBehavior` is equivalent to `thing-at-point` `symbol`. However, assuming this would make `lsp-mode` more dependent on a correct major-mode (perhaps some day `lsp-mode` could be used with `fundamental-mode`?) This will require some more thought.

kiennq:
> well, VSCode's `defaultBehavior` is equivalent to `thing-at-point` `symbol`. However, assuming this would make `lsp-mode` more dependent on a correct major-mode (perhaps some day `lsp-mode` could be used with `fundamental-mode`?) This will require some more thought.

It would require `fundamental-mode` syntax table to be aligned perfectly with VSCode. I doubt that is desired or correct. For example, lisp treat `-` as symbol whereas javascript doesn't.
The `thing-at-point` `symbol` depends on syntax table, which can change quite drastically between `major-mode` and server knowledge.

nbfalcon:
@kiennq VSCode uses a language-specific `thing-at-point` for that. It won't work with `fundamental-mode`, which is the issue currently.

PR #2357, opened 2020-11-24
change lsp-yaml-schemas to work with native json-serialize

dellison:
Hey `lsp-mode` maintainers! I was upgrading my emacs to 27.1 (finally) when I came across when I think may be a pretty tricky bug with the way the YAML backend interacts with [`lsp--json-serialize`](https://github.com/emacs-lsp/lsp-mode/blob/182cac1f5853c060ff78459a5dbfaf2e552c0d20/lsp-mode.el#L2805). This macro checks if the native `json-serialize` is available, uses it if so, and otherwise uses the elisp `json-encode`. The issue here, from what I can tell, is that `json-serialize` and `json-encode` have somewhat different capabilities when it comes to serializing an alist to a JSON string.

Here is sort of a MWE of this, inspired by `lsp--json-serialize`:

```elisp
(json-serialize '((:kubernetes "k8s.yaml"))
		:null-object nil
		:false-object :json-false)
;; Lisp error: (wrong-type-argument consp nil)
```

```elisp
(let ((json-false :json-false))
  (json-encode '((:kubernetes "k8s.yaml"))))
;; => "{\"kubernetes\":[\"k8s.yaml\"]}"
```

Changing the alist `'((:kubernetes "k8s.yaml"))` to a plist `'(:kubernetes "k8s.yaml")` seems to work in both cases:

```elisp
(json-serialize '(:kubernetes "k8s.yaml")
		:null-object nil
		:false-object :json-false)
;; => "{\"kubernetes\":\"k8s.yaml\"}"
```
```elisp
(let ((json-false :json-false))
  (json-encode '(:kubernetes "k8s.yaml")))
;; => "{\"kubernetes\":\"k8s.yaml\"}"
```

So that's what this PR does, basically. As far as I can tell, this commit fixes this issue for both situations- with and without native support for json via `json-serialize`. And as a bonus, I think this is more in line with the format the [yaml language server](https://github.com/redhat-developer/yaml-language-server#associating-a-schema-to-a-glob-pattern-via-yamlschemas) expects for the `yaml.schemas` format (thought it seemed okay with the glob as a list of a single string, so maybe the yaml backend is a little lax about that). But in any case, I'm quite new to lsp-mode, so I would really appreciate a close look at this, if it's possible. Thanks so much! :)

### DIFF
--- a/clients/lsp-yaml.el
+++ b/clients/lsp-yaml.el
@@ -27,8 +27,6 @@
 (require 'lsp-mode)
 (require 'dash)
 
-(declare-function assoc-delete-all "subr" (key alist &optional test))
-
 (defgroup lsp-yaml nil
   "LSP support for YAML, using yaml-language-server."
   :group 'lsp-mode
@@ -203,9 +201,9 @@ Set FORCE-DOWNLOADING to non-nil to force re-download the database."
             (setq lsp-yaml-schemas
                   (cl-acons uri
                             (vconcat (vector glob) current-patterns)
-                            (assoc-delete-all uri
-                                       (mapcar (lambda (x) (lsp-yaml--remove-glob x glob))
-                                               lsp-yaml-schemas)))))
+                            (assq-delete-all uri
+                                             (mapcar (lambda (x) (lsp-yaml--remove-glob x glob))
+                                                     lsp-yaml-schemas)))))
       (setq lsp-yaml-schemas
             (cl-acons uri (vector glob) (mapcar (lambda (x) (lsp-yaml--remove-glob x glob))
                                                 lsp-yaml-schemas))))
--- a/clients/lsp-yaml.el
+++ b/clients/lsp-yaml.el
@@ -25,6 +25,11 @@
 ;;; Code:
 
 (require 'lsp-mode)
+(require 'seq)
+
+(declare-function assoc-delete-all "subr" (key alist &optional test))
+(declare-function seq-contains-p"seq" (sequence elt &optional testfn))
+(declare-function seq-filter "seq" (pred sequence))
 
 (defgroup lsp-yaml nil
   "LSP support for YAML, using yaml-language-server."
@@ -88,7 +93,7 @@
 
 (defcustom lsp-yaml-schemas '()
   "Associate schemas to YAML files in a glob pattern."
-  :type '(alist :key-type (string :tag "schema") :value-type (string :tag "files (glob)"))
+  :type '(alist :key-type (symbol :tag "schema") :value-type (vector :tag "files (glob)"))
   :group 'lsp-yaml
   :package-version '(lsp-mode . "6.2"))
 
@@ -164,7 +169,7 @@
 (defconst lsp-yaml--built-in-kubernetes-schema
   '((name . "Kubernetes")
     (description . "Built-in kubernetes manifest schema definition")
-    (url . :kubernetes)
+    (url . "kubernetes")
     (fileMatch . ["*-k8s.yaml" "*-k8s.yml"])))
 
 (defun lsp-yaml-download-schema-store-db (&optional force-downloading)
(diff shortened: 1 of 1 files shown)


yyoncho:
> ```emacs-lisp
> (json-serialize '((:kubernetes "k8s.yaml"))
> 		:null-object nil
> 		:false-object :json-false)
> ;; Lisp error: (wrong-type-argument consp nil)
> ```

I have to spend a bit more time with this PR but `'((:kubernetes "k8s.yaml"))` is not alist - when you change it to `'((:kubernetes . "k8s.yaml"))` it seems to be working fine.

dellison:
> '((:kubernetes "k8s.yaml")) is not alist

Sorry, yes, you're right- that's not an alist (I think it's intended to be one). But on master, that's what `lsp-yaml-schemas` gets set to for `M-x lsp-yaml-select-buffer-schema` and choosing the built-in kubernetes schema, and this results in the error. Hopefully that makes sense.

yyoncho:
> > '((:kubernetes "k8s.yaml")) is not alist
> 
> Sorry, yes, you're right- that's not an alist (I think it's intended to be one). But on master, that's what `lsp-yaml-schemas` gets set to for `M-x lsp-yaml-select-buffer-schema` and choosing the built-in kubernetes schema, and this results in the error. Hopefully that makes sense.

I got it. Can we fix the adding of the schema to work properly and add `cons` instead of the list? I prefer this fix over the changing of the type of `lsp-yaml-schemas` since it is not a breaking change.

dellison:
Okay, I'll see if I can get it working that way and update the branch if so.

dellison:
Okay, I've updated this with an implementation that seems to be working well for both native and elisp json serialization. I couldn't *quite* make it completely non-backwards-compatible- the trouble there is that the native `json-serialize` can't handle alists with non-symbol keys. So now, cons cells are used correctly to associate the uri of the schema (now a symbol) with the glob pattern for the yaml files (a string).

Additionally, now this removes any other url-to-glob mappings that associate the current file with any other schemas, which seems necessary for the yaml-language-server to use the correct schema.

Thanks again for taking a look at this! Let me know what you think of this version. :)

clients/lsp-yaml.el
```diff
@@ -185,22 +185,27 @@ Set FORCE-DOWNLOADING to non-nil to force re-download the database."
...
   "Set yaml schema for the current buffer to URI."
   (interactive "MURI: ")
-  (let* ((workspace-path (file-relative-name
+  (let* ((uri (intern uri-string))
+         (workspace-path (file-relative-name
                           (lsp--uri-to-path (lsp--buffer-uri))
                           (lsp-workspace-root (lsp--buffer-uri))))
-         (glob-pattern (concat "/" workspace-path))
+         (glob (concat "/" workspace-path))
          (current-config (cl-assoc uri lsp-yaml-schemas :test 'equal))
```
kiennq, line 196: nit: using `assoc` is better since it's C primitive
```suggestion
         (current-config (assoc uri lsp-yaml-schemas))
```

dellison: nice, thanks

clients/lsp-yaml.el
```diff
@@ -25,6 +25,11 @@
 ;;; Code:
 
 (require 'lsp-mode)
+(require 'seq)
+
+(declare-function assoc-delete-all "subr" (key alist &optional test))
+(declare-function seq-contains-p"seq" (sequence elt &optional testfn))
+(declare-function seq-filter "seq" (pred sequence))
```
yyoncho, line 32: this shouldn't be needed.

yyoncho (commented):
Looks good, I will merge after doing manual testing.

clients/lsp-yaml.el
dellison: I was getting some errors from the byte compiler in the CI without these.

yyoncho: This means that these function versions are not present in the seq version we depend on. We should either try to bump seq or replace them with dash versions

dellison: Oh, and another thing I found doing some manual tests myself was that `seq-contains-p` is new in emacs 27, so I've added one more commit to avoid using it. I guess `seq-contains-p` was deprecated, so I thought I'd try to avoid any deprecation warnings in emacs 26 and use `member`. You have to covert the vector to a list to do that, which feels kind of awkward to me so I'm open suggestions.

dellison:
Just to document what I ran into, since it doesn't exactly match what I wrote in earlier comments- I found that the only thing that worked in all the cases I tried (emacs 26, emacs 27 with native json, emacs 27 without json) the data structure that serializes everywhere to a format that the YAML language server expects is something like:

```elisp
((https://json\.schemastore\.org/gitlab-ci . ["/.gitlab-ci.yml"])
 (kubernetes . ["/elasticsearch.yaml" "/nginx.yaml"]))
```

So, an alist with symbols for keys (the schema) and values a vector of strings, which are the globs that indicate files.

clients/lsp-yaml.el
dellison: Regarding `seq-filter`, I've updated it to `-filter` from dash. I'm not really sure about `assoc-delete-all`- it looks like it got an optional `test` argment in emacs 27, but that isn't used here.

dellison: In case it helps, here is a link to a failing CI job that removes the declare-function for `assoc-delete-all`: https://github.com/dellison/lsp-mode/actions/runs/386015329. I'm not sure why that happens.

yyoncho (approved):

yyoncho:

clients/lsp-yaml.el
dellison: Updated this to use `assq-delete-all` instead of `assoc-delete-all`. Apparently `assoc-delete-all` has a funny history, which probably explains why this was so confusing to me.
- In Emacs 21.1, [assq-delete-all replaces function assoc-delete-all](https://github.com/emacs-mirror/emacs/blob/d0e2a341dd9a9a365fd311748df024ecb25b70ec/etc/NEWS.21#L2363)
- Later, Emacs 26.2 [introduces the new function assoc-delete-all](https://github.com/emacs-mirror/emacs/blob/master/etc/NEWS.26#L150)

My own emacs was telling me that it was "Probably introduced at or before Emacs version 21.1." in the docstring for `assoc-delete-all`. Pretty funny.

dellison:
Thank you for the review, and for your great work on lsp-mode! :)

PR #2358, opened 2020-11-24
lsp-mode.el: document usage of custom client settings

Hi-Angel:
Per discussion a dozen of minutes ago: it's unclear that `lsp-client-settings` should not be used directly, nor it is clear that using `lsp-register-custom-settings` requires double parentheses. Let's add some docs on that

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7311,7 +7311,9 @@ changing the value of `foo'."
               tail (cdr (cdr tail))))))
     plist))
 
-(defvar lsp-client-settings nil)
+(defvar lsp-client-settings nil
+  "For internal use, any external users please use
+  `lsp-register-custom-settings' function instead")
 
 (defun lsp--compare-setting-path (a b)
   (equal (car a) (car b)))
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7322,7 +7324,10 @@ PROPS is list of triple (path value boolean?) where PATH is the path to the
 property; VALUE can be a literal value, symbol to be evaluated, or either a
 function or lambda function to be called without arguments; BOOLEAN? is an
 optional flag that should be non-nil for boolean settings, when it is nil the
-property will be ignored if the VALUE is nil."
+property will be ignored if the VALUE is nil.
+
+Example: `(lsp-register-custom-settings '((\"foo.bar.buzz.enabled\" t t)))'
+(note the double parentheses)"
```
kiennq, line 7330: nit: We've already said this is list of triplet `(path value boolean?)`, so the note here may not needed

kiennq (approved):

lsp-mode.el
Hi-Angel: Actually, this text, quoting `PROPS is list of triple (path value boolean?)` is the exact reason I was confused and couldn't make it work till I saw in the chat people provided examples. I read this text as: "PROPS is a list of triple, which you would write as (path value boolean?)".

yyoncho (approved):

PR #2361, opened 2020-11-25
update implementation of lsp-yaml-download-schema-store-db

dellison:
Trying to download the yaml schemas when the `~/.emacs.d/.cache/lsp` directory doesn't exist gets an error from `url-copy-file`. Also, it seems like there is an error in the logic of the `force-downloading` arg: if it's non-nil, the cache is never downloaded, even if the file doesn't exist locally. (The yaml client code doesn't use this argument, so currently the only way to get this behavior to to call `(lsp-yaml-download-schema-store-db t)` manually).

This PR fixes these two things: it creates the cache directory first if it doesn't already exist before downloading the schema db file, and updates the implementation to do match what the docstring describes.

### DIFF
--- a/clients/lsp-yaml.el
+++ b/clients/lsp-yaml.el
@@ -171,18 +171,18 @@
   "Download the remote schema store at `lsp-yaml-schema-store-uri' into local cache.
 Set FORCE-DOWNLOADING to non-nil to force re-download the database."
   (interactive "P")
-  (or force-downloading
-      (file-exists-p lsp-yaml-schema-store-local-db)
-      (url-copy-file lsp-yaml-schema-store-uri lsp-yaml-schema-store-local-db)))
+  (when (or force-downloading (not (file-exists-p lsp-yaml-schema-store-local-db)))
+    (unless (file-directory-p (file-name-directory lsp-yaml-schema-store-local-db))
+      (mkdir (file-name-directory lsp-yaml-schema-store-local-db)))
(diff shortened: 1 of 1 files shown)


clients/lsp-yaml.el
```diff
@@ -171,18 +171,18 @@
   "Download the remote schema store at `lsp-yaml-schema-store-uri' into local cache.
 Set FORCE-DOWNLOADING to non-nil to force re-download the database."
   (interactive "P")
-  (or force-downloading
-      (file-exists-p lsp-yaml-schema-store-local-db)
-      (url-copy-file lsp-yaml-schema-store-uri lsp-yaml-schema-store-local-db)))
+  (when (or force-downloading (not (file-exists-p lsp-yaml-schema-store-local-db)))
+    (if (not (file-directory-p (file-name-directory lsp-yaml-schema-store-local-db)))
```
jcs090218, line 175: I guess you can use `unless` instead of `if` and `not` here.

```suggestion
    (unless (file-directory-p (file-name-directory lsp-yaml-schema-store-local-db))
```

jcs090218: Can you help us fix the indentation the line below? Thanks!

dellison: Updated! Thanks.

jcs090218 (approved):

yyoncho (approved):

yyoncho:

PR #2367, opened 2020-11-28
ESlint server shouldn't launch for JSON files

artemkovalyov:
JSON has its own server. ESlint server produces warnings and I'm not sure if it generally works with pure JSON.
Also, the resolution should be the EOL by default for non-temporary files.

### DIFF
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -267,7 +267,7 @@ source.fixAll code action."
   :activation-fn (lambda (filename &optional _)
                    (when lsp-eslint-enable
                      (or (string-match-p (rx (one-or-more anything) "."
-                                             (or "ts" "js" "jsx" "tsx" "html" "vue")eol)
+                                             (or "ts" "js" "jsx" "tsx" "html" "vue")eos)
                                          filename)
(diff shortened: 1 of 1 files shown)


clients/lsp-eslint.el
```diff
@@ -267,7 +267,7 @@ source.fixAll code action."
   :activation-fn (lambda (filename &optional _)
                    (when lsp-eslint-enable
                      (or (string-match-p (rx (one-or-more anything) "."
-                                             (or "ts" "js" "jsx" "tsx" "html" "vue"))
+                                             (or "ts" "js" "jsx" "tsx" "html" "vue")eol)
```
yyoncho, line 270: can you use `eos` instead of `eol`?

artemkovalyov, line 270:
```suggestion
                                             (or "ts" "js" "jsx" "tsx" "html" "vue")eos)
```

artemkovalyov: Indeed, it's my first encounter with `rx`, didn't know `eos` exists.

yyoncho (approved):

yyoncho:
Thank you!

PR #2372, opened 2020-11-29
Extend clangd with switch to header/source

petr-tik:
Address #2279 

Made an interactive function to switch to header/source.

Add a cmake-based repo for intergration tests of clangd. Currently Linux-only, since I don't have a Windows machine at home. 

I am not sure if async functions that fail inside deferred::nextc show up as errors in CI logs. Even if followed by deferred:sync! 

Follow up to #2279, which I closed to sort out the CI without spamming the upstream and reviewers.

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -242,8 +242,8 @@ returned to avoid that the echo area grows uncomfortably."
          (lsp-cpp-flycheck-clang-tidy-error-explainer e))
         (t (flycheck-error-message e))))
 
-(defun lsp-clangd-to-other (&optional arg)
-  "Open the corresponding header/source file.
+(defun lsp-clangd-find-other-file (&optional arg)
+  "Switch between the corresponding C/C++ header and source file.
 
 If called with ARG universal argument, the file will open in the other window."
   (interactive "P")
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -242,5 +242,21 @@ returned to avoid that the echo area grows uncomfortably."
          (lsp-cpp-flycheck-clang-tidy-error-explainer e))
         (t (flycheck-error-message e))))
 
+(defun lsp-clangd-to-other ()
+  "Open the corresponding header/source file."
+  (if-let ((other-fname
+            (with-lsp-workspace (lsp-find-workspace 'clangd)
+              ;; TODO define an extension with lsp-interace
+              ;; similar to rust-analyzer on lines 278-292 in
+              ;; lsp-protocol.el
+              (lsp-send-request (lsp-make-request
+                                 "textDocument/switchSourceHeader"
+                                 (lsp--text-document-identifier))))
+            ))
+      ;; TODO use lsp-goto-location instead
+      (find-file (lsp--uri-to-path other-fname))
+    (lsp--info "This file doesn't have a corresponding file")
+    nil))
+
 (provide 'lsp-clangd)
 ;;; lsp-clangd.el ends here
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -242,18 +242,20 @@ returned to avoid that the echo area grows uncomfortably."
          (lsp-cpp-flycheck-clang-tidy-error-explainer e))
         (t (flycheck-error-message e))))
 
-(defun lsp-clangd-find-other-file (&optional arg)
-  "Switch between the corresponding C/C++ header and source file.
+(defun lsp-clangd-find-other-file (&optional new-window)
+  "Switch between the corresponding C/C++ source and header file.
+If NEW-WINDOW (interactively the prefix argument) is non-nil,
+open in a new window.
 
-If called with ARG universal argument, the file will open in the other window."
+Only works with clangd."
   (interactive "P")
-  (if-let ((other-fname (lsp-send-request (lsp-make-request
-                                           "textDocument/switchSourceHeader"
-                                           (lsp--text-document-identifier)))))
-      (if arg
-          (find-file-other-window (lsp--uri-to-path other-fname))
-        (find-file (lsp--uri-to-path other-fname)))
-    (lsp--info "This file doesn't have a corresponding file")))
+  (let ((other (lsp-send-request (lsp-make-request
+                                  "textDocument/switchSourceHeader"
+                                  (lsp--text-document-identifier)))))
+    (unless (s-present? other)
+      (user-error "Could not find other file"))
+    (funcall (if new-window #'find-file-other-window #'find-file)
+             (lsp--uri-to-path other))))
 
 (provide 'lsp-clangd)
 ;;; lsp-clangd.el ends here
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -242,21 +242,18 @@ returned to avoid that the echo area grows uncomfortably."
          (lsp-cpp-flycheck-clang-tidy-error-explainer e))
         (t (flycheck-error-message e))))
 
-(defun lsp-clangd-to-other ()
-  "Open the corresponding header/source file."
-  (if-let ((other-fname
-            (with-lsp-workspace (lsp-find-workspace 'clangd)
-              ;; TODO define an extension with lsp-interace
-              ;; similar to rust-analyzer on lines 278-292 in
-              ;; lsp-protocol.el
-              (lsp-send-request (lsp-make-request
-                                 "textDocument/switchSourceHeader"
-                                 (lsp--text-document-identifier))))
-            ))
-      ;; TODO use lsp-goto-location instead
-      (find-file (lsp--uri-to-path other-fname))
-    (lsp--info "This file doesn't have a corresponding file")
-    nil))
+(defun lsp-clangd-to-other (&optional arg)
+  "Open the corresponding header/source file.
+
+If called with ARG universal argument, the file will open in the other window."
+  (interactive "P")
+  (if-let ((other-fname (lsp-send-request (lsp-make-request
+                                           "textDocument/switchSourceHeader"
+                                           (lsp--text-document-identifier)))))
+      (if arg
+          (find-file-other-window (lsp--uri-to-path other-fname))
+        (find-file (lsp--uri-to-path other-fname)))
+    (lsp--info "This file doesn't have a corresponding file")))
 
 (provide 'lsp-clangd)
 ;;; lsp-clangd.el ends here
--- a/test/lsp-clangd-test.el
+++ b/test/lsp-clangd-test.el
@@ -23,6 +23,7 @@
 
 (require 'ert)
 (require 'lsp-clangd)
+(require 'lsp-integration-test)
 
 (ert-deftest lsp-clangd-extract-signature-on-hover ()
   (should (string= (lsp-clients-extract-signature-on-hover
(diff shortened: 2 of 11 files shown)


petr-tik:
I don't really understand why the windows ci run is failing. I don't think it's my clangd integration tests, but i don't think we even reach that point.

petr-tik:
Can you please review again? @nbfalcon @yyoncho still don't understand why the windows CI is failing.

yyoncho:
I am fine with the code. 

> still don't understand why the windows CI is failing.

My guess is that the current directory is not on the load-path thus it cannot find `lsp-integration-test.el`, it can be added explicitly to , I don't have an explanation why. You may try to add it explicitly to the load-path variable.

petr-tik:
> My guess is that the current directory is not on the load-path thus it cannot find `lsp-integration-test.el`. I don't have an explanation why. You may try to add it explicitly to the load-path variable.

I think it's because of the alphabetical order. test/lsp-clangd-test.el is loaded before test/lsp-integration-test.el, so when i `(require 'lsp-integration-test)` in lsp-clangd-test.el, this file hadn't been loaded yet.

petr-tik:
But then again, it somehow works in linux, so I am not sure

petr-tik:
> I am fine with the code.

If that's still the case - please take a butcher's and feel free to squash and merge. 

Thanks for guiding me all along, appreciate it was longer than expected, but I now have more context to contribute other features.

nbfalcon:
I think the windows CI being broken isn't your fault - the CI uses a command that was disabled for security resasons and wasn't updated yet.

clients/lsp-clangd.el
```diff
@@ -237,7 +237,10 @@ returned to avoid that the echo area grows uncomfortably."
 
 (defun lsp-clangd-to-other ()
```
nbfalcon: To be consistent with `projectile-find-other-file`, this should be named `lsp-clangd-find-other-file`.

clients/lsp-clangd.el
```diff
@@ -246,7 +249,7 @@ returned to avoid that the echo area grows uncomfortably."
                                  "textDocument/switchSourceHeader"
                                  (lsp--text-document-identifier))))
             ))
-      ;; TODO use lsp-goto-location instead
+      ;; TODO maybe worth using lsp-goto-location instead
```
nbfalcon: You cannot and shouldn't use `lsp-goto-location`, because that expects an `LspLocation` with line:col information, which is not what `clangd` returns here (it would also not make sense for it to do that).

nbfalcon (changes requested):

clients/lsp-clangd.el
petr-tik: Thank for addressing my TODOs. I will delete the comment.

nbfalcon: Are there still any ToDos remaining?

clients/lsp-clangd.el
petr-tik: Makes sense. Is consistency with projectile something that I should generally keep in mind? What about project.el that seems to be gaining feature parity?

nbfalcon:
If @yyoncho is ok with that (don't do it yet), can you please squash your changes using interactive-rebase? There are currently too many commits.

nbfalcon:
Things like `Wrote a free standing function` and the following two commits should be one. You are committing too often, IMHO.

nbfalcon:
Your debugging commits like "try tmate" followed by "comment tmate out" should be gone as well.

clients/lsp-clangd.el
nbfalcon: That is not something you need to keep in mind especially; we don't depend on `projectile`. This is just about having naming that is consistent with the rest of Emacs in general.

nbfalcon:
I'd like to propose the following (tested) alternative implementation:
```
(defun lsp-clangd-find-other-file ()
  "Switch between the corresponding C/C++ source and header file.
Only works with `clangd'."
  (interactive)
  (let ((other (lsp-send-request (lsp-make-request
                                  "textDocument/switchSourceHeader"
                                  (lsp--text-document-identifier)))))
    (unless (s-present? other))
      (user-error "Could not find other file"))
    (find-file (lsp--uri-to-path other))))
```
My function has the following advantages:
- It doesn't break down if users have a custom `clangd` server id for tramp
- It `user-error`s instead of calling `lsp--info`, which gives a better hint that operation failed. 

It should be clear from both the naming and the docstring that this function cannot be used with anything but `clangd`, which is why I don't consider the clangd workspace check necessary. Servers that won't support this will error intuitively (IMHO):
- `jdtls`: "Unsupported request method: textDocument/switchSourceHeader"
- `pyright`: "Unhandled method textDocument/switchSourceHeader"

nbfalcon:
@yyoncho do we prefer `lsp--info`, `lsp--error`, `message`, or `user-error`?

test/lsp-clangd-test.el
```diff
@@ -102,6 +102,7 @@ and starts lsp. After the test BODY runs - tidy up."
      (setq lsp-clients-clangd-args actual-clangd-args)))
 
 (ert-deftest lsp-clangd-initialised-workspace ()
+  :tags '(no-win)
```
nbfalcon, line 105: Can you not do that? It should work on all platforms, so we should check on windows, too. I can have a look at the CI logs.

nbfalcon (changes requested):

clients/lsp-clangd.el
yyoncho: Generally, for the extension methods, we follow the name of the server method. In this case, something like lsp-clangd-switch-source-header should do.

yyoncho:
> @yyoncho do we prefer `lsp--info`, `lsp--error`, `message`, or `user-error`?

For that particular use case, I would pick `user-error`. In general, I am less picky about non-core files.

yyoncho (approved):
Looks good to me! If @nbfalcon is fine we could merge it is as it is, to get the functionality in and then address the non-critical issues separately.

petr-tik:
I can address the code review, function rename and rebase tidy up in the next couple of hours. Not sure about running clangd tests on windows CI though, as it’s been flaky for me and I don’t have a windows machine to test locally. 

I can add windows ci separately

yyoncho:
> Not sure about running clangd tests on windows CI though, as it’s been flaky for me and I don’t have a windows machine to test locally.

Ideally, better to have these green but IMHO if you feel that you are spending too much time on it I am fine to not have these tests. After all, we don't that many resources to invest them in relatively low priority issues.

clients/lsp-clangd.el
nbfalcon: @yyoncho but in this case I'd argue that there is benefit to be had from being consistent with the `projectile` naming scheme:
1. That name is good and intuitive (IMHO)
2. More familiarity: users of `projectile` could discover this function more easily trough `M-x`

yyoncho:
> > Not sure about running clangd tests on windows CI though, as it’s been flaky for me and I don’t have a windows machine to test locally.
> 
> Ideally, better to have these green but IMHO if you feel that you are spending too much time on it I am fine to not have these tests. After all, we don't that many resources to invest them in relatively low priority issues.

.. especially given the fact that we dont have tests at all, so adding them only for Linux is still a huge gain.

nbfalcon:
I might be able to test on windows.

petr-tik:
> I'd like to propose the following (tested) alternative implementation:
> 
> ```emacs-lisp
> (defun lsp-clangd-find-other-file ()
>   "Switch between the corresponding C/C++ source and header file.
> Only works with `clangd'."
>   (interactive)
>   (let ((other (lsp-send-request (lsp-make-request
>                                   "textDocument/switchSourceHeader"
>                                   (lsp--text-document-identifier)))))
>     (unless (s-present? other))
>       (user-error "Could not find other file"))
>     (find-file (lsp--uri-to-path other))))
> ```

I was reviewing what you proposed vs what I have. I would quite like to keep the `(interactive "P") to open the file in another buffer. @yyoncho please be the tie breaker

petr-tik:
> I might be able to test on windows.

Are you if we merge my PR without windows tests and you can add them later? To be honest, I don't really want to drag this relatively small change out any longer and think some tests is better than no tests.

nbfalcon:
Here is a new version with prefix argument support:
```emacs-lisp
(defun lsp-clangd-find-other-file (new-window)
  "Switch between the corresponding C/C++ source and header file.
If NEW-WINDOW (interactively the prefix argument) is non-nil,
open the other file in a new window.

Only works with `clangd'."
  (interactive "P")
  (let ((other (lsp-send-request (lsp-make-request
                                  "textDocument/switchSourceHeader"
                                  (lsp--text-document-identifier)))))
    (unless (s-present? other)
      (user-error "Could not find other file"))
    (funcall (if new-window #'find-file-other-window #'find-file)
             (lsp--uri-to-path other))))
```

nbfalcon:
I looked at the code, and I'd like you to do the following before it can be merged:
- [ ] use my revised method instead, for the reasons stated above
- [ ] create a new lsp-test-utils file, and move the common code there; in addition, refactor `lsp-with-pyls` to be generic, e.g. `lsp-with-workspace`, and define `lsp-clangd` and `lsp-with-pyls` in terms of the former
- [ ] (*optional*) make tests work on windows
- [ ] (optional) interactive-rebase
  - [ ] remove debugging commits
  - [ ] squash the free-standing implementation until implementation commits
  - [ ] squash the text fixture commits

nbfalcon:
Note that my new `lsp-clangd-find-other-file` method `user-error`s; this means that you'll need to wrap calls to it for cases where the call should not switch the file in `should-error`, as otherwise the tests would fail.

petr-tik:
>     * create a new lsp-test-utils file, and move the common code there; in addition, refactor `lsp-with-pyls` to be generic, e.g. `lsp-with-workspace`, and define `lsp-clangd` and `lsp-with-pyls` in terms of the former

Can you please create a separate issue for that and assign to me? I am not too proficient with elisp macros and would need a bit of time to work out how to make it generic over clangd and pyls considering they have different setup routines, which will drag this PR out, while extending outside the scope of the original ticket.

petr-tik:
I have rebased to leave only 4 commits. Overall, I am pretty keen to have this merged and I will happily tidy up the integration testing code and get it to work for windows in another PR.

petr-tik:
ok, for some reason my clangd tests aren't running anymore - pretty sure it's the tags in make unix-test

petr-tik:
Sigh. I guess I will need to make it windows compatible after all.

test/lsp-clangd-test.el
```diff
@@ -147,7 +144,7 @@ and starts lsp. After the test BODY runs - tidy up."
       ;; after opening a new buffer - need to make sure we are in the same lsp session
       (lsp)
       (sleep-for 1)
-      (lsp-clangd-find-other-file)
+      (should-error (lsp-clangd-find-other-file nil) :type 'user-error)
```
nbfalcon, line 147:
```suggestion
      (should-error (lsp-clangd-find-other-file) :type 'user-error)
```

test/lsp-clangd-test.el
```diff
@@ -112,30 +111,28 @@ and starts lsp. After the test BODY runs - tidy up."
...
-  :tags '(no-win)
   (lsp-in-sample-cpp-project
    (->
     (lsp-test-wait (eq 'initialized
                        (lsp--workspace-status (cl-first (lsp-workspaces)))))
 
     (deferred::nextc
       (should (string= (buffer-name) "main.cpp"))
-      (lsp-clangd-find-other-file)
+      (lsp-clangd-find-other-file nil)
```
nbfalcon, line 121:
```suggestion
      (lsp-clangd-find-other-file)
```

test/lsp-clangd-test.el
```diff
@@ -112,30 +111,28 @@ and starts lsp. After the test BODY runs - tidy up."
...
       (should (string= (buffer-name) "main.h"))
       (lsp)
       ;; give enough time for clangd to respond for main.h
       ;; otherwise will get errors like
       ;; The connected server(s) does not support
       ;; method textDocument/switchSourceHeader.
       (sleep-for 1)
-      (lsp-clangd-find-other-file)
+      (lsp-clangd-find-other-file nil)
```
nbfalcon, line 129:
```suggestion
      (lsp-clangd-find-other-file)
```

clients/lsp-clangd.el
```diff
@@ -234,19 +234,20 @@ returned to avoid that the echo area grows uncomfortably."
                                        :value (lsp-clangd-join-region (point-min) (point-max)))))
         (car (s-lines (lsp--render-element contents)))))))
 
+(defun lsp-clangd-find-other-file (new-window)
```
nbfalcon, line 237:
```suggestion
(defun lsp-clangd-find-other-file (&optional new-window)
```
Sorry, new-window should be &optional.

nbfalcon (changes requested):
NEW-WINDOW should be optional.

nbfalcon:
If you address my latest review, we can merge. This was a mistake on my part, sorry.

No windows support is good enough for now; I can look into that and do the macro refactoring in a later PR if you don't want to/can't do that.

Thank you for contributing!

petr-tik:
> No windows support is good enough for now; I can look into that and do the macro refactoring in a later PR if you don't want to/can't do that.

Ok, I am back to fixing windows tests. 

> Thank you for contributing!

Thank you for helping me with this PR. It's been quite arduous for you - i appreciate that

nbfalcon:
LGMT. I have tested this manually (`lsp-clangd-find-other-file` works) and the tests pass locally on my machine. This just needs to be rebased against master.

nbfalcon (approved):

petr-tik:
@nbfalcon excuse the delay - i came back to this, rebased on master (looks like you added macos tests - nice!) and it looks fine bar the windows test fails, which I thought you were ok with? Please let me know what I should do

nbfalcon:
Great! Though note that it wasn't me who has added the `macos` tests, I think it was @jcs090218. I can't merge this with failing tests, so @yyoncho will need to do that.

yyoncho:
> Great! Though note that it wasn't me who has added the `macos` tests, I think it was @jcs090218. I can't merge this with failing tests, so @yyoncho will need to do that.

Are we going to merge with broken tests? If you are planning to fix them, can we comment on them?

kiennq:
I'm not a fan of broken tests since it will make us not sure if new changes will broken the package further.
Can we at least has the `system-type` check so the tests for these new functionalities will run only when is not in `windows-nt` or `darwin`?

nbfalcon:
I agree, broken CI's don't seem like a good idea. Perhaps something like `vagrant` might help with local cross-platform testing? Now the tests can be fixed by wrapping them in `(unless (memq system-type '(windows-nt darwin)))`.

petr-tik:
skipping clangd integration tests unless on linux atm. macos seems to fail to install dependencies.

kiennq (approved):

yyoncho (approved):

yyoncho:
Thank you!

PR #2379, opened 2020-12-05
[Documentation] Relocated file

mtekman:
lsp-client.el no longer exists and the equivalent function is in lsp-mode.el


yyoncho (approved):

yyoncho:
Thank you!

Issue #2049, opened 2020-08-11
"LSP :: message" hides minibuffer prompt in the echo area


JohnC32:
lsp--info (and I assume lsp--warn, lsp--error) are called asynchronously. Suppose you

    C-x C-f somefile.cpp

and are using the ccls language server. If it's a large project, it will take a while for the ccls project to startup. When things are running, we see messages like "LSP :: ccls:197652 initialized successfully" displayed in the echo area. These displays take a bit. After opening somefile.cpp, you quickly try opening another file by typing:

    C-x C-f 

and then pause typing for a bit in the minibuffer, the minibuffer prompt will get overwritten with the "LSP :: ...." messages, yielding a little confusion because the minibuffer prompt is gone.

I'm not sure the best way to solve this. A coworker suggested looking at https://sourceforge.net/p/cedet/git/ci/master/tree/lisp/cedet/fame.el which has options for logging messages that will go away.

Using minibuffer-depth also seems to work to avoid showing the messages when the minibuffer is active:

``` emacs-lisp
(defun lsp--info (format &rest args)
  "Display lsp info message with FORMAT with ARGS."
  (when (= (minibuffer-depth) 0)
    (message "%s :: %s" (propertize "LSP" 'face 'success) (apply #'format format args))))
```

PR #2383, opened 2020-12-06
lsp--info/lsp--warn, don't overwrite prompts in the echo area

JohnC32:
This is a solution to https://github.com/emacs-lsp/lsp-mode/issues/2049

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1181,11 +1181,15 @@ Deprecated. Use `lsp-repeatable-vector' instead. "
 
 (defun lsp--info (format &rest args)
   "Display lsp info message with FORMAT with ARGS."
-  (message "%s :: %s" (propertize "LSP" 'face 'success) (apply #'format format args)))
+  ;; Display the message when the echo area isn't in use. Consider the case of find-file active and
+  ;; then lsp wants to display a message, in this case lsp shouldn't overwrite the find-file prompt.
+  (when (= (minibuffer-depth) 0)
+    (message "%s :: %s" (propertize "LSP" 'face 'success) (apply #'format format args))))
 
 (defun lsp--warn (format &rest args)
   "Display lsp warn message with FORMAT with ARGS."
-  (message "%s :: %s" (propertize "LSP" 'face 'warning) (apply #'format format args)))
+  (when (= (minibuffer-depth) 0)
+    (message "%s :: %s" (propertize "LSP" 'face 'warning) (apply #'format format args))))
 
 (defun lsp--error (format &rest args)
   "Display lsp error message with FORMAT with ARGS."


yyoncho:
Hm, I am unable to reproduce your issue with the following code:

```
(progn 
  (run-with-idle-timer 3 nil (lambda ()
                               (message "Foo message")))

  (call-interactively 'find-file))
```
and then `emas -q`. 

About this PR - I dont like the fact that we will waste these messages and they won't be present at all. We should at least log them via `lsp-log` or try something with `inhibit-message`. But let first find a reproducer.

JohnC32:
I will see if I can make a simpler reproduction step. The overwrite of items in the echo area was constantly confusing "new" people to emacs (and even some of us with more experience in emacs) because they don't realize the echo area is waiting for input.

If you look at `kmacro-call-macro' it contains the following to prevent overwriting a message in the echo area. Using current-message is more aggressive than checking the mini-buffer depth.

````emacs-lisp
      (unless (current-message)
	(message "(Type %s to repeat macro%s)"
		 (format-kbd-macro (vector repeat-key) nil)
		 (if (and kmacro-call-repeat-with-arg
			  arg (> arg 1))
		     (format " %d times" arg) "")))
````

**Comment by yyoncho** [comment]
I understand the issue and we should solve it, but if we hide a message like "Your project is not loaded correctly, feature FOO wont work" will cause trouble too.

**Comment by JohnC32** [comment]
I now see that my proposed solution doesn't work - it works in one case when minibuffer depth is > 0, but for some reason when find-file is active it is 0. Even so the point about always displaying the message is valid.

What about delaying the display of the message. What about something like this where lsp--info, lsp--warn, lsp--error would call lsp--show-message, which contains:

````emacs-lisp
(defun lsp--show-message (msg)
  "Show a message in the echo area when the echo area isn't busy"
  (if (> (with-current-buffer (window-buffer (minibuffer-window)) (point-max)) 1)
      (lexical-let ((msg-copy msg))
        (run-at-time 1 nil
                     (lambda () (lsp--show-message msg-copy))))
    (message "%s" msg)))

(progn
  (run-with-idle-timer 3 nil (lambda ()
                               (lsp--show-message "foo message")))
  (call-interactively 'find-file))
````

I haven't tried this in lsp-mode as I want to get your take before looking at this more.

Thanks

**Comment by yyoncho** [comment]
I suggest first get a reproducer. I suspect that the issue is fixed in the latest versions of emacs. In case it is, we may patch lsp-mode for older emacsens . In case it is not, we may ask in emacs devel since it is a generic issue and it should have a generic solution. If it doesn't have such a solution your proposal seems to make sense. We also may try also to append the message to the existing message.

**Comment by JohnC32** [comment]
This is reproducible in our environment. We have a large C++ code base and ccls takes about 6 minutes to load on a high-end computer (wish that were faster - this was one advantage of RTags when we used that). Below is a gif where we find-file and then attempt to find-file again. While the 2nd find-file prompt is active, lsp writes a message indicating the file isn't indexed. This overwrites the find-file prompt. The LSP message is because ccls hasn't finished starting up. After about 6 minutes the file becomes indexed, even so the message isn't updated and it's still hiding the find-file prompt. If you do try typing while the find-file prompt is active, it will be quickly overwritten by the LSP not index message. Once ccls is loaded, the messages due to not indexed stop appearing.

lsp-overwrite

Getting a test case outside of our environment could be done by using a large C++ code base, e.g. several million lines of C++ and several 10,000's files. Index all files offline using ccls. This will take many hours. Then use lsp-mode in emacs, do a find-file followed by another find-file where you pause a moment on the 2nd find file. You should see the prompt for 2nd find-file overwritten by lsp-mode.

I've been trying the above suggestion locally and it seems to work nicely. After a little more testing, I'm going to deploy it within our environment (many people will use it). 

Having a core emacs solution would be nice. Special handling for this in lsp mode would also be okay as we see this being done elsewhere, e.g. with Emacs macros. We are using Emacs 26 as this comes with Debian 10. The logistics of deploying Emacs 27 are quite costly, so we'd like to stick with stock Debian emacs.

One other option would be to not use the message function and to use a specific lsp log buffer. The mode-line for the buffers already have lsp info. Short "keywords" for lsp could be shown there and clicking them would open the log buffer, e.g. lsp-not-loaded could be displayed in mode line and clicking it would give more details. I personally find the number of messages displayed by lsp mode distracting. Having them logged and accessible from the mode-line would make lsp more usable to me. I could ignore them as I'm working and visit them when convenient. 

Thanks for taking the time to help us with this

**Comment by yyoncho** [comment]
I confirm that it reproduces only in 26. Also, using inhibit-message fixes the issue as well. 

I propose two options: 

1. Use inhibit-message to prevent the message to pop up when there is already a message and when the user is on emacs 26. It should work as expected, the message will be present in the Messages buffer, but it won't be presented for the user. This hurts the experience but the impact is relatively low since it will happen only on emacs 26. 

2. Reimplement emacs 27 fix for message function, again only for emacs 26. 

What you are suggesting with the separate buffer and indication on the modeline will work but it will diverge from the typical emacs packages are using  `message` for notifying the user.

**Comment by yyoncho** [comment]
Actually, I just realized that we want to hide the text in case it is not waiting for user input(e. g. two calls of info function should result in  second message hiding the first one). 

@eli-zaretskii, the following piece of code in emacs 26.3 results in `message` call hiding the `find-file` content. The same works fine in emacs 27+. Is there a recipe to fix that usability issue for older emacs versions? Ideally, we would like to be able to find out that we should not hide current content and just log it in `messages` buffer or we want emacs 27 behavior. 

```
(progn 
  (run-with-idle-timer 3 nil (lambda ()
                               (message "Foo message")))
  (call-interactively 'find-file))
```

Eli-Zaretskii:
> Is there a recipe to fix that usability issue for older emacs versions?

Not really: this functionality in Emacs 27 is based on changes on the C level, which allow to control what `message` does via a Lisp function.  I doubt that this could be emulated easily in older Emacs versions.  Sorry.

yyoncho:
@Eli-Zaretskii thank you!

@JohnC32 is this your issue only with these error messages? If yes, we may think about a different solution.

JohnC32:
Hi

The issue occurs quite frequently when files aren't index (e.g. ccls is loading), or the file hasn't been indexed because it's new. It does feel like some annotation in the mode-line could suffice for some set of these high frequency messages. I don't see many packages sending messages asynchronously which is why I was hoping there could be a mode-line solution.

Thanks
John

yyoncho:
Closing in favor of #2399 . Please test it and let me know.

PR #2385, opened 2020-12-07
Add `lsp-rust-analyzer-rust-src-path` to lsp-rust

worr:
`lsp-rust-analyzer-rust-src-path` allows users to hint to the the rust-analyzer
client where it can find the standard library if they have installed it through
a means other than `rustup`.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -491,6 +491,12 @@ The command should include `--message=format=json` or similar option."
   :group 'lsp-rust
   :package-version '(lsp-mode . "7.1.0"))
 
+(defcustom lsp-rust-analyzer-rust-src-path nil
+  "The location of the rust-src component or checkout of the rust standard library source. The location should should contain a `core' directory."
+  :type 'string
+  :group 'lsp-rust
+  :package-version '(lsp-mode . "7.1.0"))
+
 (defun lsp-rust-analyzer--make-init-options ()
   "Init options for rust-analyzer"
   `(:diagnostics (:enable ,(lsp-json-bool lsp-rust-analyzer-diagnostics-enable)
@@ -629,6 +635,8 @@ The command should include `--message=format=json` or similar option."
                             "rust-analyzer")
                        ,@(cl-rest lsp-rust-analyzer-server-args))))
   :major-modes '(rust-mode rustic-mode)
+  :environment-fn (lambda ()
+                    '(("RUST_SRC_PATH" . lsp-rust-analyzer-rust-src-path)))
   :priority (if (eq lsp-rust-server 'rust-analyzer) 1 -1)
   :initialization-options 'lsp-rust-analyzer--make-init-options
   :notification-handlers (ht<-alist lsp-rust-notification-handlers)
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -11,6 +11,7 @@
   * Automatically download [[https://github.com/eclipse/lemminx][XML language server Lemminx]]
   * Add Vala support.
   * Add [[https://github.com/sorbet/sorbet][Sorbet Language Server]] for typechecking Ruby code.
+  * Added ~lsp-rust-analyzer-rust-src-path~
 ** Release 7.0.1
   * Introduced ~lsp-diagnostics-mode~.
   * Safe renamed ~lsp-flycheck-default-level~ -> ~lsp-diagnostics-flycheck-default-level~


yyoncho (approved):

yyoncho:
Looks good to me! @flodiebold, willing to take a look?

clients/lsp-rust.el
```diff
@@ -491,6 +491,12 @@ The command should include `--message=format=json` or similar option."
   :group 'lsp-rust
   :package-version '(lsp-mode . "7.1.0"))
 
+(defcustom lsp-rust-analyzer-rust-src-path nil
+  "The location of the rust-src component or checkout of the rust standard library source. The location should should contain a `core' directory."
```
flodiebold, line 495: TBH I'm not convinced we need a setting for this -- it's only required in special setups, and you can set the environment variable just as well on the emacs process, or put `(setenv "RUST_SRC_PATH" "...")` into your init file.

If we add it, I'd insist on at least adding something like "You do not need to set this if you're using rustup" to the doc string.

flodiebold: Although I think a setting that allows setting env variables for rust-analyzer in general would actually be useful :thinking:

worr: Personally, I added this because this is my typical setup (and likely a setup that may be more common than you'd expect). I use OpenBSD as my daily driver, and unfortunately there's no rustup support for it whatsoever. For anyone on an OS that has support in the compiler, but no rustup support, this is a typical setup. This is admittedly a small number of users (from what I can tell from the [rust components list](https://rust-lang.github.io/rustup-components-history/). It looks like the people that need this are primarily OpenBSD, DragonflyBSD and Haiku users.

If this is enough reason to either add this explicitly, or to add support for injecting env vars generically to rust-analyzer, I'm happy to amend the comment or even implement generic env support.

flodiebold: @yyoncho What do you think?

@worr I'm wondering what would be necessary to make your setup work without this though. I assume `rustc` is still on the path in your setup, right? What does `rustc --print sysroot` print? Could the rust source be put there?

worr: I had actually thought of doing that first before making this change, but unfortunately that's not feasible.  `rustc --print sysroot` on the Rust from OpenBSD packages shows `/usr/local` as the sysroot.

yyoncho: Support for injecting env vars sounds good to me.

flodiebold: @worr so where is the rust source installed in your setup, and where are the `rlib`s?

worr: rlibs are stored in `/usr/local/lib/rustlib/x86_64-unknown-openbsd/lib/`

The source I have pulled down in my home directory separately, checked out to the version of rust that they package.

flodiebold: So couldn't whatever mechanism installs the rlibs into that folder also install the corresponding source into `/usr/local/lib/rustlib/src`? That would also make sure you've got the correct version of the source.

worr: There's no package that includes the source (I will ask some OpenBSD devs about this, since there are genuine use-cases for it, like `rust-analyzer`). That still does not resolve the problem though, since `rust-analyzer` locates the source [by getting the sysroot path, and looks for the source for the crates in the sysroot path.](https://github.com/rust-analyzer/rust-analyzer/blob/f925735e64cb6aed85f28cacc0a91c1c1bc06bb4/crates/project_model/src/sysroot.rs#L59-L107)

So in my case, with a sysroot of `/usr/local`, it'd expect to find to find effectively the contents of `rust/library`. Even if OpenBSD put the source in a separate package, their packaging would certainly not allow for distributing all of the library source in the top of  `/usr/local` like that.

flodiebold: It'd expect to find the sources in [`/usr/local/lib/rustlib/src/rust`](https://github.com/rust-analyzer/rust-analyzer/blob/f925735e64cb6aed85f28cacc0a91c1c1bc06bb4/crates/project_model/src/sysroot.rs#L155), i.e. an `src` folder on the same level as the `x86_64-unknown-openbsd` folder. That seems to me like a reasonable location. And yeah, it'd be great if the sources could be packaged.

worr: Thanks for fixing my misunderstanding about how this worked. Adding a symlink to my checkout there worked beautifully. I'll close out this PR. Sorry about all of this!

PR #2387, opened 2020-12-08
Fix keyword auto-completion in classes for iph

nbfalcon:
When auto-completing public, ... at the start of a class, Intelephense
sends `CompletionList`s that have no :isIncomplete fields (a spec
violation), causing `lsp-completion--sort-completions` to get a
`CompletionList` instead of a list of `CompletionItem`, `error`ing.

Extend `lsp-interface` to take an optional fourth list of semi-optional
arguments (like optional but without a ?) for each interface tuple.
Define `CompletionList` so that is-optional is semi-optional.

### DIFF
--- a/lsp-protocol.el
+++ b/lsp-protocol.el
@@ -44,7 +44,7 @@
   (defvar lsp-use-plists nil))
 
 (defmacro lsp-interface (&rest interfaces)
-  "Generate LSP bindings from INTERFACES triplet.
+  "Generate LSP bindings from INTERFACES tuples.
 
 Example usage with `dash`.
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Is this reported upstream? Also, can we replace this one with a local patch(e. g. check only if :items are present to determine the type of the object) instead of introducing the patch on the global level?

nbfalcon:
@yyoncho this is more-or-less a local patch: `lsp-interface` is extended to support semi-required fields, and the only interface definition that changes is `CompletionList`, which now doesn't require `:isIncomplete` to be present. All other interfaces are completely unaffected.

nbfalcon:
I don't know whether this can be raised upstream - I cannot submit a PR, given that `iph` is proprietary.

yyoncho:
`lsp-interface` part of the public `lsp-mode`'s interface so it is not local.

nbfalcon:
My change doesn't break anything - all previous uses of `lsp-interface` will still continue to work like they did, if the use triples. This is because `-let` + list destructuring sets variables to nil if the list ends first:
```emacs-lisp
(funcall (-lambda ((a b c d)) d) '((a))) ; nil, not error, even though we are missing three elements
```

That also means that interface triples can also be pairs, or even one-tuples.

yyoncho:
The issue is that we add a parameter that makes little sense and it is hard to explain. Using `vectorp` instead of `lsp-completion-list?`  will fix the code too and it won't introduce awkward changes to public lsp-mode functions.

kiennq:
> Using `vectorp` instead of `lsp-completion-list?` will fix the code too and it won't introduce awkward changes to public lsp-mode functions.

I think we will actually need to check for `items` is presented in response from server to fix this error.

Anyway, this is clearly spec violation, and the more we support those behaviors, the more our code become unmanageable. Extending the interface to accommodate the kind of semi-optional is not making sense too.

We can actually ask the user of `iph` to actually advice the `lsp-require-while-no-input` to specially insert the `isIncomplete` field into `textDocument/completion` response. That requires no hack from `lsp-mode` (so we keep spec compliant) and discourage from using spec violated language server.
If you pay for `iph`, you should have the rights to file bug and ask for its changes to compliant with actual specs.

yyoncho:
> I think we will actually need to check for `items` is presented in response from server to fix this error.

This is what I initially suggested. Spent some time analyzing the code and you will end up with the conclusion that using `vectorp` is much easier to implement.

kiennq:
Ah, yes, that would work too.

yyoncho:
> Ah, yes, that would work too.

The issue is with the merge function and we do not expose `member` func via lsp-protocol.el.

PR #2388, opened 2020-12-08
Bugfix/lsp lens support check

nbfalcon:
- In `lsp--auto-configure`: add `lsp--lens-enable` instead of `lsp-lens-mode`, so that the latter won't be enabled for servers that don't support lenses. Unlike stated in the commit messages, this doesn't error. The errors were either fixed for some other reason or caused by a broken environment of mine.
- `lsp-lens-mode` can now be disabled, without the lenses reappearing shortly thereafter.


nbfalcon:
I'm making a new PR because my commit didn't make it in.

PR #2389, opened 2020-12-08
Bugfix/lens check if supported

nbfalcon:
- In `lsp--auto-configure`: add `lsp--lens-enable` instead of `lsp-lens-mode`, so that the latter won't be enabled for servers that don't support lenses. Unlike stated in the commit messages, this doesn't error. The errors were either fixed for some other reason or caused by a broken environment of mine.
- `lsp-lens-mode` can now be disabled, without the lenses reappearing shortly thereafter.

### DIFF
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -125,9 +125,10 @@ Results are meaningful only if FROM and TO are on the same line."
   "Handler for `after-save-hook' for lens mode."
   (lsp-lens--schedule-refresh t))
 
-(defun lsp-lens--schedule-refresh (buffer-modified?)
+(defun lsp-lens--schedule-refresh (&optional buffer-modified?)
   "Call each of the backend.
-BUFFER-MODIFIED? determines whether the buffer is modified or not."
+BUFFER-MODIFIED? determines whether the buffer was modified or
+not."
   (-some-> lsp-lens--refresh-timer cancel-timer)
 
   (setq lsp-lens--page (cons (window-start) (window-end)))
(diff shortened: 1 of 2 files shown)


nbfalcon:
This is the last PR that I'm making from upstream.

nbfalcon:
Note that disabling `lsp-lens-mode` and then reverting the buffer will turn it back on again. This is because doing so causes `lsp-configure-buffer` to be re-run, and is not an issue that needs to be addressed (at least here, probably at all).

lsp-lens.el
```diff
@@ -347,18 +353,20 @@ CALLBACK - callback for the lenses."
...
-    (lsp-lens-hide)
     (remove-hook 'lsp-configure-hook #'lsp-lens--enable t)
     (remove-hook 'lsp-unconfigure-hook #'lsp-lens--disable t)
     (remove-hook 'lsp-on-idle-hook #'lsp-lens--idle-function t)
-    (remove-hook 'lsp-on-change-hook (lambda () (lsp-lens--schedule-refresh nil)) t)
-    (remove-hook 'after-save-hook (lambda () (lsp-lens--schedule-refresh t)) t)
+    (remove-hook 'lsp-on-change-hook #'lsp-lens--schedule-refresh-modified t)
+    (remove-hook 'after-save-hook #'lsp-lens--schedule-refresh t)
     (remove-hook 'before-revert-hook #'lsp-lens-hide t)
+    (cancel-timer lsp-lens--refresh-timer)
```
yyoncho, line 367: dont we need a check if this is nil?

nbfalcon: It indeed can be nil. Fixed.

yyoncho (approved):

nbfalcon:
Can I just merge this (given that you approved), or do you want to merge PRs yourself?

yyoncho:
After approval you may merge it as it is or keep it open for some time if you think that more reviewers are needed for the pr.

nbfalcon:
Rerun CI.

nbfalcon:
@yyoncho I cannot merge this because the windows-test CI is broken (`::set_env` disabled).

jcs090218:
The CI failed because this PR uses `v3` instead of `master`.

nbfalcon:
@jcs090218 `v3`? How would I make it not fail?

jcs090218:
I think this PR opened while CI is still in `v3`. Currently the master branch is using `master`. I don't knw how to get around this; yet I think there should be a way to update PR's CI environment. Or else, I don't see a solution here. 😕 If there isn't a way to do this, then I assumed this should be an enhancement for GitHub. Like updating the CI testing environment respect to the master branch, etc.

PR #2399, opened 2020-12-10
Make sure calls to message don't remove the prompt(in async context)

### LINKED ISSUE #2049: "LSP :: message" hides minibuffer prompt in the echo area

yyoncho:
- fixes #2049

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1205,17 +1205,32 @@ Deprecated. Use `lsp-repeatable-vector' instead. "
 
 (make-obsolete 'lsp-string-vector nil "lsp-mode 7.1")
 
+(defun lsp--message  (format &rest args)
+  "Wrapper for `message'
+
+We `inhibit-message' the message when the cursor is in the
+minibuffer and when emacs version is before emacs 27 due to the
+fact that we often use `lsp--info', `lsp--warn' and `lsp--error'
+in async context and the call to these function is removing the
+minibuffer prompt. The issue with async messages is already fixed
+in emacs 27.
+
+See #2049"
+  (let ((inhibit-message (and (minibufferp)
+                              (version< emacs-version "27.0"))))
+    (apply #'message format args)))
+
 (defun lsp--info (format &rest args)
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

lsp-mode.el
```diff
@@ -1205,17 +1205,32 @@ Deprecated. Use `lsp-repeatable-vector' instead. "
...
+
+We `inhibit-message' the message when the cursor is in the
+minibuffer and when emacs version is before emacs 27 due to the
+fact that we often use `lsp--info', `lsp--warn' and `lsp--error'
+in async context and the call to these function is removing the
+minibuffer prompt. The issue with async messages is already fixed
+in emacs 27.
+
+See #2049"
+  (let ((inhibit-message (and (eq major-mode 'minibuffer-inactive-mode)
```
kiennq, line 1219: Why don't we use `(minibufferp)`?

kiennq (approved):

PR #2400, opened 2020-12-11
Add lsp-gunzip function to decompress gzipped files

bubba:
Also add the `:decompress` argument to lsp-download-install so that dependencies can decompress themselves.
My main use case is to be able to download binaries for `lsp-haskell.el`, which has the binaries compressed with gzip. This way we can use dependencies like:

```elisp
(lsp-dependency
 'haskell-language-server-wrapper
 '(:system "haskell-language-server-wrapper")
 `(:download :url ,(concat lsp-haskell-language-server-github-releases-url
			  (format "haskell-language-server-wrapper-%s.gz"
				  (lsp-haskell--github-system-type-suffix)))
	     :store-path ,(f-join lsp-server-install-dir "haskell" "haskell-language-server-wrapper")
	     :decompress :gzip
	     :set-executable? t))
```

So far I haven't add a script for windows Powershell yet, as I don't have a windows machine to test this on.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7201,7 +7201,7 @@ nil."
            (:gzip (concat store-path ".gz"))
            (:zip (concat store-path ".zip"))
            (`nil store-path)
-            (_ (error ":decompress must be `:gzip', `:zip' or `nil'")))))
+           (_ (error ":decompress must be `:gzip', `:zip' or `nil'")))))
     (make-thread
      (lambda ()
        (condition-case err
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7201,7 +7201,7 @@ nil."
            (:gzip (concat store-path ".gz"))
            (:zip (concat store-path ".zip"))
            (`nil store-path)
-           (_ (error ":decompress must be `gzip', `zip' or `nil'")))))
+            (_ (error ":decompress must be `:gzip', `:zip' or `nil'")))))
     (make-thread
      (lambda ()
        (condition-case err
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7198,8 +7198,8 @@ nil."
          ;; (decompress (lsp-resolve-value decompress))
          (download-path
           (pcase decompress
-           (`gzip (concat store-path ".gz"))
-           (`zip (concat store-path ".zip"))
+           (:gzip (concat store-path ".gz"))
+           (:zip (concat store-path ".zip"))
            (`nil store-path)
            (_ (error ":decompress must be `gzip', `zip' or `nil'")))))
     (make-thread
@@ -7215,8 +7215,8 @@ nil."
              (when decompress
                (lsp--info "Decompressing %s..." download-path)
                (pcase decompress
-                 (`gzip (lsp-gunzip download-path))
-                 (`zip (lsp-unzip download-path store-path)))
+                 (:gzip (lsp-gunzip download-path))
+                 (:zip (lsp-unzip download-path store-path)))
                (lsp--info "Decompressed %s..." store-path))
              (funcall callback))
          (error (funcall error-callback err)))))))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7192,19 +7192,32 @@ nil."
 
 
 ;; Download URL handling
-(cl-defun lsp-download-install (callback error-callback &key url store-path &allow-other-keys)
-  (let ((url (lsp-resolve-value url))
-        (store-path (lsp-resolve-value store-path)))
+(cl-defun lsp-download-install (callback error-callback &key url store-path decompress &allow-other-keys)
+  (let* ((url (lsp-resolve-value url))
+         (store-path (lsp-resolve-value store-path))
+         ;; (decompress (lsp-resolve-value decompress))
+         (download-path
+          (pcase decompress
+           (`gzip (concat store-path ".gz"))
+           (`zip (concat store-path ".zip"))
+           (`nil store-path)
+           (_ (error ":decompress must be `gzip', `zip' or `nil'")))))
     (make-thread
      (lambda ()
        (condition-case err
            (progn
-             (when (f-exists? store-path)
-               (f-delete store-path))
-             (lsp--info "Starting to download %s to %s..." url store-path)
-             (mkdir (f-parent store-path) t)
-             (url-copy-file url store-path)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7192,19 +7192,32 @@ nil."
 
 
 ;; Download URL handling
-(cl-defun lsp-download-install (callback error-callback &key url store-path &allow-other-keys)
-  (let ((url (lsp-resolve-value url))
-        (store-path (lsp-resolve-value store-path)))
+(cl-defun lsp-download-install (callback error-callback &key url store-path gzip? zip? &allow-other-keys)
```
nbfalcon, line 7195: ZIP? and GZIP? are exclusive, no? In that case it might be better to have a `container?` parameter, which could be one of a few keywords like `:gzip`, `:zip`.

lsp-mode.el
```diff
@@ -7192,19 +7192,32 @@ nil."
...
+  (let* ((url (lsp-resolve-value url))
+         (gzip? (lsp-resolve-value gzip?))
+         (store-path (lsp-resolve-value store-path))
+         (download-path
+          (cond
```
nbfalcon, line 7200: If you use a `:container?` field isntead, you can use `pcase` here.

lsp-mode.el
```diff
@@ -7192,19 +7192,32 @@ nil."
 
 
 ;; Download URL handling
-(cl-defun lsp-download-install (callback error-callback &key url store-path &allow-other-keys)
-  (let ((url (lsp-resolve-value url))
-        (store-path (lsp-resolve-value store-path)))
+(cl-defun lsp-download-install (callback error-callback &key url store-path decompress &allow-other-keys)
```
nbfalcon, line 7195: It is Elisp convention to use keywords if you have a fixed set of options. I.e. `:gzip`, `:zip`. Can you do this change without force-pushing?

lsp-mode.el
```diff
@@ -7192,19 +7192,32 @@ nil."
...
+             (when (f-exists? download-path)
+               (f-delete download-path))
+             (lsp--info "Starting to download %s to %s..." url download-path)
+             (mkdir (f-parent download-path) t)
+             (url-copy-file url download-path)
+             (lsp--info "Finished downloading %s..." download-path)
+             (when decompress
+               (lsp--info "Decompressing %s..." download-path)
+               (pcase decompress
+                 (`gzip (lsp-gunzip download-path))
```
nbfalcon, line 7218: This can be done asynchronously. `lsp-gunzip-async` could get CALLBACK and ERROR-CALLBACK as parameters, and call `lsp-start-process-async` instead. An async version of `lsp-unzip` should be added as well.

nbfalcon (changes requested):

nbfalcon:
Otherwise, LGTM. Perhaps @yyoncho might have some more comments.

Thank you for contributing!

lsp-mode.el
yyoncho: @bubba you may go without this

yyoncho:
> So far I haven't add a script for windows Powershell yet, as I don't have a windows machine to test this on.

@kiennq - can you help with this?

lsp-mode.el
nbfalcon: I can do that one afterwards.

lsp-mode.el
bubba: TIL about keyword symbols. Pushed the commit there, hope I'm understanding this correctly

lsp-mode.el
```diff
@@ -7198,8 +7198,8 @@ nil."
          ;; (decompress (lsp-resolve-value decompress))
          (download-path
           (pcase decompress
-           (`gzip (concat store-path ".gz"))
-           (`zip (concat store-path ".zip"))
+           (:gzip (concat store-path ".gz"))
+           (:zip (concat store-path ".zip"))
            (`nil store-path)
            (_ (error ":decompress must be `gzip', `zip' or `nil'")))))
```
nbfalcon, line 7204:
```suggestion
            (_ (error ":decompress must be `:gzip', `:zip' or `nil'")))))
```

nbfalcon: Except for this small point, LGMT.

yyoncho:
> > So far I haven't add a script for windows Powershell yet, as I don't have a windows machine to test this on.
> 
> @kiennq - can you help with this?

Actually, I just realized that PowerShell works on linux as well. I will come up with a proposal.

nbfalcon:
There is [Expand-Archive](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/Expand-Archive?view=powershell-7.1&viewFallbackFrom=powershell-6).

lsp-mode.el
```diff
@@ -7201,7 +7201,7 @@ nil."
            (:gzip (concat store-path ".gz"))
            (:zip (concat store-path ".zip"))
            (`nil store-path)
-           (_ (error ":decompress must be `gzip', `zip' or `nil'")))))
+            (_ (error ":decompress must be `:gzip', `:zip' or `nil'")))))
```
nbfalcon, line 7204:
```suggestion
           (_ (error ":decompress must be `:gzip', `:zip' or `nil'")))))
```
Sorry, that was one too much whitespace.

nbfalcon:
`Expand-Archive` won't work because it only supports `zip`, not `gzip`.

bubba:
Maybe we can use https://github.com/emacs-mirror/emacs/blob/master/lisp/jka-compr.el

yyoncho:
Code looks good to me. The only thing that is missing is the windows scripts. I found this - https://scatteredcode.net/download-and-extract-gzip-tar-with-powershell/ which has gzip script but it seems like it can be used only on windows since it depends on .net(might be wrong). I think that we can merge it as it is if we change the error message: `(error "Unable to find `gzip' on the path, please customize `lsp-gunzip-script'")` to something like:

`Unable to ... , please either customize ... or unzip manually path-to-arch(to location ...)`.

This way the user can continue the installation manually.

lsp-mode.el
```diff
@@ -40,6 +40,7 @@
 (require 'ht)
 (require 'imenu)
 (require 'inline)
+(require 'jka-compr)
```
yyoncho, line 43: leftover

yyoncho:
Thank you!

PR #2402, opened 2020-12-11
Feature/handle disabled code actions

nbfalcon:
- Handle disabled code actions (not shown in `lsp-execute-code-action` unless C-u prefix specified)
- Properly propertize preferred code actions
- Add `lsp-auto-fix`, which executes one of the preferred code actions
- Add `lsp-execute-code-action-by-type`, which executes a code action by given base type (e.g. "refactor", ...)

I have tested this a bit with `jdtls`, and this PR seems to not introduce a regression. However, I didn't have the chance to test the new disabled and preferred code action handling yet, which is why I am marking as draft.

There will be a corresponding `lsp-ui` PR soon, too (there is just one minor change I need to make, and to test this).


lsp-mode.el
```diff
@@ -4868,18 +4869,53 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
+
+(defface lsp-disabled-code-action-reason-face '((t :inherit lsp-details-face))
+  "Face used to highlight the reason a code action is disabled.
+Shown after the code action in `lsp-execute-code-action',
+... (see `lsp-preferred-code-action-face').")
+
+(lsp-defun lsp--code-action-title ((action &as &CodeAction :title :is-preferred? :disabled?))
+  ;; It would be strange for a code action to be both preferred and disabled, so
+  ;; `cond' is probably good enough.
+  (cond (is-preferred?
```
yyoncho, line 4890: Ideally we would want the default code action to be preselected in the select dialog. If you don't want to adress it in this PR you may leave a TODO.

lsp-mode.el
```diff
@@ -4868,18 +4869,53 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
-   ((and (eq (seq-length actions) 1) lsp-auto-execute-action)
-    (lsp-seq-first actions))
-   (t (let ((completion-ignore-case t))
-        (lsp--completing-read "Select code action: "
-                              (seq-into actions 'list)
-                              (-compose (lsp--create-unique-string-fn)
-                                        #'lsp:code-action-title)
-                              nil t)))))
+  (let ((actions (seq-into actions 'list)))
+    (cond
```
yyoncho, line 4903: If you are going to change it, then change it to pcase.

lsp-mode.el
```diff
@@ -5209,12 +5245,31 @@ It will show up only if current point has signature help."
...
-  (if-let ((action (->> (lsp-get-or-calculate-code-actions command-kind)
-                        (-filter (-lambda ((&CodeAction :kind?))
-                                   (and kind? (equal command-kind kind?))))
-                        lsp--select-action)))
-      (lsp-execute-code-action action)
-    (signal 'lsp-no-code-actions '(command-kind))))
+  (->> (lsp-get-or-calculate-code-actions command-kind)
+       (-filter (-lambda ((&CodeAction :kind?))
+                  (and kind? (equal command-kind kind?))))
+       lsp--select-enabled-action
```
yyoncho, line 5251: If I call explicitly a refactoring and the method is disabled I would like to see the reason for which it is disabled. E. g. if I call "Extract method" I might see an error "You have selected a range that cannot be extracted to a method".

lsp-mode.el
```diff
@@ -5224,10 +5279,27 @@ It will show up only if current point has signature help."
...
     (lsp--send-execute-command command arguments?)))
 
+(defun lsp--select-action-enabled-or-die (actions)
+  "Prompt the user to select an action from ACTIONS.
+Even disabled code actions are shown. If such an action is
+chosen, `user-error'."
+  (let ((action (lsp--select-action actions)))
+    (-when-let ((&CodeAction :disabled?) action)
+      (user-error "%s is disabled: %s" (lsp:code-action-title action)
+                  (lsp:code-action-disabled-reason disabled?)))
```
yyoncho, line 5289: this seems like should be part of `lsp-execute-code-action`. E. g. if lsp-ui calls the method with disabled action it should get that error.

lsp-mode.el
```diff
@@ -5224,10 +5279,27 @@ It will show up only if current point has signature help."
...
-  (interactive (list (lsp--select-action (lsp-code-actions-at-point))))
+Interactively, ask for a code action. If the prefix argument is
+unspecified, the set of code actions will be filtered to only
+include those that are not disabled."
+  (interactive
+   (list
+    (funcall (if current-prefix-arg
+                 #'lsp--select-action-enabled-or-die
+               #'lsp--select-enabled-action)
+             (lsp-code-actions-at-point))))
```
yyoncho, line 5302: seems like if you drop lsp--select-enabled-action/lsp--select-action-enabled-or-die and pass an optional parameter to lsp-code-actions-at-point the code will be simplified/shortened a lot.

lsp-mode.el

lsp-mode.el

lsp-mode.el

lsp-mode.el
nbfalcon: I have implemented this. That feature was surprisingly easy to implement, requiring only a small change to `lsp-completing-read` and one to `lsp--select-action`.

lsp-mode.el
nbfalcon: Note: this is now an `error` instead of a `user-error` (reasoning in the commit message).

lsp-mode.el
nbfalcon: Fixed. There might be a prompt with the reasons listed individually, depending on the number of actions and `lsp-auto-execute-action`.

nbfalcon:
Should I prepend "Review: " to the commit subject if it doesn't make it go over 50 lines?

yyoncho:
> Should I prepend "Review: " to the commit subject if it doesn't make it go over 50 lines?

We dont have such a requirement. It will be squashed in the end so it shouldn't matter.

lsp-mode.el
```diff
@@ -5203,18 +5239,44 @@ It will show up only if current point has signature help."
...
                     ,@(when kind (list :only (vector kind))))))
 
-(defun lsp-code-actions-at-point (&optional kind)
-  "Retrieve the code actions for the active region or the current line."
-  (lsp-request "textDocument/codeAction" (lsp--text-document-code-action-params kind)))
+(defun lsp-code-actions-at-point (&optional kind enabled)
+  "Retrieve the code actions for the active region or the current line.
+If ENABLED is specified, only return code actions that are not
+disabled."
+  (--doto (lsp-request "textDocument/codeAction" (lsp--text-document-code-action-params kind))
```
yyoncho, line 5246: I guess this section is on the edge of being too trickier.

lsp-mode.el
```diff
@@ -5203,18 +5239,44 @@ It will show up only if current point has signature help."
...
+  (->> (lsp-get-or-calculate-code-actions command-kind)
+       (-filter (-lambda ((&CodeAction :kind?))
+                  (and kind? (equal command-kind kind?))))
+       ;; Fail in `lsp-execute-code-action' if there is only one disabled
+       ;; action, and show a list of applicable actions with an overview of why
+       ;; they are disabled otherwise.
+       lsp--select-action
+       lsp-execute-code-action))
+
+(defun lsp-execute-code-action-by-type (kind)
```
yyoncho, line 5261: can you elaborate on why we need this method?

lsp-mode.el
```diff
@@ -5203,18 +5239,44 @@ It will show up only if current point has signature help."
...
 
 (defun lsp-execute-code-action-by-kind (command-kind)
   "Execute code action by name."
-  (if-let ((action (->> (lsp-get-or-calculate-code-actions command-kind)
-                        (-filter (-lambda ((&CodeAction :kind?))
-                                   (and kind? (equal command-kind kind?))))
-                        lsp--select-action)))
-      (lsp-execute-code-action action)
-    (signal 'lsp-no-code-actions '(command-kind))))
+  (->> (lsp-get-or-calculate-code-actions command-kind)
```
yyoncho, line 5252: don't we need to filter the disabled and only in case there is no enabled action use the disabled one(s) error message?

yyoncho (commented):
Looks good. I would have approved it if it was not marked as draft.

lsp-mode.el
nbfalcon: If there is some specific code action to be executed, e.g. `source.organizeImports`, `lsp-execute-code-action-by-kind` is enough. However, sometimes it can be useful to bring up all quickfixes, or all refactorings: in this case, `lsp-execute-code-action-by-kind` doesn't work, because it must match the kind exactly. My new function matches a base kind: `lsp-execute-code-action-by-type` "refactor" would match "refactor.move", "refactor.*", ... (tested with `jdtls`). This corresponds to the spec's "if the user requests a more specific type of code action, e.g. refactoring".

lsp-mode.el
nbfalcon: What I had in mind here was the following: currently there is a single filter: ENABLED. In the future, more filters could be added, and using `--doto` with `cl-callf` allows adding them easily (e.g. `is-preferred`). I could rewrite this in terms of `if` if you consider that to be a large enough readability gain:
```emacs-lisp
  (let ((actions (lsp-request "textDocument/codeAction" (lsp--text-document-code-action-params kind))))
    (if enabled
        (seq-remove #'lsp:code-action-disabled? actions)
      acitons))
```

I just realized that the code has a bug: `lsp-code-action-disabled?` should be `lsp:code-action-disabled?`.

lsp-mode.el

nbfalcon:
Now all that is left is to test this.

lsp-mode.el
yyoncho: I guess it could stay as it is, just it is a bit unusual. Once `-cond->>` lands in `dash` we can use it here(and few other places).

nbfalcon: `-cond->>` would certainly be awesome, and I remember that I wanted to use it somewhere but couldn't. However, I don't see how this would help here? The point of this extensible function style is to allow filters to be chained: e.g. first filter out all disabled actions, then those that don't match the type base, .... (we don't have that yet, but might in the future).

nbfalcon:
As a side note, why not directly wait for `lsp-response-timeout` [here](https://github.com/emacs-lsp/lsp-mode/blob/0349a1cc0976829fab8f73ecc033252be31a7cf6/lsp-mode.el#L3009)? I.e. `(accept-process-output nil lsp-response-timeout)`.

lsp-mode.el
yyoncho: Something like that (clj)

```

(let [even true
      small true]
  (cond->> (range 20)
    even (filter even?)
    small (filter #(< 10 %))))
```

yyoncho:
> As a side note, why not directly wait for `lsp-response-timeout` [here](https://github.com/emacs-lsp/lsp-mode/blob/0349a1cc0976829fab8f73ecc033252be31a7cf6/lsp-mode.el#L3009)? I.e. `(accept-process-output nil lsp-response-timeout)`.

We are not doing accept-process-output on the process that we need to read from and also we need to read multiple messages and potentially from different processes. As a side note, this code should be rewritten to use emacs 26 threads.

leungbk:
> As a side note, this code should be rewritten to use emacs 26 threads.

I've heard the implementation isn't thread-safe:

> ThreadSanitizer (TSan) quickly shows that Emacs’ threading implementation has many data races, making it completely untrustworthy. Until this is fixed, nobody should use Emacs threads for any purpose, and threads should disabled at compile time.

https://nullprogram.com/blog/2018/05/31/

yyoncho:
> > As a side note, this code should be rewritten to use emacs 26 threads.
> 
> I've heard Emacs' threading implementation isn't thread-safe:
> 
> > ThreadSanitizer (TSan) quickly shows that Emacs’ threading implementation has many data races, making it completely untrustworthy. Until this is fixed, nobody should use Emacs threads for any purpose, and threads should disabled at compile time.
> 
> https://nullprogram.com/blog/2018/05/31/

As per Eli's comments, it is not that bad as described in this blogpost.

lsp-mode.el
```diff
@@ -4898,9 +4898,9 @@ Shown after the code action in `lsp-execute-code-action',
         (disabled?
          (concat
           (propertize title 'face 'lsp-disabled-code-action-face)
-          " "
```
nbfalcon: Do we need this commit at all? I don't know how far you want to go to support spec-breaking LS.

yyoncho: Sorry, I got lost. Can you elaborate? There are several improvements in this PR so IMHO better to have them in.

nbfalcon: Code actions can have a `disabled?` field, which, as per the spec, *must* have a `reason` member. This commit handles the case where language servers send invalid code actions that have a `disabled` field without a `reason` (I don't know of an LS that does that, though). This is just being defensive, and I'm unsure whether we need to handle that case (it is a spec violation after all).

yyoncho: Most of the time we don't handle them. Leaving up to you to decide.

lsp-mode.el
```diff
@@ -5203,18 +5240,52 @@ It will show up only if current point has signature help."
...
-                        (-filter (-lambda ((&CodeAction :kind?))
-                                   (and kind? (equal command-kind kind?))))
-                        lsp--select-action)))
-      (lsp-execute-code-action action)
-    (signal 'lsp-no-code-actions '(command-kind))))
+  (let* ((all-actions (->> (lsp-get-or-calculate-code-actions command-kind)
+                           (-filter (-lambda ((&CodeAction :kind?))
+                                      (and kind? (equal command-kind kind?))))))
+         (enabled-actions (-remove #'lsp:code-action-disabled? all-actions))
+         (action (lsp--select-action (or enabled-actions all-actions))))
```
kiennq, line 5257: Is this ok? In case of all actions are disabled, we show all actions instead of empty list?
Also, shouldn't we need guard of `action` as the old code, in case of user not select anything.

lsp-mode.el
```diff
@@ -5203,18 +5240,52 @@ It will show up only if current point has signature help."
...
+    (lsp-execute-code-action action)))
+
+(defun lsp-execute-code-action-by-type (kind &optional enabled)
+  "Execute a code action with a given base KIND.
+When ENABLED is given, filter out all disabled code actions. This
+is in contrast with the spec's recommended behavior."
+  (->> (lsp-get-or-calculate-code-actions kind enabled)
```
kiennq, line 5264: `-some->>`?

lsp-mode.el
```diff
@@ -5203,18 +5240,52 @@ It will show up only if current point has signature help."
...
+
+(defun lsp-refactor ()
+  "Execute a \"refactor\" code action at point."
+  (interactive)
+  (lsp-execute-code-action-by-type "refactor"))
+
+(defun lsp-auto-fix ()
+  "Execute a preferred code action at point."
+  (interactive)
+  (->> (lsp-get-or-calculate-code-actions)
```
kiennq, line 5285: `-some->>`?

lsp-mode.el
```diff
@@ -4868,18 +4875,48 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
+  "Face used to highlight disabled code actions.
+See `lsp-preferred-code-action-face' for details.")
+
+(defface lsp-disabled-code-action-reason-face '((t :inherit lsp-details-face))
+  "Face used to highlight the reason a code action is disabled.
+Shown after the code action in `lsp-execute-code-action',
+... (see `lsp-preferred-code-action-face').")
+
+(lsp-defun lsp--code-action-title ((action &as &CodeAction :title :is-preferred? :disabled?))
+  ;; It would be strange for a code action to be both preferred and disabled, so
```
kiennq, line 4894: docstring?

lsp-mode.el
```diff
@@ -4868,18 +4875,48 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
+    (pcase actions
+      (`nil (signal 'lsp-no-code-actions nil))
+      ((and `(,action) (guard lsp-auto-execute-action)) action)
+      (_ (let ((completion-ignore-case t))
+           (lsp--completing-read "Select code action: "
+                                 actions
+                                 (-compose (lsp--create-unique-string-fn)
+                                           #'lsp--code-action-title)
+                                 nil t nil nil
+                                 (or (-find-index #'lsp:code-action-is-preferred? actions)
```
kiennq, line 4918: These should be `-find` instead of `-find-index`, the index passing is only working in `ivy-mode`

nbfalcon: I have implemented that for `lsp--completing-read`.

lsp-mode.el
nbfalcon: With `->>`, we throw `lsp-no-code-actions`, which is right imho.

lsp-mode.el
nbfalcon: Same as below.

lsp-mode.el
nbfalcon: Fixed. In what cases do we need docstrings? In all new code, or just in certain functions?

lsp-mode.el
nbfalcon: The idea is to show an error message in case there is no enabled code action, if the user has enabled `lsp-auto-execute-action` as t ([discussion](https://github.com/emacs-lsp/lsp-mode/pull/2402#discussion_r541268228)). However, I agree that this might be suboptimal. Do you have an idea for a better UI in this case?

As for guarding, we needn't do that, since the user cannot exit from `lsp--select-action` without either selecting an action or `quit`ting (causing a non-local exit), due to REQUIRE-MATCH being set as t.

lsp-mode.el
nbfalcon: I propose to keep it in then, because it can't hurt to have a more lenient lsp implementation.

nbfalcon:
This PR is done. I will not be adding any new commits/features except those needed to address review comments.

lsp-mode.el
kiennq: Ah, sorry, I missed that.
Also, why don't just pass using `-find` instead of `-find-index`?
It seems that you're taking a round trip of finding index and taking the list element by index here.

lsp-mode.el
kiennq: I think we should have docstring in all new function

lsp-mode.el
```diff
@@ -4868,18 +4875,44 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
+  "Face used to highlight preferred code actions.
+Used by `lsp-execute-code-action' and
+`lsp-execute-code-action-by-kind' if they have to prompt for an
+action.")
+
+(defface lsp-disabled-code-action-face '((t :inherit shadow))
+  "Face used to highlight disabled code actions.
+See `lsp-preferred-code-action-face' for details.")
+
+(defface lsp-disabled-code-action-reason-face '((t :inherit lsp-details-face))
```
kiennq, line 4888: This is not used?

lsp-mode.el
kiennq: Isn't `lsp-auto-execute-action` is only for single code action?
The case of `lsp-execute-code-action-by-kind` is falling into the bellow in the spec, should we just take the first code action and let-bind `lsp-auto-execute-action` to `t` so the error will always be shown?
```
	 * - If the user has a keybinding that auto applies a code action and only
	 *   a disabled code actions are returned, the client should show the user
	 *   an error message with `reason` in the editor.
```

lsp-mode.el
nbfalcon: If the COLLECTION is a list of strings that is to be transformed, there would be an ambiguitiy: should DEF be used to look up the corresponding string in COLLECTION, or be the DEF argument to completing read?

lsp-mode.el
nbfalcon: Fixed. I wanted to add a help-echo to disabled code actions at some point, but then realized that it wouldn't work with completing-read.

nbfalcon: Nevermind, this is a regression from the commit to support both preferred and disabled code actions at the same time. It was supposed to show the reason. I'll need to add it back.

lsp-mode.el
nbfalcon: We could `let`-bind `lsp-auto-execute-action` to t if there are no enabled ones, however I am unsure if that is the right thing: `lsp-execute-code-action-by-kind` doesn't auto-execute a single code action if available by default, either. If there is only one disabled code action, a prompt containing only it will show up, and along the code action title the `disabled-reason` will be shown:

lsp-mode.el
nbfalcon: I'll keep that in mind.

lsp-mode.el
```diff
@@ -5203,18 +5238,52 @@ It will show up only if current point has signature help."
...
+is in contrast with the spec's recommended behavior."
+  (->> (lsp-get-or-calculate-code-actions kind enabled)
+       (-filter (-lambda ((&CodeAction :kind?))
+                  (and kind? (string-match-p (format "\\`%s\\(\\.\\|\\'\\)"
+                                                     (regexp-quote kind))
+                                             kind?))))
+       lsp--select-action
+       lsp-execute-code-action))
+
+(defun lsp-quickfix ()
```
kiennq, line 5270: There's `lsp-make-interactive-code-action` macro which use `lsp-execute-code-action-by-kind`, do you want to provide similar one for `lsp-execute-code-action-by-type`?

lsp-mode.el
kiennq: Wouldn't that ambiguity be resolved if we document it? For example, DEF should be the corresponding string in COLLECTION (since it's what's returned). That would also be aligned with original `completing-read` which is what wrapped in `lsp--completing-read`.

lsp-mode.el
kiennq: Shouldn't `lsp--select-action` automatically choose the only one code action available when `lsp-auto-execute-action` is `t`?
Is there special case for disable code action that make the `lsp--select-action` behaves otherwise?

lsp-mode.el
```diff
@@ -5203,18 +5238,52 @@ It will show up only if current point has signature help."
         :context `( :diagnostics ,(lsp-cur-line-diagnostics)
                     ,@(when kind (list :only (vector kind))))))
 
-(defun lsp-code-actions-at-point (&optional kind)
-  "Retrieve the code actions for the active region or the current line."
-  (lsp-request "textDocument/codeAction" (lsp--text-document-code-action-params kind)))
+(defun lsp-code-actions-at-point (&optional kind enabled)
```
michaelpj, line 5241: While we're here... could we make this a list of kinds? The protocol supports that.

lsp-mode.el
```diff
@@ -5203,18 +5238,52 @@ It will show up only if current point has signature help."
...
+  "Retrieve the code actions for the active region or the current line.
+If ENABLED is specified, only return code actions that are not
+disabled."
+  (--doto (lsp-request "textDocument/codeAction" (lsp--text-document-code-action-params kind))
+    (when enabled
+      (cl-callf2 seq-remove #'lsp:code-action-disabled? it))))
 
 (defun lsp-execute-code-action-by-kind (command-kind)
   "Execute code action by name."
```
michaelpj, line 5250:
```suggestion
  "Execute code action by kind."
```

lsp-mode.el
```diff
@@ -5203,18 +5238,52 @@ It will show up only if current point has signature help."
...
-      (lsp-execute-code-action action)
-    (signal 'lsp-no-code-actions '(command-kind))))
+  (let* ((all-actions (->> (lsp-get-or-calculate-code-actions command-kind)
+                           (-filter (-lambda ((&CodeAction :kind?))
+                                      (and kind? (equal command-kind kind?))))))
+         (enabled-actions (-remove #'lsp:code-action-disabled? all-actions))
+         (action (lsp--select-action (or enabled-actions all-actions))))
+    (lsp-execute-code-action action)))
+
+(defun lsp-execute-code-action-by-type (kind &optional enabled)
```
michaelpj, line 5258: The difference between this and `lsp-execute-code-action-by-kind` isn't clear to me (or from the docs).

lsp-mode.el
```diff
@@ -5224,10 +5293,19 @@ It will show up only if current point has signature help."
       (funcall action-handler action)
     (lsp--send-execute-command command arguments?)))
 
-(lsp-defun lsp-execute-code-action ((action &as &CodeAction :command? :edit?))
+(lsp-defun lsp-execute-code-action ((action &as &CodeAction :command? :edit? :disabled?))
```
michaelpj, line 5296: Should this have an optional `kinds` argument? Generally, it seems like almost all the functions that operate on code actions could take an optional `kinds` argument, perhaps that would let us get away with fewer functions.

nbfalcon: Not in this case. `lsp-execute-code-action` should execute whatever action it is given. It is only additionally interactive.

lsp-mode.el
nbfalcon: `lsp-execute-code-action-by-type` matches base-types, i.e. `lsp-execute-code-action-by-type "source"` would match "source.organizeImports", while `-by-kind` matches exactly, i.e. the same example would yield nothing, and it would have to be `(lsp-execute-code-action-by-kind "source.organizeImports")`.

yyoncho (approved):

lsp-mode.el
michaelpj: I'm not sure this is a distinction the spec really knows about - certainly "base type" doesn't appear. They do talk about "base kinds", but I'm not sure they're meaningfully distinct from "kinds", it just indicates that they should be hierarchical parents of a set of kinds.

What about just having an `exact-match` argument to `lsp-execute-code-action-by-kind`? Or maybe we shouldn't even provide exact matching. The intention for spec methods that allow you to do kind-based filtering is to include kinds "under" the selected kinds (see e.g. https://github.com/microsoft/language-server-protocol/issues/970#issuecomment-633938017). Do we have a usecase for exact matching?

michaelpj:
It would be nice to get this in. The helpers for executing code actions by kind would be quite nice for clients.

PR #2406, opened 2020-12-14
Add support for Steep (Ruby type checker)

ybiquitous:
[Steep](https://github.com/soutaro/steep) is a Ruby static type checker created by soutaro who is a Ruby committer and the main developer of [RBS](https://github.com/ruby/rbs).

Steep has its own language server implementation and he has created the VSCode extension for Steep.
See <https://github.com/soutaro/steep#ides>.

### DIFF
--- a/clients/lsp-steep.el
+++ b/clients/lsp-steep.el
@@ -1,0 +1,65 @@
+;;; lsp-steep.el --- lsp-mode for Steep  -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020  Masafumi Koba
+
+;; Author: Masafumi Koba <ybiquitous@gmail.com>
+;; Keywords: languages
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; LSP client for Steep which is a Ruby type checker.
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(defgroup lsp-steep nil
+  "LSP support for Steep, using the Steep language server."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/soutaro/steep"))
+
+(defcustom lsp-steep-log-level "warn"
+  "Log level of Steep."
+  :type '(choice
+          (const "fatal")
+          (const "error")
+          (const "warn")
+          (const "info")
+          (const "debug"))
+  :group 'lsp-steep)
+
+(defcustom lsp-steep-use-bundler t
+  "Run Steep using Bunder."
+  :type 'boolean
+  :safe #'booleanp
+  :group 'lsp-steep)
+
+(defun lsp-steep--build-command ()
+  "Build a command to start the Steep language server."
+  (append
+    (if lsp-steep-use-bundler '("bundle" "exec"))
+    '("steep" "langserver")
+    (list (concat "--log-level=" lsp-steep-log-level))))
+
+(lsp-register-client
+ (make-lsp-client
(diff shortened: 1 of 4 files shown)


clients/lsp-steep.el
```diff
@@ -0,0 +1,65 @@
...
+
+(defcustom lsp-steep-use-bundler t
+  "Run Steep using Bunder."
+  :type 'boolean
+  :safe #'booleanp
+  :group 'lsp-steep)
+
+(defun lsp-steep--build-command ()
+  "Build a command to start the Steep language server."
+  (append
```
kiennq, line 52: nit: you can use `nconc` here

clients/lsp-steep.el
```diff
@@ -0,0 +1,65 @@
...
+    (if lsp-steep-use-bundler '("bundle" "exec"))
+    '("steep" "langserver")
+    (list (concat "--log-level=" lsp-steep-log-level))))
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection #'lsp-steep--build-command)
+  :major-modes '(ruby-mode enh-ruby-mode)
+  :priority -2
```
kiennq, line 61: The priority should be differnt from `sorbet-ls`

kiennq:
Can you help consolidate `lsp-steep` with `lsp-sorbet` in the same file, for example `lsp-ruby`?
So that all language servers for `ruby` will be easily viewed in same file and don't have priority conflict.

ybiquitous:
@kiennq I'd be happy to help! Implementing 3 tools (solargraph, sorbet, and steep) in one file sounds nice, but they have their own language server. Can we actually integrate them into one file for lsp-mode?

kiennq:
@ybiquitous Yes, that would be fine, each of them will be in separated sections.
You can take a look at `lsp-php`, which has `php-ls`, `intelephense`, `serenata` inside it.

ybiquitous:
Ah, I got it. I'll try it! Is it better to open a new pull request, or can I continue on this PR?

ybiquitous:
First, it seems good to create `lsp-ruby.el` integrating `lsp-solargraph.el` and `lsp-sorbet.el` on another PR. What do you think?

yyoncho:
@kiennq I am a bit concerned about removing lsp-solargraph.el it will be a breaking change.

kiennq:
@yyoncho We can always put a dummy `lsp-solargraph.el` there that require `lsp-ruby` for backward compatibility and  warning when attempt to use it. WDYT?
I think we're doing the same for `lsp.el` and also removing `lsp-intelephense` when refactoring `lsp-php` too.

yyoncho:
> @yyoncho We can always put a dummy `lsp-solargraph.el` there that require `lsp-ruby` for backward compatibility and warning when attempt to use it. WDYT?
> I think we're doing the same for `lsp.el` and also removing `lsp-intelephense` when refactoring `lsp-php` too.

In can be done (we should keep also lsp-packages references). In general, I am not big fan of grouping since the prop names became too long  `lsp-<language>-<server-id>-<property-name>`. Other than that I am fine with the change.

kiennq:
> In general, I am not big fan of grouping since the prop names became too long `lsp-<language>-<server-id>-<property-name>`.

It  can keep the original `lsp-<server-id>-<property-name>` though.
The main reason for grouping is for easier to determine the priority and also for new contributor to know exactly where to add new language server without worrying about boiler plate code when creating Elisp file.

ybiquitous:
@yyoncho @kiennq What status is this PR now?

yyoncho (approved):

yyoncho:
Can you resolve conflicts?

clients/lsp-steep.el
ybiquitous: The following change does not work on my environment, so I keep it as-is.

```diff
--- a/clients/lsp-steep.el
+++ b/clients/lsp-steep.el
@@ -49,7 +49,7 @@
 
 (defun lsp-steep--build-command ()
   "Build a command to start the Steep language server."
-  (append
+  (nconc
     (if lsp-steep-use-bundler '("bundle" "exec"))
     '("steep" "langserver")
     (list (concat "--log-level=" lsp-steep-log-level))))
```

clients/lsp-steep.el
ybiquitous: I've changed it to `:priority -3`.

https://github.com/emacs-lsp/lsp-mode/blob/d2d346c397c507b60dd3cd4a98d79b64bcb26028/clients/lsp-sorbet.el#L52

ybiquitous:
@yyoncho I've rebased the `master` branch and fixed the conflict. 👌

yyoncho:

PR #2407, opened 2020-12-14
Add the ability to run elixir tests from test lenses

wkirschbaum:
Tests running code lenses are now provided by elixir-lsp https://github.com/elixir-lsp/elixir-ls/pull/389 and implemented for vscode https://github.com/elixir-lsp/vscode-elixir-ls/pull/155. 

This adds the ability to switch on elixir test lenses and will run the tests using the compile function. 

test-lens-preview

### DIFF
--- a/clients/lsp-elixir.el
+++ b/clients/lsp-elixir.el
@@ -129,7 +129,8 @@ finding the executable with `exec-path'."
       "mix test --exclude test --include "
       test-command
       " "
-      file-path))
+      file-path
+      " --no-color"))
     file-path))
 
 (lsp-register-custom-settings
--- a/clients/lsp-elixir.el
+++ b/clients/lsp-elixir.el
@@ -100,6 +100,38 @@ finding the executable with `exec-path'."
   :group 'lsp-elixir
   :type 'file)
 
+(defcustom lsp-elixir-enable-test-lenses t
+  "Suggest Tests."
+  :type 'boolean
+  :group 'lsp-elixir
+  :package-version '(lsp-mode . "7.1"))
+
+(defun lsp-elixir--build-test-command (argument)
+  "Builds the test command from the ARGUMENT."
+  (let ((test-name (gethash "testName" argument))
+        (module (gethash "module" argument))
+        (describe (gethash "describe" argument)))
+    (cond (module (concat "\"" "module:" module "\""))
+          ((not test-name) (concat "\"" "describe:" describe "\""))
+          (describe (concat "\"" "test:test " describe " " test-name "\"" ))
+          (t (concat "\"" "test:test " test-name "\"" )))))
+
+(lsp-defun lsp-elixir--run-test ((&Command :arguments?))
+  "Runs tests."
+  (let* ((argument (lsp-seq-first arguments?))
+         (file-path (gethash "filePath" argument))
+         (test-command (lsp-elixir--build-test-command argument)))
+    (compile
+     (concat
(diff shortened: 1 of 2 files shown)


clients/lsp-elixir.el
```diff
@@ -100,6 +100,39 @@ finding the executable with `exec-path'."
...
 
+(defcustom lsp-elixir-enable-test-lenses t
+  "Suggest Tests."
+  :type 'boolean
+  :group 'lsp-elixir
+  :package-version '(lsp-mode . "7.1"))
+
+(defun lsp-elixir--build-test-command (argument)
+  "Builds the test command from the ARGUMENT."
+  (let ((test-name (gethash "testName" argument))
```
yyoncho, line 111: using gethash is not allowed. All the access to the data structures that come from the language server should happen either by defined `lsp-interface` or by using `lsp-get`. This will allow us to migrate to using plists instead of hashtables. You may refer to rust-analyzer for example: https://github.com/yyoncho/lsp-mode/blob/headerline-errors/lsp-protocol.el#L278

wkirschbaum: thanks, will have a look.

wkirschbaum: @yyoncho I changed to lsp-get, but will figure out how lsp-interface works tomorrow if it might be a better fit.

clients/lsp-elixir.el
```diff
@@ -100,6 +100,33 @@ finding the executable with `exec-path'."
...
+        (describe (lsp-get argument :describe)))
+    (cond (module (concat "\"" "module:" module "\""))
+          ((not test-name) (concat "\"" "describe:" describe "\""))
+          (describe (concat "\"" "test:test " describe " " test-name "\"" ))
+          (t (concat "\"" "test:test " test-name "\"" )))))
+
+(lsp-defun lsp-elixir--run-test ((&Command :arguments?))
+  "Runs tests."
+  (let* ((argument (lsp-seq-first arguments?))
+         (file-path (lsp-get argument :filePath))
```
kiennq, line 122: is `filePath` a new member of `Command`?

wkirschbaum: I have no clue, what do you mean? I am pretty new to elisp and lsp-mode, so please excuse my lack of knowledge.

yyoncho: @kiennq `arguments?` is a list of `any`, so you can have anything in it.

clients/lsp-elixir.el
yyoncho: `lsp-get` is good enough for non-core components given the fact that this structure will be accessed only in a few places.

As a side note, most likely vscode supports running the tests in the debugger, right? We can implement that as well.

wkirschbaum: Do you mean dap-mode? I have not really worked with it before, but can have a look. Ideally i would not want to click to run tests, but this is the baseline work for future capabilities it seems.

yyoncho: > Do you mean dap-mode?

yes.

> Ideally i would not want to click to run tests, but this is the baseline work for future capabilities it seems.

There is `lsp-avy-lens`. And then you will be able to re-run with `M-x recompile`.

yyoncho (approved):

yyoncho:

wkirschbaum:
@yyoncho @axelson thanks for helping me contribute.

PR #2411, opened 2020-12-15
Svelte LS related configuration

### LINKED ISSUE #2409: Map svelte-ls config keys to user options
**Feature request**
It would be great if the configuration for the Svelte language server could be adjusted by setting user options, similar to the setup for many other language servers (e.g., to be specific, gopls).

Concretely, it would be great to be able to disable certain warnings from Svelte. Currently this can only be done through the language server configuration. I reported the problem upstream at https://github.com/sveltejs/language-tools/issues/650#issuecomment-734930568 and was pointed to [the documentation for the VS Code extension](https://marketplace.visualstudio.com/items?itemName=svelte.svelte-vscode#settings) which contains a list of allowed configuration keys.

I would be happy to implement this, but don't currently have the bandwidth, so I thought I would at least create the issue for now.

`svelte-ls` (https://github.com/sveltejs/language-tools)

**OS**
Ubuntu 20.04

yyoncho:
- fixes #2409

----

#

### DIFF
--- a/clients/lsp-svelte.el
+++ b/clients/lsp-svelte.el
@@ -186,7 +186,10 @@
   "Svelte compiler warning codes to ignore or to treat as errors.
 Example: { 'css-unused-selector': 'ignore', 'unused-export-let':
 'error'}"
-  :type 'plist
+  :type '(alist :key-type (string :tag "Warning code")
+                :value-type (choice
+                             (const :tag "Ignore" "ignore")
+                             (const :tag "Treat as error" "error")))
   :package-version '(lsp-mode . "7.1.0"))
 
 (defcustom lsp-svelte-plugin-svelte-format-enable t
--- a/clients/lsp-svelte.el
+++ b/clients/lsp-svelte.el
@@ -37,6 +37,226 @@
                 '(:npm :package "svelte-language-server"
                        :path "svelteserver"))
 
+(defcustom lsp-svelte-plugin-typescript-enable t
+  "Enable the TypeScript plugin"
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1.0"))
+
+(defcustom lsp-svelte-plugin-typescript-diagnostics-enable t
+  "Enable diagnostic messages for TypeScript"
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1.0"))
+
+(defcustom lsp-svelte-plugin-typescript-hover-enable t
+  "Enable hover info for TypeScript"
+  :type 'boolean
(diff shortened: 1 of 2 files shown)


kiennq (approved):

clients/lsp-svelte.el
```diff
@@ -37,17 +37,263 @@
...
+(defcustom lsp-svelte-plugin-svelte-diagnostics-enable t
+  "Enable diagnostic messages for Svelte"
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1.0"))
+
+(defcustom lsp-svelte-plugin-svelte-compiler-warnings nil
+  "Svelte compiler warning codes to ignore or to treat as errors.
+Example: { 'css-unused-selector': 'ignore', 'unused-export-let':
+'error'}"
+  :type 'plist
```
raxod502, line 189: Confirmed working with this amendment to the compiler warning format! Thanks so much!

```suggestion
  :type '(alist :key-type (string :tag "Warning code")
                :value-type (choice
                             (const :tag "Ignore" "ignore")
                             (const :tag "Treat as error" "error")))
```

yyoncho: Applied.

yyoncho: Actually, native JSON cannot serialize alists with string keys. We should either change keys to symbols or use hts.

clients/lsp-svelte.el
```diff
@@ -37,17 +37,266 @@
...
+
+(defcustom lsp-svelte-plugin-svelte-compiler-warnings nil
+  "Svelte compiler warning codes to ignore or to treat as errors.
+Example: { 'css-unused-selector': 'ignore', 'unused-export-let':
+'error'}"
+  :type '(alist :key-type (string :tag "Warning code")
+                :value-type (choice
+                             (const :tag "Ignore" "ignore")
+                             (const :tag "Treat as error" "error")))
+  :package-version '(lsp-mode . "7.1.0"))
```
raxod502, line 193: might as well go all the way with symbols, then!

```suggestion
(defcustom lsp-svelte-plugin-svelte-compiler-warnings nil
  "Svelte compiler warning codes to ignore or to treat as errors.
Example: '((css-unused-selector . ignore) (unused-export-let . error))"
  :type '(alist :key-type (symbol :tag "Warning code")
                :value-type (choice
                             (const :tag "Ignore" ignore)
                             (const :tag "Treat as error" error)))
  :package-version '(lsp-mode . "7.1.0"))
```

PR #2414, opened 2020-12-16
lsp-lua.el: Fix lua-lsp connection commands are unchangable

sunlin7:
Currently the commands for lua-lsp is inited once, while it should sensitive for custom settings changes and applied in new-connection.
Here I just follow python-lsp, wrapping the commands into lambda function for :new-connection, then lua-lsp will work as expect when customer change the lsp-commands.
Python lsp-pyls: https://github.com/emacs-lsp/lsp-mode/blob/d788a13b26289d3bd88dc07f3e2e9c3c3d1fdf9d/clients/lsp-pyls.el#L473-L474.

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -71,7 +71,7 @@
 
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection lsp-clients-emmy-lua-command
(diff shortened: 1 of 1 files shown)


kiennq (approved):

jcs090218 (approved):

jcs090218:
LGTM, Thanks!

sunlin7:
Thanks

PR #2418, opened 2020-12-18
[feature] Add hooks to run after text edit

ericdallo:
Every time I `lsp-rename` a symbol, I need to check other buffers and save them to make the code consistent and fix lsp errors, for that, I use this advice for months as a workaround:
```elisp
 (advice-add #'lsp-rename :after (lambda (&rest _) (projectile-save-project-buffers)))
```

It looks to me a good option to add this to `lsp-mode`, making it optional for users that wants this same behavior.

I know emacs user doesn't like things changing buffers automatically, but IMO it looks IDE like feature, also I'm adding it with the default as `false`.

This works too for formatting buffers, which is recommended for some languages like Dart

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -664,6 +664,13 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-auto-save-changed-buffers nil
+  "If non-nil, `lsp-mode' will save the buffer after apply text edits.
+Useful for multiple file changes like `lsp-rename`"
+  :type 'boolean
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
 (defcustom lsp-modeline-code-actions-enable t
   "Whether to show code actions on modeline."
   :type 'boolean
@@ -4056,7 +4063,9 @@ LSP server result."
                            (-when-let ((&SnippetTextEdit :range (&RangeToPoint :start)
                                                          :insert-text-format? :new-text) edit)
                              (when (eq insert-text-format? lsp/insert-text-format-snippet)
-                               (lsp--expand-snippet new-text start (+ start (length new-text)))))))))
+                               (lsp--expand-snippet new-text start (+ start (length new-text))))))
+                         (when lsp-auto-save-changed-buffers
+                           (save-buffer)))))
           (undo-amalgamate-change-group change-group)
           (progress-reporter-done reporter))))))
 
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -664,10 +664,51 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
-(defcustom lsp-auto-save-changed-buffers nil
-  "If non-nil, `lsp-mode' will save the buffer after apply text edits.
-Useful for multiple file changes like `lsp-rename`"
-  :type 'boolean
+(defcustom lsp-before-save-text-edit-applied-hook nil
+  "Hooks to run when text edit is applied from a before save operation."
+  :type 'hook
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-code-action-text-edit-applied-hook nil
+  "Hooks to run when text edit is applied from a code action operation."
+  :type 'hook
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-color-presentation-text-edit-applied-hook nil
+  "Hooks to run when text edit is applied from a color presentation operation."
+  :type 'hook
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-completion-text-edit-applied-hook nil
+  "Hooks to run when text edit is applied from a completion operation."
+  :type 'hook
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-completion-cleanup-text-edit-applied-hook nil
+  "Hooks to run when text edit is applied from a completion cleanup operation."
+  :type 'hook
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-format-text-edit-applied-hook nil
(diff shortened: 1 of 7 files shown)


lsp-mode.el
```diff
@@ -4056,7 +4063,9 @@ LSP server result."
                            (-when-let ((&SnippetTextEdit :range (&RangeToPoint :start)
                                                          :insert-text-format? :new-text) edit)
                              (when (eq insert-text-format? lsp/insert-text-format-snippet)
-                               (lsp--expand-snippet new-text start (+ start (length new-text)))))))))
+                               (lsp--expand-snippet new-text start (+ start (length new-text))))))
+                         (when lsp-auto-save-changed-buffers
```
kiennq, line 4067: This apply for every edits in a file, which can be expensive, can you it just one when the apply text edits finished?

kiennq, line 4067: This applied for all text edits operation, which sometimes undesirable (for example, every completion inserted will automatically save buffer).
Can you just put the code for auto-saving for `lsp-rename` only?

yyoncho:
Can we introduce a hook `lsp-after-apply-edits-hook` hook instead?

ericdallo:
I thought about a hook, but user would need to manually listen to the hook and save the buffer, right?
Also like @kienqq said, it would be nice to just enable this for lsp-rename, but I don't know how would be to trigger this hook for each buffer, WDYT?

nbfalcon:
@ericdallo we could add an optional parameter to `lsp--workspace-apply-text-edit` called `SAVE`, which, if set, saves each modified buffer.

ericdallo:
Yes, it's a option, but how this would fit with the hook mentioned by @yyoncho ?

yyoncho:
The idea with the hook is that if we continue to add settings there are virtually a million options that the users might want to customize and if we don't cut them at some point, lsp-mode code will explode. Of course, there should be a balance. 

> I thought about a hook, but user would need to manually listen to the hook and save the buffer, right?

Yes. Also, it will allow the user to add other functionality.

> Also like @kienqq said, it would be nice to just enable this for lsp-rename, but I don't know how would be to trigger this hook for each buffer, WDYT?

We can pass a parameter to indicate that we are in the rename case(or have a generic parameter `operation`). I don't get what is the issue with triggering this for each buffer? For the operation, you may use dynamic parameter, e. g. `lsp-operation-context` which may be useful for scriptability.

ericdallo:
Got it, it looks a good approach indeed, I'll try to refactor for something like you suggested

lsp-mode.el
```diff
@@ -3872,8 +3882,9 @@ Need to handle the case when FILE indicates virtual buffer."
          ,@body)
      ,@body))
 
-(defun lsp--apply-text-document-edit (edit)
+(defun lsp--apply-text-document-edit (edit &optional operation)
```
kiennq, line 3885: This require rather big change all the way until `lsp--apply-text-edits`.
Just my idea, would it be better to modify `edit` itself to include new value (for example `:_op?`) instead? And let `lsp--apply-text-edits` pull that value out and pass to the hook.
That can be nicely integrated with `lsp-protocol` too.

lsp-mode.el
```diff
@@ -664,6 +664,13 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-text-edit-applied-hook nil
```
kiennq, line 667: We have `lsp-before-apply-edits-hook`, shouldn't this be `lsp-after-apply-edits-hook`?

lsp-mode.el
ericdallo: I see your point, but op is not part of the protocol, not sure if this would trigger a whole new discussion

kiennq: Yeah, I think we've already adding new member to that structure started with `_`, with that it will not collide with protocol in any way.
The other approach is using dynamic binding like @yyoncho suggested.

yyoncho: IMHO after @ericdallo fixed all these places current approach seems to be good enough. Although he touched a lot of places the change is low risk. @kiennq are you fine to proceed as it is?

lsp-mode.el
```diff
@@ -3844,24 +3851,27 @@ in that particular folder."
...
-        (->> document-changes
-             (seq-filter (-lambda ((&CreateFile :kind))
-                           (not (or (not kind) (equal kind "edit")))))
-             (seq-do #'lsp--apply-text-document-edit)))
-    (lsp-map
-     (lambda (uri text-edits)
-       (with-current-buffer (-> uri lsp--uri-to-path find-file-noselect)
-         (lsp--apply-text-edits text-edits)))
-     changes?)))
+(defun lsp--apply-workspace-edit (workspace-edit &optional operation)
```
kiennq, line 3854: Can't we do
```suggestion
(lsp-defun lsp--apply-workspace-edit ((&WorkspaceEdit :document-changes? :changes?) &optional operation)
```

kiennq: nvm, the `&optional` is bound wrong here.

lsp-mode.el
```diff
@@ -3844,24 +3851,27 @@ in that particular folder."
...
+  "Apply the WorkspaceEdit object WORKSPACE-EDIT.
+OPERATION is symbol representing the source of this text edit."
+  (let ((document-changes? (lsp:workspace-edit-document-changes? workspace-edit))
```
kiennq, line 3857: `-let`?

lsp-mode.el
kiennq: Okay, I'm fine with that.

ericdallo:
Are we good to go here @kiennq @yyoncho ?

lsp-mode.el
```diff
@@ -664,6 +664,13 @@ If this is set to nil, `eldoc' will show only the symbol information."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-after-apply-edits-hook nil
```
yyoncho, line 667: as per conventions hooks that receive arguments should be named `*-functions`. We don't follow it(initially, I didn't know about that convention).

yyoncho (approved):

lsp-mode.el
ericdallo: Also, it does not seem a good convention IMO, since this hook has nothing related to `functions`

kiennq (approved):

PR #2424, opened 2020-12-20
[lsp-lua]: Fix connection commands not following custom changes

sunlin7:
* lsp-lua.el: `lsp-clients-emmy-lua-command' is assemblied at initilize, won't
change after user customize the `lsp-clients-emmy-lua-jar-path'. This changes
will make `:new-command' senssible the custom change and be applied with new
connection.

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -55,9 +55,7 @@
   :risky t
   :type  '(repeat string))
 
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
Thank you!

sunlin7:
Thank you ~

PR #2425, opened 2020-12-20
Fix lsp-vhdl executable PATH search

raxod502:
Using f-executable? is not appropriate for finding the lsp-vhdl server executable, as it does not do a PATH search -- rather it just checks whether a literal file path is executable. We want to use executable-find instead, which allows lsp-mode to correctly find the server executable from the PATH.

The concrete impact of this is that the language server can be started, rather than failing with the "server is not installed, and automatic installation is not enabled" message.

From my browsing of the changelog, it seems like bugfixes this small are not recorded there. But let me know if I should update it!


raxod502:
I think the CI failure is not related to this change.

yyoncho (approved):

yyoncho:
Thank you!

PR #2432, opened 2020-12-22
[icons] Add lsp-icons to centralize icons management

### LINKED ISSUE #2430: Is there a way to remove icons from the Header Line?
Now that Header Line is enabled by default, is there away to remove/disable icons/images and keep only text? They are too big and ugly, and `treemacs-resize-icons` has no effect at all, and if can't get them get smaller I'd rather just remove them totally.

I got the Header Line disabled till I can find a solution, any ideas?

ericdallo:
Fix #2430

* Add `lsp-icons.el`, the idea is to every feature that needs icons on lsp-mode, should use it since will check if user supports third party packages and if the icon to feature  is enabled.
* Change `lsp-headerline.el` to use `lsp-icons.el`

### DIFF
--- a/lsp-icons.el
+++ b/lsp-icons.el
@@ -1,0 +1,68 @@
+;;; lsp-icons.el --- LSP icons management -*- lexical-binding: t; -*-
+;;
+;; Copyright (C) 2020 emacs-lsp maintainers
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;;  LSP icons management
+;;
+;;; Code:
+
+(declare-function all-the-icons-material "ext:all-the-icons" t t)
+(declare-function lsp-treemacs-symbol-icon "ext:lsp-treemacs" (kind))
+(declare-function lsp-treemacs-get-icon "ext:lsp-treemacs" (icon-name))
+
+(defun lsp-icons--fix-image-background (image)
+  "Fix IMAGE background if it is a file otherwise return as an icon."
+  (if image
+      (let ((display-image (get-text-property 0 'display image)))
+        (if (and (listp display-image)
+                 (plist-member (cl-copy-list (cl-rest display-image)) :type))
+            (propertize " " 'display
+                        (cl-list* 'image
+                                  (plist-put
+                                   (cl-copy-list
+                                    (cl-rest display-image))
+                                   :background (face-attribute 'header-line :background nil t))))
+          (if (stringp display-image)
+              (replace-regexp-in-string "\s\\|\t" "" display-image)
+            (replace-regexp-in-string "\s\\|\t" "" image))))
+    ""))
+
+(defun lsp-icons-get-by-file-ext (file-ext)
+  "Get an icon by file FILE-EXT."
+  (when (and file-ext
+             (require 'lsp-treemacs nil t))
+    (lsp-icons--fix-image-background
+     (lsp-treemacs-get-icon file-ext))))
+
+(defun lsp-icons-get-by-symbol-kind (kind)
+  "Get an icon by symbol KIND."
+  (when (and kind
+             (require 'lsp-treemacs nil t))
+    (lsp-icons--fix-image-background
+     (lsp-treemacs-symbol-icon kind))))
+
+(defun lsp-icons-all-the-icons-material-icon (icon-name face fallback)
+  "Get a material icon from all-the-icons by ICON-NAME using FACE.
+Fallback to FALLBACK string if not found or not available."
+  (if (require 'all-the-icons nil t)
+      (all-the-icons-material icon-name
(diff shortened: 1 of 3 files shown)


lsp-icons.el
```diff
@@ -0,0 +1,93 @@
...
+             (require 'lsp-treemacs nil t))
+    (lsp-icons--fix-image-background
+     (lsp-treemacs-symbol-icon kind))))
+
+(defun lsp-icons-all-the-icons-material-icon (icon-name face fallback &optional feature)
+  "Get a material icon from all-the-icons by ICON-NAME using FACE.
+Fallback to FALLBACK string if not found or not available.
+FEATURE is the feature that will use the icon which we should check
+if its enabled."
+  (if (and (require 'all-the-icons nil t)
```
kiennq, line 86: nit: Can we write this slightly differently by put all `require` expression in global scope and check for function available via `fboundp` instead?
That would ensure if the optional library ever change the function name, our package will not be broken

ericdallo: Not sure I understand, how would it be?

jcs090218: I would suggest using `featurep` instead of `fboundp`? `featurep` is like `require` but it only check for module existence and not going to load it. Anyway, I think @kiennq has the point here so we can keep the code consistent.

```el
(require 'all-the-icons nil t)  ; load `all-the-icons` in global level

(if (featurep 'all-the-icons)
  ;; other execution...
)
```

lsp-icons.el
```diff
@@ -0,0 +1,93 @@
...
+(defcustom lsp-headerline-breadcrumb-icons-enable t
+  "If non-nil, icons support is enabled for headerline-breadcrumb."
+  :type 'boolean
+  :group 'lsp-icons)
+
+(declare-function all-the-icons-material "ext:all-the-icons" t t)
+(declare-function lsp-treemacs-symbol-icon "ext:lsp-treemacs" (kind))
+(declare-function lsp-treemacs-get-icon "ext:lsp-treemacs" (icon-name))
+
+(defun lsp-icons--enabled-for-feature (feature)
```
yyoncho, line 38: the issue is that the features that will benefit from that are not part of lsp-mode project, so we cannot enumerate them here.

I propose the following structure(or at least using this as base).

1. Have a functions get-file-icon-ALL-THE-ICONS/treemacs/whatever.
2. Have a property lsp-FEATURE-file-icon-function which defaults to something that make sense. The same for symbols. 

The global killswitch will be in the default function. At the same time, one could configure a particular feature to use different set of icons(e. g. headerlien to use all the icons, dap-mode to use smt else).

WDYT?

lsp-icons.el
wyuenho: Or better yet, just feature detect the function via `functionp` instead of the package, unless you are using a lot of it.

lsp-icons.el
```diff
@@ -0,0 +1,93 @@
...
+            (replace-regexp-in-string "\s\\|\t" "" image))))
+    ""))
+
+(defun lsp-icons-get-by-file-ext (file-ext &optional feature)
+  "Get an icon by file FILE-EXT.
+FEATURE is the feature that will use the icon which we should check
+if its enabled."
+  (when (and file-ext
+             (lsp-icons--enabled-for-feature feature)
+             (require 'lsp-treemacs nil t))
```
wyuenho, line 67: Don't require lsp-treemacs, just check for the function definition with `(functionp lsp-treemacs...`. Same for the other require below.

lsp-icons.el
```diff
@@ -0,0 +1,96 @@
...
+            (replace-regexp-in-string "\s\\|\t" "" image))))
+    ""))
+
+(defun lsp-icons-get-by-file-ext (file-ext &optional feature)
+  "Get an icon by file FILE-EXT.
+FEATURE is the feature that will use the icon which we should check
+if its enabled."
+  (when (and file-ext
+             (lsp-icons--enabled-for-feature feature)
+             (featurep 'lsp-treemacs-get-icon))
```
wyuenho, line 70: `(featurep)` is used to check for module names exposed by `(provide)`, not functions. Please change to `(functionp)` here and elsewhere.

lsp-icons.el
```diff
@@ -0,0 +1,96 @@
...
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+;;
+;;; Commentary:
+;;
+;;  LSP icons management
+;;
+;;; Code:
+
+(require 'lsp-treemacs nil t)
+(require 'all-the-icons nil t)
```
wyuenho, line 25: Please just delete these 2 requires. You don't need it as all the function calls are guard, and there are legitimate reasons for the users to install but not require these packages. Please don't overrule the end-user.

ericdallo:
Done @wyuenho

lsp-icons.el
wyuenho: I don't even know why this function is needed. Isn't providing overridable functions sufficient?

wyuenho:
What's holding up the merge of this?

lsp-icons.el
nbfalcon: @jcs090218 `featurep` does not check if the package is installed, but rather if it is *loaded*. DOOM lazy-loads `lsp-mode`, for example, so when I start it, `(featurep 'lsp-mode)` yields nil, even though it is installed. `fboundp` is better to check for the existence of packages, because it returns true for `auto-loaded` functions, which works even if the package isn't loaded yet.

ericdallo:
I think @yyoncho suggestion, I'm not sure how to implement it

yyoncho:
We can merge it as it is and then work on that.

wyuenho:
@ericdallo  I'd just remove `lsp-icons--enabled-for-feature` and make a global `lsp-use-icons` defcustom to deal with that. I can't think of a compelling reason to use icons in one feature but not another.

ericdallo:
Ok, rebased and I'm merging it to not become stale, then we can work on improvements for that

PR #2438, opened 2020-12-24
Add tailwindcss support

itome:
## Tailwind css support
- https://marketplace.visualstudio.com/items?itemName=bradlc.vscode-tailwindcss
- Runs as add on
- Tailwind intellisense has no installable binary, so I directory download it from vscode extension (Same as eslint lsp)

### DIFF
--- a/clients/lsp-tailwindcss.el
+++ b/clients/lsp-tailwindcss.el
@@ -181,9 +181,9 @@
   :activation-fn (lambda (filename &optional _)
                    (when lsp-tailwindcss-enable
                      (or (string-match-p (rx (one-or-more anything) "."
-                                             (or "ts" "js" "jsx" "tsx" "html" "vue")eos)
+                                             (or "ts" "js" "jsx" "tsx" "html" "vue" "css")eos)
                                          filename)
-                         (derived-mode-p 'js-mode 'js2-mode 'typescript-mode 'html-mode))))
+                         (derived-mode-p 'js-mode 'js2-mode 'typescript-mode 'html-mode 'css-mode))))
   :notification-handlers (ht
                           ("tailwindcss/configUpdated" #'ignore)
                           ("tailwindcss/configError" #'ignore)
--- a/clients/lsp-tailwindcss.el
+++ b/clients/lsp-tailwindcss.el
@@ -1,0 +1,212 @@
+;;; lsp-tailwindcss.el --- lsp-mode tailwindcss integration    -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2020  Takeshi Tsukamoto
+
+;; Author: Takeshi Tsukamoto <dev@itome.team>
+;; Keywords: languages
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;;
+
+;;; Code:
+
+(require 'lsp-protocol)
+(require 'lsp-mode)
+
+
+(defgroup lsp-tailwindcss nil
+  "Tailwind CSS language server group."
+  :group 'lsp-mode
+  :link '(url-link "https://tailwindcss.com/docs/intellisense"))
+
+(defcustom lsp-tailwindcss-unzipped-path (f-join lsp-server-install-dir "tailwindcss/unzipped")
+  "The path to the file in which `tailwindcss' will be stored."
+  :type 'file
+  :group 'lsp-tailwindcss
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-enable t
+  "Controls whether tailwindcss is enabled for JavaScript files or not."
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-server-command `("node"
+                                            ,(f-join lsp-tailwindcss-unzipped-path "extension/dist/server/index.js")
+                                            "--stdio")
+  "Command to start tailwindcss intellisense server."
+  :risky t
+  :type '(repeat string)
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-include-languages '()
+  "Enable features in languages that are not supported by default.
+  Add a mapping here between the new language and an already supported language.
+  E.g.: `{\"plaintext\": \"html\"}`"
+  :type 'alist
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-emmet-completions nil
+  "Enable completions when using Emmet-style syntax,
+  for example div.bg-red-500.uppercase. Default: false"
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-color-decorators "inherit"
+  "Controls whether the editor should render inline color decorators
+  for Tailwind CSS classes and helper functions."
+  :type '(choice
+          (const "inherit")
+          (const "on")
+          (const "off"))
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-validate t
+  "Enable linting. Rules can be configured individually
+  using the `tailwindcss.lint.*` settings"
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-tailwindcss-lint-css-conflict "warning"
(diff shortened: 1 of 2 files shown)


clients/lsp-tailwindcss.el
```diff
@@ -0,0 +1,212 @@
...
+
+(defun lsp-tailwindcss--get-configuration ()
+  ;; FIXME: Handle get configuration request
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/extension.ts#L160
+  t)
+
+(defun lsp-tailwindcss--get-document-symbols ()
+  ;; FIXME: Handle get document symbols request
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/extension.ts#L168
+  t)
```
itome, line 178: `HELP NEEDED` I have no ideas how to implement this handlers. Could anyone help me?

yyoncho: This one is: the tailwindcss server is calling the typescript-language-server to get the symbol information. In this case you should use textDocument/documentSymbols call to obtain the data and return it to the tailwindcss server. Most likely just doing `lsp--get-document-symbols` will work but it may need rework to make it async.

clients/lsp-tailwindcss.el
```diff
@@ -0,0 +1,212 @@
...
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lib/registerColorDecorator.ts#L125
+  t)
+
+(defun lsp-tailwindcss--config-error ()
+  ;; FIXME: Update decorations in editor and show error message, see original implementation
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lib/registerColorDecorator.ts#L129
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lib/registerConfigErrorHandler.ts#L5
+  t)
+
+(defun lsp-tailwindcss--get-configuration ()
```
yyoncho, line 170: this one you will have to return a map with tab size in it: `(symbol-value (lsp--get-indent-width major-mode))`

clients/lsp-tailwindcss.el
```diff
@@ -0,0 +1,212 @@
...
+
+(defun lsp-tailwindcss--config-updated ()
+  ;; FIXME: Update decorations in editor, see original implementation
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lib/registerColorDecorator.ts#L125
+  t)
+
+(defun lsp-tailwindcss--config-error ()
+  ;; FIXME: Update decorations in editor and show error message, see original implementation
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lib/registerColorDecorator.ts#L129
+  ;; https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lib/registerConfigErrorHandler.ts#L5
```
yyoncho, line 167: for these you will have to port updateDecorationsInEditor - here I need more specific questions because it is a huge method

clients/lsp-tailwindcss.el
itome: The original implementation seems to be get all configuration related to tailwindcss by calling `...Workspace.getConfiguration('tailwindCSS', scope)`.
Is there any function in lsp-mode equivalent to that of vscode?

yyoncho: `lsp-configuration-section`

itome:
@yyoncho ↑ Does this commit fit your intentions?

yyoncho:
> @yyoncho ↑ Does this commit fit your intentions?

Looks good, but the best way to ensure that it works fine is to compare the emacs behaviour with vscode one. I also tend to compare the client/server logs in both.

clients/lsp-tailwindcss.el
itome: `updateDecorationInEditor` provides syntax highlight (see the [doc](https://tailwindcss.com/docs/intellisense)).
Is lsp-mode support syntax highlighting?

itome:
Thanks, I'll check it later.
And I want to implement hover preview feature in `tailwindcss-intellisense`

Here is original implementation.
https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lsp/server.ts#L216

`lsp-mode` (or `lsp-ui` ?) support this?

clients/lsp-tailwindcss.el
yyoncho: We do syntax highlighting via sematic tokens. This looks more like textDocument/documentColor handling.

yyoncho:
> Here is original implementation.
> https://github.com/tailwindlabs/tailwindcss-intellisense/blob/b3c17cf5c9d1d74091a4a5ee0504696bb2ce1c46/packages/tailwindcss-intellisense/src/lsp/server.ts#L216
> 
> `lsp-mode` (or `lsp-ui` ?) support this?

You have linked the server, not the client. This is normal onHover which is supported in both lsp-mode via eldoc and in lsp-ui

clients/lsp-tailwindcss.el
itome: So it's already implemented and I have nothing to do about this?

yyoncho:
Yes.

itome:
In my environment, tailwind hover preview seems not working. Any additions configuration is required for hover preview?

yyoncho:
Check what textDocument/hover is returning, call it explicitly via M-x lsp-describe-thing-at-point

itome:
`lsp-describe-thing-at-point` timeout with following error message

```
lsp-request: Timeout while waiting for response. Method: textDocument/hover.

```

yyoncho:
> `lsp-describe-thing-at-point` timeout with following error message
> 
> ```
> lsp-request: Timeout while waiting for response. Method: textDocument/hover.
> ```

I am not quite sure what might be causing that - you have to check the server.

Reductions:
Is there any progress on this pull request?

merrickluo:
for anyone interested, I made a tailwindcss lsp client [here](https://github.com/merrickluo/lsp-tailwindcss), didn't notice this PR when I started 😥️. Although I haven't implemented all the options this PR has, but it works. feel free to try it out!

nicwhitts:
Could someone help me out, im new to emacs i'v moved from vscode. 
I can set extra classNames on vscode quite easily ;

```
  "tailwindCSS.experimental.classRegex": ["tailwind\\('([^)]*)\\')", "'([^']*)'"]
```
How can i set this in emacs?

I'v tried:
```
(defcustom lsp-tailwindcss-experimental-class-regex '("tailwind\\('([^)]*)\\')" "'([^']*)'")
  ""
  :type '(repeat string)
  :package-version '(lsp-mode . "7.1"))

(require 'lsp-protocol)
(require 'lsp-mode)
(lsp-register-custom-settings '(("tailwindCSS.experimental.classRegex" lsp-tailwindcss-experimental-class-regex )))


```
But the lsp server hangs with an error `Wrong type argument: symbolp, "tailwind\\('([^)]*)\\')"
`

What's wrong here?

yyoncho:
> What's wrong here?

It should be vector, not list.

nicwhitts:
> > What's wrong here?
> 
> It should be vector, not list.

Could you give an example with the above as its not behaving as expected

```
(vector "tailwind\\('([^)]*)\\')" "'([^']*)'")
```
doesn't work

I seen there is potentially a type of lsp-vector-string, so i also tried:
```
["tailwind\\('([^)]*)\\')", "'([^']*)'"]

```

Which will only use the first string in the array?

VisionaryAppDev:
Just curious if tailwind support will come to life anytime soon? it's been a while now and haven't seen any update at all.

VisionaryAppDev:
Yes it is a great project. But I found multiple tailwind projects is suspended and mentioned this pull request. And since I saw this pull request is inactive for so long so i just wondering if it's going to merge and bring lsp mode out of the box. And I wanted to see if this one will be faster? I have tried lsp-tailwindcss but it's quite slow. merrickluo mentioned that it's because Emacs is using a single thread.

yyoncho:
> And I wanted to see if this one will be faster?

@VisionaryAppDev - the speed is not relevant to the server configuration. If M-x lsp-doctor is all green then there isn't a lot we can do ATM.

carusog:
@merrickluo https://github.com/merrickluo/lsp-tailwindcss is awesome, but, AFAIK, it's missing auto-completion of classes in Svelte mode, as well as color preview.
It would be great to have a complete solution to support TW out of the box. Alternatively, we could help @merrickluo with their implementation.

PR #2439, opened 2020-12-24
[svelte] Workaround: Use ht-get over ht-get* because of upstream bug

leungbk:
See https://github.com/Wilfred/ht.el/issues/38

----

#

### DIFF
--- a/clients/lsp-svelte.el
+++ b/clients/lsp-svelte.el
@@ -270,9 +270,12 @@ Example: '((css-unused-selector . ignore) (unused-export-let . error))"
                             "svelte"))
   :initialization-options
   (lambda ()
-    (list :config (ht-get* (lsp-configuration-section "svelte.plugin")
-                           "svelte"
-                           "plugin")
+    ;; XXX: workaround for https://github.com/Wilfred/ht.el/issues/38
+    ;; Use `ht-get*' instead of `lsp--ht-get' when
+    ;; https://github.com/Wilfred/ht.el/pull/39 is merged
+    (list :config (lsp--ht-get (lsp-configuration-section "svelte.plugin")
+                               "svelte"
+                               "plugin")
           :prettierConfig (lsp-configuration-section "prettier")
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
as a side note, there is lsp--ht-get which I was planning to delete.

leungbk:
> as a side note, there is lsp--ht-get which I was planning to delete.

OK, I used that instead.

clients/lsp-svelte.el
```diff
@@ -270,9 +270,12 @@ Example: '((css-unused-selector . ignore) (unused-export-let . error))"
                             "svelte"))
   :initialization-options
   (lambda ()
-    (list :config (ht-get* (lsp-configuration-section "svelte.plugin")
```
nbfalcon, line 273: Are you sure this is needed? `ht-get*` has a compiler macro which seems to work. If I expand that `ht-get*` expression in `macrostep` with `macrostep-expand-compiler-macros` enabled, I get:
```emacs-lisp
(ht-get
 (ht-get
  (lsp-configuration-section "svelte.plugin")
  "svelte")
 "plugin")
 ```
 which is correct.

Comment (leungbk) [reply clients/lsp-svelte.el:273]: Here's what I get in `ielm` when testing the two variants:

```
ELISP> (ht-keys (ht-get* (lsp-configuration-section "svelte.plugin")
                               "svelte"
                               "plugin"))
("svelte")

ELISP> (ht-keys (lsp--ht-get (lsp-configuration-section "svelte.plugin")
                               "svelte"
                               "plugin"))
("typescript" "css" "html" "svelte")
```

According to https://github.com/Wilfred/ht.el/pull/39#issue-533898005, the [`ht-get*` in the latest upstream of `ht.el`](https://github.com/Wilfred/ht.el/blob/2850301d19176b8d3bb6cc8d95af6ab7e529bd56/ht.el#L109-L119) returns the value the hash table had prior to the manipulations in the inner loop.

PR #2476, opened 2020-12-28
Set correct indentation width for `ess-mode`

w9:
Currently, any mode derived from `ess-mode` (`ess-r-mode`, for example), falls back to using `standard-indent`, which is surprising since `ess-mode` has a variable for indentation offset -- `ess-ident-offset`.


yyoncho (approved):

yyoncho:

PR #2477, opened 2020-12-28
Add an `iedit' command to edit documentHighlights

nbfalcon:
Add a new command, `lsp-iedit-highlights`, which will invoke `iedit` on
the documentHighlights at point. It can be used as a primitive
`lsp-rename` replacement if the former is unsupported (e.g. `haskell-language-server`).

Implement it on top the new function, `lsp-iedit--on-ranges`, which
invokes `iedit` on a list of `&Range`s. This separation will allow
implementing a similar `iedit` commmand for `LinkedEditingRanges`.

@yyoncho would you consider this out of the scope of `lsp-mode`?

### DIFF
--- a/lsp-iedit.el
+++ b/lsp-iedit.el
@@ -25,11 +25,13 @@
 
 (require 'lsp-mode)
 
+(declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
+(declare-function iedit-start-buffering "iedit-lib" ())
+
 (defvar iedit-mode)
 (defvar iedit-auto-buffering)
 (defvar iedit-occurrences-overlays)
-(declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
-(declare-function iedit-start-buffering "iedit-lib" ())
+
 (defun lsp-iedit--on-ranges (ranges)
   "Start an `iedit' operation using RANGES.
 RANGES shall be a list of lsp-`&Range's. They can be acquired
--- a/lsp-iedit.el
+++ b/lsp-iedit.el
@@ -1,0 +1,55 @@
+;;; lsp-iedit.el --- `iedit' integration -*- lexical-binding: t -*-
+
+;;; Commentary:
+
+;; This module provides features that allow starting `iedit' on various
+;; different lsp-based, semantic units (like documentHighlights, and
+;; linkedEditingRanges in the future).
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(defvar iedit-mode)
+(defvar iedit-auto-buffering)
+(defvar iedit-occurrences-overlays)
+(declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
+(declare-function iedit-start-buffering "iedit-lib" ())
+(defun lsp-iedit--on-ranges (ranges)
+  "Start an `iedit' operation using RANGES.
+RANGES shall be a list of lsp-`&Range's. They can be acquired
+from various lsp protocol requests, e.g.
+`textDocument/documentHighlight', ...."
+  (require 'iedit)
+  (unless (seq-empty-p ranges)
+    (mapc (-lambda ((&RangeToPoint :start :end))
+            (push (iedit-make-occurrence-overlay start end)
+                  iedit-occurrences-overlays))
+          ranges)
+    ;; See `iedit-start'; TODO: upstream this
+    (setq iedit-mode t)
+    (when iedit-auto-buffering
+      (iedit-start-buffering))
+    (run-hooks 'iedit-mode-hook)
+    (add-hook 'before-revert-hook 'iedit-done nil t)
+    (add-hook 'kbd-macro-termination-hook 'iedit-done nil t)
+    (add-hook 'change-major-mode-hook 'iedit-done nil t)
+    (add-hook 'iedit-aborting-hook 'iedit-done nil t)
+    (message "%d occurrences of \"%s\""
+             (seq-length ranges)
+             (lsp--range-text (lsp-seq-first ranges)))))
+
+;;;###autoload
+(defun lsp-iedit-highlights ()
+  "Start an `iedit' operation on the documentHighlights at point.
+This can be used as a primitive `lsp-rename' replacement if the
+language server doesn't support renaming.
+
+See also `lsp-enable-symbol-highlighting'."
+  (interactive)
+  (let ((highlights (lsp-request "textDocument/documentHighlight"
+                                 (lsp--text-document-position-params))))
+    (lsp-iedit--on-ranges (mapcar #'lsp:document-highlight-range highlights))))
+
+(provide 'lsp-iedit)
+;;; lsp-iedit.el ends here
--- a/lsp-iedit.el
+++ b/lsp-iedit.el
@@ -1,4 +1,19 @@
 ;;; lsp-iedit.el --- `iedit' integration -*- lexical-binding: t -*-
+;;
+;; Copyright (C) 2020 emacs-lsp maintainers
+;;
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -8385,6 +8385,43 @@ This avoids overloading the server with many files when starting Emacs."
     (lsp--error "Nothing to disconnect from?")))
 
 
+;;; `lsp-iedit'
```
ericdallo, line 8388: Can we move this to a separate file? otherwise `lsp-mode.el`  will explode 😅

nbfalcon: Sure.

lsp-mode.el
```diff
@@ -8385,6 +8385,43 @@ This avoids overloading the server with many files when starting Emacs."
...
+             (lsp--range-text (lsp-seq-first ranges)))))
+
+(defun lsp-iedit-highlights ()
+  "Start an `iedit' operation on the documentHighlights at point.
+This can be used as a primitive `lsp-rename' replacement if the
+language server doesn't support renaming.
+
+See also `lsp-enable-symbol-highlighting'."
+  (interactive)
+  (let ((highlights (lsp-request "textDocument/documentHighlight"
```
ericdallo, line 8422: We should check if user has `iedit` and if is required via `(require 'iedit nil t)`
Also we should declare the `iedit` methods, check how we do for `avy` integration

lsp-mode.el
```diff
@@ -8385,6 +8385,43 @@ This avoids overloading the server with many files when starting Emacs."
...
+    ;; See `iedit-start'; TODO: upstream this
+    (setq iedit-mode t)
+    (when iedit-auto-buffering
+      (iedit-start-buffering))
+    (run-hooks 'iedit-mode-hook)
+    (add-hook 'before-revert-hook 'iedit-done nil t)
+    (add-hook 'kbd-macro-termination-hook 'iedit-done nil t)
+    (add-hook 'change-major-mode-hook 'iedit-done nil t)
+    (add-hook 'iedit-aborting-hook 'iedit-done nil t)
+    (message "%d occurrences of \"%s\""
```
ericdallo, line 8411: Should we use `lsp-info` instead of raw `message`?

lsp-mode.el
nbfalcon: I will do that.

ericdallo: This will probably fix the CI

lsp-mode.el
nbfalcon: I would argue against that, for the sake of consistency with `iedit` itself.

lsp-mode.el
nbfalcon: Fixed. Although `(require 'iedit nil t)` is not correct here, as that function would fail much less intuitively if `iedit` wasn't available.

lsp-iedit.el
```diff
@@ -0,0 +1,55 @@
+;;; lsp-iedit.el --- `iedit' integration -*- lexical-binding: t -*-
```
ericdallo, line 1: Could you follow other files template?
https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-headerline.el#L3-L16

nbfalcon:
`lsp-iedit`


vs `iedit-mode`


Note how in the first case, irrelevant textual occurrences of `coll` aren't edited.

lsp-iedit.el
```diff
@@ -0,0 +1,55 @@
...
+;; different lsp-based, semantic units (like documentHighlights, and
+;; linkedEditingRanges in the future).
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(defvar iedit-mode)
+(defvar iedit-auto-buffering)
+(defvar iedit-occurrences-overlays)
```
ericdallo, line 15:
```suggestion
(defvar iedit-occurrences-overlays)

```

lsp-iedit.el
```diff
@@ -0,0 +1,55 @@
...
+(declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
+(declare-function iedit-start-buffering "iedit-lib" ())
```
ericdallo, line 17:
```suggestion
(declare-function iedit-start-buffering "iedit-lib" ())

```

lsp-iedit.el
nbfalcon: Should this by `Copyright lsp-maintainers` or `Copyright <me>`?

lsp-iedit.el
```diff
@@ -0,0 +1,55 @@
...
+(defun lsp-iedit--on-ranges (ranges)
+  "Start an `iedit' operation using RANGES.
+RANGES shall be a list of lsp-`&Range's. They can be acquired
+from various lsp protocol requests, e.g.
+`textDocument/documentHighlight', ...."
+  (require 'iedit)
```
ericdallo, line 23: Maybe its better to change to a:
```elisp
(if (require 'iedit nil t)
      ....
   (lsp-warn "iedit package is not installed"))
```

lsp-iedit.el
nbfalcon: Fixed.

lsp-iedit.el
nbfalcon: Fixed.

lsp-iedit.el
ericdallo: lsp-maintainers

nbfalcon: Fixed, with `Copyright emacs-lsp maintainers`.

lsp-iedit.el
nbfalcon: This would have to be an `lsp-warn` followed by an `error` anyway, since we mustn't continue trying to use `iedit`-functions without it loaded. Besides, `lsp-avy-lens` doesn't do that either. And lastly, I'd argue that such a warning is redundant, given that users of that function would probably already know about `iedit` and have it installed.

ericdallo: Yeah, a lsp-error would fit better. 
I think a nice and clear UX for users is always better :)

nbfalcon: I absolutely agree that we should strive for having a good and intuitive UX. However, I don't believe using `lsp--error` is the right thing here, because it conflicts with doing things in a way that is consistent with the rest of Emacs. Few (perhaps none?) packages are putting `require nil t` + `(warning)` guards around `require`, since that is cumbersome. Calling this function without `iedit` installed should be a regular `error`, not just demoted to a pretty `lsp` message. And the error message from an invalid `require` isn't too bad: `(file-missing "Cannot open load file" "No such file or directory" "q")`.

ericdallo: I see, my point is user may not know what is `iedit` and wanna try that lsp feature and understand what it does.
And a `Package x is not installed` may be a better message than the error that may happen at any time.
But is just a suggestion, no hard feeling about that

lsp-iedit.el
```diff
@@ -0,0 +1,72 @@
...
+          ranges)
+    ;; See `iedit-start'; TODO: upstream this
+    (setq iedit-mode t)
+    (when iedit-auto-buffering
+      (iedit-start-buffering))
+    (run-hooks 'iedit-mode-hook)
+    (add-hook 'before-revert-hook 'iedit-done nil t)
+    (add-hook 'kbd-macro-termination-hook 'iedit-done nil t)
+    (add-hook 'change-major-mode-hook 'iedit-done nil t)
+    (add-hook 'iedit-aborting-hook 'iedit-done nil t)
```
yyoncho, line 54: can you elaborate on why this is needed? Seems like this should be part of the iedit package.

lsp-iedit.el
```diff
@@ -0,0 +1,72 @@
...
+    (message "%d occurrences of \"%s\""
+             (seq-length ranges)
+             (lsp--range-text (lsp-seq-first ranges)))))
```
yyoncho, line 57: How ranges with different length will work? E. g. in java you could have FooClass and foo.bar.FooClass selected.

yyoncho (approved):
Looks good.

lsp-iedit.el
nbfalcon: `iedit` works like `multiple-cursors`, so that will still work. Erasing the first char of `FooClass` will erase the first char of `foo.bar.FooClass`.

lsp-iedit.el
nbfalcon: Sadly, this isn't. The closest function to this functionality would be `iedit-start`, but it takes a regex, so we cannot use it here.

ericdallo:
@nbfalcon I think we fogot to add a entry in CHANGELOG and some documentation about that, would you mind creating a second PR with that?

nbfalcon:
Sure, I can do that.

PR #2479, opened 2020-12-29
`lsp-unzip-script': prefer unzip over powershell

nbfalcon:
The powershell unzip script does not work on linux, but the former is available
on e.g. Ubuntu as a snap. Prefer unzip if available, so that `lsp-unzip-script`
works out of the box even if the user has powershell installed.

If I try to install `eslint` using `M-x lsp-install-server`, I get the following error:
```
The argument 'Expand-Archive' is not recognized as the name of a script file. Check the spelling of the name, or if a path was included, verify that the path is correct and try again.
```
which leads me to believe that the powershell unzip script doesn't work on Linux.
I have `powershell` installed as a snap and `Expand-Archive` works if I run `powershell` in a regular terminal. `unzip` works just fine, and I believe it should be the default, also because it is less heavy.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6950,8 +6950,8 @@ STORE-PATH to make it executable."
 (defconst lsp-ext-unzip-script "bash -c 'mkdir -p %2$s && unzip -qq %1$s -d %2$s'"
   "Unzip script to unzip file.")
 
-(defcustom lsp-unzip-script (cond ((executable-find "powershell") lsp-ext-pwsh-script)
-                                  ((executable-find "unzip") lsp-ext-unzip-script)
+(defcustom lsp-unzip-script (cond ((executable-find "unzip") lsp-ext-unzip-script)
+                                  ((executable-find "powershell") lsp-ext-pwsh-script)
                                   (t nil))
   "The script to unzip."
   :group 'lsp-mode


yyoncho:
This order was switched in `dap-mode` as a result of an issue report. I think that it was a similar issue but for windows. Maybe it should be: prefer PowerShell unless you are on Linux and there is `unzip`(it can be written in multiple ways).

kiennq:
Did the binary for newest powershell appeared as `powershell` on linux? I though it should be `pwsh` instead.
That said, we can just append the `.exe` to `powershell.exe` so that the file is guaranteed on Windows.

nbfalcon:
In my case, I have installed powershell from [`snap`](https://snapcraft.io/) as `snap install powershell` (no other packages, symlinks or configuration was done) and I have both `pwsh` and `powershell` in my `PATH`.

```console
$ whereis powershell
/snap/bin/powershell
$ whereis pwsh
/snap/bin/pwsh
$ ll /snap/bin/pwsh
/snap/bin/pwsh -> powershell
```

nbfalcon:
@yyoncho what is the reason to prefer powershell over unzip on windows? Could there be an `unzip` binary in the PATH which doesn't work like unzip?

yyoncho:
> @yyoncho what is the reason to prefer powershell over unzip on windows? Could there be an `unzip` binary in the PATH which doesn't work like unzip?

The issue comes from gitter - https://gitter.im/emacs-lsp/lsp-mode?at=5f70d1b9e1dd7c19549c1b20 and it caused by the fact that you may have unzip but you may miss `bash` and it will fail.

kiennq:
So we can add more condition like @yyoncho suggested or add suffix `.exe` for `powershell` to make sure that's Windows binary.

nbfalcon:
I don't see why we'd need bash for that though - `shell-command` already spawns a subshell, so I propose that we just don't use bash in `lsp-ext-unzip-script`.

nbfalcon:
Also, shouldn't we `shell-quote-arguments` the format strings? Because paths with `'` in them will break.

nbfalcon:
It would be great if `format` had a `shell-quote-argument` specifier.

kiennq:
`fish` shell doesn't support `&&` as per https://github.com/emacs-lsp/dap-mode/issues/81
Thus `bash` is needed.

nbfalcon:
This seems fixed as of https://github.com/fish-shell/fish-shell/issues/4620 and `&&` works in `fish` `3.1.0`:
```console
$ fish -c "true && echo q"
q
$ fish --version
fish, version 3.1.0
```

Do we need to support older versions of fish? In that case, I think checking for `powershell.exe` might be better, because that isn't in the PATH of a standard Ubuntu snap install.

PR #2481, opened 2020-12-29
Run pcase-defmacro for interfaces

### LINKED ISSUE #2285: Make pcase match protocol-specified types
`lsp-pcase` would work like `pcase`, but we would pattern-match against protocol-specified types. This would be helpful for type-checking server results that could be one of several types (for example, `Location | Location [] | LocationLink []`).

Actually, we might instead use `pcase-defmacro` to simply define new cases for regular `pcase`. I haven't thought about it enough.

leungbk:
Closes #2285.

An alternate approach is to write a separate wrapper entirely (untested):

``` elisp
(defmacro lsp-dcase (expr-val &rest clauses)
  "Destructure EXPR-VAL upon reaching the first matching clause,
then execute the corresponding body form."
  (-when-let (((bindings body) . remaining-clauses) clauses)
    `(-if-let (,bindings ,expr-val)
         (progn ,@body)
       (lsp-dcase ,expr-val ,remaining-clauses))))
```

But using `pcase-defmacro` instead lets us use the existing `pcase` patterns, most notably the `seq` pattern.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4434,12 +4434,13 @@ and the position respectively."
   "Return a list of `xref-item' given LOCATIONS, which can be of
 type Location, LocationLink, Location[] or LocationLink[]."
   (setq locations
-        (if (and (sequencep locations)
-                 (let ((fst (lsp-seq-first locations)))
-                   (or (lsp-location? fst)
-                       (lsp-location-link? fst))))
-            (append locations nil)
-          (when locations (list locations))))
+        (pcase locations
+          ((seq (or (Location)
+                    (LocationLink)))
+           (append locations nil))
+          ((or (Location)
+               (LocationLink))
+           (list locations))))
 
   (cl-labels ((get-xrefs-in-file
                (file-locs)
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -4707,17 +4708,15 @@ When language is nil render as markup if `markdown-mode' is loaded."
...
-       (-let [(&MarkedString :language :value) content]
-         (lsp--render-string value language)))
-      ((lsp-markup-content? content)
-       (-let [(&MarkupContent :value :kind) content]
-         (lsp--render-string value kind)))
-      ;; plain string
-      ((stringp content) (lsp--render-string content "markdown"))
-      ((null content) "")
-      (t (error "Failed to handle %s" content)))
+     (pcase content
```
ericdallo, line 4711: shouldn't we use `lsp-pcase`?

ericdallo: oh got it, it's an existing macro

yyoncho (approved):
Looks very good! Consider adding a unit test for that.

leungbk:
OK, I added some tests.

kiennq (approved):

yyoncho:
> OK, I added some tests.

Thank you. FYI the reason I am requesting tests is that we already have a good test suite for that particular feature and it is relatively straightforward to add them.

PR #2482, opened 2020-12-30
Fix Clangd's Flycheck Explainer

danielmartin:
(Requires a recent version of Flycheck with this fix: https://github.com/flycheck/flycheck/pull/1801)

- The first fix is an update to the `declare-function` to match our recent folder structure.
- The second fix involves storing the LSP Diagnostic's `source` that Clangd sends as `clang-tidy` in `flycheck-error`'s `group` field. That gives the error explainer enough context to special treat clang-tidy checks in a reliable way.
- Flycheck explainer logic is now decoupled by using `cl-defgeneric`.
- Finally, we add one more new line to the function that loads the explanation asynchronously, to make Flycheck show it in a new buffer and not in the echo area.


lsp-diagnostics.el
```diff
@@ -75,12 +75,13 @@ on top the flycheck face for that error level."
...
 (declare-function flycheck-error-new "ext:flycheck" t t)
 (declare-function flycheck-error-message "ext:flycheck" (err) t)
+(declare-function flycheck-error-group "ext:flycheck" (err) t)
 (declare-function flycheck-define-error-level "ext:flycheck" (level &rest properties))
 (declare-function flycheck-buffer "ext:flycheck")
 (declare-function flycheck-valid-checker-p "ext:flycheck")
 (declare-function flycheck-stop "ext:flycheck")
 
-(declare-function lsp-cpp-flycheck-clang-tidy-error-explainer "lsp-cpp")
+(declare-function lsp-cpp-flycheck-clang-tidy-error-explainer "clients/lsp-clangd")
```
kiennq, line 84: There should not be `clients/` here, since when package is installed, all of clients directory is flattened

lsp-diagnostics.el
```diff
@@ -202,7 +204,7 @@ See https://github.com/emacs-lsp/lsp-mode."
       :modes '(lsp-placeholder-mode) ;; placeholder
       :predicate (lambda () lsp-mode)
       :error-explainer (lambda (e)
-                         (cond ((string-prefix-p "clang-tidy" (flycheck-error-message e))
+                         (cond ((string-equal "clang-tidy" (flycheck-error-group e))
                                 (lsp-cpp-flycheck-clang-tidy-error-explainer e))
```
yyoncho, line 208: FTR here we can use cl-defgeneric to get rid of clangd reference.

yyoncho (approved):

kiennq (approved):

PR #2483, opened 2020-12-30
lsp-completion: company integration, don't invalid cache on cancel action without pos-completion

### LINKED ISSUE #2475: (error "Marker does not point anywhere") on post-completion
Hi,

I've got an issue with Rust Analyzer's post-completion functions. I previously [reported this to Company mode](https://github.com/company-mode/company-mode/issues/1043), but was told to take it here.

# Describe the bug

When using [Rust Analyzer's](https://github.com/rust-analyzer/rust-analyzer) post-completion functionality, I came across an issue where it seemingly doesn't work unless the previous action removed text from the buffer or there is text following the items to be expanded.

More specifically, given this scenario (inside a function and where `|` is the cursor):
```rust
true.if|
```

I can trigger `company-complete` to get the list of completion candidates (which contains a single item). If I try to `company-complete-selection`, I get an error message saying

```
company-call-backend-raw: Company: backend company-capf error "Marker does not point anywhere" with args (post-completion if)
```

However, if I add a character and then delete it, say `true.if` -> `true.ifh` -> `true.if`, expanding the snippet works just fine, expanding to

```rust
if true {

}
```

The same happens if what I delete is elsewhere in the buffer too (such as on a different line) and it doesn't seem to matter what character it is (whitespace or otherwise).

Also, it turns out that as long as I enter any characters after `if` (e.g. `true.ifh`) and move the cursor back to before the extra characters, expansions works as expected and removes the remaining characters: `true.if|h` -> `company-complete-selection` becomes

```rust
if true {

}
```

If a string follows the `if`, then it remains intact for some reason. `true.if"string"` becomes:
```rust
if true {

}"string"
```

**To Reproduce**

Sample project: https://github.com/thomasheartman/lsp-rust-analyzer-post-completion-issue

The sample project is a fresh rust project (as created by `cargo new`). In `src/main.rs`, line 6 is `true.if`. Expanding this snippet may or may not work based on previous action. It seems to always work if I delete characters before expanding. It never works if I try and expand it just after typing it out.

**Expected behavior**

I expect the post-completion to work regardless of what my previous action in the buffer was, and especially if I have just typed out the necessary letters for expansion to take place. I expect it to not throw any errors.

Rust Analyzer (as of 2020-11-09).

**OS**
NixOS 20.09

**Error callstack**

<details>
<summary>Callstack</summary>

```
Debugger entered--Lisp error: (error "Marker does not point anywhere")
  delete-region(271 #<marker (moves after insertion) in no buffer>)
  apply(delete-region (271 #<marker (moves after insertion) in no buffer>))
  lsp-completion--exit-fn(#("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)) finished #f(compiled-function () #<bytecode 0x32995c9>))
  apply(lsp-completion--exit-fn (#("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)) finished #f(compiled-function () #<bytecode 0x32995c9>)))
  #f(compiled-function (&rest args-before) #<bytecode 0x24b93ed>)(#("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)) finished)
  company--capf-post-completion(#("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  #f(compiled-function (command &optional arg &rest args) "`company-mode' backend using `completion-at-point-functions'." (interactive #f(compiled-function () #<bytecode 0x34d527d>)) #<bytecode 0x32ddce1>)(post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  apply(#f(compiled-function (command &optional arg &rest args) "`company-mode' backend using `completion-at-point-functions'." (interactive #f(compiled-function () #<bytecode 0x34d5cc9>)) #<bytecode 0x32ddce1>) (post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>))))
  company-flx-company-capf-advice(#f(compiled-function (command &optional arg &rest args) "`company-mode' backend using `completion-at-point-functions'." (interactive #f(compiled-function () #<bytecode 0x2502e7d>)) #<bytecode 0x32ddce1>) post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  apply(company-flx-company-capf-advice #f(compiled-function (command &optional arg &rest args) "`company-mode' backend using `completion-at-point-functions'." (interactive #f(compiled-function () #<bytecode 0x25036a1>)) #<bytecode 0x32ddce1>) (post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>))))
  company-capf(post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  apply(company-capf (post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>))))
  company-call-backend-raw(post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  apply(company-call-backend-raw (post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>))))
  company--force-sync(company-call-backend-raw (post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>))) company-capf)
  company-call-backend(post-completion #("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  company-cancel(#("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  company-finish(#("if" 0 2 (face (completions-common-part) lsp-completion-prefix "if" lsp-completion-markers (271 #<marker (moves after insertion) in no buffer>) lsp-completion-start-point 271 lsp-sort-text nil lsp-completion-item #<hash-table equal 9/12 0x2c4bd95>)))
  company-complete-selection()
  my-company-complete-selection-or-first()
  funcall-interactively(my-company-complete-selection-or-first)
  call-interactively(my-company-complete-selection-or-first nil nil)
  command-execute(my-company-complete-selection-or-first)
```

</details>

<details>
<summary>`*lsp-log*`</summary>

```
Command "rls" is not present on the path.
Command "/home/thomas/.emacs.d/.cache/lsp/rust/rust-analyzer" is present on the path.
Command "rls" is not present on the path.
Command "/home/thomas/.emacs.d/.cache/lsp/rust/rust-analyzer" is present on the path.
Found the following clients for /home/thomas/projects/rust-analyzer-setup/src/main.rs: (server-id rust-analyzer, priority 1)
The following clients were selected based on priority: (server-id rust-analyzer, priority 1)
Command "rls" is present on the path.
Command "/nix/store/99y2l9alga74r3fzd55a8y7nkzw0qzd8-rust-analyzer-unstable-2020-08-24/bin/rust-analyzer" is present on the path.
Command "rls" is present on the path.
Command "/nix/store/99y2l9alga74r3fzd55a8y7nkzw0qzd8-rust-analyzer-unstable-2020-08-24/bin/rust-analyzer" is present on the path.
Found the following clients for /home/thomas/projects/rust-analyzer-setup/src/main.rs: (server-id rls, priority -1), (server-id rust-analyzer, priority 1)
The following clients were selected based on priority: (server-id rust-analyzer, priority 1)
Creating watch for /home/thomas/projects/rust-analyzer-setup
Creating watch for /home/thomas/projects/rust-analyzer-setup/src
Command "rls" is present on the path.
Command "/nix/store/99y2l9alga74r3fzd55a8y7nkzw0qzd8-rust-analyzer-unstable-2020-08-24/bin/rust-analyzer" is present on the path.
Command "rls" is present on the path.
Command "/nix/store/99y2l9alga74r3fzd55a8y7nkzw0qzd8-rust-analyzer-unstable-2020-08-24/bin/rust-analyzer" is present on the path.
Found the following clients for /home/thomas/projects/lsp-post-completion/src/main.rs: (server-id rls, priority -1), (server-id rust-analyzer, priority 1)
The following clients were selected based on priority: (server-id rust-analyzer, priority 1)
Creating watch for /home/thomas/projects/lsp-post-completion
Creating watch for /home/thomas/projects/lsp-post-completion/src
Cancelling textDocument/onTypeFormatting(97) in hook after-change-functions
Cancelling textDocument/codeAction(145) in hook after-change-functions
Cancelling textDocument/hover(159) in hook after-change-functions
Cancelling textDocument/codeAction(158) in hook after-change-functions
Cancelling textDocument/hover(176) in hook after-change-functions
Cancelling textDocument/hover(206) in hook after-change-functions
Cancelling textDocument/hover(211) in hook after-change-functions
Cancelling textDocument/codeAction(210) in hook after-change-functions
Cancelling textDocument/hover(218) in hook after-change-functions
Cancelling textDocument/codeAction(217) in hook after-change-functions
Cancelling textDocument/hover(226) in hook after-change-functions
Cancelling textDocument/documentHighlight(225) in hook after-change-functions
Cancelling textDocument/codeAction(224) in hook after-change-functions
Cancelling textDocument/codeAction(327) in hook after-change-functions
```

</details>

The `lsp-workspace-show-log` output was too long to include in this comment. I'll try and put it in a separate comment.

# Versions

- Emacs: GNU Emacs 27.1 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.21, cairo version 1.16.0)
- lsp-mode: lsp-mode 20201113.1903, Emacs 27.1, gnu/linux
- rust-analyzer: rust-analyzer 2020-08-24 and 2020-11-09 
- company: Company version: 0.9.13

## Reproducing the issue in a clean environment:

I tried reproducing the issue in a clean environment but wasn't able to install the correct major mode. I'm not sure why or what's going on. If there is a simple solution, I'd be more than happy to try again. I've included the error message below.

<details>
<summary>Install error message</summary>

```
  string-match("\\`https?:" nil nil)
  package--with-response-buffer-1(nil #f(compiled-function () #<bytecode 0x13b318d>) :file "rust-mode-20201204.1527.el" :async nil :error-function #f(compiled-function () #<bytecode 0xc62f05>) :noerror nil)
  package-install-from-archive(#s(package-desc :name rust-mode :version (20201204 1527) :summary "A major emacs mode for editing Rust source code" :reqs ((emacs (25 1))) :kind single :archive "melpa" :dir nil :extras ((:commit . "c5c7ed31a2e1106ab4835b135618a34570796dc7") (:authors ("Mozilla")) (:maintainer "Mozilla") (:keywords "languages") (:url . "https://github.com/rust-lang/rust-mode")) :signed nil))
  mapc(package-install-from-archive (#s(package-desc :name rust-mode :version (20201204 1527) :summary "A major emacs mode for editing Rust source code" :reqs ((emacs (25 1))) :kind single :archive "melpa" :dir nil :extras ((:commit . "c5c7ed31a2e1106ab4835b135618a34570796dc7") (:authors ("Mozilla")) (:maintainer "Mozilla") (:keywords "languages") (:url . "https://github.com/rust-lang/rust-mode")) :signed nil)))
  package-download-transaction((#s(package-desc :name rust-mode :version (20201204 1527) :summary "A major emacs mode for editing Rust source code" :reqs ((emacs (25 1))) :kind single :archive "melpa" :dir nil :extras ((:commit . "c5c7ed31a2e1106ab4835b135618a34570796dc7") (:authors ("Mozilla")) (:maintainer "Mozilla") (:keywords "languages") (:url . "https://github.com/rust-lang/rust-mode")) :signed nil)))
  package-install(rust-mode nil)
  funcall-interactively(package-install rust-mode nil)
  call-interactively(package-install record nil)
  command-execute(package-install record)
  execute-extended-command(nil "package-install" "package-inst")
  funcall-interactively(execute-extended-command nil "package-install" "package-inst")

```

</details>

kiennq:
Fix #2475 
Also always showing the candidates popup even when there's only one candidate, so that the pos completion logic can be properly triggered.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -399,9 +399,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
             (lambda ()
               (lsp--catch 'input
                   (let ((lsp--throw-on-input lsp-completion-use-last-result)
-                        (same-session? (and (not done?)
-                                            (not result)
-                                            lsp-completion--cache
+                        (same-session? (and lsp-completion--cache
                                             (equal (cl-first lsp-completion--cache) bounds-start)
                                             (s-prefix?
                                              (plist-get (cddr lsp-completion--cache) :prefix)
@@ -444,7 +442,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
                                                           (cond
                                                            ((and done? (not (seq-empty-p items)))
                                                             (lsp-completion--to-internal items))
-                                                            ((not done?) :incomplete))
+                                                           ((not done?) :incomplete))
                                                           :lsp-items nil
                                                           :markers markers
                                                           :prefix prefix)
@@ -675,8 +673,9 @@ The CLEANUP-FN will be called to cleanup."
   :lighter ""
   (let ((completion-started-fn (lambda (&rest _)
                                  (setq-local lsp-inhibit-lsp-hooks t)))
-        (after-completion-fn (lambda (&rest _)
-                               (lsp-completion--clear-cache)
+        (after-completion-fn (lambda (result)
+                               (when (stringp result)
+                                 (lsp-completion--clear-cache))
                                (setq-local lsp-inhibit-lsp-hooks nil))))
     (cond
      (lsp-completion-mode


nbfalcon:
I can confirm that this also fixes the auto-completion `error` I had with `iph`. Thank you!

lsp-completion.el
```diff
@@ -397,11 +398,11 @@ The MARKERS and PREFIX value will be attached to each candidate."
            result done?
            (candidates
             (lambda ()
+              ;; Force manual completion action so post completion logic can be run
+              (setq-local company--manual-action t)
               (lsp--catch 'input
                   (let ((lsp--throw-on-input lsp-completion-use-last-result)
-                        (same-session? (and (not done?)
```
nbfalcon, line 402: Why did you drop `(not done?)` and `(not result)`? Was it because they are always nil here?

lsp-completion.el
```diff
@@ -397,11 +398,11 @@ The MARKERS and PREFIX value will be attached to each candidate."
            result done?
            (candidates
             (lambda ()
+              ;; Force manual completion action so post completion logic can be run
+              (setq-local company--manual-action t)
```
nbfalcon, line 402: Why is this needed? Post-completion hooks should work without `company--manual-action` (as they did previously), or am I missing something? Removing this line still fixes my `iph` error.

lsp-completion.el
kiennq: The check will not make much different since with the first `cond` clause is checking for `(or done? result)` already.
I'm planning to remove that cached `result` (company-capf is caching it) too if there's error.

lsp-completion.el
kiennq: In case of there's only one completion result, `company-mode` will do some check of completion result and prefix, if they're the same, the completion popup will not be shown, and `company-cancel` will be triggered with `unique` as result.
In that case, the post-completion will not be run and user will need to manually trigger `company-complete` for the completion to show up, which is kind of annoying.
If we force set `company--manual-action`, the completion popup will always be shown, make the post-completion logic will always be run.

kiennq: The `true.if` in this bug is actually falling on that case too

nbfalcon: I see. Can't we just run the post-completion hook in `company-after-completion-fn` though, if `RESULT` = `unique`, so that we don't end up relying on implementation details?

lsp-completion.el
```diff
@@ -675,8 +676,9 @@ The CLEANUP-FN will be called to cleanup."
   :lighter ""
   (let ((completion-started-fn (lambda (&rest _)
                                  (setq-local lsp-inhibit-lsp-hooks t)))
-        (after-completion-fn (lambda (&rest _)
-                               (lsp-completion--clear-cache)
+        (after-completion-fn (lambda (result)
+                               (when (stringp result)
```
nbfalcon, line 680: This would mean that, when aborting, the cache wouldn't be cleared. That might be better though, because if the user auto-completes again at the same point, the result will be cached.

lsp-completion.el
nbfalcon: Aside from this implementation detail, I approve of this PR.

nbfalcon: From looking at the company source, it seems that if there is a sole match, some company functions should not just complete it, which would be the case for all other company backends, too? Are you sure that `lsp-mode` is the right place to fix that behaviour?

yyoncho: If we have to do that to make things work then most likely there is a bug/limitation in company-mode that has to be fixed. Seems like we should include company's maintainer in the loop.

lsp-completion.el
kiennq: Yes, beside we have a cache clear again at completion start so the cache should not be leaked to become too large anyway.

lsp-completion.el
kiennq: Let's loop @dgutov here.
I think that's intentional behavior from `company-mode`. When there's sole match, the completion popup is showing up may seem a little bit silly and intrusive. Thus for automatically invoked completion, not showing the popup (let's call it candidate window) is actually makes sense.
However, in case of manual invoked completion (`company--manual-action` is `t`), the candidate window will always be shown as it's user intention.

For `lsp-mode`, since we actually relied on post-completion, always show the candidate window is better. Thus this hack.

yyoncho: IMHO showing/not showing the completion popup should not be related to invoking post-completion. It should happen no matter where the insertion comes from(as long as the candidate is inserted).

kiennq: I think for `company-mode` it's actually related.
Here is the commit (and there's linked discussion) about that behavior https://github.com/company-mode/company-mode/commit/201577cd7ca815b8be2dc62e1f433ee12c1d654f.
`company-mode` auto-completion cannot actually decided if user want to insert the pop-up or just want to type `RET` with current popup. Note that for non `company-tng-mode`, the popup is always preselected.
For `lsp-mode`, user actually relies on completion more so it makes more sense to always showing the popup and let user runs the post-completion

dgutov: This logic is basically for the case when the popup has been automatically shown, but then the user, say, types out the completion in full manually, and then continues typing. It would be weird to do template expansion, etc, in that case.

dgutov:
Hi!

Could you explain what's the problem with having a cache invalidated? Just populate it the next time it's needed.

kiennq:
> Hi!
> 
> Could you explain what's the problem with having a cache invalidated? Just populate it the next time it's needed.

Indeed, this is `lsp-mode` problem so that we have to populate the cache next time if needed.

However, while fixing that I discover the sole completion case that the `exit-fn` doesn't get called.
`lsp-mode` actually relies on that to work properly, so instead of just dismiss the completion popup, it's better to keep showing it as user explicitly invoke completion.

lsp-completion.el
kiennq: @yyoncho @nbfalcon With @dgutov 's answer, WDYT?
I feel that we should always show the completion popup and let user to run the `exit-fn` properly since `lsp-mode` relies on that to work properly.
Note that when user keeps typing, unless user presses `RET` or `SPC`, the sole match doesn't do anything weird.
Also, with the `company-tng-mode`, user can press `RET` and `SPC` normally since there's no selection by default.

dgutov:
> Indeed, this is lsp-mode problem so that we have to populate the cache next time if needed.

Why is that a problem, though? Does it take a long time?

> However, while fixing that I discover the sole completion case that the exit-fn doesn't get called.

You can still use `company-after-completion-hook` for cleanup, right?

> it's better to keep showing it as user explicitly invoke completion

I'm not sure I like that UI.

But more importantly, by touching a private variable and creating a behavior that is only present together with `lsp-mode` but not "stock" `company`, you increase the risk of having lsp-specific bugs in that area. If the proposed change in the behavior is objectively better for all users, we can discuss changing it. But if the main motivation is to solve a technical problem, there should be other ways of doing that.

yyoncho:
> Indeed, this is lsp-mode problem so that we have to populate the cache next time if needed.

Don't we match the cache with start/end points? Also, seems like hiding the popup when the user has typed something that does not exist must result in cleaning the cache and that seems to be equivalent to typing the full name and then continuing to type.  

I agree with @dgutov that the behaviour should be controlled by public variables(I think that we all agree on that).

kiennq:
> Don't we match the cache with start/end points? Also, seems like hiding the popup when the user has typed something that does not exist must result in cleaning the cache and that seems to be equivalent to typing the full name and then continuing to type.

@yyoncho, yes we match that. However in case of sole completion, we do cache cleaning too early and have to avoid that.

> If the proposed change in the behavior is objectively better for all users, we can discuss changing it

The current experience that when user type `true.if`, the popup disappeared since it's sole match. Leave user with no other choice but to invoke `company-complete` again to be able to expand it via `exit-fn`.
I think that would be better (at least for `lsp-mode`) to continue showing up completion popup here instead.

Of course having public variable for that is better.

yyoncho:
@dgutov 

> This logic is basically for the case when the popup has been automatically shown, but then the user, say, types out the completion in full manually, and then continues typing. It would be weird to do template expansion, etc, in that case.

But you are also removing the opportunity for the user to press `RET` and insert the snippet. This is observable with company-yasnippet - if you narrow to a single candidate match, the popup disappears and now you have to invoke the completion manually which IMHO is inconvenient UE.

dgutov:
> But you are also removing the opportunity for the user to press RET and insert the snippet. This is observable with company-yasnippet - if you narrow to a single candidate match, the popup disappears and now you have to invoke the completion manually which IMHO is inconvenient UE.

That is true. This behavior prioritizes completion being less in-your-face, meaning, if you just continue typing, you run less of a risk of having `company-active-map` (e.g. `RET`) bindings conflicting which what you meant to do.

Perhaps it's not what everybody expects, though. Please feel free to open an issue about adding a user option for changing this behavior. Some example of how VS Code works should help too.

I do believe that `lsp-mode` should support both behaviors, though, so if you can fix this another way, that would be preferable.

dgutov:
BTW, what happens to your cache if the user just types `C-g` to abort completion? That doesn't run `post-completion` either.

kiennq:
> BTW, what happens to your cache if the user just types `C-g` to abort completion? That doesn't run `post-completion` either.

With new logic in this, the cache is left intact.

dgutov:
Then the fix doesn't actually require `post-completion` being run, and the change in visible behavior?

kiennq:
> Then the fix doesn't actually require `post-completion` being run, and the change in visible behavior?

Strictly speaking, yes. The fix is not requiring that.
That's why I put it as also in the first description.

When I first trying to repro this bug, actually I cannot do that since my `company-backends` contains `(company-capf :with company-dabbrev-code)`, thus almost make the sole completion case never happen.
Then I realized that with the current UX of hiding popup with full typing and have to trigger `company-complete` again, it's kind of easily lead user strays from golden path of expecting user to commit a completion candidate and clear the cache.
With this path, the cache is being clear earlier and left more memory back to Emacs.
Thus, the change in UX is also being pushed together with this fix

dgutov:
If memory is an issue, you could clear the cache in `company-after-completion-hook` instead, right?

yyoncho:
> If memory is an issue, you could clear the cache in `company-after-completion-hook` instead, right?

we had issues with that in the past, it is not called in certain cases which I don't recall ATM.

dgutov:
> we had issues with that in the past, it is not called in certain cases which I don't recall ATM

We might need to fix those, then.

But it certainly covers all cases when `post-completion` is called, and more. See the definition of `company-cancel`.

kiennq:
> If memory is an issue, you could clear the cache in `company-after-completion-hook` instead, right?

We're currently doing that. However it also include the case of sole completion (treated as abort) which we should not clear the cache, since user will just trigger recompletion again with `company-complete`.
Note that `company-capf` is also caching the data and will call into cached cleared (i.e,: this bug) closure in sole completion case.

yyoncho:
@kiennq IIUC company--manual-action is not part of the fix right, just a UE improvement, right?

If yes, can you remove that, so we can get the fix in, and then we will move the discussion in company repo for improving the display of a single match.

dgutov:
> since user will just trigger recompletion again with company-complete

That's debatable. But if true, yes, it might call for a change in behavior.

kiennq:
> @kiennq IIUC company--manual-action is not part of the fix right, just a UE improvement, right?
> 
> If yes, can you remove that, so we can get the fix in, and then we will move the discussion in company repo for improving the display of a single match.

I've removed it.

kiennq:
> > since user will just trigger recompletion again with company-complete
> 
> That's debatable. But if true, yes, it might call for a change in behavior.

I think this is case by case.
For normal use with spell checking the current behavior is perfectly fine and actually preferable.
But for programing with highly accurate and focused completion candidates has been provided, the completion is actually more preferable.
Having something as public interface to configure that is better for `company-mode` using package.

dgutov:
All right, so if you can file an issue, that would be great. And if you have an example of some other editor's behavior, even better.

yyoncho (approved):

PR #2484, opened 2020-12-30
Insert space to the lighter string for modeline

melito00:


kiennq:
Do we need lighter for `lens-mode` in the first place?

jcs090218 (approved):

jcs090218:
I will approve this first since this is a valid changes. 

> Do we need lighter for lens-mode in the first place?

We should just remove it?

nbfalcon:
I'd say that lighter is needed here because `lsp-lens-mode` provides a relatively important feature, so should show up by default, to be consistent with the rest of Emacs. 

Users of custom modelines like `doom-modeline`, `spaceline` by default won't see lighters anyway, and users of the standard modeline can use something like `use-package :diminish`.

From looking at `yasnippet`'s code, I see that they also have a space before the lighter.

nbfalcon:
LGTM.

nbfalcon:

PR #2488, opened 2020-12-30
Add support for `phpactor'

nbfalcon:
Add it as an lsp client to lsp-php.

Expose its extension installation feature trough an `interactive`
command, `lsp-phpactor-install-extension`.

Add a CHANGELOG entry and add `phpactor` to lsp-clients.json.

### DIFF
--- a/clients/lsp-php.el
+++ b/clients/lsp-php.el
@@ -352,7 +352,49 @@ already present."
                       (lsp--set-configuration
                        (lsp-configuration-section "serenata"))))
   :server-id 'serenata))
+
+;;; phpactor
+(defcustom lsp-phpactor-path "~/.composer/vendor/phpactor/phpactor/bin/phpactor"
+  "Path to the `phpactor' command."
+  :group 'lsp-phpactor
+  :type "string")
 
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection (list lsp-phpactor-path "language-server"))
+  :major-modes '(php-mode)
+  ;; `phpactor' is not really that feature-complete: it doesn't support
+  ;; `textDocument/showOccurence' and sometimes errors (e.g. find references on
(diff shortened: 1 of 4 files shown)


clients/lsp-php.el
```diff
@@ -352,7 +352,49 @@ already present."
...
+                                          ("PHPUnit" . "phpactor/phpunit-extension"))
+  "Alist mapping extension names to `composer' packages.
+These extensions can be installed using
+`lsp-phpactor-install-extension'."
+  :type '(alist :key-type "string" :value-type "string")
+  :group 'lsp-phpactor)
+
+(defun lsp-phpactor-install-extension (extension)
+  "Install a `phpactor' EXTENSION.
+See `lsp-phpactor-extension-alist' and "
```
yyoncho, line 387: the sentence is not finished.

yyoncho (approved):
Looks good!

clients/lsp-php.el
nbfalcon: Fixed.

clients/lsp-php.el
```diff
@@ -352,7 +352,56 @@ already present."
...
+  :group 'lsp-mode)
+
+(defcustom lsp-phpactor-path "~/.composer/vendor/phpactor/phpactor/bin/phpactor"
+  "Path to the `phpactor' command."
+  :group 'lsp-phpactor
+  :type "string")
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection (list lsp-phpactor-path "language-server"))
```
kiennq, line 370: you should change parameter of `lsp-stdio-connection` to lambda instead of static list so user can customize `lsp-phphactor-path` after `lsp-php` has been loaded

nbfalcon: Fixed.

PR #2490, opened 2020-12-31
Do not require all-the-icons and lsp-treemacs for the user

wyuenho:
If the user has installed but not required all-the-icons or lsp-treemacs, there must be a reason, so I don't think LSP mode should require them eagerly for the user. This PR fixes this issue.

### DIFF
--- a/lsp-modeline.el
+++ b/lsp-modeline.el
@@ -82,7 +82,7 @@
 
 (defun lsp-modeline--code-actions-icon (face)
   "Build the icon for modeline code actions using FACE."
-  (if (require 'all-the-icons nil t)
+  (if (functionp 'all-the-icons-octicon)
       (all-the-icons-octicon "light-bulb"
                              :face face
                              :v-adjust -0.0575)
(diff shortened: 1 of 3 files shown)


lsp-modeline.el
```diff
@@ -243,7 +243,7 @@ The `:global' workspace is global one.")
     (-> (s-join "/" strs)
         (propertize 'mouse-face 'mode-line-highlight
                     'help-echo "mouse-1: Show diagnostics"
-                    'local-map (when (require 'lsp-treemacs nil t)
+                    'local-map (when (functionp 'lsp-treemacs-errors-list)
```
yyoncho, line 246: this is autoloaded so it won't have any effect.

yyoncho (commented):
Let's see what others would say. IMHO we should extract these as functions (e. g. `lsp-headerline-get-icon-for-file-function`) to allow users to customize them and have the behavour wanted in this PR. The default should work OOTB, at least this is what we are trying to achieve with the other functionality. @ericdallo has PR for that, but for a different issue (disabling icons).

wyuenho:
Ah ok, I didn't realize there was another PR over there. I'll close this one.

PR #2493, opened 2021-01-03
lsp-completion: disable company-abort-on-unique-match

kiennq:
Per #2483 and https://github.com/company-mode/company-mode/issues/1046.
It should be better UX for always show the completion window even when there's unique match

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -122,6 +122,7 @@ This will help minimize popup flickering issue in `company-mode'."
 
 (defvar yas-indent-line)
 (defvar company-backends)
+(defvar company-abort-on-unique-match)
 
 (defvar lsp-completion--no-reordering nil
   "Dont do client-side reordering completion items when set.")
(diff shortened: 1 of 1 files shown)


jcs090218:
No problem to the changes. But variable `company-abort-on-unique-match` is a custom variable. Should it just respect to the user's decision?

yyoncho:
To address @jcs090218 concern I suggest having a mirror variable on our side with a different default value.

kiennq:
I think we can just move the change under provider `:capf` section.
We've already force put `company-capf` in the backends too. WDYT?

yyoncho:
For me it is not a big deal, most likely nobody will notice changing this property and most likely they will prefer our default. I am fine with any approach as long as we have it set to nil in lsp-mode buffers.

kiennq:
@jcs090218 WDYT?

jcs090218 (approved):

jcs090218:
@kiennq I respect to the team and community. ;) Feel free to merge this if you are ready.

PR #2494, opened 2021-01-03
Follow upstream change of iedit and add evil multiedit support

geza-herman:
Currently `lsp-iedit-highlights` is broken because of an upstream change, this PR fixes it. The problem was that occurrences were correctly highlighted, but edits were applied only to the current occurrence, not all of them.

And I created a new `lsp-evil-multiedit-highlights` function, which has a similar functionality as `lsp-iedit-highlights`, but it works with `evil-multiedit` mode.

### DIFF
--- a/lsp-iedit.el
+++ b/lsp-iedit.el
@@ -28,12 +28,14 @@
 (declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
 (declare-function iedit-start-buffering "iedit-lib" ())
 (declare-function iedit-lib-start "iedit-lib" ())
+(declare-function evil-multiedit-state "evil-multiedit" ())
 
 (defvar iedit-mode)
 (defvar iedit-auto-buffering)
 (defvar iedit-occurrences-overlays)
 (defvar iedit-occurrence-keymap)
 (defvar iedit-mode-occurrence-keymap)
+(defvar evil-multiedit--dont-recall)
 
 (defun lsp-iedit--on-ranges (ranges)
   "Start an `iedit' operation using RANGES.
@@ -72,6 +74,20 @@ See also `lsp-enable-symbol-highlighting'."
   (let ((highlights (lsp-request "textDocument/documentHighlight"
                                  (lsp--text-document-position-params))))
     (lsp-iedit--on-ranges (mapcar #'lsp:document-highlight-range highlights))))
+
+;;;###autoload
+(defun lsp-evil-multiedit-highlights ()
+  "Start an `evil-multiedit' operation on the documentHighlights at point.
+This can be used as a primitive `lsp-rename' replacement if the
+language server doesn't support renaming.
+
+See also `lsp-enable-symbol-highlighting'."
+  (interactive)
+  (require 'evil-multiedit)
+  (if (fboundp 'ahs-clear) (ahs-clear))
+  (setq evil-multiedit--dont-recall t)
+  (lsp-iedit-highlights)
+  (evil-multiedit-state))
 
 (provide 'lsp-iedit)
 ;;; lsp-iedit.el ends here
--- a/lsp-iedit.el
+++ b/lsp-iedit.el
@@ -27,7 +27,8 @@
 
 (declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
 (declare-function iedit-start-buffering "iedit-lib" ())
-(declare-function iedit-lib-start "iedit-lib" ())
+(declare-function iedit-lib-start "iedit-lib" (mode-exit-func))
+(declare-function iedit-done "iedit" ())
 (declare-function evil-multiedit-state "evil-multiedit" ())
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Can you adjust the PR to make the CI happy?

geza-herman:
> Can you adjust the PR to make the CI happy?

Done. Note, I'm not sure whether it's the correct way to add `evil-multiedit` support, because now there are `evil-multiedit` related declarations in `lsp-iedit.el`, so people who doesn't use `evil-multiedit` will see some unnecessary functions.

lsp-iedit.el
```diff
@@ -73,5 +75,19 @@ See also `lsp-enable-symbol-highlighting'."
...
+(defun lsp-evil-multiedit-highlights ()
+  "Start an `evil-multiedit' operation on the documentHighlights at point.
+This can be used as a primitive `lsp-rename' replacement if the
+language server doesn't support renaming.
+
+See also `lsp-enable-symbol-highlighting'."
+  (interactive)
+  (require 'evil-multiedit)
+  (if (fboundp 'ahs-clear) (ahs-clear))
+  (setq evil-multiedit--dont-recall t)
```
nbfalcon, line 88: What does setting this variable achieve, and can it be done without depending on the internals of `evil-multiedit`? Perhaps something needs to be addressed upstream?

nbfalcon: Aside from that, LGTM. I have made an `iedit` PR to improve `iedit-lib-start`, so most of the set-up will be gone soon, hopefully.

geza-herman: It is connected to `evil-multiedit-restore`. If this variable is `t`, then the selected symbols are not saved for recalling. I didn't choose to use this setting by my own, I followed how `evil-multiedit-match-all` sets things up. We need to use `t`, because restore doesn't work correctly if `lsp-evil-multiedit-highlights` did the highlighting. The reason for this is that it doesn't actually saves the selected symbols, but just the area and the regexp. So this feature doesn't work correctly even in non-lsp situations: for example, if the user selected symbols one-by-one, skipping some, then the skipped ones will be incorrectly re-selected at restore.

lsp-iedit.el
```diff
@@ -68,5 +75,19 @@ See also `lsp-enable-symbol-highlighting'."
...
+;;;###autoload
+(defun lsp-evil-multiedit-highlights ()
+  "Start an `evil-multiedit' operation on the documentHighlights at point.
+This can be used as a primitive `lsp-rename' replacement if the
+language server doesn't support renaming.
+
+See also `lsp-enable-symbol-highlighting'."
+  (interactive)
+  (require 'evil-multiedit)
+  (if (fboundp 'ahs-clear) (ahs-clear))
```
nbfalcon, line 87:
```suggestion
(when (fboundp 'ahs-clear) (ahs-clear))
```
If there is no `else` branch, use `when`.

lsp-iedit.el
```diff
@@ -68,5 +75,19 @@ See also `lsp-enable-symbol-highlighting'."
                                  (lsp--text-document-position-params))))
     (lsp-iedit--on-ranges (mapcar #'lsp:document-highlight-range highlights))))
 
+;;;###autoload
+(defun lsp-evil-multiedit-highlights ()
```
ericdallo, line 79: Since we are adding non `iedit` functions/features to this file, I think we should rename it or move this code to another file.
IMO rename to something common between the 2 would be better though

lsp-iedit.el
nbfalcon: I see. Perhaps a new command, `evil-multiedit-from-custom-iedit` or something like that should be added upstream. However, this needn't be addressed now.

lsp-iedit.el
geza-herman: `evil-multiedit` uses `iedit` (it's just a "thin" layer above `iedit`), so in a way it is an `iedit` feature. But yep, we can consider what you're suggesting. My first version of this function had the name `lsp-iedit-evil-multiedit-highlights`, but I didn't like this name too much. But I can rename it back :)

lsp-iedit.el
kiennq: Do you need to set `evil-multiedit--dont-recall` or just let-bind is enough?

lsp-iedit.el
ericdallo: Got it, I think we can keep it as it is so :)

lsp-iedit.el
geza-herman: It is need to be set because this variable is checked at a later time, in `evil-multiedit-abort` (when the mode is aborted, it is checked whether the current selections need to be saved so they can be recalled later).

yyoncho (approved):
I tested it and it and it works fine. I will merge it if there are no more comments on it.

yyoncho:
Thank you!

PR #2500, opened 2021-01-05
Change headerline separator until we find a way to detect missing glyph

yyoncho:
This char is not present in a lot of the popular fonts and it is causing trouble.

----

#


ericdallo (approved):
Good suggestion, IMO that's the best approach for now

jcs090218 (approved):

PR #2501, opened 2021-01-05
Add csharp-tree-sitter-mode to known C# modes

theothornhill:
I `csharp-mode` we now have a major mode with preliminary tree-sitter support. It would be nice to be able to use it on equal footing with the "legacy" `csharp-mode` to ease debugging going forward. This commit aims to add that. Are there any more places that need to be changed to get this right?


jcs090218 (approved):

jcs090218:
LGTM!

PR #2502, opened 2021-01-06
Try restoring most of flycheck state when toggling diagnostics mode

thomasf:


lsp-diagnostics.el
```diff
@@ -202,10 +209,13 @@ provided by lsp-mode.
 See https://github.com/emacs-lsp/lsp-mode."
       :start #'lsp-diagnostics--flycheck-start
       :modes '(lsp-placeholder-mode) ;; placeholder
-      :predicate (lambda () lsp-mode)
+      :predicate (lambda () lsp-diagnostics-mode)
```
thomasf, line 212: this is a bug fix because otherwise the lsp flycheck checker will ill be activated even if it's in lsp-diagnostics-disabled-modes

thomasf:
Hmm the CI job failed, I'm not sure why..

```
In toplevel form:
lsp-diagnostics.el:212:29:Error: reference to free variable ‘lsp-diagnostics-mode’
Makefile:40: recipe for target 'unix-compile' failed
make: *** [unix-compile] Error 1
```

define-minor-mode should declare that variable, right?

lsp-diagnostics.el
```diff
@@ -214,6 +224,16 @@ See https://github.com/emacs-lsp/lsp-mode."
   (add-hook 'lsp-diagnostics-updated-hook #'lsp-diagnostics--flycheck-report nil t)
   (add-hook 'lsp-managed-mode-hook #'lsp-diagnostics--flycheck-report nil t))
 
+(defun lsp-diagnostics-flycheck-disable (&rest _)
```
nbfalcon, line 227: Is there a reason for it to take `&rest` arguments?

thomasf: no, just becayse -enable had it

thomasf: i removed it

thomasf:
The reason that I needed this patch btw is because it was too fiddly implementing this function for without it.

I don't know if it's anything you are interested to have included in lsp-diagnostics or not. I find it useful.

```el
    (defun lsp-diagnostics-disabled-modes-toggle ()
      "Toggles the current major mode in the lsp-diagnostics-disabled-modes list and enables or disables lsp-diagnostics-mode in all matching open buffers."
      (interactive)
      (let ((disabled (member major-mode lsp-diagnostics-disabled-modes)))
        (if disabled
            (setq lsp-diagnostics-disabled-modes (remove major-mode lsp-diagnostics-disabled-modes))
          (add-to-list 'lsp-diagnostics-disabled-modes major-mode))
        (dolist (buf (buffer-list (current-buffer)))
          (when (eq major-mode (buffer-local-value 'major-mode buf))
            (with-current-buffer buf
              (lsp-diagnostics-mode (if disabled 1 -1)))))))
```

lsp-diagnostics.el
```diff
@@ -321,8 +345,9 @@ See https://github.com/emacs-lsp/lsp-mode."
       (lsp--warn "Unable to autoconfigure flycheck/flymake. The diagnostics won't be rendered.")))
 
     (add-hook 'lsp-unconfigure-hook #'lsp-diagnostics--disable nil t))
-   (t
-    (remove-hook 'lsp-unconfigure-hook #'lsp-diagnostics--disable t))))
+   (t (progn
+        (lsp-diagnostics-flycheck-disable)
+        (remove-hook 'lsp-unconfigure-hook #'lsp-diagnostics--disable t)))))
```
leungbk, line 350: You don't need `progn` here.

lsp-diagnostics.el
```diff
@@ -202,10 +212,14 @@ provided by lsp-mode.
 See https://github.com/emacs-lsp/lsp-mode."
       :start #'lsp-diagnostics--flycheck-start
       :modes '(lsp-placeholder-mode) ;; placeholder
-      :predicate (lambda () lsp-mode)
+      :predicate #'(lambda () lsp-diagnostics-mode)
```
leungbk, line 215: Does this need to be sharp-quoted?

thomasf: not really

thomasf:
So, any further comments? I really depend on something like this to get my daily work done. 
For some combinations of modes/projects lsp based diagnostics are great, for others they don't work at all. I really need a convenient way to toggle lsp-diagnostics on and off.

thomasf:
```diff
@@ -202,10 +212,14 @@ provided by lsp-mode.
See https://github.com/emacs-lsp/lsp-mode."
      :start #'lsp-diagnostics--flycheck-start
      :modes '(lsp-placeholder-mode) ;; placeholder
      :predicate (lambda () lsp-mode)
      :predicate (lambda () lsp-diagnostics-mode)
```
Without this change nothing really works since lsp will activate itself within flycheck regardless if it's disabled or not for a particular mode.

lsp-diagnostics.el
```diff
@@ -202,10 +212,14 @@ provided by lsp-mode.
 See https://github.com/emacs-lsp/lsp-mode."
       :start #'lsp-diagnostics--flycheck-start
       :modes '(lsp-placeholder-mode) ;; placeholder
-      :predicate (lambda () lsp-mode)
+      :predicate (lambda () lsp-diagnostics-mode)
       :error-explainer (lambda (e)
                          (lsp-diagnostics-flycheck-error-explainer
                           e (lsp--workspace-server-id (car-safe (lsp-workspaces)))))))
+  (and lsp-diagnostics--flycheck-enabled
```
yyoncho, line 219: shouldn't that be `(not  lsp-diagnostics--flycheck-enabled)`

yyoncho (commented):
Looks good

lsp-diagnostics.el
thomasf: yes, fixed now. It worked for me since I only tested with an original value of nil most of the time. Verified that it works enabling/disabling with flycheck-checker actually being set not as well.

yyoncho (approved):

yyoncho:
Thank you!

PR #2505, opened 2021-01-07
Define missing lsp-face-semhl-deprecated face

sergv:
I noticed that `lsp-face-semhl-deprecated` face is already used in https://github.com/emacs-lsp/lsp-mode/blob/ee4d07f178174e935099971affe12e2e13af908f/lsp-mode.el#L6051 and https://github.com/emacs-lsp/lsp-mode/blob/ee4d07f178174e935099971affe12e2e13af908f/lsp-mode.el#L6062 but it's not defined anywhere.


ericdallo (approved):
Good catch! thank you for your contribution!

PR #2506, opened 2021-01-07
`lsp-workspace-*`: don't prompt for one workspace

nbfalcon:
Similarly to `lsp-auto-execute-action`, `lsp-auto-accept-workspace`
controls whether `lsp-workspace-restart` and `lsp-workspace-shutdown`
should automatically act if there is only one workspace. By default,
that option is enabled, behaving like `lsp-shutdown-workspace`.

To implement this, refactor the common workspace prompting code
`lsp-workspace-*` share into a separate function, `lsp--read-workspace`,
which respects `lsp-auto-accept-workspace`.

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7534,7 +7534,7 @@ such."
 
 (make-obsolete 'lsp-shutdown-workspace 'lsp-workspace-shutdown "lsp-mode 6.1")
 
-(defcustom lsp-auto-accept-workspace t
+(defcustom lsp-auto-select-workspace t
   "Shutdown or restart a single workspace.
 If set and the current buffer has only a single workspace
 associated with it, `lsp-shutdown-workspace' and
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7534,11 +7534,26 @@ such."
 
 (make-obsolete 'lsp-shutdown-workspace 'lsp-workspace-shutdown "lsp-mode 6.1")
 
+(defcustom lsp-auto-accept-workspace t
```
kiennq, line 7537: nit: Should this be `lsp-auto-select-workspace`?
We have `lsp-restart` to automatically accept to restart workspace already. The functionality in this PR is more like automatically select workspace if there's only one.

nbfalcon: You are right, accept kind of implies a yes/no prompt. Fixed.

kiennq (approved):

PR #2511, opened 2021-01-09
Add mjs to ts-ls activation regexp

wyuenho:
Node 14 now requires the ["mjs"](https://nodejs.org/dist/latest-v14.x/docs/api/esm.html#esm_enabling) file extension to opt into ES modules.

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -45,7 +45,7 @@
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
   (or (string-match-p "\\`mjs\\|[jt]sx?\\'" (or (file-name-extension filename) ""))
-      (and (derived-mode-p 'js-mode 'js2-mode 'typescript-mode)
+      (and (derived-mode-p 'js-mode 'typescript-mode)
            (not (derived-mode-p 'json-mode)))))
 
 (lsp-register-client
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -44,7 +44,7 @@
 
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
-  (or (string-match-p (rx (one-or-more anything) "." (or "ts" "js") (opt "x") string-end) filename)
+  (or (string-match-p "\\`mjs\\|[jt]sx?\\'" (or (file-name-extension filename) ""))
       (and (derived-mode-p 'js-mode 'js2-mode 'typescript-mode)
            (not (derived-mode-p 'json-mode)))))
 
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -44,7 +44,7 @@
 
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
-  (or (string-match-p "\\`mjs\\|[jt]sx?\\'" (or (file-name-extension filename) ""))
+  (or (string-match-p "\\.mjs\\|[jt]sx?\\'" filename)
       (and (derived-mode-p 'js-mode 'typescript-mode)
            (not (derived-mode-p 'json-mode)))))
 


ericdallo (approved):

clients/lsp-javascript.el
```diff
@@ -44,7 +44,7 @@
 
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
-  (or (string-match-p (rx (one-or-more anything) "." (or "ts" "js") (opt "x") string-end) filename)
+  (or (string-match-p (rx (one-or-more anything) "." (or "mjs" (or "ts" "js") (opt "x")) string-end) filename)
```
nbfalcon, line 47: "mjs" should be part of the `or`, after "js", something like that (beware indentation!):
```suggestion
  (or (string-match-p (rx (one-or-more anything) "." (or "ts" "js" "mjs") (opt "x") string-end) filename)
 ```

**Review by nbfalcon — CHANGES_REQUESTED** [review]

File: clients/lsp-javascript.el (hunk 1 above)
Comment (nbfalcon) [reply clients/lsp-javascript.el:47]: Nevermind.

Comment (nbfalcon) [inline clients/lsp-javascript.el:47]: Actually, I think this causes a regression: `tsx` wouldn#t be matched anymore, because the `opt` is now part of another `or` branch:
```suggestion
  (or (string-match-p (rx (one-or-more anything) "." (or "mjs" (: (or "ts" "js") (opt "x"))) string-end) filename)
```

nbfalcon: Is `mjsx` a thing?

wyuenho: mjsx is not a thing

nbfalcon: Thanks for clarifying; then my `suggestion` should be correct.

clients/lsp-javascript.el
```diff
@@ -44,7 +44,7 @@
 
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
-  (or (string-match-p (rx (one-or-more anything) "." (or "ts" "js") (opt "x") string-end) filename)
+  (or (string-match-p (rx (seq (+ nonl) (submatch (or "mjs" (submatch (any ?j ?t) "s" (? "x")))) string-end)) filename)
```
nbfalcon, line 47: This is more complicated than what I have proposed (which isn't necessary since `rx` is an optimizing regex compiler) and will break for people with newlines in filenames (more of a theoretical point, since lots of other things would probably break at that point; it is possible on Linux, though).

Why not go with my `suggestion`?

nbfalcon (commented):
.

clients/lsp-javascript.el
wyuenho: I've just converted it to a read syntax to match only the extension. It's so simple, there's no need to use rx.

clients/lsp-javascript.el
```diff
@@ -44,7 +44,7 @@
 
 (defun lsp-typescript-javascript-tsx-jsx-activate-p (filename &optional _)
   "Check if the javascript-typescript language server should be enabled based on FILENAME."
-  (or (string-match-p (rx (one-or-more anything) "." (or "ts" "js") (opt "x") string-end) filename)
+  (or (string-match-p "\\`mjs\\|[jt]sx?\\'" (or (file-name-extension filename) ""))
```
nbfalcon, line 47: Can you please not change this more than needed? The `rx` expression was simpler and more maintainable. The `(one-or-more anything)` can be dropped, though:
```suggestion
  (or (string-match-p (rx "." (or "mjs" (: (or "ts" "js") (opt "x"))) string-end) filename)
```

wyuenho: Nobody speaks rx, and it offers no performance benefits since most functions that accepts regular expressions accepts a string. Regular expression is annoying enough, Emacs regular expression even more so, and now with rx, you are asking the reader to translate what they want to do 3 times in their head before arriving to a syntax (does any know what ":"  is without RTFM?). Rx only makes sense If you have a really long and complicated regular expression with lots of branches and groups that you want to compose together with functions. There are only 5 cases here that you'll likely never change again. This client and eslint are the only clients in LSP that uses rx, neither have a compelling reason to do so.

Just keep it simple.

wyuenho: Additional benefit to using my solution - you can test it in re-builder.

nbfalcon, line 47:
```suggestion
  (or (string-match-p "\\.mjs\\|[jt]sx?\\'" filename)
```

nbfalcon (approved):

nbfalcon:
It might be debatable whether `.js` constitutes a real JavaScript file, but I think that the behaviour of this commit (yes) is correct and hence this can be merged.

PR #2513, opened 2021-01-10
README: link the opencollective page from the image

leungbk:
We prefer to take users to the opencollective page, and not to a page
with the image alone.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ we need to implement new features and help the community on a lot of issues.
 
 You can help us keep going and improving it by **[supporting the project](https://opencollective.com/emacs-lsp)**
 
-<img src="https://opencollective.com/emacs-lsp/tiers/backer.svg">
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -113,7 +113,7 @@ we need to implement new features and help the community on a lot of issues.
 
 You can help us keep going and improving it by **[supporting the project](https://opencollective.com/emacs-lsp)**
 
-<img src="https://opencollective.com/emacs-lsp/tiers/backer.svg">
+<a href="https://opencollective.com/emacs-lsp"><img src="https://opencollective.com/emacs-lsp/tiers/backer.svg" /></a>
```
ericdallo, line 116: Odd, for me the image already has a link on Github README

leungbk: Yes, but the point of this commit is to change it so that clicking on the image directs to the user to the donation page rather than github's hosted copy of the image.

ericdallo (approved):

PR #2519, opened 2021-01-12
[3.16] Add support for codeAction/resolve

### LINKED ISSUE #2469: [3.16] Support CodeActionsResolve request
[SPEC](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#codeAction_resolve)

* Add client capabilities `textDocument.codeAction.resolveSupport`
* :leftwards_arrow_with_hook: `codeAction/resolve`

This will increase performance if the server supports it

ericdallo:
Fixes #2469


lsp-mode.el
```diff
@@ -5103,10 +5106,24 @@ It will show up only if current point has signature help."
...
+(defun lsp-execute-code-action (action)
   "Execute code action ACTION.
-If ACTION is not set it will be selected from `lsp-code-actions-at-point'."
+If ACTION is not set it will be selected from `lsp-code-actions-at-point'.
+Request codeAction/resolve for more info if server supports."
   (interactive (list (lsp--select-action (lsp-code-actions-at-point))))
+  (if (-> (lsp--server-capabilities)
+          (lsp:server-capabilities-code-action-provider?)
+          (lsp:code-action-options-resolve-provider?))
+      (lsp-request-async
```
yyoncho, line 5117: I would expect that to be sync. Imagine that you perform the code action and you switch the buffer - the code action won't be performed.

lsp-mode.el
```diff
@@ -5103,10 +5106,24 @@ It will show up only if current point has signature help."
...
     (lsp--send-execute-command command arguments?)))
 
-(lsp-defun lsp-execute-code-action ((action &as &CodeAction :command? :edit?))
+(defun lsp-execute-code-action (action)
   "Execute code action ACTION.
-If ACTION is not set it will be selected from `lsp-code-actions-at-point'."
+If ACTION is not set it will be selected from `lsp-code-actions-at-point'.
+Request codeAction/resolve for more info if server supports."
   (interactive (list (lsp--select-action (lsp-code-actions-at-point))))
+  (if (-> (lsp--server-capabilities)
```
yyoncho, line 5114: please use `lsp-feature?` for that(refer to `lsp-method-requirements`).

yyoncho, line 5114: shouldn't we perform resolve only if the property we are looking for is not present. This is how resolve works for code lenses if memory serves me right.

lsp-mode.el
```diff
@@ -3200,7 +3200,9 @@ disappearing, unset all the variables related to it."
                                                                                                   "refactor.inline"
                                                                                                   "refactor.rewrite"
                                                                                                   "source"
-                                                                                                  "source.organizeImports"])))))))
+                                                                                                  "source.organizeImports"])))))
+                                     (resolveSupport . ((properties . ["edit" "command" "kind"])))
```
yyoncho, line 3204: hm, we perform actions like filtering based on "kind". In that case, it seems to me that it is better to always request kind. WDYT?

lsp-mode.el
ericdallo: I thought this is expected since we wouldn't block user input, but I'll change it

lsp-mode.el
ericdallo: I searched about it but for server capabilities, I could not find how to do that, do you know how I can check for the resolveSupport via that?

ericdallo: Yeah, I thought about that too but I was not sure, in this case we should check if `command` or `workspaceEdit` is present, right?

lsp-mode.el
ericdallo: Hum, I think it makes sense, I'll remove it from here

lsp-mode.el
yyoncho: I think it is edit not workspaceEdit but yes.

yyoncho: Check how textDocument/prepareRename is handled in lsp-method-requirements.

nbfalcon: You'll need to use `(lsp-feature? "codeAction/resolve")` and to add `("codeAction/resolve" :capability :resolveSupport)` to `lsp-method-requirements`.

ericdallo: Nice, I'll change it

yyoncho: > You'll need to use `(lsp-feature? "codeAction/resolve")` and to add `("codeAction/resolve" :capability :resolveSupport)` to `lsp-method-requirements`.

resolveProvider is not top level, so a custom check-command will be needed.

yyoncho (approved):

PR #2526, opened 2021-01-14
Fix headerline disable

ericdallo:


alexey0308:
Hi @ericdallo ,

the initial value of the header-line-format may be of the form like
```
(:eval (list (propertize " " 'display '((space :align-to 0))) (semantic-stickyfunc-fetch-stickyline)))
```
I think, before enabling, the original value should be wrapped into a list and added as the last element,
so in the end it would have a value like
```
(t (:eval lsp-headerline--string) (old-value))
```
The `t` symbol results in evaluation of the first item and the second is not being evaluated then.

I guess this is the reason of the [issue](https://github.com/syl20bnr/spacemacs/issues/14313),
where one gets  `header-line-format` set by add-to-list instead to
```
((t (:eval lsp-headerline--string))
 :eval (list (propertize " " 
                         'display
                         '((space :align-to 0)))
             (semantic-stickyfunc-fetch-stickyline)))
```
and Emacs attempts to evaluate all the elements, not only the lsp one.
> A list whose car is a string or list is processed by processing each
 of the list elements recursively, as separate mode line constructs,
 and concatenating the results.
>
> A list of the form ‘(:eval FORM)’ is processed by evaluating FORM and
 using the result as a mode line construct.  Be careful--FORM should
 not load any files, because that can cause an infinite recursion.
>
> A list whose car is a symbol is processed by examining the symbol’s
 value, and, if that value is non-nil, processing the cadr of the list
 recursively; and if that value is nil, processing the cadr of the
 list recursively.

PR #2528, opened 2021-01-16
lsp-start-plain: set session-wide package-user-dir

kiennq:
Make changing `package-user-dir` and stuff in `lsp-start-plain.el` become emacs session-wide  instead of local scope like now.
This will make sure even if user install something, it will not mess up with user current Emacs installation.
And also make sure that if `package-initialize` is run again, nothing from current user package installation will get loaded into `lsp-start-plain` session.

### DIFF
--- a/scripts/lsp-start-plain.el
+++ b/scripts/lsp-start-plain.el
@@ -33,26 +33,26 @@
 
 (require 'package)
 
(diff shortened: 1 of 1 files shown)


yyoncho:
Good catch!

yyoncho (approved):

PR #2531, opened 2021-01-16
Update lsp-tramp-connection

petr-tik:
Copy the implementation that Michael Albinus suggested when
graciously helping people understand why lsp wasn't working over tramp.

https://lists.gnu.org/archive/html/emacs-devel/2020-12/msg00896.html

This replaces start-file-process-shell-command with make-process 
and moving from binary encoding to none.

> One disadvantage is, that direct asynch processes work only with the
upcoming Tramp 2.5 (that's already in Emacs master), and it works only
if the asynchronous process does not require password handling. The
latter is true, if your ssh authentication is based on keys, or if you
use Tramp's control master arguments (enabled by default).

Added a Troubleshooting section to the remote doc with examples
of the errors. Suggested solution and tips to debug other
remote LSP errors.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6734,28 +6734,33 @@ should return the command to start the LS server."
   "Create LSP stdio connection named name.
 LOCAL-COMMAND is either list of strings, string or function which
 returns the command to execute."
+  ;; Force a direct asynchronous process.
+  (add-to-list 'tramp-connection-properties
+               (list (regexp-quote (file-remote-p default-directory))
+                     "direct-async-process" t))
   (list :connect (lambda (filter sentinel name environment-fn)
-                   (let* ((final-command (lsp-resolve-final-function local-command))
-                          ;; wrap with stty to disable converting \r to \n
+                   (let* ((final-command (lsp-resolve-final-function
+                                          local-command))
+                          (_stderr (or (when generate-error-file-fn
+                                         (funcall generate-error-file-fn name))
+                                       (format "/tmp/%s-%s-stderr" name
+                                               (cl-incf lsp--stderr-index))))
                           (process-name (generate-new-buffer-name name))
-                          (wrapped-command (append '("stty" "raw" ";")
-                                                   final-command
-                                                   (list
-                                                    (concat "2>"
-                                                            (or (when generate-error-file-fn
-                                                                  (funcall generate-error-file-fn name))
-                                                                (format "/tmp/%s-%s-stderr" name
-                                                                        (cl-incf lsp--stderr-index)))))))
                           (process-environment
-                           (lsp--compute-process-environment environment-fn)))
-                     (let ((proc (apply 'start-file-process-shell-command process-name
-                                        (format "*%s*" process-name) wrapped-command)))
-                       (set-process-sentinel proc sentinel)
-                       (set-process-filter proc filter)
-                       (set-process-query-on-exit-flag proc nil)
-                       (set-process-coding-system proc 'binary 'binary)
-                       (cons proc proc))))
-        :test? (lambda () (-> local-command lsp-resolve-final-function lsp-server-present?))))
+                           (lsp--compute-process-environment environment-fn))
+                          (proc (make-process
+                                 :name process-name
+                                 :buffer (format "*%s*" process-name)
+                                 :command final-command
+                                 :connection-type 'pipe
+                                 :coding 'no-conversion
+                                 :noquery t
+                                 :filter filter
+                                 :sentinel sentinel
+                                 :file-handler t)))
+                     (cons proc proc)))
+        :test? (lambda () (-> local-command lsp-resolve-final-function
+                              lsp-server-present?))))
 
 (defun lsp--auto-configure ()
   "Autoconfigure `company', `flycheck', `lsp-ui', etc if they are installed."
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6749,10 +6749,10 @@ returns the command to execute."
   (list :connect (lambda (filter sentinel name environment-fn)
                    (let* ((final-command (lsp-resolve-final-function
                                           local-command))
-                          (_stderr (or (when generate-error-file-fn
-                                         (funcall generate-error-file-fn name))
-                                       (format "/tmp/%s-%s-stderr" name
-                                               (cl-incf lsp--stderr-index))))
+                          (stderr-buf (or (when generate-error-file-fn
+                                            (funcall generate-error-file-fn name))
+                                          (format "/tmp/%s-%s-stderr" name
+                                                  (cl-incf lsp--stderr-index))))
                           (process-name (generate-new-buffer-name name))
                           (process-environment
                            (lsp--compute-process-environment environment-fn))
@@ -6765,6 +6765,7 @@ returns the command to execute."
                                  :noquery t
                                  :filter filter
                                  :sentinel sentinel
+                                 :stderr stderr-buf
                                  :file-handler t)))
                      (cons proc proc)))
         :test? (lambda () (-> local-command lsp-resolve-final-function
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6734,6 +6734,14 @@ should return the command to start the LS server."
   "Create LSP stdio connection named name.
 LOCAL-COMMAND is either list of strings, string or function which
 returns the command to execute."
+  ;; 2.5.0-pre (as built from native-comp branch before M Albinus released tramp-2.5)
+  ;; worked fine
+  (defvar tramp-version)
+  (defvar tramp-connection-properties)
+  (when (version< "2.5.0-pre" tramp-version)
+    (lsp-warn
+     "Your tramp version - %s - might fail to work with remote LSP. Update to tramp-2.5 for tested reliability improvements"
+     tramp-version))
   ;; Force a direct asynchronous process.
   (add-to-list 'tramp-connection-properties
                (list (regexp-quote (file-remote-p default-directory))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6738,9 +6738,9 @@ returns the command to execute."
   ;; worked fine
   (defvar tramp-version)
   (defvar tramp-connection-properties)
-  (when (version< "2.5.0-pre" tramp-version)
+  (when (version< tramp-version "2.5.0-pre")
     (lsp-warn
-     "Your tramp version - %s - might fail to work with remote LSP. Update to tramp-2.5 for tested reliability improvements"
+     "Your tramp version - %s - might fail to work with remote LSP. Update to version 2.5 or greater (available on elpa)"
      tramp-version))
   ;; Force a direct asynchronous process.
   (add-to-list 'tramp-connection-properties
--- a/docs/page/remote.md
+++ b/docs/page/remote.md
@@ -30,6 +30,59 @@ _Note:_ when you do not have root privileges on the remote machine to put the la
 With TRAMP, Emacs does not have an easy way to distinguish stdout and stderr, so when the underlying LSP process writes to stderr, it breaks the `lsp-mode` parser. As a workaround, `lsp-mode` is redirecting stderr to `/tmp/<process-name>-<id>~stderr`.
 
 
+### Troubleshooting
+
+#### Remote LSP fails to starts/initializes - spinner going for a long time
+
+If you have been waiting for LSP to connect to the remote server for longer than 1-2 minutes (and your connection isn't the problem), it might be an encoding error during the first `initialize` request sent by the client. This can often be observed by an active spinner that never settles to a green tick or whatever symbol you have for successful connection. 
+
+This seems to be caused by incorrect JSON encoding leaving out the last curly brace and thus causing the LSP server to fail to parse it as valid JSON. 
+
+Examples of the same error caused by the client to different servers. 
+
+##### pyls 
+
+```bash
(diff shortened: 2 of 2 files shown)


lsp-mode.el
```diff
@@ -6490,28 +6490,41 @@ should return the command to start the LS server."
   "Create LSP stdio connection named name.
 LOCAL-COMMAND is either list of strings, string or function which
 returns the command to execute."
+  ;; 2.5.0-pre (as built from native-comp branch before M Albinus released tramp-2.5)
+  ;; worked fine
+  (defvar tramp-version)
+  (defvar tramp-connection-properties)
+  (when (version< "2.5.0-pre" tramp-version)
+    (lsp-warn
+     "Your tramp version - %s - might fail to work with remote LSP. Update to tramp-2.5 for tested reliability improvements"
```
yyoncho, line 6499: can you change the wording to mention that you can install that version from elpa?

lsp-mode.el
```diff
@@ -6490,28 +6490,41 @@ should return the command to start the LS server."
...
+  ;; Force a direct asynchronous process.
+  (add-to-list 'tramp-connection-properties
+               (list (regexp-quote (file-remote-p default-directory))
+                     "direct-async-process" t))
   (list :connect (lambda (filter sentinel name environment-fn)
-                   (let* ((final-command (lsp-resolve-final-function local-command))
-                          ;; wrap with stty to disable converting \r to \n
+                   (let* ((final-command (lsp-resolve-final-function
+                                          local-command))
+                          (_stderr (or (when generate-error-file-fn
```
yyoncho, line 6508: this should be appended to the command. Otherwise, the stderr will end up on stdout. (Unless this is fixed in the latest tramp).

yyoncho (commented):
Looks good.

lsp-mode.el
petr-tik: what do you mean by `this` here?

yyoncho: _stderr

lsp-mode.el
fjl: BTW, when I tested this function with tramp 2.5.0 from elpa, it showed the warning.

petr-tik: Thanks for testing my branch! I think I have the arguments in the wrong order

fjl:
Just some more feedback: I tried this a bit yesterday, and it didn't fix my issue with running `gopls` remotely. It could never get beyond some JSON parsing error.

I'm not sure if this PR really improves things in `lsp-tramp-connection`. There were two issues
when I tried to integrate it into my connection:

- When `lsp-tramp-connection` attempts to add the "direct-async-process" property, it tries to
  find the tramp connection using `(file-remote-p default-directory)`. This means that the call
  to `lsp-tramp-connection` must be made from within a context where the remote buffer is current,
  like a language mode hook. If I just put the `lsp-register-client` invocation into my init file, it crashes
  during init because `(file-remote-p default-directory)` is `nil`.
- Using the bare `make-process` means that stderr cannot be handled independently from stdout.
  The previous version using `start-file-process-shell-command` could do this because with a shell,
  `2> ...` can be appended to the command. Not sure if there is an equally 'direct' alternative to
  `make-process` that runs a shell command, but if there is one, it should be used here.
  I tried to hack it by setting the `make-process` command to `("/bin/sh" "-c" ... "2>" "/tmp/lsp-stderr")`
  but it didn't fix my issue with the language server.

petr-tik:
@fjl thanks for trying the branch and the feedback. To be entirely honest, I ported the implementation that was suggested by Michael Albinus and may have changed something with my carelessness. 

> Just some more feedback: I tried this a bit yesterday, and it didn't fix my issue with running gopls remotely. It could never get beyond some JSON parsing error.

Is it the same JSON parsing error that the troubleshooting section of the remote doc lists or something else? I haven't tried gopls, so would be great to get a repro.

fjl:
Tried again with lsp-mode from master branch and it seems that remote connection to gopls is working for me, even without this PR. I'm using tramp version 2.5.0. I think this might be the most important thing to recommend for people with issues: just upgrade tramp from ELPA.

The other changes here seem kind of weird, IMHO it's better to use `start-file-process-shell-command` than trying to make it work with low-level `make-process`.

fjl:
> Is it the same JSON parsing error that the troubleshooting section of the remote doc lists or something else? I haven't tried gopls, so would be great to get a repro.

I'm getting a message like `gopls: unmarshaling jsonrpc message: unexpected end of JSON input`

Also, while debugging this, I noticed that `make-process` has a `:stderr` argument. This could be used instead of the shell redirection, if it actually works with tramp.

fjl:
Now I'm wondering why `lsp-tramp-connection` even exists. The changes in this PR, make it 95% the same as `lsp-stdio-connection`.

lsp-mode.el
yyoncho: ping? The fix is just to concat `"2>" stderr` just like we are doing now to avoid stderr polluting the stdout(this was an issue with old tramp).

jdtsmith:
This PR (together with upgrading to Tramp v2.5.0.1) fixed my [issues](https://github.com/emacs-lsp/lsp-pyright/issues/10#issuecomment-780056762) with `lsp-pyright` blocking and/or freezing over tramp using Emacs 27.1. Thanks!

lsp-mode.el
petr-tik: According to the comment from fjl, this PR doesn't really fix the problem for him. 

I will change the method and then try to wire-up some integration tests with tramp - though I am not sure if that's possible on github actions

jdtsmith: The fix works for me with pyright, whereas upgrading tramp alone did not.  I do see it requires key based (non-prompting) access, which might be an inconvenience for some.

lsp-mode.el
```diff
@@ -6734,28 +6734,42 @@ should return the command to start the LS server."
...
+  ;; Force a direct asynchronous process.
+  (add-to-list 'tramp-connection-properties
+               (list (regexp-quote (file-remote-p default-directory))
+                     "direct-async-process" t))
   (list :connect (lambda (filter sentinel name environment-fn)
-                   (let* ((final-command (lsp-resolve-final-function local-command))
-                          ;; wrap with stty to disable converting \r to \n
+                   (let* ((final-command (lsp-resolve-final-function
+                                          local-command))
+                          (stderr-buf (or (when generate-error-file-fn
```
yyoncho, line 6752: if stderr buffer works for tramp 2.5 then buffer name can be generated like this `(stderr-buf (format "*%s::stderr*" process-name))`. This is what we do for normal stdio.

shanemcd:
I tried this PR and am seeing this when starting emacs:

```
  2   regexp-quote(nil)
  3   (list (regexp-quote (file-remote-p default-directory)) "direct-async-process" t)
  4   (add-to-list 'tramp-connection-properties (list (regexp-quote (file-remote-p default-directory)) "direct-async-process" t))
  5   lsp-tramp-connection("pyls")
  6   (make-lsp-client :new-connection (lsp-tramp-connection "pyls") :major-modes '(python-mode) :remote\? t :server-id 'pyls-remote)
  7   (lsp-register-client (make-lsp-client :new-connection (lsp-tramp-connection "pyls") :major-modes '(python-mode) :remote\? t :server-id 'pyls-remote))
  8   (lambda nil (lsp-register-client (make-lsp-client :new-connection (lsp-tramp-connection "pyls") :major-modes '(python-mode) :remote\? t :server-id 'pyls-remote)))()
  9   eval-after-load("lsp-mode" (lambda nil (lsp-register-client (make-lsp-client :new-connection (lsp-tramp-connection "pyls") :major-modes '(python-mode) :remote\? t :server-id 'pyls-remote))))
 10   load-with-code-conversion("/Users/shanemcd/.emacs.d/init.el" "/Users/shanemcd/.emacs.d/init.el" t t)
 11   load("/Users/shanemcd/.emacs.d/init" noerror nomessage)
 12   startup--load-user-init-file(#f(compiled-function () #<bytecode -0x148f9fb91d5576b>) #f(compiled-function () #<bytecode -0x1f3c686ddc0d7fb5>) t)
 13   command-line()
 14   normal-top-level()
```

Blade6570:
after I apply this patch with tramp 2.5.1, I am getting this error `tramp-signal-hook-function: Wrong type argument: "bufferp *pyright-remote::stderr*"`. Any idea?

Blade6570:
Okay the above error can be fixed by calling generate-new-buffer command. Now in the stderr buffer, I only get this message `Process pyright-remote stderr finished` and lsp does not connect to server. It disconnects immediately.

lsp-mode.el
Blade6570: It complains that stderr-buf does not exist so (stderr-buf (generate-new-buffer stderr-buf)) worked.

Blade6570:
This fork with tramp 2.5.0 works in my case also. My config is given below. First call M-x register-pyright and then call M-x lsp. 

```
       (defun register-pyright ()
       (interactive)
       ;; (lsp-register-client (make-lsp-client :new-connection (lsp-tramp-connection "pyright")
       ;; 				    :major-modes '(python-mode)
       ;; 				    :remote? t
       ;; 				    :multi-root t
       ;; 				    :server-id 'pyright-remote)))
       (setq lsp-log-io t)
       (lsp-register-client
	 (make-lsp-client
	   :new-connection (lsp-tramp-connection (lambda ()
					 (cons "pyright-langserver"
					       lsp-pyright-langserver-command-args)))
	   :major-modes '(python-mode)
	   :remote? t
	   :server-id 'pyright-remote
	   :multi-root nil
	   :priority 3
    :initialization-options (lambda () (ht-merge (lsp-configuration-section "pyright")
							(lsp-configuration-section "python")))
			     (with-lsp-workspace workspace
			       (lsp--set-configuration
			       (ht-merge (lsp-configuration-section "pyright")
					 (lsp-configuration-section "python")))))
	   :download-server-fn (lambda (_client callback error-callback _update?)
			       (lsp-package-ensure 'pyright callback error-callback))
	   :notification-handlers (lsp-ht ("pyright/beginProgress" 'lsp-pyright--begin-progress-callback)
				       ("pyright/reportProgress" 'lsp-pyright--report-progress-callback)
				       ("pyright/endProgress" 'lsp-pyright--end-progress-callback)))))

```

billop:
I can't figure out how to get this working. I'm using latest emacs with tramp version 2.5.1-pre. Using https://github.com/petr-tik/lsp-mode/tree/tramp_connection_compliant_with_2_5 for lsp-mode and copied the custom server registration config.

But all I get is:

```
Wrong type argument: "Wrong type argument", "bufferp /tmp/pyright-remote-1-stderr"
File mode specification error: (wrong-type-argument bufferp /tmp/pyright-remote-1-stderr)
```

Blade6570:
@billop please try my fork of lsp-mode. You need to create this buffer like this `(generate-new-buffer stderr-buf)` before passing it to `lsp-tramp-connection`.

Ergus:
Hi:

I added this to my config:
```
(lsp-register-client
   (make-lsp-client :new-connection (lsp-tramp-connection "clangd")
                    :major-modes '(c-mode c++-mode)
                    :remote? t
                    :server-id 'clangd-remote))
```
trying to make lsp to work with tramp.

But when I start the process I get an error:
```
: (wrong-number-of-arguments (3 . 3) 7)
```

activating the debug command I get this backtrace 

```
  start-file-process-shell-command("clangd-remote" "*clangd-remote*" "stty" "raw" ";" "clangd" "2>/tmp/clangd-remote-2-stderr")
  apply(start-file-process-shell-command "clangd-remote" "*clangd-remote*" ("stty" "raw" ";" "clangd" "2>/tmp/clangd-remote-2-stderr"))
  #f(compiled-function (filter sentinel name environment-fn) #<bytecode 0xd063bb24a084754>)(#f(compiled-function (proc input) #<bytecode 0x15fa3033f0f64413>) #f(compiled-function (&rest args2) #<bytecode -0x1ae4bbec3439043a>) "clangd-remote" nil)
  (...)
  lsp--ensure-lsp-servers(
  lsp--try-project-root-workspaces(nil nil)
  lsp(nil)
  funcall-interactively(lsp nil)
  call-interactively(lsp record nil)
  command-execute(lsp record)
  counsel-M-x-action("lsp")
```

So it seems that lsp-tramp-connection is somehow outdated?

lsp-mode.el
```diff
@@ -6734,28 +6734,42 @@ should return the command to start the LS server."
...
+  ;; worked fine
+  (defvar tramp-version)
+  (defvar tramp-connection-properties)
+  (when (version< tramp-version "2.5.0-pre")
+    (lsp-warn
+     "Your tramp version - %s - might fail to work with remote LSP. Update to version 2.5 or greater (available on elpa)"
+     tramp-version))
+  ;; Force a direct asynchronous process.
+  (add-to-list 'tramp-connection-properties
+               (list (regexp-quote (file-remote-p default-directory))
```
muirdm, line 6747: I don't think this is right. `lsp-tramp-connection` is run when the user defines the lsp client, which may not be in a remote file (in which case the user gets an error). Plus, this is adding the entire directory to `tramp-connection-properties`, so it will grow forever as the user connects to different hosts. Don't we just need to add `("ssh:/" "tramp-direct-async" t)` to the list)?

lsp-mode.el
muirdm: Something like this seems to work:
```elisp
:stderr (get-buffer-create (format "*%s::stderr*" process-name))
```

lsp-mode.el
yyoncho: I think that like comes from the tramp maintainer. I am not 100% sure how significant it is.

antifuchs: Trying this out, I can confirm that this errors if you're defining a tramp connection with elisp code in an indirect buffer (that is, one without an associated file name). I think the add-to-list here should happen inside the :connect lambda.

antifuchs:
I tried this on my system, and the stdio connection over SSH does indeed work much better than it did before, once I applied @Blade6570's stderr change, and moved the tramp hook modification into the connection lambda. 

One langserver that still doesn't work quite right for me is `pyrite-langserver --stdio` (it hangs on startup): from strace'ing the ssh child processes on the remote system, it looks like the data that lsp-mode sends over the tramp connection is distorted (the lsp process expects `\r\n` and receives `\n\n` line endings):

```
read(17</dev/pts/12<char 136:12>>, "Content-Length: 2313\n", 65536) = 21
write(19</dev/pts/12<char 136:12>>, "Content-Length: 119\r\n\r\n", 23) = 23
write(19</dev/pts/12<char 136:12>>, "{\"jsonrpc\":\"2.0\",\"method\":\"window/logMessage\",\"params\":{\"type\":3,\"message\":\"Pyright language server 1.1.145 starting\"}}", 119) = 119
write(19</dev/pts/12<char 136:12>>, "Content-Length: 197\r\n\r\n", 23) = 23
write(19</dev/pts/12<char 136:12>>, "{\"jsonrpc\":\"2.0\",\"method\":\"window/logMessage\",\"params\":{\"type\":3,\"message\":\"Server root directory: /nix/store/bh95vd3sk72mw5lm7v23864lm0fryf70-node_pyright
-1.1.145/lib/node_modules/pyright/dist/\"}}", 197) = 197
epoll_wait(13<anon_inode:[eventpoll]>, [{events=EPOLLIN, data={u32=17, u64=17}}], 1024, 0) = 1
read(17</dev/pts/12<char 136:12>>, "\n", 65536) = 1
epoll_wait(13<anon_inode:[eventpoll]>, [{events=EPOLLIN, data={u32=17, u64=17}}], 1024, 0) = 1
read(17</dev/pts/12<char 136:12>>, "\n", 65536) = 1
epoll_wait(13<anon_inode:[eventpoll]>, [{events=EPOLLIN, data={u32=17, u64=17}}], 1024, 0) = 1
read(17</dev/pts/12<char 136:12>>, "\n", 65536) = 1
epoll_wait(13<anon_inode:[eventpoll]>, [{events=EPOLLIN, data={u32=17, u64=17}}], 1024, 0) = 1
read(17</dev/pts/12<char 136:12>>, "{\"jsonrpc\":\"2.0\",\"method\":\"initialize\",\"params\":{\"processId\":null,\"rootPath\":\"/home/asf/Hacks/cloud\",\"clientInfo\":{\"name\":\"emacs\",\"version\":\"GNU E[...]
```

yyoncho:
@antifuchs this sounds like  https://github.com/emacs-lsp/lsp-mode/issues/2709#issuecomment-864473829 

it seems like tramp does this odd conversation when the process is not initialized on the remote(which IMHO is quite odd behavior).

antifuchs:
Hm, I can confirm that an `stty raw` seems necessary in this case, as well. Seems thoroughly weird that it would be, especially considering that the LSP's process's coding system (as well as the backing tramp process's coding system) is reported as `(utf-8-unix . utf-8-unix)`. 

Wonder if tramp forces any conversion in the data being sent for any other reason.

Ergus:
Hi:

I haven't followed too much this thread for a while but I reported something similar to the coding issues with tramp and lsp servers in eglot some time ago. Please check there if this is related with an existing issue in tramp already fixed in the newest tramp release. Otherwise check if this needs to be reported as an Emacs or tramp bug please.

https://github.com/joaotavora/eglot/issues/662

lsp-mode.el
toshokan: It might not even be useful in this context.
See [the tramp source code](https://github.com/emacs-mirror/emacs/blob/1ff02c53b5f21dfcaff1794f3eb2f42fb6160dae/lisp/net/tramp.el#L4077)
`:buffer` and `:stderr` will be defined for this `make-process`, but those seem to prevent using `direct-async`.

sunhs:
Any update on this ?

yyoncho:
closing in favour of: https://github.com/emacs-lsp/lsp-mode/pull/4204

PR #2533, opened 2021-01-17
Apply other Emacs version for CI test

jcs090218:


jcs090218:
Hmm... interesting fact. CI does not passed on windows snapshot.

yyoncho (approved):

jcs090218:
@yyoncho Would you prefer to fix the CI in this PR or in the separate PR (next)?

yyoncho:
It is better to keep the ci green

jcs090218:
Okay, I will work in this PR then.

jcs090218:
There is an error on windows version `26.1`.

```
1 unexpected results:
   FAILED  lsp--uri-to-path--handle-utf8
```

Not quite sure how to fix this. 😕

yyoncho:
most likely we should run only the windows specific part of the test on windows.

jcs090218:
@yyoncho Aren't we already doing that? This looks strange to me because this only failed on 26.1. Other Emacs version 26.2, 26.3, 27.1, snapshot on Windows have successfully passed the test. I doubt this is an OS specific error.

yyoncho:
> @yyoncho Aren't we already doing that? This looks strange to me because this only failed on 26.1. Other Emacs version 26.2, 26.3, 27.1, snapshot on Windows have successfully passed the test. I doubt this is an OS specific error.

I was thinking about something like:

```
(ert-deftest lsp--uri-to-path--handle-utf8 ()
  (when (eq system-type 'windows-nt)
    (let ((lsp--uri-file-prefix "file:///"))
      (should (equal (lsp--uri-to-path "file:///c:/Users/%E4%BD%A0%E5%A5%BD/") "c:/Users/你好/"))))
  (when (eq system-type 'gnu/linux)
    (let ((lsp--uri-file-prefix "file://"))
      (should (equal (lsp--uri-to-path "/root/%E4%BD%A0%E5%A5%BD/%E8%B0%A2%E8%B0%A2") "/root/你好/谢谢")))))
```

yyoncho:
Or even:

```
(ert-deftest lsp--uri-to-path--handle-utf8 ()
  (when (eq system-type 'windows-nt)
    (should (equal (lsp--uri-to-path "file:///c:/Users/%E4%BD%A0%E5%A5%BD/") "c:/Users/你好/")))
  (when (eq system-type 'gnu/linux)
    (should (equal (lsp--uri-to-path "/root/%E4%BD%A0%E5%A5%BD/%E8%B0%A2%E8%B0%A2") "/root/你好/谢谢"))))
```

jcs090218:
Hmm... no luck with both tests.

d802ff4d982e920c5d3087e7dcacf74babb399b2 and 245687eec3bf611862f6e27f014500c97c307f9a.

yyoncho:
I think that it may be related to something env specific (os, build flags, etc).

```
  ((should (equal
            	(lsp--uri-to-path "file:///c:/Users/%E4%BD%A0%E5%A5%BD/")
	         "c:/Users/  /"))
      :form  (equal  #("c:/Users/你好/" 0 16 (charset iso-8859-1))
                             "c:/Users/  /")
````

As you can see the second part of the form is not "c:/Users/  /" but hardcoded literal:  "c:/Users/你好/". For some reason, emacs is converting it. I don't have enough knowledge about that. Someone from the core team might know about that.

jcs090218:
Who would you recommend me to ask? Maybe the author who created this specific test?

yyoncho:
Eli will know I guess.

jcs090218:
> Eli will know I guess.

I am sorry. Who?

---

I have figured out the issue. In Emacs 26.1 the variable `local-coding-system` is default to `cp1252` and not `utf-8`. Should we just ignore the variable `local-coding-system`? See 2ee7d9595eacdc96f263d6d25251b9bd9f663c31

yyoncho:
> > Eli will know I guess.
> 
> I am sorry. Who?

Eli Zaretskii - Emacs maintainer.

> 
> I have figured out the issue. In Emacs 26.1 the variable `local-coding-system` is default to `cp1252` and not `utf-8`. Should we just ignore the variable `local-coding-system`? See [2ee7d95](https://github.com/emacs-lsp/lsp-mode/commit/2ee7d9595eacdc96f263d6d25251b9bd9f663c31)

I don't know what is the right thing here. IMHO it will be safer to fix it the test.

jcs090218:
How about by setting `local-coding-system` to `utf-8` temporarily in the scope of the test? I wonder what's the best solution here. If there is no reason to respect to `local-coding-system` and `utf-8` is mandatory to LSP then I think we should fix it in inside the core?

yyoncho:
> How about by setting `local-coding-system` to `utf-8` temporarily in the scope of the test?

This is what I am proposing.

jcs090218:
Okay, I think this is good to merge! Let me know if anything you want me to do.

PR #2534, opened 2021-01-17
Implement lsp-eslint--confirm-local and update ESLint language server

Sorixelle:
- Implement `lsp-eslint--confirm-local` to handle `eslint/confirmESLintExecution` reqests
- Update the ESLint language server from 2.1.8 -> 2.1.14
- Create `lsp-ask-question` as a way to queue up minibuffer prompts

We should probably find a way to give users the ability to persist the value of `lsp-eslint--stored-libraries`. Using desktop.el and `desktop-globals-to-save` is one option, but some users may not want to use all the other features desktop.el provides. We should probably figure this one out before merging.

### DIFF
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -26,6 +26,7 @@
 
 (require 'lsp-protocol)
 (require 'lsp-mode)
+(require 'desktop)
 
 (defconst lsp-eslint/status-ok 1)
 (defconst lsp-eslint/status-warn 2)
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -7054,6 +7054,35 @@ VERSION is the version of the extension, defaults to `latest'"
...
+    (lsp--process-question-queue)))
+
+(defun lsp--process-question-queue ()
+  "Take the first question from `lsp--question-queue', process it, then process
+the next question until the queue is empty."
+  (-let* (((&alist "question" "options" "callback") (car lsp--question-queue))
+         (answer (completing-read question options nil t)))
+    (pop lsp--question-queue)
+    (funcall callback answer)
+    (unless (eq lsp--question-queue nil)
```
yyoncho, line 7083: nit: no need from `eq`

clients/lsp-eslint.el
```diff
@@ -253,10 +253,30 @@ source.fixAll code action."
         first-argument-exist
       (executable-find (cl-first eslint-server-command)))))
 
-(defun lsp-eslint--confirm-local (_workspace _params)
-  ;; don't bother implementing since it will be dropped in next version of the
-  ;; server
-  t)
+(defvar lsp-eslint--stored-libraries (ht)
```
yyoncho, line 256: nit: make this public if we are going to use it in combination with desktop-mode

yyoncho (commented):
Looks good.

> We should probably find a way to give users the ability to persist the value of lsp-eslint--stored-libraries.

Feel free to implement persistence. In general, we favour a bit more usability over purity.

clients/lsp-eslint.el
```diff
@@ -253,10 +253,30 @@ source.fixAll code action."
...
+(lsp-defun lsp-eslint--confirm-local (_workspace (&eslint:ConfirmExecutionParams :library-path) callback)
+  (if-let (remembered-answer (gethash library-path lsp-eslint--stored-libraries))
+      (if (eq remembered-answer t)
+          (funcall callback 4)
+        (funcall callback 1))
+    (lsp-ask-question
+     (format "Allow lsp-mode to execute %s?" library-path)
+     '("Always" "Yes" "No" "Never")
+     (lambda (response)
+       (cond ((equal response "Always")
```
yyoncho, line 270: consider using alist to map String->Number or restructoring the code to something like

```
(let ((a "Yes"))
  (funcall 'identity 
           (cond 
            ((string= "Yes" a) 10))))
```

clients/lsp-eslint.el
Sorixelle: I've added it to `desktop-globals-to-save` and added a `(require 'desktop)` at the top for persistence - is this the right way to do that?

yyoncho: If we are going to use desktop-save-mode it should be in the docs, not in the code. If we do it in the code we will force heavy dependency unconditionally.

You may implement the persistence similar to what we do for the session, check `lsp--persist-session`.

Sorixelle: Ahh, gotcha. Was talking a look at `lsp--persist-session` and noticed that sessions have a `metadata` field on them - would that be suitable for storing this information?

yyoncho: In general, yes, but the issue is that dap stores in the metadata information whcih cannot be persisted. As you can see, lsp--persist-session strips the metadata. We may strip the dap-specific stuff though and use that field.

yyoncho: What you can do is to use the metadata in this PR and later we can fix dap-mode issue.

Sorixelle: Bah, just missed that comment - I implemented it using `lsp--persist` and `lsp--read-from-file`. Which one do we want to go for?

yyoncho: I am personally fine with this solution(in general, the requirements are less strict for non-core files). The only issue is that we are creating more work for https://github.com/emacscollective/no-littering . I can merge it as it is.

yyoncho: I can merge it as it is.

yyoncho (approved):

clients/lsp-eslint.el
Sorixelle: Sounds good to me. We can move it to metadata once the dap-mode issue is fixed.

yyoncho:
@Sorixelle can you fix the CI - https://github.com/emacs-lsp/lsp-mode/pull/2534/checks?check_run_id=1721245305#step:6:125

Sorixelle:
> @Sorixelle can you fix the CI - https://github.com/emacs-lsp/lsp-mode/pull/2534/checks?check_run_id=1721245305#step:6:125

Looks good now.

yyoncho:

yyoncho:
@Sorixelle I noticed that you might get 

> Warning (lsp-mode): Unknown request method: eslint/probeFailed

with eslint - willing to fix that as well?

Sorixelle:
I think all that needs to be done for that one is to display a message to the user saying there was a misconfiguration somewhere (don't remember the specifics off the top of my head). I'll look into it later today.

PR #2543, opened 2021-01-19
Update use-package declaration in official website! 

utkarsh181:


ericdallo (approved):

yyoncho:
There were complaints that it does not work in :init sections and it was moved before the use-package

utkarsh181:
I have tested it with emacs-27.1 and emacs-master branch.

yyoncho:
> There were complaints that it does not work in :init sections and it was moved before the use-package

I tested it now, and it seems like it is working fine.

yyoncho:
Thank you!

PR #2555, opened 2021-01-22
Fixed bad request while install package through gnu elpa

jcs090218:


test/windows-bootstrap.el
```diff
@@ -30,7 +30,7 @@
                                          user-emacs-directory))
 
 (let* ((package-archives '(("melpa" . "https://melpa.org/packages/")
-                           ("gnu" . "https://elpa.gnu.org/packages/")))
+                           ("gnu" . "http://elpa.gnu.org/packages/")))
```
nbfalcon, line 33: This has security and privacy implications. Are you sure there is no other way?

jcs090218: Sometimes it works; sometimes it does not work. 😕 I have attempted to fix for all other my packages and my config but no luck. Feel free to correct, and I indeed want a better solution. Otherwise, I don't think there is other way to get around this issue.

jcs090218:
Now I have re-run the same job and now it works. See [here](https://github.com/emacs-lsp/lsp-mode/actions/runs/503369705). It wasn't working because of the `Bad request` from GNU ELPA.

jcs090218:
I think this no longer happens! Close this now! :)

PR #2558, opened 2021-01-24
Add reason-mode to ocaml-lsp-server client

briangorman:
I verified that this works.

I'm not sure why `ocaml-ls` has reason-mode added, but `ocaml-lsp-server` did not.


jcs090218 (approved):
LGTM

PR #2559, opened 2021-01-24
lsp--persist: create parent directories if needed

DamienCassou:
This makes it easier for users to configure variables storing filenames, e.g., `lsp-eslint-library-choices-file`.


yyoncho (approved):

yyoncho:
Thank you!

PR #2567, opened 2021-01-27
Implemented progress reporting in the modeline

yyoncho:
- The new reporting uses the modeline
- the old reporting can be accessed via
```
(setq lsp-progress-function #'lsp-on-progress-legacy)
```
- disable progress reporting via

```
(setq lsp-progress-function #'ignore)
```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -93,6 +93,13 @@
 
 (defconst lsp--empty-ht (make-hash-table))
 
+(eval-and-compile
+  (defun dash-expand:&lsp-wks (key source)
+    `(,(intern-soft (format "lsp--workspace-%s" (eval key))) ,source))
+
+  (defun dash-expand:&lsp-cln (key source)
+    `(,(intern-soft (format "lsp--client-%s" (eval key))) ,source)))
+
 (define-obsolete-variable-alias 'lsp-print-io 'lsp-log-io "lsp-mode 6.1")
 
 (defcustom lsp-log-io nil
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -1795,9 +1802,57 @@ PARAMS - the data sent from WORKSPACE."
...
+                                 (format "%s%%%% " percentage?)
+                               "")
+                             (or message? title)))
+                   (ht-values tokens)
+                   "|"))))
+            (lsp-workspaces)))))
+    (unless (s-blank? progress-status)
+      (concat lsp-progress-prefix progress-status))))
+
+(lsp-defun lsp-on-progress-modeline (workspace (&ProgressParams :token :value
```
ericdallo, line 1842: Shouldn't this be on `lsp-modeline.el`?

yyoncho: The progress reporting is defined in lsp-mode.el and also it can target different display options(modeline, echo area, logs(eventually)). What is your idea?

ericdallo: Yes I got it, my opinion is just that modeling functions could stay on lsp-modeline.el, this is the only function that I think that makes sense go to there

yyoncho: I think that we should group by feature, not by display target. Automatic installation and major modes in lsp-mode also use the modeline but splitting them will just cause trouble to do the fact that lsp-mode.el will start depending on lsp-modeline.el and vice versa.

ericdallo: I see, I thought this function was kind of optional since user can change it to use other progress function, but I'm ok keeping it here so

ericdallo (approved):

lsp-mode.el
```diff
@@ -1795,9 +1802,57 @@ PARAMS - the data sent from WORKSPACE."
...
+  :type '(choice
+          (const :tag "Use modeline" lsp-on-progress-modeline)
+          (const :tag "Legacy(uses either `progress-reporter' or `spinner' based on `lsp-progress-via-spinner')"
+                 lsp-on-progress-legacy)
+          (const ignore :tag "Ignore")
+          (function :tag "Other function"))
+  :package-version '(lsp-mode . "7.1.0"))
+
+(defun lsp--progress-status ()
+  (-let ((progress-status
```
ericdallo, line 1823: Missing docstring :p

yyoncho: Fixed.

PR #2576, opened 2021-02-01
Leave default way to render str in lsp--render-str

liushihao456:
By using `if-let` instead of `when-let`, we can return `str` itself if `mode` becomes `nil`. I made this change because in some cases, e.g. where I use web-mode to edit typescript files (this mode-language pair does not exist in `lap-language-id-configuration`), `mode` is `nil`, and then `when-let` will just devour the str and return nil so no sideline hovers will show up in the end.


yyoncho (approved):
Good catch!

PR #2577, opened 2021-02-01
Update CPP-guide.md

youngpm:
`(add-hook 'cpp-mode-hook 'lsp)` doesn't work (took me a while to figure out why things weren't auto starting), needs to be `++`.


jcs090218 (approved):

jcs090218:
Thanks!

PR #2579, opened 2021-02-02
Make sure that json server runs for json files even in js-mode

yyoncho:
- introduced `lsp-activate-on` which mimic what vscode does - it selects the
language server based on the language in the file. This will make the language
servers a bit easier to configure and it will give more control to the users via
`lsp-language-id-configuration`.

### DIFF
--- a/clients/lsp-json.el
+++ b/clients/lsp-json.el
@@ -62,9 +62,6 @@
   `(:provideFormatter t
     :handledSchemaProtocols ["file" "http" "https"]))
 
-(defvar lsp-json--major-modes '(json-mode jsonc-mode)
-  "List of supported JSON major modes.")
-
 (defvar lsp-json--schema-associations
   `(:/*.css-data.json ["https://raw.githubusercontent.com/Microsoft/vscode-css-languageservice/master/docs/customData.schema.json"]
     :/package.json ["http://json.schemastore.org/package"]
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
Nice!

yyoncho:
@alanz - I think that this may help with erlang/elp server selection issue in case the languages are different for both.

alanz:
> I think that this may help with erlang/elp server selection issue in case the languages are different for both.

Thanks, I'll see how it goes when it updates on MELPA

jcs090218 (approved):

kiennq:
@yyoncho Is there a way to override buffer language from clients (`lsp-json` in this case)?
With this change, the buffer language is fixed into `json` for `.json` files and `jsonc` for `.jsonc` files instead of base on current major modes.
For VSCode `settings.json`, it's language id is `jsonc` instead of `json` so won't we need to let the `lsp-mode` clients to be able to override `lsp-buffer-language` in this case?

yyoncho:
> @yyoncho Is there a way to override buffer language from clients (`lsp-json` in this case)?
> With this change, the buffer language is fixed into `json` for `.json` files and `jsonc` for `.jsonc` files instead of base on current major modes.

json-mode/jsonc-mode can be put before the pattern match.

PR #2581, opened 2021-02-04
lsp-buffer-language: use lsp--client-language-id first to determine language-id

kiennq:
Fix for proposal in https://github.com/emacs-lsp/lsp-mode/pull/2579#issuecomment-772968197

### DIFF
--- a/clients/lsp-go.el
+++ b/clients/lsp-go.el
@@ -141,14 +141,14 @@ The returned type provides a tri-state that either:
   - sets element to true (actually, t)
 "
   (let ((list '()))
-	(dolist (v alist)
-	  (push `(cons
-			  :tag ,(cdr v)
-			  (const :format "" ,(car v))
-			  (choice (const :tag "Enable" t) (const :tag "Disable" :json-false)))
-			list))
-	(push 'set list)
-	list))
+    (dolist (v alist)
+      (push `(cons
+              :tag ,(cdr v)
+              (const :format "" ,(car v))
+              (choice (const :tag "Enable" t) (const :tag "Disable" :json-false)))
+            list))
+    (push 'set list)
+    list))
 
 (define-obsolete-variable-alias
   'lsp-gopls-codelens
@@ -217,10 +217,14 @@ $GOPATH/pkg/mod along with the value of
 (defcustom lsp-go-link-target "godoc.org"
   "Which website to use for displaying Go documentation."
   :type '(choice (const "godoc.org")
-		 (const "pkg.go.dev")
-		 (string :tag "A custom website"))
+                 (const "pkg.go.dev")
+                 (string :tag "A custom website"))
   :group 'lsp-go
   :package-version '(lsp-mode "7.0.1"))
+
+(defvar lsp-go--language-id-configuration
+  '((".*\\.go$" . "go")
+    (".*/go.mod$" . "go.mod")))
 
 (lsp-register-custom-settings
  '(("gopls.usePlaceholders" lsp-go-use-placeholders t)
@@ -234,7 +238,9 @@ $GOPATH/pkg/mod along with the value of
  (make-lsp-client :new-connection (lsp-stdio-connection
                                    (lambda () (cons lsp-go-gopls-server-path lsp-go-gopls-server-args)))
                   :major-modes '(go-mode go-dot-mod-mode)
-                  :language-id "go"
+                  :language-id (lambda (buf)
+                                 (with-current-buffer buf
+                                   (lsp--buffer-language lsp-go--language-id-configuration)))
                   :priority 0
                   :server-id 'gopls
                   :completion-in-comments? t
--- a/clients/lsp-json.el
+++ b/clients/lsp-json.el
@@ -84,6 +84,11 @@
     :/.eslintrc ["http://json.schemastore.org/eslintrc"])
   "Default json schemas.")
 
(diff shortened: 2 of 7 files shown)


kiennq:
@yyoncho Can you check if we can use this PR?
Instead of relying on centralized `lsp-language-id-configuration` to determine language id (and need sophisticated ordering), we can let language client decide that, which is easier to manage for each client

yyoncho:
> @yyoncho Can you check if we can use this PR?
> Instead of relying on centralized `lsp-language-id-configuration` to determine language id (and need sophisticated ordering), we can let language client decide that, which is easier to manage for each client

In vscode it is not the server telling what the language is but it is the language telling which language server to start. I want to be able to tell - use that language server for this language. Using `lsp-language-id-configuration` is not more complicated than `auto-mode-alist`. If your concern is the fact that lsp-language-id-configuration will grow we could still move it to the language extension file, like currently major modes do.

```
;;;###autoload (add-to-list 'auto-mode-alist '("\\.dart\\'" . dart-mode))
```

Here it is how vscode does it - https://github.com/platelk/vscode-dart/blob/master/package.json#L76 .

kiennq:
> Using `lsp-language-id-configuration` is not more complicated than `auto-mode-alist`. If your concern is the fact that lsp-language-id-configuration will grow we could still move it to the language extension file, like currently major modes do.

No it's not because of `lsp-language-id-configuration` will grow, it's because we're using same `lsp-language-id-configuration` for every client, which make the order of entry in it becomes significant.
If we move it to the language extension file, I doubt we can have correct/desirable order in it.

Not only that, for example if there are two clients activated on the same file, but will sent different language-id to their own servers, the centralized list cannot support that.

This PR is reenable the usage of `lsp--client-language-id` so  that each client can do closely resemble to VsCode, declare its own language id when it has been activated.

yyoncho:
> If we move it to the language extension file, I doubt we can have correct/desirable order in it.

there is no such issue so let's keep the code as simple as possible. If VScode is using one central language repository and it is doing fine, then most likely it will be good enough for us. Let's revisit this PR once we have a practical issue to solve. For now, it seems to me that we can handle all cases we currently have - at least I am not aware of a pending issue without a relativelly simple solution(like reordering of the language id).

kiennq:
> If VScode is using one central language repository and it is doing fine, then most likely it will be good enough for us.

Is VsCode using one central language repository?
I've try to find the `go.mod` language id in [VsCode repository](https://github.com/microsoft/vscode/search?q=go.mod) and seems there's nothing. Similarly, even the language id in the [LSP spec](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-17.md#-textdocumentitem-) doesn't contains `clojure`/`clojurescript` too.
I thought each LSP client for VsCode declares their own language id (based on extension?) via their `package.json`, don't them?


> Let's revisit this PR once we have a practical issue to solve. For now, it seems to me that we can handle all cases we currently have - at least I am not aware of a pending issue without a relativelly simple solution(like reordering of the language id).

I see.
The change in this PR is really simple, only around 30 lines of real code changes. The other stuff that go mixed in is the reformat of the buffer to reindent, uniformly using spaces and document ...
Since we've already has the `lsp--client-language-id` (and some clients already use it, just with no effect), with this PR it will work again, do you think this PR is still redundance?

yyoncho:
> > If VScode is using one central language repository and it is doing fine, then most likely it will be good enough for us.
> 
> Is VsCode using one central language repository?
> I've try to find the `go.mod` language id in [VsCode repository](https://github.com/microsoft/vscode/search?q=go.mod) and seems there's nothing. Similarly, even the language id in the [LSP spec](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-17.md#-textdocumentitem-) doesn't contains `clojure`/`clojurescript` too.
> I thought each LSP client for VsCode declares their own language id (based on extension?) via their `package.json`, don't them?

By central I mean that it is registered in central place and it is not tight to the client. E. g. for angular language server you don't define which files are html you just say that you want to start it in when the current language is html.

> 
> > Let's revisit this PR once we have a practical issue to solve. For now, it seems to me that we can handle all cases we currently have - at least I am not aware of a pending issue without a relativelly simple solution(like reordering of the language id).
> 
> I see.
> The change in this PR is really simple, only around 30 lines of real code changes. The other stuff that go mixed in is the reformat of the buffer to reindent, uniformly using spaces and document ...
> Since we've already has the `lsp--client-language-id` (and some clients already use it, just with no effect), with this PR it will work again, do you think this PR is still redundance?

The change itself is simple - I am worried that we are introducing new properties related to client registration while in fact, we can remove one(:major-modes). ATM I think that we could delete all custom :activate-fn + :major-mode and use lsp-activate-on + language-id configuration which should lead to easy to understand interface.

yyoncho:
As a side note, the other thing that we need is "activate-on" when the root has that stuff.

kiennq:
> . ATM I think that we could delete all custom :activate-fn + :major-mode and use lsp-activate-on + language-id configuration which should lead to easy to understand interface.

I agree we can remove `:major-mode` since it has been super set by `:activate-fn`.
On other hand, I feel we try to shoes horn too much of thing on language-id configuration, even the syntax rendering is using the major-mode/function on that list to render, which makes it has hard dependency on `major-mode`.

Well, I still need `settings.json` to be in different language-id though, will open another PR for that

PR #2585, opened 2021-02-06
Fix option :none for lsp-diagnostics-provider

yuezhu:
According to https://emacs-lsp.github.io/lsp-mode/tutorials/how-to-turn-off/ and the code comment, setting `:none` for `lsp-diagnostics-provider` should use neither flycheck nor flymake. So, when it is set, `lsp-diagnostics-mode` should not be enabled.

### DIFF
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -34,7 +34,7 @@
     (const :tag "Pick flycheck if present and fallback to flymake" :auto)
     (const :tag "Pick flycheck" :flycheck)
     (const :tag "Pick flymake" :flymake)
-    (const :tag "Use neither flymake nor lsp" :none)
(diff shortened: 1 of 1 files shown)


kiennq:
Perhaps we need new `defcustom` of `lsp-enable-diagnostic` for this.
The purpose of `lsp-diagnostic-provider` as `none` is not to let `lsp-mode` auto configure the diagnostic renderer (either `flycheck` or `flymake`) and let user to configure it instead.
So do not enable `lsp-diagnostics-mode` is not correct.

yuezhu:
Thanks for the explanation. Yep, I agree we need a separate `defcustom` to control if `lsp-diagnostics-mode` should be enabled by auto configure for user to use native `flycheck` for syntax check.

PR #2584, opened 2021-02-05
lsp--action-trigger-suggest: Fix company-complete not fully working

kiennq:
The current implementation of `lsp--action-trigger-suggest` is not fully working for `company-mode`.
https://github.com/company-mode/company-mode/issues/1062

PR #2587, opened 2021-02-06
Use auto-begin fro company

jcs090218:
Response to #2584.

@dgutov @kiennq

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5163,11 +5163,11 @@ It will show up only if current point has signature help."
   "Handler for editor.action.triggerSuggest."
   (cond
    ((and company-mode
-         (fboundp 'company-complete)
+         (fboundp 'company-auto-begin)
          (fboundp 'company-post-command))
     (run-at-time nil nil
                  (lambda ()
-                   (company-complete)
+                   (company-auto-begin)
                    (company-post-command))))
    (t
     (completion-at-point))))


kiennq:
It doesn't seem to work though? @dgutov
I've tried with same repro step, as original bug, only switch to use `(run-at-time nil nil (lambda () (company-auto-begin)  (company-post-command)))`. However, the popup isn't showing up like the case of using `company-complete`

kiennq (changes requested):
It doe

dgutov:
Yes, sorry. `company-post-command` includes a check for `this-command`, so try this:

```el
(run-at-time 0 nil
             (lambda ()
               (company-auto-begin)
               (let ((this-command 'company-idle-begin))
                 (company-post-command))))
```

kiennq:
Thanks, the code does look more complicated.
Is there draw back of using `company-complete` here?

jcs090218:
Sorry I have accidentally remove my previous branch.

Move PR to #2593.

dgutov:
> Thanks, the code does look more complicated.

FWIW, the `this-command` value is arbitrary; it just shouldn't be nil.

> Is there draw back of using company-complete here?

The only reason `company-complete` works here is that it explicitly changes `this-command` as well (to `company-complete-common`). As for why it's undesirable: it calls the aforementioned `company-complete-common`. Which is likely to surprise the user, though it's only noticeable when all completions have a common prefix.

Really, it's a user-level, dwim-ish command, it's not designed for use in a Lisp program.

PR #2596, opened 2021-02-08
Added customizable diagnostics filter

### LINKED ISSUE #1520: How to turn off typescript lintings while preserving eslint lintings
After setting up typescript-language-server and vscode-eslint, typescript and eslint show the same errors.

image

yyoncho:
- Fixes #1520

- introduced `lsp-diagnostic-filter` which can be used to limit the reported
diagnostics.

Here it is sample filter to filter diagnostics from typescript

``` elisp

(lsp-defun my/filter-typescript ((params &as &PublishDiagnosticsParams :diagnostics)
                                 _workspace)
  (lsp:set-publish-diagnostics-params-diagnostics
   params
   (or (seq-filter (-lambda ((&Diagnostic :source?))
                     (not (string= "typescript" source?)))

                   diagnostics)
       []))
  params)

(setq lsp-diagnostic-filter 'my/filter-typescript )

```

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1903,6 +1903,16 @@ WORKSPACE is the workspace that contains the progress token."
 
 ;; diagnostics
 
+(defvar lsp-diagnostic-filter nil
+  "A a function which will be called with
+  `&PublishDiagnosticsParams' and `workspace' which can be used
+  to filter out the diagnostics. The function should return
+  `&PublishDiagnosticsParams'.
+
+Common usecase are:
+1. Filter the diagnostics for a particular language server.
+2. Filter out the diagnostics under specific level.")
+
 (defvar lsp-diagnostic-stats (ht))
 
 (defun lsp-diagnostics (&optional current-workspace?)
@@ -1954,8 +1964,7 @@ The result format is vector [_ errors warnings infos hints] or nil."
                                           (directory-file-name path)))))
       (lsp-diagnostics--update-path path new-stats))))
 
-(lsp-defun lsp--on-diagnostics (workspace (params &as
-                                                  &PublishDiagnosticsParams :uri :diagnostics))
+(defun lsp--on-diagnostics (workspace params)
   "Callback for textDocument/publishDiagnostics.
 interface PublishDiagnosticsParams {
     uri: string;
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
LGTM, do you know why CI fail?

vikigenius:
This works pretty well, although after additional reading and looking into the eslint plugins and such, I feel that, the right course of action to solve https://github.com/emacs-lsp/lsp-mode/issues/1520https://github.com/emacs-lsp/lsp-mode/issues/1520 is for the user to add rules to the eslint config to ignore errors that are already reported by the compiler.

Also, @yyoncho I want to know which language service are we actually using for typescript? I saw this one: https://github.com/theia-ide/typescript-language-server .

yyoncho:
> Also, @yyoncho I want to know which language service are we actually using for typescript? I saw this one: https://github.com/theia-ide/typescript-language-server .

This the recommended one. 

> I feel that, the right course of action to solve #1520 is for the user to add rules to the eslint config to ignore errors that are already reported by the compiler.

Let me think about that but I feel that it will be interfering with users config a bit too much.

vikigenius:
@yyoncho I agree with you, but maybe you are misunderstanding my comment. This PR Is a good option and should be merged.

But coming back specifically to typescript and eslint: If people want to use both the servers together, it is better to actually configure the eslint using `.eslintrc` to ignore some of the rules that produce duplicates.
Because disabling the whole of `ts-ls` diagnostics is not a very good solution because for the cases that it handles it will be better than eslint. `eslint` just handles additional linter stuff.

We should maybe document this in the react tutorial, so that people know how to use both of them together. I will probably work on a PR for that when I manage to configure it properly myself.

vikigenius:
Also, in the example filter that you provided how can I filter by language and level. For instance I only want to filter the info level diagnostics for typescript.

yyoncho:
Something like that should do:

``` emacs-lisp
(lsp-defun my/filter-typescript ((params &as &PublishDiagnosticsParams :diagnostics)
                                 _workspace)
  (lsp:set-publish-diagnostics-params-diagnostics
   params
   (or (seq-filter (-lambda ((&Diagnostic :source? :severity?))
                     (and (not (string= "typescript" source?))
                          (< severity? lsp/diagnostic-severity-information)))
                   diagnostics)
       []))
  params)

```

jacksonludwig:
Hi, @yyoncho , using your example config, I have the following:

```emacs
  (with-eval-after-load 'lsp-mode
    (lsp-defun my/filter-typescript ((params &as &PublishDiagnosticsParams :diagnostics)
                                     _workspace)
               (lsp:set-publish-diagnostics-params-diagnostics
                params
                (or (seq-filter (-lambda ((&Diagnostic :source? :severity?))
                                  (and (not (string= "typescript" source?))
                                       (< severity? lsp/diagnostic-severity-information)))
                                diagnostics)
                    []))
               params)

    (setq lsp-diagnostic-filter 'my/filter-typescript))
```
However, with that filter, I get the following error whenever I make a newline in a typescript file:
```
Error processing message (wrong-type-argument listp #s(hash-table size 6 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("message" "Delete `⏎··`" "severity" 1 "source" "eslint" "range" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("start" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("line" 16 "character" 2)) "end" #s(hash-table size 2 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("line" 17 "character" 2)))) "code" "prettier/prettier" "codeDescription" #s(hash-table size 1 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("href" "https://github.com/prettier/eslint-plugin-prettier#options"))))).
```
This does not occur without the filter. Should this be working? 

Thanks

yyoncho:
@jacksonludwig this kind of issue can happen when you have compiled this code with lsp-use-plists with different values. Can you provide error callstack?

PR #2601, opened 2021-02-10
Update clojure lsp server install

ericdallo:
* Update `lsp-clojure.el` to download native binaries compiled with GraalVM instead of legacy embedded jar executable.
* Update `lsp-rust.el` and `lsp-clojure.el` to prioritize lsp installed executables instead of system executables.
* Update `lsp-download-install` to keep the same path of store-path when unzipping.

### DIFF
--- a/clients/lsp-clojure.el
+++ b/clients/lsp-clojure.el
@@ -42,15 +42,23 @@
   :risky t
   :type '(repeat string))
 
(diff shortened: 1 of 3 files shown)


jcs090218 (approved):
LGTM

yyoncho (approved):

PR #2602, opened 2021-02-10
Bump ht.el version

leungbk:
This release forces an update on those users who may still be running
a rev of ht.el containing a buggy ht-get*.

There are tiny improvements to performance in this release as well.


ericdallo (approved):

yyoncho:
FYI this version does not have any effect when using melpa, because melpa version is generated from the date and it is always bigger...

Issue #2618, opened 2021-02-18
lsp-workspace-root: nested projects


muradm:
For mono repo with nested projects `lsp-workspace-root` selects the first matching root.
https://github.com/emacs-lsp/lsp-mode/blob/8166a1fe04891efcb7f1262609e77fc48500537b/lsp-mode.el#L4417

If project structure like below:
```
<root>/build.gradle
<root>/web-site/{package,angular}.json
```
`(lsp-session-folders (lsp-session))` returns list of two folders `'("<root>" "<root>/web-site")`. Then the above mentioned code seems to resolve to `<root>` instead of `<root>/web-site` for file `<root>/web-site/src/some.ts`.
This breaks clients which relay on `lsp-workspace-root`, for me it is `lsp-angular` which looks for `angular.json`.

Should not it make best match instead of first?

muradm:
This works better:
```
(defun lsp-workspace-root (&optional path)
  "Find the workspace root for the current file or PATH."
  (-when-let* ((file-name (or path (buffer-file-name)))
               (file-name (lsp-f-canonical file-name)))
    (->> (lsp-session)
         (lsp-session-folders)
         (--first (and (lsp--files-same-host it file-name)
                       (or (lsp-f-ancestor-of? it file-name)
                           (equal it file-name))))
         (--max-by (> (length it) (length other))))))
```

PR #2619, opened 2021-02-18
lsp-workspace-root: to return best match instead of first

muradm:
Better support for nested projects  #2618


yyoncho (approved):

yyoncho:
Looks good, we should remove duplication here and in lsp-find-session-folder

PR #2620, opened 2021-02-18
lsp-angular: can process typescript as well

muradm:
Angular lsp server can support inline HTML templates, but `activation-fn` prevents from `angular-ls` to be executed for Typescript files. Also extensions should match end of line, other wise "some-file.html.json" will match.


muradm:
This pattern does not work

PR #2630, opened 2021-02-20
Implement eslint/probeFailed in lsp-eslint

Sorixelle:
This PR implements the `eslint/probeFailed` request in the ESLint language server. From my investigations, this request seems to be fired on the following conditions:

- The ESLint configuration does not exist
- ESLint is not configured for the language in use (eg. `@typescript-eslint/parser` isn't set as the parser when using ESLint on TypeScript files)

The best way for us to handle this is probably just to display an error to the user, telling them to check their ESLint configurations.

### DIFF
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -295,6 +295,10 @@ to allow or deny it.")
              (lsp--persist lsp-eslint-library-choices-file lsp-eslint--stored-libraries)))
          (funcall callback (car option)))))))
 
+(defun lsp-eslint--probe-failed (_workspace _message)
+  "Called when the server detects a misconfiguration in ESLint."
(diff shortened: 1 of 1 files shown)


clients/lsp-eslint.el
```diff
@@ -295,6 +295,10 @@ to allow or deny it.")
              (lsp--persist lsp-eslint-library-choices-file lsp-eslint--stored-libraries)))
          (funcall callback (car option)))))))
 
+(defun lsp-eslint--probe-failed (_workspace _message)
+  "Called when the server detects a misconfiguration in ESLint."
+  (user-error "ESLint is not configured correctly. Please ensure your eslintrc is set up for the languages you are using"))
```
yyoncho, line 300: given the fact that this happens in async code, I think that lsp--error will be a better fit here.

yyoncho (approved):

PR #2631, opened 2021-02-20
Make paths for typescript-language-server configurable

Sorixelle:
This PR adds two new configuration options, `lsp-clients-typescript-tls-path` for setting an path to `typescript-language-server`, and `lsp-clients-typescript-tsserver-path` for setting a path to `tsserver`. The reasoning for adding these options is to better support [Yarn 2](https://yarnpkg.com/getting-started/editor-sdks) and [PnPify](https://yarnpkg.com/advanced/pnpify). Because of the way Yarn 2 and PnP work, Yarn needs to create wrapper scripts around `tsserver` and `typescript-language-server` for them to work correctly. With this PR, users can set the aformentioned configuration options to the path to Yarn's generated wrappers so they can use lsp-mode in Yarn 2 projects.

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -68,6 +68,18 @@
   :group 'lsp-mode
   :link '(url-link "https://github.com/theia-ide/typescript-language-server"))
 
+(defcustom lsp-clients-typescript-tls-path nil
+  "Path to the typescript-language-server binary."
+  :group 'lsp-typescript
+  :risky t
+  :type 'string)
+
+(defcustom lsp-clients-typescript-tsserver-path nil
+  "Path to the tsserver binary."
+  :group 'lsp-typescript
+  :risky t
+  :type 'string)
+
 (defcustom lsp-clients-typescript-server-args '("--stdio")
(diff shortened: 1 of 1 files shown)


clients/lsp-javascript.el
```diff
@@ -68,6 +68,18 @@
...
+  :group 'lsp-typescript
+  :risky t
+  :type 'string)
+
+(defcustom lsp-clients-typescript-tsserver-path nil
+  "Path to the tsserver binary."
+  :group 'lsp-typescript
+  :risky t
+  :type 'string)
+
```
yyoncho, line 82: can you put everything in one setting `lsp-clients-typescript-args`? I am planning to do restructoring in this code and this will make it easier for me. In particular, I  want to remove passing the tsserver as a param and let the typescript-language-server discover it.

Sorixelle: We already have a `lsp-clients-typescript-server-args` - is this what you were asking for? Or did you want to just keep `lsp-clients-typescript-tls-path` and get rid of the tsserver one?

yyoncho (approved):

PR #2632, opened 2021-02-20
lsp-clojure: remove bash -c from launch cmd

kiennq:
Since we switched to use native binary, `bash -c` is no longer needed. Removing it will allow `lsp-clojure` to work OTB in Windows.
Tested on WSL2 and Windows.


ericdallo (approved):
Thank you for that @kiennq !

PR #2635, opened 2021-02-20
Better defaults for clangd

### LINKED ISSUE #2503: `clangd` is adding incorrect headers
As discussed in https://github.com/syl20bnr/spacemacs/issues/12936, `clangd` tries to add headers automatically by default. The problem is that these normally are incorrect. This can be done by setting `lsp-clients-clangd-args` to `"--header-insertion=never"`. Do you think that this should be the default?

### LINKED ISSUE #2634: C/C++ Clangd completion for struct members adds a space
**Describe the bug**
In a C/C++ project, where `clangd` is used as the backend for `lsp-mode`, when using `company-complete`' to show the results, it adds a space character instead of showing all the possibilities for the completion.

An example, using a sample project of mine (https://github.com/RomeuG/SelfDwarfDumper):

I have this struct and function:

image

If in the function I write `a->`, the three different struct results appear without any problems:

image

If I move the cursor and then I get back to `a->` with the cursor right after `>` and use `company-complete` a space is added:

image

I assume that the results should appear instead of adding a space.

Another thing that should be noted, in the second image, is that the results window seems to add a space character in the beginning. I don't know if this is problematic or not.

**To Reproduce**

You can use a sample project of mine (https://github.com/RomeuG/SelfDwarfDumper) and Emacs should be started with `lsp-start-plain.el` using the following command: `emacs -q -l ~/.config/emacs/lsp-start-plain.el`.

More information on the bug description.

**Expected behavior**
When using `company-complete` after the `>` character to access the struct attributes, those same attributes should appear for selection.

I am using `lsp-clangd`.

**OS**

```
$ uname -a
Linux xerxes 5.10.16-zen1-1-zen #1 ZEN SMP PREEMPT Sat, 13 Feb 2021 20:51:02 +0000 x86_64 GNU/Linux

$ clangd --version
clangd version 11.0.1

$ emacs --version
GNU Emacs 28.0.50
Copyright (C) 2021 Free Software Foundation, Inc.
GNU Emacs comes with ABSOLUTELY NO WARRANTY.
You may redistribute copies of GNU Emacs
under the terms of the GNU General Public License.
For more information about these matters, see the file named COPYING.
```

Emacs version is from the `native-comp` branch.

yyoncho:
Seems like the defaults are causing confusion

Fixes #2634
Fixes #2503

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -173,7 +173,7 @@ find a suitable one. Set this variable before loading lsp."
   "Clang default executable full path when found.
 This must be set only once after loading the clang client.")
 
-(defcustom lsp-clients-clangd-args '()
+(defcustom lsp-clients-clangd-args '("--header-insertion-decorators=0")
   "Extra arguments for the clangd executable."
   :group 'lsp-clangd
   :risky t


ericdallo (approved):

jcs090218 (approved):

clients/lsp-clangd.el
```diff
@@ -173,7 +173,8 @@ find a suitable one. Set this variable before loading lsp."
   "Clang default executable full path when found.
 This must be set only once after loading the clang client.")
 
-(defcustom lsp-clients-clangd-args '()
+(defcustom lsp-clients-clangd-args '("--header-insertion=never"
```
kiennq, line 176: While `--header-insertion-decorators` is indeed not working well with Emacs, the first one `--header-insertion` is actually a feature of `clangd`.
It works fine for most case so I think we shouldn't include `--header-insertion=never` here yet

kiennq (approved):

clients/lsp-clangd.el
yyoncho: > It works fine for most case so I think we shouldn't include --header-insertion=never here yet

Good. I want to hear what others think about that as well and if there is no agreement I will revert that line and move it to FAQ section.

sebastiansturm: I personally prefer `--header-insertion=never` because I think clangd's include insertion fixit is just as convenient and I don't want my editor to sneak in potentially expensive includes just because I autocompleted the wrong candidate by accident. However, I haven't had any issues with that option either so both ways are decent defaults IMHO

sebastiansturm (approved):

clients/lsp-clangd.el
danielmartin: I prefer to have automatic header insertion by default. JetBrains's CLion also has this feature enabled by default, for instance.

IMO, we should have the same default as the clangd VSCode extension, as it's the canonical extension developed by Clangd's team.

yyoncho:
As per feedback, I kept only the "--header-insertion-decorators=0"

PR #2639, opened 2021-02-21
Improve signature help content

ericdallo:
Just make the content of signature help cleaner/less noisy

- Remove the `|` on the content, the green face already makes that a different section
- Remove the args count when length = 1, [just like vscode](https://github.com/BetterThanTomorrow/calva/blob/published/assets/howto/signature-help.gif)

signature-help-cleaner

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5019,11 +5019,13 @@ It will show up only if current point has signature help."
             (active-signature? (or lsp--signature-last-index active-signature? 0))
             (_ (setq lsp--signature-last-index active-signature?))
             ((signature &as &SignatureInformation? :label :parameters?) (seq-elt signatures active-signature?))
-            (prefix (concat (propertize (format " %s/%s"
-                                                (1+ active-signature?)
-                                                (length signatures))
-                                        'face 'success)
-                            " │ "))
+            (prefix (if (eq 1 (length signatures))
+                        ""
+                      (concat (propertize (format " %s/%s"
+                                                  (1+ active-signature?)
+                                                  (length signatures))
+                                          'face 'success)
+                              " ")))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

lsp-mode.el
```diff
@@ -5019,11 +5019,13 @@ It will show up only if current point has signature help."
             (active-signature? (or lsp--signature-last-index active-signature? 0))
             (_ (setq lsp--signature-last-index active-signature?))
             ((signature &as &SignatureInformation? :label :parameters?) (seq-elt signatures active-signature?))
-            (prefix (concat (propertize (format " %s/%s"
-                                                (1+ active-signature?)
-                                                (length signatures))
-                                        'face 'success)
-                            " │ "))
+            (prefix (if (eq 1 (length signatures))
```
yyoncho, line 5022: `(not (cl-rest signatures))` is the idiomatic way to check if the size is 1.

yyoncho (approved):

lsp-mode.el
ericdallo: Is that faster? It does not look easier to understand to me :/

yyoncho: `length` will iterate over the list.

kiennq: Yes, it's much faster, `O(1)` instead of `O(n)` like using `length`.
``` elisp
(length tmp)
91803

(benchmark-run 1000 (length tmp))
(0.665376 0 0.0)

(benchmark-run 1000 (cl-rest tmp))
(0.0002550000000000001 0 0.0)
```

ericdallo: Good to know, thanks

PR #2640, opened 2021-02-21
ido integration for lsp-mode

wandersoncferreira:
Hi,

Thanks for the great work on LSP mode, this looks very nice. I am currently updating some of my configs to use LSP for Clojure and js development.

I am kind of old and still use IDO as my main completion system and I find the `workspace-symbol` command very useful to me. Therefore, I ported the same ideas behind `lsp-ivy` and `heml-lsp` into IDO (with all its limitation and simplification). I am currently happy with the state of this patch and I am using it regularly.

Would like to know if this is a useful addition here. Thanks again for great work.


lsp-ido.el
```diff
@@ -0,0 +1,134 @@
+;;; lsp-ido.el --- `ido' integration -*- lexical-binding: t -*-
```
ericdallo, line 1: I think you should add this file here as well: https://github.com/emacs-lsp/lsp-mode/blob/master/Cask#L15

nbfalcon:
Shouldn't this be a separate page, like `lsp-ivy` and `helm-lsp`?

ericdallo:
You mean a separated package @nbfalcon ?

nbfalcon:
@ericdallo yes, I misspelled it.

ericdallo:
We could separate it, users would need to add it a new dependency, I'd like in the future to make lsp-mode manage all of that with a single `lsp-features ` variable or something like that. For now, I don't see any problem keep it here like `lsp-iedit.el` or move to another repo.
c/c @yyoncho

nbfalcon:
There are, as usual, arguments for and against that; on one hand, for consistency, it should be in a separate package. On the other hand, `ido-mode` is built-in.

ericdallo:
As a separated file and not requiring it as default, I don't see any issues keep it here

yyoncho:
I am for including it in the lsp-mode based on the argument that ido is a core component. 

What bothers me more, is that 

> (lsp-make-workspace-symbol-params :query "")

wont work with servers that requre at least one symbol. So, this component should perform as you type update of the results somehow. 

@wandersoncferreira any thoughts?

lsp-ido.el
wandersoncferreira: Added.

wandersoncferreira:
> I am for including it in the lsp-mode based on the argument that ido is a core component.
> 
> What bothers me more, is that
> 
> > (lsp-make-workspace-symbol-params :query "")
> 
> wont work with servers that requre at least one symbol. So, this component should perform as you type update of the results somehow.
> 
> @wandersoncferreira any thoughts?

I see what you mean @yyoncho . I was searching some way to provide this functionality, but IDO does not have any way to receive data asynchronously. The value as you type is stored in a variable called `ido-text` and this is modified here https://github.com/emacs-mirror/emacs/blob/24166be166ccda48353b395174da7e2bb1ca7659/lisp/ido.el#L1851. IDO's behavior seems more like an interface to help you search inside a pre-determined list of values.

I currently do not see a way to provide it, I can dig deeper if necessary. We can also intercept such error and show a message like "Feature not supported for server X" in the worst case.

yyoncho:
> I currently do not see a way to provide it, I can dig deeper if necessary.

I think that this can work - https://with-emacs.com/posts/tutorials/customize-completion-at-point/ . But in general, we may go and ask in emacs-devel how to implement that. FTR the first version of lsp-ivy and the current version of lsp-helm were hack - in elisp is relatively easy to implement async execution into sync component.

wandersoncferreira:
Update on the issue: Might be possible to do the dynamic update using `ido-completing-read-plus` https://github.com/DarwinAwardWinner/ido-completing-read-plus/blob/36a8bb93a59a16c8d5a5163c8cfde3770a1052df/ido-completing-read%2B.el#L186 but would not be fully `built in` anymore.

yyoncho:
> but would not be fully built in anymore.

Feel free to create a separate package(I can create one for you under emacs-lsp and make you the owner of that repo). I don't have a strong opinion on where to host code like that but apparently, other maintainers are more concerned about that.

wandersoncferreira:
> > but would not be fully built in anymore.
> 
> Feel free to create a separate package(I can create one for you under emacs-lsp and make you the owner of that repo). I don't have a strong opinion on where to host code like that but apparently, other maintainers are more concerned about that.

Sorry, I removed my comment by mistake. For completion I said:

Further search on the topic, https://github.com/DarwinAwardWinner/ido-completing-read-plus/blob/36a8bb93a59a16c8d5a5163c8cfde3770a1052df/ido-completing-read%2B.el#L186 might be useful to add this functionality. But in this case, the package would not be fully built in anymore.

@yyoncho would be ok for me to host it in another package.

nbfalcon:
@yyoncho looking at the code, `ido` doesn't support completion tables, so anything but querying the server for all candidates is impossible (the spec explicitly allows `""`). `icomplete` doesn't seem to have this deficiency, since `consult` supports it and it passes a completion-table to `completing-read`. I think this is the best that is possible with the `ido` API (`icomplete` might be the improvement).

Another possibility would be a `consult-lsp` package, which would give us support for `selectrum` and `icomplete`.

Also, what do you consider hacky about `lsp-ivy`'s async querying @yyoncho? (aside from the display-transformer-fn hack, removing which requires an upstream `ivy` PR of mine to be merged).

nbfalcon:
@wandersoncferreira  As for `ivy-completing-read+`, I don't think this will be an improvement over querying all symbols directly: like `ivy-read`, it simply calls `(all-completions "")` to get the values from a completion-table, see [here](https://github.com/DarwinAwardWinner/ido-completing-read-plus/blob/36a8bb93a59a16c8d5a5163c8cfde3770a1052df/ido-completing-read%2B.el#L894) and [here](https://github.com/DarwinAwardWinner/ido-completing-read-plus/blob/36a8bb93a59a16c8d5a5163c8cfde3770a1052df/ido-completing-read%2B.el#L619).

wandersoncferreira:
If the spec allows `:query ""` I would rather have this simple approach. I believe IDO users does not expect the async feature anyway as we tend to know IDO's limitations on strict lists. I would add the async feature if the spec does not allow "" for all servers.

nbfalcon:
@wandersoncferreira as far as I can see, doing this is async is impossible with `ido` without improving it, and then one might as well use `ivy`, or is `ido` somehow better (I haven't used it)? However, note that querying all workspace symbols might be very slow.

nbfalcon:
To [quote the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_symbol): "Clients may send an empty string here [in :query] to request all symbols".

wandersoncferreira:
> @wandersoncferreira as far as I can see, doing this is async is impossible with `ido` without improving it, and then one might as well use `ivy`, or is `ido` somehow better (I haven't used it)? However, note that querying all workspace symbols might be very slow.

`ido` is definitely not the brand new thing around. All these other packages have more modern approach to the same problem. I believe even emacs core is moving forward with `fido` nowdays. https://github.com/emacs-mirror/emacs/commit/213643a890913f10bac710ca8537e8b1125941d6

nbfalcon:
I understand. This situation is similar to `lsp-imenu` - it is great to have it for completeness, but we simply cannot squeeze every protocol feature out of it. The implementation now is probably the best *possible* one.

nbfalcon:
I understand. This situation is similar to `lsp-imenu` - it is great to have it for completeness, but we simply cannot squeeze every protocol feature out of it. The implementation now is probably the best *possible* one.

lsp-ido.el
```diff
@@ -0,0 +1,134 @@
...
+    (mapc (lambda (it)
+	    (lsp-ido--transform-candidate it lsp-ido--results workspace-root))
+	  raw-choices)
+    lsp-ido--results))
+
+;;;###autoload
+(defun lsp-ido-workspace-symbol ()
+  "`ido' for lsp workspace/symbol."
+  (interactive)
+  (let* ((hash-table-candidates (lsp-ido--workspace-symbol (lsp-workspaces)))
```
nbfalcon, line 127: There should probably be a function for only the current workspace, and this command should then become `lsp-ido-global-workspace-symbol`.

nbfalcon (commented):
LGTM.

lsp-ido.el
wandersoncferreira: Looks like `lsp-workspaces` is the command for only the current workspace if present, look [here](https://github.com/emacs-lsp/lsp-mode/blob/fcea313af2aa2a5da7c1f10cc7462f0104681547/lsp-mode.el#L1262) and a parallel on lsp-ivy [here](https://github.com/emacs-lsp/lsp-ivy/blob/c70ee8b54357c56d1b972393ee53e57a2e545fbb/lsp-ivy.el#L209)

wandersoncferreira: And to be fair, I tested the `global-workspacee-symbol` option some times on `ivy` and didn't find this so useful. Because of the sync nature of `ido` I got several timeouts with projects no so big. I would rather stay with the `current-workspace`option for now.

yyoncho:
> To [quote the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_symbol): "Clients may send an empty string here [in :query] to request all symbols".

I tested with jdtls and it does not respond to that query which I believe is expected because a query like that will take minutes on a big project with a lot of dependencies.


@wandersoncferreira what about the following flow:

`M-x lsp-ido-...` - reads the query string with read-string and displays the results
`C-u M-x lsp-ido-..` - works as in the PR
`M-x lsp-ido-.. RET` - works as in the PR

This will guarantee that we work correctly.

yyoncho:
> Also, what do you consider hacky about `lsp-ivy`'s async querying @yyoncho? (aside from the display-transformer-fn hack, removing which requires an upstream `ivy` PR of mine to be merged).

I am talking about the initial implementation which was using ivy's internal functions to simulate async.

wandersoncferreira:
> > To [quote the spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_symbol): "Clients may send an empty string here [in :query] to request all symbols".
> 
> I tested with jdtls and it does not respond to that query which I believe is expected because a query like that will take minutes on a big project with a lot of dependencies.
> 
> @wandersoncferreira what about the following flow:
> 
> `M-x lsp-ido-...` - reads the query string with read-string and displays the results
> `C-u M-x lsp-ido-..` - works as in the PR
> `M-x lsp-ido-.. RET` - works as in the PR
> 
> This will guarantee that we work correctly.

This sounds reasonable.

wandersoncferreira:
@yyoncho I implemented the feature using query from `read-string` however there is something off. I noticed all the symbols continued to be returned by the function no matter the value I passed as `:query`. To test this, I even hardcoded a non-sense value instead of current `""`.  Looking at the `lsp-ivy` integration, seems like the filtering of the candidates is performed after the results arrived [here](https://github.com/emacs-lsp/lsp-ivy/blob/c70ee8b54357c56d1b972393ee53e57a2e545fbb/lsp-ivy.el#L167), and [here](https://github.com/emacs-lsp/lsp-ivy/blob/c70ee8b54357c56d1b972393ee53e57a2e545fbb/lsp-ivy.el#L188);

yyoncho:
> Looking at the `lsp-ivy` integration, seems like the filtering of the candidates is performed after the results arrived [here](https://github.com/emacs-lsp/lsp-ivy/blob/c70ee8b54357c56d1b972393ee53e57a2e545fbb/lsp-ivy.el#L167), and [here](https://github.com/emacs-lsp/lsp-ivy/blob/c70ee8b54357c56d1b972393ee53e57a2e545fbb/lsp-ivy.el#L188);

AFAIK lsp-ivy performs double filtering if you do "foo bar" only foo will be sent to the server, bar will be filtered client side. Server returning all results might be something specific to the server you are testing with. Push your changes I will test it with more servers.

wandersoncferreira:
Thanks for clarification @yyoncho . PR updated

lsp-ido.el
```diff
@@ -0,0 +1,139 @@
...
+
+;;;###autoload
+(defun lsp-ido-workspace-symbol (arg)
+  "`ido' for lsp workspace/symbol.
+When called with prefix ARG the default selection will be symbol at point."
+  (interactive "P")
+  (let* ((query (if arg "" (read-string "Workspace symbol: ")))
+         (hash-table-candidates (lsp-ido--workspace-symbol (lsp-workspaces) query))
+	 (choice (ido-completing-read
+	          "Workspace symbol:"
```
yyoncho, line 131: it looks better to me with space after ":" but leaving it up to you to decide whether to apply that comment.

yyoncho (approved):
Looks good to me!

Can you apply automatic formatting to the buffer? Some of the blocks are not properly formatted. 

After that, it is good to go. You may link your doc page from the lsp-mode README.md similar to how we link dired and the other integrations.

lsp-ido.el
wandersoncferreira: Formatted, linked in README, and I added the space :)

wandersoncferreira:
Thanks everybody for all the feedbacks!

ericdallo (approved):
LGTM, thank you for the contribution!

PR #2643, opened 2021-02-23
Improve lsp-define-conditional-key to include short description for whichkey

kiennq:
Include short description for whichkey when define lsp's key bindings.
Also include several missing whichkey descriptions (for example, "Tb", "TD", "hg"), mainly because they're separated into two different tables.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2218,27 +2218,34 @@ WORKSPACE is the workspace that contains the diagnostics."
 
 
 ;; keybindings
+(defvar lsp--binding-descriptions nil
+  "List of key binding/short description pair.")
 
-(defmacro lsp-define-conditional-key (keymap key def cond &rest bindings)
+(defmacro lsp-define-conditional-key (keymap key def desc cond &rest bindings)
   "In KEYMAP, define key sequence KEY as DEF conditionally.
 This is like `define-key', except the definition disappears
 whenever COND evaluates to nil.
-BINDINGS is a list of (key def cond)."
+DESC is the short-description for the binding.
+BINDINGS is a list of (key def desc cond)."
   (declare (indent defun)
-           (debug (form form form form &rest sexp)))
-  (->> (cl-list* key def cond bindings)
-       (-partition 3)
-       (-map (-lambda ((key def cond))
-               `(define-key ,keymap ,key
-                  '(menu-item
-                    ,(format "maybe-%s" def)
-                    ,def
-                    :filter (lambda (item)
-                              (when (with-current-buffer (or (when (buffer-live-p lsp--describe-buffer)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -2218,27 +2218,34 @@ WORKSPACE is the workspace that contains the diagnostics."
...
+                        :filter
+                        (lambda (item)
+                          (when (with-current-buffer (or (when (buffer-live-p lsp--describe-buffer)
+                                                           lsp--describe-buffer)
+                                                         (current-buffer))
+                                  ,cond)
+                            item))))
+                    (when (stringp ,key)
+                      (setq lsp--binding-descriptions
+                            (nconc lsp--binding-descriptions '(,key ,desc)))))))
```
yyoncho, line 2248: nit: using cl-list* will look a bit more natural.

yyoncho (approved):
Looks good, few minor nits:

lsp-mode.el
```diff
@@ -2261,70 +2268,73 @@ BINDINGS is a list of (key def cond)."
...
-      "Th" lsp-toggle-symbol-highlight (lsp-feature? "textDocument/documentHighlight")
-      "Tb" lsp-headerline-breadcrumb-mode (lsp-feature? "textDocument/documentSymbol")
-      "Ta" lsp-modeline-code-actions-mode (lsp-feature? "textDocument/codeAction")
-      "TD" lsp-modeline-diagnostics-mode (lsp-feature? "textDocument/publishDiagnostics")
-      "TS" lsp-ui-sideline-mode (featurep 'lsp-ui-sideline)
-      "Td" lsp-ui-doc-mode (featurep 'lsp-ui-doc)
-      "Ts" lsp-toggle-signature-auto-activate (lsp-feature? "textDocument/signatureHelp")
-      "Tf" lsp-toggle-on-type-formatting (lsp-feature? "textDocument/onTypeFormatting")
-      "TT" lsp-treemacs-sync-mode (featurep 'lsp-treemacs)
+      "TD" lsp-modeline-diagnostics-mode "toggle diagnostics" (lsp-feature? "textDocument/publishDiagnostics")
```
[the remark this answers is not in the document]
kiennq: Sure

lsp-mode.el
kiennq: They are not equivalent since `lsp--binding-descriptions` is list already. What I want here is `(append A B)` instead of `(cons A B)` like in `cl-list*`.

kiennq: Oh, it seems I misunderstand your point perhaps this comment is for the `lsp-enable-which-key-integration`? The `nconc` there is not correct, indeed.

PR #2647, opened 2021-02-24
`lsp-signature-next`/prev: support cycling

nbfalcon:
If `lsp-signature-cycle` is enabled (default), `lsp-signature-next` and
`lsp-signature-previous` will cycle if called with the last/first item
selected.

Tested with `pyright` (Enabled/disabled).

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4959,22 +4959,30 @@ It will show up only if current point has signature help."
   (add-hook 'post-command-hook #'lsp-signature)
   (lsp-signature-mode t))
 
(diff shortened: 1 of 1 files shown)


yyoncho (approved):
Looks good - to me it seems like this should be the default with no option to turn it off. Leaving it up to you to decide whether to apply the comment.

nbfalcon:
@yyoncho I believe it would be better to have this customizable, as I think some users might prefer the non-cycling behaviour. Having cycling on by default is IMHO more intuitive, though, so that default is sensible.

PR #2648, opened 2021-02-24
Trigger suggests: ensure that `company-mode` bound

nbfalcon:
Also remove the `defvar` for it, since it isn't needed anywhere else.

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5145,8 +5145,9 @@ It will show up only if current point has signature help."
          (fboundp 'company-post-command))
     (run-at-time 0 nil
                  (lambda ()
-                   (company-auto-begin)
-                   (let ((this-command 'company-idle-begin))
+                   (let ((this-command 'company-idle-begin)
+                         (company-minimum-prefix-length 0))
+                     (company-auto-begin)
                      (company-post-command)))))
    (t
     (completion-at-point))))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -71,6 +71,7 @@
 (defvar yas-also-auto-indent-first-line)
 (defvar dap-auto-configure-mode)
 (defvar dap-ui-menu-items)
+(defvar company-minimum-prefix-length)
 
 (defconst lsp--message-type-face
   `((1 . ,compilation-error-face)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5137,12 +5137,10 @@ It will show up only if current point has signature help."
   "Handler for editor.action.triggerParameterHints."
   (lsp-signature-activate))
 
-(defvar company-mode)
-
 (defun lsp--action-trigger-suggest (_command)
   "Handler for editor.action.triggerSuggest."
   (cond
-   ((and company-mode
+   ((and (bound-and-true-p company-mode)
          (fboundp 'company-auto-begin)
          (fboundp 'company-post-command))
     (run-at-time 0 nil


yyoncho (approved):

nbfalcon:
@jcs090218 why use `company-auto-begin` instead? With it, trigger suggests doesn't work for me when completing CSS rules in `html-ls`. With the previous version, I had to patch company though (which might be related to my set-up) (~ line 997):
```emacs-lisp
        (apply #'company--multi-backend-adapter company-backend args)
```
```emacs-lisp
        (apply #'company--multi-backend-adapter company-backends args)
```

kiennq:
@nbfalcon It's from `company-mode` author suggestion to use `company-auto-begin` instead of `company-complete`.
I confirm it's indeed not working with `company-auto-begin`, change to `company-manual-begin` make it works though.
@dgutov

dgutov:
Whether to use `company-auto-begin` or `company-manual-begin`, depends on whether you want to indicate user's intent in initiating completion.

How does `company-auto-begin` not work? Could you show it in a self-contained example?

kiennq:
> How does `company-auto-begin` not work? Could you show it in a self-contained example?

Tried with `runemacs -Q` and manually enable `company-mode` there.
- In `*scratch*` buffer, type `(r<|>` with `<|>` as cursor position.
- Run `M-: (run-at-time 0 nil (lambda () (company-auto-begin) (let ((this-command 'company-idle-begin)) (company-post-command))))` --> No candidate window is shown
- Run `M-: (run-at-time 0 nil (lambda () (company-manual-begin) (let ((this-command 'company-idle-begin)) (company-post-command))))` --> Candidate popup is shown

nbfalcon:
@dgutov what do you mean by intent? In this case, the user isn't triggering the completion, but the LSP server, so it seems like `company-auto-begin` is more appropriate. However, it doesn't cause a completion window to pop up.

dgutov:
@kiennq Try this one, please:

```el
(run-at-time 0 nil
             (lambda ()
               (let ((this-command 'company-idle-begin)
                     (company-minimum-prefix-length 0))
                 (company-auto-begin)
                 (company-post-command))))
```

Yes, it's another extra binding (:slightly_frowning_face:), but the check for it is deeper down. And `company-manual-begin` overrides it because the assumption is if completion is invoked manually, the prefix length shouldn't matter.

dgutov:
@nbfalcon

> In this case, the user isn't triggering the completion, but the LSP server, so it seems like company-auto-begin is more appropriate

That's what I mean, yes. And it's what justifies the gymnastics we're currently discussing.

kiennq:
@dgutov Yes, it works now.
Is the different between manual and auto completion triggering is only in checking prefix length?

In this case, although it's not user intent to trigger completion but it's server intent to trigger completion. Both of them are still out side of `company-mode` so I would consider it manual completion triggering, not automatically by `company-mode`.
So using `company-manual-begin` seems to be more approriated?

@nbfalcon Can you help apply @dgutov's suggestion to fix this auto triggering feature?

nbfalcon:
@dgutov the new version works great and I'll use it in this PR. Why is the idle timer necessary though?

kiennq:
@nbfalcon The idle timer is needed to break from local binding when this function is called (most likely from `company`'s `post-completion`).
It's not really an idle timer but more like context switcher here.
Well, do you remember your patch of `(apply #'company--multi-backend-adapter company-backends args)`?
The force context switch here fixes that.

dgutov:
> Is the different between manual and auto completion triggering is only in checking prefix length?

As you can see for yourself from the definitions, the former also sets `company--manual-action` to `t`, which will affect a number of behaviors.

> the new version works great and if it is the "officially" right way, I'm going to use it here.

It's official-ish, given that you're using it in a context which I don't know of and thus couldn't anticipate.

nbfalcon:
@kiennq done.

kiennq (approved):

kiennq:
@nbfalcon The test is failing, you need to wrap `company-minimum-prefix-length` in `(ignore company-minimum-prefix-length)` or something.
(nit: Using `company-manual-begin` might be better, my arguments is here https://github.com/emacs-lsp/lsp-mode/pull/2648#issuecomment-785553864)

dgutov:
> you need to wrap company-minimum-prefix-length in (ignore company-minimum-prefix-length) or something

You need `(defvar company-minimum-prefix-length)` somewhere at the top of function's definition. Not an `ignore` call.

> Using company-manual-begin might be better, my arguments is here

Have you read mine?

kiennq:
> You need `(defvar company-minimum-prefix-length)` somewhere at the top of function's definition. Not an `ignore` call.

Since this is a `let-bind` the `defvar` is not need IMO. I'm not expert in this so anyway is fine to me though.



> > Using company-manual-begin might be better, my arguments is here
> 
> Have you read mine?

I've read it. I mean the `manual` and `auto` here, is that referring strictly to user's intent?
In another word, if another package invoke `company-mode`, should it always use `company-auto-begin` and has to manually adjust/bind the prefix length?
In this case, the server is asking `lsp-mode` to show the completion popup, thus `lsp-mode` has to manually trigger the popup. I feel that should be manual action but it might be wrong?

dgutov:
> Since this is a let-bind the defvar is not need IMO

Since `company` is not require-d, without `defvar` the byte-compiler won't know it's a special variable and will treat it as lexical (and unused).

> I mean the manual and auto here, is that referring strictly to user's intent?

More or less.

> if another package invoke company-mode, should it always use company-auto-begin

Depends on what that package is doing. If it creates a new command, with a key binding (which the user will invoke by choice), then the `manual` version is more appropriate.

>  and has to manually adjust/bind the prefix length

Unless a special function is added to company for that purpose that will do the adjusting for you. I'm not sure what to call it, though.

> In this case, the server is asking lsp-mode to show the completion popup, thus lsp-mode has to manually trigger the popup. I feel that should be manual action but it might be wrong?

In what circumstances?

Is that essentially the same situation as user typing `.` with `company-clang` when `company-clang-begin-after-member-access` is non-nil?

kiennq:
> In what circumstances?
> 
> Is that essentially the same situation as user typing `.` with `company-clang` when `company-clang-begin-after-member-access` is non-nil?

An example scenario is when user is selecting an item from completion pop up. That will trigger `post-completion` for that item, which in turn will trigger another action like showing the completion popup for the next item.
For example, in json config file, type `"lau`, select `"launchMode"` from completion popup will insert `"launchMode": ` and show the completion pop up for possible launch modes.

nbfalcon:
Done, lsp-mode byte-compiles cleanly now.

As for the discussion, I am unsure myself. Grepping trough the code of company, `company-manual-begin` causes certain messages, like "sole completion" and "no completions". @kiennq do we want that? I myself am more leaning on the side of auto-begin being more appropriate because of that, but I'd like to hear some other opinions.

@dgutov another example would be auto-completing css rules: say I am entering a css-rule like this:
```css
button:hover i {
  col| /* cursor on |, trigger auto-completion */
}
```
```css
button hover i {
  color: | /* choose `color` from the candidates; now the server will ask the client to invoke auto-completion again, so that the user can choose a colour. Here the completion candidates will be a valid list of colors at this point, but it could be other things, like pixel sizes etc. for other selectors. */
}
```

dgutov:
> Grepping trough the code of company, company-manual-begin causes certain messages, like "sole completion" and "no completions"

These two are unlikely to come up (the situations they correspond to should be avoided by the server which does the "trigger"), but the `non-essential` binding, as well as the `company-require-match` behavior, in its default setting, are going to be enabled by the `manual-` version.

I guess the key distinction is whether the user anticipates and always wants the completion after the trigger to happen. Or whether they are likely enough to go on typing, rather than want to choose a particular completion from the list. Going by your examples, maybe they do, or maybe it's still 50/50, in which case the "auto" version is more appropriate.

I guess if you're targeting users migrating from or inspired by VS Code, one could watch the promo videos and/or (better) watch some of the users at work. Like, if the odds of them picking a completion in the end are >90%, then the "manual" version can be okay.

yyoncho:
> I guess if you're targeting users migrating from or inspired by VS Code, one could watch the promo videos and/or (better) watch some of the users at work.

IMHO here it is not about being vscode or not but about smarter completion based on the context. It is similar to trigger characters in the vanilla company-mode but now with the help of the server, it might be more context-sensitive. E. g. if you autocomplete `foo = |` the autocompletion will pop-up based on whether the `foo` is int or enum.

kiennq:
> These two are unlikely to come up (the situations they correspond to should be avoided by the server which does the "trigger"), but the non-essential binding, as well as the company-require-match behavior, in its default setting, are going to be enabled by the manual- version.

Oh, I miss this.
For `lsp-mode`, `non-essential` is basically blocking so the `auto` version is better now.

jcs090218 (approved):

dgutov:
> IMHO here it is not about being vscode or not but about smarter completion based on the context. It is similar to trigger characters in the vanilla company-mode but now with the help of the server, it might be more context-sensitive. E. g. if you autocomplete foo = | the autocompletion will pop-up based on whether the foo is int or enum.

Do you mean that it will pop up in the case of enum but not if foo is int?

Anyway, sounds like "auto" is indeed the way to go because it seems similar to quick completion after one types `.`, for instance.

dgutov:
> For lsp-mode, non-essential is basically blocking so the auto version is better now

Which would make sense if expect the user to choose one of the options rather than continue typing. But apparently it's not always going to be the case.

yyoncho:
> Do you mean that it will pop up in the case of enum but not if foo is int?

Yes.

PR #2650, opened 2021-02-25
Handle after-set-visited-file-name-hook

### LINKED ISSUE #2644: Lsp-format-buffer after copy file to new file replaces file content with original file.
**Describe the bug**
In Emacs Editor
Lsp-format-buffer after copy file to new file(and edit) replaces the file content with original file.

**To Reproduce**
For example.
```
File1.scala
Class File1 {something}
File2.scala(copy of File1.scala)
Class File2{something more}
```

Now if one calls `lsp-format-buffer` on File2.scala it would replace the content of File2.scala with File1.scala and format it.

**Expected behavior**
File2 's content should not change with File1.scala

lsp-scala with metals.

**OS**
Macos

**Work Around**
close emacs and reopen it and now do lsp-format-buffer which now works.

yyoncho:
- fixes #2644

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3561,6 +3561,10 @@ yet."
                    (memq :on-trigger-char lsp-signature-auto-activate))))
       (remove-hook 'post-self-insert-hook signature-help-handler t)))))
 
+(defun lsp--after-set-visited-file-name ()
+  (lsp-disconnect)
+  (lsp))
+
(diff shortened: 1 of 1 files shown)


ericdallo (approved):

jcs090218 (approved):

jcs090218:
LGTM.

PR #2651, opened 2021-02-25
[lsp-lua] Support auto downloading and installing "lua-language-server"

sunlin7:
Support downloading the "lua-language-server" from VSCode extension site.

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -94,14 +94,22 @@
   :risky t
   :type 'directory)
 
-(defcustom lsp-clients-lua-language-server-bin (f-join lsp-clients-lua-language-server-install-dir "bin/Linux/lua-language-server")
+(defcustom lsp-clients-lua-language-server-bin
+  (f-join lsp-clients-lua-language-server-install-dir
+          "extension/server/bin/"
+          (pcase system-type
+            ('gnu/linux "Linux/lua-language-server")
+            ('darwin "macOS/lua-language-server")
+            ('windows-nt "Windows/lua-language-server.exe")))
   "Location of Lua Language Server."
   :group 'lsp-lua-language-server
   :version "7.1"
   :risky t
   :type 'file)
 
-(defcustom lsp-clients-lua-language-server-main-location (f-join lsp-clients-lua-language-server-install-dir "main.lua")
+(defcustom lsp-clients-lua-language-server-main-location
+  (f-join lsp-clients-lua-language-server-install-dir
+          "extension/server/main.lua")
   "Location of Lua Language Server main.lua."
   :group 'lsp-lua-language-server
   :version "7.1"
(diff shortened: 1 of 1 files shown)


sunlin7:
It's inspired from https://github.com/emacs-lsp/lsp-mode/blob/63c80e4feb37b9807334a157084f368b6e119c30/clients/lsp-eslint.el#L349, download the "lua-language-server" from VSCode extension site.

clients/lsp-lua.el
```diff
@@ -370,6 +378,24 @@ and `../lib` ,exclude `../lib/temp`.
    ("Lua.completion.callSnippet" lsp-lua-completion-call-snippet)
    ("Lua.color.mode" lsp-lua-color-mode)))
 
+(defun lsp-lua-language-server-install (client callback error-callback update?)
+  "Download the latest version of lua-language-server and extract it to
+`lsp-lua-language-server-install-dir'."
+  (ignore client update?)
+  (let ((tmp-zip (make-temp-file "ext" nil ".zip")))
```
kiennq, line 385: I wonder if we can automate these step? I.e., if `:store-path` is empty, automatically generate temp file name with pattern of `extXYZ.zip` since the vscode code extension file is just zip file in disguise.
That can probably new helper function `lsp-vscode-download-install`

sunlin7: Yes, if there is a helper function `lsp-vscode-download-install` will be very useful, and more, a helper function `'lsp-vscode-update-...` maybe also helpful, then it will be easy to download/install/update the vscode extensions.

sunlin7: Current pull request won't implement these helper functions, so I resolve this comment. 
Thanks.

nbfalcon: We have this support already. `lsp-download-server` `:decompess :zip`

clients/lsp-lua.el
```diff
@@ -370,6 +378,24 @@ and `../lib` ,exclude `../lib/temp`.
...
+    (delete-file tmp-zip)
+    (lsp-download-install
```
nbfalcon, line 387: Something like
```emacs-lisp
(lsp-download-install (lambda (&rest _) (set-file-mode lsp-clients-lua-language-server-bin #o0700) (funcall callback)) error-callback :url (lsp-vscode-extension-url "sumneko" "lua" "1.17.0" :store-path lsp-clients-lua-language-server-install-dir :decompress :zip)

yyoncho: If you don't mind I propose merging as it is and later addressing both lsp-eslint.el and lsp-lua.el.

sunlin7: Totally 👍

sunlin7: @nbfalcon The bin for lua-language-sever is in the subtree of the folder structure, while `lsp-download-install` is not the best choice.

nbfalcon: @sunlin7 I don't quite understand - `lsp-download-install` is downloading a zip, which is unzipped to a directory, so this shouldn't be a problem. In any case, I agree with @yyoncho that this can be addressed later.

yyoncho (approved):
I tested it manually and it works well.

sunlin7:
It will make all the files in the folder executable, maybe has secure issue.

sunlin7:
@nbfalcon Yes, your code snippet is more simple and elegant, I modify the code to follow your comments. 
Please review again. Thanks.

kiennq (approved):

PR #2654, opened 2021-02-28
Fix merging hover results

yyoncho:
- this is causing issues in lsp-java when combined with spring-boot

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1197,17 +1197,15 @@ the lists according to METHOD."
                                      (-compose #'not #'lsp-empty?)
                                      results)
                                (`(,hover) hover)
-                               (hovers (-reduce-from
-                                        (-lambda ((&Hover :contents) result)
-                                          (lsp:set-hover-contents
-                                           result (append
-                                                   (if (and (sequencep contents)
-                                                            (not (stringp contents)))
-                                                       contents
-                                                     (list contents))
-                                                   (lsp:hover-contents result))))
-                                        (lsp-make-hover)
-                                        hovers))))
+                               (hovers (lsp-make-hover
+                                        :contents
+                                        (-mapcat
+                                         (-lambda ((&Hover :contents))
+                                           (if (and (sequencep contents)
+                                                    (not (stringp contents)))
+                                               (append contents ())
+                                             (list contents)))
+                                         hovers)))))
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -1197,17 +1197,15 @@ the lists according to METHOD."
...
-                                                   (lsp:hover-contents result))))
-                                        (lsp-make-hover)
-                                        hovers))))
+                               (hovers (lsp-make-hover
+                                        :contents
+                                        (apply #'append (-map
+                                                         (-lambda ((&Hover :contents))
+                                                           (if (and (sequencep contents)
+                                                                    (not (stringp contents)))
+                                                               (append contents ())
```
nbfalcon, line 1206: Why would you need to clone `contents` here? `append` is non-destructive.

lsp-mode.el
```diff
@@ -1197,17 +1197,15 @@ the lists according to METHOD."
...
+                                                             (list contents)))
+                                                         hovers))))))
```
nbfalcon, line 1208: Wouldn't hovers always be nil here, causing hovers to always be empty in this code path? This also seems present in the old code, but I don't see it.

lsp-mode.el
```diff
@@ -1197,17 +1197,15 @@ the lists according to METHOD."
...
-                                                   (if (and (sequencep contents)
-                                                            (not (stringp contents)))
-                                                       contents
-                                                     (list contents))
-                                                   (lsp:hover-contents result))))
-                                        (lsp-make-hover)
-                                        hovers))))
+                               (hovers (lsp-make-hover
+                                        :contents
+                                        (apply #'append (-map
```
nbfalcon, line 1202: `-mapcat`

nbfalcon (commented):
Aside from a minor nitpick (`-mapcat`) and some confusion on why this works on my side, LGTM.

yyoncho:
Fixed comments.

nbfalcon (approved):

lsp-mode.el
nbfalcon: Never mind, I forgot how `pcase` works for a second...

ericdallo (approved):

jcs090218 (approved):

PR #2658, opened 2021-02-28
Don't use `cursor-sensor-mode`

nbfalcon:
`cursor-sensor-functions` isn't used anywhere, so don't enable it in
`lsp-describe-session`.

----

#


yyoncho (approved):

yyoncho:
I don't know where this came from.

Issue #2298, opened 2020-11-03
Implement rust-analyzer code lenses


nbfalcon:
- [ ] rust-analyzer.debugSingle: rust-analyzer's debugSingle works exactly like runSingle, the only difference being the "command". To implement this, we need a way to wrap cargo run commands in gdbserver (I'll make a PR), and a dap-mode command to attach to some PID. We can reuse the runSingle code and just add the wrapper-gdbserver argument.

- [x] rust-analyzer.showReferences

yyoncho:
AFAIK vscode provides the option to select which debug adapter to use for debugging.

nbfalcon:
Do we need to support that too, though? rust-gdb is the standard, AFAIK, so only supporting that should be enough for now.

yyoncho:
which adapter is that?

nbfalcon:
Its just GDB, so dap-gdb-lldb.

yyoncho:
dap-gdb-lldb is very limited regarding functionality(e. g. no threading support, poor locals handling, etc). cpptools or code-lldb are the best options for native debugging.  We do not have support for the latter. Not sure how lldb-vscode scores against them though.

nbfalcon:
Perhaps this might need to be made configurable after all. I think I'll do showReferences first.

yyoncho:
FTR - https://github.com/rust-analyzer/rust-analyzer/blob/master/editors/code/package.json#L514

Issue #1463, opened 2020-02-28
Lenses does not work in rust analyzer 


yyoncho:
There are at least 3 types of lenses in rust analyzer:

1. Implementation
2. Run test
3. Run app.

yyoncho:
cc @brotzeit @flodiebold

jacg:
As a non-expert, is there anything I could do to help with this?

yyoncho:
@adam-r-kowalski do you have something for vscode-cpp debugger? 

@jacg it is relatively easy to integrate those, we have such integration for scala metals. But IMO the integration should be with vscode-cpp which is yet to be implemented in dap-mode.

PR #2662, opened 2021-03-02
[rust-analyzer] support for debug

yyoncho:
- Fixes https://github.com/emacs-lsp/lsp-mode/issues/1463
- Fixes https://github.com/emacs-lsp/lsp-mode/issues/2298

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -640,6 +640,11 @@ them with `crate` or the crate name they refer to."
   (lsp-show-xrefs (lsp--locations-to-xref-items references) nil
                   (s-contains-p "reference" title)))
 
+(declare-function dap-debug "ext:dap-mode" (template) t)
+
+(lsp-defun lsp-rust--analyzer-debug-lens ((&Command :arguments? [args]))
+  (lsp-rust-analyzer-debug args))
+
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-stdio-connection
@@ -654,6 +659,7 @@ them with `crate` or the crate name they refer to."
   :initialization-options 'lsp-rust-analyzer--make-init-options
   :notification-handlers (ht<-alist lsp-rust-notification-handlers)
   :action-handlers (ht ("rust-analyzer.runSingle" #'lsp-rust--analyzer-run-single)
+                       ("rust-analyzer.debugSingle" #'lsp-rust--analyzer-debug-lens)
                        ("rust-analyzer.showReferences" #'lsp-rust--analyzer-show-references))
   :library-folders-fn (lambda (_workspace) lsp-rust-library-directories)
   :after-open-fn (lambda ()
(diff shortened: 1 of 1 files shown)


clients/lsp-rust.el
```diff
@@ -860,6 +866,45 @@ them with `crate` or the crate name they refer to."
        (lambda (_) (concat "*" label "*")))
       (setq lsp-rust-analyzer--last-runnable runnable))))
 
+(defun lsp-rust-analyzer-debug (runnable)
+  "Select and run a runnable action."
+  (interactive (list (lsp-rust-analyzer--select-runnable)))
+  (unless (featurep 'dap-cpptools)
```
nbfalcon, line 872: Wouldn't it be better to `require` `dap-cpptools` here instead of just throwing an error if the user hasn't done so already? As a bonus, `require` throws an error if it isn't installed itself, but this can be wrapped in an intuitive error-message.

clients/lsp-rust.el
```diff
@@ -860,6 +866,45 @@ them with `crate` or the crate name they refer to."
...
+    (user-error "You must require `dap-cpptools'."))
+  (-let (((&rust-analyzer:Runnable
+           :args (&rust-analyzer:RunnableArgs :cargo-args :workspace-root? :executable-args)
+           :label) runnable))
+    (cl-case (aref cargo-args 0)
```
nbfalcon, line 877: Shouldn't this be `seq-first`?

clients/lsp-rust.el
```diff
@@ -860,6 +866,45 @@ them with `crate` or the crate name they refer to."
...
+      ("test" (when (-contains? (append cargo-args ()) "--no-run")
+                (cl-callf append cargo-args (list "--no-run")))))
+    (->> (append (list (executable-find "cargo"))
+                 cargo-args
+                 (list "--message-format=json"))
+         (s-join " ")
+         (shell-command-to-string)
+         (s-lines)
+         (-keep (lambda (s)
+                  (condition-case _
```
nbfalcon, line 888: Nit: this could be `(ignore-error error ...)` or at least `condition-case nil`, since _ still captures the error, while `nil` doesn't.

clients/lsp-rust.el
```diff
@@ -860,6 +866,45 @@ them with `crate` or the crate name they refer to."
...
+                      (-let* ((json-object-type 'plist)
+                              ((msg &as &plist :reason :executable) (json-read-from-string s)))
+                        (when (and executable (string= "compiler-artifact" reason))
+                          executable))
+                    (error))))
+         (funcall (lambda (artifact-spec)
+                    (pcase artifact-spec
+                      (`() (user-error "No compilation artifacts or reading the "))
```
nbfalcon, line 896: Nit: this sentence is unfinished.

clients/lsp-rust.el
```diff
@@ -860,6 +866,45 @@ them with `crate` or the crate name they refer to."
...
+                      (`(,spec) spec)
+                      (_ (user-error "Multiple compilation artifacts are not supported.")))))
+         (list :type "cppdbg"
```
nbfalcon, line 899: In the future, we should consider supporting other debuggers. This isn't a critique though, as it is probably out of scope here.

nbfalcon (approved):
Aside from some minor nits, LGTM.

clients/lsp-rust.el
nbfalcon: Actually, this wouldn't work, since there is no `setf seq-first`.

nbfalcon (commented):
.

clients/lsp-rust.el
yyoncho: sure.

clients/lsp-rust.el
yyoncho: Thanks, I will update it

clients/lsp-rust.el
yyoncho: Yes, support for all of the possible rust-friendly debug adapters is with lower priority.

clients/lsp-rust.el
yyoncho: I am not quite sure about that... My concern that we have several adapters with :type = "lldb". ("cppdbg" is actually wrong, should be "lldb").

nbfalcon: `counsel-projectile-rg`ing dap-mode, I don't see `dap-register-debug-provider "lldb" anywhere other than `dap-gdb-lldb.el`. Or do you propose renaming `cppdbg` to `lldb` in the future? Isn't `cpptools` for GDB? This PR would have to be adapted in the latter case anyway, though.

yyoncho: type is set to "lldb" in dap-codelldb as well.

> Or do you propose renaming cppdbg to lldb in the future?

I think we should at some point, otherwise, we won't work with launch.json coming from vscode.

PR #2665, opened 2021-03-03
Update options for lua-language-server

FollieHiyuki:
[schema.json](https://github.com/sumneko/vscode-lua/blob/master/setting/schema.json) for reference.
I also updated the version to 1.17.3 and changed some default values to the new ones.

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -168,6 +168,12 @@ fragment.  If it is set to `0`, this feature can be disabled."
   :package-version '(lsp-mode . "7.1")
   :group 'lsp-lua-language-server)
 
+(defcustom lsp-lua-completion-workspace-word t
+  "Show words within the workspace."
+  :type 'boolean
+  :package-version '(lsp-mode . "7.1")
+  :group 'lsp-lua-language-server)
+
 (defcustom lsp-lua-develop-debugger-port 11412
   "Listen port of debugger."
   :type 'number
@@ -216,6 +222,18 @@ fragment.  If it is set to `0`, this feature can be disabled."
   :package-version '(lsp-mode . "7.1")
   :group 'lsp-lua-language-server)
 
+(defcustom lsp-lua-diagnostics-needed-file-status nil
+  "If you want to check only opened files, choice Opened; else choice Any.
+```json
+\"Lua.diagnostics.neededFileStatus\" : {
+\"ambiguity-1\" : \"Any\",
+\"circle-doc-class\" : \"Opened\"
+}
+```"
+  :type 'alist
+  :package-version '(lsp-mode . "7.1")
+  :group 'lsp-lua-language-server)
+
 (defcustom lsp-lua-diagnostics-severity nil
   "Modified diagnostic severity.
 ```json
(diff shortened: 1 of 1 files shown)


yyoncho:
Looks good to me from elisp perspective! @sunlin7 willing to take a look/test?

FollieHiyuki:
1 note: the option `Lua.runtime.fileEncoding` does not work since upstream has typo on it (see the schema file). They got noticed so it will work in the future.
I currently use lua just to write basic config files, so I only need completion. Sorry for not being able to test it much.

yyoncho:
> I currently use lua just to write basic config files, so I only need completion. Sorry for not being able to test it much.

No worries, I am inviting users who use actively use lua to hear their opinion on the changes.

yyoncho (approved):

sunlin7:
Okay, I'll try this in my local, then update to you.

sunlin7:
I try follow this change upgrading `lua-language-server` to `1.17.3`, but seems not work as preview version (on CentOS 7).
Disgrading to `1.17.2` and continue testing the options now.

sunlin7:
I need more time for testing this change, there're high CPU (~120%) and memory (~3.4g) consuming issue and emacs stop response then.
I'm try my best to figure out the root cause.
image

FollieHiyuki:
@sunlin7 I think you can change `lsp-lua-workspace-max-preload` back to  300 (the new default is 1000) and lower `lsp-diagnostics-workspace-rate`

sunlin7:
@FollieHiyuki The high CPU and memory consumming issue already be reported as https://github.com/sumneko/lua-language-server/issues/439. 
For this PR, I also add some comments, please check them.

FollieHiyuki:
I don't know what we could do here.
It appears to me that `lua-language-server` will check files in workspace based on the current directory, correct? So the solution is to manually change pwd or to use `projectile`? Correct me if I'm wrong

sunlin7:
Hi @FollieHiyuki , The high CPU consumming issue should be NOT relate to current changes, it should be relate the initial workspace folder for `lsp-language-server`, I'll try my best to find the root cause.
For current changes, I submit serveral comments, please check them. Thanks.

sunlin7:
Setting the `lsp-auto-guess-root` to `t` can resolve the high CPU consumming issue in my local, with this option, the `lsp-mode` will detect the root dir correctly (without this option, the root uri is  ~ home folder).

yyoncho:
> Setting the `lsp-auto-guess-root` to `t` can resolve the high CPU consumming issue in my local, with this option, the `lsp-mode` will detect the root dir correctly (without this option, the root uri is ~ home folder).

lsp-mode won't pick automatically the HOME folder unless you have selected it in the initial import project dialog.

sunlin7:
@yyoncho The high CPU consumming issue is my local issue, and yes, the HOME was a root dir for I had edit a "~/test.lua" in HOME dir for days ago. 
After I remove `~/.emacs.d/.lsp-session-v1`, then reopen a lua file, there is a prompt for selecting the root folder, every thing work like charm now.
Thanks.

FollieHiyuki:
Thanks @sunlin7 for the comments. I fixed them (hopfully). Please review again.

yyoncho (approved):

yyoncho:
Thank you!

PR #2666, opened 2021-03-04
Preserve major mode for markdown rendering

### LINKED ISSUE #2588: Code blocks in MarkupContent are always highlighted as Markdown code
When rendering markdown inside MarkupContent, `markdown-fontify-code-block-default-mode ` is always set to `gfm-view-mode` (or `gfm-mode`) since that is the major mode activated just before calling `lsp--setup-markdown`:
https://github.com/emacs-lsp/lsp-mode/blob/62cd1b2e569c72638ba4bd42a0290192c224c28d/lsp-mode.el#L4723-L4727

However, saving `major-mode` before switching to the major mode is not sufficient since this whole function is called inside a temp buffer in `lsp--fontlock-with-mode`:
https://github.com/emacs-lsp/lsp-mode/blob/62cd1b2e569c72638ba4bd42a0290192c224c28d/lsp-mode.el#L4789-L4791
So at the beginning of `lsp--setup-markdown`, `major-mode` is already set to `fundamental-mode` instead of the desired mode of the actual buffer. Here `mode` is `lsp--setup-markdown`, which was retrieved from `lsp-language-id-configuration` when passing the MarkupContent's "kind" `"markdown"` to `lsp--render-string`:
https://github.com/emacs-lsp/lsp-mode/blob/62cd1b2e569c72638ba4bd42a0290192c224c28d/lsp-mode.el#L4816-L4817

yyoncho:
Fixes #2588

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4690,6 +4690,10 @@ Stolen from `org-copy-visible'."
              (s-chop-suffix "\n" (buffer-string)))
       (kill-buffer temp))))
 
+(defvar lsp-buffer-major-mode nil
+  "Holds the major mode when fontification function is running.
+See #2588")
+
 (defun lsp--render-markdown ()
   "Render markdown."
 
@@ -4707,7 +4711,7 @@ Stolen from `org-copy-visible'."
         (gfm-view-mode)
       (gfm-mode))
 
-    (lsp--setup-markdown major-mode)))
+    (lsp--setup-markdown lsp-buffer-major-mode)))
 
 (defvar lsp--display-inline-image-alist
   '((lsp--render-markdown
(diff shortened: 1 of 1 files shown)


ericdallo (approved):

lsp-mode.el
```diff
@@ -4768,15 +4772,16 @@ In addition, each can have property:
...
-  (condition-case nil
-      (with-temp-buffer
-        (insert str)
-        (delay-mode-hooks (funcall mode))
-        (cl-flet ((window-body-width () lsp-window-body-width))
-          (font-lock-ensure)
-          (lsp--display-inline-image mode))
-        (lsp--buffer-string-visible))
-    (error str)))
+  (let ((lsp-buffer-major-mode major-mode))
```
nbfalcon, line 4775: How does `lsp-buffer-major-mode` get to `lsp--setup-markdown`?

nbfalcon (commented):
Question.

PR #2667, opened 2021-03-04
Drop lsp--send-execute-command

yyoncho:
- fixes 2530
- fixes 2250

There is no way to distinguish whether the function is failing due to valid
reason or whether it is failing due to lack of client handling, thus removing
lsp--send-execute-command wrapper because it seems like it is causing more
trouble than good.

### DIFF
--- a/clients/lsp-clojure.el
+++ b/clients/lsp-clojure.el
@@ -79,7 +79,7 @@
 (defun lsp-clojure--execute-command (command &optional args)
   "Send an executeCommand request for COMMAND with ARGS."
   (lsp--cur-workspace-check)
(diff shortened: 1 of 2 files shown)


ericdallo (approved):

kiennq:
Can we at least log when `lsp-workspace-command-execute` is failing?
So at least in case of `lsp-mode` not implementing command, we can easily spot which command is missing?

yyoncho:
> So at least in case of `lsp-mode` not implementing command, we can easily spot which command is missing?

I was considering doing so, but given the fact that we cannot rethrow the error and keep the error callstack I decided not to do it. But if you wish I can add the logging.

PR #2670, opened 2021-03-05
[bash] Turn on lsp-mode only in sh/bash files

leungbk:
I saw this in @raxod502's config and thought it might be worth adding. I committed using his name.

### DIFF
--- a/clients/lsp-bash.el
+++ b/clients/lsp-bash.el
@@ -62,11 +62,20 @@ See instructions at https://marketplace.visualstudio.com/items?itemName=mads-har
                 '(:npm :package "bash-language-server"
                        :path "bash-language-server"))
 
+(defvar sh-shell)
+
+(defun lsp-bash-check-sh-shell (&rest _)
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

ErkiDerLoony:
I don’t quite understand why you disabled this by default. Is the syntax of most zsh files not quite similar to bash?
Is there an alternative for zsh files?

raxod502:
To my knowledge there is no language server for zsh at present, unfortunately. Zsh is quite a complex language, much more so than Bash.

> Is the syntax of most zsh files not quite similar to bash?

Well, yes, but there is plenty of Zsh syntax that will cause bash-language-server to produce an ugly error message, and that's highly misleading since the syntax is totally valid---plus, there's no way to fix the issue, other than disabling `lsp-mode` at that point.

ErkiDerLoony:
OK, thank you for the clarification. 🙃

Issue #2324, opened 2020-11-11
[3.16] [Completion] Support resolveSupport


kiennq:
This is new capability that will be replacing `completion.completionItem.resolveAdditionalTextEditsSupport`.
Please refer to new spec https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#textDocument_completion

ericdallo:
Also the spec adds: 
* `insertReplaceSupport`
* `insertTextModeSupport`

PR #2675, opened 2021-03-05
Try to get rust-analyzer flyimport completions working

flodiebold:
rust-analyzer can provide completions that automatically insert imports as needed, but this doesn't currently work with lsp-mode. These completions require `completionItem/resolve` support, which doesn't seem to be quite working.

I've fixed two things so far: the `resolveSupport` capability wasn't set; and `additional-text-edits?` was used in a condition, but it's a vector, which is truthy even when empty.

Sadly, it still doesn't work: The `completionItem/resolve` request gets sent, but rust-analyzer returns null. I think this is because lsp-mode sends a `didChange` notification before sending the request, which invalidates the completion item in rust-analyzer; but I haven't debugged it in detail yet.

You can reproduce the problem by applying this PR, then opening a new, empty cargo project and typing `repla` in the main function. You should get a completion for `std::mem::replace(...)` among others; if you select it, lsp-mode inserts `replace()` and nothing else. It should also insert a `use` statement.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -528,7 +528,7 @@ Others: CANDIDATES"
 
         (when lsp-completion-enable-additional-text-edit
           (if (or (get-text-property 0 'lsp-completion-resolved candidate)
-                  additional-text-edits?)
+                  (lsp-seq-first additional-text-edits?))
               (lsp--apply-text-edits additional-text-edits? 'completion)
             (-let [(callback cleanup-fn) (lsp--create-apply-text-edits-handlers)]
               (lsp-completion--resolve-async
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3258,6 +3258,7 @@ disappearing, unset all the variables related to it."
                                                                              :json-false)
                                                                             (lsp-enable-snippet t)
                                                                             (t :json-false)))
+                                                        (resolveSupport . ((properties . ["additionalTextEdits"])))
                                                         (documentationFormat . ["markdown"])
                                                         (resolveAdditionalTextEditsSupport . t)))
                                      (contextSupport . t)))


yyoncho:
I don't think that we can fix that without changing company-mode (or at least having an ugly hack in it)...

@flodiebold can you confirm that the following will work with RA

1. User presses complete 
2. We perform resolve of the item without changing the buffer.
3. Without waiting for the result of 2 we insert the completion item
4. When received the response for 2 we apply the additional text edits.

@dgutov it seems like we cannot do without company-mode calling lsp-mode completion backend to perform the actual insertion after `company-complete-selection`. I think that this is a relatively easy change on company-mode's side. WDYT?

@kiennq - willing to take care of this?

flodiebold:
I think that would likely cause the resolve request to be canceled, unless it completes fast enough to be finished before the onChange arrives :confused:

yyoncho:
> If we can send the request before changing the buffer, can we do it synchronously and get the result before sending the change?

This will lead to poor UE. See discussion here https://github.com/microsoft/language-server-protocol/issues/1003#issuecomment-653499069 . I wonder how vscode works? Maybe they resolve in advance?

dgutov:
> I wonder how vscode works?

I'm curious, too.

> we cannot do without company-mode calling lsp-mode completion backend to perform the actual insertion after company-complete-selection

What would `company-complete-selection` do instead, then?

I wonder if you can make do simply with a pre-insertion callback.

kiennq:
@flodiebold 
> I've fixed two things so far: the `resolveSupport` capability wasn't set;

Thank you, this is tracked by #2324. I haven't work on that though :)


> and `additional-text-edits?` was used in a condition, but it's a vector, which is truthy even when empty.

I would expect the `CompletionItems` doesn't contain `additionalTextEdits?` if it needs resolve for additional properties since `additionalTextEdits?` is optional.
Currently we're still using hash table which can differentiate between `null` json element and empty json element (`{}` and `[]`) so in case server has return `additionalTextEdits` with empty json value, we considered that server has passed `additionalTextEdits` already and don't look forward to trigger resolving again.

Can you take a look at why RA sending empty `additionalTextEdits`? If it's because of `didChange` is being sent, that means the server somehow force the client to apply additional text edits synchronously? Which is kind of against the spec of LSP and behavior of VsCode.

Note that we've already support lazy resolve for properties that's not in `label`, `sortText`, `filterText`, `insertText` and `textEdit`

kiennq:
Also can you try https://github.com/kiennq/lsp-mode/commit/a79201ef1cb5b4f9e4826a2cb0168e22b5ed062a to see if it's somehow working?

yyoncho:
> differentiate between `null` json element and empty json element (`{}` and `[]`)

Just want to clarify that we cannot easily differentiate `null` from `{}`, `[]` works fine. To differentiate `null` from `{}` we have to parse the response twice.

kiennq:
> > differentiate between `null` json element and empty json element (`{}` and `[]`)
> 
> Just want to clarify that we cannot easily differentiate `null` from `{}`, `[]` works fine. To differentiate `null` from `{}` we have to parse the response twice.

Here is when I try to run `lsp--read-json`
``` elisp
(lsp--read-json "{\"a\": {}}")
#s(hash-table size 1 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("a" #s(hash-table size 1 test equal rehash-size 1.5 rehash-threshold 0.8125 data ())))
(lsp--read-json "{\"a\": null}")
#s(hash-table size 1 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("a" nil))
(lsp--read-json "{\"a\": []}")
#s(hash-table size 1 test equal rehash-size 1.5 rehash-threshold 0.8125 data ("a" []))
```

It seems that `{}` and `null` can be differentiate by empty `hashtable` vs `nil`?

yyoncho:
> It seems that `{}` and `null` can be differentiate by empty `hashtable` vs `nil`?

Yes, unless you are using `lsp-use-plists = t`.

flodiebold:
rust-analyzer always returns an empty list for `additionalTextEdits`, never null. That seems to work fine with VSCode. I think the client should try to resolve if the edits are null or empty, unless this is explicitly specified otherwise?

I will try to verify my theory that this isn't working because of the change events; after reading the RA code, I'm a bit less convinced.

kiennq:
I think this maybe because of `autoimport` is not specified.
```
                "rust-analyzer.completion.autoimport.enable": {
                    "markdownDescription": "Toggles the additional completions that automatically add imports when completed. Note that your client must specify the `additionalTextEdits` LSP client capability to truly have this feature enabled.",
                    "default": true,
                    "type": "boolean"
                },
```

Will try to verify it on my branch.

flodiebold:
Ah no, it's a JSON roundtripping problem... The `data` field for these completion items has a field set to `false`, and when sending it back to the server, that field gets written as `null`, which RA doesn't accept.

yyoncho:
Here is what the spec says:

"The Completion request is sent from the client to the server to compute completion items at a given cursor position. Completion items are presented in the IntelliSense user interface. If computing full completion items is expensive, servers can additionally provide a handler for the completion item resolve request (‘completionItem/resolve’). This request is sent when a completion item is selected in the user interface. A typical use case is for example: the ‘textDocument/completion’ request doesn’t fill in the documentation property for returned completion items since it is expensive to compute. When the item is selected in the user interface then a ‘completionItem/resolve’ request is sent with the selected completion item as a parameter. "

Most likely vscode resolves the completion items no matter what are their properties when the item is selected in the UI. I guess we can implement that on lsp-mode side. 

But rust analyzer being unable to resolve an item after the buffer has changed means that we cannot perform any client-side caching of the completion items which IMO is bad and again, I am not sure how it works with vscode given the fact that vscode caches completion items by default(or at least it used to do so).

flodiebold:
> But rust analyzer being unable to resolve an item after the buffer has changed means that we cannot perform any client-side caching of the completion items which IMO is bad and again, I am not sure how it works with vscode given the fact that vscode caches completion items by default(or at least it used to do so).

Yeah I think I was just wrong about that, see above.

yyoncho:
> Ah no, it's a JSON roundtripping problem... The `data` field for these completion items has a field set to `false`, and when sending it back to the server, that field gets written as `null`, which RA doesn't accept.

We have that problem in a few other places and they are fixed by using hacks. Unfortunately, when the project started it picked json settings that make the serialization/deserialization asymmetrical and it is too late to fix that.

yyoncho:
@kiennq proposed fix: Patch the field in rust analyzer's CompletionItem.data to be :json-false. Eventually, we might add an integration test to check if rust analyzer has changed its CompletionItem.data format. The other option is to ask the RA team to accept null or to avoid using booleans in data which can be rejected because after all, it is our issue...

kiennq:
Yeah, we have to patch this somehow ...
I would prefer a more throughout approach but for the mean time, the patching `data` field seems to be most feasible.

flodiebold:
The field in question is `import_for_trait_assoc_item`, btw.

yyoncho:
I will do the RA related patch.

yyoncho:
Just want to confirm that after patching `import_for_trait_assoc_item` the import works fine.


@kiennq - seems like lsp-completion--resolve-async is called 2 times per item.

yyoncho:
PR at #2695 , I will close this one.

PR #2676, opened 2021-03-05
lsp-csharp.el: add "lsp-csharp-solution-file" customizable

razzmatazz:
This can be used to select a particular solution file to load out of
many in a large project to reduce loading time by loading just the
projects that are included in the solution only.


yyoncho (approved):
Merging as it is. Although, we typically have one generic param `args` which is a list of strings.

PR #2677, opened 2021-03-05
Use compilation for `lsp-install` buffers (`lsp-async-start-process`)

nbfalcon:
- There is now a separate buffer for each `M-x lsp-install-server` invocation,
  instead of a single `*lsp-install*` buffer.
- `ansi` escapes now install buffers, so `npm install`/everything else now looks
  pretty and colourful instead of gray with escapes intermingled
- The modeline button to select the `*lsp-install*` buffer still works, but it
  now uses `completing-read`, with the code refactored to a new function
  + this function colourises buffer names based on whether the installation is
    still running
  + if there is only a single installation buffer, it is switched to immediately
    (the various commits go into more detail)

----

#

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6997,6 +6997,20 @@ JavaScript file, tsserver.js (the *.js is required for Windows)."
       (error nil)
       (args-out-of-range nil))))
 
+(defvar-local lsp--installation-buffer nil)
+
+(defun lsp-select-installation-buffer ()
+  "Interactively choose an installation buffer."
+  (interactive)
+  (let ((bufs (--filter (and (buffer-local-value 'lsp--installation-buffer it)
+                             (get-buffer-process it))
+                        (buffer-list))))
+    (pcase bufs
+      (`nil (user-error "No installation buffers"))
+      (`(,buf) (pop-to-buffer buf))
+      (bufs (pop-to-buffer (completing-read "Select installation buffer: "
+                                            (mapcar #'buffer-name bufs)))))))
+
 (defun lsp--download-status ()
   (-some--> #'lsp--client-download-in-progress?
     (lsp--filter-clients it)
@@ -7006,9 +7020,7 @@ JavaScript file, tsserver.js (the *.js is required for Windows)."
     (format " Installing following servers: %s" it)
     (propertize it
                 'local-map (make-mode-line-mouse-map
-                            'mouse-1 (lambda ()
-                                       (interactive)
-                                       (switch-to-buffer (get-buffer-create " *lsp-install*"))))
+                            'mouse-1 #'lsp-select-installation-buffer)
                 'mouse-face 'highlight)))
 
 (defun lsp--install-server-internal (client &optional update?)
@@ -7081,22 +7093,24 @@ When prefix UPDATE? is t force installation even if the server is present."
 
 (defun lsp-async-start-process (callback error-callback &rest command)
   "Start async process COMMAND with CALLBACK and ERROR-CALLBACK."
-  (make-process
-   :name (cl-first command)
-   :command command
-   :sentinel (lambda (proc _)
-               (when (eq 'exit (process-status proc))
-                 (if (zerop (process-exit-status proc))
-                     (condition-case err
-                         (funcall callback)
-                       (error
-                        (funcall error-callback (error-message-string err))))
-                   (display-buffer " *lsp-install*")
-                   (funcall error-callback
-                            (format "Async process '%s' failed with exit code %d"
-                                    (process-name proc) (process-exit-status proc))))))
-   :buffer " *lsp-install*"
-   :noquery t))
+  (let ((name (cl-first command)))
+    (with-current-buffer (compilation-start (combine-and-quote-strings command) t
+                                            (lambda (&rest _)
+                                              (generate-new-buffer-name (format "*lsp-install: %s*" name))))
+      (setq-local lsp--installation-buffer t)
+      (add-hook
+       'compilation-finish-functions
+       (lambda (buf status)
+         (if (string= "finished\n" status)
+             (condition-case err
+                 (funcall callback)
+               (error
+                (funcall error-callback (error-message-string err))))
+           (funcall error-callback
+                    (format "Async process '%s' failed with exit code %d"
+                            name (-> (current-buffer) get-buffer-process
+                                     process-exit-status)))))
+       nil t))))
 
 (defun lsp-resolve-value (value)
   "Resolve VALUE's value.
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6999,11 +6999,13 @@ JavaScript file, tsserver.js (the *.js is required for Windows)."
 
 (defvar-local lsp--installation-buffer nil)
 
-(defun lsp-select-installation-buffer ()
-  "Interactively choose an installation buffer."
-  (interactive)
+(defun lsp-select-installation-buffer (&optional show-finished)
+  "Interactively choose an installation buffer.
+If SHOW-FINISHED is set, leftover (finished) installation buffers
+are still shown."
+  (interactive "P")
   (let ((bufs (--filter (and (buffer-local-value 'lsp--installation-buffer it)
-                             (get-buffer-process it))
+                             (or show-finished (get-buffer-process it)))
                         (buffer-list))))
     (pcase bufs
       (`nil (user-error "No installation buffers"))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6999,6 +6999,16 @@ JavaScript file, tsserver.js (the *.js is required for Windows)."
 
 (defvar-local lsp--installation-buffer nil)
 
+(defface lsp-installation-finished-buffer-face '((t :foreground "orange"))
+  "Face used for finished installation buffers.
+Used in `lsp-select-installation-buffer'."
+  :group 'lsp-mode)
+
+(defface lsp-installation-buffer-face '((t :foreground "green"))
+  "Face used for installation buffers still in progress.
+Used in `lsp-select-installation-buffer'."
+  :group 'lsp-mode)
+
 (defun lsp-select-installation-buffer (&optional show-finished)
   "Interactively choose an installation buffer.
 If SHOW-FINISHED is set, leftover (finished) installation buffers
@@ -7011,7 +7021,11 @@ are still shown."
       (`nil (user-error "No installation buffers"))
       (`(,buf) (pop-to-buffer buf))
       (bufs (pop-to-buffer (completing-read "Select installation buffer: "
-                                            (mapcar #'buffer-name bufs)))))))
+                                            (--map (propertize (buffer-name it) 'face
+                                                               (if (get-buffer-process it)
+                                                                   'lsp-installation-buffer-face
+                                                                 'lsp-installation-finished-buffer-face))
+                                                   bufs)))))))
 
 (defun lsp--download-status ()
   (-some--> #'lsp--client-download-in-progress?


nbfalcon:
Actually, one more thing is missing: `lsp-pwsh` calls `lsp-async-start-process` twice; it will have to be refactored into a single command.

lsp-mode.el
```diff
@@ -7081,22 +7109,24 @@ When prefix UPDATE? is t force installation even if the server is present."
...
-                   (funcall error-callback
-                            (format "Async process '%s' failed with exit code %d"
-                                    (process-name proc) (process-exit-status proc))))))
-   :buffer " *lsp-install*"
-   :noquery t))
+  (let ((name (cl-first command)))
+    (with-current-buffer (compilation-start (combine-and-quote-strings command) t
+                                            (lambda (&rest _)
+                                              (generate-new-buffer-name (format "*lsp-install: %s*" name))))
+      (setq-local lsp--installation-buffer t)
```
kiennq, line 7116: This is already local buffer var, `setq is enough?

lsp-mode.el
```diff
@@ -7081,22 +7109,24 @@ When prefix UPDATE? is t force installation even if the server is present."
...
+       'compilation-finish-functions
+       (lambda (buf status)
+         (if (string= "finished\n" status)
+             (condition-case err
+                 (funcall callback)
+               (error
+                (funcall error-callback (error-message-string err))))
+           (funcall error-callback
+                    (format "Async process '%s' failed with exit code %d"
+                            name (-> (current-buffer) get-buffer-process
```
kiennq, line 7127: Shouldn't this be `buf`?

lsp-mode.el
```diff
@@ -7081,22 +7109,24 @@ When prefix UPDATE? is t force installation even if the server is present."
...
-                       (error
-                        (funcall error-callback (error-message-string err))))
-                   (display-buffer " *lsp-install*")
-                   (funcall error-callback
-                            (format "Async process '%s' failed with exit code %d"
-                                    (process-name proc) (process-exit-status proc))))))
-   :buffer " *lsp-install*"
-   :noquery t))
+  (let ((name (cl-first command)))
+    (with-current-buffer (compilation-start (combine-and-quote-strings command) t
```
kiennq, line 7113: Is this `combine-and-quote-string` works on Windows? Its document is saying
>Note that this is not intended to protect STRINGS from
interpretation by shells, use ‘shell-quote-argument’ for that.

nbfalcon: It indeed doesn't; I am currently refactoring the powershell installation script to use a single lsp-async-start-process instead, and this caused breakage. I will change it to `(mapconcat #'shell-quote-argument args " ")` soon.

nbfalcon:
For some reason, powershell inserts lots spurious whitespace after commands like Expand-Archive and Invoke-WebRequest:
```shell
/tmp λ powershell -Command 'Out-Null -inputObject (Expand-Archive -LiteralPath "zip.zip" -DestinationPath "out-zip" -Force)' | xxd
00000000: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000010: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000020: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000030: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000040: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000050: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000060: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000070: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000080: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000090: 2020 2020 2020 2020 2020 2020 2020 2020                  
000000a0: 2020 2020 2020 2020 2020 2020 2020 2020                  
000000b0: 2020 2020 2020 2020 2020 2020 2020 2020                  
000000c0: 2020 2020 2020 2020 2020 2020 2020 2020                  
000000d0: 2020 2020 2020 2020 2020 2020 2020 2020                  
000000e0: 2020 2020 2020 2020 2020 2020 2020 2020                  
000000f0: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000100: 2020 2020 2020 2020 2020 2020 2020 2020                  
00000110: 2020 2020 2020 2020 2020 2020 2020 2020              
[...]
```
where those "20" hexadecimals represent whitespace characters. @kiennq any ideas? I could only fix this for InvokeWebRequest by using `System.Net.WebClient`.

In any case, here is the updated `lsp-pwsh-setup` function:
```emacs-lisp
(defun lsp-pwsh--quote (s)
  "Quote S for powershell.
Like `shell-quote-argument' but for powershell instead, which
uses a different escape character (`)."
  (replace-regexp-in-string "\\\\\\([^z-a]\\)" "`\\1"
                            (shell-quote-argument s) t))

(defun lsp-pwsh-setup (_client callback error-callback update)
  "Downloads PowerShellEditorServices to `lsp-pwsh-dir'.
CALLBACK is called when the download finish successfully otherwise
ERROR-CALLBACK is called.
UPDATE is non-nil if it is already downloaded.
FORCED if specified with prefix argument."

  (unless (and lsp-pwsh-exe (file-executable-p lsp-pwsh-exe))
    (user-error "Use `lsp-pwsh-exe' with the value of `%s' is not a valid powershell binary"
                lsp-pwsh-exe))

  (let ((url (format lsp-pwsh-github-asset-url "PowerShell"
                     "PowerShellEditorServices" "PowerShellEditorServices.zip")))
    (unless (f-exists? lsp-pwsh-log-path)
      (mkdir lsp-pwsh-log-path 'create-parent))
    (unless (and (not update) (f-exists? lsp-pwsh-dir))
      ;; since we know it's installed, use powershell to download the file
      ;; (and avoid url.el bugginess or additional libraries)
      (lsp-async-start-process
       callback error-callback
       lsp-pwsh-exe "-NoProfile" "-NonInteractive" "-NoLogo" "-Ex" "bypass"
       ;; We can't use Invoke-WebRequest here, because it prints a lot of
       ;; spurious whitespace in conjunction with `Write-Host'. This is probably
       ;; a powershell bug (not Even Out-Null helps here):
       ;;
       ;; powershell -Command 'Write-Output test; $req = Invoke-WebRequest "https://example.com/index.html";' | xxd
       "-Command" (format "
$url = \"%s\"; $dest = \"%s\";
Write-Host \"`e[33mDownloading`e[0m `e[34;4m$url`e[0m...\"; $req = (New-Object System.Net.WebClient).DownloadString($url);
Write-Host \"`e[32mDone!`e[0m `e[33mExtracting to `e[0m`e[34;4m$dest`e[0m...\";
$ProgressReference = 'SilentlyContinue';
$req | Expand-Archive -DestinationPath $dest;
Write-Host \"`e[32mDone! Installation finished.`e[0m\""
                          (lsp-pwsh--quote url)
                          (lsp-pwsh--quote lsp-pwsh-dir))))))
```

lsp-mode.el
nbfalcon: Imho `setq-local` is clearer in such cases: the `defvar-local` tells the reader that the variable will always be local, and the `setq-local` here also does. This way, one needn't jump to `lsp--installation-buffer` first to check if it is local.

lsp-mode.el
nbfalcon: This was eliminated, because after compilation finishes, `get-buffer-process` always returns nil. I decided to simply use the `status` string as the argument to error-callback.

nbfalcon:
Ideas for the future (different PRs, otherwise this one will explode in scope even more):
- when installing multiple servers, clicking a specific server id in the modeline should switch to its installation buffer
- the buffer names for installation buffers should be more intuitive: instead of the command run, it should be something like `*lsp-installation: npm vls*` and should show the server id being installed in the modeline.

This could be merged as is; using a single call to `lsp-async-start-process` for powershell is better, but doesn't work right now because apparently powershell is buggy.

lsp-mode.el
kiennq: Well, I just want to write code without worrying about variable is buffer local or not, so using ubiquitous `setq` everywhere.
The `setq-local` is just a macro that expanded to `(set (make-local-variable xyz) <value>)` so we spends cycles to make variable local everytime this macro is called.
Declaring variable as local once also make it easier if we decide to change the variable back to global, just changing from `defvar-local` -> `defvar` and it will be fine.

kiennq:
I don't think switch to `compilation-start` on Windows is good option, Windows has very poor supports for command line, especially the `cmd-proxy` that comes with Emacs package on Windows.
Can we keep the old `make-process` code and turn the `compilation-mode` on the log buffer instead?

lsp-mode.el
nbfalcon: The performance overhead is insanely small: doing `make-local-variable` a million times takes less than a second, and in this case, the variable really has to be buffer local for it to make any sense. With that in mind, what do you think? Should I prefer `setq`? There are arguments for both, since the variable's name also implies that it would be local.

Benchmark:
```emacs-lisp
(benchmark-run 1000000 (make-local-variable 'abc))
(0.070333... 0 0.0)
```

yyoncho: If you are interested what is the proper way, ask in emacs-devel what is the convention to use defvar-local variables. Except for the naming we tend to follow their conventions.

nbfalcon:
@kiennq I just booted up Windows 10 and did some general `compilation-start` testing (Emacs-27.1):
- `shell-quote-argument` works
- Powershell does not support `\`e`; however, this can be worked around by using \033, I.e. doing the escaping in Emacs directly, not in Powershell
- There are some pitfalls, like "echo \"Meow\"" yielding `"Meow"`, i.e. with quotes

I will fix the escape handling in the windows powershell install script. What other issues did you encounter? The terminal support doesn't really matter, since Emacs will be doing the ansi-escape-handling.

kiennq:
> I will fix the escape handling in the new (not-yet pushed) windows powershell install script. What other issues did you encounter? The terminal support doesn't really matter, since Emacs will be doing the ansi-escape-handling.

Doing manually escaping in Emacs lisp seems to be catching up work though. Personally I would like to avoid the quoting stuff entirely since they're brittle and need good coverage test to ensure our script just not stop working unexpectedly.
One of the issue that maybe encounter is https://github.com/abo-abo/swiper/issues/2517, which may cause hang if this is invoked multiple time.
Is there strong reason to switch to using `compilation-start` than just setting compilation mode in the result buffer?

nbfalcon:
The quoting is only an issue if we implement our own pretty custom installers, like what I was trying to do for `powershell`. The change is not in this PR though, and I will test it on windows once I manage to iron out the space bug. We should not encounter the swiper bug, because `lsp-async-start-process` always creates a new buffer (`generate-new-buffer-name`), and we don't ever call `delete-process`. Using `compilation-start` here is much simpler and more elegant, since it sets up a lot of stuff behind the scenes (mode-line indicator, ...).

nbfalcon:
Upon further investigation, the shell-quoting difference I mentioned doesn't exist. `(compilation-start (mapconcat #'shell-quote-argument '("echo" "\"meow\"") " "))` works the same way on Linux, since we are calling shell-quote-argument on the second string, so it should be quoted in the output.

@kiennq what version of Emacs are you using on Windows, and version of Windows are you using?

kiennq:
I'm using `GNU Emacs 28.0.50 (build 1, x86_64-w64-mingw32) of 2021-01-16` with Windows 10 `2004 build 21301.1010`

@nbfalcon Did you try with Unicode character to see if it works properly with `compilation-start` version? Something like including folder named `令和` in the command.

nbfalcon:
@kiennq I will try later, I don't want to boot windows right now. Do you mean in the command, or in the output? I'd expect the latter to work fine, since Emacs is doing the displaying; the former might cause problems, but we don't use unicode characters (the `\\033` escape is still ASCII), unless npm packages can have non-ascii names (I don't expect that to be the case).

nbfalcon:
@kiennq Actually, nevermind, I just did some testing: the characters you mentioned work just fine in `compilation-start`, both as input and as output, even with ANSI escapes added (`4 = underlining effect`), as you can see from this screenshot:
Capture
taken on Windows 10 with Emacs 27.1.

lsp-mode.el
```diff
@@ -7081,22 +7126,21 @@ When prefix UPDATE? is t force installation even if the server is present."
...
-                            (format "Async process '%s' failed with exit code %d"
-                                    (process-name proc) (process-exit-status proc))))))
-   :buffer " *lsp-install*"
-   :noquery t))
+  (let ((name (cl-first command)))
+    (with-current-buffer (compilation-start (mapconcat #'shell-quote-argument command " ") t
+                                            (lambda (&rest _)
+                                              (generate-new-buffer-name (format "*lsp-install: %s*" name))))
+      (lsp-installation-buffer-mode +1)
+      (add-hook
```
kiennq, line 7134: Don't you need to clean this lambda from `compilation-finish-functions`? Else the unrelated callback might be called for unrelated async command

nbfalcon: Since that hook is run only once, I'd say no. Are you worried about unecessary garbage?

nbfalcon:
I just tested the current `pwsh` installer; it works, but it causes two compilation buffers. It also suffers from the whitespace bug in powershell.

lsp-mode.el
kiennq: It seems that `compilation-finish-functions` is global variable, so if you run `lsp-async-start-process` multiple times to install multiple servers, shouldn't each call will add new closure to the hook?
Also, without checking the `_buf`, how can you make sure that the status is intended for the callback to react to?

nbfalcon: The fourth parameter, if set (as in this case) makes the hook buffer-local before setting it, and adds `t` to the hook variable given (making `run-hooks` run the global one as well).

kiennq: You're right, I didn't notice the `local` flag being set

nbfalcon:
Is there something blocking this?

yyoncho (approved):

yyoncho:
I tested it briefly, seems like everything is working as expected. We might think of restructuring the code to allow better naming of the compilation buffer, but that is a minor issue.

nbfalcon:
Thanks!

nbfalcon:
@yyoncho does this fall under
> the buffer names for installation buffers should be more intuitive: instead of the command run, it should be
> something like *lsp-installation: npm vls* and should show the server id being installed in the modeline.

yyoncho:
yes, I didn't follow the whole discussion(I was more focused on the testing side for this PR).

Issue #2655, opened 2021-02-28
Test emacs-lsp packages


yyoncho:
It will be nice if we have once centralized test suite which checks whether
emacs-lsp packages byte compile properly since a lot of the extension packages
use lsp-mode's internal functions(e. g. `lsp--point-to-position`).

yyoncho:
@jcs090218 willing to take a look?

jcs090218:
Sure! How would you like to proceed this? I think I have applied a CI test for most of emacs-lsp packages.

yyoncho:
> How would you like to proceed this?

The goal is to byte-compile the packages part of emacs-lsp organization against the version of lsp-mode under test. I am not sure what will be the best/easiest way to achieve this - we can either clone their repos or install them via package.el and then filter out the byte-compilation errors coming from non-emacs-lsp packages.

jcs090218:
> we can either clone their repos or install them via package.el and then filter out the byte-compilation errors coming from non-emacs-lsp packages.

Hmm.. I think we will need to create a new repository to do this task? If yes, what name would you prefer? I think I can work on this over the weekend. :)

yyoncho:
> I think we will need to create a new repository to do this task

Why? Can we use that in order to get the PRs covered?

jcs090218:
> Why? Can we use that in order to get the PRs covered?

Okay, I think I misunderstand your plan. Can you elaborate it so I can get a better image of what you are proposing? I am guessing that you want to filter out error/warning message from the emacs-lsp packages' dependencies? So we can get debugging easier if there is something wrong from their dependencies?

I thought the line below should already work? Since it already triggered the error flag while dependency goes wrong.

https://github.com/emacs-lsp/lsp-mode/blob/89a1525be9606e60f708052f86f501e934a14397/Makefile#L55

yyoncho:
> Can you elaborate it so I can get a better image of what you are proposing?

I want the following: Developer FOO is opening a PR which renames `lsp--position-compare` to `lsp--position-compare-new-name` . Developer FOO opens a PR and CI fails, becase `lsp-java` is using `lsp--position-compare`.

yyoncho:
FTR seems like we should include also external packages like `ccls`. This will help us be more aggressive with fixing poor function names and with refactoring.

ericdallo:
[...]

PR #2678, opened 2021-03-06
CI test for emacs-lsp packages

jcs090218:
This if for #2655

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,10 @@ unix-build:
 	$(CASK) install
 
 # TODO: add 'checkdoc' and 'lint' here when they pass
-unix-ci: clean unix-build unix-compile prepare_cpp_project unix-test test-pkgs
+unix-ci: clean unix-build unix-compile prepare_cpp_project unix-test test-downstream-pkgs
 
 windows-ci: CASK=
-windows-ci: clean windows-compile windows-test test-pkgs
+windows-ci: clean windows-compile windows-test test-downstream-pkgs
 
 unix-compile:
 	@echo "Compiling..."
@@ -58,7 +58,7 @@ windows-compile:
 		--eval '(setq byte-compile-error-on-warn t)' \
 		-f batch-byte-compile $(LSP-FILES)
 
-test-pkgs:
+test-downstream-pkgs:
 	@echo "Test emacs-lsp packages..."
 	@$(CASK) $(EMACS) -Q --batch \
 		-l $(WIN-BOOTSTRAP) \
@@ -117,4 +117,4 @@ clean:
 	rm -rf test/fixtures/SampleCppProject/build test/fixtures/SampleCppProject/.cache
 
 
-.PHONY: all unix-build	 ci unix-compile windows-compile checkdoc lint unix-test windows-test docs local-webpage clean
+.PHONY: all unix-build ci unix-compile windows-compile checkdoc lint unix-test windows-test docs local-webpage clean
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,7 @@ windows-compile:
 test-pkgs:
 	@echo "Test emacs-lsp packages..."
 	@$(CASK) $(EMACS) -Q --batch \
+		-l $(WIN-BOOTSTRAP) \
 		--eval '(setq byte-compile-error-on-warn t)' \
 		-l $(TEST-PKGS)
 
--- a/Makefile
+++ b/Makefile
@@ -63,6 +63,7 @@ test-pkgs:
 	@$(CASK) $(EMACS) -Q --batch \
 		-l $(WIN-BOOTSTRAP) \
 		--eval '(setq byte-compile-error-on-warn t)' \
+		--eval '(setq emacs-lsp-ci t)' \
 		-l $(TEST-PKGS)
 
 checkdoc:
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,10 @@ LSP-FILES := lsp-protocol.el lsp-mode.el lsp.el lsp-completion.el \
 		lsp-diagnostics.el lsp-lens.el lsp-modeline.el \
 		$(wildcard clients/*.el)
 
-TEST-FILES := test/windows-bootstrap.el test/test-helper.el \
+WIN-BOOTSTRAP=test/windows-bootstrap.el
+TEST-PKGS=test/test-packages.el
+
+TEST-FILES := $(WIN-BOOTSTRAP) $(TEST-PKGS) test/test-helper.el \
 		$(shell ls test/lsp-*.el)
 LOAD-FILE = -l $(test-file)
 LOAD-TEST-FILES := $(foreach test-file, $(TEST-FILES), $(LOAD-FILE))
@@ -50,7 +53,7 @@ prepare_cpp_project:
 windows-compile:
 	@echo "Compiling..."
 	@$(CASK) $(EMACS) -Q --batch \
-		-l test/windows-bootstrap.el \
+		-l $(WIN-BOOTSTRAP) \
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
 		-f batch-byte-compile $(LSP-FILES)
@@ -83,11 +86,11 @@ lint:
 		$(LSP-FILES)
 
 unix-test:
-	$(CASK) exec ert-runner -L . -L clients  -t '!no-win' -t '!org'
+	$(CASK) exec ert-runner -L . -L clients	 -t '!no-win' -t '!org'
 
 windows-test:
 	@$(EMACS) -Q --batch \
-		-l test/windows-bootstrap.el \
+		-l $(WIN-BOOTSTRAP) \
 		-L . -L clients \
 		$(LOAD-TEST-FILES) \
 		--eval "(ert-run-tests-batch-and-exit \
--- a/Makefile
+++ b/Makefile
@@ -34,10 +34,10 @@ unix-build:
 	$(CASK) install
 
 # TODO: add 'checkdoc' and 'lint' here when they pass
-unix-ci: clean unix-build unix-compile prepare_cpp_project unix-test
+unix-ci: clean unix-build unix-compile prepare_cpp_project unix-test test-pkgs
 
 windows-ci: CASK=
-windows-ci: clean windows-compile windows-test
+windows-ci: clean windows-compile windows-test test-pkgs
 
 unix-compile:
 	@echo "Compiling..."
@@ -57,6 +57,12 @@ windows-compile:
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
 		-f batch-byte-compile $(LSP-FILES)
+
+test-pkgs:
+	@echo "Test emacs-lsp packages..."
+	@$(CASK) $(EMACS) -Q --batch \
+		--eval '(setq byte-compile-error-on-warn t)' \
+		-l $(TEST-PKGS)
 
 checkdoc:
 	$(eval LOG := $(shell mktemp -d)/checklog.log)
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,6 @@ test-pkgs:
 	@echo "Test emacs-lsp packages..."
 	@$(CASK) $(EMACS) -Q --batch \
 		-l $(WIN-BOOTSTRAP) \
-		--eval '(setq byte-compile-error-on-warn t)' \
 		--eval '(setq emacs-lsp-ci t)' \
 		-l $(TEST-PKGS)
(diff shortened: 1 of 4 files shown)


jcs090218:
This is 99% done except `lsp-dart`. I am getting the following error but not able to find the function `lsp-make-notification` anywhere inside `lsp-dart` package. This looks very strange to me. 😕 

Here is the full backtrace from GitHub Actions, 

```
d:/a/lsp-mode/lsp-mode/lsp-mode.el:2784:8:Error: function lsp-make-notification used to take 0+ arguments, now takes 1-2
Test lsp-byte-compilation-test backtrace:
  signal(ert-test-failed (((should (byte-compile-file (save-excursion 
  ert-fail(((should (byte-compile-file (save-excursion (find-library l
  (if (unwind-protect (setq value-701 (apply fn-699 args-700)) (setq f
  (let (form-description-703) (if (unwind-protect (setq value-701 (app
  (let ((value-701 (quote ert-form-evaluation-aborted-702))) (let (for
  (let* ((fn-699 (function byte-compile-file)) (args-700 (condition-ca
  (let ((byte-compile-error-on-warn t)) (message "Testing file %s" lib
  (closure (t) (library) (let ((byte-compile-error-on-warn t)) (messag
  mapc((closure (t) (library) (let ((byte-compile-error-on-warn t)) (m
  seq-do((closure (t) (library) (let ((byte-compile-error-on-warn t)) 
  (closure (t) nil (seq-do (function (lambda (library) (let ((byte-com
  ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
  ert-run-test(#s(ert-test :name lsp-byte-compilation-test :documentat
  ert-run-or-rerun-test(#s(ert--stats :selector (and (not (tag no-win)
  ert-run-tests((and (not (tag no-win)) (not (tag org))) #f(compiled-f
  ert-run-tests-batch((and (not (tag no-win)) (not (tag org))))
  ert-run-tests-batch-and-exit((and (not (tag no-win)) (not (tag org))
  eval((ert-run-tests-batch-and-exit (quote (and (not (tag no-win)) (n
  command-line-1(("-l" "test/windows-bootstrap.el" "-L" "." "-L" "clie
  command-line()
  normal-top-level()
    (ert-test-failed
     ((should
       (byte-compile-file
	(save-excursion ... ...)))
      :form
      (byte-compile-file "d:/a/lsp-mode/lsp-mode/lsp-mode.el")
      :value nil))
```

P.S. This only occurs in Windows

ericdallo:
Let me check @jcs090218

test/windows-bootstrap.el
```diff
@@ -41,6 +40,16 @@
...
-            `(lambda () (delete-directory ,user-emacs-directory t))))
+            `(lambda ()
+               ;; NOTE: If you are testing this on your own PC, then you
+               ;; might want to uncomment the line below to do the clean up
+               ;; after the test. (This will take more time to complete the
+               ;; CI process)
+               ;;
+               ;; Please leave it comment in CI so we don't accidentally install
+               ;; depedencies twice.
+               ;;(delete-directory ,user-emacs-directory t)
```
ericdallo, line 51: Instead of this, we could check a ENV var that we would always set on CI, like `EMACS-LSP-CI=1`, so local tests would just work, WDYT?

ericdallo:
> This is 99% done except lsp-dart. I am getting the following error but not able to find the function lsp-make-notification anywhere inside lsp-dart package. This looks very strange to me. confused

@jcs090218 I couldn't find anything related to lsp-dart, but we do have a [Notification protocol not used here](https://github.com/emacs-lsp/lsp-dart/blob/master/lsp-dart-protocol.el#L57) 🤔, @yyoncho declaring unused protocols would make this warn? this already works on lsp-dart CI without issues

yyoncho:
> This is 99% done except lsp-dart. I am getting the following error but not able to find the function lsp-make-notification anywhere inside lsp-dart package. This looks very strange to me. confused

@jcs090218 can you link the error?

test/windows-bootstrap.el
jcs090218: Good idea. Added variable `emacs-lsp-ci`.

jcs090218:
Error is here, https://github.com/emacs-lsp/lsp-mode/runs/2046252831.

test/test-packages.el
```diff
@@ -0,0 +1,75 @@
...
+;;  CI test for emacs-lsp packages
+;;
+;; See https://github.com/emacs-lsp/lsp-mode/issues/2655 for more information.
+;;
+;;; Code:
+
+(require 'package)
+
+(setq user-emacs-directory (expand-file-name (make-temp-name ".emacs.d")
+                                             "~")
```
ericdallo, line 29: Does that works for Windows?

jcs090218:
> @jcs090218 I couldn't find anything related to lsp-dart, but we do have a Notification protocol not used here 🤔,

Yeah, this looks strange to me. But I sometime do get this warning. See if I can reproduce this. 😕

test/test-packages.el
jcs090218: Yeah, I think.

yyoncho:
> Error is here, https://github.com/emacs-lsp/lsp-mode/runs/2046252831.

The issue is that we have 2 lsp-make-notification functions, one in lsp-mode.el and one in lsp-protocol.el, we should remove the one in lsp-mode.el

jcs090218:
> The issue is that we have 2 lsp-make-notification functions, one in lsp-mode.el and one in lsp-protocol.el, we should remove the one in lsp-mode.el

I could not find the second `lsp-make-notification` in lsp-protocol.el. Is the line generates it's function? 

https://github.com/emacs-lsp/lsp-mode/blob/800591f724c87da74a63d1317d45e92cfae019a1/lsp-protocol.el#L110

I think I should delete the one in lsp-mode?

ericdallo:
@jcs090218 I think @yyoncho meant https://github.com/emacs-lsp/lsp-dart/blob/master/lsp-dart-protocol.el#L57

yyoncho:
> @jcs090218 I think @yyoncho meant https://github.com/emacs-lsp/lsp-dart/blob/master/lsp-dart-protocol.el#L57
> 
> I think we can rename the function on lsp-mode or rename on dart side, I don't mean to remove that on Dart side as is not being used ATM, I just added to follow the spec

Lets remove the dart side.

yyoncho:
But generally, we should prefix the non-core interfaces.

ericdallo:
I'll remove it and later add the prefix, it makes sense @yyoncho

ericdallo:
Done on master @jcs090218, it should be available on MELPA in 1hour

Makefile
```diff
@@ -50,11 +53,19 @@ prepare_cpp_project:
...
+		-l $(WIN-BOOTSTRAP) \
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
 		-f batch-byte-compile $(LSP-FILES)
 
+test-pkgs:
+	@echo "Test emacs-lsp packages..."
+	@$(CASK) $(EMACS) -Q --batch \
+		-l $(WIN-BOOTSTRAP) \
+		--eval '(setq byte-compile-error-on-warn t)' \
```
yyoncho, line 65: can we move that in the test-package.el? Can we structure the code like that:

1. Install lsp-* package dependencies(e. g. lsp-treemacs depends on treemacs)
2. After that set byte-compile-error-on-warn to t and install our packages.

jcs090218: You mean by setting `byte-compile-error-on-warn` to `t` only for emacs-lsp packages?

yyoncho: yes. I don't want the CI to be red if some of the deps has byte-compilation issue.

jcs090218: Any suggestion achieving this? Not sure if there is a clean way to do this. 😕

test/test-packages.el
```diff
@@ -34,6 +34,12 @@
   "Get version of the package by NAME."
   (let ((pkg (cadr (assq name package-alist)))) (when pkg (package-desc-version pkg))))
 
+(defun test-packages--install-from-archive (pkg-desc)
+  "Advice execute beofre function `package-install-from-archive'."
+  (setq byte-compile-error-on-warn
+        (ignore-errors (string-prefix-p "lsp-" (symbol-name (package-desc-name pkg-desc))))))
```
jcs090218, line 40: Does this works for you? Or do you want to compare it with a emacs-lsp package list?

Makefile
```diff
@@ -50,11 +53,18 @@ prepare_cpp_project:
 windows-compile:
 	@echo "Compiling..."
 	@$(CASK) $(EMACS) -Q --batch \
-		-l test/windows-bootstrap.el \
+		-l $(WIN-BOOTSTRAP) \
 		-L . -L clients \
 		--eval '(setq byte-compile-error-on-warn t)' \
 		-f batch-byte-compile $(LSP-FILES)
 
+test-pkgs:
```
ericdallo, line 61: I would suggest something like to be more clear, but it's your choice
```suggestion
test-downstream-pkgs
```

jcs090218: Changed! 👍

test/test-packages.el
ericdallo: That won't work for `helm-lsp` :/

ericdallo: I can't see a easy way without adding a list of all known downstreams

jcs090218: I have changed to 

```el
(advice-add
   'package-install-from-archive
   :before (lambda (pkg-desc)
             (setq byte-compile-error-on-warn
                   (ignore-errors (memq (package-desc-name pkg-desc) pkgs)))))
```

test/test-packages.el
```diff
@@ -0,0 +1,81 @@
...
+               lsp-python-ms
+               lsp-sonarlint
+               lsp-sourcekit
+               lsp-treemacs
+               lsp-ui)))
+  (advice-add
+   'package-install-from-archive
+   :before (lambda (pkg-desc)
+             (setq byte-compile-error-on-warn
+                   (ignore-errors (memq (package-desc-name pkg-desc) pkgs)))))
```
jcs090218, line 60: @ericdallo How about this?

test/test-packages.el
```diff
@@ -0,0 +1,81 @@
...
+               lsp-python-ms
+               lsp-sonarlint
+               lsp-sourcekit
+               lsp-treemacs
+               lsp-ui)))
+  (advice-add
+   'package-install-from-archive
+   :before (lambda (pkg-desc)
+             (setq byte-compile-error-on-warn
+                   (if (ignore-errors (memq (package-desc-name pkg-desc) pkgs)) t nil))))
```
jcs090218, line 60: @ericdallo How about this? Just added the if statement to ensure the value is either `t` or `nil`.

ericdallo: I think that works :)

Makefile
yyoncho: My idea was to have 2 loops of package-install, first will install the deps of the emacs lsp packages, the second will install the actual packages. But I am fine with anything that covers the initial requirements.

test/test-packages.el
jcs090218: @yyoncho I have applied it this way.

jcs090218:
All checks have passed! 🎉 Let me know if I still need to change anything! ;)

ericdallo:
LGTM, thank you for that @jcs090218 :)

test/test-packages.el
```diff
@@ -0,0 +1,81 @@
...
+      package-user-dir (expand-file-name (make-temp-name "tmp-elpa")
+                                         user-emacs-directory))
+
+(defun package-version (name)
+  "Get version of the package by NAME."
+  (let ((pkg (cadr (assq name package-alist)))) (when pkg (package-desc-version pkg))))
+
+(let* ((package-archives '(("melpa" . "https://melpa.org/packages/")
+                           ("gnu" . "https://elpa.gnu.org/packages/")))
+       (pkgs '(dap-mode
```
yyoncho, line 39: I wonder whether we should add lsp-grammarly, ccls, lsp-tailwindcss to this pipeline, WDYT?

yyoncho (approved):

yyoncho:
Looks good to me! I think this is a great addition to our infrastructure. Now it will be much easier to perform cleanup.

test/test-packages.el
ericdallo: I think we should

jcs090218: I have no objection for this. Added! :)

jcs090218:
All checks have passed (again)! Let me merge this for now before this PR get larger and larger. I would apply another patch if needed! :)

Thanks for everyone who reviewed this PR. 🎉

PR #2688, opened 2021-03-07
Run GitHub Actions with fail-fast false

conao3:
Hi, Thanks for maintaining such a great project!

And I'm sorry for your inconvenience on my Cask issue.
I got an issue from @jcs090218 and see your [CI log](https://github.com/emacs-lsp/lsp-mode/runs/2046150488).

I suggest put strategy#fail-fast as `false`. This option changes
the default behavior of canceling other jobs when a job fails, so
that each CI job completes to the end without canceling.

This option allows you to see the results of jobs that work
without being disturbed by jobs that randomly fail due to Cask
factors.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -6,6 +6,7 @@ jobs:
     unix-test:
         runs-on: ${{ matrix.os }}
         strategy:
+            fail-fast: false
             matrix:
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
LGTM

jcs090218 (approved):

PR #2689, opened 2021-03-07
Remove unsafe dir-locals config from .dir-locals.el

conao3:
CC: @leungbk
Remove unsafe dir-locals config from .dir-locals.el

Hi, thanks for develop/maintaining such a great project!

I found your .dir-locals include unsafe config.
This `eval` feature has a LARGE security concern.

The contributor Emacs will always execute this S-expression just
by opening the file.  The user may set the warning to be
permanently hidden in order to hide the warning window.  What
happens if this dir-locals is mixed with code you don't intend?

It is very easy to introduce an S-expression that causes a
strange server to execute code that scp's the user's `~/.ssh` folder.
If you need such a configuration, I think the right way is to
mention it in CONTRIBUTING.md.


conao3:
New contributors see this warning window, but it is very difficult to be sure that this Sexp is **absolutely** safe!
If he is going to continuously contribute, he must first make sure that this `dir-locals.el` is not changed.
This is really a security issue, it creates an unnecessarily dangerous situation.

Screenshot_2021-03-07_21-20-22

nbfalcon:
@conao3 to check whether the sexp doesn't change is simple: just press `!`, and the current sexp will be remembered. As for the safety, I would expect a contributor to somehow test the code, which means that they have to actually use `lsp-mode`. However, removing this doesn't increase security: if you don't trust scripts in `*lsp-mode*`, why trust the package itself? Why wouldn't it be malicious?

conao3:
Of course, I know how dangerous it is to install an Emacs package. I know how dangerous it is to open unknown files.
But the fact that it is dangerous to install a package is no reason to accept other dangers.
That is called the broken window theory in Japan.
What do we gain by using such an inherently problematic method?
If there is an S-expression that you simply want the contributor to execute, you should mention it in CONTRIBUTE.md, not have it executed automatically using `.dir-locals.el`.

nbfalcon:
@conao3 the s-expression is not pointless: it sets up proper indentation for `lsp-mode` and makes functions defined using `lsp-defun` show up in `imenu`. There is nothing malicious about it, which is quite easy to prove since it doesn't open/read files, and it does nothing for non-`lsp-mode` files. No one can just sneak in malicious code into the .dir-locals file either, since we have code review.

nbfalcon:
Anyway, if you are not comfortable with running this code, you can just press `n` at the prompt and no code at all will be executed.

conao3:
I understand that this code is not dangerous. But I don't have a problem with the code, I'm asking you not to use such an inherently dangerous feature.

And maybe there is another issue that is frustrating me with this setup.
Start with `emacs -q`, open some lsp-mode file, I get warning window and reject `.dir-locals.el` unsafe config with `n`. Then open another file. I will get another warning!
The safe use of dir-locals is not a problem (And my suggestion is to put .dir-locals in that state).

nbfalcon:
@conao3 using dir-locals is the only way to achieve what we need to. If we used a contributing section, people would have to copy that code to their own .dir-locals2.el, which increases the friction.

As for your issue, you can simple disable `.dir-locals.el` globally, using `(setq enable-dir-locals nil)`.

conao3:
Why I disable the whole dir-locals feature because of only your unsafe file?
BTW, I cannot find `enable-dir-locals` in Emacs-27.1 source tree.

nbfalcon:
Sorry, I meant `enable-dir-local-variables`; it is sensible for a project to want to configure indentation, etc. so IMHO lsp-mode does nothing wrong here. You can disable this for only lsp-mode like this:
```emacs-lisp
(defun +hack-dir-local-variables-non-lsp-a ()
  (string-match-p "lsp-mode" default-directory))
(advice-add #'hack-dir-local-variables :before-until #'+hack-dir-local-variables-non-lsp-a)
```
I think this should be fixed in core-emacs, there should be a denylist for disallowed `.dir-locals`.

nbfalcon:
Digging even further, `(add-to-list 'ignored-local-variables 'eval)` would disable `(eval . )` dir-locals globally, which seems to be more what you want.

conao3:
Thanks. Then when I write a patch with that setting, it will be different from the indentation style of the whole file, how will that patch be handled?
What indentation style are you trying to enforce by using such a dangerous feature in the first place? Does it need to use this dangerous feature?

nbfalcon:
@conao3 it just uses the Emacs 28 indentation function, which might have some cleaner semantics wrt. plists:
```emacs-lisp
(:abc
 t)
;; vs
(:abc
   t) ; indented
```
I think DOOM-Emacs does this by default, but this setting hasn't done any harm. I think you should try to stick to lsp-mode conventions where possible (the former in my example). If you don't use `aggressive-indent-mode`, this shouldn't be too hard even without that advice.

yyoncho:
I tend to agree with @conao3 that it is a bit scary to see the warning when you open lsp-mode file. As an alternative, we may create a file lsp-dev.el (or similar) under scripts folder and create a minor mode lsp-elisp-indent-buffer-mode and mention it in the contributing section. If the new contributor does not follow the indentation guide, we will reformat after they commit the code. 

As a side note, with the introduction of lsp-protocol.el we have fewer plists, so the impact of removing the formatting rules will be not that big.

conao3:
@nbfalcon, Thanks. If that's the case, it might be better to encourage contributors to use Emacs-28 to develop lsp-mode rather than having them run Elisp with dir-locals.
If you want to further enforce the indentation style, you can use GitHub Actions to indent files changed in Emacs-28 and drop tests if there is a diff.

@yyoncho, Thanks for your comment. I think that's a good idea too.
(I know your idea is just a suggestion and doesn't mean much) But I feel like letting contributors run that minor mode is just moving the problem around.

nbfalcon:
@conao3 I don't think it is possible for everyone to use Emacs 28, since some distributions don't ship it (it isn't released yet, AFAIK) so users would need to build from source. It isn't in Ubuntu 20.04, for example.

conao3:
I know I'm *conservative* enough to get Cask to work with Emacs-24, but this recommendation for lsp-mode *contributors*.
it's not a bad idea we assume he is an Emacs hacker and a reasonable knowledge of Linux and Emacs builds.

yyoncho:
> But I feel like letting contributors run that minor mode is just moving the problem around.

From my POV the problem is that there is a warning huge text when you clone and open a file from lsp-mode repo. Contributors executing the script from lsp-mode's directory is not an issue, we already have a set of scripts for different tasks(e. g. generating defcustoms from package.json). Furthermore, these scripts will be used mostly by regular contributors. I doubt that occasional contributors will care about lsp-defuns not showing in imenu.

conao3:
Thanks, @yyoncho. If it's an optional collection of tips for maintainers, rather than demand on contributors, then I agree.

nbfalcon:
@yyoncho In my opinion, the `eval` block is useful and should stay, but it is ultimately your decision (I won't mind). If it is to be removed, we should keep the eval block in another optional file so that users can symlink it to .dir-locals2.el. There could be a script for that.

nbfalcon:
Core Emacs really should have a way to say "no and don't ask again for this directory" in the prompt.

leungbk:
I’m fine with moving the eval to a lsp-devel-mode.el.

I am on phone right now, but will PR later today.

leungbk (approved):

leungbk:
Thanks!

yyoncho:
> @yyoncho In my opinion, the `eval` block is useful and should stay, but it is ultimately your decision (I won't mind). Since I think DOOM sets up indentation that way anyway, I'd just loose imenu, which in any case is just a very minor inconvenice.
> 
> If it is to be removed, we might want to keep the eval block in another optional file so that users can symlink it to .dir-locals2.el. There could be a script for that.

I am proposing to keep the functionality but enable it in a different way. All this for the sake of making lsp-mode more contribution-friendly. The only thing that we will lose is that we will get poorly formatted PR from time to time which we could fix after the merge or on a regular basis(CI reformatting the code is an option too). But even without this setting, we are getting poorly formatted patches either way...

ericdallo:
I'm kind of late here but I totally agree with @conao3, This warning is something that always bothered me when developing lsp-mode, imagine new users with the same issue.
If the issue is an indentation problem, we should totally move that to a CI IMO, we could format the code after the commit or after the merge, fixing our indentation problem :)

kiennq:
One thing that's also got deleted with this PR is
``` elisp
  (flycheck-disabled-checkers . '(emacs-lisp-checkdoc))
```

Just to make sure if we want to remove that together with this

ericdallo:
I always have emacs-lisp-checkdoc enable for any package, I don't know why we would add to disable that @kiennq

conao3:
Changing that *unsafe* variable using `.dir-locals` make that warning, the deletion is intended.

nbfalcon:
@ericdallo it generates a lot of warnings for `lsp-mode`, since it wasn't completely written with checkdoc in mind. Many symbols are undocumented, or not properly (by checkdoc standards).

conao3:
I think we should document along the lines of checkdoc's point.
Turning off just because the warnings are annoying is not a good choice.

nbfalcon:
@conao3 I tend to agree, but there are too many symbols in `lsp-mode` for this to be feasible. I am not against this change though.

yyoncho:
If someone wants to fix the checkdoc warnings - that is fine. But until then, having checkdoc on just makes flycheck unusable. FWIW even some core packages like company-mode do not pass checkdoc.

conao3:
I think `dir-locals-2.el` is the place if you want to disable checkdoc.

But follow checkdoc is [mandatory for MELPA](https://github.com/melpa/melpa/blob/master/CONTRIBUTING.org#making-your-package-ready-for-inclusion).
see https://github.com/melpa/melpa/blob/master/.github/PULL_REQUEST_TEMPLATE.md#checklist
> #### `M-x checkdoc` is happy with my docstrings

PR #2693, opened 2021-03-07
git-commit-major-mode is not safe buffer-local variable

conao3:
CC: @leungbk
This is remained work on #2689.
`git-commit-major-mode` is also not safe buffer-local variable.

Sorry, this variable for `git-commit-mode`, I couldnot catch it.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,6 +1,4 @@
 ((nil
   (require-final-newline . t))
- (git-commit-mode
-  (git-commit-major-mode . git-commit-elisp-text-mode))
  (emacs-lisp-mode
   (indent-tabs-mode . nil)))


nbfalcon:
I disagree with this one - `git-commit-major-mode` makes `magit` buffers highlight symbols. This variable is used by many projects and the main way to set it.

conao3:
Again I should ask, why set this variable **on every contributor** environment?

nbfalcon:
The point of `.dir-locals` is not to only specify safe variables, but to specify an editor configuration for a given project in general. Using `git-commit-major-mode` here makes all commit messages more uniform. It is unsafe because it could theoretically be used to run anything, but `git-commit-elisp-text-mode` is safe and designed for it.

conao3:
Please tell me *many projects*.

nbfalcon:
Many might not be the right word, but DOOM Emacs does for example. 
> Why set the variable on every contributor environment.
So that contributions can be uniform and follow guidelines.

nbfalcon:
You can accept the current .dir-locals variables by pressing `!`, because they are not unsafe (especially with the removal of `eval`), so that you won't be asked to accept them again.

conao3:
OK, then, I don't know `git-commit-elisp-text-mode`. What does this mode force the contributor to do?

nbfalcon:
@conao3 do you `magit`? It has a companion module called `git-commit` and this variable sets the major-mode used for commit buffers. If you don't use magit, this variable will have no effect whatsoever.

conao3:
Yes, again, `git-commit-elisp-text-mode` gives us what?

nbfalcon:
On

vs off:

nbfalcon:
On

vs off:

conao3:
OK, this gives us just buffer font-lock.
This is obviously a user setting, not something that is enforced by the project.
If this was something that enforced a style like the main Emacs source, automatically inserted changed functions, I would have considered it, but this is not it.

nbfalcon:
No, it doesn't by default, but users are free to set custom keybindings etc. in `git-commit-elisp-text-mode`. For example, I have smartparens to auto-close \` with `'` set-up, which will work with any project using that setting. This cannot be done generically in a way that is elegant.

What is your problem with .dir-locals variables that are not explicitly safe? Is the prompt annoying you? Why can't you accept them using `!`?

nbfalcon:
Note that other projects, for example `projectile`, also use dir-locals that are not `safe-local-variable`s.

leungbk:
I didn't mind removing the giant `eval` block since I think it's reasonable for someone to be taken aback by, and skeptical of, a giant blob of arbitrary code. But customizing `git-commit-major-mode` in this manner is idiomatic (Magit, Forge, and Ghub all do this in their `.dir-locals`) and does not permit the shenanigans that `eval` might. People are free to skim the Help pages and the source for `git-commit-major-mode` and `git-commit-elisp-text-mode` to confirm what they do, and anyone who doesn't want to modify the variable can press `n` to work without the setting.

tarsius:
I've just added this to `git-commit.el`:

```elisp
+;;;###autoload(put 'git-commit-major-mode 'safe-local-variable
+;;;###autoload     (lambda (val)
+;;;###autoload       (memq val '(text-mode markdown-mode org-mode
+;;;###autoload                   git-commit-elisp-text-mode))))
```

I thought I had already done that a long time ago but turns out I had just manually accepted the setting in my own configuration.

tarsius:
There are probably other modes that would makes sense as well. I'll add them as they are being pointed out to me, as you have just done.

nbfalcon:
@conao3 do you still believe this PR to be necessary, given that `git-commit-major-mode` as set in .dir-locals is safe and as such will no longer produce that warning in the latest magit?

conao3:
No, thanks @tarsius.
One comment, I think below patch is better.
```diff
+;;;###autoload
+(put 'git-commit-major-mode 'safe-local-variable
+     (lambda (val)
+       (memq val '(text-mode markdown-mode org-mode
+                   git-commit-elisp-text-mode))))
```

PR #2694, opened 2021-03-07
Add GitHub Actions Lint job

conao3:
Add GitHub Actions Lint job.
As #2689 discussion, we need some indentation linter.

This job indent using Emacs-28 and check diff.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,38 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: purcell/setup-emacs@master
+        with:
+          version: snapshot
+
+      - uses: actions/setup-python@v2
+        with:
+          python-version: "3.6"
+          architecture: "x64"
+
+      - uses: conao3/setup-cask@master
+        with:
+          version: 0.8.4
+
+      - name: Resolve dependency
+        run: cask install
+
+      - name: Run Indentation Lint
+        run: |
+          git diff-tree --no-commit-id --name-only -r ${{ github.sha }} \
+          | grep \.el | grep -v \.dir-locals\.el \
+          | xargs -t -I^ cask emacs -q --batch ^ --eval "\
+              (let ((require-final-newline t)) \
+                (require 'flycheck) \
+                (eval-buffer) \
+                (untabify (point-min) (point-max)) \
+                (indent-region (point-min) (point-max)) \
+                (save-buffer))" \
+          ; git diff --exit-code


.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
+name: Lint
+
+on: [push, pull_request]
```
ericdallo, line 3: This could bother us every commit, WDYT about just running this when a PR is merged or a commit is made on master?
This way we'll do that as a cleanup only on the last commit of the pr (the merge)

conao3: I think no. This linter is running for contributors, not maintainers. You, maintainers, are shared properly setting, there is no reason commits with wrong indentation.
However, we need to notice immediately that the maintainer's commits have been mixed in as well.
That's why I think this trigger is appropriate.

ericdallo: This linter will run for pushes on master by maintainers as well, but my concern is more if I have a PR and push a commit, the lint will run, when I try to push another commit, it won't work, I'll need to pull the last commit, that's why I don't like CI committing on PRs until it's finished.

Running it on master pushes or PR merges would work for both contributors and maintainers

kiennq: This job is not pushing anything. It's only check if there is diff with what the auto indentation and current code, and fail if there's different

conao3: >  if I have a PR and push a commit, the lint will run,

yes

> when I try to push another commit, it won't work

but, really?
I think it should run too.

ericdallo: Oh sorry, I thought the CI was pushing, failing when is not linted correctly is enough for me :)

conao3: > This job is not pushing anything. It's only check if there is diff with what the auto indentation and current code, and fail if there's different

Yes. That is CI. You just don't merge PR if any CI fail.

ericdallo: Yes, is just that I already worked on projects that the linter pushes to the branch every commit and I thought this a PR that was doing that, sorry for that

conao3: No need to say sorry. Let me know if it doesn't work as I say :)

kiennq (approved):

ericdallo (approved):

.github/workflows/lint.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          git diff-tree --no-commit-id --name-only -r ${{ github.sha }} \
+          | grep \.el | grep -v \.dir-locals\.el \
+          | xargs -t -I^ cask emacs -q --batch ^ --eval "\
+              (let ((require-final-newline t)) \
+                (require 'flycheck) \
+                (eval-buffer) \
+                (untabify (point-min) (point-max)) \
+                (indent-region (point-min) (point-max)) \
+                (save-buffer))" \
```
ericdallo, line 37: I think we could move that to our Makefile, then we could lint locally, and CI just call the makefile like the other CI workflows, WDYT?

ericdallo: I don't get why it needs to be specified on cask files, I thought of just moving that piece of code to makefile and calling it from here, just like other makefile tasks we already do that

conao3: because current shell script depends `${{ github.sha }}` placeholder.

ericdallo: Hum, got it
Ok, we can keep it here for now I think

yyoncho (approved):

yyoncho:
Thank you!

PR #2696, opened 2021-03-08
Use inclusive comparison to construct a document-symbol hierarchy

leungbk:
The previous way of doing it disregarded any current-position that did
not satisfy a_start < b_start, b_end < a_end. The current comparison
is more like a_start <= b_start, b_end <= b_end.

This fixes the following headerline bug:

output-2021-03-07-15:41:33

When on the first character of a heading, the headerline does not include that heading entry, and only when we move to the second character do we see the headerline include that heading.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5480,11 +5480,8 @@ perform the request synchronously."
 (defun lsp--document-symbols->document-symbols-hierarchy (document-symbols current-position)
   "Convert DOCUMENT-SYMBOLS to symbols hierarchy on CURRENT-POSITION."
   (-let (((symbol &as &DocumentSymbol? :children?)
-          (seq-some (-lambda ((symbol &as &DocumentSymbol :range (&Range :start start-position
-                                                                         :end end-position)))
-                      (when (and (lsp--position-compare current-position start-position)
-                                 (lsp--position-compare end-position current-position))
-                        symbol))
+          (seq-find (-lambda ((&DocumentSymbol :range))
+                      (lsp-point-in-range? current-position range))
                     document-symbols)))
     (if children?
         (cons symbol (lsp--document-symbols->document-symbols-hierarchy children? current-position))


ericdallo (approved):
LGTM

leungbk:
OK, either I'm completely wrong in this PR or every server I tested (basically just `pyright` and `yaml`) is failing to implement the protocol correctly.

The protocol says

``` typescript
export interface DocumentSymbol {
        /**
         [...]
        */

	/**
	 * The range enclosing this symbol not including leading/trailing whitespace
	 * but everything else like comments. This information is typically used to
	 * determine if the clients cursor is inside the symbol to reveal in the
	 * symbol in the UI.
	 */
	range: Range;


}
```

Going off this, you would never expect a set of pairs whose ranges look like `([)]` where `( )` and `[ ]` are both top-level pairs and the `[` position is `<=` the `)` position . But this is exactly what I observe with the following Python file in Pyright:

``` python
def hi():
    return 42


def yo():
    return 42
```

The ranges are

``` elisp
(
 (:name "hi" :kind 12 :range (:START (:line 0 :character 0) :END (:line 4 :character 1)) :selectionRange (:start (:line 0 :character 4) :end (:line 0 :character 6)) :children #1=[])

 (:name "yo" :kind 12 :range (:START (:line 4 :character 0) :END (:line 5 :character 14)) :selectionRange (:start (:line 4 :character 4) :end (:line 4 :character 6)) :children #1#))
```

Something similar happens with the above `yaml` file (our `.github/workflows/labeler.yml`). In this case, we have `([)]` where the `[` position equals the `)` position, allowing either one of two symbols to be returned from the `seq-find`.

The patch here seems consistent with the requirements of the protocol, while the results I see from the servers do not seem to meet the demands of the protocol. The protocol specifically mentions that leading/trailing whitespace is not included in the `ranges`, yet in the `yaml` file above, the `range` for the `on` heading extends past the trailing whitespace and into the `jobs` heading. Similarly for the Pyright example. Am I wrong here?

leungbk:
Haskell seems to return agreeable `ranges`:

2021-03-07-174228_1920x1080_scrot

leungbk:
OK, the Pyright maintainer says he fixed the DocumentSymbol-reporting in the latest Pyright, so I think we should be OK here.

PR #2703, opened 2021-03-11
Remove go-bingo language server from documentation

rawenous:
Go bingo docs are removed but it was still visible in the language server list https://emacs-lsp.github.io/lsp-mode/page/languages/. It is now removed.


yyoncho (approved):

yyoncho:
Thank you!

PR #2708, opened 2021-03-12
Fix angular-ls starting exited error when using nvm

ralph-cheng:
when using [nvm](https://github.com/nvm-sh/nvm) managed NodeJS installation, angular-ls cannot
start because the node_modules directory is located some where
under user home, instead of under the system location `/usr/lib'.

### DIFF
--- a/clients/lsp-angular.el
+++ b/clients/lsp-angular.el
@@ -36,13 +36,15 @@
   :link '(url-link "https://github.com/angular/vscode-ng-language-service"))
 
 (defcustom lsp-clients-angular-language-server-command
-  '("node"
-    "/usr/lib/node_modules/@angular/language-server"
-    "--ngProbeLocations"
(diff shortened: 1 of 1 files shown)


clients/lsp-angular.el
```diff
@@ -36,13 +36,15 @@
   :link '(url-link "https://github.com/angular/vscode-ng-language-service"))
 
 (defcustom lsp-clients-angular-language-server-command
-  '("node"
```
yyoncho, line 39: We should not run programs as part of the initialization of a properly since it will run for all users using lsp-mode. The option is to move the calculation in a function so it will run only when the user is actually in an angular project but IMHO better to create an automatic installation for that server.

yyoncho (approved):

yyoncho:

PR #2711, opened 2021-03-13
[3.16] Insert replace support

### LINKED ISSUE #2032: [3.16] [Completion]  Support insertReplaceSupport
https://github.com/microsoft/language-server-protocol/issues/846
and 
https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/specification-3-16.md#completion-request-leftwards_arrow_with_hook

yyoncho:
Fixes #2032

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,7 @@
 * Changelog
 ** Release 7.1
+  * Add support for insert/replace when performing completion. When using
+    ~company~ pressing ~C-u RET~ will perform replace.
   * Add ~lsp-auto-save-changed-buffers~ which save the changed buffers after any text edit operation by lsp-mode.
   * Safe renamed ~lsp-diagnose~ to ~lsp-doctor~.
   * Add ~lsp-modeline-code-actions-segments~ for a better customization.
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -186,6 +186,14 @@ KEEP-LAST-RESULT if specified."
   (setq lsp-completion--cache nil)
   (unless keep-last-result (setq lsp-completion--last-result nil)))
 
+(defcustom lsp-completion-default-behaviour :replace
+  "Default behaviour of `InsertReplaceEdit'."
+  :type '(choice
+          (const :insert :tag "Default completion inserts")
+          (const :replace :tag "Default completion replaces"))
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "7.1"))
+
 (lsp-defun lsp-completion--guess-prefix ((item &as &CompletionItem :text-edit?))
   "Guess ITEM's prefix start point according to following heuristics:
 - If `textEdit' exists, use insertion range start as prefix start point.
@@ -194,6 +202,8 @@ KEEP-LAST-RESULT if specified."
   - The character before prefix is not word constitute
 Return `nil' when fails to guess prefix."
   (cond
+   ((lsp-insert-replace-edit? text-edit?)
+    (lsp--position-to-point (lsp:range-start (lsp:insert-replace-edit-insert text-edit?))))
    (text-edit?
     (lsp--position-to-point (lsp:range-start (lsp:text-edit-range text-edit?))))
    (t
@@ -519,7 +529,16 @@ Others: CANDIDATES"
          (text-edit?
           (apply #'delete-region markers)
           (insert prefix)
-          (lsp--apply-text-edit text-edit?))
+          (pcase text-edit?
+            ((TextEdit) (lsp--apply-text-edit text-edit?))
+            ((InsertReplaceEdit :insert :replace :new-text)
+             (lsp--apply-text-edit
+              (lsp-make-text-edit
+               :new-text new-text
+               :range (if (or (and current-prefix-arg (eq lsp-completion-default-behaviour :replace))
+                              (and (not current-prefix-arg) (eq lsp-completion-default-behaviour :insert)))
+                          insert
+                        replace))))))
          ((or (unless (lsp-falsy? insert-text?) insert-text?) label)
           (apply #'delete-region markers)
           (insert prefix)
--- a/docs/page/keybindings.md
+++ b/docs/page/keybindings.md
@@ -6,45 +6,46 @@ enabled/disabled based on the server functionality. All the commands are
 configured `lsp-command-map` which is bound to `lsp-keymap-prefix`
 (default `s-l`).
 
-| Keybinding | Description                                                                                                                |
-| ---------- | -------------------------------------------------------------------------------------------------------------------------- |
-| `s-l s s`  | Entry point for the server startup.                                                                                        |
-| `s-l s r`  | Restart language server                                                                                                    |
-| `s-l s q`  | Shutdown language server                                                                                                   |
-| `s-l s d`  | Describes current session                                                                                                  |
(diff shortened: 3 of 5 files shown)


ericdallo (approved):

yyoncho:
Demo

insert-replace-gif

yyoncho:
@dgutov, you might be interested as well.

dgutov:
Okay, so insert-replace behavior can be implemented by an individual backend in `post-completion` without any extra help from the framework. That makes sense. Two questions:

* Do we want to keep it that way? I.e. continue the situation where individual backends make that decision. Perhaps based on some backend-specific prefs.
* If `C-u RET` an optimal trigger for that feature? Depending on the odds of user preferring this behavior, it might not be ideal. I've  kind of figured this might be a good default, and RET would do the replacement (the blocker was learning the appropriate bounds of the text to replace, for each backend).

Maybe trying to collect some statistics on this would be a good idea. And/or a user poll, sometime after the feature is released and is widely available.

lsp-completion.el
```diff
@@ -510,7 +520,18 @@ Others: CANDIDATES"
...
           (apply #'delete-region markers)
           (insert prefix)
-          (lsp--apply-text-edit text-edit?))
+          (pcase text-edit?
+            ((TextEdit) (lsp--apply-text-edit text-edit?))
+            ((InsertReplaceEdit :insert :replace :new-text)
+             (lsp--apply-text-edit
+              (lsp-make-text-edit
+               :new-text new-text
+               :range (if (or (and current-prefix-arg
```
kiennq, line 529: Shouldn't this be
``` elisp
(if (or (and current-prefix-arg
             (eq lsp-completion-default-behaviour :replace))
        (and (not current-prefix-arg)
             (eq lsp-completion-default-behaviour :insert)))
    insert
  replace)
```

So without current prefix, it will execute `lsp-completion-default-behavior` but will do otherwise with current prefix.
Currently it will execute `lsp-completion-default-behavior` with `C-u` but do otherwise without current prefix, which is strange to the name and doc of `lsp-completion-default-behavior`

kiennq: Given that there's only `:replace` or `:insert` of `lsp-completion-default-behavior` and also binary state of `current-prefix-arg`. You can probably just use `xor` here too (nit)
``` elisp
if (xor current-prefix-arg (eq lsp-completion-default-behavior :insert)) insert replace)
```

yyoncho: Yes, will do. I will change the default as well. Seems like replace is closer to what we have now.

yyoncho:
> * Do we want to keep it that way? I.e. continue the situation where individual backends make that decision. Perhaps based on some backend-specific prefs.

I am fine with other proposals as long as the functionality is exposed to the users.

> * If `C-u RET` an optimal trigger for that feature? 

Thinking more about that, the best option from lsp-mode point of view is the vscode shortcut for that which most likely is `C-RET`.

lsp-completion.el
yyoncho: xor is added in emacs 27.1...

kiennq (approved):

CHANGELOG.org
```diff
@@ -1,5 +1,7 @@
 * Changelog
 ** Release 7.1
+  * Add support for insert/replace when performing completion. When using
+    ~company~ pressing ~C-u RET~ will perform replace.
```
michaelpj, line 4:
```suggestion
  * Add support for insert/replace when performing completion. When using
    ~company~ pressing ~C-u RET~ will perform the alternative action (the default is replace, but can be changed).
```

michaelpj:
Add something to the `Keybindings` doc page?

CHANGELOG.org
yyoncho: applied.

dgutov:
@yyoncho 

> I am fine with other proposals as long as the functionality is exposed to the users.

I was thinking I could use the approach that `completion-at-point-functions` takes: change (and probably rename) the `prefix` action to return a cons of positions: the beginning of the prefix and the end of suffix, and then delete the suffix as well in the "insert replace" scenario. That would also enable completion in the middle of the word if the user initiates it explicitly.

But IIUC the approach taken by LSP is 100% incompatible with this: both prefix and suffix bounds could only be known in advance as approximate values, and "insert replace" is defined in terms of server-generated TextEdit. And so it might be really unwise to mandate the API change described above across all backends/. Or do you perhaps see it working somehow?

yyoncho:
> I was thinking I could use the approach that `completion-at-point-functions` takes: change (and probably rename) the `prefix` action to return a cons of positions: the beginning of the prefix and the end of suffix, and then delete the suffix as well in the "insert replace" scenario. That would also enable completion in the middle of the word if the user initiates it explicitly.

In the lsp-mode case, the suffix and prefix are per item so IIUC this design wont work for lsp-mode. 

One potential solution is to have a backend action `insert` which receives the candidate and a second parameter `insert|replace`. And then have `company-complete-selection-insert` and  `company-complete-selection-replace`. This is simple and it will work for lsp-mode but I am not sure whether you want to go in that direction.

dgutov:
> In the lsp-mode case, the suffix and prefix are per item so IIUC this design wont work for lsp-mode.

Well, if that the only problem, suppose we could add a `suffix` backend action instead, which will be called right before a completion is inserted by `company-complete-selection-replace`.

> One potential solution is to have a backend action insert which receives the candidate and a second parameter insert|replace.

Or that. I'll have to think about it, thanks.

The downside of both of these approaches is there'd be no way to disable idle completion when in the middle of a symbol, which I figured would be a nice feature to have.

yyoncho:
> If that's the only problem, suppose we could instead add a `suffix` backend action, which will be called right before a completion is inserted by `company-complete-selection-replace`.

This won't be enough since the insertion performed by company is generally reverted by lsp-mode in post-completion and we apply the actual text edit. Or at least lsp-mode should guess what was the action performed by company based on the line content. 

> The downside of both of these approaches is there'd be no way to disable idle completion when in the middle of a symbol, which I figured would be a nice feature to have.

I am not familiar with what is the issue here.

yyoncho:
@michaelpj 

> Add something to the `Keybindings` doc page?

will do, once we settle what the user interface will be.

yyoncho:
Merged as it is. 

@dgutov once you figure out the company-mode support we will migrate to it.

jscheid:
This is awesome, I'd love to use it but can't seem to get it to work with `GNU Emacs 28.0.50`, `lsp-mode 20210521.446`, `company 20210522.1921` and `typescript-language-server 0.5.1` (`TS Server 4.2.4`).

I've added debug output in `lsp-completion.el` and it seems `text-edit?` is always nil and so the new code never gets run.

I'll try with Emacs 27, but in the meantime, do you happen to know if there is anything else I need to install, upgrade, or configure?

yyoncho:
@jscheid - I dont think that this is available for that language server given their development speed. You should open a feature request against the language server but IMHO realistically it will be fixed once https://github.com/microsoft/TypeScript/issues/39459 is in.

jscheid:
@yyoncho I see... thanks for the pointer. Hmm, that looks to be a ways off still.

As a workaround, is there a way to disable completion inside identifiers, as opposed to at the end of the identifier?

yyoncho:
@jscheid you mean something like disabling completion if there is a character after the cursor?

jscheid:
Yes, if the cursor isn't at a word boundary `(unless (looking-at "\\_>"))`

PR #2713, opened 2021-03-15
Fix CI and check with Emacs minimum requirement

jcs090218:


jcs090218:
@ericdallo The PR labeler kept removing the CI label. Is there a way so it doesn't do this? 😕

ericdallo:
Oh, really odd, it seems you manually added it so the bot it's removing :facepalm: 
I don't know how to fix that for this pr :/

ericdallo:
We could open an issue on the pr labeler GH Action

jcs090218:
It’s okay. This isn’t a big deal for now. Can anyone review this PR? This should be ready to go! :)

yyoncho (approved):
Thank you!

PR #2719, opened 2021-03-16
Copy-editing for CCP-guide.md

cgay:
There were various problems that looked like paste-o's or something like that, where old docs seemed to be left in place when new doc was written. Plus a few other edits. 

I am just getting started writing an LSP server and trying to figure out `lsp-mode`, so take my edits with a grain of salt. If you want to reject this PR and make better edits yourselves, at least it can point out a few issues with the current doc. Thanks.


jcs090218 (approved):

jcs090218:
LGTM! Thanks for contributing to `lsp-mode`! :)

PR #2720, opened 2021-03-17
Avoid incorrect package content

jcs090218:


jcs090218:
This should avoid issue like, https://github.com/emacs-lsp/lsp-mode/runs/2119968336.

ericdallo (approved):

PR #2721, opened 2021-03-17
Add support for dlang (serve-d)

cnasc:
This PR adds bare-bones support for [the D programming language](https://dlang.org/) using [the serve-d language server](https://github.com/Pure-D/serve-d).

image


yyoncho:
The issue you are seeing is most likely caused by lsp-d.el file not being loaded. You have to add it to lsp-client-packages to have it loaded automatically.

cnasc:
Indeed that did it! Thank you for the tip

yyoncho (approved):

PR #2723, opened 2021-03-17
Add support for Zig (zls)

rbino:
This PR adds support for the [Zig programming language](https://ziglang.org/) using the [zls](https://github.com/zigtools/zls) language server.

### DIFF
--- a/clients/lsp-zig.el
+++ b/clients/lsp-zig.el
@@ -1,0 +1,36 @@
+;;; lsp-zig.el --- lsp-mode Zig integration -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2021 Riccardo Binetti
+
+;; Author: Riccardo Binetti <rbino@gmx.com>
+;; Keywords: languages,tools
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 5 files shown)


kiennq (approved):

clients/lsp-zig.el
```diff
@@ -0,0 +1,36 @@
...
+;;  client for zls, the Zig language server
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection "zls")
+  :major-modes '(zig-mode)
```
kiennq, line 32: We're in transition from `:major-mode` to better `lsp-activate-on`, so you may want to change this like as 
```suggestion
  :activation-fn (lsp-activate-on "zig")
```

kiennq: @yyoncho Should we mention this in the docs when creating new client?

rbino: @kiennq thanks for the heads-up, fixed it

yyoncho: @kiennq yes, we should.

yyoncho (approved):

yyoncho:

PR #2724, opened 2021-03-18
lsp-zig: allow zls executable path to be customized

bnjmnt4n:
zls is still in active development, and no up-to-date binary releases are maintained. Hence, building zls is a common occurrence and being able to customize the path to the zls executable would be beneficial.

I've tried to follow similar conventions as other clients in the codebase.

### DIFF
--- a/clients/lsp-zig.el
+++ b/clients/lsp-zig.el
@@ -26,9 +26,21 @@
 
 (require 'lsp-mode)
 
+(defgroup lsp-zig nil
+  "LSP support for Zig via zls."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/zigtools/zls"))
+
+(defcustom lsp-clients-zls-executable "zls"
+  "The zls executable to use.
+Leave as just the executable name to use the default behavior of
+finding the executable with variable `exec-path'."
+  :group 'lsp-zig
+  :type 'string)
+
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection "zls")
+  :new-connection (lsp-stdio-connection lsp-clients-zls-executable)
   :activation-fn (lsp-activate-on "zig")
   :server-id 'zls))
 


yyoncho (approved):

yyoncho:
FYI @rbino

clients/lsp-zig.el
```diff
@@ -26,9 +26,21 @@
...
+  "The zls executable to use.
+Leave as just the executable name to use the default behavior of
+finding the executable with variable `exec-path'."
+  :group 'lsp-zig
+  :type 'string)
+
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection "zls")
+  :new-connection (lsp-stdio-connection lsp-clients-zls-executable)
```
kiennq, line 43: you should create a lamba so you customize `lsp-clients-zls-executable` even after `lsp-zig` has been loaded
```suggestion
  :new-connection (lsp-stdio-connection (lambda () lsp-clients-zls-executable))
```

clients/lsp-zig.el
```diff
@@ -26,9 +26,21 @@
 
 (require 'lsp-mode)
 
+(defgroup lsp-zig nil
+  "LSP support for Zig via zls."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/zigtools/zls"))
+
+(defcustom lsp-clients-zls-executable "zls"
```
kiennq, line 34: This should be named something like with `lsp-zig` prefix
```suggestion
(defcustom lsp-zig-zls-executable "zls"
```

clients/lsp-zig.el
bnjmnt4n: Sure, i can send another PR for this.

clients/lsp-zig.el
bnjmnt4n: I was actually following some of the other LSP clients in following the `lsp-clients-` prefix naming scheme. I can also update this in a PR if desired.

kiennq: Yeah, the `lsp-clients-` prefix is from when we have a gigantic `lsp-clients.el` file and group.
Now it's separated into several smaller one, this should be prefixed as group name

PR #2726, opened 2021-03-19
[diagnostics] Use a lighter gray on unnecessary face

ericdallo:
The current default unnecessary face is too dark, this PR change to use a lighter gray instead of one for the darkest grays.

Before:
image
After:
image

### DIFF
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -53,7 +53,7 @@
   :group 'lsp-mode)
 
 (defcustom lsp-diagnostics-attributes
-  `((unnecessary :foreground "dim gray")
+  `((unnecessary :foreground "gray")
     (deprecated  :strike-through t))
   "The Attributes used on the diagnostics.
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

yyoncho:
how does it look like on the default theme?

ericdallo:
Good question @yyoncho, for a better interop with all themes, we would need to just make the current face dim, but I have no idea how to do that in elisp and I'm not sure elisp has support for that just like vscode:
image

When the import and class name is dimmed

yyoncho:
I am not sure if we can implement this approach without changing flycheck. I am not even sure if there are relative faces.

ericdallo:
Yeah, me too, I think we can go this way and check if we have any issue with that.

A elisp package that gets the relative dimmed face related to one face would be really nice maybe

ericdallo:
Merging this, I think it will improve UX for most cases, I'll keep an eye on issues if related to that

PR #2729, opened 2021-03-19
Renamed ra-cargo-load-out-dirs-from-check to -run-build-scripts

teozkr:
See https://github.com/rust-analyzer/rust-analyzer/pull/7868

Also turned it on by default to align with upstream (https://github.com/rust-analyzer/rust-analyzer/pull/7914).

The old `loadOutDirsFromCheck` is also still sent, since Rustup's Rust-Analyzer tends to lag behind a bit.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -26,6 +26,7 @@
   * Add Grammarly support.
   * Add D support.
   * Add Zig support.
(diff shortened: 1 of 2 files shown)


ericdallo (approved):

yyoncho:
Looks good to me! @flodiebold, any thoughts?

flodiebold:
Hmm. Worth noting that this changes the default even for those older RA versions where the server-side default is still off. Probably this is the best approach nonetheless.

teozkr:
True, but it's not like it breaks anything on older RA versions.

PR #2731, opened 2021-03-19
docs: add new conventions when creating new client

kiennq:
See https://github.com/emacs-lsp/lsp-mode/pull/2723#discussion_r596681029

### DIFF
--- a/docs/page/adding-new-language.md
+++ b/docs/page/adding-new-language.md
@@ -19,7 +19,7 @@ corresponding mode -\> language id - in this case `(python-mode .
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection "pyls")
-                  :major-modes '(python-mode)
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
nice

docs/page/adding-new-language.md
```diff
@@ -19,7 +19,7 @@ corresponding mode -\> language id - in this case `(python-mode .
```
yyoncho, line 19: can you drop somewhere a note, something like: 

`lsp-mode` is using `lsp-language-id-configuration` to determine what is the buffer language. When the major-mode is not sufficient to determine the language (e. g. web-mode is used for javascript and for html) you can put regex.

yyoncho (approved):

docs/page/adding-new-language.md
kiennq: Done

PR #2736, opened 2021-03-21
Fix handling of lsp-file-watch-ignored-[directories|files]

crabbkw:
When either lsp-file-watch-ignored-directories or
lsp-file-watch-ignored-files are buffer local, the correct value may
not be used when a new directory is added to an already watched
directory. This is due to the watch callback invoking at any time. A
different buffer could be the active buffer when the callback is
invoked causing a different value of either of those two variables to
be used.

This was solved by capturing the value of those two variables at the
time the original watch was created and passing those values into the
watch function.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -339,7 +339,8 @@ the server has requested that."
     "[/\\\\]bin/Debug\\'"
     "[/\\\\]obj\\'")
   "List of regexps matching directory paths which won't be monitored when
-creating file watches."
+creating file watches. Customization of this variable is only honored at
+the global level or at a root of an lsp workspace."
   :group 'lsp-mode
   :type '(repeat string)
   :package-version '(lsp-mode . "7.1.0"))
@@ -367,7 +368,10 @@ This setting has no impact on whether a file-watch is created for
 a directory; it merely prevents notifications pertaining to
 matched files from being sent to the server.  To prevent a
 file-watch from being created for a directory, customize
-`lsp-file-watch-ignored-directories'"
+`lsp-file-watch-ignored-directories'
+
+Customization of this variable is only honored at the global
+level or at a root of an lsp workspace."
   :group 'lsp-mode
   :type '(repeat string)
   :package-version '(lsp-mode . "7.1.0"))
@@ -1644,15 +1648,15 @@ This set of allowed chars is enough for hexifying local file paths.")
   (descriptors (make-hash-table :test 'equal))
   root-directory)
 
-(defun lsp--folder-watch-callback (event callback watch)
+(defun lsp--folder-watch-callback (event callback watch ignored-files ignored-directories)
   (let ((file-name (cl-third event))
         (event-type (cl-second event)))
     (cond
      ((and (file-directory-p file-name)
            (equal 'created event-type)
-           (not (lsp--string-match-any (lsp-file-watch-ignored-directories) file-name)))
+           (not (lsp--string-match-any ignored-directories file-name)))
 
-      (lsp-watch-root-folder (file-truename file-name) callback watch)
+      (lsp-watch-root-folder (file-truename file-name) callback ignored-files ignored-directories watch)
 
       ;; process the files that are already present in
       ;; the directory.
@@ -1661,12 +1665,12 @@ This set of allowed chars is enough for hexifying local file paths.")
                      (unless (file-directory-p f)
                        (funcall callback (list nil 'created f)))))))
      ((and (not (file-directory-p file-name))
-           (not (lsp--string-match-any lsp-file-watch-ignored-files file-name))
+           (not (lsp--string-match-any ignored-files file-name))
            (memq event-type '(created deleted changed)))
       (funcall callback event)))))
 
-(defun lsp--directory-files-recursively (dir regexp &optional include-directories)
-  "Copy of `directory-files-recursively' but it skips `lsp-file-watch-ignored-directories'."
+(defun lsp--directory-files-recursively (dir regexp ignored-directories &optional include-directories)
+  "Copy of `directory-files-recursively' but it skips directories matching any regex in IGNORED-DIRECTORIES."
   (let* ((result nil)
          (files nil)
          (dir (directory-file-name dir))
@@ -1677,14 +1681,14 @@ This set of allowed chars is enough for hexifying local file paths.")
                         'string<))
       (unless (member file '("./" "../"))
         (if (and (directory-name-p file)
-                 (not (lsp--string-match-any (lsp-file-watch-ignored-directories) (f-join dir (f-filename file)))))
+                 (not (lsp--string-match-any ignored-directories (f-join dir (f-filename file)))))
             (let* ((leaf (substring file 0 (1- (length file))))
                    (full-file (f-join dir leaf)))
               ;; Don't follow symlinks to other directories.
               (unless (file-symlink-p full-file)
                 (setq result
                       (nconc result (lsp--directory-files-recursively
-                                     full-file regexp include-directories))))
+                                     full-file regexp ignored-directories include-directories))))
               (when (and include-directories
                          (string-match regexp leaf))
                 (setq result (nconc result (list full-file)))))
@@ -1709,12 +1713,15 @@ Do you want to watch all files in %s? "
      (concat "You can configure this warning with the `lsp-enable-file-watchers' "
              "and `lsp-file-watch-threshold' variables"))))
 
-(defun lsp-watch-root-folder (dir callback &optional watch warn-big-repo?)
+(defun lsp-watch-root-folder (dir callback ignored-files ignored-directories &optional watch warn-big-repo?)
   "Create recursive file notification watch in DIR.
 CALLBACK will be called when there are changes in any of
 the monitored files. WATCHES is a hash table directory->file
 notification handle which contains all of the watch that
-already have been created."
+already have been created. Watches will not be created for
+any directory that matches any regex in IGNORED-DIRECTORIES.
+Watches will not be created for any file that matches any
+regex in IGNORED-FILES."
   (let* ((dir (if (f-symlink? dir)
                   (file-truename dir)
                 dir))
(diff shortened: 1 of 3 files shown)


yyoncho:
Won't that cause ignored-files|dirs to be picked from the buffer that is currently active at the time of the startup? This is kind of improving the current situation but IMHO we should try to read the dir locals from the root folder that we are establishing the watches for. For example, the following snippet will extract  the dir local value for `foo` for lsp-mode folder:

```
(buffer-local-value
 'foo
 (find-file-noselect "/home/yyoncho/Sources/lsp/lsp-mode/.dir-locals.el"))
```

crabbkw:
This patch captures the values of ignored-files|dirs for each invocation of `lsp-watch-root-folder`. It is captured for use in callbacks that arise from the current invocation of `lsp-watch-root-folder`.  Each time a new LSP workspace is started a separate invocation to `lsp-watch-root-folder` occurs and would therefore be capturing a potentially different value of ignored-files|dirs.


I note now that the callback `lsp--folder-watch-callback` recurses into `lsp-watch-root-folder` -- so there is still an issue where if the created folder is *not* ignored, then when iterating over all the folders in the new folder we'll use whatever the current value of ignored-files|dirs is, rather than the appropriate value for the project. We may need a version of `lsp-watch-root-folder` that is passed in the values to use for ignored-files|dirs.


Dir locals is only one way that a buffer-local variable might be set, so we should not rely on reparsing that file to discover the correct value of the ignored-files|dirs. In my personal setup I use a find-file-hook to setup buffer locals as I cannot commit my .dir-locals.el to source control and I very often `git clean -xfd` to blow away all untracked files in the repositories I work it.

crabbkw:
An alternative approach would be to copy the buffer-local values on language server start from the initial buffer to the buffer that represents the language server process. Then whenever the values are read they should be read from the context of the language server buffer. This has some appeal to me as the ignored-files|dirs values should not really be buffer-local. They should conceptually be "project-local". It is a bit of a cludge to initialize the project-local values from buffer-local values, but since Emacs does not natively have a concept of "project" this may be a reasonable workaround.

yyoncho:
There are multi-root projects for which we start only the server process. It makes sense to have ignored per workspace folder. The proposal to capture the ignored folders from the initial buffer is an improvement over what we currently have but might lead to confusion. 

> Dir locals is only one way that a buffer-local variable might be set, so we should not rely on reparsing that file to discover the correct value of the ignored-files|dirs. In my personal setup I use a find-file-hook to setup buffer locals as I cannot commit my .dir-locals.el to source control and I very often git clean -xfd to blow away all untracked files in the repositories I work it.

You can configure dir-locals without having an actual .dir-local file using directory class - https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html 

The open question is how to actually obtain the dir local value. The snippet from above will work with any file and most likely there will be a way to ask - give me dir locals for this path.

crabbkw:
Thanks for the tip on directory classes -- I may adapt to using that rather than my current setup. I still maintain that there are multiple ways to set buffer-local variables, so if we support configuring ignored-files|dirs as a buffer-local, then we should support all ways that it could be set. I'm not certain that ignore-files|dirs really ought to be buffer-local. I think it is much closer to "workspace-local", and thus might require some special handling.

For multi-root projects:
1. how does the server process get started?
2. is there support for users specifying a different ignored-files|dirs on a per-root basis?

yyoncho:
> how does the server process get started?

It is started in a file under one of the roots

> 2\. is there support for users specifying a different ignored-files|dirs on a per-root basis?

This is what my proposal is covering - dir local variable for the project root seems to be the best fit to achieve that.

crabbkw:
So then `lsp-watch-root-folder` needs to:
1. identify the workspace root from the path passed in
2. temporarily open a buffer to the workspace root
3. read the buffer-local value of the ignored-files|dirs
4. (optionally) capture that value for use by the callback (if we do not capture it, then the callback has to repeat steps 1-3 above)

We would need to make sure that if a buffer is already open to the workspace root that we do not close it when we're done. Otherwise we must close it.

We need to make sure this temporary buffer work is not done more often than necessary (might require a significant refactor in the way that lsp-watch-root-folder recurses)

We need to make sure that when the lsp--folder-watch-callback invokes lsp-watch-root-folder (or a new method which is pulled out of lsp-watch-root-folder) we pass in the ignored-files|dirs. Alternatively we could simply have lsp-watch-root-folder do all the buffer work again.

We need to document that customization of lsp-watch-file-ignored-directories|files is only supported at the root(s) of a workspace.

Does that sum it up?

yyoncho:
> 1. identify the workspace root from the path passed in

we watch every root for multi root project, check lsp--server-register-capability 

2/3/4 I wish we somehow directly read the value without doing this "hack". I proposed reading .dir-locals, but as you said, it might be non-existent. I will need some time to read the which does that in emacs core. 

> We need to make sure this temporary buffer work is not done more often than necessary (might require a significant refactor in the way that lsp-watch-root-folder recurses)

I think that we need to capture the value in lsp--server-register-capability similar to what you do in this PR.

yyoncho:
Hm, seems like this will work too:

```
(buffer-local-value
 'foo
 (find-file-noselect "/home/yyoncho/Sources/lsp/lsp-mode/"))
```

(the logic about buffer being already open applies)

crabbkw:
Is `lsp-watch-root-folder` a public function? It doesn't have the double dashes that other functions do leading me to expect that it is a supported public function. Is it acceptable to change the signature of this function so that it takes as parameters the lists of regexes of ignored files and directories?

`find-file-noselect` does not close buffers if it was responsible for opening them. We need to use a set of helper functions to detect if the buffer is already open and use it directly or to open, process, then close the buffer. Alternatively we could use a file name that is unlikely to have been used by the user -- say `.lsp-mode-temp-buffer` and then unilaterally close the buffer when we're done. When opening the buffer we need to do so as silently as possible. For example, if a user has an unsafe variable in .dir-locals.el we do not want Emacs to prompt about setting the variable. Instead we want to not set the unsafe variables.

yyoncho:
> Is `lsp-watch-root-folder` a public function? It doesn't have the double dashes that other functions do leading me to expect that it is a supported public function. Is it acceptable to change the signature of this function so that it takes as parameters the lists of regexes of ignored files and directories?

You can add optional param which is fine. For that particular function, it won't be a problem even if you break it. 

> find-file-noselect does not close buffers if it was responsible for opening them. We need to use a set of helper functions to detect if the buffer is already open and use it directly or to open, process, then close the buffer. Alternatively we could use a file name that is unlikely to have been used by the user -- say .lsp-mode-temp-buffer and then unilaterally close the buffer when we're done. When opening the buffer we need to do so as silently as possible. For example, if a user has an unsafe variable in .dir-locals.el we do not want Emacs to prompt about setting the variable. Instead we want to not set the unsafe variables.

yyoncho:
You can use this snippet as well:

```
(with-temp-buffer
  (setq-local buffer-file-name  "/home/yyoncho/Sources/lsp/lsp-mode/")
  (hack-local-variables)
  foo)
```

It solves some of the issues mentioned above. I wouldn't bother if there are risky variables though.

crabbkw:
I've added a commit on top that moves the capturing up to `lsp--server-register-capability`. I've added a helper function to get buffer-local values for the ignored-dirs|files given a workspace root. These values are now passed through several functions (as required parameters) rather than looking them up. I don't think they should be optional parameters as the behaviour of the functions that need these values will be broken unless they are correctly specified. Tests have been updated to account for the fact that methods now have additional parameters.

Based on feedback I get; I'd like to squash this down into a single commit, update the changelog and the issue with the resolution we decided upon before it is merged.

lsp-mode.el
```diff
@@ -3355,17 +3364,37 @@ disappearing, unset all the variables related to it."
                            (ht-keys created-watches))))
       ;; create watch for each root folder without such
       (dolist (folder root-folders)
-        (let ((watch (make-lsp-watch :root-directory folder)))
+        (let* ((watch (make-lsp-watch :root-directory folder))
+               (ignored-things (lsp--get-ignored-regexes-for-workspace-root folder))
+               (ignored-files-regex-list (car ignored-things))
+               (ignored-directories-regex-list (cadr ignored-things))
```
yyoncho, line 3370: nit: close on the same line

lsp-mode.el
```diff
@@ -3355,17 +3364,37 @@ disappearing, unset all the variables related to it."
...
 
   (push
    (make-lsp--registered-capability :id id :method method :options register-options?)
    (lsp--workspace-registered-server-capabilities lsp--cur-workspace)))
 
+(defun lsp--get-ignored-regexes-for-workspace-root (workspace-root)
+  "Return a list of the form (lsp-file-watch-ignored-files lsp-file-watch-ignored-directories) for the given WORKSPACE-ROOT."
+  ;; The intent of this function is to provide per-root workspace-level customization of the
+  ;; lsp-file-watch-ignored-directories and lsp-file-watch-ignored-files variables.
+  (let* (
```
yyoncho, line 3388: Can you use the snippet from https://github.com/emacs-lsp/lsp-mode/pull/2736#issuecomment-803670625 ?

The advantage is that it does not run any irrelevant hooks like fundamental-mode-hook, find-file-hooks and also it does not access the disk.

yyoncho (commented):
Looks good, two minor issues:

lsp-mode.el
crabbkw: fixed -- sorry about that

lsp-mode.el
crabbkw: switched to `with-temp-buffer` as per that comment -- tested that it worked even if dir locals are set via the class.

crabbkw:
Again, if the code looks acceptable, I'd still like to:
1. squash this down to a single commit with a good message (does GitHub handle force-pushes cleanly in PRs?)
2. fix the changelog appropriately
3. write out the resolution in the GitHub issue

Please let me know when you're ready for me to start that list above.

yyoncho:
Sure, go ahead.

crabbkw:
Commits squashed
Commit message updated appropriately
Changelog updated appropriate

I'll update the corresponding GitHub issue if/when this is accepted.

Thanks much for the attention on this issue -- it was causing me quite a bit of pain in my day to day workflow on Windows.

yyoncho (approved):

yyoncho:
Thank you for contributing to `lsp-mode`! :tada:

PR #2573, opened 2021-01-31
clients/lsp-csharp.el: Implement go-to definition in metadata.

razzmatazz:
This commit adds a handler for 'osmd://' uri when returned from
omnisharp-roslyn, parses the uri and issues corresponding 'o#/metadata'
request that is used to retrieve metadata source to display to the user.

The design is modelled after clients/lsp-clojure.el and lsp-java. I am not 100% about the directory: currently I save metadata .cs files under `<project-dir>/.cache/lsp-csharp/metadata` -- but maybe that needs to be a customizable variable or some other dir?

This PR is marked as draft as there is some changes pending to be merged into `omnisharp-roslyn` yet but open for review.

PR #2740, opened 2021-03-24
Trigger update server on start

jcs090218:
Ask user to update server if needed on start.

### DIFF
--- a/clients/lsp-csharp.el
+++ b/clients/lsp-csharp.el
@@ -27,6 +27,7 @@
 (require 'lsp-mode)
 (require 'gnutls)
 (require 'f)
+(require 'ht)
 
 (defgroup lsp-csharp nil
   "LSP support for C#, using the Omnisharp Language Server.
@@ -93,7 +94,9 @@ Usually this is to be set in your .dir-locals.el on the project root directory."
 (defun lsp-csharp--latest-available-version ()
   "Returns latest version of the server available from github."
   (lsp-csharp--version-list-latest
-   (seq-map (lambda (elt) (s-trim (cdr (assq 'name elt))))
+   (seq-map (lambda (elt)
+              (or (ignore-errors (s-trim (cdr (assq 'name elt))))
+                  (ignore-errors (ht-get elt "name"))))
             (lsp-csharp--fetch-json "https://api.github.com/repos/OmniSharp/omnisharp-roslyn/releases"))))
 
 (defun lsp-csharp--server-dir (version)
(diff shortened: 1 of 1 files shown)


clients/lsp-csharp.el
```diff
@@ -448,6 +461,11 @@ using the `textDocument/references' request."
       (lsp-show-xrefs (lsp--locations-to-xref-items locations-found) nil t)
     (message "No references found")))
 
+(defun lsp-csharp--after-open-fn (&rest _)
```
yyoncho, line 464: Ideally, this should run one time and before the server startup, not in after open. I propose you to set lsp-csharp-auto-update-server to nil after the update and also, ask the user to restart the server if they want new version to take effect. At some point we should move the update logic in the core.

yyoncho (commented):
Looks good.

@razzmatazz, willing to take a look?

clients/lsp-csharp.el
razzmatazz: Yes, I agree with @yyoncho . I think this setting should be lsp-global, no need to instrument lsp-Csharp in particular IMHO.

Btw I have some changes that would move LSP-Sharp to use the standard LSP-mode package/dependency mechanism. @jcs090218 I would propose you rebase the changes after I make my PR, later this week, so a lot of custom version determination logic here would go away for good.

razzmatazz: And yes, hitting github APIs to check for new server version on every file open is too much.

can we hook into the “server started” event like @yyoncho has suggested?

yyoncho: I suggest introducing an additional property in lsp--client named `can-update?`(or similar) which is a function. lsp-mode will call it only once per session and if the user confirms an update lsp-mode will call :download-server-fn with update=t.

jcs090218: Okay, I have currently made this for check once per session by setting `lsp-csharp-auto-update-server` to `nil` after the first check.

> I suggest introducing an additional property in lsp--client named can-update?

That works for me. :)

> @jcs090218 I would propose you rebase the changes after I make my PR, later this week, so a lot of custom version determination logic here would go away for good.

@razzmatazz Sure, we can hang this PR until that point. I think you are referring to #2573?

jcs090218: Ah, okay. I thought you were referring to the existing PR.

razzmatazz: The PR is here: #2748 – That PR, however, if merged would render this PR unmergable as there would be no more custom update code in lsp-csharp.

My humble opinion, @jcs090218, is that we should strive to make this "autoupdate" functionality part of lsp core so no custom code would be added to the clients, unless required for the extra functionality.

Not sure what @yyoncho feels about this conflict of PRs – which one should go.

jcs090218: If this is going to core, I don't see point merging this PR. Close this now! :)

PR #2742, opened 2021-03-25
Make signature posframe colors customizable

wyuenho:
Currently the colors of the signagure posframe is hardcoded to be the tooltips, this PR makes it customizable so people can alias the face to company-tooltip.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4965,13 +4965,16 @@ RENDER-ALL - nil if only the signature should be rendered."
 (declare-function posframe-hide "ext:posframe")
 (declare-function posframe-poshandler-point-bottom-left-corner-upward "ext:posframe")
 
+(defface lsp-signature-posframe
+  '((t :inherit tooltip))
+  "Background and foreground for `lsp-signature-posframe'."
+  :group 'lsp-faces)
+
 (defvar lsp-signature-posframe-params
   (list :poshandler #'posframe-poshandler-point-bottom-left-corner-upward
-        :background-color (face-attribute 'tooltip :background)
         :height 6
         :width 60
         :border-width 10
-        :border-color (face-attribute 'tooltip :background)
         :min-width 60)
   "Params for signature and `posframe-show'.")
 
@@ -4984,8 +4987,12 @@ RENDER-ALL - nil if only the signature should be rendered."
                (insert str)
                (visual-line-mode 1)
                (current-buffer))
-             :position (point)
-             lsp-signature-posframe-params)
+             lsp-signature-posframe-params
+             (list
+              :position (point)
+              :background-color (face-attribute 'lsp-signature-posframe :background)
+              :foreground-color (face-attribute 'lsp-signature-posframe :foreground)
+              :border-color (face-attribute 'lsp-signature-posframe :background)))
     (posframe-hide "*lsp-signature*")))
 
 (defun lsp--handle-signature-update (signature)


ericdallo (approved):

lsp-mode.el
```diff
@@ -4939,15 +4939,21 @@ RENDER-ALL - nil if only the signature should be rendered."
 (declare-function posframe-hide "ext:posframe")
 (declare-function posframe-poshandler-point-bottom-left-corner-upward "ext:posframe")
 
-(defvar lsp-signature-posframe-params
+(defface lsp-signature-posframe
```
yyoncho, line 4942: dropping lsp-signature-posframe-params removes the possibility to customize other properties, e. g. width.

wyuenho: Well, we aren't customizing those either now right? When there's a need, someone somewhere will introduce a few new deffaces for the sizes.

ericdallo: Users could still change the value of that var ATM, right?

wyuenho: Well, maybe, if you are afraid of breaking existing user configs. But they can still change the defaults without a PR in the future with an `advice-add`. My preference now is to not introduce too many deffaces until someone asks for them.

ericdallo: I see no problems add one or more faces, my concern is we break someone else config because of that change, we could have a define-obsolete-variable for the old var just to keep things working for users that are already using that var

yyoncho: > Well, we aren't customizing those either now right? When there's a need, someone somewhere will introduce a few new deffaces for the sizes.

We do, the variable is public and it is used for customizing the popup appearance. Similar variables exist for other packages  using posframe, see dap-mouse-posframe-properties, company-posframe-show-params .

wyuenho: Fixed

lsp-mode.el
```diff
@@ -4959,6 +4962,9 @@ RENDER-ALL - nil if only the signature should be rendered."
                (visual-line-mode 1)
                (current-buffer))
              :position (point)
+             :background-color (face-attribute 'lsp-signature-posframe :background)
+             :foregorund-color (face-attribute 'lsp-signature-posframe :foreground)
```
yyoncho, line 4966: typo: foregorund

yyoncho (commented):
Looks good to me. One typo:

lsp-mode.el
wyuenho: Good catch!

yyoncho (approved):

kiennq:
Can you mention this will be overrided by `lsp-signature-posframe-params`?

wyuenho:
@kiennq later keyword arguments supplied to cl-defuns have no effect.

kiennq:
Ah, I see, didn't notice that.
So with this change, the whatever user setting of border color / background / foreground in `lsp-signature-posframe-params` will be ignored?
Can you reorder so the old `lsp-signature-posframe-params` will be placed front so at least we won't break user's configuration.

wyuenho:
done

lsp-mode.el
```diff
@@ -4958,8 +4961,11 @@ RENDER-ALL - nil if only the signature should be rendered."
                (insert str)
                (visual-line-mode 1)
                (current-buffer))
+             lsp-signature-posframe-params
              :position (point)
-             lsp-signature-posframe-params)
+             :background-color (face-attribute 'lsp-signature-posframe :background)
+             :foreground-color (face-attribute 'lsp-signature-posframe :foreground)
+             :border-color (face-attribute 'lsp-signature-posframe :background))
```
yyoncho, line 4968: I havent tested it but it seems to me that it should be: `(append lsp-signature-posframe-params (list :position ...))`. The last argument of `apply` should be list.

wyuenho: fixed

kiennq (approved):

Issue #2738, opened 2021-03-22
There is no way to always add or remove projects from a given directory to the workspace


Andre0991:
(I guess this is more like a feature request, sorry if this is not the appropriate forum for that).

Sometimes I’m on a project and I use ‘go to definition’ in some function that comes from a lib in  `~./m2`.

Then, LSP asks if I want to add the project to the workspace list etc.

The annoying thing is that LSP always asks me if I want to add it.

So it would be nice to be able to _always_ ignore projects in a given folder (in this case, `~./m2`. (or maybe always add them)

ericdallo:
@yyoncho I thought [this](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-clojure.el#L228-L231) was the way to fix that, but it seems it's not working, do you know why?

yyoncho:
> @yyoncho I thought [this](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-clojure.el#L228-L231) was the way to fix that, but it seems it's not working, do you know why?

Maybe this

> (string-match-p ".m2/repository" (buffer-file-name))

does not work on @Andre0991 PC.

ericdallo:
It seems to doesn't work for everyone, including me, I'm just not sure if that's is the correct way to do that  😅

Andre0991:
I confirmed the at least the string match part does work (it's true and the when's body is executed).

```
(defun lsp-clojure--library-folders (_workspace)
  "Return the library folders path to analyze for WORKSPACE."
  (when (string-match-p ".m2/repository" (buffer-file-name))
    (message "in when")
    (list (file-name-directory (buffer-file-name)))))```

ericdallo:
Yeah, I could repro the same thing, still lsp-mode asks for the root, not sure why

ericdallo:
@Andre0991, Could you test on latest lsp-mode, I made some changes that may fix the issue, not sure, but it worths a shot

Andre0991:
Sure, I'll test it. But it seems that `2021.03.24-00.41.55` is the latest version for the `clojure-lsp-native` formula yet.

ericdallo:
I meant lsp-mode, not clojure-lsp :p

Andre0991:
Oops, of course 😅

I confirm it works, thanks ⭐️

Note: `lsp-version` returns an error:

[...]

PR #2746, opened 2021-03-27
[clojure] Use same approach as lsp-java for opening library files

ericdallo:
Related to #2738

### DIFF
--- a/clients/lsp-clojure.el
+++ b/clients/lsp-clojure.el
@@ -64,6 +64,18 @@
   :group 'lsp-clojure
   :package-version '(lsp-mode . "7.1"))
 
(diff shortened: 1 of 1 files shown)


clients/lsp-clojure.el
```diff
@@ -230,22 +242,32 @@ If there are more arguments expected after the line and column numbers."
   (when (string-match-p ".m2/repository" (buffer-file-name))
```
yyoncho, line 242: method lsp-clojure--library-folders no longer needed?

yyoncho (approved):

jcs090218 (approved):

PR #2748, opened 2021-03-28
csharp: Remove custom installation code

razzmatazz:
This removes custom in-sync server installation code in lsp-csharp.el in favor
of the standard lsp-dependency/package mechanism.

This removes lsp-csharp-update-server interactive fn, but hopefully
ability to update/upgrade will come as standard functionality in lsp-mode.

### DIFF
--- a/clients/lsp-csharp.el
+++ b/clients/lsp-csharp.el
@@ -60,189 +60,67 @@ Usually this is to be set in your .dir-locals.el on the project root directory."
   :group 'lsp-csharp
   :type 'string)
 
-(defun lsp-csharp--version-list-latest (lst)
-  (->> lst
-       (-sort (lambda (a b) (not (version<= (substring a 1)
-                                            (substring b 1)))))
-       cl-first))
+(defcustom lsp-csharp-omnisharp-roslyn-download-url
+  (concat "https://github.com/omnisharp/omnisharp-roslyn/releases/latest/download/"
+          (cond ((eq system-type 'windows-nt)
+                 ; On Windows we're trying to avoid a crash starting 64bit .NET PE binaries in
+                 ; Emacs by using x86 version of omnisharp-roslyn on older (<= 26.4) versions
+                 ; of Emacs. See https://lists.nongnu.org/archive/html/bug-gnu-emacs/2017-06/msg00893.html"
+                 (if (and (string-match "^x86_64-.*" system-configuration)
+                          (version<= "26.4" emacs-version))
+                     "omnisharp-win-x64.zip"
+                   "omnisharp-win-x86.zip"))
 
-(defun lsp-csharp--latest-installed-version ()
-  "Returns latest version of the server installed on the machine (if any)."
-  (lsp-csharp--version-list-latest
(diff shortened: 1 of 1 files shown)


clients/lsp-csharp.el
```diff
@@ -60,189 +60,53 @@ Usually this is to be set in your .dir-locals.el on the project root directory."
...
-        (eq system-type 'darwin))
-    (call-process "tar" nil nil t "xf" filename "-C" target-dir))
-
-   (t (error "lsp-csharp cannot extract \"%s\" on platform %s (yet)" filename system-type))))
+(defcustom lsp-csharp-omnisharp-roslyn-download-url
+  (format "https://github.com/omnisharp/omnisharp-roslyn/releases/latest/download/%s.zip"
+          (pcase system-type
+            ('gnu/linux "omnisharp-linux-x64")
+            ('darwin "omnisharp-osx")
+            ('windows-nt "omnisharp-win-x64")))
```
kiennq, line 68: Don't we need to care about this anymore?
> On Windows we're trying to avoid a crash starting 64bit .NET PE binaries in
Emacs by using x86 version of `omnisharp-roslyn` on older (<= 26.4) versions
of Emacs.

razzmatazz: Good catch! I have ammended the commit to restore checks.

yyoncho (approved):
Looks good.

kiennq (approved):

jcs090218 (approved):

lafleurh:
The automatic install does not work on Arm:

- The Arm / Mono download does not include Mono and there is no run script to start OmniSharp
- Because the install points at run, during the install it will fail towards the end.
- This limitation was documented with OmniSharp-emacs
- Perhaps check for a prerequisite of Mono
- Perhaps generate the run script

Environment:

- Raspberry Pi 4
- Raspbian/Debian 10.9
- Emacs 26.1

Work around:

- Install mono
- Before installing the server, run:

```
mkdir ~/.emacs.d/.cache/lsp/omnisharp-roslyn/                            
mkdir ~/.emacs.d/.cache/lsp/omnisharp-roslyn/latest/                     
mkdir ~/.emacs.d/.cache/lsp/omnisharp-roslyn/latest/omnisharp-roslyn     
touch ~/.emacs.d/.cache/lsp/omnisharp-roslyn/latest/omnisharp-roslyn/run

cat <<EOT >> ~/.emacs.d/.cache/lsp/omnisharp-roslyn/latest/omnisharp-roslyn/run
#!/bin/bash
exec mono ~/.emacs.d/.cache/lsp/omnisharp-roslyn/latest/omnisharp-roslyn/OmniSharp.exe $@
EOT
```

yyoncho:
@razzmatazz do you think we should handle that case or eventually only document it?

razzmatazz:
> @razzmatazz do you think we should handle that case or eventually only document it?

I will add code to handle this. We will also have to check for presence of mono, I guess to make the process more seamless. The regular versions of omnisharp-roslyn has mono integrated into distro, bet the plain version that we download for arm does not.

Issue #2741, opened 2021-03-25
Is LSP doing too much work when watching folders


rtpg:
So my understanding is that when file watching is happening, we _only watch_ directories through setting up a notification watch on various sub directories.

However, in `lsp-watch-root-folder`, I'm seeing some of the following:
- calling `(lsp--directory-files-recursively dir ".*" t)` which includes directories but also just includes files as well
- checking the file watch threshhold against the number of files, instead of the number of directories (if we're thinking of a conservative estimate of 5-10 files in each folder, we're looking at much more)
- On top of that when we're setting up the watches themselves, instead of directly generating a list of directories to watch, we are again building up a list of files and then filtering that list back down.

the full list of files thing, in particular, has caused my machine to basically hang on certain degenerate cases (in particular where I had a folder with a bunch of test data all inside a single folder)

I _think_ that it would make sense to rework this function to make it much less likely for people to hit file watch limit warnings (that might not actually be applicable) but I wanted to first throw it by people here to make sure I wasn't missing an important detail here.

I'm including a reference to the function I was looking over here:

https://github.com/emacs-lsp/lsp-mode/blob/ed9308e57ff3f3de0adfb2a5036d43852ea1179e/lsp-mode.el#L1712-L1754

yyoncho:
Refer to https://github.com/emacs-lsp/lsp-mode/issues/713#issuecomment-803901668

Closing as duplicate.

yyoncho:
This is the correct comment link: https://github.com/emacs-lsp/lsp-mode/issues/713#issuecomment-473458528

rtpg:
Sorry, I think that that comment is related to more specific optimizations, but this issue in particular is just about how (based on my reading of the elisp) we are _only creating file watches on directories_ in the current version anyways. So my proposal here would be to modify the function that builds up the watch list to only return directories (since other files will be filtered out anyways), since that function is only being called in one place. And then the "large file warning" could also happen less often.

Would you be interested in a proposal patch for this? I think it would be self-contained and I have time to work on this.

yyoncho:
I am not sure whether OS uses the same resources to watch one folder with 5 files and one folder with 1000 files. Thus it is not clear whether this solution is correct. Someone proposed changing the function which counts the files to stop when we hit the limit (this will require a change of wording in the error message). I think that this is a good improvement and also it will solve your issue as well. WDYT?

rtpg:
[...]

PR #2752, opened 2021-03-30
Improve performance of project file watch setup

rtpg:
This improves performance through a couple of strategies:

 - Only check if a directory is an ignored directory once. Previously  the filter was in multiple parts of the the chain.
 - When setting up watches, only generate a list of directories (as files end up getting filtered out later anyways)
 - Directly list directory contents when recursively fetching items,  instead of relying on completion code

(See related discussion [here](https://github.com/emacs-lsp/lsp-mode/issues/2741#issuecomment-808828943))

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1669,49 +1669,50 @@ This set of allowed chars is enough for hexifying local file paths.")
            (memq event-type '(created deleted changed)))
       (funcall callback event)))))
 
-(defun lsp--directory-files-recursively (dir regexp ignored-directories &optional include-directories)
-  "Copy of `directory-files-recursively' but it skips directories matching any regex in IGNORED-DIRECTORIES."
-  (let* ((result nil)
-         (files nil)
-         (dir (directory-file-name dir))
-         ;; When DIR is "/", remote file names like "/method:" could
-         ;; also be offered.  We shall suppress them.
-         (tramp-mode (and tramp-mode (file-remote-p (expand-file-name dir)))))
-    (dolist (file (sort (file-name-all-completions "" dir)
-                        'string<))
-      (unless (member file '("./" "../"))
-        (if (and (directory-name-p file)
-                 (not (lsp--string-match-any ignored-directories (f-join dir (f-filename file)))))
-            (let* ((leaf (substring file 0 (1- (length file))))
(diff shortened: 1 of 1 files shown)


rtpg:
I'm having a hard time understanding these parse failures, looks like _something_ is wrong with Emacs 26.x in my thing but I can't quite figure out what

lsp-mode.el
```diff
@@ -1725,41 +1727,32 @@ regex in IGNORED-FILES."
...
     (lsp-log "Creating watch for %s" dir)
     (when (or
            (not warn-big-repo?)
            (not lsp-file-watch-threshold)
-           (let ((number-of-files (length (lsp--directory-files-recursively dir ".*" ignored-directories t))))
+           (let ((number-of-files (length dirs-to-watch)))
              (or
               (< number-of-files lsp-file-watch-threshold)
               (condition-case _err
                   (lsp--ask-about-watching-big-repo number-of-files dir)
```
yyoncho, line 1740: please fix the wording in that method because we are no longer counting files.

lsp-mode.el
```diff
@@ -1713,6 +1686,35 @@ Do you want to watch all files in %s? "
...
+
+(defun lsp--path-is-watchable-directory (path dir ignored-directories)
+  "Figure out whether PATH (inside of DIR) is meant to have a file watcher set.
+IGNORED-DIRECTORIES is a list of regexes to filter out directories we don't want to watch."
+  (let
+      ((full-path (f-join dir path)))
+    (and (f-dir-p full-path)
+         (not (equal path "."))
+         (not (equal path ".."))
+         (not (lsp--string-match-any ignored-directories full-path))
```
yyoncho, line 1698: always close the brackets on the same line, here and on a few more places in the PR.

yyoncho (commented):
Looks good. I am still to perform manual testing.

@jcs090218 any idea why CI is red? Seems like we are fetching a package that depends on emacs 27.1 and all 26.x are failing?

rtpg:
Sorry, forgot to comment but I went over the comments and tried to clean up the indentation and the naming (also updated the file warning to try and make it clearer that this is a directory count)

yyoncho (approved):
Thank you for contributing to `lsp-mode` and sorry for the delay.

PR #2759, opened 2021-04-03
csharp: write `run' when missing on misc archs (like arm) during installation

razzmatazz:
This should make the installation work on arm platforms. This still
requires mono to be installed by the user on their own but this is
better than nothing.

---

A follow up on #2748 with code as suggested by @lafleurh

### DIFF
--- a/clients/lsp-csharp.el
+++ b/clients/lsp-csharp.el
@@ -102,14 +102,23 @@ Usually this is to be set in your .dir-locals.el on the project root directory."
              :store-path lsp-csharp-omnisharp-roslyn-store-path))
 
 (defun lsp-csharp--download-server (_client callback error-callback _update?)
-  "Download zip package for omnisharp-roslyn and install it."
+  "Download zip package for omnisharp-roslyn and install it.
+Will invoke CALLBACK on success, ERROR-CALLBACK on error."
   (lsp-package-ensure
    'omnisharp-roslyn
    (lambda ()
      (lsp-unzip lsp-csharp-omnisharp-roslyn-store-path
                 lsp-csharp-omnisharp-roslyn-server-dir)
      (unless (eq system-type 'windows-nt)
-       (set-file-modes (f-join lsp-csharp-omnisharp-roslyn-server-dir "run") #o700))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

clients/lsp-csharp.el
```diff
@@ -102,14 +102,25 @@ Usually this is to be set in your .dir-locals.el on the project root directory."
...
    (lambda ()
      (lsp-unzip lsp-csharp-omnisharp-roslyn-store-path
                 lsp-csharp-omnisharp-roslyn-server-dir)
      (unless (eq system-type 'windows-nt)
-       (set-file-modes (f-join lsp-csharp-omnisharp-roslyn-server-dir "run") #o700))
+       (let ((run-script (f-join lsp-csharp-omnisharp-roslyn-server-dir "run"))
+             (omnisharp-exe (f-join lsp-csharp-omnisharp-roslyn-server-dir "OmniSharp.exe")))
+
+         ; create the `run' script, if missing (e.g. when server binaries are extracted from omnisharp-mono.zip)
+         (if (and (not (f-exists-p run-script)) (f-exists-p omnisharp-exe))
```
yyoncho, line 117: nit: seems like `if` should be `when`. Not sure about set-files-mode placement - when omnisharp-exe does not exist it seems like the run file won't exist as well and set-files-mode will throw. Also, do we need to check if mono is present and eventually show a warning or this will be part of separate PR?

yyoncho (approved):
Looks good. Minor nit:

clients/lsp-csharp.el
razzmatazz: Thanks for these nits!
- I have updated the code to use `when` – I actually didn't even know this form exists in emacs/lisp
- I have pondered a bit and I think that we should keep this simple for the moment
  - checking for mono seems to be out of the scope -- the use should notice the error in `*lsp-log*`
  - this is only for non-common/more rare platforms where `omnisharp-roslyn` does not release native builds, so I *think* keeping it simple is ok for now;
  - thus I have removed the check for presence of OmniSharp.exe too

PR #2760, opened 2021-04-04
csharp: add test file in test/lsp-csharp-test.el

razzmatazz:
Adds a single test for csharp, so we can check if workspace initializes on a sample project.

### DIFF
--- a/test/lsp-csharp-test.el
+++ b/test/lsp-csharp-test.el
@@ -1,0 +1,69 @@
+;;; lsp-csharp-test.el --- tests for C# -*- lexical-binding: t -*-
+
+;; Copyright (C) 2021 Saulius Menkevičius <sauliusmenkevicius@fastmail.com>.
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; Tests for C# integration.
+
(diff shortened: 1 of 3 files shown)


test/lsp-csharp-test.el
```diff
@@ -0,0 +1,69 @@
...
+
+(ert-deftest lsp-clangd-initialised-workspace ()
+  (skip-unless (memq system-type '(gnu/linux darwin windows-nt)))
+  (lsp-in-sample-csharp-project
+   (->
+    ;; now check that the workspace has started
+    (lsp-test-wait (eq 'initialized
+                       (lsp--workspace-status (cl-first (lsp-workspaces)))))
+   (deferred:sync!))
+   ))
```
ericdallo, line 67:
```suggestion
   (deferred:sync!))))
```

razzmatazz:
@jcs090218 could you guide me a little bit into how I can make lsp-csharp-test.el to be run with ert?

Also, do I have to install the server myself somehow during the test setup process or will the automatic server installation work here? Also, do I need to add hooks or something so that `(find-file (f-join lsp-test-location "fixtures/SampleCSharpProject/Program.cs"))` will actually start lsp mode?

I copied most of the code from lsp-clangd-test.el, but I might not know the underlying context well here, on how integration tests work..

jcs090218:
> @jcs090218 could you guide me a little bit into how I can make lsp-csharp-test.el to be run with ert? It seems like CI skips lsp-csharp-test.el and there is no registration list for tests so I figure those are autodetected somehow?

Sorry for the late reply! I can see the test on Windows but not on Unix system. I think the culprit is the line,

https://github.com/emacs-lsp/lsp-mode/blob/5ca68ed5b9295f298fa45e3cc0f4901a7c471c46/Makefile#L96

> Also, do I have to install the server myself somehow during the test setup process or will the automatic server installation work here? Also, do I need to add hooks or something so that (find-file (f-join lsp-test-location "fixtures/SampleCSharpProject/Program.cs")) will actually start lsp mode?

No, I think you would have to install the server in the test file. Is there a function so we can install the server? Maybe like `(lsp-install-server 'csharp)` so it would be a lot easier even with other test cases? You would also need to install `csharp-mode`.

jcs090218:
I think the function `lsp-install-server` now support argument `server-id`. You can now install the server using this function! Thanks to @yyoncho in #2817.

PR #2762, opened 2021-04-04
lsp-completion: check falsy value for some completion members

### LINKED ISSUE #2761: Unable to filter completions when `filterText` is an empty string
**Describe the bug**
The LSP spec states that when a `CompletionItem`'s `filterText` is falsy, it should use the `label` to perform filtering. Recently, zls switched to returning an empty string for `filterText`, following the LSP spec (https://github.com/zigtools/zls/pull/251, https://github.com/microsoft/language-server-protocol/issues/1222). However, with this change, it appears that typing any characters to to filter down the list of completions using company causes all the completions to disappear, which I expect is because they are filtered out given that `filterText` is `""`.

**To Reproduce**
Install [zls](https://github.com/zigtools/zls) and open any zig file and try completions. After entering a '.' and triggering completions, company displays a list of completions. Any text following that causes completions to be filtered away.

**Expected behavior**
`filterText` should not be used for filtering when it is the empty string.

zls: https://github.com/zigtools/zls

**OS**
NixOS 21.05

kiennq:
Fix #2761

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -129,14 +129,18 @@ This will help minimize popup flickering issue in `company-mode'."
 (declare-function company-doc-buffer "ext:company")
 (declare-function yas-expand-snippet "ext:yasnippet")
 
+(defun lsp-falsy? (val)
+  "Non-nil if VAL is falsy."
+  ;; https://developer.mozilla.org/en-US/docs/Glossary/Falsy
+  (or (not val) (equal val "") (equal val 0)))
+
 (cl-defun lsp-completion--make-item (item &key markers prefix)
   "Make completion item from lsp ITEM and with MARKERS and PREFIX."
   (-let (((&CompletionItem :label
-                           :insert-text?
                            :sort-text?
                            :_emacsStartPoint start-point)
           item))
-    (propertize (or label insert-text?)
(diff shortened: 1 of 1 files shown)


lsp-completion.el
```diff
@@ -194,7 +198,7 @@ Return `nil' when fails to guess prefix."
     (lsp--position-to-point (lsp:range-start (lsp:text-edit-range text-edit?))))
    (t
     (-let* (((&CompletionItem :label :insert-text?) item)
-            (text (or insert-text? label))
+            (text (or (unless (lsp-falsy? insert-text?) insert-text?) label))
```
yyoncho, line 201: this could be rewritten as `(if (lsp-falsy...`, here and on a few more places.

yyoncho (approved):
Looks good, one minor nit:

lsp-completion.el
kiennq: Yeah, I'm purposely writing like this so we can add more term to the `or` clause` if the LSP protocol ever changes/adds new text. Do you think I can merge keep this?

yyoncho: ok, it is not a big deal...

PR #2763, opened 2021-04-04
Add a command to kill the server behind a workspace

nbfalcon:
`lsp-workspace-kill` will kill the server behind a given workspace using
`kill-process`, and optionally restart it. The naming was chosen to be
consistent with `lsp-workspace-restart/shutdown`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7979,6 +7979,16 @@ Errors if there are none."
   (lsp--warn "Stopping %s" (lsp--workspace-print workspace))
   (with-lsp-workspace workspace (lsp--shutdown-workspace)))
 
+(defun lsp-workspace-kill (workspace)
+  "Kill the language server behind WORKSPACE.
+Prefer `lsp-workspace-shutdown' if possible, since this command
+may leave the server and Emacs in an inconsistent state, so use
+only in an emergency."
+  (interactive (list (lsp--read-workspace)))
+  (lsp--warn "Killing %s" (lsp--workspace-print workspace))
+  (setf (lsp--workspace-shutdown-action workspace) 'shutdown)
+  (kill-process (lsp--workspace-proc workspace)))
+
 (defun lsp-disconnect ()
   "Disconnect the buffer from the language server."
   (interactive)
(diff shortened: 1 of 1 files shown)


nbfalcon:
This feature was asked by `LunarForce` on Discord: "Is there a way to list and kill language server processes in the current buffer [...]"

lsp-mode.el
```diff
@@ -7979,6 +7979,20 @@ Errors if there are none."
...
+may leave the server and Emacs in an inconsistent state, so use
+only in an emergency.
+
+If RESTART (prefix-arg) is non-nil, cause the server to be
+restarted afterwards."
+  (interactive (list (lsp--read-workspace) current-prefix-arg))
+  (lsp--warn "Killing %s" (lsp--workspace-print workspace))
+  (setf (lsp--workspace-shutdown-action workspace)
+        (if restart 'restart 'shutdown))
+  (kill-process (lsp--workspace-proc workspace)))
```
kiennq, line 7994: This needs to be `lsp--workspace-cmd-proc` for the server process to be killed, else you will just kill the connection in the case of tcp server.
The `lsp-workspace-shutdown/restart` is already killing (and restarting) the server process already. Can you elaborate more on why we need this?

nbfalcon: `lsp-workspace-shutdown` is more graceful, so won't work if a server hangs. `lsp-workspace-kill` just kills the server, which will bring it down immediately most of the time (and as such should only be used for debugging).

yyoncho: > `lsp-workspace-shutdown` is more graceful, so won't work if a server hangs.

IMHO we should make sure that lsp-workspace-shutdown works even with a hanging server.

kiennq: Looking at `lsp--shutdown-workspace`, if the server hangs, after the time out we will call `lsp--uninitialize-workspace`, which kills the server process too.

nbfalcon: Actually, it won't, due to `error` exiting the function on timeout:
```
(let ((lsp-response-timeout 0.01))
  (condition-case nil (with-current-buffer "<some lsp buffer>" (lsp-request "textDocument/documentSymbol" '()))
    (error t)))
 ```
 
 Perhaps `lsp-workspace-shutdown` should be async? I don't really have much justification for `lsp-workspace-kill` other than "it might be useful as a debugging tool sometime", so feel free to close this PR.

PR #2764, opened 2021-04-05
Update lsp-elixir package.

hjpotter92:
- Remove custom script for lsp-elixir server.
 - Rename `lsp-clients-elixir-server-executable` to `lsp-elixit-server-command`.
 - Mark `lsp-clients-elixir-server-executable` as obsolete.
 - Define `elixir-ls` as lsp dependency.
 - Update how new connection is done for elixir-ls.

### DIFF
--- a/clients/lsp-elixir.el
+++ b/clients/lsp-elixir.el
@@ -92,15 +92,15 @@ If value is `\"\"` then defaults to the workspace rootUri."
   :group 'lsp-mode
   :link '(url-link "https://github.com/elixir-lsp/elixir-ls"))
 
-(defcustom lsp-clients-elixir-server-executable
-  (if (equal system-type 'windows-nt)
-      "language_server.bat"
-    "language_server.sh")
-  "The elixir-language-server executable to use.
-Leave as just the executable name to use the default behavior of
-finding the executable with `exec-path'."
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
LGTM

jcs090218 (approved):

ilya-babanov:
Hey, are there any reasons behind this change?

[elixir-ls](https://github.com/elixir-lsp/elixir-ls) project still produces `language_server.sh` file as its main executable, so this change breaks the default workflow, unless I'm missing something. 

ElixirLS VSCode extension (developed by maintainers of `elixir-ls` itself) seems to be using [`language_server.sh` command name too](https://github.com/elixir-lsp/vscode-elixir-ls/blob/b8453818f5988554c77d32776f82da6416238bcf/src/extension.ts#L317).

ericdallo:
Oh, I checked the elixir-ls but didn't find any docs about that sh, it seemed some forgotten script, @hjpotter92 would mind checking that please?

hjpotter92:
ok, checking it further; on arch linux, the launcher at `/usr/bin/elixir-ls` is actually linked to the `language_server.sh` inside the `/usr/lib/elixir-ls` path.

I'm definitely sorry for not checking it deeper. Will create a new PR to revert the `elixir-ls` command, and update the docstring accordingly.

PR #2765, opened 2021-04-06
support plaintext documentation format

wyuenho:
Some servers like jedi allows the client to choose which markup format is
preferred in the documentation, but lsp-mode currently does not declare support
for it in completions, this PR fixes this situation.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -774,7 +774,8 @@ Changes take effect only when a new session is started."
                                         (vala-mode . "vala")
                                         (actionscript-mode . "actionscript")
                                         (d-mode . "d")
-                                        (zig-mode . "zig"))
+                                        (zig-mode . "zig")
+                                        (text-mode . "plaintext"))
   "Language id configuration.")
 
 (defvar lsp--last-active-workspaces nil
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4848,8 +4849,7 @@ When language is nil render as markup if `markdown-mode' is loaded."
         (lsp--render-string value language))
        ((MarkupContent :value :kind)
         (lsp--render-string value kind))
-       ;; plain string
-       ((pred stringp) (lsp--render-string content "markdown"))
+       ((pred stringp) (lsp--render-string content "plaintext"))
```
yyoncho, line 4852: As per spec, when the server is sending simple string it is markdown, so this part should stay as it is:

```
/**
 * MarkedString can be used to render human readable text. It is either a
 * markdown string or a code-block that provides a language and a code snippet.
 * The language identifier is semantically equal to the optional language
 * identifier in fenced code blocks in GitHub issues.
 *
 * The pair of a language and a value is an equivalent to markdown:
 * ```${language}
 * ${value}
 * ```
 *
 * Note that markdown strings will be sanitized - that means html will be
 * escaped.
 *
 * @deprecated use MarkupContent instead.
 */
type MarkedString = string | { language: string; value: string };
```

yyoncho (approved):

PR #2768, opened 2021-04-07
Properly build path for clj-kondo config

jcsims:
`concat` doesn't handle inserting path separators where necessary,
whereas `f-join` does.

Brief discussion of this here: https://clojurians.slack.com/archives/CPABC1H61/p1617831599255600


ericdallo (approved):
Thank you!

Issue #2612, opened 2021-02-15
Emacs hangs temporarily while trying to get completion while rust-analyzer is initializing.


RomeuG:
**Describe the bug**
Emacs hangs temporarily while trying to get completion while `rust-analyzer` is initializing.
`rust-analyzer` takes  some seconds to initialize, going from a small amount of RAM to 900MB RAM (this is normal, just giving away some details).

**To Reproduce**
Steps to reproduce the behavior(sample project + file which can be used to reproduce the issue with.)

**Expected behavior**
Emacs should not hang temporarily.

**Which Language Server did you use**
I used `lsp-rust` with `rust-analyzer`.

**OS**
GNU/Linux

**Error callstack**
Information from `*lsp-log*`:

```
Command "rls" is present on the path.
Command "/bin/rust-analyzer" is present on the path.
Command "rls" is present on the path.
Command "/bin/rust-analyzer" is present on the path.
Found the following clients for /home/romeu/Documents/Projects/rustproject/src/main.rs: (server-id rls, priority -1), (server-id rust-analyzer, priority 1)
The following clients were selected based on priority: (server-id rust-analyzer, priority 1)
```

RomeuG:
Sorry, the title is kind of wrong.
It should be: 
Emacs hangs temporarily while trying to get completion while rust-analyzer is initializing.

yyoncho:
Hanging might be caused by establishing the watches which happens during startup and causes hanging. Can you provide perf report?

RomeuG:
This is the perf report:

```
         168  66% - ...
         167  66%  - company--perform
         167  66%   - company--begin-new
         167  66%    - company-calculate-candidates
         167  66%     - company--fetch-candidates
         167  66%      - company-call-backend-raw
         167  66%       - apply
         167  66%        - company-capf
         167  66%         - company-capf--candidates
         167  66%          - completion-all-completions
         167  66%           - completion--nth-completion
         167  66%            - completion--some
         167  66%             - #<compiled 0x11b93509fa5b617b>
         165  65%              - completion-basic-all-completions
         165  65%               - completion-pcm--all-completions
         165  65%                - #<compiled -0x1c081f0461c2fb95>
         165  65%                 - #<compiled -0x1a98156bc45cc026>
          13   5%                  - lsp-request-while-no-input
          12   4%                   + #<compiled -0x10d7bc3be3860626>
           1   0%                  - lsp-completion--filter-candidates
[...]

PR #2773, opened 2021-04-11
Fix empty cache being treated as cache

aaronjensen:
Fixes the issue mentioned here: https://github.com/emacs-lsp/lsp-mode/issues/2612#issuecomment-817370085

I don't believe the `listp` check is necessary as well as it should either be set to a list or nil. I believe this essentially breaks the caching of "no results". We'd basically need some sentinel value to mean either no results or no cache.

It could also be that this is totally the wrong way to fix this, but it appears to work for me.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -419,8 +419,9 @@ The MARKERS and PREFIX value will be attached to each candidate."
                                      "textDocument/completion"
                                      (plist-put (lsp--text-document-position-params)
                                                 :context (lsp-completion--get-context trigger-chars))))
-                              (completed (or (and resp (not (lsp-completion-list? resp)))
-                                             (not (lsp:completion-list-is-incomplete resp))))
+                              (completed (and resp
+                                              (not (and (lsp-completion-list? resp)
+                                                        (lsp:completion-list-is-incomplete resp)))))
                               (items (lsp--while-no-input
                                        (--> (cond
                                              ((lsp-completion-list? resp)
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -403,6 +403,10 @@ The MARKERS and PREFIX value will be attached to each candidate."
               (lsp--catch 'input
                   (let ((lsp--throw-on-input lsp-completion-use-last-result)
                         (same-session? (and lsp-completion--cache
+                                            ;; Special case for empty prefix and empty result
+                                            (or (cl-second lsp-completion--cache)
+                                                (not (string-empty-p
+                                                      (plist-get (cddr lsp-completion--cache) :prefix))))
                                             (equal (cl-first lsp-completion--cache) bounds-start)
                                             (s-prefix?
                                              (plist-get (cddr lsp-completion--cache) :prefix)


kiennq:
> I've also noticed that occasionally the existing caching breaks. Specifically, i'll type something like:
> 
> ```ts
> const foo = data.sta
> ```
> 
> and lsp will not send a new completion request and my completion cache will look like this:
> 
> ```emacs-lisp
> (459 nil :lsp-items nil :markers
>      (459 #<marker
>           (moves after insertion)
>           at 462 in ownership-table.tsx>)
>      :prefix "")
> ```
> 
> Sometimes I have to delete the `.sta` and retype it a few times to reproduce it, but I can do it pretty consistently.
> 
> I verified in the lsp-log that it doesn't ever send a request that ends in an empty list of items, so it seems like a bug. I'm happy to open another issue if you like.



> Fixes the issue mentioned here: [#2612 (comment)](https://github.com/emacs-lsp/lsp-mode/issues/2612#issuecomment-817370085)
> 
> I don't believe the `listp` check is necessary as well as it should either be set to a list or nil. I believe this essentially breaks the caching of "no results". We'd basically need some sentinel value to mean either no results or no cache.
> 
> It could also be that this is totally the wrong way to fix this, but it appears to work for me.


`listp` check is necessary as that value can contain `:incomplete` to indicate it's an incomplete request (as responded by the server) and we don't want to cache in that case.
It's varied from server to server but for an empty prefix, with `clangd` and `rust-analyzer` it will send back an incomplete candidate results response, with `json-ls`, it will send back full candidate result and for some other language servers like `ts-ls` it just sends back a null complete candidate results response.

While I think that the language server should send back an incomplete response, in this case, to indicate it cannot fetch all data yet, we might need some work-around in `lsp-mode` side. I would suggest you change in `same-session?` block to make sure that `same-session?` would require either not null `(cl-second lsp-completion--cache)` or not empty prefix

lsp-completion.el
```diff
@@ -411,7 +411,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
                      ((or done? result) result)
                      ((and (not lsp-completion-no-cache)
                            same-session?
-                           (listp (cl-second lsp-completion--cache)))
+                           (cl-second lsp-completion--cache))
```
kiennq, line 414: I think it's better to change in `same-session?` block before to something like this

``` elisp
(same-session? (and lsp-completion--cache
                    ;; Special case for empty prefix and empty result
                    (or (cl-second lsp-completion--cache)
                        (not (string-empty-p
                              (plist-get (cddr lsp-completion--cache) :prefix))))
                    (equal (cl-first lsp-completion--cache) bounds-start)
                    (s-prefix?
                     (plist-get (cddr lsp-completion--cache) :prefix)
                     (buffer-substring-no-properties bounds-start (point)))))

```

aaronjensen: I can reproduce it with a non-empty prefix:

```
(483 nil :lsp-items nil :markers
     (483 #<marker
          (moves after insertion)
          at 486 in ownership-table.tsx>)
     :prefix "st")
```

kiennq: Do you know what's the response from the server here? You can check it via lsp trace io.
Should be in format like `[Trace - 10:38:13 AM] Received response 'textDocument/completion - (284)' in 8ms.`

Here is the spec https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion

> Response:
result: `CompletionItem[]` | `CompletionList` | `null`. If a `CompletionItem[]` is provided it is interpreted to be complete. So it is the same as `{ isIncomplete: false, items }`

So if the server returns `[]`, it will be interpreted as `CompletionItem[]` and treat as the server cannot find any more match on this prefix, and the result is complete. Thus even if you type more, there will be no new result, the cache is correct in that case.
if the server returns just `null`, we may have a bug here

aaronjensen: Log below to generate this cache. It looks like the response came in after the `ta` was typed. The `s` triggered the completion and got cached even though there was pending input.

```
  const ssat = data.sta
```

```
(460 nil :lsp-items nil :markers
     (460 #<marker
          (moves after insertion)
          at 463 in ownership-table.tsx>)
     :prefix "s")
```

```
[Trace - 11:51:45 PM] Sending notification 'textDocument/didChange'.
Params: {
  "textDocument": {
    "uri": "file:///Users/user/Source/myapp/myapp-team-app/src/components/questionnaires/incorporation/form-section/equity/foobar/ownership-table/ownership-table.tsx",
    "version": 435
  },
  "contentChanges": [
    {
      "range": {
        "start": {
          "line": 9,
          "character": 20
        },
        "end": {
          "line": 9,
          "character": 20
        }
      },
      "rangeLength": 0,
      "text": "s"
    }
  ]
}


[Trace - 11:51:46 PM] Sending request 'textDocument/completion - (198)'.
Params: {
  "textDocument": {
    "uri": "file:///Users/user/Source/myapp/myapp-team-app/src/components/questionnaires/incorporation/form-section/equity/foobar/ownership-table/ownership-table.tsx"
  },
  "position": {
    "line": 9,
    "character": 21
  },
  "context": {
    "triggerKind": 1
  }
}


[Trace - 11:51:46 PM] Sending notification 'textDocument/didChange'.
Params: {
  "textDocument": {
    "uri": "file:///Users/user/Source/myapp/myapp-team-app/src/components/questionnaires/incorporation/form-section/equity/foobar/ownership-table/ownership-table.tsx",
    "version": 436
  },
  "contentChanges": [
    {
      "range": {
        "start": {
          "line": 9,
          "character": 21
        },
        "end": {
          "line": 9,
          "character": 21
        }
      },
      "rangeLength": 0,
      "text": "t"
    }
  ]
}


[Trace - 11:51:46 PM] Sending notification 'textDocument/didChange'.
Params: {
  "textDocument": {
    "uri": "file:///Users/user/Source/myapp/myapp-team-app/src/components/questionnaires/incorporation/form-section/equity/foobar/ownership-table/ownership-table.tsx",
    "version": 437
  },
  "contentChanges": [
    {
      "range": {
        "start": {
          "line": 9,
          "character": 22
        },
        "end": {
          "line": 9,
          "character": 22
        }
      },
      "rangeLength": 0,
      "text": "a"
    }
  ]
}


[Trace - 11:51:46 PM] Received response 'textDocument/completion - (198)' in 176ms.
Result: [ ... ]
```

aaronjensen: I added some logging in and was able to observe that the result had a full list of results, but `nil` still got cached. I've also seen it have input-pending-p be set and no result within `lsp-request-while-no-input` and it not throw when it should have, therefore caching nil.

yyoncho: @kiennq based on this description seems like we cache when we were unable to fetch the candidates from the server, is that the case?

kiennq: The cache happens only when we can fetch the candidates from the server. Else it will throw and skip the assignment.
May be we should change the `completed` check to bellow to make sure the `null` response will be treated as incomplete.
``` elisp
(completed (and resp (not (and (lsp-completion-list? resp)
                               (lsp:completion-list-is-incomplete resp)))))
```

I think we've changed this back and forth a few time to work around some language servers problematic response

aaronjensen: That’s when it *should* happen. But I’m saying it is happening even when canceled. It’s also failing to cache (setting nil) in cases where it did get a response. There is something odd going on.

aaronjensen: Okay, I believe I see the issues. There are a few:

1. I have `lsp-completion-use-last-result` set to nil. I'm not sure why that impacts `lsp--throw-on-input`, but it does. I would expect that to be orthogonal. So I'm not actually throwing on input, it carries through, returning nil, which ultimately gets cached as nil.
2. The `lsp--while-no-input` used to populate `items` also effectively cares about `lsp-completion-use-last-result` so it will not throw if that's not set to `t`, meaning it will instead return `nil`, which will be cached.

So, can someone explain why `lsp-completion-use-last-result` disables throwing on input (which breaks caching?)?

kiennq: Can you try to change the `(completed ....)` block right bellow the check of `lsp-completion-use-last-result` to 

> ```emacs-lisp
> (completed (and resp (not (and (lsp-completion-list? resp)
>                                (lsp:completion-list-is-incomplete resp)))))
> ```



> So, can someone explain why `lsp-completion-use-last-result` disables throwing on input (which breaks caching?)?

If we don't use the last result, then just let it carry on. The `(complete ...)` block is the one who break caching

aaronjensen: Okay, got it. Updated, thanks. This seems to fix it for me.

kiennq: Thanks, can you also change the `(same-session? ...)` block to
``` elisp
(same-session? (and lsp-completion--cache
                    ;; Special case for empty prefix and empty result
                    (or (cl-second lsp-completion--cache)
                        (not (string-empty-p
                              (plist-get (cddr lsp-completion--cache) :prefix))))
                    (equal (cl-first lsp-completion--cache) bounds-start)
                    (s-prefix?
                     (plist-get (cddr lsp-completion--cache) :prefix)
                     (buffer-substring-no-properties bounds-start (point)))))
```

aaronjensen: Done

kiennq (approved):

kiennq:
Thank you for your contribution.

aaronjensen:
You're welcome, thank you for the help.

PR #2776, opened 2021-04-12
Add 2 rust analyzer LSP extensions

petr-tik:
Implement OpenCargoToml - takes a universal arg to open in another window similar to lsp-clangd-find-other-file. 
Still has some outstanding comments for code review. 

Add the Related Tests interactive function - for some reason this rust-analyzer endpoint doesn't return `Runnable[]` but instead a list of key value mappings, where key is always "runnable" and value is the `Runnable`, hence I couldn't find many opportunities for code reuse with other  `lsp-rust-analyzer-*-runnable` methods

Extract the common code to run a given runnable in cargo-process-mode and use it in `lsp-rust-analyzer-run` as well as `lsp-rust-analyzer-related-tests`. Tested manually in a local checkout of rust-analyzer (of course), didn't add any tests, as there isn't a sample rust project. 

Misc code tidy-up to pacify flycheck


clients/lsp-rust.el
```diff
@@ -928,5 +937,64 @@ them with `crate` or the crate name they refer to."
...
+                                                           "experimental/openCargoToml"
+                                                           params))))
+                 ;; REVIEW is there a more idiomatic to retrieve the :uri key
+                 ;; or default to nil?
+                 (cargo-toml-uri (gethash "uri" cargo-toml-location)))
+          (funcall (if new-window #'find-file-other-window #'find-file)
+                   (lsp--uri-to-path cargo-toml-uri))
+        ;; REVIEW make it an error and/or add more actionable/debuggable info?
+        (lsp--warn "Couldn't find a Cargo.toml file or your version of rust-analyzer doesn't support this extension"))
+    (lsp--warn "Not running rust-analyzer")))
```
petr-tik, line 965: whoever is reviewing - please advise on the most idiomatic way to request and process the response as well as the appropriate levels that I should log errors/warnings against

petr-tik:
@brotzeit should I add any documentation on these 2 functions or is documentation auto-generated from the code?

clients/lsp-rust.el
```diff
@@ -928,5 +937,64 @@ them with `crate` or the crate name they refer to."
...
+                          :text-document (lsp--text-document-identifier)))
+                 ;; REVIEW is there a convention how to define and call rust-analyzer
+                 ;; lsp extension methods?
+                 (cargo-toml-location (with-lsp-workspace workspace
+                                        (lsp-send-request (lsp-make-request
+                                                           "experimental/openCargoToml"
+                                                           params))))
+                 ;; REVIEW is there a more idiomatic to retrieve the :uri key
+                 ;; or default to nil?
+                 (cargo-toml-uri (gethash "uri" cargo-toml-location)))
```
kiennq, line 960: You should define new interface similar to what in `lsp-protocol` for server response, and use the generated getter instead of `gethash` for this

clients/lsp-rust.el
```diff
@@ -928,5 +937,64 @@ them with `crate` or the crate name they refer to."
...
+Cannot reuse `lsp-rust-analyzer--select-runnable' because the runnables endpoint
+responds with Runnable[], while relatedTests responds with TestInfo[], which is a wrapper
+over runnable. Also, this method doesn't set the `lsp-rust-analyzer--last-runnable' variable"
+    (lsp--completing-read
+     "Select test: "
+     ;; since the endpoint returns a list of hash tables that
+     ;; stores &rust-analyzer:Runnable as a value under key
+     ;; "runnable"
+     ;; we need to unpack them first
+     (mapcar (lambda (runnable) (gethash "runnable" runnable)) (lsp-rust-analyzer--related-tests))
```
kiennq, line 987: Same here, using `gethash` is discouraged for server response, since we can switch to use `plist` instead of `hashtable`

petr-tik: Thanks, I am quite confused why the RelatedTests endpoint doesn't conform to the `Runnable[]` interface, I have asked the author this question (and a suggestion to reuse the interface) and will action the outcome.

clients/lsp-rust.el
petr-tik: am good to change it now - will need to work out how to define and use the protocol generator

brotzeit:
> @brotzeit should I add any documentation on these 2 functions or is documentation auto-generated from the code?

@petr-tik seems to me that only documentation of `defcustom`s are generated for the wiki.

ericdallo:
Yes, only documentation of defcustom are auto generated, otherwise you should edit `docs/manual-language-docs/lsp-rust.md` @petr-tik

petr-tik:
> Yes, only documentation of defcustom are auto generated, otherwise you should edit `docs/manual-language-docs/lsp-rust.md` @petr-tik

Thanks for clarifying. Since there were no comments about function names or UX, I think I can record the gifs, while changing the implementation

clients/lsp-rust.el
petr-tik: hey, i have been trying to define and use the protocol from looking at already implemented endpoints. However, I have also noticed that `lsp-rust-find-parent-module` doesn't use a protocol definition and instead relies on `lsp-find-locations`. 

Can you please clarify the most idiomatic way to handle this kind of `Response: Location | null`?

petr-tik: Also, shouldn't it the method be called `lsp-rust-analyzer-find-parent-module` to be consistent with other functions?

clients/lsp-rust.el
```diff
@@ -928,5 +937,63 @@ them with `crate` or the crate name they refer to."
...
+Cannot reuse `lsp-rust-analyzer--select-runnable' because the runnables endpoint
+responds with Runnable[], while relatedTests responds with TestInfo[], which is a wrapper
+over runnable. Also, this method doesn't set the `lsp-rust-analyzer--last-runnable' variable"
+  (-if-let* ((resp (lsp-rust-analyzer--related-tests))
+             (runnables (seq-map
+                         (-lambda ((&rust-analyzer:RelatedTests :runnable)) runnable)
+                         resp)))
+      ;; TODO Is it possible to bind
+      ;; C-SPC preview similar to what ivy-configured doom has with
+      ;; ivy-call-and-recenter
```
petr-tik, line 981: @kiennq I am a doom emacs user with the ivy completion layer and I really like the C-SPC preview feature.

Considering each runnable has a location field, we can use it to jump to the start of the related test function definition before running it. 

Are there any generic, non-ivy APIs in emacs that we can use to implement that?

If not, can we advise the ivy module in doom to do that? 

I will delete the comment once we know how to proceed.

petr-tik:
Not sure if this belongs here, but I realised that one of those lsp extensions can replace a function defined in rustic-mode.

```elisp
(if (and (fboundp #'rustic-open-dependency-file)
         (fboundp #'lsp-rust-analyzer-open-cargo-toml))
    (advice-add 'rustic-open-dependency-file :override #'lsp-rust-analyzer-open-cargo-toml))
```

thoughts?

yyoncho:
@petr-tik rustic has lsp module, right?

petr-tik:
> @petr-tik rustic has lsp module, right?

Indeed - thanks for pointing out this basic omission of mine. I can contribute my humble patch there, should you decide that this PR is worthy of merging and they deem my little piece of advise a deserving addition. 

In the meantime, i shall await your review and verdict on this attempt at extending the already glorious lsp-mode.

yyoncho (commented):
Looks good: few minor nits.

clients/lsp-rust.el
```diff
@@ -928,5 +937,63 @@ them with `crate` or the crate name they refer to."
...
+       runnables
+       (-lambda ((&rust-analyzer:Runnable :label)) label))))
+
+(defun lsp-rust-analyzer-related-tests (runnable)
+  "Execute a RUNNABLE test related to the current document position.
+
+Rust-Analyzer LSP protocol extension
+https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#related-tests"
+  (interactive (list (lsp-rust-analyzer--select-related-test)))
+  (if-let (runnable)
```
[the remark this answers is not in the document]
petr-tik: happy to improve it, just couldn't understand what exactly doesn't make sense in this code.

Can you please give more detail and I will fix

petr-tik:
> > I can contribute my humble patch there,
> 
> It should be in lsp-mode since there is rust-mode as well.

I can leave it at the bottom of the file then? I am just not sure if people want a behind-the-scenes change in interactive functions

yyoncho:
> I can leave it at the bottom of the file then? I am just not sure if people want a behind-the-scenes change in interactive functions

It should not be part of lsp-mode. It should be part of rustic and it should run either OOTB or it should be behind a config option.

clients/lsp-rust.el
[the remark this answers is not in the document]
petr-tik: done

yyoncho:
Can you fix the conflicts?

petr-tik:
> Can you fix the conflicts?

Have addressed your careful review to the best of my limited ability and resolved the conflict in the changelog. 

Only the ivy-specific comment left - I think the rest of the PR is ready.

yyoncho (approved):

clients/lsp-rust.el
petr-tik: moving to lsp-ivy

brotzeit (approved):

PR #2780, opened 2021-04-14
Quote parentheses at bol inside of doc-strings

tarsius:
Without this `outline-minor-mode` gets confused and considers
this line to be the beginning of a top-level code-block.

In a few cases address this by either moving or removing the
offending parenthesis instead of quoting it.


ericdallo (approved):

yyoncho (approved):

yyoncho:
Thank you!

PR #2791, opened 2021-04-19
Optimized lsp--create-filter-function for memory usage

yyoncho:


jcs090218 (approved):

lsp-mode.el
```diff
@@ -6093,55 +6110,62 @@ deserialization.")
...
-                 (condition-case err
-                     (lsp--read-json lsp-parsed-message)
-                   (error
-                    (lsp-warn "Failed to parse the following chunk:\n'''\n%s\n'''\nwith message %s"
-                              (concat leftovers input)
-                              err)))
-                 workspace)))))))))
+              (lsp--parser-on-message
+               (condition-case err
+                   (with-temp-buffer
```
kiennq, line 6150: nit: we can probably just use/reuse one buffer for this, no need to creating new temp buffer everytime.

yyoncho: I tested with several modelines -  `spacemacs', `all-the-icons', `custom', `doom', `vim-powerline' and it seems like all of them are doing fine with doom being the only one interested in buffer related hooks and it does not result in a noticeable slowdown.

kiennq (approved):

lsp-mode.el
```diff
@@ -6093,55 +6110,62 @@ deserialization.")
...
+                                (prog1 (substring-no-properties chunk 0 left-to-receive)
+                                  (setf chunk (substring-no-properties chunk left-to-receive)))
                               (prog1 chunk
-                                (setf chunk ""))))
+                                (setf chunk nil))))
                  (body-bytes (string-bytes this-body)))
             (push this-body body)
             (setf body-received (+ body-received body-bytes))
             (when (>= chunk-length left-to-receive)
-              (let ((lsp-parsed-message (decode-coding-string
```
kiennq, line 6129: @yyoncho @ericdallo Since the `lsp-parsed-message` is removed from here, the fix for #1807 will likely be broken.

PR #2793, opened 2021-04-21
Avoid positional arguments to define-minor-mode

tarsius:
Back in Emacs-21.1, `define-minor-mode` grew keyword arguments to
replace its old positional arguments.  Starting with Emacs-28.1
warning will be omitted if positional arguments are still used.

Providing values for the optional arguments but using nil and then
subsequently using the corresponding keyword arguments was never
necessary.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -808,7 +808,7 @@ them with `crate` or the crate name they refer to."
 
 (define-minor-mode lsp-rust-analyzer-inlay-hints-mode
   "Mode for displaying inlay hints."
-  nil nil nil
+  :lighter nil
   (cond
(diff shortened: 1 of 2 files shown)


yyoncho:
CI seems to be unhappy with that change.

tarsius:
This is weird, I'll have to investigate.

ericdallo:
Maybe some missing keyword argument now that we removed the nil ones?

tarsius:
Yep, I missed the part in the doc-string about having to provide at least one keyword argument if BODY is non-nil.

I've added such an argument now.

ericdallo (approved):

jcs090218 (approved):
LGTM

kiennq (approved):

Issue #2295, opened 2020-11-01
Python: No connection to Jedi / Pyls / lsp-python-ms Servers


j3cc:
**Describe the bug**
Python LSP-Mode server connection: Somehow connection to Python server is not possible. I installed lsp-jedi & jedi server in python venv, then lsp-pyl and at last lsp-ms-python. I tried different hooks and all, it also shows in the lsp-logs the server type (i.e. pyls/jedi).

```Command "pyls" is present on the path.
Command "pyls" is present on the path.
Found the following clients for /xxx/xxxx/xxx/models.py: (server-id pyls, priority -1)
The following clients were selected based on priority: (server-id pyls, priority -1)
Command "pyls" is present on the path.
Command "pyls" is present on the path.
Found the following clients for /xxx/xxx/xxx/test.py: (server-id pyls, priority -1)
The following clients were selected based on priority: (server-id pyls, priority -1)
```
However, it never goes into the lsp-mode and the lsp-ui-mode. Because, if I activate via    `M x lsp-mode`
it just shows `LSP[Disconnected]` in the status bar. With only option to turn the mode off. 
However, above info shows me it finds the language servers.

When I run emacs with the provided `lsp-start-plain.el` script it runs just fine. Without any extra config.
My question is, how to properly setup lsp-mode for emacs. I just installed 
lsp-mode 7.0.1 
lsp-python-ms-0.7.0
lsp-treemacs-0.3
lsp-ui-7.0.1
lsp-jedi-0.0.1 
via package install.
my init.el contains following info:
```(pyvenv-mode 1)
(pyvenv-workon 'myvenv) ;;enable my virtual environment automatically

(use-package flycheck
  :ensure t
  :init (global-flycheck-mode))

;flycheck works like a charm. after manually calling lsp-ui-mode, the messages are displayed in lsp-ui-mode, everything below does not work


;; LSP-Mode init for python

(use-package lsp-mode
  :ensure t
  :hook (python-mode . lsp-deferred)
  :commands (lsp lsp-deferred))

; let's add the lsp company backend
(use-package company-lsp
  :ensure t
  :config
  (push 'company-lsp company-backends))

; also installs lsp as a dependency
(use-package lsp-ui
  :ensure t
  :hook (lsp-mode . lsp-ui-mode))

;; jedi lsp client, it shows to connect to jedi epc in jedi logs, but ultimately does not work (as pyls)
;(use-package lsp-jedi
;  :ensure t
;  :config
[...]

PR #2794, opened 2021-04-21
Document lsp-mode entry point

excalamus:
As far as I understand it, `lsp-mode` uses a non-conventional entry point, `M-x lsp` instead of `M-x lsp-mode`.

The [GNU Conventions for Writing Minor Modes](https://www.gnu.org/software/emacs/manual/html_node/elisp/Minor-Mode-Conventions.html) states to:

1. Define a variable whose name ends in ‘-mode’. We call this the mode variable. 
2. Define a command, called the mode command, whose name is the same as the mode variable. Its job is to set the value of the mode variable, plus anything else that needs to be done to actually enable or disable the mode’s features. 

`lsp-mode` does implement an `lsp-mode` variable and command.  The variable appears to fit the first convention, however, the command is undocumented and the `lsp` command is documented in the docstring as being the entry point.  Further, to quote @yyoncho,

> The proper way to start lsp-mode is using `M-x lsp`

https://github.com/emacs-lsp/lsp-mode/issues/2295#issuecomment-720051764

I could find nowhere in the online docs that says what the appropriate entry point is.  Since `lsp-mode` uses a non-conventional entry point, I feel it should be mentioned.  I put it in the introductory Setup > Installation paragraph.  This paragraph is an overview of the installation process.  Enabling the mode is a sensible last step.

I am new to `lsp-mode`, so I chose wording that derived from the `lsp` docstring.  It may be more appropriate to say something like,

```
Finally, call `M-x lsp` to use `lsp-mode`.
```

as users may not know that "start a language server" really means "start using `lsp-mode`".  Actually, as a new user, that's not clear to me at all, although it appears to be the case (through the docstrings and @yyoncho's comment).

### DIFF
--- a/docs/page/installation.md
+++ b/docs/page/installation.md
@@ -2,7 +2,7 @@
 
 You need first `lsp-mode`, that is a Emacs client for an LSP server.
 Then you need to install the specific LSP server for your language.
-Finally, use `M-x lsp` to start a new language server.
+Finally, call `M-x lsp` to use `lsp-mode` or use the corresponding major mode hook to autostart the server.
 
 ## Client
 
--- a/docs/page/installation.md
+++ b/docs/page/installation.md
@@ -2,7 +2,7 @@
 
 You need first `lsp-mode`, that is a Emacs client for an LSP server.
 Then you need to install the specific LSP server for your language.
-Finally, call `M-x lsp` to use `lsp-mode` or use the corresponding major mode hook to autostart the server.
+Finally, call `M-x lsp` or use the corresponding major mode hook to autostart the server.
 
 ## Client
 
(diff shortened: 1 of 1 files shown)


ericdallo:
AFAIK users don't need to manually use `lsp` if have their languages configured correctly to auto start lsp

yyoncho:
> AFAIK users don't need to manually use `lsp` if have their languages configured correctly to auto start lsp

We can change the wording to something like:

> Finally, call `M-x lsp` to use `lsp-mode` or use the corresponding major mode hook to autostart the server.

excalamus:
It looks like the vanilla setup has users call `lsp` through the appropriate language mode hook:

```
(add-hook 'XXX-mode-hook #'lsp)
```

excalamus:
Sorry, accidentally closed the PR when trying to cancel my previous comment.

excalamus (commented):
Update based on @yyoncho's feedback

yyoncho:
Thank you!

PR #2801, opened 2021-04-23
Add ability to verify GPG signatures.

kurnevsky:
Usage example: https://github.com/emacs-lsp/lsp-metals/pull/47/commits/7b5d6a3d3f7736a382fc7da5d339f82460775b5b

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -31,6 +31,7 @@
 (require 'cl-lib)
 (require 'compile)
 (require 'dash)
+(require 'epg)
 (require 'ewoc)
 (require 'f)
 (require 'filenotify)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7031,6 +7032,12 @@ SESSION is the active session."
   :package-version '(lsp-mode . "6.3")
   :group 'lsp-mode)
 
+(defcustom lsp-verify-signature t
```
kurnevsky, line 7035: Should it be enabled by default? Not sure how gpg works on general user setup, e.g. mac.

yyoncho: You may create a test - our test suite runs on both mac and win.

kurnevsky: I doubt it will be indicative - It will require gnupg to be installed, so either I'll need to install it or it won't pass :)

kurnevsky: I'd suppose that most linux users will have it, but mac users - won't.

kurnevsky: Hm, but I can just check it :)

kurnevsky: Changed its value to `(executable-find epg-gpg-program)` - should solve most of the cases, I think.

yyoncho: I guess you can do executable-find and in case it is not present print warning/info message that the validation will be skipped and continue. Eventually, you can fail the installation and ask either to disable the validation or to install the needed binaries.

kurnevsky: ok, changed to a warning.

yyoncho (approved):

Issue #2804, opened 2021-04-23
Cannot set lsp-yaml-custom-tags


FredrikMeyer:
Before logging the bug, please make sure that:

**Describe the bug**

Cannot set  `lsp-yaml-custom-tags`.

**To Reproduce**

Try to set `lsp-yaml-custom-tags`.

For example
```
(setq lsp-yaml-custom-tags '("!And"))
```

**Expected behavior**

Use the customTags feature without erros.

**Which Language Server did you use**
lsp-yaml

**OS**
MacOs

**Error callstack**

```
Debugger entered--Lisp error: (wrong-type-argument symbolp "!And")
  lsp--make-message((:jsonrpc "2.0" :method "workspace/didChangeConfiguration" :params (:settings #<hash-table equal 1/65 0x1fe5877837c5>)))
[...]

PR #2807, opened 2021-04-24
Fix type of lsp-yaml-custom-tags

FredrikMeyer:
The language server expects a JSON array. This adds the correct type. (discovered in this issue https://github.com/emacs-lsp/lsp-mode/issues/2804)


yyoncho (approved):

yyoncho:
Thank you!

PR #2809, opened 2021-04-24
Responde workspace configuration with workspace buffer

### LINKED ISSUE #2806: Workspace configuration response may not use file-local variables
## Description

Under some conditions, when `find-file` is not called interactively, `lsp-mode` will not use file-local variables while responding to workspace configuration, because the `current-buffer` haven't switched to the newly opened buffer yet.

---

## Possible fix

The following patch would fix this issue, but I'm not sure if it's OK so I didn't submit a PR

```diff
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5978,8 +5978,9 @@ WORKSPACE is the active workspace."
                                                     (error-message-string err))
                                         :json-false))))
                      ((equal method "workspace/configuration")
-                      (with-lsp-workspace workspace
-                        (lsp--build-workspace-configuration-response params)))
+                      (lsp-with-current-buffer (car buffers)
+                        (with-lsp-workspace workspace
+                          (lsp--build-workspace-configuration-response params))))
                      ((equal method "workspace/workspaceFolders")
                       (let ((folders (or (-> workspace
                                              (lsp--workspace-client)
```

---

## How to reproduce

(I encountered this issue while opening files using [libvterm message parsing](https://github.com/akermu/emacs-libvterm#message-passing). Here I provide a non-realistic but simpler way to reproduce.)

(Personally I use pyright to reproduce. But it should be the same for others.)

1. Install `lsp-pyright`. Add `lsp` to `prog-mode` hook. Enable `lsp-log-io`.
2. In `/tmp/py/.dir-locals.el` : `((python-mode . ((lsp-pyright-extra-paths . ["/some/extra/path"]))))`
3. In scratch buffer, run: `(run-with-timer 1 nil #'find-file "/tmp/py/hello.py")`

Expected behavior:

"/some/extra/path" should be sent in workspace configuration

Current behavioor:

"/some/extra/path" is not sent in workspace configuration

blahgeek:
Fixes #2806

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3371,15 +3371,21 @@ disappearing, unset all the variables related to it."
    (make-lsp--registered-capability :id id :method method :options register-options?)
    (lsp--workspace-registered-server-capabilities lsp--cur-workspace)))
 
+(defmacro lsp--with-workspace-temp-buffer (workspace-root &rest body)
+  "With a temp-buffer under `WORKSPACE-ROOT' and evaluate `BODY', useful to access dir-local variables."
+  (declare (indent 1) (debug t))
+  `(with-temp-buffer
+     ;; Set the buffer's name to something under the root so that we can hack the local variables
+     ;; This file doesn't need to exist and will not be created due to this.
+     (setq-local buffer-file-name (expand-file-name "lsp-mode-temp" (expand-file-name ,workspace-root)))
+     (hack-local-variables)
+     ,@body))
+
 (defun lsp--get-ignored-regexes-for-workspace-root (workspace-root)
   "Return a list of the form (lsp-file-watch-ignored-files lsp-file-watch-ignored-directories) for the given WORKSPACE-ROOT."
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -5979,7 +5985,11 @@ WORKSPACE is the active workspace."
                                         :json-false))))
                      ((equal method "workspace/configuration")
                       (with-lsp-workspace workspace
-                        (lsp--build-workspace-configuration-response params)))
+                        (if-let ((buf (car buffers)))
+                            (with-current-buffer buf
```
yyoncho, line 5989: can you use lsp-with-current-buffer?

yyoncho (commented):
Looks good, one minor nit:

lsp-mode.el
blahgeek: fixed

yyoncho (approved):

yyoncho:
Thank you!

PR #2810, opened 2021-04-24
lsp-request: reduce the active call inside waiting loop

kiennq:
Another attempt after has been reverted by #2805 

@yyoncho Can you help check if this doesn't cause the previous issue?


yyoncho (approved):
Looks good, I was unable to reproduce the old issue.

PR #2812, opened 2021-04-25
`lsp-command-map`: change bindings to lowercase where possible

iyefrat:
Hi!

This PR changes some of the default bindings on `lsp-command-map`. In my humble opinion changing the `F` and `T` prefixes to be lowercase is a no-brainer because it doesn't even break existing bindings, and there is nothing bound under the lowercase versions. It also changes the `lsp-treemacs-sync-mode` binding from `TT` to `tt` under the same logic. Changing `G` prefix to `p` might be undesired on your end as it does break existing bindings, but I think it's more sensible as `p` is a) lowercase and b) is a memonic for "peek". 

Regradless, thank you for all your work on `lsp-mode`.

### DIFF
--- a/docs/page/keybindings.md
+++ b/docs/page/keybindings.md
@@ -25,7 +25,7 @@ configured `lsp-command-map` which is bound to `lsp-keymap-prefix`
 | `s-l t d`  | Toggle minor mode for showing hover information in child frame. (requires `lsp-ui`)                                                 |
 | `s-l t s`  | Toggle signature auto activate.                                                                                                     |
 | `s-l t f`  | Toggle on type formatting.                                                                                                          |
-| `s-l t T`  | Toggle global minor mode for synchronizing `lsp-mode` workspace folders and `treemacs` projects. (requires `lsp-treemacs`)          |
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
LGTM, but not sure this will not break existent configs that already have bindings on lowercase ? 
c/c @yyoncho

yyoncho:
Current bindings follow the spacemacs conventions(most likely doom follow the same conventions). The idea to have `T` and `F` for these commands is to have `t` for tests related methods while the `F` is used due to the fact that the folder related operations are not something that you are meant to use often, thus it has more inconvenient binding to keep the convenient bindings free.

iyefrat:
As a doom emacs user, I'm don't think these are conventions there, but it's entirely possible that I just haven't run into that part of the framework. Do LSP's have a notion of tests? From looking around I couldn't really find anything. Either way feel free to close the issue as I can always just change things around on my end.

yyoncho:
> Do LSP's have a notion of tests?

Some servers expose commands related to tests(e. g. rust analyzer, jdtls). Generally, most likely the test functionality will be filled up by dap-mode or by package extending dap-mode.

PR #2813, opened 2021-04-25
Fixed several issues discovered when testing latest changes

yyoncho:
1. Filter the buffers that have been killed before initialize has been received
2. Make sure that lsp-mode-temp is not registered as a buffer(it was breaking
spacemacs's reopen last file function
2. Cleanup some


ericdallo (approved):
Was this some regression introduced recently?

yyoncho:
Only 2)

1) was lurking for a long time(more than 2 years). That is why we have a lot of checks whether buffer is live before processing the buffers.

PR #2814, opened 2021-04-26
lsp-xml: Enable on all XML file types

foldu:
activation-fn limited it to files ending with xml, which doesn't cover
.xsd and other XML formats.


yyoncho (approved):

yyoncho:
Thank you! I will add mapping from nxml-mode to xml in lsp-language-id-configuration.

PR #2821, opened 2021-04-28
Change s bindings to w

yyoncho:
As per discussion in discord


ericdallo (approved):
Maybe we should post in #annoucements this breaking change

jcs090218 (approved):

PR #2825, opened 2021-05-01
SystemVerilog lsp client

kanadkanhere:
This adds lsp-mode support for the SystemVerilog language server called svlangserver


ericdallo (approved):
Looks good!

clients/lsp-svlangserver.el
```diff
@@ -0,0 +1,137 @@
...
+(defun lsp-clients-svlangserver-command ()
+  (list lsp-clients-svlangserver-node-command lsp-clients-svlangserver-module-path "--stdio"))
+
+(defun lsp-clients-svlangserver-get-workspace-additional-dirs (_workspace)
+  lsp-clients-svlangserver-workspace-additional-dirs)
+
+(lsp-register-client
+    (make-lsp-client :new-connection (lsp-stdio-connection 'lsp-clients-svlangserver-command)
+                     :major-modes '(verilog-mode)
+                     :priority -1
```
kiennq, line 122: The priority should be different from the priority of `lsp-verilog`.
IMO you can just move this code to `lsp-verilog.el` so we have all language sever for a language in the same file, thus easier to decide the priority there. You will not need the boiler plate comment at file header too.

kanadkanhere: That looks doable. But won't that break it for existing user since they won't have my language server installed?

kiennq: The language server will be decided by priority, in case `lsp-mode` cannot find the language server executable, the current client will be skipped and the next one (has lower priority) will be considered.

kanadkanhere: Ok. Will do. Is there a reference example for this? Specifically how to update the mkdocs.yml so that the Verilog page has entry for both (hdl checker and svlangserver)? Do I need to create it manually under docs/manual-language-docs?
Thanks

ericdallo: No, the docs check for the different registered `:group`s, so you can keep it on the same file with different defgroups and it should work

ericdallo: You can check `lsp-php.el`

kanadkanhere:
Done

clients/lsp-verilog.el
```diff
@@ -32,6 +32,117 @@
...
+  :type 'string
+  :safe (lambda (x) (stringp x)))
+
+(defcustom lsp-clients-svlangserver-workspace-additional-dirs nil
+  "Additional directories to be managed by this instance of svlangserver"
+  :group 'lsp-svlangserver
+  :type '(repeat string)
+  :safe (lambda (x) (seq-every-p #'stringp x)))
+
+(defcustom lsp-clients-svlangserver-includeIndexing '("**/*.{sv,svh}")
```
yyoncho, line 60: list of string cannot be serialized with native jsons and it will fail with emacs 27.x. You have to replace all lists that are going to be used in lsp-register-custom-settings with vectors and use lsp-string-vector type.

kanadkanhere: oh ok. I see the below in lsp-mode.el . So should I use lsp-repeatable-vector instead then?
```elisp
(define-widget 'lsp-repeatable-vector 'lsp--editable-vector
  "A variable length homogeneous vector."
  :tag "Repeat"
  :format "%{%t%}:\n%v%i\n")

(define-widget 'lsp-string-vector 'lazy
  "A vector of zero or more elements, every element of which is a string.
Appropriate for any language-specific `defcustom' that needs to
serialize as a JSON array of strings.

Deprecated. Use `lsp-repeatable-vector' instead. "
  :offset 4
  :tag "Vector"
  :type '(lsp-repeatable-vector string))

(make-obsolete 'lsp-string-vector nil "lsp-mode 7.1")
```

kanadkanhere:
done

yyoncho (approved):
Thank you!

Issue #2448, opened 2020-12-27
[3.16] Support semanticTokens missing methods


ericdallo:
[SPEC](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens)
lsp-mode is missing the implementation of:
* :leftwards_arrow_with_hook: `textDocument/semanticTokens/full/delta` witch should improve performance bringing only the tokens relative to existing ones
* :arrow_left: `workspace/semanticTokens/refresh` not critical, which allows server request client to refresh the tokens at any time.

ericdallo:
Hey @sebastiansturm willing to take a took? specially the delta request?

I implemented on clojure-lsp both range and full semantic-tokens, it's missing only the delta for full support :D

sebastiansturm:
yep, will have a look!

sebastiansturm:
if anyone would like to test this, I have a preliminary implementation at sebastiansturm/lsp-mode, but I haven't opened a PR yet as I'll need to test it with `lsp-use-plists` enabled, and with a language server that sends out refresh requests. Plan to do that later this week

ericdallo:
cool @sebastiansturm I can test it, I implemented clojure lsp server support for semantic tokens recently, but no refresh yet, but I can do if it's needed

ericdallo:
I noticed some bugs on lsp-mode regarding delaying the tokens updates and for a small amount of time the tokens just got white before apply the tokens instead of use the major mode faces, I hope this branch fixes it too!

sebastiansturm:
I'm not a clojure developer but I'll gladly use the clojure server for testing if you'll implement support, otherwise I'll try with the lua language server (as far as I understand, someone recently had issues with lua lsp sending out refresh requests, so apparently the support is there).
As for the delay, I haven't noticed that even with the current implementation, except when tree-sitter-hl mode was enabled; that's something I'll have to take a look at. Did you use tree-sitter-hl when you noticed that issue?

ericdallo:
I'm not sure when should server should send refresh notification, do you have an example?
Never heard of that tree-sitter-hl 😆

ericdallo:
Hey @sebastiansturm, good news, my issue with some tokens not being updates were gone with your branch :D
I'll keep testing it with clojure and let you know if any bugs

ericdallo:
Actually, still happens not so small files, check the gif:
[...]

PR #2826, opened 2021-05-02
[semantic-tokens] Add support for semantic tokens faces overrides

ericdallo:
Related to https://github.com/emacs-lsp/lsp-mode/issues/2448#issuecomment-826385506

- Add support for new option on client definition: `semantic-tokens-faces-overrides` with support for `:types` and `:modifiers`.
This should allow specific clients to replace the default lsp faces with custom ones.

Rationale: It's common for major modes to use different faces for their symbols, with that we could share the same faces that major modes like ` clojure-mode` uses, reducing the impact to users and following the same standard.

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -378,6 +378,23 @@ IS-RANGE-PROVIDER is non-nil when server supports range requests."
                        (lsp-warn "No face has been associated to the %s '%s': consider adding a corresponding definition to %s"
                                  category id varname)) maybe-face)) identifiers)))
 
+(defun lsp-semantic-tokens--replace-alist-values (a b)
+  "Replace alist A values with B ones if exists."
+  (-map
+   (-lambda ((ak . av))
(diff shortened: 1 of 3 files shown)


lsp-semantic-tokens.el
```diff
@@ -378,6 +378,23 @@ IS-RANGE-PROVIDER is non-nil when server supports range requests."
                        (lsp-warn "No face has been associated to the %s '%s': consider adding a corresponding definition to %s"
                                  category id varname)) maybe-face)) identifiers)))
 
+(defun lsp-semantic-tokens--replace-alist-values (a b)
+  "Replace alist A values with B ones if exists."
```
sebastiansturm, line 382: `if exists` sounds strange; not sure how to formulate this correctly without blowing up the docstring's length though. Maybe "if present" or "where available" would be slightly better?
Apart from that, the changes look good to me and seem to work as intended (tested with clojure-lsp)

sebastiansturm (approved):

PR #2833, opened 2021-05-04
Add nix-direnv to lsp-file-watch-ignored-directories

foldu:
In its default configuration, nix-direnv creates a .direnv directory next to direnv's .envrc. This directory contains no source code and links to all the projects dependencies, so it should be ignored.


yyoncho (approved):
Thank you!

PR #2837, opened 2021-05-06
set company-capf backend to buffer local

tpeacock19:
Prevents the capf backend from overtaking personal configurations in
buffers outside lsp-mode.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -740,7 +740,7 @@ The CLEANUP-FN will be called to cleanup."
              (fboundp 'company-mode))
         (setq-local company-abort-on-unique-match nil)
         (company-mode 1)
-        (add-to-list 'company-backends 'company-capf))
+        (setq-local company-backends (cl-adjoin 'company-capf company-backends :test #'equal)))
        (t
         (lsp--warn "Unable to autoconfigure company-mode.")))
 


lsp-completion.el
```diff
@@ -740,7 +740,7 @@ The CLEANUP-FN will be called to cleanup."
              (fboundp 'company-mode))
         (setq-local company-abort-on-unique-match nil)
         (company-mode 1)
-        (add-to-list 'company-backends 'company-capf))
+        (setq-local company-backends (append '(company-capf) company-backends)))
```
jcs090218, line 743: No, I don't think you should use `append` or else it would have duplicate company backends if user already has one.

tpeacock19: Right, I've changed it to use `cl-adjoin` which I believe should also respect the order of the original list as well.

lsp-completion.el
```diff
@@ -740,7 +740,7 @@ The CLEANUP-FN will be called to cleanup."
              (fboundp 'company-mode))
         (setq-local company-abort-on-unique-match nil)
         (company-mode 1)
-        (add-to-list 'company-backends 'company-capf))
+        (setq-local company-backends (cl-adjoin 'company-capf company-backends :test #'cl-equalp)))
```
kiennq, line 743: The `cl-equalp` is comparing string case-insensitively, I think you should just use `equal` here.
Or even better, since we're comparing symbol, the `:test` stuff might be omitted?

tpeacock19: I agree `equal` should be used. I had just used `cl-equalp` first. Omitting `:test` uses `eql` and that does not compare symbols. I'll update now.

tpeacock19: actually, in this case it would not make a difference. `cl-equalp` will use `equal` when comparing symbols.

wyuenho:
So this is why my company-backends are messed after turning on lsp-mode 👀

kiennq (approved):

yyoncho (approved):

PR #2838, opened 2021-05-06
Add lsp-sql-execute-paragraph

Lenbok:
This is the `lsp-mode` equivalent to `sql-send-paragraph`. Slightly generalizes `lsp-sql-execute-query` to ease construction of this and other similar equivalent functions (possibly an equivalent to `sql-send-line-and-next`?)

### DIFF
--- a/clients/lsp-sqls.el
+++ b/clients/lsp-sqls.el
@@ -85,8 +85,10 @@
       (erase-buffer)
       (insert result))))
 
-(defun lsp-sql-execute-query (&optional command)
-  "Execute COMMAND on selected region/whole buffer against current database."
+(defun lsp-sql-execute-query (&optional command start end)
+  "Execute COMMAND on buffer text against current database.
+Buffer text is between START and END.  If START and END are nil,
+use the current region if set, otherwise the entire buffer."
   (interactive)
   (lsp-sqls--show-results
    (lsp-request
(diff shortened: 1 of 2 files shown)


clients/lsp-sqls.el
```diff
@@ -99,13 +101,26 @@
...
+(defun lsp-sql-execute-paragraph (&optional command)
+  "Execute COMMAND on paragraph against current database."
+  (interactive)
+  (let ((start (save-excursion
+		 (backward-paragraph)
+		 (point)))
+	(end (save-excursion
+	       (forward-paragraph)
+	       (point))))
+    (lsp-sql-execute-query command start end)))
```
jcs090218, line 123: Can you remove tabs here and replace with spaces? And make sure the indentation is correct. I would prefer making `let` variables' lambda definition on the same line since this isn't very long. See below,

```el
  (let ((start (save-excursion (backward-paragraph) (point)))
	    (end (save-excursion (forward-paragraph) (point))))
```

Lenbok: Thanks for reviewing. Done.

yyoncho (approved):

Lenbok:
Is that single MacOS CI failure an actual problem, or some issue with the CI?

Lenbok:
@yyoncho can you please merge this?

yyoncho (approved):

yyoncho:

PR #2839, opened 2021-05-06
Elixir: Use either local build server or system binary

easbarbosa:
Hi,

Objective:  
Use local binary if found or system one.

Changes: 
- check if there is a local binary in `{EMACS}/.cache/lsp/elixir-ls/language-server.sh` and set command to it.
- if no local binary is found, use system one.

### DIFF
--- a/clients/lsp-elixir.el
+++ b/clients/lsp-elixir.el
@@ -111,10 +111,8 @@ Leave as default to let `executable-find' search for it."
   "Elixir-ls local server Directory")
 
 (defcustom lsp-elixir-local-server-command
-  (let ((file (f-join lsp-elixir-ls-server-dir
-                      (cl-first lsp-elixir-server-command))))
-    (when (f-exists? file)
-      file))
+  (f-join lsp-elixir-ls-server-dir
+          (cl-first lsp-elixir-server-command))
   "Command to start local elixir-ls binary."
   :group 'lsp-elixir
   :type '(repeat string)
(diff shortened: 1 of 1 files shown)


yyoncho:
@axelson - willing to take a look?

clients/lsp-elixir.el
```diff
@@ -105,6 +105,21 @@ Leave as default to let `executable-find' search for it."
   :type '(repeat string)
   :package-version '(lsp-mode . "7.1"))
 
+
+(defconst lsp-elixir-ls-server-dir
+  (f-join lsp-server-install-dir "elixir-ls")
+  "Elixir-ls local server Directory")
+
+(defcustom lsp-elixir-local-server-command
+  (let ((file (f-join lsp-elixir-ls-server-dir
```
kiennq, line 114: This checks for file existence whenever lsp's clients are loaded, which make `lsp-mode` loads slower when user is not using `lsp-elixir`.
Can you move this default value to somewhere around here?
``` elisp
                                     `(,(or lsp-elixir-local-server-command
                                            (or (executable-find
                                                 (cl-first lsp-elixir-server-command))
                                                (lsp-package-path 'elixir-ls)))
```

easbarbosa: oh, my bad I forgot about that.

axelson (approved):
This change makes sense to me. I didn't take a close look at the elisp, but the described algorithm sounds correct.

easbarbosa:
@axelson 

I understand that `.cache` are for transient files and `.local` should be better suited for local builds.

Nonetheless, Guix/Nix are an even better solution.

yyoncho (approved):

yyoncho:
Thank you, sorry for the delay.

PR #2844, opened 2021-05-07
From juliosueiras/terraform-lsp to hashicorp/terraform-ls

komi1230:
This PR changes juliosueiras/terraform-lsp to hashicorp/terraform-ls.

[juliosueiras/terraform-lsp](https://github.com/juliosueiras/terraform-lsp) is not actively maintained since last commit is a half year ago.

But [hashicorp/terraform-ls](https://github.com/hashicorp/terraform-ls) is really active, and also the maintainer of this is Hashi Corp.
I think this repository is stabler.

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -27,40 +27,32 @@
 (require 'lsp-mode)
 
 (defgroup lsp-terraform nil
-  "LSP support for Terraform, using terraform-lsp"
+  "LSP support for Terraform, using terraform-ls"
   :group 'lsp-mode
-  :link '(url-link "https://github.com/juliosueiras/terraform-lsp")
-  :package-version `(lsp-mode . "6.2"))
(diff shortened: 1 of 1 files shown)


juliosueiras:
@ericdallo is not a fork(the README on terraform-ls does explain it), basically terraform-lsp is less stable but more experimental features and terraform-ls is more stable but maybe less features. also terraform-ls and terraform-lsp(atleast from config) is drop-in replaceable (both have the same subcommand and args etc)

tilgovi:
Came here looking for this. I think it'd be great to use the hashicorp-provided LSP server.

tilgovi:
As a workaround, you can customize the value of `lsp-terraform-server`:

```
`(,"terraform-ls" "serve")
```

ericdallo:
@juliosueiras @tilgovi we can add hashicorp/terreform-ls as a new server along with the legacy one, and change, then soon change the default eventually removing the old one instead of just removing on this PR and causing a breaking change

PR #2850, opened 2021-05-09
fix(lsp-lua): treat boolean settings as booleans

jimeh:
This seems specifically important for the `Lua.telemetry.enable` setting. As
right now you can effectively only set it to `true` or `null` by customizing the
`lsp-lua-telemetry-enable` setting.

According to the privacy policy as of writing, the setting defaults to `null`,
which does not enable telemetry, but only asks the user if they want to
explicitly enable/disable telemetry:

https://github.com/sumneko/lua-language-server/wiki/Privacy-Policy#opt-in-policy

To quote the "Opt-in policy" section:

    By default `Lua.telemetry.enable` is set to `null`. With that
    setting, no data is sent to the development team, but UI will appear
    asking users to agree to enabling these settings. Only when a user
    explicitly opts-in will any data ever be sent.

This prompt seems to be VSCode specific though, and does not happen in Emacs.

Also, in my experience when `Lua.telemetry.enable` is set to `null`, the
language server always attempts to connect to the telemetry server
(`moe-loli.love`) on startup. After this change `Lua.telemetry.enable` is
explicitly set to `true` or `false`. Setting it to `false` causes the language
server to not make any connection attempts to the telemetry server.

As for the `Lua.hint.enable` setting, I haven't tested it, but according to the
settings schema file, it is a boolean:

https://github.com/sumneko/vscode-lua/blob/ac0b4b4d599fae1b116b3c083b312844a612df1b/setting/schema.json#L931-L936


yyoncho (approved):

yyoncho:
Thank you!

PR #2853, opened 2021-05-09
Bugfix, rust-analyzer arguments were dropped

Lunderberg:
Bug was introduced in [PR#2197](https://github.com/emacs-lsp/lsp-mode/pull/2197), which searched for the executable defined in `lsp-rust-analyzer-server-command`, but dropped any additional arguments.  This prevented running `rust-analyzer` as `rustup run nightly rust-analyzer`, if it had been [installed through rustup](https://rust-analyzer.github.io/manual.html#rustup).

While there is a workaround, to define `(setq lsp-rust-analyzer-server-args lsp-rust-analyzer-server-command)`, this conflicts with the way `lsp-rust-rls-server-command` is used, and is unexpected.  This bugfix is backwards compatible with the workaround.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -327,13 +327,6 @@ PARAMS progress report notification data."
   :group 'lsp-rust
   :package-version '(lsp-mode . "6.2"))
 
-(defcustom lsp-rust-analyzer-server-args '()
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
Thank you!

PR #2854, opened 2021-05-09
Remove duplicate highlights for iedit

geza-herman:
If a language server reports the same symbol multiple times (ccls can
behave like this for template functions, see https://github.com/MaskRay/ccls/issues/769),
 lsp-iedit will select the same region multiple times, so each modification to an area will be
duplicated.  For example, deleting a character will delete multiple
characters.  This PR filters duplicate symbols.

Note I'm not a lisp programmer, I'm not sure this is the best way to do this. But it seems to work for me.

### DIFF
--- a/lsp-iedit.el
+++ b/lsp-iedit.el
@@ -24,6 +24,7 @@
 ;;; Code:
 
 (require 'lsp-mode)
+(require 'dash)
 
 (declare-function iedit-make-occurrence-overlay "iedit-lib" (begin end))
 (declare-function iedit-start-buffering "iedit-lib" ())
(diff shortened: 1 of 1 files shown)


ericdallo:
@nbfalcon willing to take a look on this?

lsp-iedit.el
```diff
@@ -73,8 +74,29 @@ language server doesn't support renaming.
 See also `lsp-enable-symbol-highlighting'."
   (interactive)
   (let ((highlights (lsp-request "textDocument/documentHighlight"
-                                 (lsp--text-document-position-params))))
-    (lsp-iedit--on-ranges (mapcar #'lsp:document-highlight-range highlights))))
+                                 (lsp--text-document-position-params)))
+        (-compare-fn (lambda (a b)
+                       (let* ((a-range (gethash "range" a))
```
yyoncho, line 79: you should not use gethash with structures coming from the server since a might be plist when lsp-use-plist is t. You should use either `lsp-range-*` or &Range destructuring. Als, there is `lsp--position-equal`.

yyoncho: Havent tested but something like that should do:

``` elisp
(-lambda ((&Range :start l-start :end l-end) (&Range :start r-start :end r-end))
  (and (lsp--position-equal l-start l-end)
       (lsp--position-equal r-start r-end)))
```

yyoncho:
Thank you!

PR #2790, opened 2021-04-18
Address #2448

sebastiansturm:
this PR adds support for delta and refresh requests and fixes flickering seen in combination with some underlying highlighting mechanisms (tree-sitter-hl, clojure's syntax highlighter)

Fixes #2448

PR #2859, opened 2021-05-11
ensure callbacks and timers are called in "current tokenizing buffer" (fix #2858)

### LINKED ISSUE #2858: race condition in `lsp-semantic-tokens.el` (when using `desktop.el`)
**Describe the bug**
`lsp-semantic-tokens.el` has some callbacks/timers that may end up being called in a different buffer that the requesting buffer.

**To Reproduce**
This does not have a way of reproducing consistently due to many factors (using Tramp is a big one, since I'm using a far-away host). This is a description of a model, the best I could narrow the problem to:
- Use `desktop-mode` in init.el, configure it to use lazy mode (which is implemented using `run-with-idle-timer`)
- Load a (remote) file that uses lsp. More buffers like this increases the chance of hitting the race condition.
- Load a file that does not use lsp.  More buffers like this increases the chance of hitting the race condition.
- Exit Emacs (`desktop-mode` saves config)
- Open Emacs (`desktop-mode` starts restoring buffers)
- Wait (`desktop-mode` opens an lsp buffer and selects it for restoring)
- (`lsp--semantic-tokens-request` is `run-with-idle-timer`-ed)
- (another non-lsp buffer is opened and is being restored)
- (`lsp--semantic-tokens-request` is triggered in the non-lsp buffer since it's the one currently selected)
- Emacs errors because the non-lsp buffer has no lsp capability for semantic tokens (even though `lsp-mode` is not activated)

**Expected behavior**
Restoration using `desktop.el` doesn't error

Any that supports semantic tokens. Particularly [a ccls fork that supports semantic tokens](https://github.com/FelipeLema/ccls/tree/semantic-highlight)

**OS**
debian 9

**Error callstack**
Described in last step above

~~Will do a PR in the next minutes to fix this one.~~ [PR](https://github.com/emacs-lsp/lsp-mode/pull/2859)

FelipeLema:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/2858

Capture buffer that's being tokenized in closures/lambdas. This will ensure that they will be called in the proper buffer even if `(current-buffer)` changes for any reason.

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -230,15 +230,18 @@ If REGION is non-nil, it will request tokens only for given region
 otherwise it will request for whole document.
 If FONTIFY-IMMEDIATELY is non-nil, it will fontify when receive the response
 ignoring the timer."
-  (let ((request-full-token-set
-         (lambda (fontify-immediately)
-           (when lsp--semantic-tokens-idle-timer
-             (cancel-timer lsp--semantic-tokens-idle-timer))
-           (setq lsp--semantic-tokens-idle-timer
-                 (run-with-idle-timer
-                  lsp-idle-delay
-                  nil
-                  (lambda () (lsp--semantic-tokens-request nil fontify-immediately)))))))
+  (let* ((semantic-tokenizing-buffer (current-buffer))
+         (request-full-token-set
+          (lambda (fontify-immediately)
+            (when lsp--semantic-tokens-idle-timer
+              (cancel-timer lsp--semantic-tokens-idle-timer))
+            (setq lsp--semantic-tokens-idle-timer
+                  (run-with-idle-timer
+                   lsp-idle-delay
+                   nil
+                   (lambda ()
+                     (with-current-buffer semantic-tokenizing-buffer
+                       (lsp--semantic-tokens-request nil fontify-immediately))))))))
     (when lsp--semantic-tokens-idle-timer
       (cancel-timer lsp--semantic-tokens-idle-timer))
     (lsp-request-async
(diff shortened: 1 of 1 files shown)


ericdallo:
Not sure if @sebastiansturm already fixed it on #2790

sebastiansturm:
thanks for the fix, I'll have a look tomorrow at the latest, or perhaps this evening

lsp-semantic-tokens.el
```diff
@@ -230,15 +230,19 @@ If REGION is non-nil, it will request tokens only for given region
...
+         (request-full-token-set
+          (lambda (fontify-immediately)
+            (when lsp--semantic-tokens-idle-timer
+              (cancel-timer lsp--semantic-tokens-idle-timer))
+            (setq lsp--semantic-tokens-idle-timer
+                  (run-with-idle-timer
+                   lsp-idle-delay
+                   nil
+                   (lambda ()
+                     (when (buffer-live-p semantic-tokenizing-buffer)
```
sebastiansturm, line 243: this change looks reasonable to me, I'm actually doing the same in my pending semantic-tokens PR, so to the extent of my current knowledge I'd assume it to be correct :)

lsp-semantic-tokens.el
```diff
@@ -250,12 +254,14 @@ ignoring the timer."
      `( :textDocument ,(lsp--text-document-identifier)
         ,@(if region (list :range (lsp--region-to-range (car region) (cdr region))) '()))
      (lambda (response)
-       (setq lsp--semantic-tokens-cache response)
-       (lsp-put lsp--semantic-tokens-cache :_documentVersion lsp--cur-version)
-       (lsp-put lsp--semantic-tokens-cache :_region region)
-       (when fontify-immediately (font-lock-flush))
-       ;; request full token set to improve fontification speed when scrolling
-       (when region (funcall request-full-token-set nil)))
+       (when (buffer-live-p semantic-tokenizing-buffer)
```
sebastiansturm, line 257: did you run into a situation where this change made a difference? Because I think `:mode 'tick` should already evaluate the callback within the calling buffer, or cancel it altogether if the buffer is killed inbetween?

FelipeLema: I remember adding `buffer-live-p` because the following `with-current-buffer` failed indeed.

I don't remember the details right now, but I remember I _was_ pushing the limits for this code when I got that error. I personally say the cost is low (a little harder to read, more to think about) and the benefit is huge (user is not confounded by a hard-to-reproduce error), so this is worth in the long run.

sebastiansturm: I think it's redundant, so I'd prefer removing it, thanks!

FelipeLema: done

sebastiansturm: LGTM, thanks!

sebastiansturm (approved):

yyoncho (approved):

Issue #2857, opened 2021-05-11
Empty completion after space


gexplorer:
If **space** is among a servers **triggerCharacters** for completion, it does not return any results after a **space**.

This happens because `lsp-completion--get-context` returns `CompletionTriggerKind = 2` when completion is invoked manually but the previous character is one of **triggerCharacters**.

## Steps to Reproduce

See [example project](https://github.com/gexplorer/veturpack)

* Inside a **vue** file

### Failed completion

* Try completion with cursor separated one **space** from the curly bracket:
```js
import { | } from 'vue';
```
* The request is:
```json
Params: {
  "textDocument": {
    "uri": "file:///Users/eder/Projects/veturpack/src/Test.vue"
  },
  "position": {
    "line": 7,
    "character": 9
  },
  "context": {
    "triggerKind": 2,
    "triggerCharacter": " "
  }
}
```
* It won't return results

### Correct completion

* Try completion with cursor right after the curly bracket
```js
import {|  } from 'vue';
```
* The request is:
```json
Params: {
  "textDocument": {
    "uri": "file:///Users/eder/Projects/veturpack/src/Test.vue"
  },
  "position": {
    "line": 7,
    "character": 8
  },
  "context": {
    "triggerKind": 1
  }
}
```
* It returns the expected candidates

## Expected behavior
Completing after a space should yield the same results as right after the bracket. When completion is triggered manually the `CompletionTriggerKind` parameter sent to the server should be 1, regardless of the previous character.

## Which Language Server did you use
I encountered the problem with the **lsp-vetur** server which has **space** as a trigger character for completion.

## OS
macOS

kiennq:
Let me see, perhaps in the case of triggered by trigger characters, we can set some variable to distinguish that

gexplorer:
I found a possible solution and I've been using it for a few hours without any problems.

Basically before looking back for trigger characters I check that the last executed command has been `self-insert-command` which AFAIK means that the user has inserted a char.

```elisp
(defun lsp-completion--get-context (trigger-characters)
  "Get completion context with provided TRIGGER-CHARACTERS."
  (let* ((triggered-by-char (equal last-command 'self-insert-command))
         (trigger-char (if triggered-by-char
                           (lsp-completion--looking-back-trigger-characterp
                            trigger-characters)))
         (trigger-kind (cond
                        (trigger-char
                         lsp/completion-trigger-kind-trigger-character)
[...]

PR #2864, opened 2021-05-12
lsp-completion: Fix triggerKind on manual invokation

gexplorer:
Related to issue #2857.

If **space** is among a servers **triggerCharacters** for completion, it does not return any results after a **space**.

This happens because `lsp-completion--get-context` returns `CompletionTriggerKind = 2` when completion is invoked manually but the previous character is one of **triggerCharacters**.

## Steps to Reproduce

See [example project](https://github.com/gexplorer/veturpack)

* Inside a **vue** file

### Failed completion

* Try completion with cursor separated one **space** from the curly bracket:
```js
import { | } from 'vue';
```
* The request is:
```json
Params: {
  "textDocument": {
    "uri": "file:///Users/eder/Projects/veturpack/src/Test.vue"
  },
  "position": {
    "line": 7,
    "character": 9
  },
  "context": {
    "triggerKind": 2,
    "triggerCharacter": " "
  }
}
```
* It won't return results

### Correct completion

* Try completion with cursor right after the curly bracket
```js
import {|  } from 'vue';
```
* The request is:
```json
Params: {
  "textDocument": {
    "uri": "file:///Users/eder/Projects/veturpack/src/Test.vue"
  },
  "position": {
    "line": 7,
    "character": 8
  },
  "context": {
    "triggerKind": 1
  }
}
```

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -370,10 +370,12 @@ The MARKERS and PREFIX value will be attached to each candidate."
 
 (defun lsp-completion--get-context (trigger-characters)
   "Get completion context with provided TRIGGER-CHARACTERS."
-  (let* (trigger-char
+  (let* ((triggered-by-char (equal last-command 'self-insert-command))
+         (trigger-char (when triggered-by-char
+                           (lsp-completion--looking-back-trigger-characterp
+                            trigger-characters)))
          (trigger-kind (cond
-                        ((setq trigger-char (lsp-completion--looking-back-trigger-characterp
-                                             trigger-characters))
+                        (trigger-char
                          lsp/completion-trigger-kind-trigger-character)
                         ((equal (cl-second lsp-completion--cache) :incomplete)
                          lsp/completion-trigger-kind-trigger-for-incomplete-completions)
--- a/test/lsp-completion-test.el
+++ b/test/lsp-completion-test.el
@@ -26,6 +26,7 @@
 
 (require 'lsp-completion)
 (require 'ert)
+(require 'el-mock)
 
 (ert-deftest lsp-completion-test-candidate-kind ()
   (should (eq (lsp-completion--candidate-kind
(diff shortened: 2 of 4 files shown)


gexplorer:
Hey @kiennq do you know about the errors in the build? I'm not sure they are related to what I changed 😖

yyoncho:
> Hey @kiennq do you know about the errors in the build? I'm not sure they are related to what I changed

Errors seem to be unrelated.

kiennq:
> Hey @kiennq do you know about the errors in the build? I'm not sure they are related to what I changed 😖

We're not using Cask on Windows so I think you will need to update `windows_boostrap.el` to include `el-mock`

lsp-completion.el
```diff
@@ -370,10 +370,12 @@ The MARKERS and PREFIX value will be attached to each candidate."
 
 (defun lsp-completion--get-context (trigger-characters)
   "Get completion context with provided TRIGGER-CHARACTERS."
-  (let* (trigger-char
+  (let* ((triggered-by-char (equal last-command 'self-insert-command))
+         (trigger-char (if triggered-by-char
```
kiennq, line 374: you can using `when` instead of `if` here

test/lsp-completion-test.el
```diff
@@ -79,5 +80,16 @@
              '("F" "daFo" "safo")
              '("F" "daFo" "safo"))))
 
+(ert-deftest lsp-completion-test-get-context ()
+  (cl-labels ((do-test (command)
+                       (let ((last-command command))
+                         (hash-table-values (lsp-completion--get-context '("_"))))))
```
kiennq, line 86: you should use something else other than `hash-table` directly here. When we switch to using `plist`, this will become plist.
@yyoncho Do you have any suggestion on which function should we use?

yyoncho: > @yyoncho Do you have any suggestion on which function should we use?

I guess one of `lsp:completion-context-trigger-kind` `lsp:completion-context-trigger-character?` should do?

gexplorer:
retest this please

kiennq (approved):

kiennq:
Thank you

PR #2868, opened 2021-05-16
[lens] Add `lsp-lens-place-position`

ericdallo:
This should allow lsp-lens to be placed at end of line like neovim and IDEA:

Before (`'above-line`):
image

After (`'end-of-line`):
image

Note: I kept the default as `'above-line` but maybe we should consider changing the default before release lens as enabled to users as the `'above-line` adds an extra unnecessary line on the lens position as you can see in the print.

### DIFF
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -88,8 +88,10 @@ Results are meaningful only if FROM and TO are on the same line."
                                        (save-excursion
                                          (beginning-of-line-text)
                                          (point))))
-         (str (concat (make-string offset ?\s)
-                      (overlay-get ov 'lsp--lens-contents))))
+         (str (if (eq 'end-of-line lsp-lens-place-position)
+                  (overlay-get ov 'lsp--lens-contents)
+                (concat (make-string offset ?\s)
+                        (overlay-get ov 'lsp--lens-contents)))))
     (save-excursion
       (goto-char (overlay-start ov))
       (if (eq 'end-of-line lsp-lens-place-position)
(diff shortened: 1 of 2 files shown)


jcs090218 (approved):
Awesome! 👍

lsp-lens.el
```diff
@@ -82,18 +89,21 @@ Results are meaningful only if FROM and TO are on the same line."
                                          (beginning-of-line-text)
                                          (point))))
          (str (concat (make-string offset ?\s)
```
yyoncho, line 91: Hm, I think that offset should not be used when lens type = end-of-line

yyoncho:
> Note: I kept the default as `'above-line` but maybe we should consider changing the default before release lens as enabled to users as the `'above-line` adds an extra unnecessary line on the lens position as you can see in the print.

The consensus is to match vscode defaults.

lsp-lens.el
ericdallo: I didn't notice any changes, but I agree it's unnecessary, done!

yyoncho: It will make a difference when the offset is != 0.

yyoncho (approved):

yyoncho:
> > Note: I kept the default as `'above-line` but maybe we should consider changing the default before release lens as enabled to users as the `'above-line` adds an extra unnecessary line on the lens position as you can see in the print.
> 
> The consensus is to match vscode defaults.

We can revisit that... I mean, vscode does above lenses better than us so in the end of the day we might use end-of-line.

ericdallo:
Yeah I noticed vscode has the same  "issue" with the extra line:
image

I just don't like that because when you page-up/down buffer you notice the buffer increasing and changing code's location because of that extra line on every code lenses, but that could be a personal opinion though

PR #2872, opened 2021-05-18
Don't assume web-mode is html

### LINKED ISSUE #2044: web-mode,when open .scss file, lsp auto load html-ls instead of css-ls
image

billop:
This PR makes it so that web-mode buffers aren't assumed to be html. Rather, the file extension is checked so that css and js files invoke the appropriate lsp clients instead of `html-ls`.

Would close #2044


jcs090218 (approved):
I think changes make sense? LGTM.

yyoncho (approved):

PR #2878, opened 2021-05-21
Fix broken link

popoppo:
The link of `settings` has been broken.
I'm not 100% sure if the fixed link is correct, but it works for now.

Thanks for your review.


ericdallo (approved):
Thanks for that, I didn't notice that !

popoppo:
Thanks!! I always appreciate your great work ✨

PR #2882, opened 2021-05-22
Expose lsp-client-download-server-fn

### LINKED ISSUE #2879: No clear, discoverable option via defcustom for disabling server auto-install
**Describe the bug**

A previous PR tried to address this (#1620) but it was shot down because it can be hacked around by changing private things in lsp-mode. This was also not documented anywhere besides in that PR. I don't think that's a desirable solution, and I don't imagine it will continue to work as long as it's undocumented private API.

Let me describe why I think this is a problem that deserves a bit more than a private hack:

---

Whenever I open a file that I have enabled lsp for (because I do sometimes write projects in the language, i.e. anything that launches `sh-mode`), but doesn't require a language server in that instance, I am asked which language server I want to automatically install.

Though it's nice that this usually fails to work anyway because I rarely have `npm` on my path, I never want it to happen in the first place; yet there is seemingly no way to say "no thanks" (though this may be due to helm), and no way to disable the prompt in the future (or globally).

This is also problematic when offline, or behind a restrictive firewall, because it will prompt to install and just fail. I don't even want to think about how this affects users on a metered connection, or in an organization that will escalate incidents of un-permitted download attempts.

If I want to use a language server, I will install it using my distro package manager that will ensure updates, or through the project package manager that respects its lockfile. Not uncontrollably to a directory in `~/.config/emacs`. This is exacerbated when using [NixOS](https://nixos.org/), where language servers are only on `$PATH` in specific project directories of my choosing, turning almost every file type I open that happens to sometimes be part of a project into an annoyance.

---

Sorry for the slightly ranty report. If it's any excuse, I just spent 10 minutes tapping enter and waiting for a failure while `desktop-mode` opened a bunch of disparate shell scripts while starting Emacs, so I'm possibly not as concise as I should be.

The feature is probably really cool if you don't care too much where, when and how software is installed, but I think being unable to disable it is a big oversight that isn't taken as seriously as it should.

I absolutely love this package otherwise, it makes it really easy to get good dev environments to work, so I feel somewhat guilty bringing this up in this way. Thanks for all the effort!

**To Reproduce**

Open any file that lsp is enabled for without having a server installed and be annoyed.

**Expected behavior**

This is about a missing feature in a way? A configuration option to always disable installation should be present.

The language in question in this particular case was bash, but no language server was used (since it can't be installed without `npm`).

**OS**

NixOS 20.09 (Nightingale)

**Error callstack**

N/A, this is a UX bug.

yyoncho:
- Fixes #2879

### DIFF
--- a/docs/page/faq.md
+++ b/docs/page/faq.md
@@ -33,3 +33,17 @@
 
 - When using `clojure` with `clojure-lsp` `lsp-mode` is interfering with typing (e. g. deleting whitespace while typing), how to fix that?
     - The issue is caused by `clojure-lsp` server being more aggressive with formatting(deleting whitespaces) and `aggressive-indent-mode` being on and calling it while typing. The solution is either to disable `aggressive-indent-mode` or to disable `lsp-mode` formatting via `lsp-enable-indentation`.
+
+- How do I disable automatic installation for particular language server?
+    - Solution:
+    Disable for one client:
+    ``` elisp
+    (with-eval-after-load 'lsp-bash
+       (setf (lsp-client-download-server-fn (gethash 'bash-ls lsp-clients)) nil))
+    ```
+    Disable for all clients:
+    ``` elisp
+    (add-hook 'lsp-mode-hook
+       (lambda () (mapc (lambda (client) (setf (lsp-client-download-server-fn client) nil))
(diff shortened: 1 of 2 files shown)


ericdallo (approved):
Nice!

PR #2883, opened 2021-05-22
Facilities for finding use-plist inconsistencies

yyoncho:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/1835

First step before migrating to plists. Next we should add
`(lsp-consistency-check <package>)` for all downstream deps. The last step is to
switch to plists and uncomment the section in this PR which will do the actual
check.


ericdallo (approved):
Looks like a good approach, I imagine when we uncomment and enable plists as default, it should affect a lot of users, right?

yyoncho:
> Looks like a good approach, I imagine when we uncomment and enable plists as default, it should affect a lot of users, right?

most likely everyone who is updating their packages via package.el.

PR #2887, opened 2021-05-24
Add lsp-face-semhl-method to lsp-semantic-token-faces

JamieCunliffe:
I noticed that the method face wasn't added to `lsp-semantic-token-faces` which stopped the face from being used.


ericdallo (approved):

jcs090218 (approved):

jcs090218:
Thanks for contributing to `lsp-mode`! 🎉

PR #2898, opened 2021-05-28
Support additional `gopls` settings

schellj:
Add support for additional as-yet unsupported `gopls` settings (`analyses`, `importShortcut`, `symbolMatcher`, `symbolStyle`).

### DIFF
--- a/clients/lsp-go.el
+++ b/clients/lsp-go.el
@@ -82,7 +82,7 @@ completing function calls."
   'lsp-go-env
   "lsp-mode 7.0.1")
 
-(defcustom lsp-go-env (make-hash-table)
+(defcustom lsp-go-env nil
   "`gopls' has the unusual ability to set environment variables,
   intended to affect the behavior of commands invoked by `gopls'
   on the user's behalf. This variable takes a hash table of env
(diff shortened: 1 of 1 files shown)


clients/lsp-go.el
```diff
@@ -257,6 +257,52 @@ $GOPATH/pkg/mod along with the value of
   :group 'lsp-go
   :package-version '(lsp-mode "7.1"))
 
+(defcustom lsp-go-analyses (make-hash-table)
```
yyoncho, line 260: the default is not good - hash-table is not alist. You should either change the default or the type of the defcustom.

schellj: Sorry. Pushed a commit to change the default to `nil`. I also changed the default for `lsp-go-env`, as it had been `(make-hash-table)` also.

Tangentially, I'm not sure how to set a member of this setting to `false`. `'((unreachable . t))` seems to work to set `unreachable` to `true`, but `'((unreachable . nil))` or `'((unreachable . ()))` results in a `parsing setting "analyses": invalid type bool for map key "unreachable"` error, which appears to be coming from here: https://github.com/golang/tools/blob/5ab822f631a3f7b79c55b1e1dd18978ed2fcc312/internal/lsp/source/options.go#L1077. I'm not sure how to see the JSON that's actually being sent to `gopls`, but that error suggests that it's not `false`.

yyoncho:
Thank you!

clients/lsp-go.el
yyoncho: use :json-false

PR #2903, opened 2021-05-30
Add rust-analyzer.cargo.target setting

yyoncho:


jcs090218 (approved):

yyoncho:
cc @flodiebold

clients/lsp-rust.el
```diff
@@ -175,6 +175,12 @@ the latest build duration."
   :group 'lsp-rust
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-rust-analyzer-cargo-target nil
+  "Compilation target (target triple)."
+  :type 'lsp-string-vector
```
flodiebold, line 180: should be a string or nil

yyoncho: Fixed and force pushed.

PR #2916, opened 2021-06-05
Add `lsp-steep-server-path` custom variable

ybiquitous:
This change adds a new custom variable to the Steep client: `lsp-steep-server-path`.
This variable aims to help users who want to specify any `steep` executable path.

For example, the [ruby/rbs](https://github.com/ruby/rbs) repository uses the `bin/steep` script,
instead of `bundle exec steep` or globally `steep`.

In such a case, I think users want to custom the variable like this:

```elisp
;; .dir-locals.el
((enh-ruby-mode . ((lsp-steep-server-path . "/Users/ybiquitous/git/rbs/bin/steep"))))
```

### DIFF
--- a/clients/lsp-steep.el
+++ b/clients/lsp-steep.el
@@ -47,12 +47,17 @@
   :safe #'booleanp
   :group 'lsp-steep)
 
+(defcustom lsp-steep-server-path nil
+  "Path of the Steep language server executable.
+If specified, `lsp-steep-use-bundler' is ignored."
+  :type 'file
+  :group 'lsp-steep)
+
 (defun lsp-steep--build-command ()
   "Build a command to start the Steep language server."
   (append
-    (if lsp-steep-use-bundler '("bundle" "exec"))
-    '("steep" "langserver")
-    (list (concat "--log-level=" lsp-steep-log-level))))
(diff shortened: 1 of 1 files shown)


ybiquitous:
I‘m not sure why CI fails... 🤷🏼

yyoncho:
> I‘m not sure why CI fails... 🤷🏼

It is infra issue

clients/lsp-steep.el
```diff
@@ -47,12 +47,17 @@
   :safe #'booleanp
   :group 'lsp-steep)
 
+(defcustom lsp-steep-server-path nil
+  "Path of the Steep language server executable.
+If specified, `lsp-steep-use-bundler' is ignored."
+  :type 'file
```
yyoncho, line 53: can you add version 7.1 as we do for the other properties?

ybiquitous: Sure. 👍🏼 
Is it desirable to add `:package-version \`(lsp-mode . "7.1.0")` also to the existing variables `lsp-steep-log-level` and `lsp-steep-use-bundler`?

ybiquitous: Done 638c0200. It seems good to add 7.1 only to the new variable.

yyoncho (approved):

ybiquitous:
@yyoncho Thank you for the merge! 😄

PR #2919, opened 2021-06-08
Add .yarn folder to lsp-file-watch-ignored-directories

Sorixelle:
Similar to node_modules, but for the Yarn 2+ package manager.

Solves performance issues in projects using Yarn 2+ due to too many file watches being created. No reason anything in this folder should be watched - it only contains Node package dependencies + some other metadata Yarn 2+ uses.


yyoncho (approved):

yyoncho:
Thank you!

PR #2921, opened 2021-06-09
Make sure lsp--describe-buffer isn't persisted after buffer switch

yyoncho:


schellj:
For posterity, I was running into undefined lsp keybindings in buffers associated with a given language server where an action was actually supported. I use `which-key`, both generally and with `lsp-enable-which-key-integration`. Turned out that if the `which-key` info minibuffer-popup had been invoked in some other buffer where the action was not supported, the related keybinding in the original buffer would be undefined. I've confirmed that the change in this PR resolves that issue.

Thanks!

ericdallo (approved):

kiennq (approved):

PR #2922, opened 2021-06-10
Add Elixir deps to ignored directories

julianrubisch:


yyoncho:
deps seems to be too common and other servers might be interested in monitoring that folder...

julianrubisch:
Gotcha, thanks!

PR #2925, opened 2021-06-12
Handle line,column in the url

### LINKED ISSUE #2924: DocumentLink: navigate to the position specified by range field
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)?
No

Describe your feature here.

image

`textDocument/documentLink` request returns a list of [DocumentLink](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#documentLink), which are highlighted using an underline. When I click the link, I expected it to take me to the start of the `range` specified in the documentLink, but it just goes to the file specified by the `target`, but not to the position specified by the `range` field. Moving to the specific position would make it very useful in cases like JSON schema file example above.

yyoncho:
fixes #2924

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4476,7 +4476,12 @@ Applies on type formatting."
   (let* ((parsed-url (url-generic-parse-url (url-unhex-string url)))
          (type (url-type parsed-url)))
     (pcase type
-      ("file" (find-file (lsp--uri-to-path url)))
+      ("file"
+       (find-file (lsp--uri-to-path url))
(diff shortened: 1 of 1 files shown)


ananthakumaran:
looks like it's following 1 based index. Both line number and column number are off by +1

yyoncho:
updated the PR

ericdallo (approved):
I didn't test it/don't know a server which supports it but LGTM

Issue #2863, opened 2021-05-11
Add support for beancount


wangkev:
There exists a [language server](https://github.com/polarmutex/beancount-language-server) for beancount, can we please add support for it in `lsp-mode`?

This was also discussed https://github.com/hlissner/doom-emacs/issues/5035#issuecomment-839161479.

jcs090218:
The fastest way to get this to lsp-mode is to open a PR for this. I believed the community would love to see this implemented.

jcs090218:
@wangkev We have implemented the basic support for beancount language server, see #2926.

Close this now! Thanks!

wangkev:
Thanks!

PR #2926, opened 2021-06-13
Add support for beancount

jcs090218:
For #2863

### DIFF
--- a/clients/lsp-beancount.el
+++ b/clients/lsp-beancount.el
@@ -38,16 +38,14 @@
   :group 'lsp-beancount
   :package-version '(lsp-mode . "7.1.0"))
 
-(defcustom lsp-beancount-journal-file nil
-  "Pathg to Beancount journal file."
+(defcustom lsp-beancount-python-interpreter nil
+  "Path to Python executable."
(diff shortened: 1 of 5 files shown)


clients/lsp-beancount.el
```diff
@@ -0,0 +1,68 @@
...
+  :group 'lsp-beancount
+  :package-version '(lsp-mode . "7.1.0"))
+
+(defcustom lsp-beancount-journal-file nil
+  "Pathg to Beancount journal file."
+  :type 'string
+  :group 'lsp-beancount
+  :package-version '(lsp-mode . "7.1.0"))
+
+(defcustom lsp-beancount-python-interpreter
```
yyoncho, line 47: can you leave that `nil` and then do the executable-find in `lambda` in lsp-stdio-connection? This will avoid calling executable-find which will scan the classpath when loading the packages.

yyoncho (approved):
Looks good! One minor nit:

clients/lsp-beancount.el
jcs090218: Sure. Make sense.

PR #2935, opened 2021-06-18
Add lsp-load/save-vscode-workspace

yyoncho:
- related to gitpod effort, we need standartized way to load workspace folders

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1826,6 +1826,35 @@ regex in IGNORED-FILES."
 ;;  lsp-downstream-deps)
 
 
+;; loading code-workspace files
+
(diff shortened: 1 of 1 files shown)


ericdallo (approved):

kiennq (approved):

kiennq:
Should we use this one as default instead of `.lsp-session-v1`?

jcs090218 (approved):

yyoncho:
> Should we use this one as default instead of `.lsp-session-v1`?

I like that idea. We store a bit more data, (e. g. multi root folders -> language server association). We can put that in the settings section. I will file a separate issue to track that effort.

PR #2940, opened 2021-06-18
Add Roblox Lua Language Server

glinesbdev:
Hey all :wave:

This PR bases the Roblox Lua LS on the `lua-language-server` from `sumneko`, which is already a client of `lsp-mode`. The Roblox LS source code is actually based on the work from `sumneko`, as well. The Roblox Lua LS can be found here: https://github.com/NightrainsRbx/RobloxLsp

This PR adds the lsp-mode Roblox Lua client as well as the client configuration in `docs/lsp-clients.json`. I ran this code in my local emacs installation and didn't find any issues.

Thanks!

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -631,14 +631,14 @@ and `../lib` ,exclude `../lib/temp`.
 
 (defcustom lsp-lua-roblox-server-download-url
   (lsp-vscode-extension-url "Nightrains" "robloxlsp" "0.15.8")
-  "Download url for Roblox Lua vscode extension"
+  "Download url for Roblox Lua vscode extension."
   :group 'lsp-lua-roblox-language-server
   :version "7.1"
   :type 'string)
 
-(defcustom lsp-lua-roblox-server-store-url
+(defcustom lsp-lua-roblox-server-store-path
   (expand-file-name "vs-lua-roblox" lsp-lua-roblox-language-server-install-dir)
-  "Server file name for the vscode extension"
+  "Server file name for the vscode extension."
   :group 'lsp-lua-roblox-language-server
   :version "7.1"
   :type 'string)
@@ -648,13 +648,17 @@ and `../lib` ,exclude `../lib/temp`.
   (and (f-exists? lsp-lua-roblox-language-server-main-location)
        (f-exists? lsp-lua-roblox-language-server-bin)))
 
-(lsp-dependency
- 'lua-roblox-lsp
- `(:download :url lsp-lua-roblox-server-download-url
-   :store-path lsp-lua-roblox-server-store-url
-   :set-executable? t
-   :decompress :zip)
- '(:system "lua-roblox-lsp"))
+(defun lsp-lua-roblox-language-server-install (_client callback error-callback _update?)
+  "Download the latest version of lua-language-server and extract it to
+`lsp-lua-roblox-language-server-download-url'."
+  (lsp-download-install
+    (lambda (&rest _)
+     (set-file-modes lsp-lua-roblox-language-server-bin #o0700)
+     (funcall callback))
+     error-callback
+     :url lsp-lua-roblox-server-download-url
+     :store-path lsp-lua-roblox-server-store-path
+     :decompress :zip))
 
 (lsp-register-client
  (make-lsp-client
@@ -666,8 +670,7 @@ and `../lib` ,exclude `../lib/temp`.
   :major-modes '(lua-mode)
   :priority -4
   :server-id 'lua-roblox-language-server
-  :download-server-fn (lambda (_client callback error-callback _update?)
-                        (lsp-package-ensure 'lua-roblox-lsp callback error-callback))))
+  :download-server-fn #'lsp-lua-roblox-language-server-install))
 
 (lsp-consistency-check lsp-lua)
 
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -593,72 +593,80 @@ and `../lib` ,exclude `../lib/temp`.
   :server-id 'lsp-lua-lsp))
 
 ;;; lua-roblox-language-server
-(defgroup lsp-roblox-lua-language-server nil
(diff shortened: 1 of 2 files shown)


clients/lsp-lua.el
```diff
@@ -592,6 +592,73 @@ and `../lib` ,exclude `../lib/temp`.
...
   :server-id 'lsp-lua-lsp))
 
+;;; lua-roblox-language-server
+(defgroup lsp-roblox-lua-language-server nil
+  "Roblox Lua LSP client, provided by the Roblox Lua Language Server."
+  :group 'lsp-mode
+  :version "7.1"
+  :link '(url-link "https://github.com/NightrainsRbx/RobloxLsp"))
+
+(defcustom lsp-clients-roblox-lua-language-server-install-dir (f-join lsp-server-install-dir "roblox-lua-language-server/")
```
kiennq, line 602: nit: the `lsp-clients` prefix is the relic from previous `lsp-clients` which has been refactored out.
You can skip the `clients` part here and use `lsp-roblox` directly (or use `lsp-lua-roblox` as prefix since the file name is `lsp-lua.el` already)

clients/lsp-lua.el
```diff
@@ -592,6 +592,73 @@ and `../lib` ,exclude `../lib/temp`.
...
+
+(defun lsp-clients-roblox-lua-language-server-test ()
+  "Test Lua language server binaries and files."
+  (and (f-exists? lsp-clients-roblox-lua-language-server-main-location)
+       (f-exists? lsp-clients-roblox-lua-language-server-bin)))
+
+(defun lsp-roblox-lua-language-server-install (client callback error-callback update?)
+  "Download the latest version of roblox-lua-language-server and extract it to
+`lsp-roblox-lua-language-server-install-dir'."
+  (ignore client update?)
```
kiennq, line 640: You can add `_` prefix to `_client` `_update` for it to be ignored instead of this

clients/lsp-lua.el
```diff
@@ -592,6 +592,73 @@ and `../lib` ,exclude `../lib/temp`.
...
+  :version "7.1"
+  :risky t
+  :type 'file)
+
+(defun lsp-clients-roblox-lua-language-server-test ()
+  "Test Lua language server binaries and files."
+  (and (f-exists? lsp-clients-roblox-lua-language-server-main-location)
+       (f-exists? lsp-clients-roblox-lua-language-server-bin)))
+
+(defun lsp-roblox-lua-language-server-install (client callback error-callback update?)
```
kiennq, line 637: You can take a look at `lsp-clojure` and define something similar to
``` elisp
(lsp-dependency
 'clojure-lsp
 `(:download :url lsp-clojure-server-download-url
             :decompress :zip
             :store-path lsp-clojure-server-store-path
             :set-executable? t)
 '(:system "clojure-lsp"))
```
After that, the `:download-server-fn` can be just
``` elisp
(lambda (_client callback error-callback _update?)
                        (lsp-package-ensure 'roblox-lua-language-server callback error-callback))
```

kiennq (approved):

kiennq:
Is it still working okay with new change? Especially around the set executable file modes for the language server bin?

glinesbdev:
> Is it still working okay with new change? Especially around the set executable file modes for the language server bin?

@kiennq yes, during my local testing it was working but only if I excluded `:set-executable? t`. If I set that, then it would give an error which was that the file in `lsp/lsplua-roblox/extension/server/bin/macOS/lua-language-server` was not executable i.e. a command like `chmod +x lua-language-server` wasn't run on the file.

jcs090218 (approved):

glinesbdev:
> Is it still working okay with new change? Especially around the set executable file modes for the language server bin?

@kiennq, I was doing some more testing and lsp doesn't set the executable into a state in which it can use it and lsp will throw an error upon installation. If I manually go to the executable file and run `chmod +x ./lua-language-server` then it works. I updated the code to include `:set-executable t` like you suggested but still no luck. Thoughts?

kiennq:
> I updated the code to include `:set-executable t` like you suggested but still no luck. Thoughts?

I think you need to modify `lsp-mode` for this, since the `lsp-download-path` only set the executable flag for the `:store-path` (in this case the folder where you expand the package).

You can probably change `lsp-mode` like this
``` diff
diff --git a/lsp-mode.el b/lsp-mode.el
index 0c5f185a..89d0c7a5 100644
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7473,13 +7473,16 @@ nil."
 (cl-defun lsp-download-path (&key store-path set-executable? &allow-other-keys)
   "Download URL and store it into STORE-PATH.

-SET-EXECUTABLE? when non-nil change the executable flags of
-STORE-PATH to make it executable."
-  (let ((store-path (lsp-resolve-value store-path)))
+SET-EXECUTABLE? when non-nil change the executable flags of its value to
+make it executable.  When its value is t, change the executable flags of
+STORE-PATH instead."
+  (let* ((store-path (lsp-resolve-value store-path))
+         (exe-path (if (and (booleanp set-executable?) set-executable?) store-path
+                      (lsp-resolve-value set-executable?))))
     (cond
-     ((executable-find store-path) store-path)
-     ((and set-executable? (f-exists? store-path))
-      (set-file-modes store-path #o0700)
+     ((executable-find exe-path) store-path)
+     ((and set-executable? (f-exists? exe-path))
+      (set-file-modes exe-path #o0700)
       store-path)
      ((f-exists? store-path) store-path))))
```

And set the value of `:set-executable?` to `lsp-clients-roblox-lua-language-server-bin` instead of `t`

glinesbdev:
@kiennq after implementing the changes you suggested, I could never get the `(set-file-modes)` line to run. When I evaluated that same code with the same values, it would work; just not during the download / installation process. I reverted the `:download-server-fn` for this addition and opted to not use the `lsp-dependency` mechanism for now to get it into a working state.

kiennq (approved):

PR #2950, opened 2021-06-22
Automatic installation for clangd

yyoncho:
Tested only on linux. After the merge I will update cpp tutorial.

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -47,6 +47,45 @@
 (declare-function flycheck-error-group "ext:flycheck" (err) t)
 (declare-function flycheck-error-message "ext:flycheck" (err) t)
 
+(defcustom lsp-clangd-version "12.0.0"
+  "Clangd version to download.
+It has to be set before `lsp-clangd.el' is loaded and it has to
+be available here: https://github.com/clangd/clangd/releases/"
+  :type 'string
+  :group 'lsp-clangd
+  :package-version '(lsp-mode . "7.1"))
+
+(defcustom lsp-clangd-download-url
+  (format (pcase system-type
+            ('gnu/linux "https://github.com/clangd/clangd/releases/download/%s/clangd-linux-%s.zip")
+            ('darwin "https://github.com/clangd/clangd/releases/download/%s/clangd-mac-%s.zip")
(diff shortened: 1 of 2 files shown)


clients/lsp-clangd.el
```diff
@@ -47,6 +47,35 @@
 (declare-function flycheck-error-group "ext:flycheck" (err) t)
 (declare-function flycheck-error-message "ext:flycheck" (err) t)
 
+(defcustom lsp-clangd-download-url
+  (pcase system-type
+    ('gnu/linux "https://github.com/clangd/clangd/releases/download/12.0.0/clangd-linux-12.0.0.zip")
```
ericdallo, line 52: Maybe have a different defcustom for the version? to avoid users changing both here and the `lsp-clangd-binary-path` too

lsp-mode.el
```diff
@@ -7502,12 +7502,13 @@ nil."
              (funcall callback))
          (error (funcall error-callback err)))))))
 
-(cl-defun lsp-download-path (&key store-path set-executable? &allow-other-keys)
+(cl-defun lsp-download-path (&key store-path binary-path set-executable? &allow-other-keys)
   "Download URL and store it into STORE-PATH.
```
kiennq, line 7506: Add docs for `BINARY-PATH` too?

clients/lsp-clangd.el
danielmartin: On the VS Code side they parse the latest version from https://api.github.com/repos/clangd/clangd/releases/latest, but I don't know if it's worth the complexity; we could simply add a variable for the version.

kiennq (approved):

lsp-mode.el
yyoncho: sure.

jcs090218 (approved):

clients/lsp-clangd.el
yyoncho: > On the VS Code side they parse the latest version from https://api.github.com/repos/clangd/clangd/releases/latest, but I don't know if it's worth the complexity; we could simply add a variable for the version.

Yes. Given the 6months release cycle, it does not pay the effort.

ericdallo (approved):
Nice!

PR #2953, opened 2021-06-23
Introducing lsp-ensure-server and use-package integration

yyoncho:
Sample usage:

``` elisp
(use-package lsp-mode
  :ensure-servers (ts-ls json-ls xmlls))
```

We have to update the documentation to use it

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7277,6 +7277,10 @@ are still shown."
                 'mouse-face 'highlight)))
 
 (defun lsp--install-server-internal (client &optional update?)
+  (unless (lsp--client-download-server-fn client)
+    (user-error "There is no automatic installation for `%s', you have to install it manually following lsp-mode's documentation."
+                (lsp--client-server-id client)))
+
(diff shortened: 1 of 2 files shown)


yyoncho:
@jwiegley, I am not 100% sure that a package can extend use-package at the time use-package is installing that same package. Can you advise?

lsp-mode.el
```diff
@@ -7336,6 +7341,16 @@ When prefix UPDATE? is t force installation even if the server is present."
         t))
    update?))
 
+;;;###autoload
+(defun lsp-ensure-server (server-id)
```
kiennq, line 7345: Is this needed? Can this be a wrapper of `with-demoted-errors` around `(lsp-install-sever nil server-id)`?

yyoncho: The idea is to put that in your config and hitting remote endpoints each time you start emacs is not acceptable.

kiennq (approved):

yyoncho:
I am merging it as it is. I will check if use-package thing works on melpa. If it doesn't work I will revert it.

PR #2955, opened 2021-06-24
Add init option for rust-analyzer to set rustcSource

atsmtat:
rust-analyzer provides config option `rust-analyzer.rustcSource` to
set path to the Cargo.toml of the rust compiler workspace, which is
useful while developing on clippy/miri which uses rustc APIs. Add a
new lsp-rust option `lsp-rust-analyzer-rustc-source` to set this
config value.

Manually verified that rust-analyzer ops work on rustc repo by
setting the new option to `discover`.


yyoncho (approved):

yyoncho:
Thank you!

PR #2956, opened 2021-06-24
Add v vls support

remimimimi:

PR #2960, opened 2021-06-25
Fix issue that caused vls server not to run

ylluminarious:
When I tried to run `M-x lsp` in a [`v-mode`](https://github.com/damon-kwok/v-mode) buffer, I would consistently get an error as follows:

```
LSP :: There are no language servers supporting current mode `v-mode' registered with `lsp-mode'.
2. The language server that you expect to run is not configured to run for major mode `v-mode'. You may check that by checking the `:major-modes' that are passed to `lsp-register-client'.
3. `lsp-mode' doesn't have any integration for the language behind `v-mode'. Refer to https://emacs-lsp.github.io/lsp-mode/page/languages and https://langserver.org/ .
4. You are over `tramp'. In this case follow https://emacs-lsp.github.io/lsp-mode/page/remote/.
5. You have disabled the `lsp-mode' clients for that file. (Check `lsp-enabled-clients' and `lsp-disabled-clients').
You can run the command ‘lsp’ with s-l w s
```

This was fixed by setting `major-modes` instead of `activation-fn`, as given in https://github.com/emacs-lsp/lsp-mode/pull/2956

### DIFF
--- a/clients/lsp-v.el
+++ b/clients/lsp-v.el
@@ -41,8 +41,8 @@ finding the executable with variable `exec-path'."
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-stdio-connection (lambda () lsp-v-vls-executable))
-  :activation-fn (lsp-activate-on "V")
-  :server-id 'v-ls))
+  :major-modes '(v-mode)
+  :server-id 'vls))
 
 (lsp-consistency-check lsp-v)
 


yyoncho:
major-modes is deprecated in favour of using :activation-fn also vls id is already taken.

yyoncho:
most likely you have tested without https://github.com/emacs-lsp/lsp-mode/pull/2957

ylluminarious:
@yyoncho

> also vls id is already taken

Oh, gotcha. I'll change `vls` back to `v-ls`.

> most likely you have tested without #2957

No, I did try this with #2957 and it still didn't work. Only changing it to use `major-modes` worked. If that's deprecated, do you have any other suggestions on what can be done to fix it instead?

yyoncho:
Check why   `(funcall (lsp-activate-on "V") nil nil)` does not return `t` in your buffer.

ylluminarious:
@yyoncho I've investigated it, but I'm still not entirely sure why it doesn't return `t`. Strangely, using a lowercase `"v"` via:

```
(funcall (lsp-activate-on "v") nil nil)
```

Does return `t`, but using an uppercase `"V"` returns `nil`. Not only so, but even when I change `:activation-fn` to use `(lsp-activate-on "v")` (lowercase `"v"`), the server still refuses to start. Even more perplexingly, `(lsp-buffer-language)` also returns `"v"`, meaning that a lowercase `"v"` in the `:activation-fn` should definitely be working. And running `(-contains? '("v") (lsp-buffer-language))` also returns `t` (which is a sanity check I performed to make sure there wasn't somehow something wrong with `lsp-activate-on`).

So you would think it really *should* be working—and that #2957 should actually be breaking it instead of fixing it. But it happens to be broken either way; with `"V"` or `"v"`.

I have no custom settings for `lsp-mode` in my personal config whatsoever, so this seems to be an issue which is not dependent on my own setup.

yyoncho:
It should be "v" indeed and it should work. I tested in in v-mode buffer and `(lsp--filter-clients #'lsp--matching-clients?)` returns non-nil result in v-mode buffers.

ylluminarious:
@yyoncho 

> and it should work after that change

Just for clarity's sake, which change are you referring to? Do you mean that it should work after #2957 or after changing `"V"` to `"v"`?

> I tested in in v-mode buffer and `(lsp--filter-clients #'lsp--matching-clients?)` returns non-nil result in v-mode buffers.

That same expression returns `nil` for me.

Have you tried actually running `M-x lsp` in a `v-mode` buffer with the `vls` server installed? Are you able to reproduce the error I pasted above?

Also, could this issue possibly have to do with my Emacs version? I'm using Emacs 26.3.

yyoncho:
> @yyoncho
> 
> > and it should work after that change
> 
> Just for clarity's sake, which change are you referring to? Do you mean that it should work after #2957 or after changing `"V"` to `"v"`?

It should be changed back to "v".

> 
> > I tested in in v-mode buffer and `(lsp--filter-clients #'lsp--matching-clients?)` returns non-nil result in v-mode buffers.
> 
> That same expression returns `nil` for me.
> 
> Have you tried actually running `M-x lsp` in a `v-mode` buffer with the `vls` server installed? Are you able to reproduce the error I pasted above?

No, the other part is trivial, so it should work.

 
> Also, could this issue possibly have to do with my Emacs version? I'm using Emacs 26.3.

No, just (lsp-buffer-language-id) should match the one in lsp-activate-on

ylluminarious:
@remimimimi Could you offer any advice / insight here?

remimimimi:
@ylluminarious It doesn't work for me either. I tried to get it work yesterday and it somehow worked for me when I changed `v` to` V` but after that it loaded endlessly(after emacs restart this behavior changed to error same as yours), so I don't know how to start it normally either. I'll try to figure it out today

remimimimi:
Work's with this in lsp-v.el @ylluminarious 
```elisp
(lsp-register-client
 (make-lsp-client
  :new-connection (lsp-stdio-connection (lambda () lsp-v-vls-executable))
  :major-modes '(v-mode)
  :server-id 'v-ls))
```
But again infinite loading, so maybe this is `vls` issue?

shadowninja55:
> Work's with this in lsp-v.el when I manually evaluate elisp in lsp-v.el @ylluminarious
> 
> ```emacs-lisp
> (lsp-register-client
>  (make-lsp-client
>   :new-connection (lsp-stdio-connection (lambda () lsp-v-vls-executable))
>   :major-modes '(v-mode)
>   :server-id 'v-ls))
> ```
> 
> But again infinite loading, so maybe this is `vls` issue?

infinite loading and then crash does happen sometimes. you'll probably have the best success testing with the master vls branch (not use-tree-sitter) on a small file.

PR #2965, opened 2021-06-28
Fix `lsp-headerline-breadcrumb-segments` doc

schellj:
`lsp-headerline-breadcrumb-segments` is a list of segments, not a face. Used on headerline, not modeline.


ericdallo (approved):
I was probably needing to sleep when I wrote that :p

PR #2972, opened 2021-06-30
Handle 'renamed' file notify event

### LINKED ISSUE #2973: LSP mode ignores 'renamed' file watcher event

- [ ] You may also try reproduce the issue using clean environment using the following command
```bash
```

### Bug description

lsp-mode only handles 'created', 'deleted' and 'changed' file watch events. At least on Windows 10, file watchers as created by file-notify-add-watch raise a 'renamed' event when files are renamed (e.g. through dired). Since this event is currently ignored by lsp-mode, the lsp server is not notified that the file was renamed, and this typically results in the server believing that both the original file and new file exist concurrently, which in itself lead to spurious error reports, double or ambiguous declarations, etc, depending on the language itself.

See https://github.com/emacs-lsp/lsp-mode/blob/fc30baf3d66b9b5c695e851f7e54af926a453c12/lsp-mode.el#L1681 and the documentation from lisp function 'file-notify-add-watch' for references.

### Steps to reproduce

1. Open a file in a lsp workspace (ex: A.cs),
2. C-x C-j (dired-jump),
3. R (dired-rename) and choose a new filename (ex: B.cs),
4. The lsp-server (omnisharp-roslyn in my case) now thinks that both A.cs and B.cs exists.

### Expected behavior

The lsp-server should be notified that A.cs does not exist anymore, and that B.cs is a new file to be managed by the workspace. In particular, the lsp server should not be left thinking that both A.cs and B.cs currently exist.

omnisharp-roslyn

### OS

Windows

### Error callstack

_No response_

### Anything else?

_No response_

thomashilke:
lsp-mode only handles 'created', 'deleted' and 'changed' file watch events. At least on Windows 10, file watchers as created by file-notify-add-watch raise a 'renamed' event when files are renamed (e.g. through dired). Since this event is currently ignored by lsp-mode, the lsp server is not notified that the file was renamed, and this typically results in the server believing that both the original file and new file exist concurrently, which in itself lead to spurious error reports, double or ambiguous declarations, etc, depending on the language itself. 

This commit handles properly the 'rename' event by by notifying the lsp server of the deletion of the file and creation of the file with the new name.

Fixes #2973


lsp-mode.el
```diff
@@ -1679,7 +1679,12 @@ This set of allowed chars is enough for hexifying local file paths.")
      ((and (not (file-directory-p file-name))
            (not (lsp--string-match-any ignored-files file-name))
            (memq event-type '(created deleted changed)))
-      (funcall callback event)))))
+      (funcall callback event))
+     ((and (not (file-directory-p file-name))
+           (not (lsp--string-match-any ignored-files file-name))
+           (memq event-type '(renamed)))
```
yyoncho, line 1685: can you put both memq checks on the first place in the `and`-s?

yyoncho (commented):
Looks good. One minor nit:

lsp-mode.el
thomashilke: Yes sure. Like so?

yyoncho (approved):

jcs090218 (approved):

thomashilke:
Some checks fail with the error `Failed to download ‘melpa’ archive.` Is there something that I should do?

yyoncho:
> Some checks fail with the error `Failed to download ‘melpa’ archive.` Is there something that I should do?

No.

PR #2975, opened 2021-06-30
lsp-capf: add pass-through completion style + support prefix filtering on completion table

### LINKED ISSUE #2970: The completion table does not perform prefix filtering

- [x] You may also try reproduce the issue using clean environment 

### Bug description

The lsp completion table is non conforming to the specification. In particular for the all-completions action (t) the returned candidates must be filtered with the given prefix string.

The problem is here:

https://github.com/emacs-lsp/lsp-mode/blob/fc30baf3d66b9b5c695e851f7e54af926a453c12/lsp-completion.el#L504

Original issue:

https://github.com/minad/consult/issues/346

### Steps to reproduce

See https://github.com/minad/consult/issues/346

### Expected behavior

The completion table should be comforming to the specification.

Any

### OS

Linux

### Anything else?

The issue here is that lsp wants to avoid double filtering. My proposal is to not use basic but a custom completion style which does not make use of prefix filtering for the lsp-capf category. Given such a completion style allows to use a fully compliant completion table. In most scenarios the end result will be equivalent. However not in all scenarios, e.g., not in the case when consult-completion-in-region is used, where the completion is transferred to the minibuffer. I can help out with a patch. 

Cc @kiennq

kiennq:
Fix #2970.


lsp-completion.el
```diff
@@ -715,6 +706,12 @@ The CLEANUP-FN will be called to cleanup."
   "Disable LSP completion support."
   (lsp-completion-mode -1))
 
+(defun lsp-completion-pass-through-try-completion (string table pred _point)
+  (try-completion string table pred))
+
+(defun lsp-completion-pass-through-all-completions (string table pred _point)
+  (all-completions string table pred))
```
minad, line 713: This is not correct here. It should be `(all-completions "" table pred)` in order to disable the prefix filtering.

lsp-completion.el
```diff
@@ -715,6 +706,12 @@ The CLEANUP-FN will be called to cleanup."
   "Disable LSP completion support."
   (lsp-completion-mode -1))
 
+(defun lsp-completion-pass-through-try-completion (string table pred _point)
```
minad, line 709: Why not call it `lsp-completion-passthrough-try-completion`? "passthrough" is a word - why add the hypen?

lsp-completion.el
```diff
@@ -731,8 +728,13 @@ The CLEANUP-FN will be called to cleanup."
      (lsp-completion-mode
       (setq-local completion-at-point-functions nil)
       (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
-      (setq-local completion-category-defaults
-                  (add-to-list 'completion-category-defaults '(lsp-capf (styles basic))))
+      (make-local-variable 'completion-category-defaults)
+      (setf (alist-get 'lsp-capf completion-category-defaults) '((styles . (pass-through))))
+      (make-local-variable 'completion-styles-alist)
+      (setf (alist-get 'pass-through completion-styles-alist)
```
minad, line 734: I would call the style `lsp-passthrough`, not `pass-through` in order to make the provenance of the completion style clear and in order to avoid name clashes.

lsp-completion.el
```diff
@@ -758,6 +760,8 @@ The CLEANUP-FN will be called to cleanup."
       (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
       (setq-local completion-category-defaults
                   (cl-remove 'lsp-capf completion-category-defaults :key #'cl-first))
+      (setq-local completion-styles-alist
+                  (cl-remove 'pass-through completion-styles-alist :key #'cl-first))
```
minad, line 764: lsp-passthrough here too.

lsp-completion.el
minad: Furthermore - using `all-completions` does not apply highlighting. You have to add that separately based on the argument `string`.

lsp-completion.el
```diff
@@ -715,6 +706,12 @@ The CLEANUP-FN will be called to cleanup."
   "Disable LSP completion support."
   (lsp-completion-mode -1))
 
+(defun lsp-completion-pass-through-try-completion (string table pred _point)
+  (try-completion string table pred))
```
minad, line 710: Using `try-completion` is not correct here. The function must follow the calling convention of `completion-try-completion` which (confusingly) is different from `try-completion`.

yyoncho:
@kiennq CI is red. I will do manual testing once @minad is done with his changes.

kiennq:
> @kiennq CI is red. I will do manual testing once @minad is done with his review.

Fixed PR. @minad can you help taking a look to see if this works for you?

lsp-completion.el
```diff
@@ -501,7 +502,17 @@ The MARKERS and PREFIX value will be attached to each candidate."
           ;; always be shown
           ((equal action 'lambda) nil)
           ;; retrieve candidates
-          ((equal action t) (funcall candidates))))
+          ((equal action t)
+           (if (string-empty-p probe)
+               ;; when triggered via `lsp-passthrough'
+               (funcall candidates)
```
minad, line 508: Checking for an empty probe is an unnecessary optimization here. It may lead to problems later on since the returned list is mutated (assuming the list is cached).

minad (changes requested):

lsp-completion.el
```diff
@@ -501,7 +502,17 @@ The MARKERS and PREFIX value will be attached to each candidate."
...
+             ;; otherwise, doing double filtering
+             (let* ((re (completion-all-completions probe (funcall candidates)
```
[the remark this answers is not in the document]
kiennq: Ah, I got it now. Thanks

lsp-completion.el
```diff
@@ -715,6 +726,10 @@ The CLEANUP-FN will be called to cleanup."
   "Disable LSP completion support."
   (lsp-completion-mode -1))
 
+(defun lsp-completion-passthrough-all-completions (_string table pred _point)
+  "Like `completion-basic-all-completions' but have prefix ignored."
+  (completion-basic-all-completions "" table pred 0))
```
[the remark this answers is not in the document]
kiennq: Yes, please do apply the highlighting manually from the front end package side.
We only provide the highlighting cue via `:company-match` extension and let the frontend package do the highlighting itself when needed.
I think that's more robust and much less resource-consuming than have to provide highlighting for all the items.
Some server returns a few thousand items, doing the highlighting for all of them guaranteed in a degrade perf.

lsp-completion.el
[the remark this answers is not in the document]
kiennq: I think this part is puzzling me the most here. In most cases when `completion-styles` doesn't contain style like `basic` and `emacs...`, the prefix will be ignored anyway and you will get the result filtered later.

lsp-completion.el
kiennq: We don't cache that list, rather we cache the list that's the evaluation of `(funcall candidates)`.
But yes, this optimization might not need.

lsp-completion.el
```diff
@@ -501,7 +502,8 @@ The MARKERS and PREFIX value will be attached to each candidate."
           ;; always be shown
           ((equal action 'lambda) nil)
           ;; retrieve candidates
-          ((equal action t) (funcall candidates))))
+          ((equal action t)
```
minad, line 505: Why not use complete-with-action? This will be shorter and equivalent to what you have now. I think you had that in the version of this PR after my first review.

lsp-completion.el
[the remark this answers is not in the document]
minad: Could you clarify where the highlighting data comes from? Is the highlighting data in  the company-match function provided by the server?

But I agree with the performance issue. In my frontend Vertico and Corfu I am implementing deferred highlighting for the completion styles. I have also proposed upstream to extend completion styles with an ability to defer highlighting which hopefully would obsolete :company-match.

lsp-completion.el
minad: Furthermore you did not update the null action (try-completion). I would simply go with complete-with-action and be done. Is there a point in the current implementation?

minad:
~~~ elisp
(lambda (str pred action) ;; compliant completion table
  (if (eq action 'metadata)
      '(metadata (category . lsp-capf)
                 (cycle-sort-function . identity)
                 (display-sort-function . identity))
    (complete-with-action action (funcall candidates) str pred)))
~~~

lsp-completion.el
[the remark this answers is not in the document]
minad: Yes, only a few completion styles make use of the prefix filtering. It is an optimization basically.

But note that even complex completion styles like Orderless could in principle make use of it. The first word could be passed as prefix and the subsequent words are used for filtering outside the table.

lsp-completion.el
kiennq: There are, you can see that in the comment for each case there.

kiennq: I would rather not risk changing all of them to use the `complete-with-action`, if you think the implementation of some certain part is not working with your case, please state that

lsp-completion.el
minad: If you cache the list returned by funcall candidates the optimization is even incorrect, since the frontend may mutate the returned list. But maybe I misunderstood you.

lsp-completion.el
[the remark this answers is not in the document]
kiennq: The `:company-match` provides a function that you can call on the candidates to get the data where's the highlight should be.

lsp-completion.el
minad: Okay fair enough. (My reason would be: I don't trust the implementation. So I have to review it point by point. In comparison complete-with-action is already provided and does the right thing. Furthermore why not go with the simpler code, why make things complicated?)

lsp-completion.el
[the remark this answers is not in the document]
minad: Yes of course. But this is bad. Company unnecessarily forces its extensions onto every frontend. Not good.

minad:
Okay, I took a look now at the whole function. I add some annotations.

~~~ elisp
       (lambda (probe pred action)
         (cond
          ;; OK: metadata
          ((equal action 'metadata)
           `(metadata (category . lsp-capf)
                      (display-sort-function . identity)
                      (cycle-sort-function . identity)))
          ;; OK: boundaries
          ((equal (car-safe action) 'boundaries) nil)
          ;; WRONG: try-completion
          ;; try-completion should return the common prefix of all the candidates.
          ;; Take a look at the emacs21 completion style which makes very direct use of this.
          ;; The basic completion style makes use of this too, but it is a bit more involved.
          ((null action)
           (when-let ((cands (funcall candidates)))
             (if (cl-rest cands) probe (cl-first cands))))
          ;; WRONG: test-completion: not return exact match so that the selection will always be shown
          ;; Your table is specifying that the completion is never a valid completion.
          ;; The argument given here, that the selection should always be shown is dubious.
          ;; The completion table (backend!) should not take such decisions for the frontend.
          ((equal action 'lambda) nil)
          ;; OK: retrieve candidates
          ((equal action t)
           (all-completions probe (funcall candidates) pred))))
~~~

After having taken a look I stand by my point that `complete-with-action` would be the better solution here. 

Please take a look at the eglot implementation, which implements every single action correctly. It handles each case separately (instead of using `complete-with-action`) since it does not want to use `all-completions` in order to rely on the `filterText`. This is okay, but probably unnecessary given the fact that most completion styles will pass the empty string to the backend, effectively disabling the filtering (compare your lsp-passthrough style which also deliberately passes the empty string).

~~~ elisp
       (lambda (probe pred action)
         (cond
          ((eq action 'metadata) metadata)               ; metadata
          ((eq action 'lambda)                           ; test-completion
           (test-completion probe (funcall proxies)))
          ((eq (car-safe action) 'boundaries) nil)       ; boundaries
          ((null action)                                 ; try-completion
           (try-completion probe (funcall proxies)))
          ((eq action t)                                 ; all-completions
           (cl-remove-if-not
            (lambda (proxy)
              (let* ((item (get-text-property 0 'eglot--lsp-item proxy))
                     (filterText (plist-get item :filterText)))
                (and (or (null pred) (funcall pred proxy))
                     (string-prefix-p
                      probe (or filterText proxy) completion-ignore-case))))
            (funcall proxies)))))
~~~

kiennq:
> ```emacs-lisp
> ;; WRONG: try-completion
>           ;; try-completion should return the common prefix of all the candidates.
>           ;; Take a look at the emacs21 completion style which makes very direct use of this.
>           ;; The basic completion style makes use of this too, but it is a bit more involved.
> ```

No, the `lsp-mode` capf is not designed to work with any other completion styles other than the passthrough.
Unless something is broken for other packages because of this, I would just keep this.


> ```emacs-lisp
>           ;; WRONG: test-completion: not return exact match so that the selection will always be shown
>           ;; Your table is specifying that the completion is never a valid completion.
>           ;; The argument given here, that the selection should always be shown is dubious.
>           ;; The completion table (backend!) should not take such decisions for the frontend.
> ```

This is intentional, in the past there's case when the `test-completion` returns an exact match, thus make the post-completion function not invoked. The problem may have been resolved (I haven't test it) but some older Emacs version may still suffer from it. We need to run the post-completion function to cleanup/expand snippets and do various actions specified by the language server.
I would rather make the completion is never a valid completion and force the user to choose from selection (even with only one item) than making completion stop working after doing a valid completion.

kiennq:
@minad Can you check if this change is working for you (except the highlighting part)?

minad:
@kiennq 

> Can you check if this change is working for you (except the highlighting part)?

I didn't find time to test this yet. But by inspection and as I already mentioned, the completion table is still not correct.

> ...but some older Emacs version may still suffer from it...

My recommendation is to drop support for very old Emacs which suffer from these issues. Which versions are affected?

PR #2976, opened 2021-06-30
Better visuals for signature help

yyoncho:
- use page-break-lines-mode to separate the sections

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5122,13 +5122,24 @@ RENDER-ALL - nil if only the signature should be rendered."
   (funcall lsp-signature-function nil)
   (lsp-signature-mode -1))
 
+(declare-function page-break-lines--update-display-tables "ext:page-break-lines")
+
+(defun lsp--setup-page-break-mode-if-present ()
+  "Enable `page-break-lines-mode' in current buffer."
+  (when (fboundp 'page-break-lines-mode)
+    (page-break-lines-mode)
+    ;; force page-break-lines-mode to update the display tables.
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -57,6 +57,7 @@
 (require 'minibuffer)
 (require 'yasnippet nil t)
 (require 'lsp-protocol)
+(require 'page-break-lines)
```
kiennq, line 60: Is this built-in package? Else you have to put that one into package requires.

kiennq:
Can you add some screenshot for this?
Do we need to use another package for this or can just walk away with insert bunches of `-` in place of `^L`?

lsp-mode.el
ericdallo: true, also on makefile for CI tests I think

yyoncho: I thought that it is built-in, it is in melpa and built-in for spacemacs.  I will look for a solution

yyoncho:
> Can you add some screenshot for this?
> Do we need to use another package for this or can just walk away with insert bunches of `-` in place of `^L`?

Screenshot: 
image


I don't want to go in the direction of inserting a bunch of `-` - in the end in order to look good we will have to reimplement the 
page-break-mode package.

ericdallo:
Looks way better with this package indeed, maybe try to reimplements only the necessary?

jcs090218:
Looks awesome! Can't wait to see this to merge! :)

ericdallo (approved):

jcs090218 (approved):

PR #2977, opened 2021-07-01
feat: use `name` instead of `description` for bettter readability

Aaronzinhoo:
This is more of a ui-enhancement when working with lsp-yaml

Previously:

image

Updated:
image

Maybe this is more in line with what people see on the schemastore and might be easier to search but I will leave that up to yall.

### DIFF
--- a/clients/lsp-yaml.el
+++ b/clients/lsp-yaml.el
@@ -225,7 +225,7 @@ Set FORCE-DOWNLOADING to non-nil to force re-download the database."
   (let* ((schema (lsp--completing-read "Select buffer schema: "
                                        (lsp-yaml--get-supported-schemas)
                                        (lambda (schema)
-                                         (alist-get 'name schema))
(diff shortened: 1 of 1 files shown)


yyoncho:
I will let @kiennq decide on that. 

As a side note, you could concat name and description.

kiennq:
Yeah, I'm in the favor of concatenating name + description, perhaps in the format of `"%s (%s)"`.
The name alone is a little bit cryptic sometimes

Aaronzinhoo:
Got it, will make the change!

Aaronzinhoo:
@kiennq @yyoncho updated visual

image

Lemme know if there are any other changes possible!

kiennq (approved):

kiennq:
Thank you

Aaronzinhoo:
@kiennq glad I could help in some way!

PR #2983, opened 2021-07-03
Format lsp-rust variable docstrings

Wilfred:
Replace \n with newlines and use quoting syntax that Emacs recognises.


yyoncho (approved):

yyoncho:
Thank you!

(Most likely caused by issue in https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-generate-settings.el )

PR #2988, opened 2021-07-06
Make enum behave as type in semantic highlighting

amnore:
According to [lsp spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens), 'enum' means an enum type, not an enum constant. It should have the same face as type.

Below are two screenshots showing the differences in highlighting.

Before:
cpp-before

After:
cpp-after


yyoncho (approved):
CC @sebastiansturm

ericdallo (approved):

sebastiansturm (approved):

PR #2992, opened 2021-07-09
Use hack-indent-offset for on-type-formatting in hack mode

dneiter:
Previously for lsp--get-indent-width was falling back to "default" case for hack-mode, resulting in inconsistent formatting (hack mode defaults to 2 spaces, default is 4).

Add hack-mode case using correct hack-indent-offset.


yyoncho:
Thank you!

PR #3000, opened 2021-07-14
Consider activation-fn from client during library-fn check

ericdallo:
I realized lsp-dart `:library-fn` was not being called because lsp-mode only check for `:major-modes` instead of checking `:activation-fn` and `:major-modes`

Related to https://github.com/emacs-lsp/lsp-dart/issues/106

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8077,10 +8077,7 @@ The library folders are defined by each client for each of the active workspace.
                              (-sort (lambda (a _b)
                                       (-contains? lsp--last-active-workspaces a)))
                              (--first
-                              (and (or (-contains? (-> it lsp--workspace-client lsp--client-major-modes)
-                                                   major-mode)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -8070,16 +8070,17 @@ Returns nil if the project should not be added to the current SESSION."
...
   (when-let ((workspace (->> (lsp-session)
                              (lsp--session-workspaces)
                              ;; Sort the last active workspaces first as they are more likely to be
                              ;; the correct ones, especially when jumping to a definition.
                              (-sort (lambda (a _b)
                                       (-contains? lsp--last-active-workspaces a)))
                              (--first
-                              (and (-contains? (-> it lsp--workspace-client lsp--client-major-modes)
-                                               major-mode)
+                              (and (or (-contains? (-> it lsp--workspace-client lsp--client-major-modes)
```
yyoncho, line 8080: lsp--matching-clients?

yyoncho (approved):

Issue #2920, opened 2021-06-09
File mode specification error: (wrong-type-argument lsp--workspace nil) 


jaaved:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).
Alternatively, it will be great if you can reproduce the issue using [lsp-docker](https://github.com/emacs-lsp/lsp-docker/) which provides the minimal configurations for `lsp-mode` and ships with most of the language servers.

### Bug description

File mode specification error: (wrong-type-argument lsp--workspace nil)
Error during redisplay: (eval (mapconcat #'lsp--workspace-print lsp--buffer-workspaces "][")) signaled (wrong-type-argument lsp--workspace nil) [2 times]
Error processing message (wrong-type-argument lsp--workspace nil).

### Steps to reproduce

opening any python file is resulting in this error, lsp feature do not work.

### Expected behavior

expected to use lsp mode with out errors.

### Which Language Server did you use?

pyright

### OS

Linux

### Anything else?

using emacs 28

yyoncho:
can you test with lsp-start-plain.el?

jaaved:
@yyoncho 
```
emacs -q -l lsp-start-plain.el
```
Error:
```
Debugger entered--Lisp error: (error "Variable binding depth exceeds max-specpdl-size")
  (cdr list)
  (setq list (cdr list))
  (prog1 list (setq list (cdr list)))
  (car-safe (prog1 list (setq list (cdr list))))
  (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i)
  (and list (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i) (or (not (string-match it str)) (ignore (setq needle it))))
  (while (and list (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i) (or (not (string-match it str)) (ignore (setq needle it)))) (setq it elt it-index i i (1+ i)))
[...]

PR #3001, opened 2021-07-14
Minor semantic tokens cleanup

sebastiansturm:
this should fix bug #2920 (not the original issue, the one discussed further down in the issue thread), and removes one unneeded call to `lsp--semantic-tokens-initialize-buffer`.

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -61,6 +61,15 @@ preferred over both full and ranged token requests."
   :group 'lsp-semantic-tokens
   :type 'boolean)
 
+(defcustom lsp-semantic-tokens-honor-refresh-requests nil
+  "Whether to honor semanticTokens/refresh requests.
+
+When set to nil, refresh requests will be silently discarded.
+When set to t, semantic tokens will be re-requested for all buffers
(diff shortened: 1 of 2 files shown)


lsp-semantic-tokens.el
```diff
@@ -61,6 +61,15 @@ preferred over both full and ranged token requests."
   :group 'lsp-semantic-tokens
   :type 'boolean)
 
+(defcustom lsp-semantic-tokens-honor-refresh-requests nil
```
ericdallo, line 64: This frequency should be different form language server to language server so maybe it's a good idea to have it disabled or even in the future we allow/enable for servers we know do that "right"

ericdallo (approved):

sebastiansturm:
strange that I'm getting CI failures on macos again; last time, simply rerunning the tests made those checks pass. Is there anything special about the macos CI environment?

ericdallo:
@sebastiansturm AFAIK no, we noticed infra issues on the CI only, like `cask` not being able to find a dep on melpa or something like this, but nothing related with a specific job

ericdallo:
I'd suggest re-run the jobs

PR #3004, opened 2021-07-15
lsp-completion: fix completion context by using non-essential value

### LINKED ISSUE #2998: c++-mode ':', '>' and '/' unnecessarily begin completion

- [X] You may also try reproduce the issue using clean environment using the following command
```bash
```

### Bug description

When editing c++ with `lsp-mode` enabled typing `:`, `>` or `/` pops up the completion box in situations where there couldn't be a relevant completion.

### Steps to reproduce

```C++
int main() {
    // on the line below try typing ':', '>' or '/'
    
    return 0;
}
```

### Expected behavior

Don't show the completion box unnecessarily.

lsp-clangd.

### OS

Linux

### Error callstack

_No response_

### Anything else?

Whilst I've no clue how `lsp-mode` implements things. It seems that a possible acceptable fix could be to check what precedes `:`, `>` or `/`. For `:` it should be preceded by another `:` as part of a `::` sequence to trigger the completion box. Similarly for `>` it should be preceded by `-` as part of a `->` sequence to trigger the completion box. Not sure why `/` would ever trigger the completion box.

kiennq:
Fix #2998

Currently, we decide if completion is triggered by character by using `last-command`, which is not correct since other elisp's prog can intercept it.
To better detect completion's context, we can assume that all auto-triggered completion should have `non-essential` set as `t` and only try to determine if the insert character is triggered character in that condition.


kiennq:
@ville-h can you see if this fixes your issue?

yyoncho (commented):
I will approve it, but TBO I do not like this fix. I wish somehow we are explicitly notified what is the trigger.

yyoncho (approved):

dgutov:
> I wish somehow we are explicitly notified what is the trigger.

It should be possible to add some finer kind of indication to the completion-at-point API, but that's a discussion for emacs-devel, I guess. AFAIK `non-essential` is the sole bit of frontend agnostic information that's available now.

kiennq:
> I will approve it, but TBO I do not like this fix. I wish somehow we are explicitly notified what is the trigger.

Yeah, I wish we have a better indicator for that too. For now, let's go with `non-essential` and update the code to use new indicator once it's out

ville-h:
@kiennq yes seems to have solved it on my end. haven't had the chance to give it a proper go if anything new has propped as a result for example.

PR #3007, opened 2021-07-17
Add missing semantic tokens from lsp spec

amnore:
There are two token types in [lsp spec](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_semanticTokens) that are not present in lsp-mode. According to [vscode doc](https://code.visualstudio.com/api/language-extensions/semantic-highlight-guide#semantic-token-classification), they are:

- enumMember: For identifiers that declare an enumeration property, constant, or member.
- event: For identifiers that declare an event property.

This commit adds these two tokens.


ericdallo (approved):
Good catch! thanks for the contribution

PR #3008, opened 2021-07-18
Miscellaneous ESLint client changes and additions

Sorixelle:
A few changes to the ESLint config to support more configurations options, and tidy up customs.

- Support running the ESLint client on Svelte files
- Use `lsp-eslint-validate` to force enabling validation for files with given language ids. The variable was previously unused, and the functionality added was used for Svelte support.
- Give `lsp-eslint-code-action-disable-rule-comment` and `lsp-eslint-code-action-show-documentation` a nicer configuration interface:
  - `lsp-eslint-code-action-disable-rule-comment` is now only a boolean value for toggling it's associated code action
  - `lsp-eslint-code-action-disable-rule-comment-location` is used to specify the location of the comment
  - `lsp-eslint-code-action-show-documentation` is now only a boolean value for toggling it's associated code action, instead of an unnecessary alist
- Update the ESLint server downloaded (2.1.14 -> 2.1.23)
- Add 2 new configuration options:
  - `lsp-eslint-warn-on-ignored-files`: Controls if a warning should be emitted when a file is ignored.
  - `lsp-eslint-rules-customizations`: Defines severity overrides for ESLint rules.

A note: In server version 2.1.22, the `eslint/confirmESLintExecution` command was removed, meaning a large portion of [this commit](https://github.com/emacs-lsp/lsp-mode/commit/a0c41e489c656bc69a3cdd0d04ee549f45e42a59) could be reverted. Do we still want to keep it in order to support older versions?

### DIFF
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -118,10 +118,10 @@ source.fixAll code action."
   :type 'lsp-string-vector
   :package-version '(lsp-mode . "6.3"))
 
-(defcustom lsp-eslint-validate ["javascript" "javascriptreact"]
-  "An array of language ids which should be validated by ESLint"
-  :type 'lsp-string-vector
-  :package-version '(lsp-mode . "6.3"))
+(defcustom lsp-eslint-validate '()
+  "An array of language ids which should always be validated by eslint."
+  :type '(repeat string)
+  :package-version '(lsp-mode . "7.1"))
 
 (defcustom lsp-eslint-provide-lint-task nil
   "Controls whether a task for linting the whole workspace will be available."
@@ -254,7 +254,7 @@ stored."
                                (buffer (find-buffer-visiting file))
                                (workspace-folder (lsp-find-session-folder (lsp-session) file)))
                     (with-current-buffer buffer
-                      (list :validate "probe"
+                      (list :validate (if (member (lsp-buffer-language) lsp-eslint-validate) "on" "probe")
                             :packageManager lsp-eslint-package-manager
                             :codeAction (list
                                          :disableRuleComment (list
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -118,7 +118,7 @@ source.fixAll code action."
   :type 'lsp-string-vector
   :package-version '(lsp-mode . "6.3"))
 
-(defcustom lsp-eslint-validate '()
+(defcustom lsp-eslint-validate '("svelte")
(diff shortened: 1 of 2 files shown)


clients/lsp-eslint.el
```diff
@@ -346,7 +389,7 @@ to allow or deny it.")
                                    (funcall callback))
                                (error (funcall error-callback err))))
                            error-callback
-                           :url (lsp-vscode-extension-url "dbaeumer" "vscode-eslint" "2.1.14")
+                           :url (lsp-vscode-extension-url "dbaeumer" "vscode-eslint" "2.1.23")
```
yyoncho, line 392: I wish we do that change in a way to force downloading of the new version in case it is already downloaded the previous one.

clients/lsp-eslint.el
```diff
@@ -308,9 +351,9 @@ to allow or deny it.")
   :activation-fn (lambda (filename &optional _)
```
yyoncho, line 351: not part of this review, but at some point we might use `lsp-activate-on` here, which will simplify the code a lot.

yyoncho (commented):
Looks good:

> A note: In server version 2.1.22, the eslint/confirmESLintExecution command was removed, meaning a large portion of this commit could be reverted. Do we still want to keep it in order to support older versions?

As I mentioned in the review, I wish we force downloading of the new version somehow. We may add a warning in the notification handler - "Please upgrade to ..."

clients/lsp-eslint.el
Sorixelle: I'm not 100% sure what the best way to do this is - to my knowledge, the version isn't stored anywhere so it's tricky to know if the server needs to be updated. Any suggestions?

yyoncho: I think that slightly changing the download location should work. We might include the server version in the new store path path. `lsp-eslint-unzipped-path` controls that.

Sorixelle: Added the version number to `lsp-eslint-unzipped-path`, that should do it for most cases. If the user is setting it to a custom value though, the update won't be forced. There's very little reason for users to be changing this anyway though - do we think this is much of a problem?

yyoncho: I think this PR is the best we can do. Can you also drop a note in the changelog? After that, it is good to go.

yyoncho (approved):

yyoncho:
Thank you!

PR #3009, opened 2021-07-18
[lsp-rust] Add stopAtEntry for debugging lens

gagbo:
This ensures to get control at least once when running a single test

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -862,6 +862,18 @@ https://rust-analyzer.github.io/manual.html#auto-import.
   :group 'lsp-rust
   :package-version '(lsp-mode . "7.1"))
 
+(defcustom lsp-rust-analyzer-debug-lens-extra-dap-args
+  '(:MIMode "gdb" :miDebuggerPath "gdb" :stopAtEntry t :externalConsole :json-false)
+  "Extra arguments to pass to DAP template when debugging a test from code lens.
+
+As a rule of the thumb, do not add extra keys to this plist unless you exactly what
+you are doing, it might break the \"Debug test\" lens otherwise.
+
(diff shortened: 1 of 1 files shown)


yyoncho:
can we have that configurable?

yyoncho:
Even better will be to have a list of default settings to append to the template. That way you will be able to change everything.

gagbo:
I can do that for sure, but I really think that only the `stopAtEntry` is really necessary:
- all the arguments that are already present are computed from lsp-rust and I don't think it's wise to try to change that from user settings
- I think that maybe having an optional `dap-compilation(-dir)?` step added could be nice, but I found out recently with Cargo workspaces that it's not always trivial to find the correct compilation directory, so I'd rather leave it out as well.

yyoncho:
I would say that we want to have any of these configurable: `MIMode` `miDebuggerPath` `externalConsole`(and probably more), WDYT?

(JFYI you may run a test in vterm using an external console).

gagbo:
Seems all good :+1: , I ignored those MI settings because last time you told me that there was an issue with LLDB in cpptools anyway (and I just use Linux+GDB so I forgot).

yyoncho:
> Seems all good  , I ignored those MI settings because last time you told me that there was an issue with LLDB in cpptools anyway (and I just use Linux+GDB so I forgot).

I havent investigated, but I suspect that the issue is exactly that - there is no way to specify that you want to run the tests with lldb.

gagbo:
Is there a function is lsp-mode or dap-mode to safely merge plists ? I'm only using append here, but I'm not really sure what will happen if there are duplicate keys in the `defcustom` and the programmatically filled `list`. So I only added a warning in the defcustom docstring

yyoncho (approved):

Issue #3011, opened 2021-07-22
`lsp--all-watchable-directories` unable to handle symlink loops 


sata:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

Given a directory structure of:
```
── testdata
    ├── symlinks
    │   ├── dir-symlink -> ../../testdata
    │   ├── file-symlink -> ../test.file
    │   ├── invalid-symlink -> /non/existing/file
    │   └── windows-file-symlink -> C:/Users/ibrahim/go/src/github.com/golang/dep/internal/fs/testdata/test.file
    └── test.file
```
`lsp--all-watchable-directories` will fail with `recursive (error "Lisp nesting exceeds ‘max-lisp-eval-depth’")` as the function recursively will chase the `dir-symlink` without checking depth or if it's a cyclic symlink.

Backtrace:
```
Debugger entered--Lisp error: (error "Lisp nesting exceeds ‘max-lisp-eval-depth’")
  (prog1 list (setq list (cdr list)))
  (car-safe (prog1 list (setq list (cdr list))))
  (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i)
  (and list (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i) (or (not (string-match it str)) (ignore (setq needle it))))
  (while (and list (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i) (or (not (string-match it str)) (ignore (setq needle it)))) (setq it elt it-index i i (1+ i)))
  (let ((list regex-list) (i 0) elt it it-index) (ignore it it-index) (while (and list (setq elt (car-safe (prog1 list (setq list (cdr list)))) it elt it-index i) (or (not (string-match it str)) (ignore (setq needle it)))) (setq it elt it-index i i (1+ i))))
[...]

PR #3013, opened 2021-07-24
Fix inf recursion in all-watchable-directories

sata:
Fix inf recursion in all-watchable-directories
    
    By tracking which paths we have visited, we can avoid infinite
    recursions when encountering symlink loops in a repository.
    
    See https://github.com/emacs-lsp/lsp-mode/issues/3011 for more details.


yyoncho:
This issue is a regression.

I think that we already keep the watched directories in `lsp-watch-descriptors`, check https://github.com/yyoncho/lsp-mode/blob/master/lsp-mode.el#L1776 - can we used that?

PR #3014, opened 2021-07-24
Create a more sane default for phpactor

elken:
The executable for phpactor doesn't need to be installed from composer, it can
just be added to `$PATH`, as such the default should be the result of
`executable-find`.

Personally, I think it's cleaner to just rely on `executable-find` & maybe add a note to the docs RE: installation

### DIFF
--- a/clients/lsp-php.el
+++ b/clients/lsp-php.el
@@ -385,8 +385,8 @@ already present."
 
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection (list (or (executable-find "phpactor")
-                                                  (f-join (lsp-php-get-composer-dir) "vendor/phpactor/phpactor/bin/phpactor")) "language-server"))
+  :new-connection (lsp-stdio-connection (lambda () (list (or (executable-find "phpactor")
+                                                        (f-join (lsp-php-get-composer-dir) "vendor/phpactor/phpactor/bin/phpactor")) "language-server")))
   :activation-fn (lsp-activate-on "php")
   ;; `phpactor' is not really that feature-complete: it doesn't support
   ;; `textDocument/showOccurence' and sometimes errors (e.g. find references on
--- a/clients/lsp-php.el
+++ b/clients/lsp-php.el
@@ -378,15 +378,15 @@ already present."
   :link '(url-link "https://github.com/phpactor/phpactor")
   :group 'lsp-mode)
 
-(defcustom lsp-phpactor-path (or (executable-find "phpactor")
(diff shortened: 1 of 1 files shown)


clients/lsp-php.el
```diff
@@ -378,20 +378,15 @@ already present."
   :link '(url-link "https://github.com/phpactor/phpactor")
   :group 'lsp-mode)
 
-(defcustom lsp-phpactor-path nil
+(defcustom lsp-phpactor-path (or (executable-find "phpactor")
```
yyoncho, line 381: execute-find should not be executed on the top level. This means that it will be executed for all lsp-mode users. All calculations should happen in the lsp-stdio-connection lambda

elken: Ah that makes sense, apologies! Amending

clients/lsp-php.el
```diff
@@ -385,13 +385,8 @@ already present."
 
 (lsp-register-client
  (make-lsp-client
-  :new-connection (lsp-stdio-connection
-                   (lambda ()
-                     (unless lsp-php-composer-dir
-                       (setq lsp-php-composer-dir (lsp-php-get-composer-dir)))
-                     (unless lsp-phpactor-path
-                       (setq lsp-phpactor-path (f-join lsp-php-composer-dir "vendor/phpactor/phpactor/bin/phpactor")))
```
yyoncho, line 393: It seems to me that you should change:


> (unless lsp-phpactor-path
                       (setq lsp-phpactor-path (f-join lsp-php-composer-dir "vendor/phpactor/phpactor/bin/phpactor")))

to 

> (unless lsp-phpactor-path
  (setq lsp-phpactor-path 
        (or (executable-find "phpactor")
            (f-join lsp-php-composer-dir "vendor/phpactor/phpactor/bin/phpactor"))))


That way you will allow customizing lsp-phpactor-path when phpactor is not on the path. WDYT?

elken: I agree with the sentiment, if that's the desired behaviour surely

```lisp
(or lsp-phpactor-path
    (executable-find "phpactor"))
```

makes most sense?

yyoncho: You should try lsp-php-composer-dir as well if executable-find does not return, right?

yyoncho: ... but yes

elken: Yes I can make that the fallback but if the first two checks fail that likely will too, but equally there's no reason _not_ to add. I'll commit the requested change now :)

yyoncho (approved):

yyoncho:
Thank you!

elken:
> Thank you!

No worries! Nice working with you again :)

PR #3031, opened 2021-08-03
Provide a better working minimal example

trev-dev:
Please let me know if I'm off base here.

I'm assuming most people will just be adding their custom configs to their inits and may overlook the need to wait for lsp-mode to load before attempting to use it's functions/variables, so I've updated the minimal example to reflect what I needed to do to get a new language working.

I've also added an example of what setting web-mode up with a specific file-type looks like.

### DIFF
--- a/docs/page/adding-new-language.md
+++ b/docs/page/adding-new-language.md
@@ -30,6 +30,20 @@ buffer language. When the `major-mode` is not sufficient to determine the
 language (e.g. `web-mode` is used for `javascript`, `html`, and `css`) you can put
 regex.
 
+Here's an example of how to specify a filetype along with web-mode. In this case it's Shopify `.liquid` files.
+
+```elisp
+;; Shopify template lsp with theme-check
+(with-eval-after-load 'lsp-mode
+  (add-to-list 'lsp-language-id-configuration
+    '(web-mode . "\\.liquid\\'"))
+
+  (lsp-register-client
+    (make-lsp-client :new-connection (lsp-stdio-connection "theme-check-language-server")
+                     :activation-fn (lsp-activate-on "\\.liquid\\'")
+                     :server-id 'theme-check)))
+```
+
 If the language server supports environment variables to control
 additional behavior, you can register that by using the
 `:environment-fn` function, like the Bash language client does:
--- a/docs/page/adding-new-language.md
+++ b/docs/page/adding-new-language.md
@@ -11,17 +11,18 @@ corresponding mode -\> language id - in this case `(python-mode .
 
 
 ``` elisp
-(defvar lsp-language-id-configuration
-  '(...
-    (python-mode . "python")
-    ...))
-;; if you are adding the support for your language server in separate repo use
-;; (add-to-list 'lsp-language-id-configuration '(python-mode . "python"))
+(with-eval-after-load 'lsp-mode
+  (defvar lsp-language-id-configuration
+    '(...
+      (python-mode . "python")
+      ...))
+  ;; if you are adding the support for your language server in separate repo use
+  ;; (add-to-list 'lsp-language-id-configuration '(python-mode . "python"))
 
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
LGTM

jcs090218 (approved):

trev-dev:
Hey, sorry guys. I don't know how to make Circle Ci happy with a failed check on a document. Please let me know how I may be of assistance if need be.

docs/page/adding-new-language.md
```diff
@@ -11,24 +11,39 @@ corresponding mode -\> language id - in this case `(python-mode .
...
-
-(lsp-register-client
- (make-lsp-client :new-connection (lsp-stdio-connection "pyls")
-                  :activation-fn (lsp-activate-on "python")
-                  :server-id 'pyls))
+(with-eval-after-load 'lsp-mode
+  (defvar lsp-language-id-configuration
+    '(...
+      (python-mode . "python")
+      ...))
```
yyoncho, line 18: this should not be in with-eval-after-load. If you are changing the defvar then you should do a PR not put in with-eval-after-load

docs/page/adding-new-language.md
```diff
@@ -11,24 +11,39 @@ corresponding mode -\> language id - in this case `(python-mode .
...
 language (e.g. `web-mode` is used for `javascript`, `html`, and `css`) you can put
 regex.
 
+Here's an example of how to specify a filetype along with web-mode. In this case it's Shopify `.liquid` files.
+
+```elisp
+;; Shopify template lsp with theme-check
+(with-eval-after-load 'lsp-mode
+  (add-to-list 'lsp-language-id-configuration
+    '(web-mode . "\\.liquid\\'"))
```
yyoncho, line 39: In this example, "\\.liquid\\'" should be a language like `"java"`, `"c"`. Also, selecting web-mode as an example is not appropriate since web-mode can be used for multiple languages.

docs/page/adding-new-language.md
trev-dev: This was in the context of adding an lsp-language-id in one's init.el, where if 'lsp-mode is not yet loaded the variables and functions used become meaningless.

I may have misunderstood the context of this documented example and am happy to remove this change. Is this intended for people contributing to the supported languages?

docs/page/adding-new-language.md
trev-dev: This specific example is an edge-case for Shopify's own flavor of liquid template syntax. There is no solid, pre-existing program mode for it. This is why web-mode is used in this example. It is able to highlight Shopify liquid syntax with only a few minor issues. All I have to go on is the extension for this case.

I am open to better ways to do this aside from writing my own "shopify-mode" package. That's something I'm wanting to do in the long term. For everyone else right now, I'm not sure what to suggest.

docs/page/adding-new-language.md
trev-dev: I have reverted the commit that included this change.

docs/page/adding-new-language.md
trev-dev: I have refined both my config and this example to use a derived major mode. It's important to note that the documentation already stated that regexp is an acceptable means of setting up a custom language server, but I do see the benefit in being explicit with your own major major mode. I included a link to some more documentation to help new users.

Commit f086163ff1e87a2a3d6058a0aca70ae32d88fe10

trev-dev: I hope these updates have been satisfactory 🙂. Please let me know if there's more I may do

trev-dev:
Please let me know if there's anything more I may do to be helpful here.

yyoncho (approved):

yyoncho:
Thank you!

Sorry for the delay, I was on a vacation.

trev-dev:
Hope you had a good one!

PR #3034, opened 2021-08-04
Use run-with-idle-timer for lsp-deferred

### LINKED ISSUE #3033: LSP over tramp doesn't start while opening the file, but works fine when running M-x lsp after the file is opened
### Bug description

The LSP configuration
```elisp
(after! lsp-solargraph
  (lsp-register-client
   (make-lsp-client
    :new-connection (lsp-tramp-connection
                     #'lsp-solargraph--build-command)
    :major-modes '(ruby-mode enh-ruby-mode)
    :priority -1
    :remote? t
    :multi-root lsp-solargraph-multi-root
    :library-folders-fn (lambda (_workspace) lsp-solargraph-library-directories)
    :server-id 'ruby-ls-remote
                      (with-lsp-workspace workspace
                        (lsp--set-configuration
                         (lsp-configuration-section "solargraph")))))))

```

NOTE: This only happens for the remote projects. Opening localhost files would start LSP servers fine.

While opening file in the buffer through `SPC p f` or `SPC f f` the LSP initialization is attempted but fails to start with the following warning:

```
LSP :: The following servers support current file but do not have automatic installation configuration: ruby-ls-remote
You may find the installation instructions at https://emacs-lsp.github.io/lsp-mode/page/languages.
(If you have already installed the server check *lsp-log*).
```

On debugging, it seems that the `solargraph` binary isn't found while opening file.
```elisp
       ((setq clients (unless matching-clients
                        (lsp--filter-clients (-andfn #'lsp--matching-clients?
                                                     (-not #'lsp--server-binary-present?)))))
        (lsp--warn "The following servers support current file but do not have automatic installation configuration: %s
You may find the installation instructions at https://emacs-lsp.github.io/lsp-mode/page/languages.
\(If you have already installed the server check *lsp-log*)."
                   (mapconcat (lambda (client)
                                (symbol-name (lsp--client-server-id client)))
                              clients
                              " ")))
```

However, while running `(lsp)` or `M-x lsp` works after the file is opened in a buffer.
```
LSP :: Connected to [ruby-ls-remote:1615687].
```

### Steps to reproduce

- Setup LSP for the remote project.
- Ensure LSP server binary is present on a remote server.
- Open the remote file.
- Check if the LSP server has started or not
- Run `M-x lsp` and see if it has started.

It can be reproduced if the tramp takes long enough that the `lsp-deferred` timer triggers before the environment is set and `solargraph` binary is not found. That's what my hunch is based on the what I found by setting `run-with-timer 5` and `run-with-idle-timer 0` since that worked as expected.

### Expected behavior

The LSP is enabled for ruby-mode (configured for remote). While opening any ruby file in a remote project, the LSP server should start if not started already or the file should connect to the server that's already running.

`lsp-solargraph` (ruby)

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

prashantvithani:
`run-with-timer` triggers immediately after the buffer is visible. But
the environment is not set for the remote files at the time when LSP
binary is looked up.  This results in `matching-clients` to be nil.

Changing it to `run-with-idle-timer` ensures the buffer is done with
loading remote file and setting the environment and runs immediately
after the file is loaded in the buffer.

FIX #3033

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8320,11 +8320,11 @@ This avoids overloading the server with many files when starting Emacs."
   (let ((buffer (current-buffer)))
     ;; Avoid false positives as desktop-mode restores buffers by deferring
     ;; visibility check until the stack clears.
-    (run-with-timer 0 nil (lambda ()
-                            (when (buffer-live-p buffer)
-                              (with-current-buffer buffer
-                                (unless (lsp--init-if-visible)
-                                  (add-hook 'window-configuration-change-hook #'lsp--init-if-visible nil t))))))))
+    (run-with-idle-timer 0 nil (lambda ()
+                                 (when (buffer-live-p buffer)
+                                   (with-current-buffer buffer
+                                     (unless (lsp--init-if-visible)
+                                       (add-hook 'window-configuration-change-hook #'lsp--init-if-visible nil t))))))))
 
 
 


prashantvithani:
ping @yyoncho @vibhavp

yyoncho (approved):

jcs090218 (approved):

prashantvithani:
Is this good to merge?

ericdallo (approved):

PR #3035, opened 2021-08-05
Pass workspace to lsp diagnostics hook.

kurnevsky:
This fixes diagnostics handling in case user switched to a non-lsp buffer.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -541,7 +541,7 @@ It contains the operation source."
 (defcustom lsp-after-diagnostics-hook nil
   "Hooks to run after diagnostics are received.
 Note: it runs only if the receiving buffer is open. Use
-`lsp-diagnostics-updated-hook'if you want to be notified when
+`lsp-diagnostics-updated-functions'if you want to be notified when
 diagnostics have changed."
   :type 'hook
   :group 'lsp-mode)
@@ -550,6 +550,14 @@ diagnostics have changed."
   'lsp-diagnostics-updated-hook "lsp-mode 6.4")
 
 (defcustom lsp-diagnostics-updated-hook nil
+  "Hooks to run after diagnostics are received."
+  :type 'hook
+  :group 'lsp-mode)
+
+(make-obsolete-variable
+ 'lsp-diagnostics-updated-hook "lsp-diagnostics-updated-functions" "7.1" 'set)
+
+(defcustom lsp-diagnostics-updated-functions nil
   "Hooks to run after diagnostics are received."
   :type 'hook
   :group 'lsp-mode)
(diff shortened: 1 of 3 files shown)


ericdallo:
c/c @kiennq

lsp-mode.el
```diff
@@ -2084,7 +2084,7 @@ WORKSPACE is the workspace that contains the diagnostics."
         (remhash file workspace-diagnostics)
       (puthash file (append diagnostics nil) workspace-diagnostics))
 
-    (run-hooks 'lsp-diagnostics-updated-hook)))
+    (run-hook-with-args 'lsp-diagnostics-updated-hook workspace)))
```
kiennq, line 2087: Since you have changed this hook to have an argument, can you change its name to `lsp-diagnostics-updated-functions` instead? https://www.gnu.org/software/emacs/manual/html_node/elisp/Hooks.html

We need to declare the old hook as obsolete too.

yyoncho: Changing this will break also the clients like lsp-treemacs. Can we let bind lsp--cur-workspace instead?

kurnevsky: Sure, didn't know about this convention.

yyoncho: Check also if it is already bound (we bind it for the notifications so they are executed in the context of the source server).

kurnevsky: @yyoncho I preserved old hook, just made it obsolete. Should be fine?

> Check also if it is already bound (we bind it for the notifications so they are executed in the context of the source server).

Not sure what you mean here. That lsp--cur-workspace should be defined even if user switched buffer? Or I didn't get you?

kurnevsky: > Can we let bind lsp--cur-workspace instead?

Can do this if you think it's better. I'm not sure which approach would be preferable :)

yyoncho:
I spent some time reading the PR and I wonder why the current buffer is relevant? Even if lsp-diagnostics--flycheck-report runs in non-lsp buffer it will look in the current workspace and trigger flycheck update only if the lsp-mode enabled buffer is visible. Also in this handler lsp--cur-workspace should be let bound from here - https://github.com/yyoncho/lsp-mode/blob/master/lsp-mode.el#L6271

kurnevsky:
@yyoncho you are right, indeed. I thought it's a problem with current workspace, but looks like it's something different. Often when I edit something and switch to flycheck buffer I don't get errors list refreshed, even after switching back. Closing it for now, will reopen if I manage to find the cause.

yyoncho:
@kurnevsky your issue is caused by the fact that we refresh only the visible buffer. This is for performance reasons because `(find-buffer-visiting file)` is dead slow. You may try to add idle-buffer-switch to flycheck-check-syntax-automatically. I havent tested it but it should work.

kurnevsky:
> @kurnevsky your issue is caused by the fact that we refresh only the visible buffer.

Visible or active? I usually have flycheck buffer opened in a separate window so main code buffer is still visible. But it still happens - usually in this case flycheck has empty errors list while lsp-treemacs shows errors. And errors in flycheck reapper only I modify code buffer.

yyoncho:
```
(or
                             (not (bufferp buffer))
                             (and (get-buffer-window buffer)
                                  (not (-contains? (buffer-local-value 'lsp-on-idle-hook buffer)
                                                   'lsp-diagnostics--flycheck-buffer))))
````

This is the check.

PR #3037, opened 2021-08-06
[clojure] Refresh semantic tokens after cider disconnected

ericdallo:
When using cider with lsp-mode, after the REPL is disconnected cider flush the font-faces, removing the tokens from semantic-tokens, adding the need to edit the file or `M-x` `lsp`.

### DIFF
--- a/clients/lsp-clojure.el
+++ b/clients/lsp-clojure.el
@@ -309,7 +309,8 @@ If there are more arguments expected after the line and column numbers."
 
 (defun lsp-clojure-semantic-tokens-refresh ()
   "Force refresh semantic tokens."
-  (when (lsp-workspaces)
+  (when (and lsp-semantic-tokens-enable
+             (lsp-find-workspace 'clojure-lsp (buffer-file-name)))
     (lsp-semantic-tokens--enable)))
 
 (with-eval-after-load 'cider
--- a/clients/lsp-clojure.el
+++ b/clients/lsp-clojure.el
@@ -304,5 +304,19 @@ If there are more arguments expected after the line and column numbers."
 
 (lsp-consistency-check lsp-clojure)
 
+;; Cider integration
+
+(declare-function lsp-semantic-tokens--enable "lsp-semantic-tokens")
+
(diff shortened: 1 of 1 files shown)


clients/lsp-clojure.el
```diff
@@ -304,5 +304,19 @@ If there are more arguments expected after the line and column numbers."
 
 (lsp-consistency-check lsp-clojure)
 
+;; Cider integration
+
+(declare-function lsp-semantic-tokens--enable "lsp-semantic-tokens")
```
kiennq, line 309: Why not require `lsp-semantic-tokens` directly?
Also, the `lsp-semantice-tokens--enabled` is autoloader already, so this declaration may not needed

ericdallo: oh good to know, so I don't need to even require? how to suppress the warning?

clients/lsp-clojure.el
```diff
@@ -304,5 +305,17 @@ If there are more arguments expected after the line and column numbers."
...
 
+;; Cider integration
+
+(defun lsp-clojure-semantic-tokens-refresh ()
+  "Force refresh semantic tokens."
+  (when (lsp-workspaces)
+    (lsp-semantic-tokens--enable)))
+
+(with-eval-after-load 'cider
+  (when lsp-semantic-tokens-enable
```
yyoncho, line 316: seems like this check should go in lsp-clojure-semantic-tokens-refresh as well.

clients/lsp-clojure.el
```diff
@@ -304,5 +305,17 @@ If there are more arguments expected after the line and column numbers."
 
 (lsp-consistency-check lsp-clojure)
 
+;; Cider integration
+
+(defun lsp-clojure-semantic-tokens-refresh ()
+  "Force refresh semantic tokens."
+  (when (lsp-workspaces)
```
yyoncho, line 312: maybe check if clojure-lsp server is running

yyoncho (approved):

ericdallo:
Done @yyoncho !

jcs090218 (approved):

PR #3042, opened 2021-08-12
Improve the Angular client

daanturo:
* Set lsp-clients-angular-language-server-command after the first run
to avoid the lag introduced by running the shell command.

Previously, if I add `benchmark-progn` after `lsp-stdio-connection                   (lambda ()`

Then after each `revert-buffer` & new file opened:
```
Elapsed time: 0.221614s
Elapsed time: 0.220887s
Elapsed time: 0.228161s
```

Certainly it's not a good experience for users.

Now:

First connection:
```
Elapsed time: 0.220515s
Elapsed time: 0.000001s [2 times]
```
After that, for each `revert-buffer` & new file opened:
```
Elapsed time: 0.000002s [2 times]
Elapsed time: 0.000001s
```
That's a huge speedup!

* Make Angular server's NodeJS prefix path configurable.

Because sometimes we don't use npm, and maybe there are multiple
installed versions of NodeJS, too.

Edit: I don't know how to edit documentation on [Angular - LSP Mode - LSP support for Emacs ](https://emacs-lsp.github.io/lsp-mode/page/lsp-angular/)

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,7 @@
 * Changelog
 ** Release 7.1
+  * Add ~lsp-clients-angular-node-get-prefix-command~ to get the Angular server from another location which is still has ~/lib/node_modules~ in it.
+  * Set ~lsp-clients-angular-language-server-command~ after the first connection to speed up subsequent connections.
   * Add ~lsp-sql-execute-paragraph~ to execute the current paragraph (like ~sql-send-paragraph~).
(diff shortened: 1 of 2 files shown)


yyoncho (approved):

daanturo:
There are many overly long lines, should we split them, too?

yyoncho (approved):

yyoncho:
Thank you!

PR #3045, opened 2021-08-14
Add very basic lsp client for netbeans server

timfel:
This mixes the dap support with the lsp support, unfortunately. I guess this can be split, but there's a bit of coupling, since I just discover the DAP port from the output of the language server, and that means the language server installation and run script has to be a bit coupled to the dap code.

This is more a suggestion, please let me know if this even makes sense to include at this stage (it's missing a bunch of features the VSCode extension exposes).

### DIFF
--- a/clients/lsp-netbeans.el
+++ b/clients/lsp-netbeans.el
@@ -1,0 +1,129 @@
+;;; lsp-netbeans.el --- Apache Netbeans Language Server Client settings         -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2021  Tim Felgentreff
+
+;; Author: Tim Felgentreff <timfelgentreff@gmail.com>
+;; Keywords: java groovy graalvm lsp
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; LSP and DAP client for the Apache Netbeans LSP server.
+
+;; Currently missing many features that are in the VSCode extension, including, but not limited to:
+;; - All custom commands, such as
+;;     java.build.workspace
+;;     java.generate.getters
+;;     java.generate.setters
+;;     java.generate.getters.setters
+;;     java.generate.implementMethod
+;;     java.generate.overrideMethod
+;;     java.generate.equals.hashCode
+;;     java.generate.delegateMethod
+;;     java.generate.constructor
+;;     java.generate.logger
+;;     java.generate.toString
+;;     graalvm.pause.script
+;;     java.super.implementation
+;; - Groovy support
+;; - Managing/running/debugging unittest directly
+;; - GraalVM Native Image launch/debug configuration
+;; - Attach to JVM process debug configs
+;; - Launch JVM debug config
+
+;;; Code:
+
+(require 'lsp-mode)
+(require 'dap-mode)
+
+(defcustom lsp-netbeans-download-url "https://marketplace.visualstudio.com/_apis/public/gallery/publishers/ASF/vsextensions/apache-netbeans-java/12.4.301/vspackage"
+  "URL to download the Apache Netbeans LSP server from"
+  :group 'lsp-netbeans
+  :type 'string)
+
+(defcustom lsp-netbeans-install-dir (f-join lsp-server-install-dir "asf.apache-netbeans-java")
+  "Apache Netbeans language server installation dir"
+  :group 'lsp-netbeans
+  :type 'string)
+
+(defun lsp-netbeans-server-command (main-port)
+  `(,(f-join lsp-netbeans-install-dir "run.sh")
+    "--start-java-debug-adapter-server=listen:0"
+    ,(format "--start-java-language-server=listen:%d" main-port)))
+
+(defun lsp-netbeans--install-server (client callback error-callback update?)
+  (unless (and (not update?)
+               (f-exists? lsp-netbeans-install-dir))
+    (make-directory lsp-netbeans-install-dir t)
+    (let ((download-path (f-join lsp-netbeans-install-dir "vspackage")))
+      (lsp-download-install
+       (lambda ()
+         (let ((run-script-path (f-join lsp-netbeans-install-dir "run.sh")))
+           (with-temp-file run-script-path
+             (insert "#!/bin/bash
+                      cd `dirname $0`
+                      cd extension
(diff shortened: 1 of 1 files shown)


clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+ (make-lsp-client
+  :new-connection (lsp-tcp-connection 'lsp-netbeans-server-command)
+  :activation-fn (lsp-activate-on "java")
+  :server-id 'netbeans
+  :initialization-options (list :nbcodeCapabilities
+                                (list
+                                 :testResultsSupport (lsp-json-bool nil)
+                                 :statusBarMessageSupport (lsp-json-bool nil)
+                                 :wantsGroovySupport (lsp-json-bool nil)))
+  :priority 10
```
timfel, line 96: I'm not sure what to choose as default here. I guess most people would prefer the Eclipse server.

ericdallo:
Could you please add a documentation entry following https://emacs-lsp.github.io/lsp-mode/page/adding-new-language/ ?

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+       :decompress :zip))))
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-tcp-connection 'lsp-netbeans-server-command)
+  :activation-fn (lsp-activate-on "java")
+  :server-id 'netbeans
+  :initialization-options (list :nbcodeCapabilities
+                                (list
+                                 :testResultsSupport (lsp-json-bool nil)
```
timfel, line 93: These three should probably be settings, as in the VSCode extension, but I'm not sure what they even imply.

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+;; - Groovy support
+;; - Managing/running/debugging unittest directly
+;; - GraalVM Native Image launch/debug configuration
+;; - Attach to JVM process debug configs
+;; - Launch JVM debug config
+
+;;; Code:
+
+(require 'lsp-mode)
+(require 'dap-mode)
```
ericdallo, line 49: The CI is failing because dap-mode is not a dependency of lsp-mode, you need to check if dap-mode is installed or move this client to a separated repo which I don't think it makes sense right now

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+    (make-directory lsp-netbeans-install-dir t)
+    (let ((download-path (f-join lsp-netbeans-install-dir "vspackage")))
+      (lsp-download-install
+       (lambda ()
+         (let ((run-script-path (f-join lsp-netbeans-install-dir "run.sh")))
+           (with-temp-file run-script-path
+             (insert "#!/bin/bash
+                      cd `dirname $0`
+                      cd extension
+                      node out/nbcode.js $@ >&2")
```
timfel, line 78: The `>&2` bit is the first part of the hack to discover the DAP port. Not sure what a better way would be, the VSCode extension seems to also just read the process output.

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+  :download-server-fn #'lsp-netbeans--install-server))
+
+(defun dap-netbeans--populate-attach-args (conf)
+  (dap--put-if-absent conf :hostName (read-string "Enter host: " "localhost"))
+  (dap--put-if-absent conf :port (read-string "Enter port: " "8000"))
+  (dap--put-if-absent conf :host "localhost")
+  (dap--put-if-absent conf :name (format "%s(%s)" (plist-get conf :host) (plist-get conf :port)))
+  conf)
+
+(defun dap-netbeans--populate-default-args (conf)
```
ericdallo, line 107: Please follow the convention of using the prefix of functions/vars as the client like `lsp-netbeans--...`

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+(defun lsp-netbeans--install-server (client callback error-callback update?)
+  (unless (and (not update?)
+               (f-exists? lsp-netbeans-install-dir))
+    (make-directory lsp-netbeans-install-dir t)
+    (let ((download-path (f-join lsp-netbeans-install-dir "vspackage")))
+      (lsp-download-install
+       (lambda ()
+         (let ((run-script-path (f-join lsp-netbeans-install-dir "run.sh")))
+           (with-temp-file run-script-path
+             (insert "#!/bin/bash
```
timfel, line 75: This run script is a bit weird. I don't know enough about node, it seems I have to run it from the extension module, otherwise, it's not finding the dependencies? The script of course assumes that `bash` works and that a recent `node` version is on the `PATH`, very sub-optimal. Any suggestions for a better solution here are very welcome :)

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+  conf)
+
+(defun dap-netbeans--populate-default-args (conf)
+  (setq conf (plist-put conf :type "java8+"))
+
+  (setq conf (pcase (plist-get conf :request)
+               ("attach" (dap-netbeans--populate-attach-args conf))
+               (_ (dap-netbeans--populate-attach-args conf))))
+
+  (plist-put conf :debugServer (with-current-buffer (get-buffer "*netbeans::stderr*")
```
timfel, line 114: Relying on the `*netbeans::stderr*` buffer name here also seems very fragile to me.

clients/lsp-netbeans.el
yyoncho: Alternatively, can  you extract the dap part in dap-mode repo?

clients/lsp-netbeans.el
yyoncho: those shipping with lsp-mode should be -1, thus if you install lsp-java it will be preffered.

yyoncho (commented):
Thank you, It will be great if you can provide a short comparison vs lsp-java.

clients/lsp-netbeans.el
```diff
@@ -0,0 +1,129 @@
...
+;;     java.generate.getters.setters
+;;     java.generate.implementMethod
+;;     java.generate.overrideMethod
+;;     java.generate.equals.hashCode
+;;     java.generate.delegateMethod
+;;     java.generate.constructor
+;;     java.generate.logger
+;;     java.generate.toString
+;;     graalvm.pause.script
+;;     java.super.implementation
```
yyoncho, line 39: seems like this file will grow, maybe start lsp-netbeans as a separate package?

ericdallo: Agree with that

yyoncho: I checked the feature set here - https://marketplace.visualstudio.com/items?itemName=ASF.apache-netbeans-java and I would say that we definitely need this in a separate package. But it is still up to you to decide. @ericdallo or I can create the project under emacs-lsp and make you maintainer if you agree.

timfel: I agree that probably this should be it's own package. However, I doubt I'll be able to look into this more than a couple of days a month, so I would be a bad maintainer. I guess if no one cares and there's not a lot of issues being opened, it's fine, but otherwise I could be a bad look for the project. Maybe it's then better if I keep it in my private github account, we just put a link there from the readme and this way it won't look bad for the emacs-lsp project when I take days to respond to issues and PRs

timfel:
Closing in favor of https://github.com/timfel/lsp-netbeans

PR #3047, opened 2021-08-15
Add support for `workspace/codeLens/refresh`

ericdallo:
[SPEC](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#codeLens_refresh)

I intend to use it with [clojure-lsp](https://clojure-lsp.github.io/clojure-lsp/)

### DIFF
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -309,6 +309,13 @@ CALLBACK - callback for the lenses."
           (funcall callback lsp-lens--backend-cache lsp--cur-version)
         (lsp-lens--backend-fetch-missing lsp-lens--backend-cache callback lsp--cur-version)))))
 
+(defun lsp--lens-on-refresh (workspace)
+  "Clear lens within all buffers of WORKSPACE, refreshing all workspace buffers."
+  (cl-assert (not (eq nil workspace)))
+  (cl-loop
+   for ws-buffer in (lsp--workspace-buffers workspace) do
+   (lsp-lens-refresh t ws-buffer)))
+
 ;;;###autoload
 (defun lsp-lens--enable ()
   "Enable lens mode."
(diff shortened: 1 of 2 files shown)


lsp-lens.el
```diff
@@ -309,6 +309,13 @@ CALLBACK - callback for the lenses."
           (funcall callback lsp-lens--backend-cache lsp--cur-version)
         (lsp-lens--backend-fetch-missing lsp-lens--backend-cache callback lsp--cur-version)))))
 
+(defun lsp--lens-on-refresh (workspace)
+  "Clear lens within all buffers of WORKSPACE, refreshing all workspace buffers."
+  (cl-assert (not (eq nil workspace)))
+  (cl-loop
```
kiennq, line 315: nit: why using `cl-loop` instead of `mapc`?

kiennq (approved):

lsp-lens.el
ericdallo: I just copied from the same did on semantic-tokens 😅 I don't like cl-loop either

jcs090218 (approved):

lsp-lens.el
```diff
@@ -309,6 +309,13 @@ CALLBACK - callback for the lenses."
...
+   for ws-buffer in (lsp--workspace-buffers workspace) do
+   (lsp-lens-refresh t ws-buffer)))
```
yyoncho, line 317: can we avoid calling refresh for visible buffers unconditionally, and schedule the refresh once the buffer is visible?

ericdallo: I think I missed this comment, sure, that seems a good idea, how can we do that exactly?

yyoncho: we have a similar piece of code for flycheck errors. We refresh only the visible buffer. And we should ensure that non visible buffers will refresh when switching to them.

PR #3051, opened 2021-08-20
lsp-rust/rust-analyzer: expose experimental_procAttrMacros setting

japaric:
this config flag was added in rust-analyzer/rust-analyzer#9128

when enabled, the existing `lsp-rust-analyzer-expand-macro` function will also
expand proc macro attributes

as this is an experimental feature, the default value of the setting is nil


flodiebold (approved):

brotzeit (approved):
:+1:

PR #3072, opened 2021-08-31
lsp-rust: Add support for opening external docs

ALSchwalm:
Add support for using the rust-analyzer 'experimental/externalDocs' request to get a URL for external documentation. These 'experimental' endpoints are used elsewhere in lsp-rust, so this isn't a change on that front.

An alternative here is that we just return the URL instead of using `browse-url`. I will change to that if you prefer.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -1094,6 +1094,18 @@ open in a new window."
         (lsp--warn "Couldn't find a Cargo.toml file or your version of rust-analyzer doesn't support this extension"))
     (lsp--error "OpenCargoToml is an extension available only with rust-analyzer")))
 
+(defun lsp-rust-analyzer-open-external-docs ()
+  "Open a URL for documentation related to the current TextDocumentPosition.
+
+Rust-Analyzer LSP protocol documented here
+https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#open-external-documentation"
+  (interactive)
+  (-if-let* ((params (lsp-make-rust-analyzer-open-external-docs-params
+                      :text-document (lsp--text-document-identifier)
+                      :position (lsp--cur-position)))
+             (url (lsp-request "experimental/externalDocs" params)))
+      (browse-url url)
+    (lsp--warn "Couldn't find documentation URL or your version of rust-analyzer doesn't support this extension")))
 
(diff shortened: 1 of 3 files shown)


clients/lsp-rust.el
```diff
@@ -1094,6 +1095,22 @@ open in a new window."
         (lsp--warn "Couldn't find a Cargo.toml file or your version of rust-analyzer doesn't support this extension"))
     (lsp--error "OpenCargoToml is an extension available only with rust-analyzer")))
 
+(defun lsp-rust-analyzer-open-external-docs ()
+  "Open a URL for documentation related to the current TextDocumentPosition.
+
+Rust-Analyzer LSP protocol documented here
+https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#open-external-documentation"
+  (interactive)
+  (-if-let (workspace (lsp-find-workspace 'rust-analyzer))
```
yyoncho, line 1104: can you remove that -if-let? The if-let in `lsp-rust-analyzer-open-cargo-toml` because you can call this function in `dired` buffer for example and it will work. In this case, there is no need to look for workspace - just execute the function using the current context.  You may follow `lsp-rust-analyzer-move-item` as an example.

clients/lsp-rust.el
```diff
@@ -1094,6 +1095,22 @@ open in a new window."
...
+      (-if-let* ((response (with-lsp-workspace workspace
+                             (lsp-send-request (lsp-make-request
```
yyoncho, line 1106: can you use lsp-request instead of lsp-send-request(it is a bit simpler)?

yyoncho (commented):
Looks good! Few stylistic improvements:

clients/lsp-rust.el
```diff
@@ -27,6 +27,7 @@
 (require 'lsp-mode)
 (require 'ht)
 (require 'dash)
+(require 'browse-url)
```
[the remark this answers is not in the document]
yyoncho: (this is save load time for those who are not going to use browse function)

ALSchwalm:
Thanks for the quick review! I believe I've addressed your feedback

yyoncho:
> EDIT: actually, should I go ahead and update the docs?

If you wish.

ALSchwalm:
Done

yyoncho:

PR #3073, opened 2021-08-31
add missing option in texlab install command (documentation)

gamboz:
I think the same option (`--locked`) should be added to l.619, but have not tested.
Please see also see https://github.com/latex-lsp/texlab/issues/474 that points to https://github.com/latex-lsp/texlab#building-from-source


yyoncho:

(and sorry for the delayed review)

PR #3078, opened 2021-09-02
When auto-configure is enabled, remove hooks for disabled options

### LINKED ISSUE #2525: Can't disable breadcrumbs without restarting.
**Describe the bug**

I updated lsp-mode and noticed the new breadcrumbs. I wanted to disable them so I set `lsp-headerline-breadcrumb-enable` to nil and re-opened the file but the breadcrumbs still show up.

After restarting Emacs, the setting seemed to be respected.

**To Reproduce**

1. Open a file with breadcrumbs enabled at the top.
2. Set `lsp-headerline-breadcrumb-enable` to nil.
3. Close and re-open file - notice breadcrumbs still show up.

**Expected behavior**

Breadcrumbs don't show up after disabling `lsp-headerline-breadcrumb-enable`.

gopls

**OS**
Mac

aspiers:
**N.B. please review the second commit carefully, since I am brand new to this project and not sure if it is correct - perhaps it breaks something else.**

When `lsp-auto-configure` is enabled, hooks are automatically added to `lsp-configure-hook` for any options which are enabled.  However, if any of those options are later disabled, the hooks were not removed from `lsp-configure-hook`; therefore the corresponding behaviour did not get disabled.

So add the relevant `remove-hook` calls when options are not enabled, and also call `lsp--auto-configure` from `lsp-configure-buffer`, so that any hooks which need to be removed will get removed.

Fixes #2525.


lsp-mode.el
```diff
@@ -6977,18 +6979,25 @@ returns the command to execute."
...
+    (remove-hook 'lsp-configure-hook 'lsp-headerline-breadcrumb-mode))
+  (if lsp-modeline-code-actions-enable
+      (add-hook 'lsp-configure-hook 'lsp-modeline-code-actions-mode)
+    (remove-hook 'lsp-configure-hook 'lsp-modeline-code-actions-mode))
+  (if lsp-modeline-diagnostics-enable
+      (add-hook 'lsp-configure-hook 'lsp-modeline-diagnostics-mode)
+    (remove-hook 'lsp-configure-hook 'lsp-modeline-diagnostics-mode))
+  (if lsp-modeline-workspace-status-enable
+      (add-hook 'lsp-configure-hook
+                'lsp-modeline-workspace-status-mode)
```
jcs090218, line 6993: Maybe just put these in the same line for consistency?

jcs090218 (approved):

lsp-mode.el
aspiers: Done.

aspiers:
Addressed feedback and rebased on latest `master`.

yyoncho (approved):

PR #3082, opened 2021-09-05
Use eldoc-documentation hook instead of variable

### LINKED ISSUE #3081: Migrate to the new eldoc-documentation-functions hook
Mentioned in [this](https://github.com/emacs-lsp/lsp-mode/blob/d9c8ed3a932b568614d9bbf682271cf43bb8ec73/lsp-mode.el#L1124) comment here. 

lsp-mode still uses the old `eldoc-documentation-function`  variable for hooking into eldoc meaning you can't support multiple documentation functions at the same time. For example lsp-mode has support for flymake (when you don't have flycheck installed) but it doesn't show any eldoc messages for flymake because only `eldoc-documentation-function` is configured.

There're also other benefits to using hooks instead of the current advice approach.

I'm willing to work on this, I'm just curious how some of the stuff should be handled. For example I don't understand the value of `lsp-eldoc-hook` so can I depracate it?

mohkale:
closes #3081

**note**: Still need to update the changelog.

**note**: eglot seems to disable the local value of the hook and override it with its own documentation functions only. It also then enables flymake-mode to ensure flymake is in the eldoc hook. We may want to do something similair :?

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3765,7 +3765,7 @@ yet."
    (t
     (lsp-unconfig-buffer)
 
-    (remove-hook 'eldoc-documentation-functions #'lsp-eldoc-function nil t)
+    (remove-hook 'eldoc-documentation-functions #'lsp-eldoc-function t)
     (remove-hook 'post-command-hook #'lsp--post-command t)
     (remove-hook 'after-change-functions #'lsp-on-change t)
     (remove-hook 'after-revert-hook #'lsp-on-revert t)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4764,7 +4764,7 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
    (->> lsp--cur-workspace lsp--workspace-client lsp--client-response-handlers (remhash id))
    (lsp-notify "$/cancelRequest" `(:id ,id))))
 
-(defun lsp-eldoc-function (cb &optional _ignore)
+(defun lsp-eldoc-function (cb &rest _ignored)
   "`lsp-mode' eldoc function to display hover info (based on `textDocument/signatureHelp')."
   (if (and lsp--hover-saved-bounds
            (lsp--point-in-bounds-p lsp--hover-saved-bounds))
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -563,10 +563,7 @@ The hook will receive two parameters list of added and removed folders."
   :type 'hook
   :group 'lsp-mode)
 
-(defcustom lsp-eldoc-hook '(lsp-hover)
-  "Hooks to run for eldoc."
-  :type 'hook
-  :group 'lsp-mode)
+(make-obsolete 'lsp-eldoc-hook 'eldoc-documentation-functions "lsp-mode 8.0.0")
 
 (defcustom lsp-before-apply-edits-hook nil
   "Hooks to run before applying edits."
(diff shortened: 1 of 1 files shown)


mohkale:
For some reason I'm getting an error when killing emacs while lsp-mode is running.

Screenshot_20210905_163637

Not sure how to debug this :cry:.

yyoncho:
Looks good. I think that we should depend on a newer eldoc version from elpa. Note that we support emacs 26, and it won't have that new eldoc functionality.

> For some reason I'm getting an error when killing emacs while lsp-mode is running.

AFAICS CI fails with a similar error. 

> **note**: eglot seems to disable the local value of the hook and override it with its own documentation functions only. It also then enables flymake-mode to ensure flymake is in the eldoc hook. We may want to do something similair :?

To me, it sounds like we should do it like that, otherwise, we might get a documentation function that we don't want. But enabling flymake stuff should happen only if flymake is selected as a checker.

mohkale:
@yyoncho 

> Looks good. I think that we should depend on a newer eldoc version from elpa. Note that we support emacs 26, and it won't have that new eldoc functionality.

Which version would you recommend? I currently have 1.11.0 installed and it seems to work with emacs `26.3`.

> AFAICS CI fails with a similar error

Found it. Remove hook has too many arguments :sweat:.

> To me, it sounds like we should do it like that, otherwise, we might get a documentation function that we don't want. But enabling flymake stuff should happen only if flymake is selected as a checker.

There might be an issue on that front. Last I checked flycheck doesn't use eldoc. It has its own bespoke system for showing eldoc messages that might conflict with lsp-mode.

lsp-mode.el
```diff
@@ -3743,7 +3729,7 @@ yet."
   (cond
    (lsp-managed-mode
     (when (lsp-feature? "textDocument/hover")
-      (add-function :before-until (local 'eldoc-documentation-function) #'lsp-eldoc-function)
+      (add-hook 'eldoc-documentation-functions #'lsp-eldoc-function nil t)
```
kiennq, line 3732: The `eldoc-documentation-functions` seems to be the new one introduced at emacs 28 right?
You will probably need to check for its availability before use and fall back to the previous one since we support from emacs 26

kiennq (changes requested):

yyoncho:
> Which version would you recommend? I currently have 1.11.0 installed and it seems to work with emacs `26.3`.

We should test with the newest version available.

lsp-mode.el
mohkale: @kiennq 

Looks to me like the hook function was added in commit [c0fcbd2c119b8418855f0931aceefbef717c5e53](https://github.com/emacs-mirror/emacs/commit/c0fcbd2c119b8418855f0931aceefbef717c5e53) meaning it should've been in the 26.2 release. Plus @yyoncho suggested downloading a newer version of eldoc from MELPA for use with lsp-mode so I don't see much value in trying to support both hooks and the old add-function mechanism.

yyoncho:
Checking https://elpa.gnu.org/packages/eldoc-1.11.0.el it seems like it requires emacs 26.3 but we currently support emacs 26.1/2. If others are fine, we can drop the support for 26.1/2. If we decide to do so, then we have to update the CI and add changelog entry. IMHO ATM it is a good time to drop these versions considering that we just released a stable lsp-mode release.

kiennq:
> Checking https://elpa.gnu.org/packages/eldoc-1.11.0.el it seems like it requires emacs 26.3 but we currently support emacs 26.1/2. If others are fine, we can drop the support for 26.1/2. If we decide to do so, then we have to update the CI and add changelog entry. IMHO ATM it is a good time to drop these versions considering that we just released a stable lsp-mode release.

Fine with me.

ericdallo:
Ping @mohkale

mohkale:
Hi, sorry for the long time without an update. I just started a new job and I'm not sure when I'll have the time to continue working on this, although I do intend to.

yyoncho:
changes from this pr merged in #3275

PR #3083, opened 2021-09-05
[code-lens ]Refresh lens with lsp-idle-timer

ericdallo:
Following https://github.com/emacs-lsp/lsp-mode/pull/3047#discussion_r689525106

### DIFF
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -312,7 +312,8 @@ CALLBACK - callback for the lenses."
 (defun lsp-lens--refresh-buffer ()
   "Trigger lens refresh on buffer."
   (remove-hook 'lsp-on-idle-hook #'lsp-lens--refresh-buffer t)
-  (lsp-lens-refresh t (current-buffer)))
+  (when (bound-and-true-p lsp-lens-mode)
+    (lsp-lens-refresh t)))
 
 (defun lsp--lens-on-refresh (workspace)
   "Clear lens within all buffers of WORKSPACE, refreshing all workspace buffers."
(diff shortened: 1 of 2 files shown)


lsp-lens.el
```diff
@@ -309,12 +309,19 @@ CALLBACK - callback for the lenses."
...
 (defun lsp--lens-on-refresh (workspace)
   "Clear lens within all buffers of WORKSPACE, refreshing all workspace buffers."
   (cl-assert (not (eq nil workspace)))
-  (cl-loop
-   for ws-buffer in (lsp--workspace-buffers workspace) do
-   (lsp-lens-refresh t ws-buffer)))
+  (->> (lsp--workspace-buffers workspace)
+       (mapc (lambda (buffer)
+               (lsp-with-current-buffer buffer
+                 (add-hook 'lsp-on-idle-hook #'lsp-lens--refresh-buffer nil t)
```
yyoncho, line 323: I think that here you should call refresh for the visible buffer or in case it is not visible, you should attach the hook as you do. Also, I suspect that you have to check if lens mode is enabled in the buffers that you refresh the lenses.

ericdallo: Done!

lsp-mode.el
```diff
@@ -8306,10 +8306,14 @@ This issue might be caused by:
 5. You have disabled the `lsp-mode' clients for that file. (Check `lsp-enabled-clients' and `lsp-disabled-clients')."
                     major-mode major-mode major-mode))))))
 
+(defun lsp--buffer-visible-p ()
+  "Return non nil if current buffer is visible."
+  (or (buffer-modified-p) (get-buffer-window nil t)))
```
ericdallo, line 8311: Is this check if buffer visible correct?

ericdallo:
Ping @yyoncho

yyoncho (approved):

PR #3087, opened 2021-09-08
Dispatch each lsp--parser-on-message in next message loop

### LINKED ISSUE #3079: Messages won't be processed if there is lsp-request-while-no-input

- [ ] You may also try reproduce the issue using clean environment using the following command
```bash
```

### Bug description

See https://github.com/emacs-lsp/lsp-mode/pull/3075#pullrequestreview-746561695 

### Steps to reproduce

-

### Expected behavior

-

-

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

kiennq:
Dispatch `lsp--parser-on-message` in a separated message loop so that the callback function can throw without affects others.
Fix #3079

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1890,7 +1890,7 @@ regex in IGNORED-FILES."
   "Helper macro for invoking BODY in WORKSPACE context."
   (declare (debug (form body))
            (indent 1))
-  `(let ((lsp--cur-workspace ,workspace)) ,@body))
+  `(when ,workspace (let ((lsp--cur-workspace ,workspace)) ,@body)))
 
 (defmacro with-lsp-workspaces (workspaces &rest body)
   "Helper macro for invoking BODY against multiple WORKSPACES."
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -6368,7 +6376,8 @@ deserialization.")
                              (concat leftovers input)
                              err)))))))
         (mapc (lambda (msg)
```
yyoncho, line 6378: It seems to me that this mapc is no longer needed? The dispatch will effectively do the loop.

yyoncho: So, it will work like this too, due to the guard. IMHO it might be more readable if you move the dispatch out of the loop.

yyoncho (approved):

ericdallo:
Ping @kiennq, are this good? the CI seems to be failing

kiennq:
> Ping @kiennq, is this good? the CI seems to be failing

Yeah, the CI is failing due to this change to run into another message loop. This is purely test issue though.

yyoncho:
BTW If you want to play with this, we may suppress dispatching any of the messages if there is a completion request in progress thus giving priority to the completion. E. g. if we receive diagnostics it sometimes slows down the completion and the completion IMHO is with higher priority.

ericdallo:
That sounds like a nice improvement

PR #3091, opened 2021-09-10
Add python .venv folder to lsp-file-watch-ignored-directories

hemanta212:
The .venv folder stores project dependencies for a python project. 

Although there is no standard folder name in python for these currently,  .venv is the most common naming convention used when keeping the python virtual environment locally in the project folder.


ericdallo (approved):
Thank you!

PR #3100, opened 2021-09-14
lsp-csharp.el: remove `--version xxx` when installing csharp-ls using dotnet tool

razzmatazz:
A small future-proofing change so we don't need to update lsp-csharp.el every time csharp-ls gets an update released:

* New versions of csharp-ls no longer have -alpha suffix we don't need to
bind to a particular version anymore;

* Shortened naming style for lsp-csharp.el functions related to csharp-ls.

### DIFF
--- a/clients/lsp-csharp.el
+++ b/clients/lsp-csharp.el
@@ -413,13 +413,10 @@ is returned so lsp-mode can display this file."
   "Install/update csharp-ls language server using `dotnet tool'.
 
 Will invoke CALLBACK or ERROR-CALLBACK based on result. Will update if UPDATE? is t"
-  (let ((exit-code (call-process "dotnet" nil nil nil
-                                 "tool"
-                                 (if update? "update" "install")
-                                 "-g" "csharp-ls")))
-    (if (zerop exit-code)
-        (funcall callback)
-      (funcall error-callback))))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

clients/lsp-csharp.el
```diff
@@ -409,24 +409,26 @@ is returned so lsp-mode can display this file."
...
+  "Install/update csharp-ls language server using `dotnet tool'.
+
+Will invoke CALLBACK or ERROR-CALLBACK based on result. Will update if UPDATE? is t"
+  (let ((exit-code (call-process "dotnet" nil nil nil
+                                 "tool"
+                                 (if update? "update" "install")
+                                 "-g" "csharp-ls")))
+    (if (zerop exit-code)
+        (funcall callback)
+      (funcall error-callback))))
```
yyoncho, line 422: here you should pass a message to the error callback.  FTR we tend to use `lsp-async-start-process` to execute install processes which will pop up compilation buffer which will help the users figure out what the issue is.

razzmatazz: ok, I changed the code to use `lsp-async-start-process` -- seems to work

yyoncho:
Thank you!

Issue #2000, opened 2020-08-05
lsp-fsharp prints "No ident at this location" error


fehmud:
Hello everybody and thank you for this package.

I am experiencing a problem with `lsp-fsharp` that should be easy to reproduce, therefore I am skipping the full bug report for now.

The problem is that when point is on an operator - like `|` - `lsp-mode` prints the message:

    LSP :: Error from the Language Server: No ident at this location (Internal Error)

yyoncho:
Do you have lsp-ui enabled? I think that this "error" is hidden in lsp-mode.

fehmud:
@yyoncho 

Thanks for your suggestion. I have disabled `lsp-ui`, but the warning persists.

As a workaround until I will investigate this better, I have redefined `lsp--create-default-error-handler` to generate a handler that bypasses that warning.

yyoncho:
It will be great if you can provide the location of the code that triggers the error(or someone using FSharp). Note that we have 40+ languages and even configuring each one of them + sample project is time-consuming if you are not using that language.

fehmud:
I have put a conditional breakpoint on `log--warn`, and the warning seems coming from a "textDocument/documentHighlight" response (the earliest visible function call on the stack is `lsp--parser-on-message`).

When the token under point is not an identier, then `lsp-fsharp` may:
1. warn of "No ident at this location" (e.g. on `|`, `10`);
2. warn of "No symbol information found" (e.g. on `=`, `_`);
3. like #2 and then echo the symbol in the minibuffer (e.g. on `let`, `->`).

s-kostyaev:
Is there any examples for that. I can't understand how to use it.

yyoncho:
@s-kostyaev no, I mean that we can change lsp-mode to allow that pattern. E. g. `(defcustom lsp-default-error-handler ...` which then can be passed at https://github.com/yyoncho/lsp-mode/blob/posframe-better/lsp-mode.el#L3258

yyoncho:
or it can be `lsp-default-create-error-handler-fn` to allow passing the method name which can be convenient for filtering the errors.

s-kostyaev:
I think about variant where we can pass it into `make-lsp-client` for default error handler per lsp-client.

s-kostyaev:
Or another option: add something like `:ignore-errors` into `make-lsp-client` and filter it inside `lsp--create-default-error-handler`.

yyoncho:
[...]

PR #3110, opened 2021-09-22
add custom default error handler function

s-kostyaev:
See #2000

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3164,12 +3164,22 @@ ID is the request id. "
                      results
                    (lsp--merge-results (-map #'cl-rest results) method)))))))
 
+(defcustom lsp-default-create-error-handler-fn nil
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3164,12 +3164,20 @@ ID is the request id. "
                      results
                    (lsp--merge-results (-map #'cl-rest results) method)))))))
 
+(defcustom lsp-default-create-error-handler-fn nil
+  "Default error handler customization.
+Handler should give METHOD as argument and return function of one argument ERROR."
+  :type 'function
+  :group 'lsp-mode)
```
yyoncho, line 3171: can you add `:package-version '(lsp-mode . "8.0.1")`

s-kostyaev: sure

PR #3122, opened 2021-09-27
Lsp mode progress percentage precision fix.

vitek:
Only show integer part of progress percentage.


vitek:
Without modeline looks like this when indexing:

image

yyoncho (approved):

yyoncho:

PR #3126, opened 2021-09-29
fix: do not start eslint lsp in json file

braineo:
json-mode is a derived mode of js-mode but eslint cannot work on json file


yyoncho (approved):

yyoncho:
Thank you!

Merging as it is although the proper way to fix that at this point is to switch to lsp-activate-on instead of ad-hoc activation function.

PR #3129, opened 2021-09-29
Update eslint language server to use lsp-server-binaries

yyoncho:
- marketplace fails us often due to DOS prevention


ericdallo (approved):
Same issue I had on lsp-dart hehe

yyoncho:
> Same issue I had on lsp-dart hehe

Yes, lets move marketplace items in that repo.

PR #3131, opened 2021-10-02
lsp-csharp.el: add support for -s parameter 

razzmatazz:
..to load specified solution for csharp-ls

@ericdallo could you take a peek if I didn't break your recent changes..


ericdallo (approved):
LGTM

razzmatazz:
Should I do something about the tests? Restart them? I think those failures are unrelated

jcs090218:
I think tests are unrelated, it's fixed in #3134.

jcs090218 (approved):

Issue #3140, opened 2021-10-05
provide tsserver preferences to js-ts


orzechowskid:
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)?

not that I could find

Describe your feature here.

the [typescript-language-server initialization options object](https://github.com/typescript-language-server/typescript-language-server#initializationoptions) contains a field named `preferences` which allows clients to configure the underlying `tsserver` process.  `lsp-mode` does not seem to be providing this field in its `initialize` requests.

(for context, I'm working in a Javascript project where all file imports must have extensions specified - e.g. `import Foo from './Foo.js'`)

I have a small PR for this enhancement which is nearly complete but the contributor guidelines recommended opening an issue here first in case anyone else is working in the same area.

yyoncho:
> I have a small PR for this enhancement which is nearly complete but the contributor guidelines recommended opening an issue here first in case anyone else is working in the same area.

Sounds good, feel free to provide the PR.

PR #3141, opened 2021-10-06
Feature/ts-ls init opts

orzechowskid:
the typescript-language-server initialization options object contains a field named preferences which allows clients to configure the underlying tsserver process. lsp-mode does not seem to be providing this field in its initialize requests.  this PR adds a `defcustom` to allow users to define the contents of this field, and updates the `:initialization-options` passed to `make-lsp-client` to provide it to ts-ls.

addresses #3140 .

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -89,7 +89,8 @@
   :type 'string)
 
 (defcustom lsp-clients-typescript-init-opts nil
-  "Configuration options provided to tsserver."
+  "Configuration options provided to tsserver.
+See the UserPreferences interface at https://github.com/microsoft/TypeScript/blob/main/lib/protocol.d.ts for the list of options available in the latest version of TypeScript."
   :group 'lsp-typescript
   :type 'plist)
 


clients/lsp-javascript.el
```diff
@@ -88,6 +88,11 @@
   :group 'lsp-typescript
   :type 'string)
 
+(defcustom lsp-clients-typescript-init-opts nil
+  "Configuration options provided to tsserver."
```
orzechowskid, line 92: I'm not happy with this docstring but couldn't find better words to write.

the keys of this plist are symbols corresponding to the supported fields in [the `preferences` field of js-ts initializationOptions](https://github.com/typescript-language-server/typescript-language-server#initializationoptions), and the values of this plist are primitives.

an example value for `lsp-clients-typescript-init-opts` would be
```lisp
'(:importModuleSpecifierEnding "js" :generateReturnInDocTemplate t)
```

orzechowskid: I found the definition of this preferences interface in the TypeScript source itself and linked to that instead; it seemed like that would be more up-to-date than someone else's documentation.

orzechowskid:
there are [two tests failing in CI](https://github.com/emacs-lsp/lsp-mode/runs/3817927223?check_suite_focus=true) on this branch; both use emacs 27 on OSX machines and seem to be related to missing packages and not related to this PR.  I can look into it if I'm mistaken though.

yyoncho (approved):

Issue #3159, opened 2021-10-19
lsp-install-server not recognizing available servers


BuddhiLW:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

I go inside a .js file, then *messages* gives:
"LSP :: There are no language servers supporting current mode `rjsx-mode' registered with `lsp-mode'.
This issue might be caused by:
1. The language you are trying to use does not have built-in support in `lsp-mode'. You must install the required support manually. Examples of this are `lsp-java' or `lsp-metals'.
2. The language server that you expect to run is not configured to run for major mode `rjsx-mode'. You may check that by checking the `:major-modes' that are passed to `lsp-register-client'.
3. `lsp-mode' doesn't have any integration for the language behind `rjsx-mode'. Refer to https://emacs-lsp.github.io/lsp-mode/page/languages and https://langserver.org/.
4. You are over `tramp'. In this case follow https://emacs-lsp.github.io/lsp-mode/page/remote/.
5. You have disabled the `lsp-mode' clients for that file. (Check `lsp-enabled-clients' and `lsp-disabled-clients')."

Then, I press M-x lsp-install-server RET which returns in *messages*
"user-error: There are no servers with the automatic installation"

*I have npm server installed*
`sudo npm install -g typescript-language-serve`

### Steps to reproduce

1. create a file "index.js"
2. put the text `import React from 'react'`.
3. M-x lsp-install-server RET

### Expected behavior

To show the available servers, including the typescript/javascript one.

### Which Language Server did you use?

lsp-mode

### OS

Linux

yyoncho:
> > M-x lsp-install-server RET
> 
> In emacs selection dialogs you should use TAB or type the entry name.

Ignore that.

Please confirm that the issue reproduces with lsp-start-plain.el. To me it sounds more like an issue on your end.

PR #3160, opened 2021-10-19
Improve lsp-install-server and add lsp-update-server

ericdallo:
Related / fix? https://github.com/emacs-lsp/lsp-mode/issues/3159

The idea is to show already installed clients as option and if user choose it, we update it.
This should avoid users complaining about server not found on the servers list as well

image


yyoncho:
Can we keep the original signature? lsp-install-server is used in other projects as well(e. g. lsp-gitpod).

PR #3170, opened 2021-10-27
Improve semantic token modifiers

ericdallo:
This PR:

- enable by default `lsp-semantic-tokens-apply-modifiers` as it doesn't make sense I think in keep it as nil.
- Add missing modifiers faces for most important/common modifiers

Below you can find some like `defaultLibrary` and `definition` working with clojure-lsp master:

image

c/c @sebastiansturm


sebastiansturm (approved):
LGTM, thanks!

PR #3183, opened 2021-11-02
performance.md: Instruct users on generating reports

### LINKED ISSUE #3178: Documentation: Missing step in Performance
I would have proposed a PR but I don't see where the content of the doc site is maintained.

In https://emacs-lsp.github.io/lsp-mode/page/performance/#reporting-performance-problems the steps are listed but the last two are:

* `M-x profiler-stop`
* In the profiler report expand all nodes by doing `C-u TAB`.

are missing the step to run `M-x profiler-report` in between these two.  For people not already familiar with the profiler and how to use it, the above is inscrutable :)

madscientist:
Fixes #3178


ericdallo (approved):
Thank you!

Issue #2571, opened 2021-01-30
Feature Request: re-arrange mode line order


alanz:
#2567 brought in a much-improved progress reporting widget.

But it highlighted for me a niggling issue. All the interesting stuff reported in the  mode line by `lsp-mode` comes right at the end of the mode line, where for me most of the time it is invisible. (I use a fairly "vanilla" setup, tweaked slightly with the excellent `blackout` package to remove or shorten displayed items)

![image](https://user-images.githubusercontent.com/409607/106355348-2f960a00-62ef-11eb-8c26-95558ebbacca.png)

The *legacy* progress reporting modified the mode lighter, which is far to the left, so visible even when you have many windows open side by side. With the progress added to `global-mode-string`, it has moved to the end.

Perhaps we should consider moving the `lsp-mode` information to the left, between vc-info and mode info. 

Using lsp-mode at: abac0d7450d91a8f386f8be82e36f207ea36f77a

ericdallo:
Do you see any issues with that @yyoncho, also we could take the opportunity to fix that spacing issue in modeline (add space to the left instead of right, something like that)

yyoncho:
I think that we should group all lsp-mode related modeline stuff in a separate variable(e. g. lsp-modeline-string) instead of using global-mode-string and then place it on the modeline probably before the major mode list similar to what evil-mode does(I guess there are other packages doing that). When we do so, we are much more in control to do the proper spacing because only lsp-mode will use that variable.

yyoncho:
For reference, here it is mode-line-format on my side 

```
("%e" mode-line-front-space mode-line-mule-info mode-line-client mode-line-modified mode-line-remote mode-line-frame-identification mode-line-buffer-identification "   " mode-line-position evil-mode-line-tag
 (vc-mode vc-mode)
 "  " mode-line-modes mode-line-misc-info mode-line-end-spaces)
```

`global-mode-string` goes in `mode-line-misc-info`

ericdallo:
Yeah this looks a good idea, doom-modeline does that similar

yyoncho:
> Yeah this looks a good idea, doom-modeline does that similar

doom-modeline will have to be fixed separately I guess.

yyoncho:
FWIW until the issue is fixed one can use `diminish` package to clean the major/minor modes from the modeline.

PR #3188, opened 2021-11-05
[modeline] Add segments to future drop global-mode-string

ericdallo:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/2571

This PR adds 2 new variables:
- `lsp-modeline-use-global-mode-string` with `t` as default which behaves the same as ATM, adding modeling strings to the `global-mode-string`.
- `lsp-modeline-segments` used in case `lsp-modeline-use-global-mode-string` is `nil`, indicating the modeling segments features and its order, adding those segments in the `mode-line-format` before major modes following the same `evil-mode` does.

This should standardize the way we add things to modeling and make it easy to improve fixing spacing and other things.

To disable `lsp-modeline-use-global-mode-string` by default, we should at least update major modelines like `doom-emacs modeline`, `doom-modeline`, spaceemacs (?) to use the `lsp-modeline--string` in their modeline.

### DIFF
--- a/lsp-modeline.el
+++ b/lsp-modeline.el
@@ -71,8 +71,76 @@
                  (const :tag "All Projects" :global))
   :package-version '(lsp-mode . "6.3"))
 
+(defcustom lsp-modeline-use-global-mode-string t
+  "Whether add LSP features to global-mode-string instead of mode-line-format."
+  :group 'lsp-modeline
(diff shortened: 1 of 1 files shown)


ericdallo:
c/c @yyoncho @alanz @kiennq

Issue #3155, opened 2021-10-16
[semantic-tokens] Refresh tokens after apply workspaceEdit


ericdallo:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

We should refresh the semantic tokens of all affecter buffers by a workspaceEdit change sent from server.

### Steps to reproduce

call LSP rename on some symbol that is used on multiple already opened buffers, after renaming the symbol on other buffers will have outdated semantic tokens.

### Expected behavior

refresh the tokens after applying the workspaceEdit

### Which Language Server did you use?

clojure-lsp and dart

### OS

Linux

### Anything else?

c/c @sebastiansturm

ericdallo:
It seems we are sending the request semantic token before we send didChange, causing these outdated tokens c/c @sebastiansturm

sebastiansturm:
I pushed a change to my fork that refreshes tokens on `lsp-after-change`, though I'm still not sure I'm too fond of this change, will try to make up my mind tomorrow. Haven't tested this extensively but at least the clojure-lsp example seems to refontify correctly

ericdallo:
yes, I used this during the week and it works perfectly unless I jack in cider which I still have no idea how to fix that :/

Issue #3154, opened 2021-10-16
Semantic tokens not entire applied on buffer


ericdallo:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

Sometimes when opening a buffer the semantic tokens are applied only for the current visible region (probably range semantic tokens) but if scrolling the buffer up or down, I don't see tokens for those other regions unless I `M-x` `lsp` again, even so other regions would not contain the tokens.

print:
![image](https://user-images.githubusercontent.com/7820865/137595904-4f57247c-b1ea-4e1f-a7af-157658aedb7b.png)
You can see how just part of the buffer has the green tokens

I tested on vscode with clojure-lsp and it works fine, so it seems a issue on lsp-mode.

### Steps to reproduce

You can clone https://github.com/clojure-lsp/clojure-lsp/ and open `server.clj` buffer, it's possible to see in other buffers as well.

### Expected behavior

All tokens correctly applied for the entire buffer.

### Which Language Server did you use?

[clojure-lsp](https://github.com/clojure-lsp/clojure-lsp/)

### OS

Linux

### Anything else?

@sebastiansturm could you help understand why that happens? I was thinking that was some issue on server but after implementing semantic tokens support on Calva(VScode extension to code clojure) [here](https://github.com/BetterThanTomorrow/calva/pull/1336), everything worked perfectly there.
I think this is the only issue I see with semantic tokens right now, but it's quite annoying as it happens a lot :/

ericdallo:
Thank you for taking a look @sebastiansturm!

ericdallo:
I'm also having issues with `cider` package that is the REPL package that almost every clojure user uses it, it seems cider apply a lot of font lock and I can't disable that accurately, but I think this is other issue :)

ericdallo:
[...]

Issue #2834, opened 2021-05-05
Tree sitter compatibility w/ semantic tokens


theol0403:
I use semantic highlighting to improve the coloring of my c++ files. I was looking for better syntax hl for non-semantic identifiers and found tree-sitter.

However, when tree sitter is enabled, all semantic highlights disappear. Even for identifiers that don't have a tree sitter face assigned to them, the semhl face is still gone. It would be ideal if semhl would override tree-sitter.

Default font-lock:

![image](https://user-images.githubusercontent.com/16546293/117098924-9f6bc480-ad24-11eb-8ec1-b2c766ef6b2a.png)

Semhl: 

![image](https://user-images.githubusercontent.com/16546293/117098981-cd510900-ad24-11eb-9172-29b7cf17b839.png)

Semhl + tree-sitter:
![image](https://user-images.githubusercontent.com/16546293/117098994-dc37bb80-ad24-11eb-9d1f-1899fc680781.png)

As you can see, the orange parameter semhl is removed. Running `describe-char` on the most-bottom "spline" identifier in the screenshot shows that there is no tree-sitter face being assinged to it, yet the semhl face is removed.

Semhl:
![image](https://user-images.githubusercontent.com/16546293/117099242-857eb180-ad25-11eb-82f4-82711d2781af.png)

Semhl + tree-sitter:
![image](https://user-images.githubusercontent.com/16546293/117099267-94656400-ad25-11eb-8f9f-c155c7e54d94.png)

**Which Language Server did you use**
Clangd 12

**OS**
Fedora 35 w/ doom-emacs w/ ptk-nativecomp from git tree.

Thanks!

yyoncho:
cc @sebastiansturm @ubolonton

sebastiansturm:
could you please try that with PR https://github.com/emacs-lsp/lsp-mode/pull/2790? That should improve semhl compatibility with underlying font-lock mechanisms. Also, I guess it might matter in what order tree-sitter-hl-mode and lsp are enabled; during my personal tests, I typically activated tree-sitter first, lsp thereafter

theol0403:
I pinned my lsp-mode package in doom to `0a298c4`, rebuilt (and it did properly checkout the commit), and I did not notice any difference in behavior.

The order might matter, but I'm not sure how I can control that within doom. Ideally, it should not matter, of course.

ericdallo:
[...]

PR #3189, opened 2021-11-06
Fix 3 different cases of stale semantic highlights

sebastiansturm:
this PR intends to solve issues #3155, #3154 and #2834.

#2834 was caused by late tree-sitter activation overriding lsp's semantic font-lock handlers; we now have lsp-semantic-tokens-mode watch out for late tree-sitter-hl activation and, if that happens, warn the user about the configuration issue, disable both modes and reenable them in the correct order. The fix has been verified, though I'd appreciate @theol0403 having another look at this PR, as the version I had previously proposed was slightly different (didn't disable in proper LIFO order).

#3154 was caused by aggressive font-lock region widening requested by some underlying font-lock mechanism (clojure-mode's font-lock, in the reported case), which would cause font-lock to clear token highlights within regions for which no new tokens were available (if ranged requests were being used). To fix that, we now enlarge ranged request regions by several font-lock chunks, and refontify with full responses if the resulting regions still remain too small to cover the underlying fontifier's fontification region.

#3155 was caused by workspace edits not resulting in semantic token flushes; to alleviate this (as workspace edits may affect a large number of buffers simultaneously), pending full-token requests are now maintained as a global queue and full token sets are requested (on an idle timer) within `lsp--after-change`. This is the commit most likely to cause problems, as it changes the way full token requests are handled, and the tighter coupling between lsp-mode proper and lsp-semantic-tokens isn't exactly pretty. Suggestions on how to better handle this are very welcome

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -23,6 +23,7 @@
 ;;; Code:
 
 (require 'lsp-mode)
+(require 'dash)
 
 (defgroup lsp-semantic-tokens nil
   "LSP support for semantic-tokens."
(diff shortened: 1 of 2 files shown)


lsp-semantic-tokens.el
```diff
@@ -253,17 +254,72 @@ Faces to use for semantic token modifiers if
         (tokenTypes . ,(apply 'vector (mapcar #'car lsp-semantic-token-faces)))
         (formats . ["relative"])))))
 
+(defvar lsp--semantic-tokens-pending-full-token-requests '()
+  "Buffers which should have their semantic tokens refreshed on idle.
+
+This is an alist of the form ((buffer_i . fontify_immediately_i) ...); entries with
+fontify_immediately set to t will immediately refontify once their token
+request is answered.")
```
ericdallo, line 262: To fix CI
```suggestion
This is an alist of the form ((buffer_i . fontify_immediately_i) ...); entries 
with fontify_immediately set to t will immediately refontify once their 
token request is answered.")
```

sebastiansturm: thanks! Pushed those changes, and addressed a few similar complaints raised by the byte compiler

ericdallo:
@sebastiansturm seems good to me, could you please rebase and fix the conflicts?

ericdallo:
Thank you!

PR #3195, opened 2021-11-09
Fix phpactor URL & remove installation

elken:
Existing installation instructions aren't recommended due to collisions with other libraries.


yyoncho (approved):

yyoncho:
Thank you!

PR #3209, opened 2021-11-14
Add  to completion items when insert-text

ericdallo:
This is a simple modification to enhance completion doing the same neovim does, showing a `~` on the end of the completion item if it will insert some kind of text rather than just the `label`, this is useful to user know that this completion item will do something other than just complete with the label.

Note: We can rely only on snippet kind since it's possible and IMO makes sense for example to return a completion item saying that a item is of type function but will insert some other kind of text, example:

__lsp-mode__
image

__neovim__
image

c/c @kiennq @yyoncho

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -143,10 +143,13 @@ This will help minimize popup flickering issue in `company-mode'."
 (cl-defun lsp-completion--make-item (item &key markers prefix)
   "Make completion item from lsp ITEM and with MARKERS and PREFIX."
   (-let (((&CompletionItem :label
+                           :insert-text?
                            :sort-text?
                            :_emacsStartPoint start-point)
           item))
-    (propertize label
+    (propertize (if insert-text?
+                    (concat label "~")
+                  label)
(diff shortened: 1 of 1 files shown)


yyoncho:
Can you have this behind a flag and turned off by default? I haven't seen anything like that in other ides and I don't think that neovim can be our UE guideline. We tend to follow the VSCode.

kiennq:
Can we use company-mode annotation for this instead?
That would also had better performance since we only adding the annotations for visible candidates instead of the whole list

ericdallo:
I didn't know about annotation, sounds good, I'll change it

kiennq:
Also, if we're using `:annotation-function` then we can do more accurate comparisons there, i.e,: compare the `label` to `insert-text` to make sure the display and the inserted is not the same, compare the `edit-text` as well if possible.

diaevd:
> I didn't know about annotation, sounds good, I'll change itI
I'm make similar changes with the company-transform function. Why change something at the lsp-completion level? For me this change will be breaking change

PR #3211, opened 2021-11-17
Added client for Python-LSP-Server

AnselmC:
I added a client for the python-lsp-server implementation that is currently being maintained.
Not sure whether I should have to touch any other file.

### DIFF
--- a/docs/lsp-clients.json
+++ b/docs/lsp-clients.json
@@ -482,6 +482,9 @@
     "server-name": "pylsp",
     "server-url": "https://github.com/python-lsp/python-lsp-server",
     "installation": "pip install 'python-lsp-server[all]'",
+    "installation-url": "https://github.com/AnselmC/lsp-py",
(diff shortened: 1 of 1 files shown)


yyoncho:
I think that we already have a client for this https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-pylsp.el

(the entry in client.json that you are editing is for that server)

AnselmC:
Ah I didn't see this. I guess it's missing from the documentation?
https://emacs-lsp.github.io/lsp-mode/page/languages/

yyoncho:
> Ah I didn't see this. I guess it's missing from the documentation? https://emacs-lsp.github.io/lsp-mode/page/languages/

@ericdallo can you check why the entry is not generated from client.json? (it is there).

ericdallo:
Yes, I can check later today

ericdallo:
There was a missing configuration on mkdocs.yml, I fixed on 9865d3151

PR #3230, opened 2021-11-25
Introduce a flag to disable automatic download

yyoncho:


ericdallo (approved):

lsp-mode.el
```diff
@@ -5007,6 +5007,12 @@ In addition, each can have property:
   :group 'lsp-mode
   :type 'boolean)
 
+(defcustom lsp-enable-server-download t
```
michaelpj, line 5010: Maybe put "auto" in the name somewhere? Since this doesn't affect the ability to use e.g. `lsp-install-server`

ericdallo: Make sense to me WDYT @yyoncho ?

jcs090218 (approved):

abbec:
Any ETA on this? :)

lsp-mode.el
yyoncho: I added the word suggest, because the "auto" is not actually auto.

PR #3234, opened 2021-11-28
Prioritize create edits before other edits

ericdallo:
I found this issue when implementing a new code action on clojure-lsp "Create namespace + function".
When the server returns on the same `apply workspace edit` request a edit to create a file and other edit to add code to it, we should:
- create the file first before other edits
- make it "visited" to avoid `File foo changed on disk.  Discard your edits? (y or n)` messages when user visit the buffer after the changes.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4002,12 +4002,17 @@ OPERATION is symbol representing the source of this text edit."
         (progn
           (lsp--check-document-changes-version document-changes)
           (->> document-changes
-               (seq-filter (-lambda ((&CreateFile :kind))
-                             (or (not kind) (equal kind "edit"))))
+               (seq-filter (-lambda ((&CreateFile :kind)) (equal kind "create")))
                (seq-do (lambda (change) (lsp--apply-text-document-edit change operation))))
           (->> document-changes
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4043,7 +4048,8 @@ interface TextDocumentEdit {
                 (mkdir (f-dirname file-name) t)
                 (f-touch file-name)
                 (when (lsp:create-file-options-overwrite? options?)
-                  (f-write-text "" nil file-name))))
+                  (f-write-text "" nil file-name))
+                (find-file-noselect file-name)))
```
yyoncho, line 4052: why this is needed?

ericdallo: For some reason, without this, when user access the buffer, emacs prompt with `File foo changed on disk. Discard your edits? (y or n)`, if user types `y`, the buffer becomes blank, if user types `n` then buffer becomes with the changes of those edits correctly.
I realized that adding this makes emacs load the buffer file after creating the file which makes sense

yyoncho (approved):

jcs090218 (approved):

PR #3235, opened 2021-11-28
Add org-mode to language id

jcs090218:
Not sure if this is required for org-mode to work correctly with `lsp-grammarly` and `lsp-ltex`? 😕 

#### Related issue

* https://github.com/emacs-languagetool/lsp-ltex/issues/10
* https://github.com/emacs-grammarly/lsp-grammarly/issues/23


yyoncho (approved):
This is fine as long as the servers expect that language ID.

jcs090218:
I see, thanks! I reckon `plaintext` is the correct language id for both `grammarly` and `ltex` language servers.

PR #3237, opened 2021-12-01
Make installation-buffer removable by pressing Q.

josteink:
Currently, after installation is complete, you kinda of expect to be able to dismiss the `*lsp-install*` buffer by pressing Q (like compilation-mode buffers, occur-buffers, etc).

But when you do that, you find that the buffer is open to all kinds of input and editing, which IMO doesn't make much sense.

This PR changes `*lsp-install*` buffers to have `view-mode` enabled, and thus be easily dismissable.


yyoncho (approved):

yyoncho:
Thank you!

PR #3238, opened 2021-12-01
Allow a minimum PHP version to be specified in Intelephense

elken:
Introduces an option to set the PHP version for Intelephense; as it doesn't use any system PHP.

### DIFF
--- a/clients/lsp-php.el
+++ b/clients/lsp-php.el
@@ -85,6 +85,12 @@
   :link '(url-link "https://github.com/bmewburn/vscode-intelephense")
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-intelephense-php-version "8.0.1"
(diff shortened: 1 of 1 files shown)


clients/lsp-php.el
```diff
@@ -85,6 +85,12 @@
   :link '(url-link "https://github.com/bmewburn/vscode-intelephense")
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-intelephense-php-version "8.0"
+  "Minimum version of PHP to refer to. Affects code actions, diagnostic & completions."
+  :type 'string
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "6.1"))
```
yyoncho, line 92: please change it to 8.0.1

elken: Thanks, I wasn't sure what to set it to :)

yyoncho (approved):

PR #3248, opened 2021-12-09
[lsp-terraform] add hashicorp lsp server

thanhvg:
https://github.com/hashicorp/terraform-ls
tested on Linux

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -26,6 +26,9 @@
 
 (require 'lsp-mode)
 
+
+;; terraform-lsp
+
 (defgroup lsp-terraform nil
   "LSP support for Terraform, using terraform-lsp"
   :group 'lsp-mode
@@ -48,6 +51,7 @@
   :type 'boolean
   :package-version `(lsp-mode . "6.2"))
 
+
 (defun lsp-terraform--make-launch-cmd ()
   (-let [base (if (stringp lsp-terraform-server)
                   `(,lsp-terraform-server)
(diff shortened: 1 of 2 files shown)


clients/lsp-terraform.el
```diff
@@ -62,6 +65,30 @@
...
+  "Path to the `terraform-ls' binary."
+  :group 'lsp-terraform-ls
+  :risky t
+  :type '(choice
+          (file :tag "File")
+          (repeat string))
+  :package-version `(lsp-mode . "8.0.1"))
+
+(lsp-register-client
+ (make-lsp-client :new-connection (lsp-stdio-connection `(,lsp-terraform-ls-server "serve"))
```
yyoncho, line 87: we tend to wrap this in lambda to allow changing the property after lsp-mode is loaded. We do so for args.

clients/lsp-terraform.el
```diff
@@ -62,6 +65,30 @@
...
+                  :major-modes '(terraform-mode)
+                  :priority 1
```
yyoncho, line 89: do you really want to have that server with a bigger priority?

yyoncho (commented):
Can you update lsp-clients.json with the new client?

clients/lsp-terraform.el
thanhvg: I have no preference, I chose `1` because `-1` is already taken. I can change to `-2` if it makes any sense. In my experience both servers are so so. But the Hashicorp one's repo seems to be more active recently.

yyoncho: It is really up to you to pick the server we recommend to the users.

yyoncho (approved):

yyoncho:
Thank you!

PR #3250, opened 2021-12-10
[clojure] Add support for test tree feature

ericdallo:
This adds support for a new custom protocol on clojure-lsp that I'm implementing with the experimental capability if client supports it

image


clients/lsp-clojure.el
```diff
@@ -327,8 +412,10 @@ and the third the column."
   :uri-handlers (lsp-ht ("jar" #'lsp-clojure--file-in-jar))
   :action-handlers (lsp-ht ("resolve-macro-as" #'lsp-clojure--resolve-macro-as)
                            ("code-lens-references" #'lsp-clojure--show-references))
+  :notification-handlers (lsp-ht ("clojure/textDocument/testTree" #'lsp-clojure--handle-test-tree))
   :initialization-options '(:dependency-scheme "jar"
                             :show-docs-arity-on-same-line? t)
+  :custom-capabilities '((experimental . ((testTree . t))))
```
yyoncho, line 418: maybe here we should enable if lsp-treemacs is present.

yyoncho (approved):

PR #3275, opened 2021-12-20
Move to new eldoc

yyoncho:
- continue #3082

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4,7 +4,7 @@
 
 ;; Author: Vibhav Pant, Fangrui Song, Ivan Yonchovski
 ;; Keywords: languages
-;; Package-Requires: ((emacs "26.1") (dash "2.18.0") (f "0.20.0") (ht "2.3") (spinner "1.7.3") (markdown-mode "2.3") (lv "0") (eldoc "1.11"))
+;; Package-Requires: ((emacs "26.3") (dash "2.18.0") (f "0.20.0") (ht "2.3") (spinner "1.7.3") (markdown-mode "2.3") (lv "0") (eldoc "1.11"))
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
I don't see any problems with that.

We need to remove the check for these removed version on github settings after merge

lsp-mode.el
```diff
@@ -4851,10 +4837,33 @@ If INCLUDE-DECLARATION is non-nil, request the server to include declarations."
...
    (lsp-notify "$/cancelRequest" `(:id ,id))))
 
-(defun lsp-eldoc-function ()
-  "`lsp-mode' eldoc function."
-  (run-hooks 'lsp-eldoc-hook)
-  eldoc-last-message)
+(defvar-local lsp--hover-saved-bounds nil)
+
+(defun lsp-eldoc-function (cb &rest _ignored)
+  "`lsp-mode' eldoc function to display hover info (based on `textDocument/signatureHelp')."
```
michaelpj, line 4843:
```suggestion
  "`lsp-mode' eldoc function to display hover info (based on `textDocument/hover')."
```

jcs090218 (approved):
IMO, we can drop the entire Emacs 26.x. Esp. without the native json support; it's quite slow on windows machines.

lsp-mode.el
yyoncho: Applied, thanks.

yyoncho:
> IMO, we can drop the entire Emacs 26.x. Esp. without the native json support; it's quite slow on windows machines.

I don't think that we do anything special to support emacs 26 (disregarding the native json stuff, but you can still compile emacs 27 without native json). Let's drop emacs 26.3 when it turns into a burden.

PR #3283, opened 2021-12-26
replace unified-language-server with remark-language-server

remcohaszing:
### What is all this?

- [`unified`](https://github.com/unifiedjs/unified) is a plugin based ecosystem for processing files on an AST level.
- [`unified-engine`](https://github.com/unifiedjs/unified-engine) is a wrapper around unified which makes it easy to create tooling around `unified`. For example it deals with loading plugins, configuration files and ignore files.
- [`remark`](https://github.com/remarkjs/remark) is an implementation of `unified` which deals with markdown.

### What changed?

- [`unified-language-server`](https://github.com/unifiedjs/unified-language-server) **was** a language server which had its own logic for dealing with loading configuration files and plugins
- In version 1.0.0, `unified-language-server` has been changed from a this language server to a library for creating language servers based on `unified-engine`.
- [`remark-language-server`](https://github.com/remarkjs/remark-language-server) is a new `remark` based implementation of `unified-language-server` which follows the same logic as [`remark-cli`](https://github.com/remarkjs/remark/tree/main/packages/remark-cli).

---

I reworked how this language server works, but I’m not familiar with Emacs or Lisp. So sorry if I made silly mistakes.


ericdallo:
What happens with people using the unified-language-server? it will just stop working right? maybe we should support both with a higher priority for the remark one?

remcohaszing:
I wouldn’t recommend to keep supporting the old `unified-language-server`, but yes, you could. In that case it explicitly needs to depend on version 0.3.0.

yyoncho:
> I wouldn’t recommend to keep supporting the old `unified-language-server`, but yes, you could.

We tend to avoid breaking users configs unless it is unavoidable. We may put a warning to migrate to the other server in :initialization-fn.

remcohaszing:
Should I just restore `lsp-markdown.el` and change the priority in `lsp-remark.el` to `0` then so it takes precedence over `lsp-markdown.el`?

I’m interested in integrating `remark-language-server` in various lsp clients, but not in fixing backwards compatibility for `unified-language-server@0.3.0`, because I simply don’t have the knowledge to do so.

yyoncho:
> Should I just restore `lsp-markdown.el` and change the priority in `lsp-remark.el` to `0` then so it takes precedence over `lsp-markdown.el`?

Yes.

yyoncho:
(and you can mark the unified-ls as deprecated in clients.json similar to what we do for "pyls")

remcohaszing:
> (and you can mark the unified-ls as deprecated in clients.json similar to what we do for "pyls")

Done!

ericdallo (approved):

jcs090218 (approved):

ericdallo:
Thank you for your contribution!

ericdallo:
@remcohaszing could you add an entry on changelog please?

PR #3291, opened 2022-01-01
Add automatic kotlin-language-server installation

### LINKED ISSUE #3290: Add automatic installation for kotlin-language-server
lsp-mode could [download a release from the repository](https://github.com/fwcd/kotlin-language-server/releases/tag/1.2.0). The binary can be found in a subdirectory.

shivjm:
Since I’m entirely new to the codebase, I’ve modelled this PR after (if not outright copied it from) parts of lsp-fsac and lsp-clangd, to try to reuse as much as possible of the lsp-mode machinery. It seems to work on my machine. The two questions I have are:

1. Is it okay to use `lsp-clients-kotlin-server-executable`? The code relies on the evaluation of its initial value, so it may not work correctly for someone who’s upgrading.
2. ~What is the right way to fetch the latest GitHub version? At present, I’m doing it when the package is loaded (or rather when the `defcustom` form is evaluated) so that I can use it in the `lsp-dependency` call, which is obviously not a good experience. Ideally, I would only fetch it when the server is about to be downloaded.~

Closes #3290

### DIFF
--- a/clients/lsp-kotlin.el
+++ b/clients/lsp-kotlin.el
@@ -36,7 +36,10 @@
   'lsp-clients-kotlin-server-executable
   "lsp-mode 6.4")
 
-(defcustom lsp-clients-kotlin-server-executable "kotlin-language-server"
+(defcustom lsp-clients-kotlin-server-executable
+  (if (eq system-type 'windows-nt)
+      "kotlin-language-server.bat"
+    "kotlin-language-server")
   "The kotlin-language-server executable to use.
 Leave as just the executable name to use the default behavior of finding the
 executable with `exec-path'."
@@ -94,6 +97,13 @@ to Kotlin."
   :group 'lsp-kotlin
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-kotlin-server-download-url
+  "https://github.com/fwcd/kotlin-language-server/releases/download/latest/server.zip"
+  "The URL for the language server download."
+  :type 'string
+  :group 'lsp-kotlin
+  :package-version '(lsp-mode . "8.0.1"))
+
 (lsp-register-custom-settings
  '(("kotlin.externalSources.autoConvertToKotlin" lsp-kotlin-external-sources-auto-convert-to-kotlin t)
    ("kotlin.externalSources.useKlsScheme" lsp-kotlin-external-sources-use-kls-scheme t)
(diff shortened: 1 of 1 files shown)


clients/lsp-kotlin.el
```diff
@@ -36,7 +36,12 @@
   'lsp-clients-kotlin-server-executable
   "lsp-mode 6.4")
 
-(defcustom lsp-clients-kotlin-server-executable "kotlin-language-server"
+;; TODO is it okay to reuse this?
+(defcustom lsp-clients-kotlin-server-executable
+  (f-join lsp-server-install-dir
+          "kotlin" "server" "bin" (if (eq system-type 'windows-nt)
+                                      "kotlin-language-server.bat"
+              "kotlin-language-server"))
```
ericdallo, line 44: Indentation seems odd here

clients/lsp-kotlin.el
```diff
@@ -94,6 +99,12 @@ to Kotlin."
   :group 'lsp-kotlin
   :package-version '(lsp-mode . "6.1"))
 
+(defcustom lsp-kotlin-server-download-url
+  "https://github.com/fwcd/kotlin-language-server/releases/download/latest/server.zip"
+  "The URL prefix for the language server download."
+  :type 'string
+  :group 'lsp-kotlin) ; TODO add `:package-version'
```
ericdallo, line 106: You can add a package-version with 8.0.1 version

ericdallo, line 106: You can add a package-version with 8.0.1 version

clients/lsp-kotlin.el
```diff
@@ -36,7 +36,12 @@
   'lsp-clients-kotlin-server-executable
   "lsp-mode 6.4")
 
-(defcustom lsp-clients-kotlin-server-executable "kotlin-language-server"
+;; TODO is it okay to reuse this?
```
ericdallo, line 39: Does changing this variable default value would cause any breaking changes? if a user has a custom value for that variable, with this pr would we break that? if so, it's not a good idea to re-use it, otherwise sounds ok to me

shivjm: If I’ve understood correctly, it shouldn’t, because the custom value will be passed to `lsp-dependency` as the `:binary-path` and it shouldn’t be downloaded again. But what about if I do this, to be safe?

1. Put the generated value I have here in a new variable.
2. Let `lsp-clients-kotlin-server-executable` remain unchanged except to add `.bat` to the default value on Windows. Anyone already setting the variable should be unaffected.
3. Change this form to use `(:system lsp-client-kotlin-server-executable)` if found, and the `:download :url` second.

This way, those who’ve set `lsp-clients-kotlin-server-executable` to a valid executable will see it work automatically, while everyone else (including those who’ve set it to an invalid path) will be prompted to download the language server.

ericdallo: Sounds good to me

ericdallo (approved):

shivjm:
Thanks for the review! Sorry, I think a couple of commits got mixed up. Fixing it now.

shivjm:
Looks like everything’s resolved from my end. Thank you for your patience.

ericdallo:
Thank you for the contribution!

PR #3297, opened 2022-01-04
Remove godoc.org from lsp-go-link-target

l2dy:
All requests to godoc.org redirect to pkg.go.dev, see https://github.com/golang/go/issues/43178.


yyoncho:

PR #3304, opened 2022-01-07
Add capability to customize lighter

jcs090218:
In GIF, I restart `lsp-mode` and set `lsp-lighter-delay` to `3.0`.

ezgif-2-c576ce8056

Everything should default to the same as the current settings.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2703,6 +2703,54 @@ and end-of-string meta-characters."
          (when choice
            (call-interactively action)))))))
 
+(defcustom lsp-lighter-delay nil
+  "Delay to shorten lighter in seconds; nil means never shorten lighter."
+  :type '(choice (const :tag "Never Hide" nil)
+                 (number :tag "Seconds"))
+  :group 'lsp-mode)
+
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -3864,7 +3907,8 @@ yet."
     (remove-hook 'xref-backend-functions #'lsp--xref-backend t)
     (remove-hook 'change-major-mode-hook #'lsp-disconnect t)
     (remove-hook 'after-set-visited-file-name-hook #'lsp--after-set-visited-file-name t)
-    (setq-local lsp-buffer-uri nil))))
+    (setq-local lsp-buffer-uri nil)))
+  (lsp--run-lighter-timer))
```
jcs090218, line 3911: Not sure if this is a appropriate place to add this line?

yyoncho:
`diminish` package is able to modify mode's lighter without the internal changes. Can we turn this PR into a recipe without changing the lsp-mode's source?

jcs090218:
I think this was worth to discuss hence I have the PR opened. I think it's good to customize it in my own config! Thanks! :)

PR #3310, opened 2022-01-10
Update versions of clangd to look for

berquist:
https://apt.llvm.org/ has both 13 and 14.

This also replaces the hard-coded list with a reversed `number-sequence`.

### DIFF
--- a/clients/lsp-clangd.el
+++ b/clients/lsp-clangd.el
@@ -234,9 +234,9 @@ This must be set only once after loading the clang client.")
                       (-map (lambda (version)
                               (concat "clangd" version))
                             ;; Prefer `clangd` without a version number appended.
-                            (append (list "") (-map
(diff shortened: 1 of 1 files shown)


clients/lsp-clangd.el
```diff
@@ -233,7 +233,10 @@ This must be set only once after loading the clang client.")
               (-first #'executable-find
                       (-map (lambda (version)
                               (concat "clangd" version))
-                            '("" "-12" "-11" "-10" "-9" "-8" "-7" "-6")))
+                            ;; Prefer `clangd` without a version number appended.
+                            (append (list "") (-map
```
yyoncho, line 237: nit: cl-list*

yyoncho (commented):
Looks good, one minor nit:

We should bump `lsp-clangd-version` as well.

yyoncho:
Thank you!

PR #3314, opened 2022-01-11
Allow disabling timeouts

umanwizard:


ericdallo:
@umanwizard do you have an example why you would like to disable the timeout?

umanwizard:
Because requests in large Rust projects (e.g. https://github.com/MaterializeInc/materialize , which I work on as my day job) can take a few minutes to return when rust-analyzer’s cache is still cold.

Since one can cancel long-running requests with C-g anyway, I see no reason to have a timeout.

If you choose not to land this, I can just set the timeout to 9999, but that seems less elegant.

ericdallo (approved):
Makes sense to me, thanks!

umanwizard:
Thanks @ericdallo ! Is any action needed from me in order to merge this?

jcs090218 (approved):

ericdallo:
@umanwizard I was waiting for more input from other maintainers, could you add an entry on CHANGELOG? then I think we can merge this

umanwizard:
Done!

ericdallo:
Thank you!

PR #3330, opened 2022-01-29
Fix: Local variables not safe

rvlo:
lsp-file-watch-ignored-directories and lsp-file-watch-ignored-files were
not being respected as safe, when opening the first LSP project.

Added autoload magic comments to the safe-local-variable definitions and
to the predicate function lsp--string-listp.


yyoncho (approved):

yyoncho:
Thank you!

PR #3332, opened 2022-02-01
Add lsp-remark to the list of required client packages

picsel2:
The new client was added with emacs-lsp#3283 but forgotten to actually
enable it.


yyoncho:
Thank you!

PR #3349, opened 2022-02-12
lsp-csharp.el: fix breakage with the latest emacs

razzmatazz:
See https://lists.gnu.org/archive/html/emacs-devel/2022-02/msg00461.html
for more context.

I will provide a corresponding PR for lsp-fsharp later too.

I didn't hear other runtimes than dotnet experiencing issues with a switch to posix_spawn in emacs, but just FYI -- the latest emacs@master (@emacs-28 on macos too) launches subprocesses with SIGCHLD and other signals masked and unless subprocess fixes it up by itself, that may cause problems.


razzmatazz:
@yyoncho I have applied a couple more commits to make this PR even better :)

yyoncho (approved):

PR #3350, opened 2022-02-13
Lsp fsharp installation update

razzmatazz:
This PR builds on top of https://github.com/emacs-lsp/lsp-mode/pull/3327 by @drvink

- This includes all changes from #3327 sans the startup code changes
- Updates installation mechanism for fsautocomplete as FsAutoComplete has switched to `dotnet tool`-based distribution mechanism completely;
- Provides a workaround for statup on emacs-29 (and emacs-28 on macOS) -- see #3349

PR #3362, opened 2022-02-22
make fsautocomplete working without change user config

s-kostyaev:
There is no need to force user update their configuration if we already know where to find executable.

@razzmatazz 

Relates #3350

### DIFF
--- a/clients/lsp-fsharp.el
+++ b/clients/lsp-fsharp.el
@@ -194,7 +194,8 @@ Will invoke CALLBACK or ERROR-CALLBACK based on result. Will update if UPDATE? i
 
                                (t nil)))
         (fsautocomplete-exec (or (executable-find "fsautocomplete")
-                                 "fsautocomplete")))
+                                 (f-join (or (getenv "USERPROFILE") (getenv "HOME"))
+                                         ".dotnet" "tools" "fsautocomplete"))))
     (append startup-wrapper
             (list fsautocomplete-exec "--background-service-enabled")
             lsp-fsharp-server-args)))


razzmatazz:
Hmm this will not work on windows and actually depends on where dotnet_home env points to..

razzmatazz:
Again this is more generally a queation to @yyoncho on how we deal with broken/unconfigured  exec-path in lsp mode? What is the policy?

s-kostyaev:
> dotnet_home

How it looks like on windows? I can add more branches...

razzmatazz:
> > dotnet_home
> 
> How it looks like on windows? I can add more branches...

There is some info here:
- https://docs.microsoft.com/en-us/dotnet/core/tools/global-tools#install-a-global-tool

but to reiterate, I think we should check what other lsp clients impl do with broken exec-path before merging. I agree with the intent but I am a bit wary for adding too much magic if other clients (node npm?) fix this some other way in lsp mode..

s-kostyaev:
@razzmatazz like this?

s-kostyaev:
> Again this is more generally a question to @yyoncho on how do we deal with broken/unconfigured exec-path in lsp mode? What is the policy?

example:
```elisp
(defcustom lsp-clients-flow-server "flow"
  "The Flow executable to use.
Leave as just the executable name to use the default behavior of
finding the executable with variable `exec-path'."
  :group 'lsp-flow
  :risky t
  :type 'file)
```

razzmatazz:
I don’t know. A good approach maybe would be to indicate the possible fix to the user somehow once we cannot find fsautocomplete on exec-path? But when and how?

There is lsp-doctor I think.., but that is not easily discoverable..

s-kostyaev:
@yyoncho we need you :smile:

razzmatazz:
@jcs090218 too :)

s-kostyaev:
If we decide to merge it - it will work with automatically installed `fsautocomlete`. In other case we will install it, but it will not work without changing user configuration.

yyoncho:
It looks fine to me. I read the `flow` example as an instruction.

yyoncho:
IMHO if the user typically does not have this binary on the path and the case handled by this PR will be hit by all users if they follow the default installation steps then it makes sense to have the PR merged. lsp-mode should work as much OOTB as possible. If this piece of code improves the experience I don't see anything against it as long as it will be useful for the users.

s-kostyaev:
> IMHO if the user typically does not have this binary on the path and the case handled by this PR will be hit by all users if they follow the default installation steps then it makes sense to have the PR merged. lsp-mode should work as much OOTB as possible. If this piece of code improves the experience I don't see anything against it as long as it will be useful for the users.

That's why I have created this PR. @razzmatazz @TOTBWF What do you think?

TOTBWF:
I don't write a ton of F# these days so I'm not super equipped to answer this, but this seems reasonable enough to me. Anything that makes the initial onboarding experience simpler is an easy win!

s-kostyaev:
@yyoncho can we merge it?

razzmatazz:
> @yyoncho can we merge it?

I think the agreement is that this is a good approach after all, to minimize setup where possible. I will test the Linux code-path and then I can merge.

razzmatazz (approved):
I have tested the changes and it seems to work for me, on Linux at least (when simulating `(executable-find "fsac")` to be failing

yyoncho:
> @yyoncho can we merge it?

@s-kostyaev 

I believe that decisions like that should be taken by fsharp/csharp maintainers/contributors/users. I mean, the decision is mainly affecting them.

PR #3366, opened 2022-02-23
Disable lsp-steep-use-bundler by default

buzztaiki:
When `lsp-steep-use-bundler` is `non-nil`, `lsp-steep` is enabled even so `steep` command does not exists, because a first value of `lsp-steep--build-command` is `bundle` instead of `steep`. I think this is not default behavior.

https://github.com/emacs-lsp/lsp-mode/blob/cf87368054f32f9ecd3960f79f0815fbf97d798b/clients/lsp-steep.el#L57-L61

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -21,6 +21,7 @@
   * Fix fsautocomplete (lsp-fsharp) startup on emacs@master (and emacs@emacs-28 on macOS) due to
     a switch ~posix_spawn~ and not setting proper sigmask on child processes.
   * Add TypeProf support.
+  * Disable ~lsp-steep-use-bundler~ by default.
 ** Release 8.0.0
   * Add ~lsp-clients-angular-node-get-prefix-command~ to get the Angular server from another location which is still has ~/lib/node_modules~ in it.
   * Set ~lsp-clients-angular-language-server-command~ after the first connection to speed up subsequent connections.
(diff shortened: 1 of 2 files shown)


yyoncho:
CC @ybiquitous

ybiquitous:
@buzztaiki Thanks for the proposal!

Other Ruby lang servers disabled the Bundler support by default, so it's better to disable it for Steep as well. 👍🏼 
(I forgot why I enabled it by default, though... 😅 )

- https://github.com/emacs-lsp/lsp-mode/blob/cf87368054f32f9ecd3960f79f0815fbf97d798b/clients/lsp-solargraph.el#L106
- https://github.com/emacs-lsp/lsp-mode/blob/cf87368054f32f9ecd3960f79f0815fbf97d798b/clients/lsp-sorbet.el#L35
- https://github.com/emacs-lsp/lsp-mode/blob/cf87368054f32f9ecd3960f79f0815fbf97d798b/clients/lsp-typeprof.el#L35

BTW, the latest Ruby 3.1 [bundles the TypeProf gem by default](https://docs.ruby-lang.org/en/3.1/doc/standard_library_rdoc.html#label-Bundled+gems), so `lsp-typeprof` will be selected in a clean environment of Ruby. This means Ruby users can the Ruby lang server with zero configuration. Great. 🙌🏼

yyoncho:
@ybiquitous are you fine with this change? It is beyond my expertise.

ybiquitous:
I agree with this change. 👍🏼

buzztaiki:
Thanks guys!

PR #3373, opened 2022-02-26
Fix `lsp-rust-analyzer-debug` hangs with long running processes

mkhattab:
Calling this function `lsp-rust-analyzer-debug` to debug apps with long running processes, e.g. when debugging a web server, causes this function to hang due to `cargo run` being called here instead of `cargo build` as intended.

The use of `cl-case` here is inappropriate for string values because it uses `eql` according to the [docs](https://www.gnu.org/software/emacs/manual/html_node/elisp/Pattern_002dMatching-Conditional.html). Updating this line to use `pcase` fixes this issue.


yyoncho (approved):

yyoncho:
Thank you!

Issue #3381, opened 2022-03-06
Support emmet-ls


sei40kr:
I want emmet-ls support in lsp-mode.

Actually I implemented on my own & it worked.

But `make docs` didn't work and couldn't update docs.
https://github.com/sei40kr/lsp-mode/commit/2bc0b1be9066c23a9ef00ea850fec9c65c1087c4

<details>
<summary>The results of $ make docs</summary>

```
$ make docs
make -C docs/ generate
make[1]: Entering directory '/home/sei40kr/ghq/github.com/sei40kr/lsp-mode/docs'
cask install

Error: void-variable (package-recipe--working-tree)
  debug-early-backtrace()
  debug-early(error (void-variable package-recipe--working-tree))
  byte-code(    \"
\"!#
        \"$     \"      \"\"#!\"!\"\"##$" [package-recipe--working-tree rcp package-build-working-dir package-recipe--upstream-url defmethod (rcp package-recipe) file-name-as-directory expand-file-name eieio-oref name (rcp package-recipe) url format url-format repo defalias package-git-recipe-p eieio-make-class-predicate package-git-recipe package-git-recipe--eieio-childp eieio-make-child-predicate package-git-recipe-child-p make-obsolete "use (cl-typep ... 'package-git-recipe) instead" "25.1" define-symbol-prop cl-deftype-satisfies eieio-defclass-internal (package-recipe) ((tag-regexp :initform "\\([0-9]\\{4\\}-[0-9]\\{2\\}-[0-9]\\{2\\} [0-9]\\{2\\}:[0-9]\\{2\\}:[0-9]\\{2\\}\\( [+-][0-9]\\{4\\}\\)?\\)")) nil] 8)
  require(package-recipe)
  load-with-code-conversion("/home/sei40kr/.config/emacs/.cask/29.0/bootstrap/package-build-2.4/package-build.el" "/home/sei40kr/.config/emacs/.cask/29.0/bootstrap/package-build-2.4/package-build.el" nil t)
  require(package-build)
  (if (featurep pkg) nil (if (package-installed-p pkg) nil (if package-archive-contents nil (package-refresh-contents)) (package-install pkg)) (require pkg))
  (let ((pkg (car --dolist-tail--))) (if (featurep pkg) nil (if (package-installed-p pkg) nil (if package-archive-contents nil (package-refresh-contents)) (package-install pkg)) (require pkg)) (setq --dolist-tail-- (cdr --dolist-tail--)))
[...]

PR #3376, opened 2022-03-02
Add client for Ansible

fredericgiquel:
Add `lsp-ansible` using https://github.com/ansible/ansible-language-server

PR #3384, opened 2022-03-06
Add emmet-ls support

sei40kr:
Implements #3381.

Add [emmet-ls](https://github.com/aca/emmet-ls) support.

### DIFF
--- a/clients/lsp-emmet.el
+++ b/clients/lsp-emmet.el
@@ -1,0 +1,65 @@
+;;; lsp-emmet.el --- lsp-mode Emmet integration -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2022 emacs-lsp maintainers
+
+;; Author: lsp-mode maintainers
+;; Keywords: lsp, emmet
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; LSP Client for Emmet
+
+;;; Code:
+
+(require 'lsp-mode)
+
+;;; emmet-ls
+(defgroup lsp-emmet-ls nil
+  "Settings for emmet-ls."
+  :group 'lsp-mode
+  :link '(url-link "https://github.com/aca/emmet-ls")
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defcustom lsp-emmet-ls-command '("emmet-ls" "--stdio")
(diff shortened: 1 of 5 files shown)


yyoncho:
can you follow this PR https://github.com/emacs-lsp/lsp-mode/pull/3376 as an example?

yyoncho:
> can you follow this PR #3376 as an example?

more specifically, the places you have to change.

sei40kr:
@yyoncho Sorry, I've fixed them.

ericdallo (approved):
Docs seems ok

clients/lsp-emmet.el
```diff
@@ -0,0 +1,68 @@
...
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection
+                   (lambda ()
+                     `(,(or (executable-find (cl-first lsp-emmet-ls-command))
+                            (lsp-package-path 'emmet-ls))
+                       ,@(cl-rest lsp-emmet-ls-command))))
+  :activation-fn (lambda (filename &optional _mode)
+                   (or (string-suffix-p ".jsx" filename)
+                       (string-suffix-p ".tsx" filename)
```
yyoncho, line 56: I can merge it as it is, but it will be a bit better to associate these extensions with a particular language(I guess it is well known) and use solely activate-on. Let me know if you want to fix that. If not, I will merge it as it is. 

`(lsp-activate-on "html" "css" "scss" "less")` returns a function, so you should so `(funcall   (lsp-activate-on "html" "css" "scss" "less") ...`)

sei40kr: @yyoncho 
I want to fix it.

`(string-suffix-p ".tsx" filename)` -> `(lsp-activate-on "typescriptreact")` worked, but it also enables `emmet-ls` in `.ts` file (I don't know why).
I'm still investigating, so let me make this PR wip again.

yyoncho (approved):

yyoncho:
Thank you!

Out of curiosity, and also for future reference: why would one pick emmet-ls over let's say emmet-mode? (I am not using emmet-mode, so I am not very familiar with what emmet-ls will be better for).

sei40kr:
@yyoncho 
Thanks.

IMO, the biggest reason is fuzzy matching.
Fuzzy matching is very useful when we write CSS/Less/SCSS because it's hard for us to remember the Emmet abbreviations of all CSS properties.

In addition, [emmet-mode](https://github.com/smihica/emmet-mode) is no longer maintained. In contrast, [emmet-ls](https://github.com/aca/emmet-ls) is being actively maintained.

But there's still a problem with `emmet-ls`. When we try to expand

```
p+p
```

`emmet-mode` can expand this correctly by pressing Tab.
`emmet-ls` can complete this, but can't expand this by pressing Tab. It results in

```
p+<p></p>
```

ericdallo:
Sounds like a server issue on emmet-ls indeed @sei40kr

S0mbr3:
I don't think it is a server issue, I am using both Emacs and Neovim and with the exact same server, p+p or p*2 does well trigger lsp completion menu and expands it correctly on the Neovim side.

pedroprieto:
Hi, @sei40kr . I can't make emmet-ls work in Emacs. I have yasnippet, company, LSP and LSP-UI. LSP loads emmet-ls, but I can't trigger any expansion. Also, no suggestions are shown. Other LSP languages work (JS, for example). Any thoughts? Is there any option missed? Or any binding key?

Thanks!!

PR #3400, opened 2022-03-11
Added support for the PerlNavigator LSP server along with automatic i…

ketbra:
I've added support for PerlNavigator which is a new Perl LSP server from https://github.com/bscan/PerlNavigator/ .  In my opinion it provides an enhanced experience compared with the existing Perl Language Server included with LSP mode.  My changes also include automatic installation by downloading the latest self contained executables from the project's github releases. Thanks for LSP mode.  I use it daily and love it!


yyoncho (approved):

yyoncho:

PR #3401, opened 2022-03-12
Fix a few rust-analyzer customs

wyuenho:
`lsp-rust-analyzer-cargo-target` was set to the wrong group, and it and `lsp-rust-analyzer-rustc-source`'s types are wrong.


yyoncho (approved):

yyoncho:
Thank you!

PR #3404, opened 2022-03-15
Update support for rust-analyzer's inlay hints

### LINKED ISSUE #3403: lens-mode / inlayHints broken with rust-analyzer since 2022-03-04

### Bug description

No inlay hints are displayed when using lsp-mode + lsp-lens-mode with rust-analyzer since https://github.com/rust-analyzer/rust-analyzer/pull/11445. lsp-log reports

```
[Trace - 06:19:24 PM] Received response 'rust-analyzer/inlayHints - (204)' in 23ms.
Result: {
  "message": "unknown request",
  "code": -32601
}
```

as the inlayHints method was renamed to `experimental/inlayHints` and has apparently gotten some semantic changes.

### Steps to reproduce

install latest lsp-mode + rust-analyzer,

```
    let x = 123;
```

should have a type annotation.

### Expected behavior

inlay hints are shown.

rust-analyzer

### OS

MacOS

### Error callstack

```emacs-lisp
[Trace - 06:19:24 PM] Received response 'rust-analyzer/inlayHints - (204)' in 23ms.
Result: {
  "message": "unknown request",
  "code": -32601
}
```

### Anything else?

_No response_

rjmac:
I believe that this is no longer actually rust-analyzer-specific and
instead will work with any LSP backend that supports the draft inlay
hint feature, but I have no other backend to try with.

Fixes #3403

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -933,9 +933,8 @@ meaning."
          (dolist (hint res)
            (-let* (((&rust-analyzer:InlayHint :position :label :kind :padding-left :padding-right) hint)
                    (pos (lsp--position-to-point position))
-                   (overlay (make-overlay pos (+ pos 1) nil 'front-advance 'end-advance)))
+                   (overlay (make-overlay pos pos nil 'front-advance 'end-advance)))
              (overlay-put overlay 'lsp-rust-analyzer-inlay-hint t)
-             (overlay-put overlay 'evaporate t)
              (overlay-put overlay 'before-string
                           (format "%s%s%s"
                                   (if padding-left " " "")
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -850,7 +850,7 @@ https://rust-analyzer.github.io/manual.html#auto-import.
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "8.0.0"))
 
-(defcustom lsp-rust-analyzer-inlay-type-format ": %s"
+(defcustom lsp-rust-analyzer-inlay-type-format "%s"
   "Format string for variable inlays (part of the inlay face)."
   :type '(string :tag "String")
   :group 'lsp-rust-analyzer
@@ -869,7 +869,7 @@ https://rust-analyzer.github.io/manual.html#auto-import.
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "8.0.0"))
 
-(defcustom lsp-rust-analyzer-inlay-param-format "%s:"
+(defcustom lsp-rust-analyzer-inlay-param-format "%s"
   "Format string for parameter inlays (part of the inlay face)."
   :type '(string :tag "String")
   :group 'lsp-rust-analyzer
@@ -910,20 +910,20 @@ meaning."
   (if (and (lsp-rust-analyzer-initialized?)
            (eq buffer (current-buffer)))
       (lsp-request-async
-       "rust-analyzer/inlayHints"
+       "experimental/inlayHints"
        (lsp-make-rust-analyzer-inlay-hints-params
         :text-document (lsp--text-document-identifier))
        (lambda (res)
          (remove-overlays (point-min) (point-max) 'lsp-rust-analyzer-inlay-hint t)
          (dolist (hint res)
-           (-let* (((&rust-analyzer:InlayHint :range :label :kind) hint)
-                   ((&RangeToPoint :start :end) range)
-                   (overlay (make-overlay start end nil 'front-advance 'end-advance)))
+           (-let* (((&rust-analyzer:InlayHint :position :label :kind) hint)
+                   (pos (lsp--position-to-point position))
+                   (overlay (make-overlay pos (+ pos 1) nil 'front-advance 'end-advance)))
              (overlay-put overlay 'lsp-rust-analyzer-inlay-hint t)
              (overlay-put overlay 'evaporate t)
              (cond
               ((equal kind lsp/rust-analyzer-inlay-hint-kind-type-hint)
-               (overlay-put overlay 'after-string
+               (overlay-put overlay 'before-string
                             (format lsp-rust-analyzer-inlay-type-space-format
                                     (propertize (format lsp-rust-analyzer-inlay-type-format label)
                                                 'font-lock-face 'lsp-rust-analyzer-inlay-type-face))))
@@ -935,7 +935,7 @@ meaning."
                                                 'font-lock-face 'lsp-rust-analyzer-inlay-param-face))))
 
               ((equal kind lsp/rust-analyzer-inlay-hint-kind-chaining-hint)
-               (overlay-put overlay 'after-string
+               (overlay-put overlay 'before-string
                             (format lsp-rust-analyzer-inlay-chain-space-format
                                     (propertize (format lsp-rust-analyzer-inlay-chain-format label)
                                                 'font-lock-face 'lsp-rust-analyzer-inlay-chain-face))))))))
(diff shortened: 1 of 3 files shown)


clients/lsp-rust.el
```diff
@@ -910,20 +910,20 @@ meaning."
...
         :text-document (lsp--text-document-identifier))
        (lambda (res)
          (remove-overlays (point-min) (point-max) 'lsp-rust-analyzer-inlay-hint t)
          (dolist (hint res)
-           (-let* (((&rust-analyzer:InlayHint :range :label :kind) hint)
-                   ((&RangeToPoint :start :end) range)
-                   (overlay (make-overlay start end nil 'front-advance 'end-advance)))
+           (-let* (((&rust-analyzer:InlayHint :position :label :kind) hint)
+                   (pos (lsp--position-to-point position))
+                   (overlay (make-overlay pos (+ pos 1) nil 'front-advance 'end-advance)))
```
rjmac, line 921: I'm not actually sure what that second parameter should be; `pos` doesn't work, but the position we get doesn't really have an _end_ per se.

lsp-protocol.el
```diff
@@ -397,9 +397,9 @@ See `-let' for a description of the destructuring mechanism."
 
 (lsp-interface (rls:Cmd (:args :binary :env :cwd) nil))
 
-(defconst lsp/rust-analyzer-inlay-hint-kind-type-hint "TypeHint")
-(defconst lsp/rust-analyzer-inlay-hint-kind-param-hint "ParameterHint")
-(defconst lsp/rust-analyzer-inlay-hint-kind-chaining-hint "ChainingHint")
+(defconst lsp/rust-analyzer-inlay-hint-kind-type-hint 1)
+(defconst lsp/rust-analyzer-inlay-hint-kind-param-hint 2)
+(defconst lsp/rust-analyzer-inlay-hint-kind-chaining-hint nil)
```
rjmac, line 402: This `nil` seems like it's a bug in rust-analyzer?  It feels like it should be `3` but it definitely isn't on my machine.

rjmac: Ah, so not a bug but an extension to the draft spec.

lnicola: Fixed in https://github.com/rust-analyzer/rust-analyzer/pull/11445

rjmac:
As I mentioned in that latest commit message, the emacs-side formatting of hints looks like it'll become less useful over time.  The new inlay hint protocol doesn't appear to be expressive enough to tell us what these hints _are_ (being restricted to "parameter" and "type") and indeed the new hints once again provide a null kind.  What it _does_ do is tell us about what padding should appear.

So anyway, I've undone the change where I asked rust-analyzer to never format the hints itself and made that configurable (defaulting to on).  When it is on, the formatting and spacing stuff is all ignored - the code assumes that the `label` comes back pre-formatted and uses the `paddingLeft` and `paddingRight` fields in the response instead of a spacing configuration.

I'm not super happy with it, to be honest, but I thought I'd put it out there and see what other people have to say.  My current inclination is to remove the elisp formatting code altogether, always have rust-analyzer format it, and use the `kind` field only to determine the exact font-face to display the inlay with, but I honestly have no _extremely_ strong opinion about what it should do, and I'm really only a drive-by contributor, so I'll happily take input from anybody (especially regular maintainers!)

rjmac:
But all that said, if you're curious, here are the new lifetime elision, closure return type, and reborrowing hints showing up in Emacs: 
image

yyoncho:
@flodiebold @brotzeit willing to take a look?

Veykril:
That sounds wrong to me, those kinds are solely meant for styling, with null falling back to the default styling the user chooses. I'm not sure what the problem here is though, `null` is a valid kind, the one that applies the default styling?

It's a classic case of the LSP being limited to only a few known things and not allowing custom kinds though which I believe is a serious flaw in the spec :/

clients/lsp-rust.el
```diff
@@ -342,6 +342,13 @@ PARAMS progress report notification data."
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "6.2"))
 
+(defcustom lsp-rust-analyzer-server-format-inlay-hints t
```
flodiebold, line 345: in general I'd prefer if the defcustoms mirror the naming of the options in rust-analyzer, but I recognize it's too late here :/

clients/lsp-rust.el
```diff
@@ -665,9 +696,13 @@ https://rust-analyzer.github.io/manual.html#auto-import.
             :unsetTest ,lsp-rust-analyzer-cargo-unset-test)
     :rustfmt (:extraArgs ,lsp-rust-analyzer-rustfmt-extra-args
               :overrideCommand ,lsp-rust-analyzer-rustfmt-override-command)
-    :inlayHints (:typeHints ,(lsp-json-bool lsp-rust-analyzer-server-display-inlay-hints)
+    :inlayHints (:renderColons ,(lsp-json-bool lsp-rust-analyzer-server-format-inlay-hints)
+                 :typeHints ,(lsp-json-bool lsp-rust-analyzer-server-display-inlay-hints)
                  :chainingHints ,(lsp-json-bool lsp-rust-analyzer-display-chaining-hints)
+                 :closureReturnTypeHints ,(lsp-json-bool lsp-rust-analyzer-display-closure-return-type-hints)
+                 :lifetimeElisionHints ,lsp-rust-analyzer-display-lifetime-elision-hints
```
flodiebold, line 703: I think the structure of this setting might change again soon (right, @Veykril ?), so maybe it's best to just leave it out

clients/lsp-rust.el
flodiebold: My guess would be that `pos` doesn't work because we set `evaporate`, which makes Emacs remove the hints when they have length 0. I don't know if that's necessary, since we also remove the overlays manually when we update them. Maybe @yyoncho knows more here

flodiebold (approved):
LGTM mostly. I agree that letting rust-analyzer format the hints is probably the way forward.

clients/lsp-rust.el
flodiebold: Left that on the wrong line, it was referring to `lsp-rust-analyzer-display-lifetime-elision-hints` etc.

rjmac:
Ah ok, that makes more sense.  Looking at it in that light, I think I can make it a fair bit less scattered-feeling today, then I'll consider the PR "done".

clients/lsp-rust.el
rjmac: I believe you're right, that's rust-analyzer/rust-analyzer#11778.  Hasn't actually happened yet, but I'm aware it almost certainly will.

Veykril: I went ahead and changed one of the configs here https://github.com/rust-analyzer/rust-analyzer/pull/11789/files

Note that r-a might have to change a bunch of configs in the future for consistency ( we dug ourselves a bad hole here https://github.com/rust-analyzer/rust-analyzer/issues/11790 )

rjmac:
Ok, I'm pretty happy with the PR as it is now.  I incorporated most of the feedback above, specifically:
* The new-as-of-today lifetime hint configuration is used
* The overlay creation code has been substantially simplified so there's really only one majro path through it.  Spacing is now controlled purely by the padding values in the inlay information rust-analyzer returns rather than emacs-side config.  I've left in emacs-side config for formatting "type" and "parameter" hints (default disabled, just letting rust-analyzer handle it) but always use the kind to determine which face to lay out the overlays in.
* I _didn't_ do the the `defvaralias` thing mainly because it wasn't obvious to me how it interacted with the customization system.

brotzeit:
> I didn't do the the defvaralias thing mainly because it wasn't obvious to me how it interacted with the customization system.

AFAIK it shouldn't make a difference if we don't change `:type` for the new `defcustom`. But you don't have to apply it, maybe I'll update the names. I merge this now, thanks for your work.

brotzeit (approved):

rksm:
Thank you @rjmac, very much appreciated!

PR #3407, opened 2022-03-15
disable failing tests with emacs snapshot

brotzeit:


brotzeit:
I think it always looks better, especially for new users, when tests are green. And I doubt cask will get fixed soon.

PR #3409, opened 2022-03-16
Add support for TypeScript Language Server's format options

AlynxZhou:
TypeScript Language Server has [some format options](https://github.com/typescript-language-server/typescript-language-server#workspacedidchangeconfiguration) that needs to be send via `workspace/didChangeConfiguration` notification, currently user has no way to pass those settings to language server, this commit will set those options.

However, it seems that Vue Language Server shares the same options with TypeScript Language Server for TypeScript and JavaScript, so the related custom settings are already registered in [lsp-vetur.el](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-vetur.el#L728-L800), I just simply load those options here. And user should also use the same [custom options for lsp-vetur](https://emacs-lsp.github.io/lsp-mode/page/lsp-vetur/).


AlynxZhou:
I am not sure if it's OK for two servers sharing the same custom setting paths, but it seems that it's chosen by server upstream, and we can not split them with current code (in other words, we have no things like hasp maps to store different settings for different servers, but just using the prefix of a path, so if upstream server choose the same prefix, we can do nothing).

yyoncho:
I think we should do the following: 

1. Move all typescript/javascript properties in lsp-javascript.el
2. Require lsp-javascript.el from lsp-vetur.el

Let me know if you are willing to do that. If not, I can merge the PR as it is.

AlynxZhou:
> I think we should do the following:
> 
>     1. Move all typescript/javascript properties in lsp-javascript.el
> 
>     2. Require lsp-javascript.el from lsp-vetur.el
> 
> 
> Let me know if you are willing to do that. If not, I can merge the PR as it is.

OK, I can try to do this.

AlynxZhou:
Please check my new commit, I am a newbie to elisp and not sure whether a single require in the head of the file will work.

clients/lsp-vetur.el
```diff
@@ -727,21 +551,6 @@ Code's JavaScript and TypeScript support."
...
-   ("javascript.format.insertSpaceAfterOpeningAndBeforeClosingTemplateStringBraces" lsp-javascript-format-insert-space-after-opening-and-before-closing-template-string-braces t)
-   ("javascript.format.insertSpaceAfterSemicolonInForStatements" lsp-javascript-format-insert-space-after-semicolon-in-for-statements t)
-   ("javascript.format.insertSpaceBeforeAndAfterBinaryOperators" lsp-javascript-format-insert-space-before-and-after-binary-operators t)
-   ("javascript.format.insertSpaceBeforeFunctionParenthesis" lsp-javascript-format-insert-space-before-function-parenthesis t)
-   ("javascript.format.placeOpenBraceOnNewLineForControlBlocks" lsp-javascript-format-place-open-brace-on-new-line-for-control-blocks t)
-   ("javascript.format.placeOpenBraceOnNewLineForFunctions" lsp-javascript-format-place-open-brace-on-new-line-for-functions t)
    ("javascript.implicitProjectConfig.checkJs" lsp-javascript-implicit-project-config-check-js t)
    ("javascript.implicitProjectConfig.experimentalDecorators" lsp-javascript-implicit-project-config-experimental-decorators t)
    ("javascript.preferences.importModuleSpecifier" lsp-javascript-preferences-import-module-specifier)
```
yyoncho, line 556: can you move these as well? At some point ts-ls will start consuming them as well.

yyoncho:
> Please check my new commit, I am a newbie to elisp and not sure whether a single require in the head of the file will work.

Looks good.

clients/lsp-vetur.el
yyoncho: I mean, can you move *all* js/ts properties from lsp-vetur.el to lsp-javascript.el

AlynxZhou: Those options are not in ts-ls's server documents, I am not sure whether those are supported, I just started to use lsp recently and not so familiar with it. So I only move format options I found in <https://github.com/typescript-language-server/typescript-language-server#workspacedidchangeconfiguration>.

yyoncho: Yes, that is true. They are part of tsserver configuration and most likely they will be exposed at some point from typescript-language-server as well. So better have everything in place right now.

AlynxZhou: I'll try to do it, but not sure whether it will work.

yyoncho: Effectively it won't change how the code behaves.

AlynxZhou: I just did a rebase to make the git log more clear.

yyoncho (approved):

yyoncho:
Thank you!

hungnh17:
hey @yyoncho , sorry for bothering you but what about insertSpaceAfterOpeningAndBeforeClosingEmptyBraces ? and how can I set it via config.el ? I am a lisp newbie

PR #3416, opened 2022-03-17
Fix documentation editing

### LINKED ISSUE #3413: All of the Github "Edit this page" links on https://emacs-lsp.github.io/lsp-mode/ are broken

### Bug description

On https://emacs-lsp.github.io/lsp-mode/ for example, if you click the "Edit this page" (pencil icon) link, it goes to https://github.com/emacs-lsp/lsp-mode/edit/master/docs/README.md which is a 404. I can't work out what the correct link should be.

This also means that I can't work out how to contribute help or fixes to any of the pages.

I also can't work out if this is the right place to report this bug, but there doesn't seem to be anywhere else.

### Steps to reproduce

-

### Expected behavior

-

-

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

elken:
Correctly resolve all editable links on the documentation (blog posts
and changelogs are left as-is, this can be easily modified).

There is now an expectation to find a `root_file` which is a string of a
file to resolve from the root of the lsp-mode tree.

The way we discover which files belong where is by:
- First checking if we can evaluate `symbol-file` for the group that
matches the name of the client (eg "lsp-csharp")
- Then checking for a file name in clients for a common group name
- Lastly checking for a file in clients (eg "clients/lsp-csharp.el")

Fixes #3413


yyoncho:
Looks good to me! I will let @ericdallo take a look and proceed with the merge.

.gitignore
```diff
@@ -54,3 +54,8 @@ target
 .project
 .settings
 /ert-profile
+docs/page/lsp-*.md
+/docs/page/settings/
+/docs/page/CHANGELOG.md
+/docs/examples/
+docs/README.md
```
ericdallo, line 61: We need to take care `ghpages` branch doesn't use that as ignore, but Ithink this will apply for master only

ericdallo (approved):
Looks good! thank you!

.gitignore
elken: You're right that it _should_, if it has the potential to cause headache I can remove it though

PR #3420, opened 2022-03-18
Resolve build issues with mkdocs

elken:
Per a note[1] on the mkdocs-material page, it seems the way this is
built doesn't allow for a correct build pipeline as it was ignoring any
changes in the custom_dir.

This should hopefully resolve and correctly build the docs.

If preferred, I can instead try and update the docs-image dockerfile to build the project correctly.

Works fine on my fork: https://elken.github.io/lsp-mode/
Pipeline pass: https://github.com/elken/lsp-mode/runs/5598210113?check_suite_focus=true

[1] https://squidfunk.github.io/mkdocs-material/customization/#setup-and-theme-structure


yyoncho (approved):

ericdallo (changes requested):
@elken I don't think this is a good idea, we have emacs-lsp/docs-image exactly to keeping a docker image of the dependencies we need with the proper changes we need, if you need to change the image I think we should change the docker image, and release a new version, LMK if need any help with that

https://github.com/emacs-lsp/docs-image

elken:
Sure, I'll make changes to docs-image instead 😃

PR #3425, opened 2022-03-20
allow using only executable name for csharp server path

contrun:
Current `(setq lsp-csharp-server-path "omnisharp")` does not work even if "omnisharp" is in `$PATH`. This patch makes the csharp client also search for executable in the `$PATH`.

### DIFF
--- a/clients/lsp-csharp.el
+++ b/clients/lsp-csharp.el
@@ -132,7 +132,7 @@ Will invoke CALLBACK on success, ERROR-CALLBACK on error."
 (defun lsp-csharp--language-server-path ()
   "Resolve path to use to start the server."
   (if lsp-csharp-server-path
-      lsp-csharp-server-path
+      (executable-find lsp-csharp-server-path)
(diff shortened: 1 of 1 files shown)


jcs090218:
I think the variable `lsp-csharp-server-path` is the direct path to the server executable? Just the name of the string isn't sufficient for lsp-mode to find it's target. Maybe we can do something like,

```el
(or lsp-csharp-server-path
    (executable-find "omnisharp")
    (let ((server-dir lsp-csharp-omnisharp-roslyn-server-dir))
      (when (f-exists? server-dir)
        (f-join server-dir (cond ((eq system-type 'windows-nt) "OmniSharp.exe")
                                 (t "run"))))))
```

elken:
[This](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-php.el#L396-L410) is what phpactor does, if that helps.

yyoncho (approved):

PR #3455, opened 2022-04-08
Add support for inlay hints from TypeScript language server

shivjm:
This is an inexpert adaptation (mainly duplication) of [the recently-added support for rust-analyzer’s inlay hints](https://github.com/emacs-lsp/lsp-mode/pull/3404) to ts-ls. Here’s how it looks in JavaScript and TypeScript, with bright backgrounds for the hints so they stand out in the image:

image

[The language server suggests a fair bit of flexibility,](https://github.com/typescript-language-server/typescript-language-server#inlay-hints-typescriptinlayhints-experimental) almost all of which is reproduced in this PR.[^1] For example, the `return new Entity` line that the cursor is on in the top window would normally have hints for the parameters too, but I’ve set `lsp-javascript-display-parameter-name-hints-when-argument-matches-name` to `nil`, so, for example, passing a `status` variable as the `status` parameter suppresses the hint.

Note that the server’s support is itself marked experimental. In particular, I can’t seem to get any enum member value hints, despite the existence of the option. (Hence the commented-out case in `lsp-javascript-format-inlay`.)

[^1]: It didn’t seem useful to have separate parameters for JS and TS, so I clubbed them together. I’m open to the idea that someone else might prefer to distinguish between them.

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -600,11 +600,87 @@ TypeScript 3.0 or newer in the workspace."
   :type 'boolean
   :package-version '(lsp-mode . "6.1"))
 
+;; inlay hints
+
+(defface lsp-javascript-inlay-face
+  '((t :inherit font-lock-comment-face))
+  "The face to use for the JavaScript inlays."
+  :group 'lsp-javascript
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defface lsp-javascript-inlay-type-face
+  '((t :inherit lsp-javascript-inlay-face))
+  "Face for inlay type hints (e.g. inferred variable types)."
+  :group 'lsp-javascript
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defcustom lsp-javascript-inlay-type-format ": %s"
+  "Format string for variable inlays (part of the inlay face)."
+  :type '(string :tag "String")
+  :group 'lsp-javascript
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defface lsp-javascript-inlay-parameter-face
+  '((t :inherit lsp-javascript-inlay-face))
+  "Face for inlay parameter hints (e.g. function parameter names at
+call-site)."
+  :group 'lsp-javascript
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defcustom lsp-javascript-inlay-param-format "%s:"
+  "Format string for parameter inlays (part of the inlay face)."
+  :type '(string :tag "String")
+  :group 'lsp-javascript
+  :package-version '(lsp-mode . "8.0.1"))
+
 (defcustom lsp-typescript-surveys-enabled t
   "Enabled/disable occasional surveys that help us improve VS
 Code's JavaScript and TypeScript support."
   :type 'boolean
   :package-version '(lsp-mode . "6.1"))
+
+(defcustom lsp-javascript-display-inlay-hints nil
+  "Whether to display inlay hints."
+  :type 'boolean
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defcustom lsp-javascript-display-enum-member-value-hints nil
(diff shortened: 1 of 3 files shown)


shivjm:
There’s spec-related discussion in a few different places:

* microsoft/language-server-protocol#956 is a closed issue that mentions a few other PRs
* microsoft/language-server-protocol#1249 is a closed PR with a comment that [the draft was added to the spec](https://github.com/microsoft/language-server-protocol/commit/c63b5f5266b3ffc1c83747f782cc70f0fa1e38e8), but [that’s in the not-yet-released 3.17 version](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_inlayHint)
* microsoft/vscode-languageserver-node#772 is an open PR to add it to the protocol and the client for that server

The TypeScript language server itself puts it under a file with ‘proposed’ in the name. Its implementation is slightly different. I don’t see any mention of the spec in these:

* typescript-language-server/typescript-language-server#258 (issue about adding hints to the server)
* typescript-language-server/typescript-language-server#259 (PR for adding hints to the server)
* microsoft/TypeScript#42073 (issue about adding hints to TS itself)
* microsoft/TypeScript#42089 (PR for adding hints to TS itself)

Maybe for now this should be ts-ls–specific. We could re-evaluate it as and when the server itself adopts the official version. What do you think?

yyoncho:
> We could re-evaluate it as and when the server itself adopts the official version. What do you think?

Sounds good. It will be great if you can take care of this by monitoring ts-ls development so we can drop the ad-hoc protocol implementation in favor of a real one.

clients/lsp-javascript.el
```diff
@@ -727,6 +819,63 @@ Code's JavaScript and TypeScript support."
...
 
+(defun lsp-javascript-update-inlay-hints (buffer)
+  (if (and (lsp-javascript-initialized?)
+           (eq buffer (current-buffer)))
+      (lsp-request-async
+       "typescript/inlayHints"
+       (lsp-make-javascript-inlay-hints-params
+        :text-document (lsp--text-document-identifier))
+       (lambda (res)
+         (remove-overlays (point-min) (point-max) 'lsp-javascript-inlay-hint t)
```
yyoncho, line 830: I guess you have copied this from the rust implementation but here you should use `lsp--remove-overlays` to ensure it will work in narrowed buffers.

yyoncho (approved):
Looks good to me. I won't be that picky about this implementation since it most likely will be short-lived when the core spec implementation is added. In core implementation I would expect lsp-javascript-inlay-hints-timer to go away in favor of using `lsp-on-idle-hook`. The benefits are  that `lsp-on-idle-hook` will prevent the refresh of inlay hints when the completion dialog is running (thus making emacs more responsive).

clients/lsp-javascript.el
shivjm: Done, thanks.

shivjm:
> It will be great if you can take care of this by monitoring ts-ls development so we can drop the ad-hoc protocol implementation in favor of a real one.

Yup, happy to!

> Looks good to me. I won't be that picky about this implementation since it most likely will be short-lived when the core spec implementation is added. In core implementation I would expect lsp-javascript-inlay-hints-timer to go away in favor of using `lsp-on-idle-hook`. The benefits are that `lsp-on-idle-hook` will prevent the refresh of inlay hints when the completion dialog is running (thus making emacs more responsive).

That makes a lot of sense. I don’t want to unnecessarily bog the interface down. I’ve updated it to use `lsp-on-idle-hook` instead.

yyoncho:
Thank you! If you have a Twitter account it will be great if you can tweet about it( I will retweet it).

shivjm:
Thank **you** for the gentle and patient guidance. 😊 I unfortunately don’t have a Twitter account, or I would have been happy to.

PR #3456, opened 2022-04-11
Add support for PureScript LSP settings

alternateved:
Hello, 

I would like to add support for [purescript-language-server](https://github.com/nwolverson/purescript-language-server) commands `addSpagoSources` and `addNpmPath`. Similar configuration could be seen [here](https://github.com/nwolverson/purescript-language-server#vimcoc=). If there is anything that I would need to fix or change, please let me now.

Best wishes,
alternateved

### DIFF
--- a/clients/lsp-purescript.el
+++ b/clients/lsp-purescript.el
@@ -50,6 +50,20 @@
             (lsp-package-path 'purescript-language-server))
         lsp-purescript-server-args))
 
+(defcustom-lsp lsp-purescript-add-spago-sources t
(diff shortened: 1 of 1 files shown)


yyoncho:
Can you do it like this: https://github.com/yyoncho/lsp-mode/blob/master/clients/lsp-php.el#L88 ( this is the new official way to do it). Follow that as an example, just change package-version to 8.0.1

alternateved:
Of course! Would that work?

yyoncho (approved):

yyoncho:
> Of course! Would that work?

Looks fine to me.

PR #3460, opened 2022-04-12
Fix inlay hints being added to wrong buffer (lsp-javascript, lsp-rust)

### LINKED ISSUE #3459: Inlay hints added to wrong buffer when original buffer is not active (lsp-javascript, lsp-rust)

### Bug description

#3455 inherited a bug from #3404 where it’s possible for inlay hints to be added to the wrong buffer. For example, watch the Magit buffer in this recording:

Inlay hints are incorrectly added to the active (Magit) buffer instead of the relevant buffer

This is happening because the handler for the async request is operating on whatever the active buffer is.

### Steps to reproduce

These two steps must be performed in quick succession:

1. Open a JavaScript or Rust file while having `lsp-javascript-display-hints` or `lsp-rust-analyzer-server-display-hints` (respectively) set to `t`.
2. Immediately switch to another buffer.

### Expected behavior

Inlay hints are added to the buffer they belong to.

lsp-javascript
lsp-rust

### OS

Windows

### Error callstack

_No response_

### Anything else?

_No response_

shivjm:
Fixes #3459

The async handlers for both clients’ inlay hints were operating on the active buffer, which could differ from the original buffer. This PR ensures both operate only on the latter.

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -818,26 +818,28 @@ name (e.g. `data' variable passed as `data' parameter)."
       (lsp--info "Renamed '%s' to '%s'." name (file-name-nondirectory new)))))
 
 (defun lsp-javascript-update-inlay-hints ()
-  (if (lsp-javascript-initialized?)
-      (lsp-request-async
-       "typescript/inlayHints"
-       (lsp-make-javascript-inlay-hints-params
-        :text-document (lsp--text-document-identifier))
-       (lambda (res)
-         (lsp--remove-overlays 'lsp-javascript-inlay-hint)
-         (-each (gethash "inlayHints" res)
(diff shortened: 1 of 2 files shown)


shivjm:
Unrelated failure again, as far as I can tell.

yyoncho:
That should be handled by `:mode 'tick`. I wonder what is causing the issue here.

shivjm:
I can honestly say I have no idea, heh. I only know that selecting the buffer explicitly fixes the problem.

shivjm:
The `:mode 'tick` was outside the function call. I should have realized from the indentation. 🤦‍♂️ Sorry to have wasted your time. I’ve added the one-line fix to #3489.

PR #3465, opened 2022-04-13
Use Eask for CI

jcs090218:
This is the same PR with https://github.com/emacs-lsp/lsp-ui/pull/700 and https://github.com/emacs-lsp/dap-mode/pull/596

Special changes to this PR:

* I have temporary disabled the ert test due to this error https://github.com/jcs-PR/lsp-mode/runs/6003823570?check_suite_focus=true#step:13:1168

```
Unknown key: :something-unrelated.  Available keys: (:start :end)
```

It seems like I couldn't evaluate the test script directly `./test/lsp-protocol-test.el`? Did I miss something?


ericdallo (approved):

jcs090218:
We might need to change requirements as well for this PR to merge?

ericdallo:
Yep, I can change that after merge, LMK when we can merge

yyoncho:
The failing test is written like that:

```
    (should-error (pcase particular-range
                    ((MyRange :something-unrelated)
                     t)
                    (_ nil)))
```
So the error means the test should pass. I am not quite sure why this happens. Maybe the ci is running in a mode that means any error should break the job.

jcs090218:
Ah.. I think loading ERT shouldn't trigger any error. Now tests can be operated without stopping.

It leads to some interesting results:

#### ubuntu-latest, 27.1 (Cask)

```
LSP :: Shutdown pyls:3498 since folder /home/runner/work/lsp-mode/lsp-mode/test/fixtures is removed...
..............................................................

Ran 62 tests in 20.910 seconds
```

#### ubuntu-latest, 27.1 (Eask)

```
Ran 75 tests, 63 results as expected, 12 unexpected (2022-04-15 07:23:31+0000, 23.612242 sec)

12 unexpected results:
   FAILED  lsp-file-notifications-test
   FAILED  lsp-file-watch--adding-watches
   FAILED  lsp-file-watch--non-existing
   FAILED  lsp-file-watch--recursive
   FAILED  lsp-file-watch--relative-path-glob-patterns
   FAILED  lsp-org-position-translation-test
   FAILED  lsp-org-test-current-org-mode-content
   FAILED  lsp-org-transformation-of-did-change-events-1
   FAILED  lsp-org-transformation-of-did-change-events-2
   FAILED  lsp-test-pcase-patterns
   FAILED  lsp-transformation-of-did-change-events
   FAILED  lsp-virtual-position-to-point
```

Why are there tests `75` in Eask but only tests `62` in Cask? 😕 The tests seem to be accurate since I cannot get past these tests on my machine as well (Windows, Emacs 29.0.50).

Total of 80 `ert-deftest` appeared in the project, and 5 redefined tests. I assumed 75 is the correct number of the total test cases?

Image 1

Issue #2464, opened 2020-12-27
[3.16] Support DidRenameFiles notification


ericdallo:
[SPEC](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_didRenameFiles)

* Add client capability `workspace.fileOperations.didRename`
* :arrow_right: send `workspace/didRenameFiles` notification

PR #3482, opened 2022-04-21
Add support for rename file operations

### LINKED ISSUE #2463: [3.16] Support WillRenameFiles request
[SPEC](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#workspace_willRenameFiles)

* Add client capability `workspace.fileOperations.willRename`
* :leftwards_arrow_with_hook: support `workspace/willRenameFiles`

ericdallo:
Fixes #2463 #2464 

Tested with Dart LSP server which already supports both methods.


lsp-mode.el
```diff
@@ -6013,6 +6028,26 @@ relied upon."
...
+Applies OLD-FUNC with OLD-NAME, NEW-NAME and OK-IF-ALREADY-EXISTS?.
+
+This advice sends workspace/willRenameFiles before renaming file
+to check if server wants to apply any workspaceEdits after renamed."
+  (if (lsp--send-will-rename-files-p)
+      (let ((params (lsp-make-rename-files-params
+                     :files (vector (lsp-make-file-rename
+                                     :oldUri (lsp--path-to-uri old-name)
+                                     :newUri (lsp--path-to-uri new-name))))))
+        (when-let ((edits (lsp-request "workspace/willRenameFiles" params)))
```
ericdallo, line 6042: SPEC says clients could drop the request if taking too much time, since Emacs has C-g not sure we should add a custom logic to timeout the request and not sure how hard would be to do that, so It seems ok to me

yyoncho: We already have timeout.

yyoncho:
Can you add flags to turn off the feature?

ericdallo:
Done @yyoncho !

yyoncho (approved):
Looks good. I already did some work on creating the confirmation flow, we may plug it at some point.

ericdallo:
@yyoncho Great!
I'm working on clojure-lsp implementation side, and just noticed we should check filters set by server to avoid sending the request for every renamed file (not that would be a problem), will open a next PR soon

PR #3489, opened 2022-04-25
Only display JavaScript inlay hints for code in view

### LINKED ISSUE #3459: Inlay hints added to wrong buffer when original buffer is not active (lsp-javascript, lsp-rust)

shivjm:
This drastically speeds up working with large files. A work project has a 90,000-line JavaScript file. Before, I couldn’t edit it with inlay hints enabled since it would just keep locking up (even after the initial delay while it was being parsed and everything). In contrast, with this change, inlay hints have virtually no impact on performance regardless of the size of the file. In fact, I tried making the hints more aggressive by updating them when scrolling or changing the size of the window and noticed no impact on responsiveness. (I added a custom variable to allow disabling this in case it does impact performance on other machines.)

Bonus:
Fixes #3459

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -825,7 +825,7 @@ name (e.g. `data' variable passed as `data' parameter)."
         :text-document (lsp--text-document-identifier))
        (lambda (res)
          (lsp--remove-overlays 'lsp-javascript-inlay-hint)
-         (-each (gethash "inlayHints" res)
+         (-each (lsp-get res :inlayHints)
            #'(lambda (hint)
                (-let* (((&javascript:InlayHint :text :position :kind :whitespace-before? :whitespace-after?) hint)
                        (pos (lsp--position-to-point position))
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -857,8 +857,8 @@ name (e.g. `data' variable passed as `data' parameter)."
                                         (if (and whitespace-before? (not (string= kind lsp/javascript-inlay-hint-kind-type-hint))) " " "")
                                         (propertize (lsp-javascript-format-inlay text kind)
                                                     'font-lock-face (lsp-javascript-face-for-inlay kind))
-                                        (if whitespace-after? " " "")))))))))
-    :mode 'tick))
+                                        (if whitespace-after? " " ""))))))))
+       :mode 'tick)))
 
 (defun lsp-javascript-column-at-pos (pos)
   (save-excursion
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -682,6 +682,11 @@ name (e.g. `data' variable passed as `data' parameter)."
   :type 'boolean
   :package-version '(lsp-mode . "8.0.1"))
 
+(defcustom lsp-javascript-update-inlay-hints-on-scroll t
+  "Update inlay hints immediately when scrolling or modifying window sizes."
+  :type 'boolean
+  :package-version '(lsp-mode . "8.0.1"))
+
 (lsp-register-custom-settings
  '(("javascript.autoClosingTags" lsp-javascript-auto-closing-tags t)
    ("javascript.implicitProjectConfig.checkJs" lsp-javascript-implicit-project-config-check-js t)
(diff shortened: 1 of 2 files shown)


clients/lsp-javascript.el
```diff
@@ -817,28 +822,49 @@ name (e.g. `data' variable passed as `data' parameter)."
...
-                       (overlay (make-overlay pos pos nil 'front-advance 'end-advance)))
-                 (overlay-put overlay 'lsp-javascript-inlay-hint t)
-                 (overlay-put overlay 'before-string
-                              (format "%s%s%s"
-                                      (if (and whitespace-before? (not (string= kind lsp/javascript-inlay-hint-kind-type-hint))) " " "")
-                                      (propertize (lsp-javascript-format-inlay text kind)
-                                                  'font-lock-face (lsp-javascript-face-for-inlay kind))
-                                      (if whitespace-after? " " ""))))))))
+         (let ((hints (lsp-get res :inlayHints)))
+           (when (> (length hints) 0)
```
shivjm, line 844: Open to suggestions for how to phrase this entire `when` block. The idea is to call `overlay-recenter` only if there are hints to add.

lsp-protocol.el
```diff
@@ -425,7 +425,8 @@ See `-let' for a description of the destructuring mechanism."
 (defconst lsp/javascript-inlay-hint-kind-parameter-hint "Parameter")
 (defconst lsp/javascript-inlay-hint-kind-enum-hint "Enum")
 (lsp-interface (javascript:InlayHint (:text :position :kind) (:whitespaceBefore :whitespaceAfter))
-               (javascript:InlayHintsParams (:textDocument) (:range)))
+               (javascript:InlayHintsParams (:textDocument) (:range))
+               (javascript:Range (:start :end)))
```
yyoncho, line 429: there is no need to define that range, the normal range should work fine.

shivjm:
I removed the `window-size-change-functions` function because I kept seeing errors like this in `*Messages*`:
```
Error during redisplay: (lsp-javascript-update-inlay-hints-size-change-function #<window 3 on file.js>) signaled (wrong-type-argument stringp nil)
```
It didn’t seem to affect anything but it kept repeating, and I couldn’t see any reason why. Since the idle hook updates the inlay hints anyway, I figured there wasn’t much to be gained by trying to debug it.

shivjm:
I believe that addresses all the many issues in this small PR, ha. Should be ready to review again.

clients/lsp-javascript.el
```diff
@@ -817,27 +822,48 @@ name (e.g. `data' variable passed as `data' parameter)."
...
-                                      (if whitespace-after? " " ""))))))))
-    :mode 'tick))
+         (let ((hints (lsp-get res :inlayHints)))
+           (when (> (length hints) 0)
+             (overlay-recenter
+              (-let* (([hint] hints)
+                      ((&javascript:InlayHint :position) hint))
+                (lsp--position-to-point position))))
+           (-each hints
+             #'(lambda (hint)
```
yyoncho, line 850: this `#'` is no longer needed in modern elisp.

clients/lsp-javascript.el
yyoncho: looks ok to me, you may use `(unelss (seq-empty-p...`.

yyoncho (commented):
Looks good to me. Some minor issues. 

we might consider implementing the lenses in a similar manner because we also have perf issues there.

jcs090218 (approved):

shivjm:
Great, thanks for the review. I’ve pushed the latest changes. It should be good to go.

> we might consider implementing the lenses in a similar manner because we also have perf issues there.

Re: lenses, if it’s helpful, I could look at lsp-lens to see if I can figure out how to apply the same pattern.

yyoncho:
Thank you!

shivjm:
Thanks!

> That would be great. There are some differences(lenses are more complex because they can be lazy and there might be several lens providers) but I think that the code that places the lenses is compact.

Cool, I’ll take a look at it.

PR #3494, opened 2022-04-28
Link to python poetry + pyenv debugging guide.

samhedin:
Link to DAP-mode tutorial here: https://emacs-lsp.github.io/dap-mode/page/python-poetry-pyenv/
Related to https://github.com/emacs-lsp/dap-mode/pull/608#issuecomment-1112005381

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -157,6 +157,7 @@ nav:
       - 'React JavaScript Tutorial in Emacs': tutorials/reactjs-tutorial.md
       - 'Configuring Emacs as a Clojure IDE': tutorials/clojure-guide.md
       - 'Debugging Clojurescript': tutorials/debugging-clojure-script.md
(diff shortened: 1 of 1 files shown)


mkdocs.yml
```diff
@@ -157,6 +157,7 @@ nav:
       - 'React JavaScript Tutorial in Emacs': tutorials/reactjs-tutorial.md
       - 'Configuring Emacs as a Clojure IDE': tutorials/clojure-guide.md
       - 'Debugging Clojurescript': tutorials/debugging-clojure-script.md
+      - 'Python, debbuging with poetry + pyenv': page/python-poetry-pyenv.md
```
ericdallo, line 160: I'm a little bit confused here, this page doesn't exists on lsp-mode, but only on dap-mode right? I think you want to paste here the full URL link to dap-mode page?

samhedin: You're very correct, I forgot to push the actual URL... Is that better?

ericdallo (approved):

ericdallo:
Thank you!

Issue #3207, opened 2021-11-14
[3.17] Support `labelDetails` on completionItems


ericdallo:
[SPEC](https://microsoft.github.io/language-server-protocol/specifications/specification-3-17/#completionItemLabelDetails)

dgutov:
Seems to have been resolved in the linked PR (in 2022)?

ericdallo:
Yes!

PR #3495, opened 2022-04-29
Add support for a completion item label details

jadestrong:
#3207

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -71,6 +71,12 @@ ignored."
   :type 'boolean
   :group 'lsp-completion)
 
+(defcustom lsp-completion-show-label-description t
+  "Whether or not to show description of completion candidates."
+  :type 'boolean
+  :group 'lsp-completion
+  :package-version '(lsp-mode . "8.0.1"))
+
 (defcustom lsp-completion-no-cache nil
   "Whether or not caching the returned completions from server."
   :type 'boolean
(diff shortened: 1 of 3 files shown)


lsp-completion.el
```diff
@@ -164,10 +170,13 @@ This will help minimize popup flickering issue in `company-mode'."
 
 (defun lsp-completion--annotate (item)
   "Annotate ITEM detail."
-  (-let (((&CompletionItem :detail? :kind?) (plist-get (text-properties-at 0 item)
+  (-let (((&CompletionItem :detail? :kind? :label-details?) (plist-get (text-properties-at 0 item)
                                                        'lsp-completion-item)))
     (concat (when (and lsp-completion-show-detail detail?)
               (concat " " (s-replace "\r" "" detail?)))
+            (when (and lsp-completion-show-label-description label-details?)
+              (when-let ((description (and label-details? (gethash "description" label-details?))))
```
yyoncho, line 178: you should not use gethash since when lsp-mode is compiled with lsp-use-plists = t then label-details? will be plist. You should define a structure and then use either destructuring or the generated getters

jadestrong: Any example for it? I'm not very familiar with lisp. And I don't know how to change it?

yyoncho: Define the structure in lsp-protocol (just like the one you edited) and then use `lsp:<struct-name>-<field-name>` as an accessor.

ericdallo: @yyoncho we could try to documentate lsp-protocol.el to explain all that creates and why we should use it, it's a lsp-mode specific thing that is not that easy for new users get it

yyoncho: yep, we should do that in contribution section.

yyoncho (approved):

yyoncho:
Thank you!

PR #3496, opened 2022-04-29
lsp-rust: files/watcher should be a string

jtracey:
This was causing rust-analyzer to generate a warning for me.


yyoncho (approved):

yyoncho:
Thank you!

PR #3505, opened 2022-05-03
Pass data and kind from code action

timfel:
When a code action comes in and is routed via `lsp--execute-command` to a custom `action-handler`, we want the data and kind fields of the code action to also be passed into the handler, not just the command.


timfel:
I guess I may be misunderstanding something about how the `CodeAction` is pulled apart - how do I ensure that the entire thing with all properties ends up being passed to my action-handler?

PR #3509, opened 2022-05-05
terraform-ls: Support commands for init and validate

psibi:
We implement two functions:
- lsp-terraform-ls-validate
- lsp-terraform-ls-init

The validate is most commonly used to verify the terraform files in the workspace directory.

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -94,7 +94,7 @@
                   :server-id 'tfmls))
 
 (defun lsp-terraform-ls-validate ()
-  "Execute terraform validate on `default-directory'."
+  "Execute terraform validate on project root."
   (interactive)
   (lsp-request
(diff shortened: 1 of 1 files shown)


clients/lsp-terraform.el
```diff
@@ -93,6 +93,29 @@
                   :priority 1
                   :server-id 'tfmls))
 
+(defun lsp-terraform-ls-validate ()
+  "Execute terraform validate on `default-directory'."
```
yyoncho, line 97: the comment is not accurate - you are executing it over project root. `default-directory` is the current directory.

psibi: Ah, it was a comment from the old implementation. Fixed it, thanks!

yyoncho (approved):

yyoncho:
Thank you!

PR #3514, opened 2022-05-09
openscad: better server; support both stdio and tcp connection types

Lenbok:
A new OpenSCAD language server has entered the chat. It is available at:
https://github.com/Leathong/openscad-LSP

The Leathong server requires tcp connection type, and the dzhu server requires stdio connection type. This commit allows the use of either server implementation.

The developer of the earlier language server seems inactive, and the new server has a superset of features, so it seems reasonable to change the default.

### DIFF
--- a/clients/lsp-openscad.el
+++ b/clients/lsp-openscad.el
@@ -29,7 +29,7 @@
 (defgroup lsp-openscad nil
   "LSP support for openscad."
   :group 'lsp-mode
-  :link '(url-link "https://github.com/dzhu/openscad-language-server"))
+  :link '(url-link "https://github.com/Leathong/openscad-LSP"))
 
 (defcustom lsp-openscad-server
   "openscad-language-server"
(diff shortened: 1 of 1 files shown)


ericdallo:
@Lenbok this is kind of a breaking change for this server, right? can we add a second server with a higher priority?

Lenbok:
@ericdallo thanks for looking.  Yeah, it is a breaking change.

The story is that an initial server was implemented last Sept (the one I added the existing support for), but no further development happened. Then it was forked to the new site, and a bunch of development has been happening (extra supported features, and the change in default tcp vs stdio, server binary name etc), so the new one is better in every way. I'm not sure how (or even whether) to maintain support for the old one.

Can you point me at an example of how this should be handled I can use as a template if we want to keep support for the old one?

ericdallo:
@Lenbok , I see, I have no idea how much this server is used, if we think it's not worth to maintain it or users would probably want to use the new one, I'm ok with removing it.

@yyoncho may have some suggestions about that

yyoncho (approved):

Lenbok:
@yyoncho or @ericdallo I think this can be merged now.

yyoncho:
Thank you!

PR #3515, opened 2022-05-09
Gopls use deamon mode by default

cabuda:
Using deamon mode can save a start-up cost each time a session is created.

more: https://github.com/golang/tools/blob/master/gopls/doc/daemon.md

### DIFF
--- a/clients/lsp-go.el
+++ b/clients/lsp-go.el
@@ -48,7 +48,7 @@
   'lsp-go-gopls-server-args
   "lsp-mode 7.0.1")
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):
LGTM. Is this something on by default?

cabuda:
I noticed this feature in vim-go and lsp-bridge

https://github.com/fatih/vim-go/issues/2760#issuecomment-596096981
https://github.com/manateelazycat/lsp-bridge/pull/7/files

PR #3521, opened 2022-05-14
terraform: proper priority and ability to choose a specific server

psibi:
Right now if we both the language server executables are in path, one
of them always wins. The change allows to choose either of them.

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -26,7 +26,13 @@
 
 (require 'lsp-mode)
 
-
+(defcustom lsp-tf-server 'terraform-ls
+  "Choose LSP server for terraform."
+  :type '(choice (const :tag "terraform-lsp" terraform-lsp)
+                 (const :tag "terraform-ls" terraform-ls))
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "8.0.1"))
+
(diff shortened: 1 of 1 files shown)


yyoncho:
The way to solve this is to use lsp-enabled-clients and lsp-disabled-clients. We provide default priorities only as a recommendation.

psibi:
@yyoncho Ah, I see. Follow up question: What should I pass to `lsp-enabled-clients` to choose between terraform-ls and terraform-lsp in this case ? (I will create a follow up documentation PR for the same)

yyoncho:
You should set it like this: `(setq lsp-disabled-clients '(tfls))` to  disable `tfls` server. Refer to this as well: https://emacs-lsp.github.io/lsp-mode/page/faq/#i-have-multiple-language-servers-registered-for-language-foo-which-one-will-be-used-when-opening-a-project

psibi:
Also, I chose this approach because we seem to be doing the same for lsp-rust.el (between rust-analyzer and rls). But given that rust-analyzer has kind of took over - I'm not sure if there is much value in having the support of rls.

psibi:
> You should set it like this: (setq lsp-disabled-clients '(tfls)) to disable tfls server.

Thank you! I'm closing the PR.

PR #3523, opened 2022-05-15
Add vendor folder to lsp-file-watch-ignored-directories

takeokunn:
For programming languages like PHP and Ruby, the libraries are located in the `vendor/` directory.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -338,6 +338,7 @@ the server has requested that."
     "[/\\\\]\\.vscode\\'"
     "[/\\\\]\\.venv\\'"
     "[/\\\\]\\.mypy_cache\\'"
+    "[/\\\\]\\vendor\\'"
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

yyoncho:
`vendor` is too generic, I would rather not put it in the default set.

ericdallo (approved):

elken:
> `vendor` is too generic, I would rather not put it in the default set.

What about per-mode lists? Maybe an alist?

seagle0128:
Agree. It's too generic.

ericdallo:
The per-mode list sounds a good idea to me, would it be hard to implement @yyoncho ?

yyoncho:
We can use after open function to set buffer local or put the specific filters in the `lsp--client`

PR #3525, opened 2022-05-17
Don't set completion-at-point-functions to lsp exclusively (#3340)

### LINKED ISSUE #3340: lsp-completion-mode clobbers completion-functions-at-point

### Bug description

When started, `lsp-completion-mode` sets `completion-functions-at-point` to nil and then adds `lsp-completion-at-point` to the list.  When the server disconnects, `completion-functions-at-point` is set bak to `nil`, erasing the user's customizations.  In order to resture them, the user must reopen the file without connecting lsp.  In addition, this prevents the user form using more complex completion commands, such as [CAPE's `super-cape` functions](https://github.com/minad/cape#super-capf---merging-multiple-capfs).

My suggested solution would have two parts:
- a new defcustom which holds the symbol for the completion function
- using `add-hook` with a negative priority instead of `setq` to add the lsp completion, then removing the hook on disconnect/mode off.

If that sounds like a good idea I can prepare a minor PR.

### Steps to reproduce

turn lsp-completion-mode on and then off again.

### Expected behavior

old completion configuration would still be active.

have tried with several js servers and also with lsp-css. 

### OS

Linux

### Error callstack

_No response_

### Anything else?

thank you!

daanturo:
Fixes #3340

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -761,8 +761,10 @@ The CLEANUP-FN will be called to cleanup."
                                (setq-local lsp-inhibit-lsp-hooks nil))))
     (cond
      (lsp-completion-mode
-      (setq-local completion-at-point-functions nil)
-      (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
+      (make-local-variable 'completion-at-point-functions)
+      ;; Ensure that `lsp-completion-at-point' the first CAPF to be tried,
+      ;; unless user has put it elsewhere in the list by their own
+      (add-to-list 'completion-at-point-functions #'lsp-completion-at-point)
       (make-local-variable 'completion-category-defaults)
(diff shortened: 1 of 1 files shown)


lsp-completion.el
```diff
@@ -761,8 +761,8 @@ The CLEANUP-FN will be called to cleanup."
                                (setq-local lsp-inhibit-lsp-hooks nil))))
     (cond
      (lsp-completion-mode
-      (setq-local completion-at-point-functions nil)
-      (add-hook 'completion-at-point-functions #'lsp-completion-at-point nil t)
+      (make-local-variable 'completion-at-point-functions)
+      (add-to-list 'completion-at-point-functions #'lsp-completion-at-point)
```
kiennq, line 765: The `completion-at-point-functions` is a special hook, can we use `add-hook` here?

kiennq (approved):

drcxd:
When will this be merged?

PR #3535, opened 2022-05-21
terraform: Add semantic token support

psibi:
Summary of the changes:

- Make these variables buffer local:
  - lsp-semantic-token-faces
  - lsp-semantic-token-modifier-faces

The main reason for it is that a particular client will have more semantic tokens associated with them and making them buffer local makes it work seamlessly across different languages.

- Convert the variable lsp-semantic-tokens-capabilities to lsp--semantic-tokens-capabilities function so that the initialization
message to the language server is passed correctly.

- Add complete semantic token support for terraform lanuage. This has been tested using the official Hashicorps terraform language server.

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -25,6 +25,7 @@
 ;;; Code:
 
 (require 'lsp-mode)
+(require 'lsp-semantic-tokens)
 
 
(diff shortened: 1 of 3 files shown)


ericdallo (approved):
LGTM

PR #3537, opened 2022-05-24
terraform: Providers integration using treemacs

psibi:
lsp-treemacs-terraform


psibi:
Also, if lsp-mode is still looking for additional maintainers - I can volunteer to maintain the terraform client since I use it quite heavily.

yyoncho (approved):

PR #3538, opened 2022-05-26
terraform: Integrate terraform modules call

psibi:
lsp_tf_module_snapshot


clients/lsp-terraform.el
```diff
@@ -242,12 +260,30 @@ This is a synchronous action."
...
-         (tf-packages (-zip-with (lambda (x y) (construct-tf-package x y)) providers installed-versions) )
-         )
+         (tf-packages (-zip-with (lambda (x y) (construct-tf-package x y)) providers installed-versions)))
     tf-packages))
 
+(defun lsp-terraform-ls--modules-to-tf-module (modules-tree-data)
+  "Convert MODULES-TREE-DATA to list of `TF-MODULE'."
+  (let* ((json-object-type 'hash-table)
+         (json-key-type 'string)
+         (module-calls (gethash "module_calls" modules-tree-data))
```
yyoncho, line 270: it should not sue gethash because when lsp-mode is compiled with lsp-use-plists = t, then this structure will be plist. So you should use either lsp-get or define define the server interfaces via lsp-interface (check lsp-protocol.el)

psibi: Ah, okay. I will address it for this and the other function. Thanks!

psibi: @yyoncho I have ported it to use the server interfaces via `lsp-interface`.  But do note that I'm still using `gethash` at some places as the object keys returned by the language server are not constant and depends on the code. Example:

```
Result: {
  "installed_providers": {
    "registry.terraform.io/hashicorp/random": "3.1.3",
    "registry.terraform.io/hashicorp/aws": "4.13.0"
  },
  "provider_requirements": {
    "registry.terraform.io/hashicorp/aws": {
      "docs_link": "https://registry.terraform.io/providers/hashicorp/aws/latest",
      "version_constraint": "~> 4.13.0",
      "display_name": "hashicorp/aws"
    }
  },
  "v": 0
}
```

The key `registry.terraform.io/hashicorp/aws"` could be instead `azure` or google cloud depending on what provider a particular persion is using in the code. Given the situation, I'm using `lsp-interface` for the places where it has has fixed constant interface - for the other places I'm still using `gethash`. But do let me know if you think there is a better way to address this.

yyoncho: @psibi can you use lsp-get instead?

psibi: @yyoncho Thanks, using `lsp-get` made me remove the rest of the usages of `gethash`.  (At some places I had to couple it with `make-symbol` to get things working) :-)

clients/lsp-terraform.el
```diff
@@ -222,32 +228,55 @@ This is a synchronous action."
...
 
 (cl-defstruct tf-package display-name doc-link installed-version version-constraint)
 
+(cl-defstruct tf-module name doc-link version source-type dependent-modules)
+
 (defun construct-tf-package (provider installed-version)
   "Construct `TF-PACKAGE' using PROVIDER and INSTALLED-VERSION."
-  (make-tf-package :display-name (gethash "display_name" provider)
-                   :doc-link (gethash "docs_link" provider)
+  (make-tf-package :display-name (lsp-get provider :display_name)
```
yyoncho, line 241: typically here you can use the generated getters/setters or even better use destructoring by utilizing lsp-defun.

yyoncho (approved):

clients/lsp-terraform.el
psibi: I did consider that, but the field `installed-version` comes from a different structure.

But I will keep that in mind when writing future code, thanks!

yyoncho:
@psibi JFYI use merge squash when merging PRs.

psibi:
@yyoncho Thanks, will keep that in mind.

PR #3551, opened 2022-06-06
terraform: Ability to refresh modules widget

psibi:
Also add the corresponding documentation for the same.

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -401,6 +401,8 @@ This is a synchronous action."
 (defun lsp-terraform-ls--modules-refresh ()
   "Refresh terraform modules data."
   (interactive)
+  (unless (buffer-live-p lsp-tf--modules-control-buffer)
+    (error "Original buffer not present.  Do M-x lsp-terraform-ls-module-calls"))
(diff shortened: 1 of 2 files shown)


clients/lsp-terraform.el
```diff
@@ -357,9 +396,15 @@ This is a synchronous action."
...
   (interactive)
   (if (require 'lsp-treemacs nil t)
-      (lsp-terraform-ls--show-module-calls ignore-focus?)
+      (lsp-terraform-ls--show-module-calls ignore-focus? (lsp-workspace-root))
     (error "The package lsp-treemacs is not installed")))
 
+(defun lsp-terraform-ls--modules-refresh ()
+  "Refresh terraform modules data."
+  (interactive)
+  (with-current-buffer lsp-tf--modules-control-buffer
```
yyoncho, line 404: I guess you need a check if the buffer is alive.

yyoncho (approved):

clients/lsp-terraform.el
psibi: Good point, have pushed a fix addressing it. Thanks!

Issue #3190, opened 2021-11-07
Feature request: command to upgrade (or reinstall) all installed LSP servers


peterbecich:
Feature request

> Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? 

I don't believe so

Can all installed LSP servers be upgraded automatically? Can `lsp-mode` provide a command to upgrade or re-install all LSP servers?

Something like
`M-x lsp-upgrade`

A.f.a.i.k. the current best option is to re-install each server individually, i.e. to reinstall the Purescript LSP Server: `M-x lsp-install-server` and select `pursls (Already installed)`.

Thank you

peterbecich:
`lsp-update-server` covers most of this:
https://github.com/emacs-lsp/lsp-mode/blob/828c6c234a3231670ba5649666652e22a9bc4f35/lsp-mode.el#L7545-L7564

ericdallo:
Yeah, we already have `lsp-update-server` as a public command, maybe we could have `lsp-update-all-servers`?

peterbecich:
This may not work correctly: https://github.com/emacs-lsp/lsp-mode/issues/3754

PR #3572, opened 2022-06-12
`lsp-update-servers` updates all installed servers

peterbecich:
I have tested this and believe it works correctly.

The filter logic is copied from `lsp-update-server` above.

Closes https://github.com/emacs-lsp/lsp-mode/issues/3190

One downside is that many `*lsp-install:....` buffers are created.

Thanks

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7634,6 +7634,19 @@ When prefix UPDATE? is t force installation even if the server is present."
     (lsp--install-server-internal chosen-client t)))
 
 ;;;###autoload
+(defun lsp-update-servers ()
+  "Update all installed servers."
+  (interactive)
+  (lsp--require-packages)
+  (cl-loop for client in
+           (-filter (-andfn
+                     (-not #'lsp--client-download-in-progress?)
+                     #'lsp--client-download-server-fn
+                     #'lsp--server-binary-present?) (hash-table-values lsp-clients))
+           do (lsp--install-server-internal client t))
+  )
+
+;;;###autoload
 (defun lsp-ensure-server (server-id)
   "Ensure server SERVER-ID"
   (lsp--require-packages)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7614,7 +7614,7 @@ When prefix UPDATE? is t force installation even if the server is present."
 
 ;;;###autoload
 (defun lsp-update-server (&optional server-id)
-  "Interactively update a server."
+  "Interactively update (reinstall) a server."
   (interactive)
   (lsp--require-packages)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -7633,6 +7633,19 @@ When prefix UPDATE? is t force installation even if the server is present."
...
+(defun lsp-update-servers ()
+  "Update all installed servers."
+  (interactive)
+  (lsp--require-packages)
+  (cl-loop for client in
+           (-filter (-andfn
+                     (-not #'lsp--client-download-in-progress?)
+                     #'lsp--client-download-server-fn
+                     #'lsp--server-binary-present?) (hash-table-values lsp-clients))
+           do (lsp--install-server-internal client t))
```
yyoncho, line 7646: please close it on the same line

lsp-mode.el
```diff
@@ -7633,6 +7633,19 @@ When prefix UPDATE? is t force installation even if the server is present."
                             t))))
     (lsp--install-server-internal chosen-client t)))
 
+;;;###autoload
+(defun lsp-update-servers ()
+  "Update all installed servers."
+  (interactive)
+  (lsp--require-packages)
+  (cl-loop for client in
```
yyoncho, line 7641: consider using `mapc` here - it is a bit more coherent - e. g. it makes more sense to combine -filter with mapc because cl-loop has its own filtering facilities.

yyoncho (commented):
Looks good! Two minor nits.

lsp-mode.el
peterbecich: done

lsp-mode.el
peterbecich: done

lsp-mode.el
```diff
@@ -7614,7 +7614,7 @@ When prefix UPDATE? is t force installation even if the server is present."
 
 ;;;###autoload
 (defun lsp-update-server (&optional server-id)
-  "Interactively update a server."
+  "Interactively update (reinstall) a server."
```
peterbecich, line 7617: `lsp-update-server` and `lsp-update-servers` reinstall all servers, right? Even if already up-to-date?

lsp-mode.el
peterbecich: thanks, tested with `mapc`; it works

lsp-mode.el
yyoncho: yes, ATM we don't have a mechanism to check if it is up to date.

yyoncho (approved):

peterbecich:
Can this be merged?

PR #3582, opened 2022-06-17
refactor lsp-elixir to be activated on "elixir"

elsifaka:
If you create a derived-mode, even if you add the new major-mode to `lsp-language-id-configuration`, `lsp` is not started.

It works if you activate on "elixir"


yyoncho:
Thank you!

PR #3594, opened 2022-06-28
Fix test errors

devmotion:
This PR should fix the test errors on the master branch that I observed in #3593.

It seems the problematic function was added in https://github.com/emacs-lsp/lsp-mode/commit/92c1f08c7bdb806f00d37ec6d0bf2c05d3995d44. https://github.com/emacs-lsp/lsp-mode/commit/c44cd787a9164df683ac21ece2c05d21cb3deda7 was supposed to fix the error but I think one should have removed a different bracket, namely the last one in line 843 instead.


yyoncho:
@brotzeit

brotzeit:
Thanks!

PR #3595, opened 2022-06-28
Fix docstrings that are too wide (> 80 characters)

devmotion:
This PR fixes a docstring that is too wide which should fix (at least one of) the test errors on Emacs 28. Currently, these tests fail since a docstring is too wide (see e.g. https://github.com/emacs-lsp/lsp-mode/pull/3594#issuecomment-1168381449).

Edit: Seems there are some more docstrings, so I might have to add some more commits.


jcs090218 (approved):

yyoncho (approved):

devmotion:
Seems I finally managed to find all docstrings that were too wide, tests on Emacs 28.1 pass now :slightly_smiling_face:

yyoncho:
Thank you!

PR #3597, opened 2022-06-29
lsp-rust: set UPDATE_EXPECT=1 when expect-test on runnable

alanz:
The rust-analyzer `experimental/runnables` LSP extension includes an optional flag `expectTest`

https://github.com/rust-lang/vscode-rust/blob/b1ae67b06640ffab6e1ebb72e07364b4477dfbf1/rust-analyzer/editors/code/src/lsp_ext.ts#L59-L68

This is set when the cursor is on an `expect!` or `expect_file!` macro when the `lsp-rust-analyzer-run` command is invoked.

Update the `lsp-rust` client to include `env UPDATE_EXPECT=1` before the corgo command when this is set, so `expect` automatically updates the expected result to match actual.

image

### DIFF
--- a/lsp-protocol.el
+++ b/lsp-protocol.el
@@ -414,7 +414,7 @@ See `-let' for a description of the destructuring mechanism."
                (rust-analyzer:MoveItemParams (:textDocument :range :direction) nil)
                (rust-analyzer:RunnablesParams (:textDocument) (:position))
                (rust-analyzer:Runnable (:label :kind :args) (:location))
-               (rust-analyzer:RunnableArgs (:cargoArgs :executableArgs) (:workspaceRoot))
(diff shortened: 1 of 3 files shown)


lsp-protocol.el
```diff
@@ -414,7 +414,7 @@ See `-let' for a description of the destructuring mechanism."
                (rust-analyzer:MoveItemParams (:textDocument :range :direction) nil)
                (rust-analyzer:RunnablesParams (:textDocument) (:position))
                (rust-analyzer:Runnable (:label :kind :args) (:location))
-               (rust-analyzer:RunnableArgs (:cargoArgs :executableArgs) (:workspaceRoot))
+               (rust-analyzer:RunnableArgs (:cargoArgs :executableArgs) (:workspaceRoot) (:expectTest))
```
yyoncho: this should go near :workspaceRoot (thus the CI failure)

yyoncho (approved):

PR #3598, opened 2022-06-30
lsp-rust: fix bad cond case in inlay-hints

gagbo:
I think this is a small copy-paste that got through the net, found this out while trying to understand my other issue around plists


ericdallo (approved):

ericdallo:
Thanks!

PR #3600, opened 2022-07-01
Download aarch64 version of rust-analyzer for ARM based Macs

cedarbaum:
I was getting a lot of errors using the `x86_64` version of the `rust-analyzer` binary on a Mac using an M1 chip. Changing the download to the `aarch64` version fixed this.


yyoncho (approved):

yyoncho:
Thank you!

PR #3605, opened 2022-07-05
fix: only run if lib dir exists

elken:
Windows might need a proper fix, but for now this should attempt to resolve it.

@jcs090218


jcs090218:
Seems to be working!? I have tested in https://github.com/emacs-grammarly/lsp-grammarly/pull/39.

elken:
Per the discussion [here](https://github.com/emacs-lsp/lsp-mode/commit/c0d17320b86381da3d88903c91689934016497fc#r77701775) it looks like Windows will need custom behaviour...

Can you verify this is the case for everything or just your grammarly package?

jcs090218:
I have these language server installed, and all `lib` folders are empty. I guess it's the standard in Windows? 😕 

Image 2

I think the `lib` folder is created in L7754?

jcs090218 (approved):

elken:
@jcs090218 can you confirm the below snippet produces the correct path? Replace `package` as needed

```lisp
(let* ((package "pyright")
       (default-directory (f-join lsp-server-install-dir "npm" package "lib" "node_modules" package)))
                                       (when (f-dir-p default-directory)
                                         (last (directory-files-recursively default-directory "package.json"))))
```

jcs090218:
This works for me:

```elisp
(let* ((package "pyright")
       (default-directory (f-join lsp-server-install-dir "npm" package "node_modules" package)))
  (when (f-dir-p default-directory)
    (last (directory-files-recursively default-directory "package.json"))
    ))
```

output:

```
(c:/path/to/home/.emacs.d/.cache/lsp/npm/pyright/node_modules/pyright/package.json)
```

If your remove `lib`.

elken:
That's what I meant to do, cool thanks :D

elken:
@jcs090218 does this still work for you? please try installing graphql also and make sure that works :)

jcs090218:
I have made another test and it works! Just want to confirm this is the code your want me to test?

```elisp
(cl-defun lsp--npm-dependency-install (callback error-callback &key package &allow-other-keys)
  (if-let ((npm-binary (executable-find "npm")))
      (progn
        ;; Explicitly `make-directory' to work around NPM bug in
        ;; versions 7.0.0 through 7.4.1. See
        ;; https://github.com/emacs-lsp/lsp-mode/issues/2364 for
        ;; discussion.
        (make-directory (f-join lsp-server-install-dir "npm" package "lib") 'parents)
        (lsp-async-start-process (lambda ()
                                   (if (string-empty-p
                                        (string-trim (shell-command-to-string
                                                      (mapconcat #'shell-quote-argument `(,npm-binary "view" ,package "peerDependencies") " "))))
                                       callback
                                     (let ((default-directory (f-dirname (car (last (directory-files-recursively (f-join lsp-server-install-dir "npm" package) "package.json"))))))
                                       (when (f-dir-p default-directory)
                                         (lsp-async-start-process callback
                                                                  error-callback
                                                                  (executable-find "npx")
                                                                  "npm-install-peers")))))
                                 error-callback
                                 npm-binary
                                 "-g"
                                 "--prefix"
                                 (f-join lsp-server-install-dir "npm" package)
                                 "install"
                                 package))
    (lsp-log "Unable to install %s via `npm' because it is not present" package)
    nil))
```

I don't about the graphql, maybe we should have a CI test like what [lsp-grammarly/test/activate.el](https://github.com/emacs-grammarly/lsp-grammarly/blob/master/test/activate.el) did?

elken:
Hm, yeah adding a test is a good call.

Wonder if all clients should have tests? Probably a discussion for another thread

jcs090218:
The current CI tests seem to be very fragile (a lot of weird things going on in the ert tests), and Cask doesn't seem to work with Windows anymore.

> Wonder if all clients should have tests? Probably a discussion for another thread

But answering your question, it would be nice to have CI tests for all clients if we want this package to be stable. 😅

elken:
I am _really_ close to a working test, but yeah you're right; there should be a defined standard.

elken:
> Discuss CI tests in another thread should be good. Leave this PR clean.

https://github.com/emacs-lsp/lsp-mode/discussions/3608

elken:
@jcs090218 before I merge, just clarify that this branch does indeed work for you?

jcs090218:
Just in case, I have re-run the test.

Yes, it works!

PR #3609, opened 2022-07-05
fix: run npx without prompting

elken:


yyoncho (approved):

jcs090218 (approved):

lsp-mode.el
```diff
@@ -7762,6 +7762,7 @@ nil."
                                          (lsp-async-start-process callback
                                                                   error-callback
                                                                   (executable-find "npx")
+                                                                  "--yes"
```
dlineberger, line 7765: This flag does not exist in node v14. I think it was introduced in v16. This currently breaks installing language servers in my environment.

elken: Please submit an issue rather than commenting on a PR...

PR #3611, opened 2022-07-05
Avoid invalid face reference errors

### LINKED ISSUE #2255: Invalid face reference: lsp-flycheck-info-unnecessary
``` python
def hello():
    def abc(args):
        print(42)

    a = 1
    print("hello world")
    a += 1
    b = 5
    c = 10
    a = [1, 2]

    d = 5

    print("hello ")
    abc(1)


hello()
print("print statement")
```

Using `lsp-pyright` with the above Python file, there are numerous Flycheck errors classified as `lsp-flycheck-info-unnecessary`. When `point` is on a line with one such error, `Invalid face reference: lsp-flycheck-info-unnecessary` appears in the `Messages` buffer. This message seems to be called by `merge_face_ref` in the `xfaces.c` file in core Emacs.

The following seems to make the message go away:

``` diff
---
 lsp-diagnostics.el | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)

diff --git a/lsp-diagnostics.el b/lsp-diagnostics.el
index b7aa5daa..ea6d2113 100644
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -93,7 +93,7 @@ g. `error', `warning') and list of LSP TAGS."
                       flycheck-level
                       (mapconcat #'symbol-name tags "-"))))
     (or (intern-soft name)
-        (let* ((face (--doto (intern (format "lsp-%s-face" name))
+        (let* ((face (--doto (intern name)
                        (copy-face (-> flycheck-level
                                       (get 'flycheck-overlay-category)
                                       (get 'face))
@@ -102,7 +102,7 @@ g. `error', `warning') and list of LSP TAGS."
                                (apply #'set-face-attribute it nil
                                       (cl-rest (assoc tag lsp-diagnostics-attributes))))
                              tags)))
-               (category (--doto (intern (format "lsp-%s-category" name))
+               (category (--doto (intern (format "%s-category" name))
                            (setf (get it 'face) face
                                  (get it 'priority) 100)))
                (new-level (intern name))
-- 
2.28.0
```

Apparently, making the `face` name match `name` does the trick, but I'm not sure why. The error levels that Flycheck defines [don't look like this](https://github.com/flycheck/flycheck/blob/cd8e0a280c9980c8c7ce31fd2458df7fd81a0acf/flycheck.el#L4388). It seems weird to define some symbol that has not only the properties assigned to it by `flycheck-define-error-level` but also is the name of a face, so we should probably find a cleaner solution.

I am not able to build Emacs from source on my machine for some reason, so I can't go into `xfaces.c` to debug this.

indigoviolet:
Fixes #2255

I have no idea of the extended impact of this - I'm just making a PR that @yyoncho can easily merge if they so choose

### DIFF
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -105,7 +105,7 @@ g. `error', `warning') and list of LSP TAGS."
                       flycheck-level
                       (mapconcat #'symbol-name tags "-"))))
     (or (intern-soft name)
-        (let* ((face (--doto (intern (format "lsp-%s-face" name))
+        (let* ((face (--doto (intern (format "%s-face" name))
                        (copy-face (-> flycheck-level
                                       (get 'flycheck-overlay-category)
                                       (get 'face))
@@ -114,7 +114,7 @@ g. `error', `warning') and list of LSP TAGS."
                                (apply #'set-face-attribute it nil
                                       (cl-rest (assoc tag lsp-diagnostics-attributes))))
                              tags)))
-               (category (--doto (intern (format "lsp-%s-category" name))
+               (category (--doto (intern (format "%s-category" name))
                            (setf (get it 'face) face
                                  (get it 'priority) 100)))
                (new-level (intern name))


ericdallo (approved):
I found the same and fixed locally and forget about opening a PR, thanks!

PR #3616, opened 2022-07-07
lsp-ansible: add command to resync inventory

fredericgiquel:
The last version of Ansible Language Server add a notification event to resync the inventory cache (see https://github.com/ansible/ansible-language-server/releases/tag/v0.9.0-next.0 and https://github.com/ansible/ansible-language-server/pull/350).

This PR makes it easy for the user to send this notification with `lsp-mode`.

### DIFF
--- a/clients/lsp-ansible.el
+++ b/clients/lsp-ansible.el
@@ -25,7 +25,6 @@
 ;;; Code:
 
 (require 'lsp-mode)
-(require 'lsp-completion)
 
 ;;; Ansible
 (defgroup lsp-ansible nil
@@ -200,6 +199,7 @@ This prevents the Ansible server from being turned on in all yaml files."
   "Resync the inventory cache used by Ansible Language Server for hosts completion."
   (interactive)
   (lsp-notify "resync/ansible-inventory" nil)
+  (require 'lsp-completion)
   (lsp-completion--clear-cache))
 
 (lsp-register-client
--- a/clients/lsp-ansible.el
+++ b/clients/lsp-ansible.el
@@ -195,6 +195,8 @@ This prevents the Ansible server from being turned on in all yaml files."
        ;; emacs-ansible provides ansible, not ansible-mode
        (with-no-warnings (bound-and-true-p ansible))))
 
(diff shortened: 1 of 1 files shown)


clients/lsp-ansible.el
```diff
@@ -195,6 +196,12 @@ This prevents the Ansible server from being turned on in all yaml files."
        ;; emacs-ansible provides ansible, not ansible-mode
        (with-no-warnings (bound-and-true-p ansible))))
 
+(defun lsp-ansible-resync-inventory ()
+  "Resync the inventory cache used by Ansible Language Server for hosts completion."
+  (interactive)
+  (lsp-notify "resync/ansible-inventory" nil)
+  (lsp-completion--clear-cache))
```
yyoncho, line 203: is this really needed? This will be called automatically when needed.

fredericgiquel: There is no auto sync done by the server. I think it's because it's difficult to known when to sync if a dynamic inventory is used.

A similar command has already been added to the vscode client: https://github.com/ansible/vscode-ansible/pull/522/files#diff-04bba6a35cad1c794cbbe677678a51de13441b7a6ee8592b7b50be1f05c6f626

yyoncho: I mean that (lsp-completion--clear-cache) should not be needed

fredericgiquel: I found it useful in the following case:
1. I trigger a completion for hosts a first time
2. I notice that the inventory cache is outdate
3. so I notify the server to resync the inventory
4. I trigger the same completion as in step 1

And I get the same completion as the first time because it comes from the cache.

yyoncho: in that case can you move the require inside the function? If it is on top-level lsp-completion will be loaded unconditionally

yyoncho (approved):

clients/lsp-ansible.el
yyoncho: apparently CI is now happy with this code, can you add `declare-function` (there are several examples in the repo how to do that)

yyoncho:
Thank you!

PR #3619, opened 2022-07-09
fix: ClojureScript Debugging tutorial: support Chromium

allentiak:

PR #3620, opened 2022-07-09
Add Support for Spacemacs

allentiak:


docs/tutorials/clojure-guide.md
```diff
@@ -137,6 +137,38 @@ You may need to remove the lookup handlers conflict from `cider` and/or `clj-ref
...
+```elisp
+dotspacemacs/user-config()
+  ...
+  ;; Enable DAP debugging in ClojureScript.
+  ;; Trying almost all options...
+
+  ;; This one downloads OK.
+  (require 'dap-chrome)
+
+  ;; These two cannot be downloaded...
```
yyoncho, line 159: these cannot be downloaded when you hit DOS protection of the marketplace

allentiak: Hmmm...

Have you considered using open-vsx.org[1] as a source instead of VS Marketplace? It's a libre alternative to VS Marketplace...
Plus, VS Marketplace can only be legally used by Microsoft products [2]...

[1] https://open-vsx.org/about
[2] https://github.com/VSCodium/vscodium/#extensions-and-the-marketplace

yyoncho: @allentiak I was not aware of this option. Switching to it seems like a good idea

yyoncho (approved):

allentiak:
Related to #3619.

PR #3508, opened 2022-05-05
Add Gleam support

J3RN:
Adds support for the [recently announced Gleam language server](https://gleam.run/news/v0.21-introducing-the-gleam-language-server/).

PR #3632, opened 2022-07-21
Add client configuration for Marksman

alanz:
https://github.com/artempyanykh/marksman

Emacs config:

    (use-package markdown-mode
      :after lsp-mode
      :hook (markdown-mode . lsp)
      :config
      (require 'lsp-marksman))

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -177,7 +177,7 @@ As defined by the Language Server Protocol 3.16."
          lsp-cmake lsp-crystal lsp-csharp lsp-css lsp-d lsp-dart lsp-dhall lsp-docker lsp-dockerfile
          lsp-elm lsp-elixir lsp-emmet lsp-erlang lsp-eslint lsp-fortran lsp-fsharp lsp-gdscript lsp-go
          lsp-gleam lsp-graphql lsp-hack lsp-grammarly lsp-groovy lsp-haskell lsp-haxe lsp-idris lsp-java lsp-javascript
-         lsp-json lsp-kotlin lsp-latex lsp-ltex lsp-lua lsp-markdown lsp-nginx lsp-nim lsp-nix lsp-magik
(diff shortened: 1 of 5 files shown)


alanz:
cc @artempyanykh

yyoncho:
can you follow this PR https://github.com/emacs-lsp/lsp-mode/pull/3508/ and add the documentation stuff?

lsp-mode.el
```diff
@@ -841,7 +841,8 @@ Changes take effect only when a new session is started."
                                         (magik-mode . "magik")
                                         (idris-mode . "idris")
                                         (idris2-mode . "idris2")
-                                        (gleam-mode . "gleam"))
+                                        (gleam-mode . "gleam")
+                                        (markdown-mode . "marksman"))
```
yyoncho, line 845: This change is not needed - markdown-mode is mapped to `markdown`.

alanz: I was wondering about that.  Thanks.

yyoncho (approved):

artempyanykh:
Wow! Thanks for putting this together @alanz! 💖

artempyanykh:
I found these settings to be useful when working with markdown:
```lisp
(setq-local completion-ignore-case t)
(setq-local lsp-imenu-sort-methods '(kind position name))
```

@yyoncho is there a way to advertise those for marksman LSP? Or perhaps lsp-mode doesn't want to mess with server-specific settings?

yyoncho:
@artempyanykh we do encourage server-specific settings but these seem to be a bit more opinionated. But still, if you want you can create a custom documentation page and add these as a suggestion. If we change these defaults then we need a way to turn off this customization and it ends up very unconventional and not very user-friendly.

artempyanykh:
@yyoncho thanks, yes, this makes sense!

PR #3637, opened 2022-07-24
Allow to not cut the signature received on hover

mattiasdrp:
Some servers (i.e. the ocamllsp one) return only the signature on hover (without the documentation) but tend to put newlines between the different items. With the current implementation of `lsp-clients-extract-signature-on-hover`, the displayed signature will only contain the first item

```ocaml
config1:string list ->
config2:string list -> config3:string list -> int -> int -> int
```

will then be rendered as

```ocaml
config1:string list ->
```

With the new implementation, a custom variable allows to take control over this behaviour and the function has been reimplemented to either return the previous result or return the signature with newlines replaced by spaces

### DIFF
--- a/clients/lsp-ocaml.el
+++ b/clients/lsp-ocaml.el
@@ -77,6 +77,27 @@
   :priority 0
   :server-id 'ocaml-lsp-server))
 
+(defcustom lsp-cut-signature 'space
+  "If non-nil, signatures returned on hover will not be split on newline"
+  :group 'lsp-ocaml
+  :type '(choice (symbol :tag "Default behaviour" 'cut)
+                 (symbol :tag "Display all the lines with spaces" 'space)))
+
+(cl-defmethod lsp-clients-extract-signature-on-hover (contents (_server-id (eql ocamllsp)) &optional storable)
+  "Extract a representative line from OCaml's CONTENTS, to show in the echo area.
+This function splits the content between the signature
+and the documentation to display the signature"
+  (let ((type (lsp-make-marked-string
+               :language "ocaml"
+               :value (car (s-split "---" (lsp--render-element contents))))))
+      (if (eq lsp-cut-signature 'cut)
+          (car (s-lines (s-trim (lsp--render-element type))))
+        (if (and (eq lsp-cut-signature 'space) (equal nil storable))
+            (let* ((ntype (s-replace "\n" " " (s-trim (lsp--render-element type)))))
+              (if (>= (length ntype) (frame-width))
+                  (concat (substring ntype 0 (- (frame-width) 4)) "...")
+                ntype))
+        (s-trim (lsp--render-element type))))))
 
 (lsp-consistency-check lsp-ocaml)
 
--- a/clients/lsp-ocaml.el
+++ b/clients/lsp-ocaml.el
@@ -78,7 +78,7 @@
   :server-id 'ocaml-lsp-server))
 
 (defcustom lsp-cut-signature 'space
-  "If non-nil, signatures returned on hover will not be split on newline"
+  "If non-nil, signatures returned on hover will not be split on newline."
   :group 'lsp-ocaml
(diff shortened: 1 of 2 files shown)


mattiasdrp:
I need to check some things, don't review it right now ;-)

yyoncho:
you may do it like this: https://github.com/yyoncho/lsp-mode/blob/master/clients/lsp-clangd.el#L267

mattiasdrp:
Thanks a lot @yyoncho for the pointer, that's much better for me!

clients/lsp-ocaml.el
```diff
@@ -77,6 +77,27 @@
...
+(cl-defmethod lsp-clients-extract-signature-on-hover (contents (_server-id (eql ocamllsp)) &optional storable)
+  "Extract a representative line from OCaml's CONTENTS, to show in the echo area.
+This function splits the content between the signature
+and the documentation to display the signature"
+  (let ((type (lsp-make-marked-string
+               :language "ocaml"
+               :value (car (s-split "---" (lsp--render-element contents))))))
+      (if (eq lsp-cut-signature 'cut)
+          (car (s-lines (s-trim (lsp--render-element type))))
+        (if (and (eq lsp-cut-signature 'space) (equal nil storable))
```
yyoncho, line 95: `(equal nil storable)` -> `(not storable)`

clients/lsp-ocaml.el
```diff
@@ -77,6 +77,57 @@
   :priority 0
   :server-id 'ocaml-lsp-server))
 
+(defcustom lsp-cut-signature 'space
```
ericdallo, line 80: Wouldn't this change the current behavior for most lsps?
For example for clojure-lsp, server return the first line as the function signature and the rest as the docs and other things, and it's intended to be shown only the first line on messages buffer

mattiasdrp: Since I added it to the lsp-ocaml client I thought it wouldn't have any impact on the rest of the clients. Am I wrong about it?

clients/lsp-ocaml.el
```diff
@@ -77,6 +77,57 @@
...
+  "Extract a representative line from OCaml's CONTENTS, to show in the echo area.
+This function splits the content between the signature
+and the documentation to display the signature
+and truncate it if it's too wide.
+The STORABLE argument is used if you want to use this
+function to get the type and, for example, kill and yank it.
+
+An example of function using STORABLE is:
+
+  (defun mdrp/lsp-get-type-and-kill ()
```
yyoncho, line 96: you can call lsp-describe-thing-at-point and then copy the buffer content.

clients/lsp-ocaml.el
```diff
@@ -77,6 +77,57 @@
...
+                 (lsp-clients-extract-signature-on-hover
+                   contents
+                   nil)
+                 )))
+          (message \"Copied %s to kill-ring\" contents)
+          (kill-new contents)))))"
+  (let ((type (s-trim (lsp--render-element (lsp-make-marked-string
+                                            :language "ocaml"
+                                            :value (car (s-split "---" (lsp--render-element contents))))))))
+    (if (equal nil storable)
```
yyoncho, line 122: nit: if (not storable)

mattiasdrp: Will update tomorrow ;-)

mattiasdrp:
> Can you confirm that this is still relevant?

Still relevant, yes, rewrote the function in my .emacs.d until this is merged :-)

clients/lsp-ocaml.el
ericdallo: oh, yes, so I think it'd be better to name it as `lsp-ocaml-cut-signature` or something

yyoncho (approved):

yyoncho:
Thank you for contributing to lsp-mode, and sorry for the delay.

Issue #3638, opened 2022-07-25
Feature Request: Download `lua-language-server` directly from its GitHub page


Mateus-Carmo31:
**Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)?**
No.
**Describe your feature here.**
Currently, the lsp-mode configurations for sumneko's `lua-language-server` are hardcoded to use a specific version of the server, downloaded from VSCode's marketplace (version 1.17.4). The server has received many updates since then, adding many useful features such as per-project configuration, and is now at version 3.5.0. PR #3636 aims to address this version issue, but downloading the server from microsoft causes many additional, vscode extension specific files to be downloaded with it that serve no purpose for lsp-mode.

So I thought to try and make another client configuration that would fetch the latest release of the server (OS-appropriate) using GitHub's API. The client would be separate from the vscode one for compatibility, and would provide appropriate settings for the lsp startup commands (as sumneko has changed the language server release's inner file layouts a bit since 1.17.4).

PR #3636, opened 2022-07-23
bump: lua-language-server

basaran:
bump: lua-roblox-language-server

fix: lua-language-server download url

close #3635 3635

PR #3639, opened 2022-07-25
lua-language-server: fetch latest version from github

Mateus-Carmo31:
This is my implementation of the feature I specified in #3638. I am not quite adept with elisp yet, so please tell me if I've done any mistakes.

### DIFF
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -606,11 +606,9 @@ and `../lib` ,exclude `../lib/temp`.
       (goto-char (point-max))
       (goto-char (point-at-bol))
       (let* ((json-result (prog1 (json-read) (kill-buffer)))
-             (os-search-pred (lambda (json-entry)
-                      (string-match-p system-suffix (alist-get 'name json-entry)))))
+             (os-search-pred (lambda (json-entry) (string-match-p system-suffix (alist-get 'name json-entry)))))
         (message "Latest version found: %s" (alist-get 'name json-result))
-        (alist-get 'browser_download_url (seq-find os-search-pred (alist-get 'assets json-result)))
-        ))))
+        (alist-get 'browser_download_url (seq-find os-search-pred (alist-get 'assets json-result)))))))
 
 (defun lsp-lua-language-server-install-latest (client callback error-callback update?)
   "Download the latest version of lua-language-server and extract it to
@@ -632,7 +630,7 @@ and `../lib` ,exclude `../lib/temp`.
                                                        `(,lsp-clients-lua-language-server-latest-bin
                                                          ,@lsp-clients-lua-language-server-latest-args
                                                          ,lsp-clients-lua-language-server-latest-main-location)))
-                                        '(lambda () (and (f-exists? lsp-clients-lua-language-server-latest-bin) (f-exists? lsp-clients-lua-language-server-latest-main-location))))
+                                        (lambda () (and (f-exists? lsp-clients-lua-language-server-latest-bin) (f-exists? lsp-clients-lua-language-server-latest-main-location))))
   :major-modes '(lua-mode)
   :priority -2
   :server-id 'lua-language-server-latest
--- a/clients/lsp-lua.el
+++ b/clients/lsp-lua.el
@@ -590,26 +590,6 @@ and `../lib` ,exclude `../lib/temp`.
   :risky t
   :type '(repeat string))
 
-(defun lsp--find-latest-lua-ls-release-url ()
-  "Fetch the latest version of lua-language-server from its github repo
-(for use with `lsp-lua-language-server-install'). Search for \"linux-x64\",
-\"darwin-x64\" and \"win32-x64\" tags on releases."
-  (let ((system-suffix (pcase system-type
-                         ('gnu/linux "linux-x64")
-                         ('darwin "darwin-x64")
-                         ('windows-nt "win32-x64")
-                         (_ "linux-x64")))
-        (url-request-method "GET"))
-    (with-current-buffer (url-retrieve-synchronously "https://api.github.com/repos/sumneko/lua-language-server/releases/latest")
-      ;; Go to beginning of the last line (where the retrieved json with information
-      ;; is in the buffer from `url-retrieve-synchronously')
-      (goto-char (point-max))
-      (goto-char (point-at-bol))
-      (let* ((json-result (prog1 (json-read) (kill-buffer)))
-             (os-search-pred (lambda (json-entry) (string-match-p system-suffix (alist-get 'name json-entry)))))
-        (message "Latest version found: %s" (alist-get 'name json-result))
-        (alist-get 'browser_download_url (seq-find os-search-pred (alist-get 'assets json-result)))))))
-
 (defun lsp-lua-language-server-install-latest (client callback error-callback update?)
   "Download the latest version of lua-language-server and extract it to
 `lsp-lua-language-server-install-dir'."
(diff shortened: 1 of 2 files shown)


clients/lsp-lua.el
```diff
@@ -556,6 +556,87 @@ and `../lib` ,exclude `../lib/temp`.
...
+    (with-current-buffer (url-retrieve-synchronously "https://api.github.com/repos/sumneko/lua-language-server/releases/latest")
+      ;; Go to beginning of the last line (where the retrieved json with information
+      ;; is in the buffer from `url-retrieve-synchronously')
+      (goto-char (point-max))
+      (goto-char (point-at-bol))
+      (let* ((json-result (prog1 (json-read) (kill-buffer)))
+             (os-search-pred (lambda (json-entry)
+                      (string-match-p system-suffix (alist-get 'name json-entry)))))
+        (message "Latest version found: %s" (alist-get 'name json-result))
+        (alist-get 'browser_download_url (seq-find os-search-pred (alist-get 'assets json-result)))
```
yyoncho, line 612: nit: please close on the same line.

clients/lsp-lua.el
```diff
@@ -556,6 +556,87 @@ and `../lib` ,exclude `../lib/temp`.
...
+     :store-path store-path
+     :decompress (pcase system-type ('windows-nt :zip) (_ :targz)))))
+
+(lsp-register-client
+ (make-lsp-client
+  :new-connection (lsp-stdio-connection (lambda () (or lsp-clients-lua-language-server-command
+                                                       `(,lsp-clients-lua-language-server-latest-bin
+                                                         ,@lsp-clients-lua-language-server-latest-args
+                                                         ,lsp-clients-lua-language-server-latest-main-location)))
+                                        '(lambda () (and (f-exists? lsp-clients-lua-language-server-latest-bin) (f-exists? lsp-clients-lua-language-server-latest-main-location))))
```
yyoncho, line 635: nit: in latest version of elisp you should not quote lambdas

clients/lsp-lua.el
```diff
@@ -556,6 +556,87 @@ and `../lib` ,exclude `../lib/temp`.
...
+  :major-modes '(lua-mode)
+  :priority -2
+  :server-id 'lua-language-server-latest
```
yyoncho, line 638: can you modify the existing client instead of creating a new one? You can try first the existing variables and if they are not present.

lsp-mode.el
```diff
@@ -7905,6 +7907,26 @@ in place."
     (error "Unable to find `gzip' on the path, please either customize `lsp-gunzip-script' or manually decompress %s" gz-file))
   (shell-command (format lsp-gunzip-script gz-file)))
 
+;; tar.gz decompression
+
+(defconst lsp-ext-tar-script "bash -c 'mkdir -p %2$s; tar xf %1$s --directory=%2$s'"
+  "Script to decompress a .tar.gz file.")
+
+(defcustom lsp-tar-script (cond ((executable-find "tar") lsp-ext-tar-script)
```
yyoncho, line 7915: I missed that in the original PR. We should not run executable-find on the top level. We could use a function and `lsp-resolve-value` to make the execution lazy.

clients/lsp-lua.el
```diff
@@ -556,6 +556,87 @@ and `../lib` ,exclude `../lib/temp`.
...
+  :risky t
+  :type 'file)
+
+(defcustom lsp-clients-lua-language-server-latest-args '("-E")
+  "Arguments to run the Lua Language server (latest)."
+  :group 'lsp-lua-language-server-latest
+  :risky t
+  :type '(repeat string))
+
+(defun lsp--find-latest-lua-ls-release-url ()
```
yyoncho, line 593: can you change all variables and functions to be prefixed with `lsp-lua-`. All that are not prefixed like that are due to the fact that initially all clients were in single file `lsp-clients`

yyoncho (commented):
Looks good, few minor nits.

yyoncho:
There is also https://github.com/emacs-lsp/lsp-mode/pull/3636/

cc @basaran

clients/lsp-lua.el
Mateus-Carmo31: you mean to make the (os-search-pred ...) section of the `let` in the same line, yes?

Mateus-Carmo31:
yes, I saw that pr as well. I didn't change the original client to avoid conflicts with basaran's pr, as well as to avoid breaking any compatibility issues. I will also look into your nits as soon as I'm able.

clients/lsp-lua.el
yyoncho: closing brackets should be on the same line.

yyoncho:
> yes, I saw that pr as well. I didn't change the original client to avoid conflicts with basaran's pr, as well as to 

I generally prefer using GH as a source instead of marketplace. 

> avoid breaking any compatibility issues. I will also look into your nits as soon as I'm able.

You should be able to do that without introducing a new client. Most of the automatic download functions were added afterward and still we managed to keep the backward compatibility.

lsp-mode.el
```diff
@@ -7861,6 +7863,33 @@ archieve(e. g. when the archieve has multiple files)"
...
+             (pcase system-type
+                         ('gnu/linux "linux-x64")
+                         ('darwin "darwin-x64")
+                         ('windows-nt "win32-x64")
+                         (_ "linux-x64"))))
+        (url-request-method "GET"))
+    (with-current-buffer (url-retrieve-synchronously url)
+      ;; Go to beginning of the last line (where the retrieved json with information
+      ;; is in the buffer from `url-retrieve-synchronously')
+      (goto-char (point-max))
```
kiennq, line 7887: You should skip the header part by search for `\n\n` instead, which is part of the specs for HTML response
``` elisp
                    (goto-char (point-min))
                    (re-search-forward "\n\n" nil 'noerror)
                    (delete-region (point-min) (point))
```
Also, optionally you can decode the string to `utf-8-unix` as well if needed

lsp-mode.el
```diff
@@ -7861,6 +7863,33 @@ archieve(e. g. when the archieve has multiple files)"
       (set-file-modes store-path #o0700)
       store-path)
      ((f-exists? store-path) store-path))))
+
+(defun lsp--find-latest-release-url (url &optional system-names)
```
kiennq, line 7867: instead of system-name, can this be a regex instead? and this function will call into URL, extracting URL from the item that matches regex

kiennq, line 7867: can we rename this to mention github? The json path here is specifically for github API only.

kiennq (changes requested):

clients/lsp-lua.el
```diff
@@ -530,19 +530,19 @@ and `../lib` ,exclude `../lib/temp`.
...
   (ignore client update?)
-  (let ((store-path (expand-file-name "vs-lua" lsp-clients-lua-language-server-install-dir)))
+  (let ((store-path (expand-file-name "lua-language-server-github" lsp-clients-lua-language-server-install-dir)))
     (lsp-download-install
      (lambda (&rest _)
        (set-file-modes lsp-clients-lua-language-server-bin #o0700)
        (funcall callback))
      error-callback
-     :url (lsp-vscode-extension-url "sumneko" "lua" "1.17.4")
+     :url (lsp--find-latest-release-url "https://api.github.com/repos/sumneko/lua-language-server/releases/latest")
```
kiennq, line 543: nit: also less generic but you can get the latest Github version via `https://github.com/sumneko/lua-language-server/releases/latest/download/lua-language-server-3.5.3-win32-x64.zip` for example.

lsp-mode.el
```diff
@@ -7841,10 +7842,11 @@ nil."
                (pcase decompress
                  (:gzip
                   (lsp-gunzip download-path))
-                 (:zip (lsp-unzip download-path (f-parent store-path))))
+                 (:zip (lsp-unzip download-path (f-parent store-path)))
+                 (:targz (lsp-tar-gz-decompress download-path (f-parent store-path)))
```
kiennq, line 7846: You miss an `)` here. See my propose fix for this PR here https://github.com/kiennq/lsp-mode/commit/cff987c8b1c2746edc8d3fdae81a1aedbc039a72

PR #3645, opened 2022-07-30
[completion] Fix completion always sorting on client

ericdallo:
After some debugging, I noticed lsp-completion always sorts on client side, even if server (testing with clojure-lsp) didn't provide `sort-text?` on completion items.
It seems that `completion` was the cause, not sure why that is there, but removing it fixes the issue, I'd like to understand why that was added and if we should indeed remove.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -475,8 +475,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
                                              ((lsp-completion-list? resp)
                                               (lsp:completion-list-items resp))
                                              (t resp))
-                                         (if (or completed
-                                                 (seq-some #'lsp:completion-item-sort-text? it))
+                                         (if (seq-some #'lsp:completion-item-sort-text? it)
                                              (lsp-completion--sort-completions it)
                                            it)
                                          (-map (lambda (item)


ericdallo:
c/c @yyoncho @kiennq

yyoncho:
can you crosscheck with how it works on vscode?

kiennq:
I vaguely remember that we decided to follow vscode and always perform sorting on the client unless the server returned an incomplete list.
Can you double check the behavior with vscode?

yyoncho:
can you try it in vscode? I don't think that you have to check the source, just check how vscode is behaving in this case.

ericdallo:
@yyoncho so client sort is always enabled?

yyoncho:
> @yyoncho so client sort is always enabled?

you can alter it using sortText

yyoncho:
Also, I think that we support putting weight on each of the items, I think that this can work for you too.

ericdallo:
@yyoncho How can I use this weight from server side? is it on spec?

kiennq:
> @yyoncho How can I use this weight from server side? is it on spec? I find a little bit weird need to specify sort-text for all items

You can set the `score` in the returned `CompletionItem` https://clangd.llvm.org/extensions#code-completion-scores.
We will use that score to further weight the result.

ericdallo:
@kiennq but I guess that would be a specific solution for emacs only right? 
I still don't get it why that `completion` check is necessary on lsp-mode side, which actually makes client sort almost all the time

yyoncho:
That is what vscode does. Vscode refuses pretty much any server side sorting. This was argued a lot in protocol repo and then clangd guys introduced the score field. Not sure what are the clients supporting it but it is not only emacs for sure

ericdallo:
I see :/ I think I will use the score so on server side, thank you all!

flodiebold:
FWIW, rust-analyzer just basically puts the score into `sortText` (formatted so that it will sort correctly).

yyoncho:
I think that this will sort only initially and after that, the items will be cached locally and the sortText will be ignored. This can work if the server returns always partial results to prevent the caching. On the other hand, the score will be used on each pass even if the items are cached on the client.

flodiebold:
> This can work if the server returns always partial results to prevent the caching

That's what RA does.

Issue #3647, opened 2022-08-01
lsp-mode cannot found the installed `vscode-json-languageserver`


ccqpein:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

I installed the `vscode-json-languageserver` as the [doc](https://emacs-lsp.github.io/lsp-mode/page/lsp-json/) says. (`npm i -g vscode-json-languageserver`)

However, when I open a json file, `lsp-mode` cannot find the server and ask me to install `json-ls`.

### Steps to reproduce

`rm -rf ~/.emacs.d/.cache` for cleaning the `lsp-mode` installed server.

`npm i -g vscode-json-languageserver` making sure the server is installed in the system. 

open a json file.

### Expected behavior

`lsp-mode` starts fine.

### Which Language Server did you use?

`vscode-json-languageserver`

### OS

MacOS

### Anything else?

if I ask `lsp-mode` install `vscode-json-languageserver`  in `.emacs.d/.cache`. It works fine. The problem only is `lsp-mode` cannot find the server in `PATH`. 

I run `(executable-find "vscode-json-languageserver")` in `*scrartch*` and it returns me the right location of `vscode-json-languageserver`.

ccqpein:
Update:

I found the `lsp-mode` actually installs the `vscode-langservers-extracted` rather than the `vscode-json-languageserver`. So I update the doc and give a PR #3648 .

PR #3648, opened 2022-08-01
Update install instruction for JSON language server

ccqpein:
Update the json section of [doc](https://emacs-lsp.github.io/lsp-mode/page/lsp-json/)

[Issue](https://github.com/emacs-lsp/lsp-mode/issues/3647)


jcs090218 (approved):

jcs090218:
I think we should change for `css` and `html` as well?

jcs090218:
Just pushed the fix in #3649. Thanks! ❤️

PR #3653, opened 2022-08-02
feat(clients): add support for the syntax_tree Ruby gem

geolessel:
This adds the client for the `syntax_tree` Ruby gem. The naming on
this one was a bit odd as the github org name is "ruby-syntax-tree",
the gem name is "syntax_tree", and the executable is "stree". I
attempted to standardize on "ruby-syntax-tree".

[syntax_tree](https://github.com/ruby-syntax-tree/syntax_tree) is a
Ruby gem that at the moment doesn't provide a lot of functionality
in its lsp server, but _does_ provide auto formatting which is something
I would really like.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -39,6 +39,7 @@
   * Add ~lsp-use-workspace-root-for-server-default-directory~.
   * Add [[https://github.com/artempyanykh/marksman][marksman]] support.
   * ~lsp-find-references~ to include declaration by default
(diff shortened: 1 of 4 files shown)


yyoncho (approved):

geolessel:
@yyoncho Thanks for the approval. I noticed there is a failing test but I am unsure as to how to fix it (or if I should even worry about it at all). Would you mind taking a look at it for me?

PR #3655, opened 2022-08-04
Add configuration for excluding declarations from xrefs/references

deejayem:
Add variable `lsp-references-exclude-definition`, to allow `lsp-find-references` and `xref-find-references` to exclude declarations (keeping the current default behaviour of including them).


ericdallo (approved):
LGTM

yyoncho (approved):

PR #3659, opened 2022-08-05
Do not add empty options to initializationOptions in typescript-language-server

### LINKED ISSUE #3658: [javascript] maxTsServerMemory causes tsserver to crash if not set

### Bug description

Commit 844a9c0425039164d70469e2f5640092aa9238cf introduced new initialization options for typescript-language-server.

If they are not set (default is nil), the tsserver.js receives a empty object `{}`.
In particular, `maxTsServerMemory` set as `{}` makes the server to crash.

### Steps to reproduce

- Leave maxTsServerMemory as default
- typescript-language-server crashes at the startup

### Expected behavior

We should pass to the list of options only not nil ones

ts-ls

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

matteolel:
Empty options can cause problems because a nil is converted in `{}` object in js.

In order to avoid that, I changed the list passed to `:initialization-options`, in order to add only options that are not `nil`

close #3658

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -896,12 +896,19 @@ name (e.g. `data' variable passed as `data' parameter)."
                   :priority -2
                   :completion-in-comments? t
                   :initialization-options (lambda ()
-                                            (list :disableAutomaticTypingAcquisition lsp-clients-typescript-disable-automatic-typing-acquisition
-                                                  :logVerbosity lsp-clients-typescript-log-verbosity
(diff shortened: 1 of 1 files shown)


yyoncho (approved):
I am approving with the comment that if nil serializes to {} then there will be more issues that you will hit. Several users are complaining from nil being serialized to {} instead of json null but we havent being able to pinpoint the rootcause.

PR #3668, opened 2022-08-09
lsp-rust: add semantic token modifiers

alanz:
This is more of a proof-of-concept than anything else.

It works for the "mutable" modifier, adding an underline. This matches
what the VsCode client does, but clashes with the lsp-mode default for
read vs write highlighted variables. The rest have a no-operation face
assigned, but need to be there to get the numbers right.

It is based on the approach in `lsp-terraform.el`.  I tried setting it
in the client registration as per `lsp-clojure.el`, but the final
mapping of fonts to indexes does not work for some reason,
as the call to `lsp--semantic-tokens-initialize-workspace` does not get
the correct value from (`lsp-semantic-tokens--modifier-faces-for
client`), 
despite them being set elsewhere. I suspect a timing or update problem.

I set `  (setq lsp-semantic-tokens-enable t)` in `(use-package lsp-mode` in my personal init file to enable this.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -27,6 +27,7 @@
 (require 'lsp-mode)
 (require 'ht)
 (require 'dash)
+(require 'lsp-semantic-tokens)
 
 (defgroup lsp-rust nil
   "LSP support for Rust, using Rust Language Server or rust-analyzer."
@@ -45,6 +46,12 @@
   :group 'lsp-mode
   :link '(url-link "https://github.com/rust-analyzer/rust-analyzer")
   :package-version '(lsp-mode . "8.0.0"))
+
+(defgroup lsp-rust-analyzer-semantic-tokens nil
+  "LSP semantic tokens support for rust-analyzer."
+  :group 'lsp-rust-analyzer
+  :link '(url-link "https://github.com/rust-analyzer/rust-analyzer")
+  :package-version '(lsp-mode . "8.0.1"))
 
 (defcustom lsp-rust-server 'rust-analyzer
   "Choose LSP server."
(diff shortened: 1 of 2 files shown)


clients/lsp-rust.el
```diff
@@ -1289,6 +1290,134 @@ https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensio
...
   (lsp-rust-analyzer-move-item "Down"))
 
+
+;; Semantic tokens
+
+;; Modifier faces
+(defvar font-lock-documentation-modifier-face 'font-lock-nop-face
+  "Face name to use for documentation items.")
+
+(defvar font-lock-declaration-modifier-face 'font-lock-nop-face
```
yyoncho, line 1300: does it make sense to move these in `lsp-semantic-tokens`? Also, can you at least prefix them with `lsp-` and if there are rust specific you can prefix with `lsp-rust`

clients/lsp-rust.el
```diff
@@ -1289,6 +1290,134 @@ https://github.com/rust-analyzer/rust-analyzer/blob/master/docs/dev/lsp-extensio
...
+          ("mutable" . font-lock-mutable-modifier-face)
+          ("public" . font-lock-public-modifier-face)
+          ("reference" . font-lock-reference-modifier-face)
+          ("trait" . font-lock-trait-modifier-face)
+          ("unsafe" . font-lock-unsafe-modifier-face)
+          )))
+
+(with-eval-after-load 'rust-mode
+  (when lsp-semantic-tokens-enable
+    (lsp-rust-analyzer--set-tokens)
```
yyoncho, line 1418: I don't think this call should be here. Technically, `(require 'rust-mode)` can happen from any buffer and if you want to keep this configuration only for rust buffer then you can use :after-open-fn like this: https://github.com/yyoncho/lsp-mode/blob/master/clients/lsp-go.el#L330

psibi: I should fix this in the terraform client too.

clients/lsp-rust.el
alanz: I am prefixing them with `lsp-rust-analyzer`.  I see there are already other faces defined in `lsp-rust.el`, such as `lsp-rust-analyzer-inlay-face`.  Perhaps they should stay here, to keep all the rust config in one place.

sebastiansturm:
I missed the fact that lsp-terraform-ls uses `lsp-terraform-ls--set-tokens`; that should probably be cleaned up in a separate PR; I can do that tomorrow or on the weekend.
As for the present PR: there already is an option `:semantic-tokens-faces-overrides` for `lsp-register-client` that should allow you to define type and modifier faces on a per-client basis, cf. lsp-clojure.el. In your case, replacing `:types` by `:modifiers` and pasting in your list of face definitions should just work. Could you please try that and adapt your PR accordingly (unless of course you find some bug along the way that first needs fixing)? thanks!

alanz:
>  there already is an option :semantic-tokens-faces-overrides for lsp-register-client

I tried this first, and could not get it to work. See my description in the initial PR comment (https://github.com/emacs-lsp/lsp-mode/pull/3668#issue-1333849483).  I agree it would be the cleanest way to do it.

Note that `lsp-clojure` sets the fonts only, not modifiers.

sebastiansturm:
oh, sorry! I'll have a look this evening; if the face overrides don't work out of the box, that should at least be easy enough to fix

alanz:
> if the face overrides don't work out of the box, that should at least be easy enough to fix

I think the problem is in `lsp-semantic-tokens--replace-alist-values`. It only replaces fonts for the set of modifiers already in `lsp-semantic-token-modifier-faces`, when called from `lsp-semantic-tokens--modifier-faces-for`.  So perhaps those parameters should be swapped, or the ones in the new modifiers list should be added to the set.

sebastiansturm:
good catch, thank you! Just wanted to start debugging and then I noticed you already found the culprit :) I opened a mini-pr but haven't tested it yet

clients/lsp-rust.el
```diff
@@ -945,7 +946,9 @@ or JSON objects in `rust-project.json` format."
   :library-folders-fn (lambda (_workspace) lsp-rust-analyzer-library-directories)
   :after-open-fn (lambda ()
                    (when lsp-rust-analyzer-server-display-inlay-hints
-                     (lsp-rust-analyzer-inlay-hints-mode)))
+                     (lsp-rust-analyzer-inlay-hints-mode))
+                   (when lsp-semantic-tokens-enable
```
psibi, line 950: Does this work for you ?

I tried this approach in terraform mode and I could see that the terraform language server was not getting initialized with the proper values.

psibi (commented):
Minor comment.

psibi:
> I missed the fact that lsp-terraform-ls uses lsp-terraform-ls--set-tokens; that should probably be cleaned up in a separate PR; I can do that tomorrow or on the weekend.
As for the present PR: there already is an option :semantic-tokens-faces-overrides for lsp-register-client that should allow you to define type and modifier faces on a per-client basis, cf. lsp-clojure.el.

@sebastiansturm I tried the approach for terraform today and for some reason it doesn't seem to be working. The language server is not getting initialized with the right values. I can test your PR with terraform ls server, if you get to it.

clients/lsp-rust.el
alanz: It does work for me for lsp-rust and rust-analyzer using my branch https://github.com/alanz/lsp-mode/tree/semantic-tokens-plus-optimise-modifiers which is a combination of https://github.com/emacs-lsp/lsp-mode/pull/3668 and https://github.com/emacs-lsp/lsp-mode/pull/3670

psibi: I tried again and these are my observations:

- The buffer local variable is actually set properly.
- The terraform language server is not initialized with the right values. I believe the `after-open-fn` is too late in the lifecycle process to change the initialization values.

It could be that your other changes in your branch may cause this to work. I will wait till your changes are merged, before giving it an attempt again. 

I think a more proper fix for this would be to allow proper initialization in the `make-lsp-client` function itself.

alanz: It might relate to the initial capability negotiation.  The rust language server ignores the capabilities [the client sends](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensClientCapabilities), and always returns a fixed modifiers legend (fixed for the given version of the language server).

psibi: Yeah, that's probably it.

sebastiansturm:
@psibi I changed my PR to take face overrides into account when sending out client capabilities; I'd appreciate it if you could test that against terraform ls (with appropriate face overrides used in `lsp-register-client`)

psibi:
@sebastiansturm Thanks! I tested your PR against terraform and left my comment here: https://github.com/emacs-lsp/lsp-mode/pull/3680#issuecomment-1235749313

alanz:
What still has to happen to get this PR to land?

sebastiansturm:
I think it's fine, though I'd prefer it to use semantic token overrides, which in turn requires #3680 (and that one still requires an approving review). I'll ping the other thread

alanz:
> I think it's fine, though I'd prefer it to use semantic token overrides, which in turn requires https://github.com/emacs-lsp/lsp-mode/pull/3680 (and that one still requires an approving review). I'll ping the other thread

I updated this to use the overrides, hopefully we can now land it

clients/lsp-rust.el
```diff
@@ -926,6 +927,37 @@ or JSON objects in `rust-project.json` format."
 (lsp-defun lsp-rust--analyzer-debug-lens ((&Command :arguments? [args]))
   (lsp-rust-analyzer-debug args))
 
+(defun lsp-rust-analyzer--semantic-modifiers ()
```
sebastiansturm, line 930: first, thanks a lot for your patience! Sorry the PR has been blocked so long. I'm fine with your PR as-is, though I wonder if there's a reason this is a `defun` as opposed to a variable, defcustom or something? Not trying to be annoying here, honest question :)

alanz: The simple answer is I repurposed the prior version, and changed the function name to plug it into a new place.  I will take a look when I have time to make it something more idiomatic.

alanz:
I made a variable for each specific semantic modifier, since the set of modifiers is fixed for a given version of rust-analyzer, and has not changed yet.

Rather than having a fallback default font that does nothing, I made a face for each font, so you can use `describe-char` to understand what modifiers are active.

sebastiansturm:
thanks, that's very thorough :). LGTM, though personally I'd probably still have used a `defvar`/`defcustom` instead of a `defun` for `lsp-rust-analyzer--semantic-modifiers` (or perhaps even a `defconst` if each of its entries is now separately customizable -- not sure). I'm  fine with merging this as-is; @yyoncho / @ericdallo : do you have a strong opinion on the use of `defun` here?

ericdallo:
Both are ok to me, but personally, I'd use a variable or const for better performance trading CPU for memory

ericdallo (approved):

Issue #3497, opened 2022-04-29
Capabilites not displayed when using plists


fredericgiquel:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Without using plists, capabilites of language server are displayed when pressing associated button in "lsp-describe-session" buffer.

When using plists, nothing appens as we can see in this video :

https://user-images.githubusercontent.com/17704127/166008003-4a399a5a-eff4-4f63-b03d-c8a86ad39600.mp4

### Steps to reproduce

- Use the following "init.el" minimal configuration:
```elisp
;; Initialize package sources
(require 'package)

(setq package-archives '(("melpa" . "https://melpa.org/packages/")
                         ("elpa" . "https://elpa.gnu.org/packages/")))

(package-initialize)
(unless package-archive-contents
  (package-refresh-contents))

(unless (package-installed-p 'use-package)
  (package-install 'use-package))

(require 'use-package)
(setq use-package-always-ensure t)

(use-package lsp-mode
  :init
  (setq read-process-output-max (* 1024 1024))
  (setq gc-cons-threshold 100000000))
```
- Activate plists support as describe here: https://emacs-lsp.github.io/lsp-mode/page/performance/#use-plists-for-deserialization
- Launch "lsp" then "lsp-describe-session"
- Activate the "capabilities" button

### Expected behavior

Capabilities displayed as without plists.

### Which Language Server did you use?

pylsp but the problem is the same with others language servers

### OS

Linux

Jacobfaib:
I also have this problem. I'm on macOS 12.4. I have the following in my `.bash_profiile`:
```bash
export LSP_USE_PLISTS=true
```
with the following `lsp-mode` setup (via `use-package`):
```elisp
(use-package lsp-mode
  :commands (lsp lsp-deferred)
  :bind (:map lsp-mode-map ("C-c l M-." . #'lsp-find-references))
  :init
  ;; set prefix for lsp-command-keymap (few alternatives - "C-l", "C-c l")
  (setq lsp-keymap-prefix "C-c l"
        lsp-session-file (concat user-cache-dir "lsp.cache")
[...]

PR #3669, opened 2022-08-10
Build tree-widget when using plists

fredericgiquel:
Possible solution for #3497 

With this PR, capabilities are displayed in 'lsp-session' buffer both with plist and hash-table.


lsp-protocol.el
```diff
@@ -260,7 +260,8 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
       (defalias 'lsp-copy 'copy-sequence)
       (defun lsp-member? (from key)
         (when (listp from)
-          (plist-member from key))))
+          (plist-member from key)))
+      (defalias 'lsp-structure-p 'json-plist-p))
```
kiennq, line 264: Is using `plistp` enough? Or you may have to include `(require 'json)` for this.

yyoncho (approved):

lsp-protocol.el
fredericgiquel: AFAIK, `plistp` does not exist. So I will include the require.

kiennq: It's in the latest Emacs snapshot :)
https://github.com/emacs-mirror/emacs/commit/dedd19a2f5cbf16c8ac8a122b0c39ee4e178b9e8
Anyway, using `json-plist-p` is fine.

kiennq (approved):

yyoncho:
Thank you!

Issue #3678, opened 2022-08-17
[Feature Request] mint-lang support.


zw963:
It seem like mint supported by VSCode since 2021.

https://github.com/mint-lang/mint-vscode
https://mint-lang.com/blog/mint-0.12.0

Thank you.

yyoncho:
Adding new language support is straightforward and pretty much does not require elisp skills. Usually, the language integration is performed by someone interested in the language server. Here are the docs https://emacs-lsp.github.io/lsp-mode/page/adding-new-language/ and sample PR https://github.com/emacs-lsp/lsp-mode/pull/3653

zw963:
Hi, @yyoncho , thanks point it out.

I try add a very elementary mint support, it works!

![image](https://user-images.githubusercontent.com/549126/185156698-ee936c0f-8945-48b8-8f24-5488d9e65ba8.png)

I consider this is a good start anyway, thank you.

PR #3679, opened 2022-08-17
Add mint-lang client

zw963:
Add elementary mint support. #3678


yyoncho:
Can you add an entry in clients.json?

jcs090218:
Maybe update the CHANGELOG as well?

zw963:
Done, please check.

There is one more issue, i have to  `(require 'lsp-mint)` manually in my own to make those config take effect, but, it seem like some others client will load automatically when `lsp` is running in mode hook, right?

yyoncho:
You have to update lsp-client-packages just like the sample PR I linked.

zw963:
done, please have a look, thank you.

yyoncho (approved):

PR #3680, opened 2022-08-17
Permit semantic token overrides to extend base set

sebastiansturm:
please don't merge -- I haven't tested this yet. As @alanz noted, the current implementation of `lsp--client-semantic-tokens-faces-overrides` doesn't allow client package authors to add new faces or modifiers to the base set; the present mini-PR should fix that (plan to try this out tomorrow, won't get to it today)

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -674,8 +674,9 @@ IS-RANGE-PROVIDER is non-nil when server supports range requests."
   "Merge alist BASE with OVERRIDES.
 For keys present in both alists, the assignments made by
 OVERRIDES will take precedence."
-  (let ((all-keys (-union (mapcar #'car base) (mapcar #'car overrides))))
-    (--map (cons it (alist-get it overrides (alist-get it base) nil #'string=)) all-keys)))
+  (let* ((copy-base (copy-alist base)))
+    (mapc (-lambda ((key . value)) (setf (alist-get key copy-base nil nil #'string=) value)) overrides)
+    copy-base))
 
 (defun lsp-semantic-tokens--type-faces-for (client)
   "Return the semantic token type faces for CLIENT."
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -670,25 +670,29 @@ IS-RANGE-PROVIDER is non-nil when server supports range requests."
                        (lsp-warn "No face has been associated to the %s '%s': consider adding a corresponding definition to %s"
                                  category id varname)) maybe-face)) identifiers)))
 
-(defun lsp-semantic-tokens--apply-alist-overrides (base overrides)
-  "Merge alist BASE with OVERRIDES.
+(defun lsp-semantic-tokens--apply-alist-overrides (base overrides discard-defaults)
+  "Merge or replace BASE with OVERRIDES, depending on DISCARD-DEFAULTS.
 For keys present in both alists, the assignments made by
 OVERRIDES will take precedence."
-  (let* ((copy-base (copy-alist base)))
-    (mapc (-lambda ((key . value)) (setf (alist-get key copy-base nil nil #'string=) value)) overrides)
-    copy-base))
+  (if discard-defaults
+      overrides
+    (let* ((copy-base (copy-alist base)))
+      (mapc (-lambda ((key . value)) (setf (alist-get key copy-base nil nil #'string=) value)) overrides)
+      copy-base)))
 
 (defun lsp-semantic-tokens--type-faces-for (client)
   "Return the semantic token type faces for CLIENT."
   (lsp-semantic-tokens--apply-alist-overrides
    lsp-semantic-token-faces
-   (plist-get (lsp--client-semantic-tokens-faces-overrides client) :types)))
+   (plist-get (lsp--client-semantic-tokens-faces-overrides client) :types)
+   (plist-get (lsp--client-semantic-tokens-faces-overrides client) :discard-default-types)))
 
 (defun lsp-semantic-tokens--modifier-faces-for (client)
   "Return the semantic token type faces for CLIENT."
   (lsp-semantic-tokens--apply-alist-overrides
    lsp-semantic-token-modifier-faces
-   (plist-get (lsp--client-semantic-tokens-faces-overrides client) :modifiers)))
+   (plist-get (lsp--client-semantic-tokens-faces-overrides client) :modifiers)
+   (plist-get (lsp--client-semantic-tokens-faces-overrides client) :discard-default-modifiers)))
 
 (defun lsp--semantic-tokens-on-refresh (workspace)
   "Clear semantic tokens within all buffers of WORKSPACE,
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -277,9 +277,9 @@ Faces to use for semantic token modifiers if
      . ((dynamicRegistration . t)
         (requests . ((range . t) (full . t)))
         (tokenModifiers . ,(if lsp-semantic-tokens-apply-modifiers
-                               (apply 'vector (mapcar #'car lsp-semantic-token-modifier-faces))
+                               (apply 'vector (mapcar #'car (lsp-semantic-tokens--modifier-faces-for (lsp--workspace-client lsp--cur-workspace))))
                              []))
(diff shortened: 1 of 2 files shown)


lsp-semantic-tokens.el
```diff
@@ -670,22 +670,24 @@ IS-RANGE-PROVIDER is non-nil when server supports range requests."
...
-  "Replace alist A values with B ones where available."
-  (-map
-   (-lambda ((ak . av))
-     (cons ak (alist-get ak b av nil #'string=)))
-   a))
+(defun lsp-semantic-tokens--apply-alist-overrides (base overrides)
+  "Merge alist BASE with OVERRIDES.
+For keys present in both alists, the assignments made by
+OVERRIDES will take precedence."
+  (let ((all-keys (-union (mapcar #'car base) (mapcar #'car overrides))))
```
yyoncho, line 677: Something like that seems to be more natural:

```
(let* ((copy-base (copy-alist base)))
  (mapc (-lambda ((key . value)) (setf (alist-get key copy-base) value)) overrides)
  copy-base) 
```

yyoncho: How I tested it:

```
(let* ((base '((key1 . value1) (key2 . value2)))
       (overrides '((key1 . override1)
                    (new-key . override2)))
       (copy-base (copy-alist base)))
  (mapc (-lambda ((key . value)) (setf (alist-get key copy-base) value)) overrides)
  copy-base)
```

alanz:
Something to be aware of: the list of faces supplied for modifiers has a specific order, which is indexed by the bitmap when applying them.
So the client-supplied override sequence should be preserved when doing the override.
I actually think it should be a replace operation, given the complexity.

For example, here is the initialization reply capabilites from rust analyzer, which I have annotated with the bitmap positions

```json
     "semanticTokensProvider": {
      "legend": {
        "tokenTypes": [
        .......
        ],
        "tokenModifiers": [
          "documentation",  0
          "declaration",    1
          "definition",     2
          "static",         3
          "abstract",       4
          "deprecated",     5
          "readonly",       6
          "defaultLibrary", 7
          "async",          8
          "attribute",      9
          "callable",       10
          "constant",       11
          "consuming",      12
          "controlFlow",    13
          "crateRoot",      14
          "injected",       15
          "intraDocLink",   16
          "library",        17
          "mutable",        18
          "public",         19
          "reference",      20
          "trait",          21
          "unsafe"          22
        ]
      },
      "range": true,
      "full": {
        "delta": true
      }
    },

```

alanz:
In fact, the spec says that the [SemanticTokensLegend](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#semanticTokensLegend) defines how the modifiers will be used by the server. This is a vector of names corresponding to the bitmap order the server will use.  So the client needs to arrange things so that those face names line up with the bitmap positions when they are applied.  Which sounds like there should be another stage of processing to do this lining up, ready for use.

The simplest way to do this is to let the client registration specify the fonts in the order required, and not to change the order.  Otherwise explicitly rearrange once the name -> face mapping is in place.

sebastiansturm:
that ordering should be established by `lsp-semantic-tokens-build-face-map`, which is supposed to map the client-supplied list of recognized token types/modifier types to a vector (containing `nil` for unmapped types) that can then directly be indexed by numerical token/modifier values. But since you mention it, I guess you found that it doesn't work as intended? Haven't checked in a while, so I'd better have a look

lsp-semantic-tokens.el
sebastiansturm: thanks!

alanz:
`lsp-semantic-tokens-build-face-map` just uses the value in the client of  `semantic-tokens-faces-overrides`, which needs to keep its order as I stated in https://github.com/emacs-lsp/lsp-mode/pull/3680#issuecomment-1219762950, and that will work if the client and server match each other exactly.

But if the server ever changes the modifiers it uses, it may be more robust to explicitly use the order provided by the server in its initialize response, which is the `semanticTokenLegend`, as I mentioned https://github.com/emacs-lsp/lsp-mode/pull/3680#issuecomment-1219835086

sebastiansturm:
yes, I believe that's what `lsp-semantic-tokens-build-face-map` does: within `lsp--semantic-tokens-initialize-workspace`, it calls `lsp--semantic-tokens-as-defined-by-workspace`, which in turn refers to the client-provided `semanticTokenLegend`. To each of these client-provided token/modifier types, a face (possibly `nil`) is then assigned using `semantic-token[-modifier]-faces` and `semantic-tokens-faces-overrides`

psibi:
@sebastiansturm I like this PR and it kind of works for terraform client. But For terraform, I don't want to extend the base set - since it doesn't recognize some of them. I want only the one which I defnine and I need full control over it. For eg: there are no such tokens like this in terraform:

```
    ("dependent" . lsp-face-semhl-type)
    ("concept" . lsp-face-semhl-interface)
```

But because it's already defined in the variable `lsp-semantic-token-faces`, it passes it to the language server.

psibi:
@sebastiansturm I like this PR and it kind of works for terraform client. But For terraform, I don't want to extend the base set - since it doesn't recognize some of them. I want only the one which I defnine and I need full control over it. For eg: there are no such tokens like this in terraform:

```
    ("dependent" . lsp-face-semhl-type)
    ("concept" . lsp-face-semhl-interface)
```

But because it's already defined in the variable `lsp-semantic-token-faces`, it passes it to the language server.

sebastiansturm:
thanks for trying out the PR! I would expect any extra faces passed to the server to be ignored by said server, so they won't end up in the face vector and shouldn't cause any overhead. Do you see any practical downside to the current approach? If so, we could of course just remove all definitions explicitly overridden by `(<identifier-string> . nil)`

psibi:
@sebastiansturm I think this should be fine, but I guess I would prefer as a client user - if I didn't have to pass any additional parameters to the language server if it's not required. But this definitely improves the current situation, so thank you for that! And I'm planning to patch the terraform client once this PR get's merged.

Also one thing worth mentioning is that the parameters that are passed to the language server doesn't seem to preserve the order. While this is not a problem for terraform language server, I'm not sure if it will cause issues for other language server.

Also, probably it's worth having two options:

- :semantic-tokens-faces-overrides
- :semantic-tokens-faces-extend

Where the fist option completely overrides the one defined in `lsp-semantic-tokens.el` and the second one just extends the one defined there (current behavior).

sebastiansturm:
thanks for the hint, not sure if I'll have time this week but I'll check the spec to see if token/modifier order is somehow important. I slightly changed my PR so that default tokens/modifiers can now be completely replaced by your overrides, depending on two new (optional) elements `:discard-default-types` and `:discard-default-modifiers` of `semantic-tokens-faces-overrides`

psibi:
> but I'll check the spec to see if token/modifier order is somehow important.

I could be mis-understanding, but from reading @alanz's comment here - I think that ordering matters: https://github.com/emacs-lsp/lsp-mode/pull/3680#issuecomment-1219835086

alanz:
> I think that ordering matters: https://github.com/emacs-lsp/lsp-mode/pull/3680#issuecomment-1219835086

A [later response by @sebastiansturm](https://github.com/emacs-lsp/lsp-mode/pull/3680#issuecomment-1219912904) clarifies that it actually uses the order returned in the initialize response.

So the key thing is to make sure all the required mappings are in place.

sebastiansturm:
I guess this could be merged, or are there any outstanding issues I'm currently not aware of? @alanz have you tried to use this in your pr #3668?

alanz:
I have tested this, and it works, using

```elisp
  :semantic-tokens-faces-overrides `(:discard-default-modifiers t
                                     :modifiers
                                     ,(lsp-rust-analyzer--semantic-modifiers))
```
in `mak-lsp-client`

sebastiansturm:
thank you! I think this should be good to go; @yyoncho or @ericdallo do you agree?

alanz:
Should there be a documentation update with it, or as well, related.  It is not clear how to use this in practice.

sebastiansturm:
true. I added a comment to `semantic-tokens-faces-overrides`, do you think this description will be helpful to new users?

sebastiansturm:
I think this can be merged; or (@yyoncho @ericdallo)  is there anything else you'd like me to change?

ericdallo:
Sorry for the delay!

PR #3688, opened 2022-08-22
[eslint] Correct handling of absolute paths in lsp-eslint-working-directories

RohanHart:
As the current logic turns all absolute paths into nil.

Local tests not run as, unfortunately, `cask` doesn't seem to work with NixOS wrapped emacs.


mpsq:
FYI https://github.com/emacs-lsp/lsp-mode/pull/3747

RohanHart:
Replaced by #3747

PR #3690, opened 2022-08-23
Make more capabilities become dynamically registrable

kiennq:
Makes the following capabilities become dynamically registrable since some language servers (e.x `pwsh-ls` is sending them as dynamic capabilities)
- `onTypeFormatting`
- `signatureHelp`
- `hover`
- `selectionRange`
- `typeHierarchy`

Also add `lsp--capability-for-method` as a shortcut to return `RegistrationOptions` and switch a few `lsp--capability` check to `lsp-feature?` instead.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -428,9 +428,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
   (when (or (--some (lsp--client-completion-in-comments? (lsp--workspace-client it))
                     (lsp-workspaces))
             (not (nth 4 (syntax-ppss))))
-    (let* ((trigger-chars (-> (or (lsp--capability :completionProvider)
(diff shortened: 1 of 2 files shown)


lsp-completion.el
```diff
@@ -710,8 +708,8 @@ See #2675"
   (cl-assert item nil "Completion item must not be nil")
   (lsp-completion--fix-resolve-data item)
   (or (ignore-errors
-        (when (lsp-feature? "completionItem/resolve")
-          (lsp-request "completionItem/resolve" item)))
+        (when (lsp-feature? "textDocument/completionResolve")
```
yyoncho, line 711: I don't see such method in the spec.

kiennq: Thanks for noticing, my refactoring has gone too far :D

yyoncho (approved):

PR #3696, opened 2022-08-26
Add syntax table property for comment semantic tokens.

JimDBh:
This is important to let emacs understand that certain parts of the code (e.g. wrapped in C marcos) are not in effect and should be by-passed for things like forward-sexp.

For example, if we have the following code in emacs with c-mode:
```c
#include <stdio.h>

#define SOME_MACRO 1

int myFun(int par) { // => This bracelet has no matching
  // ...
#if SOME_MACRO
  if (par == 0) { // => This incorrectly matches with the bracelet at the end of function
    printf("example check 0\n");
#else
  if (par == 1) { // => This should have been ignored by forward-sexp
    printf("example check 1\n");
#endif
    // ...
  }
  return 0;
} // => this bracelet matches with the wrong one

void someFun() {
  // Do something.
  myFun(1);
  return; // => beginning-of-defun won't work here
}
```

With this patch, these should work fine.

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -71,6 +71,22 @@ associated with the requesting language server."
   :group 'lsp-semantic-tokens
   :type 'boolean)
 
+(defcustom lsp-semantic-tokens-set-comment-syntax nil
+  "Whether to set the local syntax table for comments.
+
+When set to nil, the syntax table will not be changed.
+When set to t, semantic tokens with type comment will be also
+marked as comment in the local syntax table. This helps a lot in
+situations when it is beneficial to ignore the comments. For
+example, in c-mode, parenthesis matching should ignore the ones
+in disabled macro blocks. Note that when turned on,
+auto-indentation may not work well in these \"disabled\" code
+blocks, so this may need to be temporarily set to nil in that
+case. Use `lsp-semantic-tokens-toggle-comment-syntax' to toggle
+the value of this variable."
+  :group 'lsp-semantic-tokens
+  :type 'boolean)
+
 (defface lsp-face-semhl-constant
   '((t :inherit font-lock-constant-face))
   "Face used for semantic highlighting scopes matching constant scopes."
@@ -458,6 +474,9 @@ modified by OLD-FONTIFY-REGION.
 LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
   ;; TODO: support multiple language servers per buffer?
   (let ((faces (seq-some #'lsp--workspace-semantic-tokens-faces lsp--buffer-workspaces))
+        (types (plist-get
+                (seq-some #'lsp--semantic-tokens-as-defined-by-workspace lsp--buffer-workspaces)
+                :token-types))
         (modifier-faces
          (when lsp-semantic-tokens-apply-modifiers
            (seq-some #'lsp--workspace-semantic-tokens-modifier-faces lsp--buffer-workspaces)))
@@ -468,11 +487,15 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
           (eq nil lsp--semantic-tokens-cache)
           (eq nil (plist-get lsp--semantic-tokens-cache :response)))
       ;; default to non-semantic highlighting until first response has arrived
+      (when lsp-semantic-tokens-set-comment-syntax
+        (lsp-semantic-tokens--remove-comment-syntax beg-orig end-orig))
       (funcall old-fontify-region beg-orig end-orig loudly))
      ((not (= lsp--cur-version (plist-get lsp--semantic-tokens-cache :_documentVersion)))
       ;; delay fontification until we have fresh tokens
       '(jit-lock-bounds 0 . 0))
      (t
+      (when lsp-semantic-tokens-set-comment-syntax
+        (lsp-semantic-tokens--remove-comment-syntax beg-orig end-orig))
       (setq old-bounds (funcall old-fontify-region beg-orig end-orig loudly))
       ;; this is to prevent flickering when semantic token highlighting
       ;; is layered on top of, e.g., tree-sitter-hl, or clojure-mode's syntax highlighting.
@@ -496,6 +519,7 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
               (line-delta)
               (column 0)
               (face)
+              (type)
               (line-start-pos)
               (line-min)
               (line-max-inclusive)
@@ -529,8 +553,14 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
                  (setq current-line (+ current-line line-delta)))
                (setq column (+ column (aref data (1+ i))))
                (setq face (aref faces (aref data (+ i 3))))
+               (setq type (aref types (aref data (+ i 3))))
                (setq text-property-beg (+ line-start-pos column))
                (setq text-property-end (+ text-property-beg (aref data (+ i 2))))
+               (when lsp-semantic-tokens-set-comment-syntax
+                 (if (equal type "comment")
+                     (progn
+                       (lsp-semantic-tokens--put-comment-syntax text-property-beg text-property-end))
+                   (lsp-semantic-tokens--remove-comment-syntax text-property-beg text-property-end)))
                (when face
                  (put-text-property text-property-beg text-property-end 'face face))
                (cl-loop for j from 0 to (1- (length modifier-faces)) do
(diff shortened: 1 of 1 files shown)


yyoncho:
@sebastiansturm @ericdallo - willing to review this one?

lsp-semantic-tokens.el
```diff
@@ -529,8 +553,14 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
                  (setq current-line (+ current-line line-delta)))
                (setq column (+ column (aref data (1+ i))))
                (setq face (aref faces (aref data (+ i 3))))
+               (setq type (aref types (aref data (+ i 3))))
                (setq text-property-beg (+ line-start-pos column))
                (setq text-property-end (+ text-property-beg (aref data (+ i 2))))
+               (when lsp-semantic-tokens-set-comment-syntax
+                 (if (equal type "comment")
+                     (progn
```
yyoncho, line 561: nit: progn looks redundant

lsp-semantic-tokens.el
```diff
@@ -541,6 +571,149 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
...
+  "Returns the start of the comment pair that contains BEG, and the
+end of the comment pair that contains END. If either of such
+comment pairs does not exist, return nil for that part. This
+function also removes dangling comment starters/ends."
+  (let (prev-beg
+        prev-end
+        next-beg
+        next-end
+        tmp)
+    (save-mark-and-excursion
```
yyoncho, line 594: nit: use `lsp-save-restriction-and-excursion` (here and on a few more places)

yyoncho (commented):
Looks good. I am not familiar with this code, I will let the others review it. Meanwhile, a few elisp nits:

lsp-semantic-tokens.el
JimDBh: Thanks for the review! Done.

lsp-semantic-tokens.el
JimDBh: Done.

ericdallo (commented):
Looks good but I'd like an OK from @sebastiansturm as well 😅

sebastiansturm:
sorry for being late, have been away for a week. The feature looks very useful, though I'm not sure how to test it. I've set `lsp-semantic-tokens-set-comment-syntax` to `t`, and `lsp-semantic-tokens--put-comment-syntax` is being called, but `comment-search-forward` fails to return a match (in a small C++ buffer with an ifdef-disabled part that correctly gets fontified using `lsp-face-semhl-comment`); searching for the regexp `\s<` doesn't turn up anything either. Any hints for me on how I might verify that the feature is actually working?
I'll add some preliminary comments to the review section now

lsp-semantic-tokens.el
```diff
@@ -541,6 +577,144 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
...
+  "Toggle the value of `lsp-semantic-tokens-set-comment-syntax'"
+  (interactive)
+  (if lsp-semantic-tokens-set-comment-syntax
+      (progn
+        (setq lsp-semantic-tokens-set-comment-syntax nil)
+        (lsp-semantic-tokens--remove-comment-syntax (point-min) (point-max)))
+    (setq lsp-semantic-tokens-set-comment-syntax t))
+  (font-lock-flush))
+
+(defun lsp-semantic-tokens--get-overlapping-comments (beg end)
```
sebastiansturm, line 590: with the way this is used, I think it'd make sense to always return non-nil `beg` and `end`, with `beg` or `end` remaining unmodified if they don't happen to overlap with comment tokens. Maybe the function could also be renamed to something like `extend-region-to-include-comment-tokens` or something like that (plus the prefix, of course), in my opinion that would be slightly more intuitive

lsp-semantic-tokens.el
```diff
@@ -541,6 +577,144 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
...
+        (if next-beg
+            (when (>= next-beg end)
+              ;; Does not overlap with (beg end)
+              (setq next-end nil))
+          (lsp-semantic-tokens--remove-comment-syntax-strict next-end next-end)
+          (setq next-end nil))))
+    (cons prev-beg next-end)))
+
+(defun lsp-semantic-tokens--remove-comment-syntax-strict (beg end)
+  "Remove all commnet syntax strictly in (BEG END), even if they
```
sebastiansturm, line 643: minor typo (`commnet`)

lsp-semantic-tokens.el
```diff
@@ -541,6 +577,144 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
...
+      (goto-char beg)
+      (cl-do ((loc (text-property-search-forward
+                    'lsp-semantic-token--comment-beg)
+                   (text-property-search-forward
+                    'lsp-semantic-token--comment-beg)))
+          ((or (not loc) (>= (point) end)))
+        (let ((beg-match (prop-match-beginning loc))
+              (end-match (prop-match-end loc)))
+          (remove-text-properties beg-match end-match '(lsp-semantic-token--comment-beg))
+          (cl-loop for i from beg-match below end-match do
```
sebastiansturm, line 657: is it necessary to loop here (i.e., can't `beg-match` and `end-match` simply be passed to `{put,remove}-text-propert{y,ies}`?

lsp-semantic-tokens.el
```diff
@@ -541,6 +577,144 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
...
+                   (put-text-property i (1+ i) 'syntax-table
+                                      (get-text-property i 'lsp-semantic-token--previous-syntax-table))
+                   (remove-text-properties i (1+ i) '(lsp-semantic-token--previous-syntax-table)))))
+      ;; Remove comment ends
+      (goto-char end)
+      (cl-do ((loc (text-property-search-backward
```
sebastiansturm, line 663: the body form of `cl-do` here is very similar to the one above, might make sense to extract it (or not, given that it's not very long -- I'm not sure)

lsp-semantic-tokens.el
```diff
@@ -541,6 +577,144 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
...
+(defun lsp-semantic-tokens--put-comment-syntax (beg end)
+  "Set the comment syntax from BEG to END."
+  (let* ((overlapping (lsp-semantic-tokens--get-overlapping-comments beg end))
+         (new-beg (car overlapping))
+         (new-end (cdr overlapping)))
+    (when new-beg
+      (setq beg new-beg))
+    (when new-end
+      (setq end new-end)))
+  (lsp-semantic-tokens--remove-comment-syntax-strict beg end)
```
sebastiansturm, line 685: given that `lsp--semantic-tokens-fontify` removes comment syntax right at the outset, this call to `lsp-semantic-tokens--remove-comment-syntax-strict` might needlessly degrade performance? Or is this call necessary for some reason I don't see right now?

JimDBh:
> either

Thanks for the review! I'll address the code reviews next. As for testing the code, in my local tests, comment-search forward does find the "disabled" lines. Also paren/bracelet matchings are ignoring them. I'm using the provide snippet in this MR to test. Could you let me know what code you used for testing so I can reproduce? Thanks!

sebastiansturm:
this is what I'm using (with clangd, `lsp-semantic-tokens-set-comment-syntax` set to `t`):
```
int main() {
  int x = 0;
#ifdef NOT_DEFINED
  int y = 1;
#endif
  return x;
}
```
`lsp-semantic-tokens--put-comment-syntax` gets called with the right begin/end arguments, no luck with `comment-search-forward` though

JimDBh:
Thanks! However I was able to do comment-search-forward from the start of the buffer, and it ends up at the "i" character in the #ifdef macro (not sure why it wouldn't just be at the "#" sign, need to look into this a bit more)

JimDBh:
Upon further testing with a very large c-mode file with MANY such marcos, this patch makes lsp-mode lag a lot. The culprit seems to be text-property-search-forward/backward. I will see if we can keep a buffer-local cache of the added comment start/end pairs, and use that in the functions. In this way it should be a faster hopefully, and we also don't have to rely on text-property-search-forward, which is not available for 26.x.

yyoncho:
> In this way it should be a faster hopefully, and we also don't have to rely on text-property-search-forward, which is not available for 26.x.

I think we can go ahead and drop 26.x at this point. Not sure if this helps solving your issue.

PR #3698, opened 2022-08-27
Allow running the Sorbet server as an add-on alongside Solargraph

tudorpavel:
A common Ruby setup is to run the Sorbet typechecker alongside a main server like Solargraph, this new variable will allow this setup by enabling it in your Emacs init file:

```elisp
(setq-default
 ;; lsp-mode run sorbet LSP as add-on alongside solargraph
 lsp-sorbet-as-add-on t)
```

I've considered making the Sorbet client default to being an add-on, however if you have Sorbet installed but not Solargraph it will try to start the TypeProf server which I found doesn't work with older versions of Ruby out of the box. So in the end I think this is the cleanest solution, but let me know your thoughts.


yyoncho (approved):
I guess it will be fine to have it on by default.

PR #3702, opened 2022-08-30
lsp-completion-at-point: simplify bound calculation

kiennq:
Since we have the logic to calculate the prefix by looking at the candidate items already, the bound calculation by using trigger characters shouldn't be needed anymore.
We should simplify the logic so this will not cause unexpected behavior even if the server define every thing as trigger chars

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -430,15 +430,7 @@ The MARKERS and PREFIX value will be attached to each candidate."
             (not (nth 4 (syntax-ppss))))
     (let* ((trigger-chars (-> (lsp--capability-for-method "textDocument/completion")
                               (lsp:completion-options-trigger-characters?)))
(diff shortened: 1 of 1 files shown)


yyoncho (commented):
If I recall correctly this fix was introduced to handle java, but I am not quite sure what the situation was.

kiennq:
> If I recall correctly this fix was introduced to handle java, but I am not quite sure what the situation was.

I vaguely remember it may relate to the `@` annotation.
It shouldn't be a problem since we've used different approaches to mitigate that.
Can you recheck again to see if you can recall the problem?

kiennq:
It's this PR https://github.com/emacs-lsp/lsp-mode/pull/1662, but we don't need that anymore since the text should be replaced correctly now by the logic to calculate prefix

yyoncho (approved):

yyoncho:
seems to work fine

PR #3710, opened 2022-09-07
clients/lsp-elixir.el: disable fetchDeps option

wingyplus:
Currently, `elixir-ls` now disable this option. This changes change it to disable by default to make it consistent with language server.


clients/lsp-elixir.el
```diff
@@ -176,7 +176,7 @@ be available here: https://github.com/elixir-lsp/elixir-ls/releases/"
    ("elixirLS.mixEnv" lsp-elixir-mix-env)
    ("elixirLS.mixTarget" lsp-elixir-mix-target)
    ("elixirLS.projectDir" lsp-elixir-project-dir)
-   ("elixirLS.fetchDeps" lsp-elixir-fetch-deps t)
+   ("elixirLS.fetchDeps" lsp-elixir-fetch-deps)
```
yyoncho, line 179: this is not how you do that. This flag marks the property as boolean

wingyplus: Sorry, my bad. I fixed by changing `lsp-elixir-fetch-deps` from `t` to `nil` instead. Please kindly review again.

yyoncho (approved):

yyoncho:
Thank you!

PR #3715, opened 2022-09-10
perl: add new client for PLS server

c-alpha:
Ref: https://github.com/FractalBoy/perl-language-server

Also updating the priority order, so that the new client pls is preferred over perl-languag-server, since the PLS server supports auto-completion, whereas Perl::LanguageServer does not.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -44,6 +44,7 @@
   * Add [[https://github.com/mint-lang/mint][mint-lang]] support.
   * Add ~lsp-sorbet-as-add-on~ variable to allow running the Sorbet server as an add-on alongside Solargraph or others
   * Add [[https://github.com/nikeee/dot-language-server][dot-language-server]] (/a.k.a./ Graphviz) support.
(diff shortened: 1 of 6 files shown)


jcs090218:
Might update the CHANGELOG as well?

c-alpha:
> Might update the CHANGELOG as well?

Ah, overlooked that. Thanks for the gentle prod; will do!

c-alpha:
> > Might update the CHANGELOG as well?
> 
> Ah, overlooked that. Thanks for the gentle prod; will do!

Done.

c-alpha:
The error in `CI / unix-test` appears to be some failing package install:
```
[...]
Package ‘lsp-pyright’ installed.
Contacting host: melpa.org:443
Debugger entered--Lisp error: (error "Cannot resolve lock conflict in batch mode")
  signal(error ("Cannot resolve lock conflict in batch mode"))
```

Will the test be re-run automatically?

yyoncho (approved):

yyoncho:
Thank you!

c-alpha:
Ivan Yonchovski ***@***.***> writes:

> Thank you!

Thank you, too!

PR #3719, opened 2022-09-13
Kotlin override/implement member functionality

themkat:
Functionality to support override/implement member was recently merged in the language server (screenshot in that PR):
https://github.com/fwcd/kotlin-language-server/pull/379

Think I described the functionality quite well in the PR above:
> Override any member that is currently not implemented in the current class. This includes open functions, functions with default implementations, open members, abstract stuff etc. The main difference here is that this is general override member functionality, NOT JUST abstract members like the quick fix does. This is super useful for cases where you need to override a default implementation (e.g toString, Thread run, lsp4j interface methods like in TextDocumentService etc.), NOT just the unimplemented abstract ones. For me it would be super useful for my day job using Kotlin 😄 Having to google api docs to find signatures are a pain in the ass just for doing simple stuff like this.

(the quick fix I'm referring to is a code action that the language server provides)

I think an interactive function makes sense here. The users can then create a key binding for it if they want to. If you feel like we should have one out of the box, let me know 🙂 

This functionality is very similar to the Java-version of implement/override methods, but has a few minor differences in how it works. Both need a multiple select field where we can select the methods/members we want. Due to this reason I stole that function from lsp-java:
https://github.com/emacs-lsp/lsp-java/blob/a1aff851bcf4f397f2a968557d213db1fede0c8a/lsp-java.el#L1065

Maybe it should be in a common place? At this point I don't know where it should really be.

### DIFF
--- a/clients/lsp-kotlin.el
+++ b/clients/lsp-kotlin.el
@@ -27,7 +27,6 @@
 (require 'lsp-mode)
 (require 'cl-lib)
 (require 'dash)
-(require 'ht)
 
 (defgroup lsp-kotlin nil
   "LSP support for Kotlin, using KotlinLanguageServer."
@@ -230,9 +229,9 @@ to Kotlin."
    (lambda (member-options)
      (-if-let* ((option-items (-map (lambda (x)
                                       (list (lsp-get x :title)
-                                            (ht-get (lsp-get (lsp-get x :edit)
+                                            (lsp-get (lsp-get (lsp-get x :edit)
                                                              :changes)
-                                                    (lsp--buffer-uri))))
+                                                     (intern (concat ":" (lsp--buffer-uri))))))
                                     member-options))
                 (selected-members (lsp-kotlin--completing-read-multiple "Select overrides" option-items nil)))
          (dolist (edit (-flatten selected-members))
--- a/clients/lsp-kotlin.el
+++ b/clients/lsp-kotlin.el
@@ -26,7 +26,8 @@
 
 (require 'lsp-mode)
 (require 'cl-lib)
-(require 'cl)
+(require 'dash)
+(require 'ht)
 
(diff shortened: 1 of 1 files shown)


clients/lsp-kotlin.el
```diff
@@ -181,6 +183,61 @@ to Kotlin."
...
+(defun lsp-kotlin-implement-member ()
+  (interactive)
+  (lsp-request-async
+   "kotlin/overrideMember"
+   (list :textDocument (list :uri (lsp--buffer-uri))
+         :position (lsp--cur-position))
+   (lambda (member-options)
+     (-if-let* ((option-items (-map (lambda (x)
+                                      (list (lsp-get x :title)
+                                            (ht-get (lsp-get (lsp-get x :edit)
```
yyoncho, line 233: `ht-get` -> `lsp-get`

yyoncho (approved):
Looks good!

clients/lsp-kotlin.el
themkat: The problem with that is that `lsp-get` seems to expect a symbol as the key, while `ht-get` allows strings as well (`(lsp--buffer-ui)` is a string). I get the following error when using `lsp-get` in place of `ht-get` here:
`Error processing message (wrong-type-argument symbolp "file:///....")`


Tried a few other solutions as well, like converting to symbol, but I couldn't get it to work properly. Let me know if there is something I'm missing 🙂

yyoncho: can you try something like this:

```
(let ((key "uri"))
  (intern (concat ":" key) ))
```

The issue is that when you run that code with lsp-use-plists that map wont the hashtable but it will be plists and it won't work.

yyoncho:
Thank you! Not that happy with having that multiselect here but we are pretty much out of options...

PR #3722, opened 2022-09-14
Fix: Don’t hardcode architecture in lsp-clojure on Linux

reifieure:
Prior to this commit, all Linuxen were assumed to be running on amd64, which is wrong.  This change attempts to more precisely determine the OS/arch combo.

I suspect that a more general mechanism which can be used across clients is needed here, but this solves the immediate problem.

### DIFF
--- a/clients/lsp-clojure.el
+++ b/clients/lsp-clojure.el
@@ -53,8 +53,9 @@
                                    ((string= "x86_64" arch) "amd64")
                                    (t arch))))
               ('darwin (concat "macos-"
-                               (cond ((string= "aarch64" arch) arch)
-                                     (t "amd64"))))
+                               (cond
+                                ((string= "x86_64" arch) "amd64")
+                                (t arch))))
               ('windows-nt "windows-amd64"))))
   "Automatic download url for lsp-clojure."
   :type 'string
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
My bad, thanks for that!
It'd be nice to have internal lsp-mode functions to handle that indeed

ericdallo:
@reifieure could you fix the CI please?

brotzeit:
`string-search` is too new, we should use a function that works with emacs26+.

reifieure:
I'll follow up later today.

clients/lsp-clojure.el
```diff
@@ -46,12 +46,17 @@
 
 (defcustom lsp-clojure-server-download-url
   (format "https://github.com/clojure-lsp/clojure-lsp/releases/latest/download/clojure-lsp-native-%s.zip"
-          (pcase system-type
-            ('gnu/linux "linux-amd64")
-            ('darwin (if (string-match "^aarch64-.*" system-configuration)
-                         "macos-aarch64"
-                       "macos-amd64"))
-            ('windows-nt "windows-amd64")))
+          (let ((arch (substring system-configuration 0 (string-search "-" system-configuration))))
```
kiennq, line 49: nit: there's a new value `lsp--system-arch` that can be used with `(lsp-resolve-value lsp--system-arch)` to get the architecture of Emacs

ericdallo: https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-lua.el#L545-L547
I'll merge the PR to unblock the issue, feel free to open a new PR

ericdallo (approved):
Thank you!

PR #3733, opened 2022-09-22
feat: add list of files to be ignored with xref

elken:
Implemented to handle PHP needed extra files to get proper completion
for Laravel

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -200,6 +200,11 @@ Takes a value accepted by `spinner-start'."
                  ,@(mapcar (lambda (c) (list 'const (car c)))
                            spinner-types)))
 
+(defcustom lsp-xref-ignored-files '("_ide_helper_models.php" "_ide_helper.php")
+  "List of files to be ignored when searching by `xref'"
+  :group 'lsp-mode
+  :type '(repeat string))
+
 (defvar-local lsp-use-workspace-root-for-server-default-directory nil
   "Use `lsp-workspace-root' for `default-directory' when starting LSP process.")
 
@@ -4924,6 +4929,13 @@ identifier and the position respectively."
       (goto-char point)
       (buffer-substring (line-beginning-position) (line-end-position)))))
 
+(defun lsp--xref-ignored-file-p (item)
+  "Returns t if item should be ignored."
+  (seq-some
+   (lambda (cand)
+     (string-suffix-p cand (oref (xref-item-location item) file)))
+   lsp-xref-ignored-files))
+
 (lsp-defun lsp--xref-make-item (filename (&Range :start (start &as &Position :character start-char :line start-line)
                                                  :end (end &as &Position :character end-char)))
   "Return a xref-item from a RANGE in FILENAME."
@@ -4991,6 +5003,7 @@ type Location, LocationLink, Location[] or LocationLink[]."
                                          filename (error-message-string err)))))))
 
     (->> locations
+         (cl-remove-if #'lsp--xref-ignored-file-p)
          (seq-sort #'lsp--location-before-p)
          (seq-group-by (-compose #'lsp--uri-to-path #'lsp--location-uri))
          (seq-map #'get-xrefs-in-file)


yyoncho:
Can you describe why exactly this is needed? If the files are problematic then we should filter them another way. If you want to remove it from the UI that should happen in the frontend part (xref) not in backend.

elken:
> Can you describe why exactly this is needed? If the files are problematic then we should filter them automatically. If you want to remove it from the UI that should happen in the frontend part (xref) not in backend.

How does this filtering occur without the change?

An example is Laravel needs a big file which basically duplicates all the models, classes and docstrings in order to provide actual completion; but when querying references it picks them up in both places so you have n+1 references all the time.

> If you want to remove it from the UI that should happen in the frontend part (xref) not in backend.

Why even send it to the UI though if it's going to be used 0% of the time?

yyoncho:
> > Can you describe why exactly this is needed? If the files are problematic then we should filter them automatically.
> 
> How does this filtering occur without the change?

I though that file might be corrupted or something.

> 
> An example is Laravel needs a big file which basically duplicates all the models, classes and docstrings in order to provide actual completion; but when querying references it picks them up in both places so you have n+1 references all the time.
> 
> > If you want to remove it from the UI that should happen in the frontend part (xref) not in backend.
> 
> Why even send it to the UI though if it's going to be used 0% of the time?

Then why the server is sending them to us in the first place? My concern is that if we add handlings like this in lsp-mode.el the code will grow indefinitely. E. g. this issue can be solved by creating a wrapper over `xref-show-xrefs-function` and eventually we can add that piece of code in FAQ/php docs.

elken:
> I though that file might be corrupted or something.

Ah I see, no this is unfortunately almost required 😛 

> Then why the server is sending them to us in the first place? 

Because it's a project file and meets all the conditions of a valid reference.

> My concern is that if we add handlings like this in lsp-mode.el the code will grow indefinitely. E. g. this issue can be solved by creating a wrapper over xref-show-xrefs-function and eventually we can add that piece of code in FAQ/php docs.

Sure, but there may also be cases for users that want to say ignore references in test folders or a folder full of mocks or something. In that case, I should probably adjust this PR to handle regex; but if you're sure you don't want this in it's fine to be closed 😄

yyoncho:
I don't want to be bossy, but I would rather not have this in especially if something can be achieved using public APIs.

elken:
> I don't want to be bossy, but I would rather not have this in especially if something can be achieved using public APIs.

What would be a preferred alternative then?

yyoncho:
Something like that should work(I haven't tested it it is just to illustrate the direction)

```
(defun my/create-xref-show-xref-filtering-function ()
  (let ((xref-function xref-show-xrefs-function))
    (lambda (fetcher alist)
      (funcall
       xref-function
       (-const (funcall fetcher)) ;; here the filtering should happen
       alist))))
```

elken:
I see, I'll try and upstream this further to xref then 😄

gjm:
@elken I'm interested in this. Did you manage to implement a solution like yyoncho suggested?

elken:
> @elken I'm interested in this. Did you manage to implement a solution like yyoncho suggested?

Nope I've just kept it in my config for now

gjm:
How do you do that? Do you modify lsp-mode locally? How about excluding those results in UI? Can you give me some pointers? My knowledge of elisp is pretty basic...

elken:
> How do you do that? Do you modify lsp-mode locally? How about excluding those results in UI? Can you give me some pointers? My knowledge of elisp is pretty basic...

https://github.com/elken/doom#ignore-files-in-xref

It relies on a couple of doom macros but it's just syntax sugar for creating a function and using it with `defadvice`. I would recommend going through the [manual](https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html) too.

gjm:
I'll look into it. Thanks!

gjm:
@elken Tried your config but it seems to work too well :smile: It removes not only the `_ide_*` files but also the correct ones. `xref` finds no references...

gjm:
@elken sorry to keep bothering you... but can you give me some pointers on what I might be doing wrong? If you want to you can send me an email, it's in my profile.

PR #3737, opened 2022-09-25
Fix F# executable finding

Gastove:
Previously, this function used a bare `expand-file-name` to try and locate the `fsharpautocomplete` binary, preferring versions installed in the LSP cache directory over global installations. However, `expand-file-name` does not verify that a file exists. `expand-file-name` will, therefore, always succeed, even if it's incorrect. This causes LSP to crash loop, trying to start a (likely) non-existent binary.

This commit updates this behavior to include an existence check, ensuring that it will correctly "fail" the LSP cache version and check more paths.

### DIFF
--- a/clients/lsp-fsharp.el
+++ b/clients/lsp-fsharp.el
@@ -179,7 +179,9 @@ available, else the globally installed tool."
 
 (defun lsp-fsharp--fsac-cmd ()
   "The location of fsautocomplete executable."
-  (or (expand-file-name "fsautocomplete" lsp-fsharp-server-install-dir)
+  (or (-let [maybe-local-executable (expand-file-name "fsautocomplete" lsp-fsharp-server-install-dir)]
+        (when (f-exists-p maybe-local-executable)
+          maybe-local-executable))
       (executable-find "fsautocomplete")
       (f-join (or (getenv "USERPROFILE") (getenv "HOME"))
               ".dotnet" "tools" "fsautocomplete")))


Gastove:
Ah, interesting -- apparently, macOS has different opinions. Will have a look in a bit here.

razzmatazz:
oh, this is my bad; just tested and `(expand-file-name "fsautocomplete" nil)` does return $CWD/fsautocomplete -- I though it would return `nil` in case second param is `nil`

razzmatazz:
> Ah, interesting -- apparently, macOS has different opinions. Will have a look in a bit here.

your changes seem to work for me on macOS -- those CI failures seem to be flukes

clients/lsp-fsharp.el
```diff
@@ -179,7 +179,9 @@ available, else the globally installed tool."
 
 (defun lsp-fsharp--fsac-cmd ()
   "The location of fsautocomplete executable."
-  (or (expand-file-name "fsautocomplete" lsp-fsharp-server-install-dir)
+  (or (-let [maybe-local-executable (expand-file-name "fsautocomplete" lsp-fsharp-server-install-dir)]
```
razzmatazz, line 182: is there any particular need to have `-let` here instead of the regular `let` form?

razzmatazz (approved):

clients/lsp-fsharp.el
```diff
@@ -179,7 +179,9 @@ available, else the globally installed tool."
...
+        (when (f-exists-p maybe-local-executable)
+          maybe-local-executable))
       (executable-find "fsautocomplete")
       (f-join (or (getenv "USERPROFILE") (getenv "HOME"))
```
razzmatazz, line 186: on related note, I just found out that we have `lsp-fsharp-use-dotnet-tool-for-fsac` var that we're not consider here.. this line should be guarded, I think with:
```
(when lsp-fsharp-use-dotnet-tool-for-fsac
    (f-join (or ...
```

but I can do it myself later..

razzmatazz (approved):

PR #3741, opened 2022-09-29
lsp-yaml: support for gitlab-ci-mode

fredericgiquel:
`gitlab-ci-mode` is a major mode derived from `yaml-mode`. So we should be able to use the YAML language server with this major mode.


brotzeit (approved):

jcs090218 (approved):

clients/lsp-yaml.el
```diff
@@ -161,7 +161,7 @@ Limited for performance reasons."
                                      `(,(or (executable-find (cl-first lsp-yaml-server-command))
                                             (lsp-package-path 'yaml-language-server))
                                        ,@(cl-rest lsp-yaml-server-command))))
-                  :major-modes '(yaml-mode k8s-mode)
+                  :major-modes '(yaml-mode k8s-mode gitlab-ci-mode)
```
kiennq, line 164: What's the file extension?
Perhaps we should go with something like `:activation-fn (lsp-activate-on "yaml")` and define the regex for file name (maybe `.*\\.(yaml|yml)`) instead

fredericgiquel: The file extension is ".yml".

So I pushed a new commit with a more generic solution.

yyoncho (approved):

Issue #3753, opened 2022-10-09
Workspace edits are reversed for no apparent reason


slotThe:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

@akshaymankar and I discovered this while trying to debug why a new Haskell LSP plugin wasn't working correctly.  Basically, https://github.com/emacs-lsp/lsp-mode/commit/ec441a1761c9a071b8648f2fd2673043df968d4c added a reverse of the `documentChange` requests, which we don't quite understand.

### Steps to reproduce

A (probably too big) reproducer is as follows:

``` emacs-lisp
(let* ((edits "{
    \"documentChanges\": [
      {
        \"edits\": [
          {
            \"newText\": \"  f :: X -> X\\n  f = _\\n\",
            \"range\": {
              \"end\": {
                \"character\": 0,
                \"line\": 13
              },
              \"start\": {
                \"character\": 0,
                \"line\": 13
              }
            }
          }
        ],
        \"textDocument\": {
          \"uri\": \"file:///home/axeman/workspace/haskell-language-server/plugins/hls-class-plugin/test/testdata/DefaultImplementation.hs\",
          \"version\": 0
        }
      },
      {
        \"edits\": [
          {
            \"newText\": \"{-# LANGUAGE InstanceSigs #-}\\n\",
            \"range\": {
              \"end\": {
                \"character\": 0,
                \"line\": 1
              },
              \"start\": {
                \"character\": 0,
                \"line\": 1
              }
            }
          }
        ],
        \"textDocument\": {
          \"uri\": \"file:///home/axeman/workspace/haskell-language-server/plugins/hls-class-plugin/test/testdata/DefaultImplementation.hs\",
          \"version\": 0
        }
      }
    ]
}")
       (input "{-# LANGUAGE DefaultSignatures #-}
module DefaultImplementation where

class Test1 a where

class Test a where
  f :: a -> a
  default f :: Test1 a => a -> a
  f = id

data X = X | Y

instance Test X where

foo = 123"))
  (with-temp-buffer
    (insert input)
    (lsp--apply-workspace-edit (lsp--read-json edits))))
[...]

PR #3756, opened 2022-10-11
docs: Fix stray comment in lsp--apply-text-edit

slotThe:
[ Something we noticed while debugging #3753 ]

In 5763c334e2927d2014c34f478eeb08fd27b95088 the reversing and sorting that the comment talks about was moved to `lsp--apply-text-edits`, so we should move the comment where it actually belongs.


yyoncho (approved):

yyoncho:
Thank you!

PR #3757, opened 2022-10-12
fix(lsp-protocol.el): Fix unknown keys in lsp-terraform

jcs090218:
When `(require 'lsp-terraform)` we get the error,

```
lsp-terraform.el:269:34: Error: Unknown key: :docs-link.  Available keys: (:name :docs_link :version :source_type :dependent_modules)
```

@psibi Not sure if this is the correct fix. Can you confirm this? 😕

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -205,8 +205,7 @@ Defaults to side following treemacs default."
   (lsp-request
    "workspace/executeCommand"
    (list :command "terraform-ls.terraform.validate"
-         :arguments (vector (format "uri=%s" (lsp--path-to-uri (lsp-workspace-root))))
-         )
+         :arguments (vector (format "uri=%s" (lsp--path-to-uri (lsp-workspace-root)))))
    :no-wait t
    :no-merge t))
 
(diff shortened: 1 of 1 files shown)


psibi:
@jcs090218 I don't think it's right as the language server documention explicitly has those keys: https://github.com/hashicorp/terraform-ls/blob/main/docs/commands.md#modulecalls

jcs090218:
So the solution should be another way around? Should we change those keys in `lsp-terraform.el`?

https://github.com/emacs-lsp/lsp-mode/blob/759fb79b4fc62b9b8d2dcd8a4e4db951af090fdb/clients/lsp-terraform.el#L269

https://github.com/emacs-lsp/lsp-mode/blob/759fb79b4fc62b9b8d2dcd8a4e4db951af090fdb/clients/lsp-terraform.el#L277

Do you have any idea on how to fix this? Thanks! 😅

psibi:
@jcs090218 Sorry, I don't think I have a good idea on how to fix it since I'm using the key name corresponding to the one used and sent at the terraform language server side.

> Should we change those keys in lsp-terraform.el?

I think I tried it before, but it didn't work. But I could be mis-remembering.

yyoncho:
the keys should be changed in the caller. It is just that lsp-protocol does not convert `_` into `-`.

jcs090218:
I made another changes but the CI failed.

```
clients/lsp-terraform.el:268:1: Error: Unknown key: :docs_link.  Available keys: (:name :docs-link :version :source-type :dependent-modules)
```

TBH, I don't know why since it does compile in my workspace.

PR #3758, opened 2022-10-12
Jump to current workspace in `lsp-describe-session`

LaurenceWarne:
Hi, this PR makes it so that after calling `lsp-describe-session` your cursor is set to the widget for the current workspace if you are in one (else the existing behaviour remains - you are set to the end of the buffer).

I hope this is helpful, most of the time I call `lsp-describe-session` it's to check the workspace I'm currently in so it saves me from navigating to the right widget every time I call it.  

I'm no expert on the widget library, but let me know what you think anyway, thanks.


yyoncho (approved):
Thank you!

PR #3759, opened 2022-10-12
*lsp-help*: allow to open link with markdown-follow-link-at-point

kiennq:


lsp-mode.el
```diff
@@ -5087,9 +5089,33 @@ If EXCLUDE-DECLARATION is non-nil, request the server to include declarations."
...
+                                                (eq (car last-input-event) 'mouse-2)))
+                                    (event (cadr last-input-event))
+                                    (win (posn-window event))
+                                    (buffer (window-buffer win)))
+                              `(,buffer ,(posn-point event))
+                            `(,(current-buffer) ,(point)))]
+      (with-current-buffer buffer
+        ;; Markdown-mode puts the url in 'help-echo
+        (-some--> (get-text-property point 'help-echo)
+          (and (string-match-p goto-address-url-regexp it)
```
yyoncho, line 5106: check how lsp--document-link-handle-target  is implemented. Here, I think we should have similar handling.

kiennq: done

PR #3763, opened 2022-10-15
Allow with-lsp-workspace for non-existed one

kiennq:
Fix issue in https://github.com/emacs-lsp/lsp-treemacs/issues/144#issuecomment-1279392725
In this change:
- Reallow passing null for workspace in `with-lsp-workspace`
- Check `workspace` available in `lsp--parse-on-message` instead of relying on `with-lsp-workspace`.
- Fix CI to use the maintained `pylsp` instead of `pyls`

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1920,7 +1920,7 @@ regex in IGNORED-FILES."
   "Helper macro for invoking BODY in WORKSPACE context."
   (declare (debug (form body))
            (indent 1))
-  `(when ,workspace (let ((lsp--cur-workspace ,workspace)) ,@body)))
+  `(let ((lsp--cur-workspace ,workspace)) ,@body))
 
 (defmacro with-lsp-workspaces (workspaces &rest body)
   "Helper macro for invoking BODY against multiple WORKSPACES."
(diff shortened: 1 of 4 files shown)


lsp-mode.el
```diff
@@ -6622,37 +6622,38 @@ server. WORKSPACE is the active workspace."
...
-                (lsp--make-log-entry method id (lsp:json-response-error-error json-data)
-                                     'incoming-resp (/ (nth 2 (time-since before-send)) 1000))
-                workspace))
-             (when callback
-               (remhash id (lsp--client-response-handlers client))
-               (funcall callback (lsp:json-response-error-error json-data)))))
-          ('notification
-           (lsp--on-notification workspace json-data))
-          ('request (lsp--on-request workspace json-data)))))))
+    (when workspace
```
yyoncho, line 6625: can you elaborate on that? When workspace will be nil here?

kiennq: They are left-over from the previous test which sent an empty workspace. I figured out I can either reset the `lsp--parsed-messages` or skip to process the `workspace` here.
Will take the second approach to manually reset the stored messages then

Issue #3543, opened 2022-05-31
Upgrade XML Language Server to 0.20.0


apupier:
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)?

Describe your feature here.

see https://github.com/eclipse/lemminx/releases/tag/0.20.0

Current version used is 0.18.0 https://github.com/emacs-lsp/lsp-mode/blob/8123c5cd33bedf3ac5e88d98ede690f73b7f1b00/clients/lsp-xml.el#L192

Walheimat:
I'm about to open a PR to upgrade to 0.21.0 and add the new formatting options.

yyoncho:
> I'm about to open a PR to upgrade to 0.21.0 and add the new formatting options.

sounds good, go ahead.

PR #3764, opened 2022-10-17
Upgrade lsp-xml version and add new formatting options

Walheimat:
This upgrades the language server version to 0.21.0 and adds all new formatting options available. The default values are the defaults used in the source code.

This also replaces obsolete `lsp-xml-format-quotations` with `lsp-xml-preferences-quote-style`. Since they're not compatible (instead of "doubleQuotes" and "singleQuotes" the values are "double" and "single"), it was not marked as obsolete.

This also changes (or fixes?) `lsp-xml-validation-schema` being defined as a boolean (when it's expected to be an object) to an alist.

This PR addresses https://github.com/emacs-lsp/lsp-mode/issues/3543


jcs090218 (approved):

yyoncho:
Thank you!

PR #3768, opened 2022-10-19
Update lsp-mode.el

tonicebrian:
Ignore the `bazel-*` symbolic links pointing to object cache that Bazel creates at the root of a project

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -355,6 +355,8 @@ the server has requested that."
     "[/\\\\]build-aux\\'"
     "[/\\\\]autom4te.cache\\'"
     "[/\\\\]\\.reference\\'"
+    ;; Bazel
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

lsp-mode.el
```diff
@@ -355,6 +355,8 @@ the server has requested that."
     "[/\\\\]build-aux\\'"
     "[/\\\\]autom4te.cache\\'"
     "[/\\\\]\\.reference\\'"
+    ;; Bazel
+    "bazel-[^/\\\\]+\\'",
```
yyoncho, line 359: There is a `,` after the string? That should be removed

tonicebrian: Yep, sorry. Typo fixed.

yyoncho:
Thank you!

PR #3778, opened 2022-10-24
Rework terraform support for semantic tokens

psibi:
This makes use of the machinery introduced by Sebastian here: https://github.com/emacs-lsp/lsp-mode/pull/3680 and avoids the whole with-eval-after-load hack that I used previously


sebastiansturm:
thanks! I wouldn't make those face and modifier definitions `defun`'s though (I know the rust client currently does this, will probably change both that as well as the defaults `lsp-semantic-token-faces` and `lsp-semantic-token-modifier-faces`). I guess the most sensible choice would be a defcustom along the lines of
```
(defcustom terraform-lsp-semantic-token-faces
  '(("namespace" . lsp-face-semhl-namespace)
    ("type" . lsp-face-semhl-type)
    ("class" . lsp-face-semhl-class)
    ("enum" . lsp-face-semhl-enum)
    ("interface" . lsp-face-semhl-interface)
    ("struct" . lsp-face-semhl-struct)
    ("typeParameter" . lsp-face-semhl-type-parameter)
    ("parameter" . lsp-face-semhl-parameter)
    ("variable" . lsp-face-semhl-variable)
    ("property" . lsp-face-semhl-property)
    ("enumMember" . lsp-face-semhl-constant)
    ("event" . lsp-face-semhl-event)
    ("function" . lsp-face-semhl-function)
    ("method" . lsp-face-semhl-method)
    ("macro" . lsp-face-semhl-macro)
    ("keyword" . lsp-face-semhl-keyword)
    ("modifier" . lsp-face-semhl-member)
    ("comment" . lsp-face-semhl-comment)
    ("string" . lsp-face-semhl-string)
    ("number" . lsp-face-semhl-number)
    ("regexp" . lsp-face-semhl-regexp)
    ("operator" . lsp-face-semhl-operator)
    ("hcl-attrName" . lsp-face-semhl-member)
    ("hcl-blockType" . lsp-face-semhl-struct)
    ("hcl-blockLabel" . lsp-face-semhl-member)
    ("hcl-bool" . lsp-face-semhl-constant)
    ("hcl-string" . lsp-face-semhl-string)
    ("hcl-number" . lsp-face-semhl-number)
    ("hcl-objectKey" . lsp-face-semhl-member)
    ("hcl-mapKey" . lsp-face-semhl-member)
    ("hcl-keyword" . lsp-face-semhl-keyword)
    ("hcl-traversalStep" . lsp-face-semhl-member)
    ("hcl-typeCapsule" . lsp-face-semhl-type)
    ("hcl-typePrimitive" . lsp-face-semhl-type))
  "Mapping between terrafom-ls tokens and fonts to apply."
  :group 'lsp-terraform
  :type '(alist :key-type string :value-type face)
  :package-version '(lsp-mode . "8.1"))
```
I never use the customize facilities and was a bit surprised at the rather ugly alist display it offered me, but not sure if that's the case in vanilla Emacs too (I use doom-emacs which tries to avoid customize in general).
In any case, I think even if users were to modify those settings via `setq`, having a defcustom in the right group should improve discoverability for customize users.

sebastiansturm (commented):
LGTM, except for a minor nit: could you please rename the two `defcustom`s `terraform-lsp-...` to `lsp-terraform-...`, so they are in line with the rest of `lsp-terraform.el`, and easier to discover using autocompletion? Thanks!

yyoncho (approved):

PR #3779, opened 2022-10-24
fix(lsp-mode.el): Make sure modules will be loaded

jcs090218:
By looking at the macro `require`'s description:

```
If the optional third argument NOERROR is non-nil, then, if
the file is not found, the function returns nil instead of signaling
an error.  Normally the return value is FEATURE.
```

It seems like it only check file not found condition; I think it's better if we ignore all conditions so one lsp packages break won't break all of them.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7720,7 +7720,7 @@ Check `*lsp-install*' and `*lsp-log*' buffer."
     (seq-do (lambda (package)
               ;; loading client is slow and `lsp' can be called repeatedly
               (unless (featurep package)
-                (require package nil t)))
+                (ignore-errors (require package nil t))))
             lsp-client-packages)
     (setq lsp--client-packages-required t)))
 


yyoncho:
But that way if the package that you want to use is broken you won't be able to understand why it doesn't load.

jcs090218:
> But that way if the package that you want to use is broken you won't be able to understand why it doesn't load.

I have few thoughts on this issue,

1. That wouldn't cause too many hassles since this will only ignore errors during the requiring time. And loading a bunch of modules at a time doesn't help you isolate the issue (unless you have `debug-on-error` on), and it will cause confusion when you are trying to figure out what is broken in the chain. I prefer ignoring it here, and debugging it by evaluating `(require 'xxx)` directly. I don't think one extra lsp package would hamper one another is a good idea.
2. Add a debug flag, maybe a flag for developer/development mode?

```elisp
;; just an example
(if lsp-debug
    (require package nil t)
  (ignore-errors (require package nil t)))
```

3. The issue detector tasks should hand it over to CI. Something like following is a good start, 

```elisp
(let ((lsp-debug t)
      (lsp-auto-configure t))  ; this is enabled by default
  (lsp))  ; start the requiring test
```

All of them are design decisions. WDYT?

yyoncho:
Have you observed packages that fail when you so `(require 'lsp-xxx)`? Because this is an indication of a bug. I see that some packages are doing `executable-find` on top level for example - we should remove these instances.

jcs090218:
`lsp-metals` is giving me several errors depends on your Emacs version,

#### Emacs 29+ snapshot (after https://github.com/emacs-mirror/emacs/commit/80cf13a3d27d8a967feafeec32fd130529635592)

```
Eager macro-expansion failure: (wrong-number-of-arguments (1 . 1) 0)
```

https://github.com/Alexander-Miller/treemacs/issues/982

#### Emacs 28.2 or earlier (before https://github.com/emacs-mirror/emacs/commit/80cf13a3d27d8a967feafeec32fd130529635592)

```
load-with-code-conversion: Symbol’s function definition is void: treemacs-define-leaf-node
```

https://github.com/Alexander-Miller/treemacs/issues/987

> we should remove these instances.

You prefer to remove `lsp-metals` for now?

The worst-case scenario is me removing `lsp-metals` from my configuration which I don't want to do since I expect these bugs will be fixed one day. 😅 (https://github.com/mjhoy/dotfiles/commit/0ee2c234d998a5e99ef0eff4a2827cba455f1f28)

yyoncho:
We should fix metals.

jcs090218:
Got it! I think there is no reason for this patch anymore! Closing this now!

PR #3791, opened 2022-10-31
Revert "Dispatch each lsp--parser-on-message in next message loop"

yyoncho:
Reverts emacs-lsp/lsp-mode#3087

This is causing trouble when there is sync request and at the same time we are receiving messages from the server. 

Fixes https://github.com/emacs-lsp/lsp-mode/issues/3790

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1922,7 +1922,7 @@ regex in IGNORED-FILES."
   "Helper macro for invoking BODY in WORKSPACE context."
   (declare (debug (form body))
            (indent 1))
-  `(when ,workspace (let ((lsp--cur-workspace ,workspace)) ,@body)))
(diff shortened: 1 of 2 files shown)


ericdallo (approved):

ericdallo:
FYI @kiennq

ericdallo:
Sounds like a possible fix to me

yyoncho:
> @yyoncho @ericdallo Can someone help me to understand the problem here? Would it be okay to skip dispatch in the case of `lsp-request` while keeping it in other cases?

the problem is that when you have sync requests you may have async responses and notifications meanwhile. These should be handled synchronously. I think it is fine to keep this feature for now given the fact that there are chances native json rpc to take over and it already performs that dispatch but from a different thread.

PR #3792, opened 2022-10-31
Support for native json-rpc

yyoncho:
- this is intended to work with https://github.com/emacs-lsp/emacs

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1932,7 +1932,7 @@ regex in IGNORED-FILES."
 
 
 
-(defconst lsp-downstream-deps
+(defvar lsp-downstream-deps
   '(;; external packages
     ccls consult-lsp dap-mode helm-lsp lsp-dart lsp-docker lsp-focus lsp-grammarly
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -7170,29 +7223,33 @@ returned by COMMAND is available via `executable-find'"
...
-                                   :noquery t)))
-                       (set-process-query-on-exit-flag proc nil)
-                       (set-process-query-on-exit-flag (get-buffer-process stderr-buf) nil)
-                       (with-current-buffer (get-buffer stderr-buf)
-                         ;; Make the *NAME::stderr* buffer buffer-read-only, q to bury, etc.
-                         (special-mode))
-                       (cons proc proc))))
+  (list :connect (lambda (filter sentinel name environment-fn workspace)
+                   (if (fboundp 'json-rpc)
+                       (lsp-create-native-jsonrpc-connection
```
ericdallo, line 7228: We are using jsonrpc on lsp-dart for talking with flutter daemon, do you think we should switch to `lsp-create-native-jsonrpc-connection` in the future?

PR #3795, opened 2022-11-01
Do not use 'nil for the nix nil server id

### LINKED ISSUE #3794: lsp-install-server no longer works when called interactively

### Bug description

With the addition of the nix-nil lsp in #3789 there is now an entry in `lsp-clients` with key `nil`. See: https://github.com/emacs-lsp/lsp-mode/pull/3789/files#diff-32a52f22910769f3a77dc26a6b03e754d28fc65eae884e25431b7500e810821fR60

This means that when calling `lsp-install-server` interactively (via M-x) this form returns the client info for nix-nil instead of an actual nil value: `(gethash server-id lsp-clients)`. Which causes `lsp-install-server` to print a message about installing the "nil" language server manually.

### Steps to reproduce

`M-x lsp-install-server`

Message appears in the minibuffer: `There is no automatic installation for ‘nil’, you have to install it manually following lsp-mode’s documentation.`

### Expected behavior

Selected list of servers to install or update shown in the minibuffer.

clojure-lsp primarily, but this bug is unrelated to the language server currently being used.

### OS

MacOS

### Error callstack

_No response_

### Anything else?

Calling the function manually with the `server-id` parameter set installs/updates the requested server as expected. Example: `(lsp-install-server t 'clojure-lsp)`

byronclark:
Some functions, notably `lsp-install-server`, rely on there not being an entry with key `nil` in the `lsp-clients` hash table. Quoting `nil` doesn't make it not the same as `nil` so we need to use a different value for the server-id.

Resolves #3794

### DIFF
--- a/clients/lsp-nix.el
+++ b/clients/lsp-nix.el
@@ -57,7 +57,7 @@
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda () lsp-nix-nil-server-path))
                   :major-modes '(nix-mode)
-                  :server-id 'nil))
+                  :server-id 'nix-nil))
 
 (lsp-consistency-check lsp-nix)
 


byronclark:
I'm not a Nix user so I haven't tested to see if this breaks the nil language server. But I have tested that the fix allows calling `lsp-install-server` interactively.

yyoncho (approved):

yyoncho:
Thank you!

PR #3797, opened 2022-11-02
add .gradle to `lsp-file-watch-ignored-directories`

thanhvg:
Ref: https://docs.gradle.org/current/userguide/directory_layout.html#dir:project_root


yyoncho:
thank you!

PR #3801, opened 2022-11-05
clients/lsp-volar.el: fix lsp-volar-get-typescript-server-path

renatofdds:
`file-exists-p` should be used for checking if directory exists, e.g. "lib" in this case.


jadestrong:
I have tried this change. It really fixes the bugs caused by my previous changes to obtain the system typescript sdk. Please review the PR. @yyoncho

yyoncho (approved):

yyoncho:
Thank you @jadestrong  and @renatofdds !

PR #3805, opened 2022-11-09
fix(lsp-volar): diagnosticMode -> diagnosticModel

jadestrong:
This typo will not affect the actual function, because the volar server has a default value, and the value is also 2.


ericdallo (approved):

yyoncho:
Hey @jadestrong, willing to join the team as the volar maintainer?

PR #3812, opened 2022-11-17
Implement optional formatting options

thejan2009:
Based on a [not so recent discussion](https://discord.com/channels/789885435026604033/789890622424219658/1034017839088148562) in lsp-mode Discord, I'm opening this in hope of sending additional formatting options to lsp servers.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#formattingOptions

I'm not sure about the default values of these new variables - they might already have some emacs-native equivalents.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -4761,6 +4761,21 @@ Added to `after-change-functions'."
   (lsp--idle-reschedule buffer))
 
 
+(defcustom lsp-trim-trailing-whitespace t
+  "Trim trailing whitespace on a line."
+  :group 'lsp-mode
+  :type 'boolean)
+
+(defcustom lsp-insert-final-newline t
+  "Insert a newline character at the end of the file if one does not exist."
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -4773,7 +4788,10 @@ Applies on type formatting."
                           :text-document (lsp--text-document-identifier)
                           :options (lsp-make-formatting-options
                                     :tab-size (symbol-value (lsp--get-indent-width major-mode))
-                                    :insert-spaces (if indent-tabs-mode :json-false t))
+                                    :insert-spaces (if indent-tabs-mode :json-false t)
+                                    :trim-trailing-whitespace? lsp-trim-trailing-whitespace
```
yyoncho, line 4792: wrap each of these with `lsp-json-bool`  (you can use the same for use-spaces).

Note that in emacs nil will be serialized to null, not false.

PR #3813, opened 2022-11-18
lsp-graphql: add locally installed command, if present, as preferred option

diegov:
lsp-graphql: add locally installed command, if present, as preferred option

I'm having some trouble with the auto-installed graphql-languageserver-cli package, and I haven't figured out whether that's an environment issue on my end, or something worth reporting as a bug, but meanwhile I'm using a working graphql-lsp that I installed manually. 

Even if the auto-installed server can be fixed, most lsp clients will prefer an already installed command if available, and I didn't see any reason for lsp-graphql to be different, so I think this would be an improvement for graphql-lsp.

### DIFF
--- a/clients/lsp-graphql.el
+++ b/clients/lsp-graphql.el
@@ -27,6 +27,7 @@
 (require 'lsp-mode)
 
 (lsp-dependency 'graphql-language-service-cli
+                '(:system "graphql-lsp")
                 '(:npm :package "graphql-language-service-cli"
                        :path "graphql-lsp"))
 


yyoncho:
If you call M-x lsp-install-server I believe that you would want to have the one installed by lsp-mode. Typically the users tend to have the global one (e. g. clangd) but if they install a local one, the local one is preferred. If you don't want to use the local one you can delete it from ~/.emacs.d/cache/lsp

diegov:
> If you call M-x lsp-install-server I believe that you would want to have the one installed by lsp-mode. Typically the users tend to have the global one (e. g. clangd) but if they install a local one, the local one is preferred. If you don't want to use the local one you can delete it from ~/.emacs.d/cache/lsp

The issue is that the current lsp-graphql can't use the system installed package, because it has no `:system` entry in the `lsp-dependency` list. 

I don't see logic either to mark a server installed with `lsp-install-server` as preferred in other clients, and in my tests (eg. in lsp-yaml) even after installing one within emacs, the system path takes precedence. So I think this change bring lsp-graphql in line with many other lsp clients in lsp-mode. They all just put the `:system` entry first:

image

yyoncho:
Thank you!

Issue #3028, opened 2021-08-02
lsp-sourcekit ;; completion stops working after lsp server returns an error


wibed:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

completion stopped working after receiving a completion error from the sourcekit-lsp

the completion stops working if one follows the reproduction right here:
```
func hello(na
------------^
```
and the following logs are produced. the communication with the sourcekit-lsp 
continues, as the stacktrace shows, but completion stops working.

### Steps to reproduce

1. open swift file
2. set the root path for the project
3. type "fun", select the proper ("func") selection from the dropdown
4. try typing out " hello(name: String)" after that

### Expected behavior

the completion to work after receiving a completion error from the sourcekit-lsp

# current behavior
completion stopped working after receiving a completion error from the sourcekit-lsp

### Which Language Server did you use?

# context
os: linux 5.12.19_1
emacs: 28.0.50 (x86_64)
language: swift (swift-mode)
lsp: lsp-mode, lsp-sourcekit

### OS

Linux

### Error callstack

```emacs-lisp
- [company-diag](https://termbin.com/2oj0)
- [backtrace](https://termbin.com/so0c)
- [lsp-log](https://termbin.com/rzgk)
- [sourcekit-ls](https://termbin.com/s5ss)
- [stacktrace](https://termbin.com/1rqs)
```

### Anything else?

- ~~trying to clean the company cache manually had no effect~~
- adding (setq yas-inhibit-overlay-modification-protection t) had no effect

UPDATE:
- manually cleaning the cache with `(lsp-completion--clear-cache t)` had some effect, manual completion at point works. but even restarting `swift-mode` does not bring back expected functionality.
- disabling cache at startup setting `(setq lsp-completion-no-cache t)` had no effect.

wibed:
@danielmartin i am fairly new to emacs. as far as i can tell `sourcekit-lsp` only establishes the connection to the lsp server

[...]

PR #3815, opened 2022-11-20
Fix usage of ...trigger-for-incomplete-completions

crabbkw:
Ensure that the completion session is the same as the previous incomplete completion before using trigger style
`lsp/completion-trigger-kind-trigger-for-incomplete-completions`.

Previously the internal function `lsp-completion--get-context` had no notion of whether the current completion request was the using the same context as the previous completion request. This internal function is only called by `lsp-completion-at-point`. `lsp-completion-at-point` does calculate whether this completion request is using the same context as the last result. We therefore now pass the information from `lsp-completion-at-point` down to `lsp-completion--get-context` and use it therein when making a determination about which trigger type of completion to use.

This addresses issue #3028 where completions would stop working with the sourcekit-lsp language server for Swift. This language server in particular sends back incomplete completions frequently and also checks on its side to make sure any requested completion continuation context is the same as it was when originally requested. If not it cancels the request.

### DIFF
--- a/test/lsp-completion-test.el
+++ b/test/lsp-completion-test.el
@@ -80,11 +80,11 @@
              '("F" "daFo" "safo")
              '("F" "daFo" "safo"))))
 
-(ert-deftest lsp-completion-test-get-context ()
+(ert-deftest lsp-completion-test-get-context-trigger-characters-no-cache ()
   (setq lsp-completion--cache nil)
   (cl-labels ((do-get-context (arg)
                               (let ((non-essential arg))
-                                (lsp-completion--get-context '("_"))))
+                                (lsp-completion--get-context '("_") nil)))
               (do-test-trigger-kind (arg)
                                     (lsp:completion-context-trigger-kind
                                      (do-get-context arg)))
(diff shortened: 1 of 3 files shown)


yyoncho:
Looks good to me. I will wait some time if @kiennq wants to review as well before merging it.

yyoncho (approved):

test/lsp-completion-test.el
```diff
@@ -102,5 +102,32 @@
       (should (equal (do-test-trigger-kind t) 2))
       (should (equal (do-test-trigger-character t) "_")))))
 
+(ert-deftest lsp-completion-test-get-context-incomplete-completion ()
+  (setq lsp-completion--cache (list "not-used" :incomplete))
```
crabbkw, line 106: I'm somewhat worried about this setting of `lsp-completion--cache` leaking out beyond the scope of this test. I think we may need to wrap this whole test in an unwind-protect so that regardless of whether the test passes or fails we set `lsp-completion--cache` back to nil. Thoughts?

yyoncho: I don't think that investing time here will be worth the effort. The cache is ignored when completion is triggered from a different position, thus it is not very likely to pollute the other tests.

yyoncho:

PR #3818, opened 2022-11-21
Let lsp-warn-no-matched-clients suppress install warnings

daanturo:
Let it ignore 2 more warnings about installation.

Personally for me when needing lsp support, I would go to https://emacs-lsp.github.io/lsp-mode/page/languages/ to find installation instructions first.


yyoncho:
There is lsp-enable-suggest-server-download?

daanturo:
> There is lsp-enable-suggest-server-download?
But then `lsp` will warn about "automatic download is disabled:" so I moved the inserted cond branch higher than that.

yyoncho (approved):

yyoncho:

I thought that it will disable automatic installation, thus my comment but it is only for the warnings.

PR #3816, opened 2022-11-20
Support for Astro language server

pawelkobojek:
Support for [Astro](https://astro.build/) [language server](https://github.com/withastro/language-tools). Related issue: #3584

PR #3820, opened 2022-11-22
Added initializaion options to astro-ls

lucafanselau:
Add typescript server lookup to language server initialization options as mentioned in pull request #3816

### DIFF
--- a/clients/lsp-astro.el
+++ b/clients/lsp-astro.el
@@ -1,6 +1,6 @@
 ;;; lsp-astro.el --- lsp-mode astro integration -*- lexical-binding: t; -*-
 
-;; Copyright (C) 2022 Paweł Kobojek
+;; Copyright (C) 2022 Paweł Kobojek, Luca Fanselau
 
 ;; Author: Paweł Kobojek
 ;; Keywords: languages,astro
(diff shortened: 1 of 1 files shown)


lucafanselau:
Is there any issue with this pull request? Should I create a related issue to track it?

clients/lsp-astro.el
```diff
@@ -25,6 +25,14 @@
 ;;; Code:
 (require 'lsp-mode)
 
+
+(defun lsp-astro--get-initialization-options ()
+  "Try to get the typescript server path, to supply to the astro language server."
+  (let ((library (f-join (lsp-workspace-root) "node_modules/typescript/lib/tsserverlibrary.js")))
+    (if (file-exists-p library)
+        `(:typescript (:serverPath ,library))
+      (lsp-warn "Unable to find typescript server path for astro-ls. Guessed: %s" library))))
```
yyoncho, line 34: I assume that it may work without having tsserverlibrary specified, right?

lucafanselau: yup, the server then logs to stdio that it wont enable any typescript services. the behavior of this pr is taken from the astro vscode plugin (regarding the initialization options)

yyoncho: Ok, does it make sense to have a variable to point it to custom location for the tsserver?

lucafanselau: I guess not really because it is supposed to take the typescript in the version bundled by the project (if you use typescript you will have it as a dev dependency). And then it should always pickup that specific tsserver library.

Again, afaik, this is done equally in the vscode extension that is developed alongside the language server.

yyoncho (approved):

clients/lsp-astro.el
yyoncho: Ok, although it looks strange to me because it seems like there is no point to have that logic in the client - if you are going to look in hardcoded location then equally well you can have it in the server.

lucafanselau: yeah seems to be an issue on their side. Maybe one could contribute upstream.

But thanks for the review 🎉

Issue #3822, opened 2022-11-24
Prefer (float-time) to doing timestamps by hand


eggert:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

There's a minor performance issue in lsp-mode: several instances of `(time-to-seconds (current-time))` should just be `(float-time)`.

[Here is a proposed patch.](https://web.cs.ucla.edu/~eggert/lsp-mode/0001-Prefer-float-time-to-doing-it-by-hand.patch)

### Steps to reproduce

Use lsp-mode to generate timestamps internally.

### Expected behavior

Cheaper timestamps.

### Which Language Server did you use?

any

### OS

Linux

ericdallo:
@eggert sounds good to me, would you open a PR with that patch?

eggert:
> @eggert sounds good to me, would you open a PR with that patch?

[PR#3823](https://github.com/emacs-lsp/lsp-mode/pull/3823) opened. It also has a more-important timestamp fix.

PR #3823, opened 2022-11-24
Timestamp fixes

eggert:
This pull request handles two timestamp-related issues.

- Issue #3822).
- More importantly, fix a timestamp bug on the next Emacs beta 29.0.50. The bug is that lsp-mode assumes timestamps are in (HI LO US PS) form, but in Emacs they can also be in (TICKS . HZ) form. This timestamp bug exists even in older Emacs (since the milliseconds count is wrong) but with bleeding-edge Emacs the bug causes a signal. See [GNU Emacs bug #55163](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=55163).

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6462,6 +6462,10 @@ PARAMS are the `workspace/configuration' request params"
                      section))))))
        (apply #'vector)))
 
+(defun lsp--ms-since (timestamp)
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

ericdallo (approved):

ericdallo:
Thank you!

PR #3825, opened 2022-11-24
Implement simplified capf mode

### LINKED ISSUE #3817: corfu for variables not working in php-mode with serenata ls

### Bug description

When using corfu as completion frontend rather than company, assuming a defined `$varname`, completing `$var` doesn't work. Typing the `var` prefix *without* the `$` does work, and adds the missing `$`.

@minad [says](https://github.com/minad/corfu/issues/253) it must be a lsp-mode issue.

### Steps to reproduce

Haven't been able to make one yet. `lsp-start-plain` didn't make me wiser as it only have company (which works).

### Expected behavior

That `$varname` can be completed by typing `$var` rather than `var`.

PHP, serenata

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

xendk:
This adds another `lsp-completion-provider` option, `:corfu-capf`,
which makes auto-completion work better with standard completion
front-ends and language servers that prettify completion labels.

In essence, it uses the `filterText` property from the language server
completion reply in preference to `label`. This makes sense as
completion is basically filtering, the drawback is that what was
previously shown as "Schema" in the popup now shows up as
"\mysql_xdevapi\Schema", and "my_var" as "$my_var".

It also disables the `lsp-passthrough` completion style, so
completions are again fed through the standard completion functions
(or whatever the user has set up). Non-company-capf front-ends expect
the completion style to filter the completions, and `lsp-passthrough`
doesn't.

This does mean that prefix filtering is back in effect with the
default Emacs completion styles, which might filter out some of the
more outlandish suggestions that the language server provides
(serenata, for instance, suggests "dummy()" as completion for "my"),
which is a bit sad, but completion styles like orderless can remedy
this.

For me, it's an acceptable compromise, but for someone not coding PHP
it might not be, so that's why it's a new option. I'll argue that it
should be the `:none` option as even the built-in
`completion-at-point` wont work with serenata for variables, but I'm
not entirely sure it's OK to change the behaviour of `:none`?

Closes #3817

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -37,6 +37,7 @@
   "The completion backend provider."
   :type '(choice
           (const :tag "Use company-capf" :capf)
+          (const :tag "Use simplified capf" :corfu-capf)
           (const :tag "None" :none))
   :group 'lsp-completion
   :package-version '(lsp-mode . "7.0.1"))
@@ -159,10 +160,18 @@ This will help minimize popup flickering issue in `company-mode'."
 (cl-defun lsp-completion--make-item (item &key markers prefix)
   "Make completion item from lsp ITEM and with MARKERS and PREFIX."
   (-let (((&CompletionItem :label
+                           :filter-text?
                            :sort-text?
                            :_emacsStartPoint start-point)
           item))
-    (propertize label
+    ;; In :corfu-capf mode, the filtering is done by completion styles
+    ;; which work only with the candidate text. As some LSP servers
+    ;; return candidates with simplified labels ("var" for "$var",
+    ;; "Schema" for "\mysql_xdevapi\Schema"), prefer filter text as
+    ;; label. As completion is filtering, this does makes sense.
+    (propertize (or (unless (or (equal lsp-completion-provider :corfu-capf)
+                                (lsp-falsy? filter-text?)) filter-text?)
+                    label)
                 'lsp-completion-item item
                 'lsp-sort-text sort-text?
                 'lsp-completion-start-point start-point
@@ -757,40 +766,43 @@ TABLE PRED"
       ;; Ensure that `lsp-completion-at-point' the first CAPF to be tried,
       ;; unless user has put it elsewhere in the list by their own
       (add-to-list 'completion-at-point-functions #'lsp-completion-at-point)
-      (make-local-variable 'completion-category-defaults)
-      (setf (alist-get 'lsp-capf completion-category-defaults) '((styles . (lsp-passthrough))))
-      (make-local-variable 'completion-styles-alist)
-      (setf (alist-get 'lsp-passthrough completion-styles-alist)
-            '(completion-basic-try-completion
-              lsp-completion-passthrough-all-completions
-              "Passthrough completion."))
+
+      (unless (equal lsp-completion-provider :corfu-capf)
+        (make-local-variable 'completion-category-defaults)
+        (setf (alist-get 'lsp-capf completion-category-defaults) '((styles . (lsp-passthrough))))
+        (make-local-variable 'completion-styles-alist)
+        (setf (alist-get 'lsp-passthrough completion-styles-alist)
+              '(completion-basic-try-completion
+                lsp-completion-passthrough-all-completions
+                "Passthrough completion.")))
 
       (cond
-       ((equal lsp-completion-provider :none))
-       ((and (not (equal lsp-completion-provider :none))
+       ((or (equal lsp-completion-provider :none)
+            (equal lsp-completion-provider :corfu-capf)))
+       ((and (equal lsp-completion-provider :capf)
              (fboundp 'company-mode))
         (setq-local company-abort-on-unique-match nil)
         (company-mode 1)
-        (setq-local company-backends (cl-adjoin 'company-capf company-backends :test #'equal)))
+        (setq-local company-backends (cl-adjoin 'company-capf company-backends :test #'equal))
+        (when (bound-and-true-p company-mode)
+          (add-hook 'company-completion-started-hook
+                    completion-started-fn
+                    nil
+                    t)
+          (add-hook 'company-after-completion-hook
+                    after-completion-fn
+                    nil
+                    t)))
        (t
         (lsp--warn "Unable to autoconfigure company-mode.")))
-
-      (when (bound-and-true-p company-mode)
-        (add-hook 'company-completion-started-hook
-                  completion-started-fn
-                  nil
-                  t)
-        (add-hook 'company-after-completion-hook
-                  after-completion-fn
-                  nil
-                  t))
       (add-hook 'lsp-unconfigure-hook #'lsp-completion--disable nil t))
      (t
       (remove-hook 'completion-at-point-functions #'lsp-completion-at-point t)
-      (setq-local completion-category-defaults
-                  (cl-remove 'lsp-capf completion-category-defaults :key #'cl-first))
-      (setq-local completion-styles-alist
-                  (cl-remove 'lsp-passthrough completion-styles-alist :key #'cl-first))
+      (unless (equal lsp-completion-provider :corfu-capf)
+        (setq-local completion-category-defaults
+                    (cl-remove 'lsp-capf completion-category-defaults :key #'cl-first))
+        (setq-local completion-styles-alist
+                    (cl-remove 'lsp-passthrough completion-styles-alist :key #'cl-first)))
       (remove-hook 'lsp-unconfigure-hook #'lsp-completion--disable t)
       (when (featurep 'company)
         (remove-hook 'company-completion-started-hook


xendk:
...and this might need some work, but input would be appreciated.

lsp-completion.el
```diff
@@ -159,10 +160,18 @@ This will help minimize popup flickering issue in `company-mode'."
...
           item))
-    (propertize label
+    ;; In :corfu-capf mode, the filtering is done by completion styles
+    ;; which work only with the candidate text. As some LSP servers
+    ;; return candidates with simplified labels ("var" for "$var",
+    ;; "Schema" for "\mysql_xdevapi\Schema"), prefer filter text as
+    ;; label. As completion is filtering, this does makes sense.
+    (propertize (or (unless (or (equal lsp-completion-provider :corfu-capf)
+                                (lsp-falsy? filter-text?)) filter-text?)
+                    label)
```
minad, line 174: You could also use `(propertize filter-text 'display label)` but then we wouldn't get completion style highlighting of the candidates. Therefore this is probably not a good option.

lsp-completion.el
```diff
@@ -37,6 +37,7 @@
   "The completion backend provider."
   :type '(choice
           (const :tag "Use company-capf" :capf)
+          (const :tag "Use simplified capf" :corfu-capf)
           (const :tag "None" :none))
```
minad, line 41: Maybe this option should be replaced with a new option `lsp-completion-frontend` which can take these values:

- `default-passthrough`: Setup passthrough style, use label
- `default-filtered`: Setup flex style (or orderless if available), use filter-text
- `company`: Auto configure Company, like `default-passthrough`
- `corfu`: Auto configure Corfu, like `default-filtered`

lsp-completion.el
minad: I took another look at the lsp capf. We should better change this, such that filter-text is used for filtering. Instead of calling `(complete-with-action ...)` here:

https://github.com/emacs-lsp/lsp-mode/blob/e5a6274fac076f0d74a6ca892467be66a90a9649/lsp-completion.el#L513-L514

We should handle the actions separately (`try-completion`, `all-completions`), like Eglot does it. Then we can display the label and still use filter-text for filtering. 

https://github.com/emacs-mirror/emacs/blob/b0fa3b1a1f31c158131325f0f451c960ec54d938/lisp/progmodes/eglot.el#L2811-L2820

Iirc I've suggested this a while ago, but then we went with the simpler solution of `(complete-with-action ...)` which unfortunately led to the PHP issue.

minad: Hmm, but if we use the Eglot approach we will still get the wrong results with completion styles like flex/substring/orderless due to completion-regexp-list which is respected by all-completions. This means there is also a bug in Eglot here.

The correct approach is this:

1. Use all-completions against filter-text strings propertized with the label
2. Map over the result, looking up the label text property

kiennq: The `filter-text` is already used for filtering (see `lsp-completion--to-internal`), however, it's filtered **before** passing that back to Emacs's completion framework. Thus, skipping Emacs's filtering stuff later with `lsp-passthrough` is the better choice as we don't want the filtered candidates to be filtered again and lower the performance.

The reason why the completion candidates are filtered **before** is because the prefix of each candidate can be different and can be completely different from what Emacs expect.
Currently, to conform with the Emacs completion framework, we've reported a phony prefix based on Emacs's major mode expectation of a symbol for a language. However, when we send the completion request to the language server, it has a different interpretation of what's a language symbol (an example is Java decoration operator or JavaScript's member - see attached example) and creates the completion candidates based on that.
image

Now if we blindly apply the candidate based on Emacs's symbol expectation, that will lead to some bugs (we have bug reports for them already). I would say, this is a quirk of integrating LSP completion with Emacs completion.

To resolve that problem, we need to do the filtering with the prefix that the language server expected, which means doing it **before** passing back to Emacs filtering, via `lsp-completion--filter-candidates`.

Now if you want to support all completion styles, what you can do is to override `lsp-completion--filter-candidates` to call `all-completions` in place of the built-in lsp's flex filtering.
We have code like that before (pre Emacs 27 with no major completion styles like flex or orderless around) but decide to just go ahead with the built-in lsp flex filtering as it provides less hassle for the user (@yyoncho for confirm or to correct me if I'm wrong somewhere).
Also, displaying the candidate using its `label` is to stay as faithfully as possible to LSP specs (vscode also does that).
@minad @xendk 

As for `corfu`, I think it would be better to configure the `corfu` to bust the cache here and call the capf on every keystroke. The `lsp-mode` also does cache already so it will not call the language server on every keystroke.
Also, depends on the language server, we have to call the language server (for example `gopls` or `clangd`) to refresh the candidate list if the server explicitly tells us that the candidates they provided are incomplete.
Busting the cache for `corfu` will not necessarily lower the perf (or it might even improve the perf as `corfu` doesn't have to do the caching logic anymore)

xendk: > As for corfu, I think it would be better to configure the corfu to bust the cache here and call the capf on every keystroke.

I'll just grab this first. How does one set this up? I've tried

```
  (defun my/lsp-mode-setup-completion ()
    (setf (elt (cl-member 'lsp-completion-at-point completion-at-point-functions) 0)
          (cape-capf-buster #'lsp-completion-at-point))
    )
```

Which as far as I can see wraps `lsp-completion-at-point`, but doesn't really work:

image

(after the initial popup, no further filtering is done). Typing `$ar` and invoking `completion-at-point` just gives "No match".

minad: @kiennq 

> The filter-text is already used for filtering (see lsp-completion--to-internal), however, it's filtered before passing that back to Emacs's completion framework. Thus, skipping Emacs's filtering stuff later with lsp-passthrough is the better choice as we don't want the filtered candidates to be filtered again and lower the performance.

The filtering by the Emacs completion style happens anyway since complete-with-action calls all-completions. You cannot prevent the completion style filtering except by installing your lsp-passthrough style. The suggestion I made in https://github.com/emacs-lsp/lsp-mode/pull/3825#discussion_r1031914554 should not degrade performance for lsp-passthrough. Furthermore it will ensure that completion style filtering works correctly if filter-text is present. This will be beneficial for Corfu users like @xendk.

> Now if you want to support all completion styles, what you can do is to override lsp-completion--filter-candidates to call all-completions in place of the built-in lsp's flex filtering.

Note that complete-with-action already calls all-completions.

> Also, displaying the candidate using its label is to stay as faithfully as possible to LSP specs (vscode also does that).

With my suggestion https://github.com/emacs-lsp/lsp-mode/pull/3825#discussion_r1031914554 this will also be the case. We will get filtering using filter-text and display using the label.

> As for corfu, I think it would be better to configure the corfu to bust the cache here and call the capf on every keystroke. The lsp-mode also does cache already so it will not call the language server on every keystroke.

No, this won't be the case. Cache busting will definitely be slower. But more importantly, we lose the benefit of completion style post filtering. Note that some Corfu users like to request the candidates from the lsp server and then post filter using Orderless matching. This benefit will be lost. But as I wrote one can also use cache busting if that is desired. Corfu users have the choice.

minad: @kiennq I should add - completion style filtering works already correctly with Corfu, except in the case where filter-text is present. We are not suggesting deep architectural changes here. We only suggest that the edge case of filter-text should be fixed. Also the lsp-passthrough mechanism will stay untouched.

minad: @xendk I haven't tested the capf buster with lsp, but I think you have to adjust the input validation:

~~~ elisp
(cape-capf-buster #'lsp-completion-at-point 'never)
~~~

See also `cape--input-valid-p`.

kiennq: > The filtering by the Emacs completion style happens anyway since complete-with-action calls all-completions. You cannot prevent the completion style filtering except by installing your lsp-passthrough style. The suggestion I made in [#3825 (comment)](https://github.com/emacs-lsp/lsp-mode/pull/3825#discussion_r1031914554) should not degrade performance for lsp-passthrough. Furthermore it will ensure that completion style filtering works correctly if filter-text is present. This will be beneficial for Corfu users like @xendk.

Yes, by using the `lsp-passthrough` the Emacs completion style is by-passed for this capf.
Note that the filter text has already been used to filter the candidates in `lsp-completion--filter-candidates`, so any more filtering from Emacs completion is double filtering and will degrade perf. 
To utilize the user's `completion-styles`, you can probably call the `completion-all-completions` with the correct prefix in `lsp-completion--filter-candidates` to filter the candidates there.
And sorry as I was confused `all-completions` with `completion-all-completions` in my previous comment.



> > Now if you want to support all completion styles, what you can do is to override lsp-completion--filter-candidates to call all-completions in place of the built-in lsp's flex filtering.
> 
> Note that complete-with-action already calls all-completions.

Yes, but it should call the `all-completions` with an empty prefix as of lsp-passthrough, so no filtering.



> With my suggestion [#3825 (comment)](https://github.com/emacs-lsp/lsp-mode/pull/3825#discussion_r1031914554) this will also be the case. We will get filtering using filter-text and display using the label.

What's the difference between this with the current approach that filters using filter-text via `lsp-completion--filter-candidates`.
I saw your suggestion of returning the `filter-text` and decorating it with `'display` of `label`, but you said that it would make the match highlighting obsolete so it's not desirable. Also, does `company-capf` aware of that display as well?

minad: > Note that the filter text has already been used to filter the candidates in lsp-completion--filter-candidates, so any more filtering from Emacs completion is double filtering and will degrade perf.

Why does lsp-mode do that?

> To utilize the user's completion-styles, you can probably call the completion-all-completions with the correct prefix in lsp-completion--filter-candidates to filter the candidates there.

No, this is wrong and works against how the completion API is designed. I've described before how the completion API works. completion-all-completions calls all-completions calls the completion table, which will then again call all-completions on a list.

> Yes, but it should call the all-completions with an empty prefix as of lsp-passthrough, so no filtering.

That's right. But if you use a different completion style, that's not the case anymore. My point is that it should be up to the user to decide if they want to use an alternative completion style. I described the advantages and disadvantages of this design in https://github.com/emacs-lsp/lsp-mode/issues/3817#issuecomment-1326867546. There is no right or wrong here, just two different options.

> What's the difference between this with the current approach that filters using filter-text via lsp-completion--filter-candidates.

lsp-completion--filter-candidates doesn't implement the completion table API correctly. It doesn't handle prefix filtering, or does it handle completion-regexp-list. I think it would be better to even remove lsp-completion--filter-candidates and instead perform the filtering via the completion style. This way we get a design which fits better to the Emacs  completion API.

> I saw your suggestion of returning the filter-text and decorating it with 'display of label, but you said that it would make the match highlighting obsolete so it's not desirable. Also, does company-capf aware of that display as well?

Disregard that. It is not a good proposal. I just wanted to show a possible encoding of for the label and filter-text.

yyoncho: > > Note that the filter text has already been used to filter the candidates in lsp-completion--filter-candidates, so any more filtering from Emacs completion is double filtering and will degrade perf.
> 
> Why does lsp-mode do that?

@kiennq already explained that: there is no single prefix to return to emacs completion but each `CompletionItem` contains the range that can be used to perform the filtering. Consider the text `foo.bar|` the prefix for the first `CompletionItem` can be `foo.bar` and for the second it can be `bar`.

minad: @yyoncho Oh, thanks, I must have missed that. That's good then. Thankfully this simplifies the discussion, since lsp-completion--filter-candidates can just stay as is.

I just propose to fix the completion style double filtering. If you want I can prepare a patch. I promise that it won't degrade performance if lsp-passthrough is used, only if a different completion style is used. It is easy to check if a different completion style is used (prefix non-empty or completion-regexp-list non-nil).

The trade off will be like this:

* lsp-passthrough: No overhead due to double filtering. More communication with the server, candidates always up to date.
* completion style filtering: Overhead due to post filtering, high flexibility for the post filtering (e.g. orderless completion style), less communication with the server, but maybe outdated candidates.

Again, completion style double filtering already works perfectly fine, except for the minor edge case where the server provides a filter-text. This is all we would want to have fixed. What do you think?

xendk: This is all going a bit over my head, but it sounds promising...

Is it correctly understood that the filtering in `lsp-completion--filter-candidates` is to ensure that when one enters `b` in the `foo.bar` case, it'll check if the completion for `foo.bar` now matches `foo.barb` and whether `bar` completion matches `barb`?

yyoncho: @xendk yes

yyoncho: @minad 

> Again, completion style double filtering already works perfectly fine, except for the minor edge case where the server provides a filter-text. This is all we would want to have fixed. What do you think?

FTR initially we were performing the filtering in emacs and we did no filtering on lsp-mode side. Later the bug reports came and we had to move the filtering on lsp-mode side - we decided to work correctly instead of being a good citizen and returning the wrong set of results. If we do second filtering it might cut off the items with miscalculated prefixes. IMHO in order for the whole thing to work the filtering strategy should be passed to the provider instead (not sure that such a thing can be achieved). But that I guess is not in the scope. 

> less communication with the server

Not sure if we will do better than what we currently have because we already have a cache locally in lsp-mode.

In general, I am fine with putting a less precise implementation of the completion for the sake of being integrated better with emacs tools but the limitations should be well documented.

minad: @yyoncho Good to know the history of the developments. Likely, I would have made the same design choices as you did.

> Not sure if we will do better than what we currently have because we already have a cache locally in lsp-mode.

In fact, we can. If you use Corfu this is already the case. Note that the caching happens on the side of Corfu.

> In general, I am fine with putting a less precise implementation of the completion for the sake of being integrated better with emacs tools but the limitations should be well documented.

Okay, I will prepare a PR when I find time. To be 100% clear and to avoid any misunderstandings: The patch I am going to create will preserve the existing filtering if lsp-passthrough is configured. There won't be a regression, no loss in precision. Everything will just behave exactly as before in the default lsp-completion configuration. The difference will only be observable if you configure a different completion style and if the server attaches filter-text to the candidates.

xendk: > The difference will only be observable if you configure a different completion style and if the server attaches filter-text to the candidates.

Any progress? I don't understand enough of this to take a stab at it.

xendk: @minad Could you spare some time for this one? This PR might halfway work, but your idea seemed better.

PR #3826, opened 2022-11-24
Fix activation of lsp-help-mode

vincentbernat:
For some reason, enabling `(lsp-help-mode)` before use of `(with-help-window)` will revert the buffer back to `(help-mode)`. Therefore, move the activation of lsp-help-mode after that.


yyoncho:
Thank you for the PR, I am closing it because the issue was addressed in another PR.

Issue #3802, opened 2022-11-06
LSP server download swallows errors


Rconybea:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

If host cannot run language server (bad config / can't run executable),   errors may be swallowed instead of being reported to *lsp-log*.

Background: 
1. setting up lsp (lsp-mode-20221104.1904) with emacs 28.0.50
2. host VM a fairly old+stable linux,  I think its RHEL 7.   uname reports as 3.10.0-1160.71.1.el7.x86_64
3. I don't have root privileges,   need to take contents of `/lib`, `/lib64`, `/usr/lib` etc as given

with minimal lsp install:
```
(use-package lsp-mode :ensure t)
(add-hook 'c++-mode-hook #'lsp)
```

Attempting `M-x lsp` would loop forever:  
1.  emacs asks to download a language server;  
2. user responds `clangd`:
3. emacs proceeds to print a few progress messages (contacting `github.com:443` ...)
4. downloaded clangd appears in `~/.emacs.d/.cache/lsp/clangd_13.0.0/bin/clangd`
5. ... repeat forever

Note that the *lsp-log* 

### Steps to reproduce

1. supply bad value for `lsp-clients-clangd-args`:

```(setq lsp-clients-clangd-args "--header-insertion-decorators=0")  ;; <--- should have been a list```

2. Invoke `M-x lsp` in a `.cpp` buffer

### Expected behavior

error on modeline or `*lsp-log*`:
 ```Wrong type argument: listp, "--header-insertion-decorators=0"```
or
```unable to execute: /path/to/clangd, error: ...```

### Which Language Server did you use?

lsp-clangd

### OS

Linux

### Error callstack

```emacs-lisp
(nothing here)
```

### Anything else?

I believe the error-swallowing code is here (from `lsp_mode.el`):
```
(defun lsp--server-binary-present? (client)
  (unless (equal (lsp--client-server-id client) 'lsp-pwsh)
    (condition-case ()
      (-some-> client lsp--client-new-connection (plist-get :test?) funcall)
     (error nil)
     (args out-of-range nil))))
```

[...]

PR #3840, opened 2022-12-06
Prefer plain unzip for extracting archives instead of PowerShell when available

josteink:
Currently on Linux, Powershell is used to unzip ZIP-archives even when plain unzip is available.

However using PowerShell has more dependencies (versions, needs correct parameters, escaping etc) than plain `unzip`. On Ubuntu powershell is also provided as a snap which may or may not cause issues wrt to environment-variables, permissions, etc.

This PR changes the probing order in `lsp-mode.el` to prefer unzip over PowerShell when unzip is available.

This PR partially adresses: https://github.com/emacs-lsp/lsp-mode/issues/3802

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8041,8 +8041,8 @@ archieve(e. g. when the archieve has multiple files)"
   "Unzip script to unzip file.")
 
 (defcustom lsp-unzip-script (lambda ()
-                              (cond ((executable-find "powershell") lsp-ext-pwsh-script)
-                                    ((executable-find "unzip") lsp-ext-unzip-script)
+                              (cond ((executable-find "unzip") lsp-ext-unzip-script)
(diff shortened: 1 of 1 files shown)


kiennq (approved):

jcs090218 (approved):
CI should be good for now.

jcs090218:
> So if it's still red, it means I broke something?

No, the ert tests seems to be broken in two snapshot (windows and linux) and there is nothing to do with this PR. This PR should be good to merge, but we will have to wait for other maintainers with admin privileges to merge this.

yyoncho:
I believe that we switched them for exactly the same problem but in the opposite direction. But lets see how it goes. We may change the code to be os specific and prefer  unzip for linux and powershell for windows.

josteink:
Makes sense. I was considering how it was "clever" to prove for tools rather than platforms, but maybe in practice that works less great than one would think 🤷

PR #3844, opened 2022-12-10
[Fix emacs-lsp/lsp-ui#452] Fix markdown link navigation in *lsp-help* buffers

vspinu:
Because markdown links are rendered during fontification I couldn't see where to integrate the link lookup except of the generic [lsp--fontlock-with-modelsp--fontlock-with-mode](https://github.com/emacs-lsp/lsp-mode/blob/ee6d0b030454218d9346b41696b3a9adaaef61e3/lsp-mode.el#L5346-L5363)

fixes emacs-lsp/lsp-ui#452

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5158,9 +5158,12 @@ If EXCLUDE-DECLARATION is non-nil, request the server to include declarations."
                               `(,buffer ,(posn-point event))
                             `(,(current-buffer) ,(point)))]
       (with-current-buffer buffer
-        ;; Markdown-mode puts the url in 'help-echo
-        (-some-> (get-text-property point 'help-echo)
-          (lsp--document-link-handle-target))))))
+        (when-let* ((face (get-text-property point 'face))
+                    (url (or (and (eq face 'markdown-link-face)
+                                  (get-text-property point 'help-echo))
+                             (and (memq face '(markdown-url-face markdown-plain-url-face))
+                                  (nth 3 (markdown-link-at-pos point))))))
+          (lsp--document-link-handle-target url))))))
 
 (defvar lsp-help-mode-map
   (-doto (make-sparse-keymap)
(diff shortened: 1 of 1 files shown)


lsp-mode.el
```diff
@@ -5218,6 +5218,23 @@ MODE is the mode used in the parent frame."
         (lambda (_start _end _match) t))
   (prettify-symbols-mode 1))
 
+(defvar lsp--markdown-link-map
```
yyoncho, line 5221: can you make this public? I guess one might wanna change the bindings to C-RET, or Ctrl-Click.

vspinu: Done. Renamed into a more friendly `lsp-help-link-keymap`.

I just realized that there are several url types in markdown: `markdown-plain-url-face` and `markdown-url-face`. 

I am afraid we are getting too much into markdown's internals with this. This is necessary because the exposed  markdown API like  `markdown-follow-link-at-point` and `markdown-link-p` breaks in `*lsp-help*` buffer. 

It breaks because `lsp` removes `invisible` text in [lsp--buffer-string-visible](https://github.com/emacs-lsp/lsp-mode/blob/1cdb46c26a137be087e374cb2496bb909c7b181f/lsp-mode.el#L5246) which is relied upon by markdown internals. Is that  invisible-text-removal really necessary? If not, things could be simplified a bit.

yyoncho: > It breaks because `lsp` removes `invisible` text in [lsp--buffer-string-visible](https://github.com/emacs-lsp/lsp-mode/blob/1cdb46c26a137be087e374cb2496bb909c7b181f/lsp-mode.el#L5246) which is relied upon by markdown internals. Is that invisible-text-removal really necessary?

It is needed for lsp-ui-sideline. The overlays does not handle invisible text.

yyoncho:
CI is not happy: https://github.com/emacs-lsp/lsp-mode/actions/runs/3669045447/jobs/6202594289#step:12:640

yyoncho:
```
In toplevel form:
lsp-mode.el:9399:1:Error: the following functions are not known to be defined: text-property-search-forward, prop-match-beginning, prop-match-end
.....................Test lsp-byte-compilation-test backtrace:
```

vspinu:
Rewrote. Hopefully it's good now.

yyoncho:
Looks good, I will do some manual testing I and will proceed with the merge.

yyoncho (approved):

dschrempf:
I still need to use `  (advice-add 'lsp--help-open-link :override #'lsp-ui-doc--open-markdown-link)` to be able to open links in LSP help buffers. Any thoughts on this? Do you think this is a problem on my side?

PR #3846, opened 2022-12-10
Delay lsp-help-mode hooks to allow users to operate on the help content

vspinu:
Currently help buffer content is inserted after the `lsp-help-mode-hook` has been run. This makes it hard for users to post-process   help content (e.g. reformatting, replacing unreadable characters etc). 

This PR fixes that by delaying the mode hooks after the content has been inserted.


yyoncho:
Thank you!

PR #3848, opened 2022-12-12
use the workspace object to resolve workspace path (fsharp)

augustfengd:
`lsp-workspace-root` will not find the workspace root if it is a new addition.


yyoncho:

PR #3850, opened 2022-12-12
feature: Add command to uninstall server

jcs090218:
For https://github.com/emacs-lsp/lsp-mode/discussions/3431.

Code is from doomemacs, see https://github.com/doomemacs/doomemacs/blob/e0385052a8004ec54a402c27357b1352840eb798/modules/tools/lsp/autoload/lsp-mode.el#L13-L21

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7784,6 +7784,16 @@ When prefix UPDATE? is t force installation even if the server is present."
     (lsp--install-server-internal chosen-client update?)))
 
 ;;;###autoload
+(defun lsp-uninstall-server (dir)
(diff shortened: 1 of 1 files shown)


ericdallo (approved):
Nice! that was on my backlog for a long time 😂 thanks for that!

ericdallo:
Also I thought about a lsp-uninstall-all-servers but we can add it later

jcs090218:
> Also I thought about a lsp-uninstall-all-servers but we can add it later

I have opened up the PR in #3868. Can you review it? 😄

Issue #3550, opened 2022-06-06
Autocompletion popups are not working with lsp-ltex


swarnendubiswas:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

I can get autocompletion to work with `txt` files, but not when I enable `lsp`. I believe this error has to do with value of `completion-at-point-functions` set by `lsp`.

### Steps to reproduce

```
(setq straight-build-dir (format "build/%d%s%d"
                                 emacs-major-version
                                 version-separator
                                 emacs-minor-version)
      straight-use-package-by-default t)

(let ((bootstrap-file
       (expand-file-name "straight/repos/straight.el/bootstrap.el" user-emacs-directory))
      (bootstrap-version 5))
  (unless (file-exists-p bootstrap-file)
    (with-current-buffer
        (url-retrieve-synchronously
         "https://raw.githubusercontent.com/raxod502/straight.el/develop/install.el"
         'silent 'inhibit-cookies)
      (goto-char (point-max))
      (eval-print-last-sexp)))
  (load bootstrap-file nil 'nomessage))

;; These variables need to best before loading `use-package'.
(setq use-package-hook-name-suffix nil)
(straight-use-package 'use-package)

(use-package vertico
  :hook (after-init-hook . vertico-mode)
  :bind
  (:map vertico-map
   ("C-M-j" . vertico-exit-input)
   ("<tab>" . vertico-insert)))

(use-package corfu
  :hook (after-init-hook . global-corfu-mode)
  :custom
  (corfu-auto t "Enable auto completion")
  (corfu-auto-prefix 3)
  :bind
  (:map corfu-map
        ("[tab]" . corfu-next)
        ("C-n" . corfu-next)
        ("[backtab]" . corfu-previous)
        ("C-p" . corfu-previous)
        ("<escape>" . corfu-quit)))

(use-package cape
  :after corfu
  :demand t
  :commands (cape-history ; Complete from Eshell, Comint or minibuffer history
             cape-file
             cape-keyword ; Complete programming language keyword
             cape-tex ; Complete unicode char from TeX command, e.g. \hbar.
[...]

PR #3854, opened 2022-12-14
Add :exclusive 'no to capf backend

vspinu:
Problem: Often lsp backend cannot find completions at point (in my case pylsp) but the good-old `python-completion-at-point` happily completes. 

This solution was also mentioned in #3550, so I would presume this also addresses that issue.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -523,7 +523,8 @@ The MARKERS and PREFIX value will be attached to each candidate."
        :company-doc-buffer (-compose #'lsp-doc-buffer
                                      #'lsp-completion--get-documentation)
        :exit-function
-       (-rpartial #'lsp-completion--exit-fn candidates)))))
+       (-rpartial #'lsp-completion--exit-fn candidates)
+       :exclusive 'no))))
 
 (defun lsp-completion--find-workspace (server-id)
   (--first (eq (lsp--client-server-id (lsp--workspace-client it)) server-id)


yyoncho:
This one is breaking the fuzzy completion. I will revert it for now until we investigate.

minad:
@vspinu My recommendation is to use `cape-wrap-nonexclusive` or `cape-capf-nonexclusive` in personal user configurations. The problem is that `:exclusive no` is implemented incorrectly in Company and default completion, such that non-prefix completion styles break (e.g. fuzzy). There is even a fixme in minibuffer.el. I only recommend to use `:exclusive no` with Corfu, where it works as it should.

yyoncho:
> I only recommend to use `:exclusive no` with Corfu, where it works as it should.

Seems like we can at least make it configurable.

vspinu:
> Seems like we can at least make it configurable.

If its a bug in company shouldn't it be fixed there instead?

yyoncho:
> > Seems like we can at least make it configurable.
> 
> If its a bug in company shouldn't it be fixed there instead?

Yep. I am just saying what we can do on our side.

minad:
@vspinu It is not so simple. 

The bug lies in the `completion--capf-wrapper` from minibuffer.el. See this comment: https://github.com/emacs-mirror/emacs/blob/48db8b68a8eb5c12d5682f2eb31cadc53186f5d7/lisp/minibuffer.el#L2736-L2743. In Corfu I replace the Capf wrapper via an advice to fix this and also to improve the auto completion handling. See this comment: https://github.com/minad/corfu/blob/948f7121c783936d4605ee713b497f839c16e128/corfu.el#L1226-L1229.

Now the problem is that Stefan Monnier disagrees with me about how to fix this properly. Prefix completion is obviously only a heuristic, but he argues that `:exclusive no` should make sure that the Capf doesn't set in too often (since it is non-exclusive). This is achieved by only checking the prefix.

Since the overall situation is a bit messy and given the disagreement, my answer is to make everything configurable via Capf transformers, see https://github.com/minad/cape#capf-adapters-and-transformers. Unfortunately it requires some deeper knowledge of the Capf mechanism to use the transformers.

PR #3855, opened 2022-12-15
Prevents CPU throttling issue when LS doesn't support semantic token

psibi:
This fixes the issue that is easily reproducible with the Haskell language server. Steps to reproduce:

- Enable semantic tokens

(setq lsp-semantic-tokens-enable t)

- Configure haskell language server and open some haskell file. Note that HLS doesn't support semantic tokens.
- You could observe that your CPU will be throttled and will have a high consumption.

Enabling the logs, you can see that the lsp client will be constantly sending semantic token requests to hls which will be continuous returning -32601.

Notes regarding the changes:

- Right now `(when (lsp-feature? "textDocument/semanticTokens"))` will always return true no matter what. In fact, even an arbitrary value like `(when (lsp-feature? "some_value"))` returns true. So, I don't think it's a correct check. And I believe this happens because of the missing `:check-command` in the `lsp-method-requirements`.
- Changing it to "textDocument/semanticTokensFull" fixes the issue. And I believe this is because it has an appropriate `:check-command`.

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -569,7 +569,7 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
   ;; which should minimize those occasions where font-lock region extension extends beyond the
   ;; region covered by our freshly requested tokens (see lsp-mode issue #3154), while still limiting
   ;; requests to fairly small regions even if the underlying buffer is large
-  (when (lsp-feature? "textDocument/semanticTokens")
+  (when (lsp-feature? "textDocument/semanticTokensFull")
     (lsp--semantic-tokens-request
      (cons (max (point-min) (- (window-start) (* 5 jit-lock-chunk-size)))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

yyoncho:
Looks good, I will let @ericdallo check it out too.

ericdallo (approved):
Looks good!

michaelpj:
This seems odd to me? There is no `textDocument/semanticTokensFull` method. There is `textDocument/semanticTokens/full`, however?

Semantic tokens is really annoying because they linked the _capability_ to `textDocument/semanticTokens`, but that isn't an actual method you can call! It's just pointlessly weird :/ So I think looking for a capability associated with `textDocument/semanticTokens` is actually correct.

ericdallo:
Good point, I forgot semantic tokens was the only feature not following the standard

PR #3856, opened 2022-12-15
Makes the lsp-headerline--string a frame-parameter

ryan-s-wilson:
This adds functionallity when the same document is open across multiple frames (e.g. multi monitor setup) allowing the headerline to display different text in each of the frames.


yyoncho:

PR #3824, opened 2022-11-24
Support some treesit prog-mode

nasyxx:
Support

+ c-ts-mode
+ c++-ts-mode
+ css-ts-mode
+ java-ts-mode
+ js-ts-mode
+ json-ts-mode
+ python-ts-mode

PR #3858, opened 2022-12-16
Support more treesit modes

meliache:
I added some more treesit-based major modes that have been added with emacs29. I just went through the [list](https://github.com/emacs-mirror/emacs/blob/4331771222fe670bd45c460b5e4f68e0dd4dc3c1/etc/NEWS.29#L3019-L3090) of new treesit major modes and added those that were missing from `lsp-language-id-configuration` but had a non-treesit variant in that configuration. Didn't do much testing except for golang, but I think the changes are straightforward.

This PR extends the list of treesit modes added in PR #3824 by @nasyxx.

In particular, I added:
- `go-ts-mode`
- `go-mod-ts-mode`
- `dockerfile-ts-mode`
- `cmake-ts-mode`
- `toml-ts-mode`


ericdallo (approved):
Looks good to me!

meliache:
Seems the ubuntu and macos tests for the snapshot are failing, but I don't see yet why :thinking:

PR #3859, opened 2022-12-16
Fix CI

jcs090218:
The redefined error is caused by `ert-runner` already loaded the test files for us, simply remove `$(LOAD-TEST-FILES` can resolve this issue.

```diff
-	$(EASK) exec ert-runner -L . -L clients	$(LOAD-TEST-FILES) -t '!no-win' -t '!org'
+	$(EASK) exec ert-runner -L . -L clients -t '!no-win' -t '!org'
```

Also few minor fix for styling, `point-at-bol`/`point-at-eol` obsolete, etc.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -184,7 +184,7 @@ This will help minimize popup flickering issue in `company-mode'."
 
 (defun lsp-completion--looking-back-trigger-characterp (trigger-characters)
   "Return character if text before point match any of the TRIGGER-CHARACTERS."
-  (unless (= (point) (point-at-bol))
+  (unless (= (point) (beginning-of-line))
     (seq-some
      (lambda (trigger-char)
        (and (equal (buffer-substring-no-properties (- (point) (length trigger-char)) (point))
--- a/lsp-dired.el
+++ b/lsp-dired.el
@@ -72,12 +72,11 @@
                    (if (dired-move-to-filename nil)
                        (let* ((file (dired-get-filename nil t))
                               (bol (progn
-                                     (point-at-bol)
                                      (search-forward-regexp "^[[:space:]]*" (line-end-position) t)
                                      (point)))
                               (face (lsp-dired--face-for-path file)))
                          (when face
-                           (-doto (make-overlay bol (point-at-eol))
+                           (-doto (make-overlay bol (line-end-position))
                              (overlay-put 'evaporate t)
                              (overlay-put 'face face))))
                      (cl-return-from :file nil))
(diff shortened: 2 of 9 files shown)


lsp-dired.el
```diff
@@ -72,12 +72,11 @@
                    (if (dired-move-to-filename nil)
                        (let* ((file (dired-get-filename nil t))
                               (bol (progn
-                                     (point-at-bol)
```
jcs090218, line 75: This does nothing, should we replace it with `(beginning-of-line)`?

yyoncho (approved):

yyoncho:
Thank you!

lsp-completion.el
```diff
@@ -184,7 +184,7 @@ This will help minimize popup flickering issue in `company-mode'."
 
 (defun lsp-completion--looking-back-trigger-characterp (trigger-characters)
   "Return character if text before point match any of the TRIGGER-CHARACTERS."
-  (unless (= (point) (point-at-bol))
+  (unless (= (point) (beginning-of-line))
```
danp, line 187: Just updated to get this change and now getting this backtrace when I try completing in my Go project:

```
Debugger entered--Lisp error: (wrong-type-argument number-or-marker-p nil)
  =(118 nil)
  (if (= (point) (beginning-of-line)) nil (seq-some #'(lambda (trigger-char) (and (equal (buffer-substring-no-properties (- ... ...) (point)) trigger-char) trigger-char)) trigger-characters))
  lsp-completion--looking-back-trigger-characterp(["."])
  lsp-completion-at-point()
  completion--capf-wrapper(lsp-completion-at-point optimist)
  company--capf-wrapper(lsp-completion-at-point optimist)
  company--capf-data-real()
  company--capf-data()
  company-capf(prefix)
  apply(company-capf prefix)
  company-call-backend-raw(prefix)
  company--force-sync(company-call-backend-raw (prefix) company-capf)
  company-call-backend(prefix)
  company--begin-new()
  company--perform()
  company-auto-begin()
  company-idle-begin(#<buffer main.go> #<window 9 on main.go> 102 118)
  apply(company-idle-begin (#<buffer main.go> #<window 9 on main.go> 102 118))
  timer-event-handler([t 25502 835 435308 nil company-idle-begin (#<buffer main.go> #<window 9 on main.go> 102 118) nil 0 nil])
```

If I change back to `point-at-bol` it works. (The top of the backtrace here is maybe funky because I eval'd to test changing back to `point-at-bol`, then eval'd again using `beginning-of-line` to get the backtrace to paste here)

jcs090218: Ops, it should be `line-beginning-position`.

danp: 👍 Tried that and it does work.

yyoncho:
I quickly reverted the PR until we have a solution.

PR #3860, opened 2022-12-17
json language server: Fix the binary name

psibi:
The binary name currently isn't used correct and this patch helps to fix it. The current situation makes the JSON language server present in my $PATH unusable. 

For reference, this is the confirmed binary named in the package.json
file: https://github.com/microsoft/vscode/blob/09c0b1f799d75edf284bc89c824bdd724c5b0fa7/extensions/json-language-features/server/package.json#L11

### DIFF
--- a/clients/lsp-json.el
+++ b/clients/lsp-json.el
@@ -98,7 +98,7 @@
                   (list callback))))
 
 (lsp-dependency 'vscode-json-languageserver
-                '(:system "vscode-json-language-server")
+                '(:system "vscode-json-languageserver")
                 '(:npm :package "vscode-langservers-extracted"
                        :path "vscode-json-language-server"))
 


clients/lsp-json.el
```diff
@@ -98,7 +98,7 @@
                   (list callback))))
 
 (lsp-dependency 'vscode-json-languageserver
-                '(:system "vscode-json-language-server")
+                '(:system "vscode-json-languageserver")
```
ericdallo, line 101: if the binary was renamed in some release we could support both adding two `:system`?

ericdallo (approved):

yyoncho (approved):

clients/lsp-json.el
psibi: @ericdallo Maybe, but I'm not sure. This is the first time I'm trying to use JSON language server and I got hit with this.

Also, I'm not able to find any CHANGELOG associated with the vscode project.

ericdallo: I think we can go ahead and if anyone complain we can add back again the old system

ericdallo:
Thank you

PR #3863, opened 2022-12-17
Fix from #3859

jcs090218:


yyoncho:
can you rebase? (I reverted the original PR to give some more time for testing).

PR #3867, opened 2022-12-19
Fixe makdown help rendering

vspinu:
- Fixes emacs-lsp/lsp-ui#742
 - Use with-demoted-errors instead of the condition case
 - Display the processed string regardless of the errors


yyoncho (approved):

yyoncho:
The code looks good to me. @jcs090218 can you test and merge it?

jcs090218 (approved):
Tested, it works!

jcs090218:
@vspinu Thank you for the quick response and came up with the patch! ❤️ 😄

PR #3873, opened 2022-12-23
lsp-languague-id-configuration: enable file-name base dectection for more languages

kiennq:
Using `lsp-activate-on` for:
- rust
- go | go.mod
- cmake
- dockerfile
- csharp

And add adequate pattern matchings for them in `lsp-language-id-configuration`.
All of those are new languages that have the native tree-sitter support, we should switch to using file name-based pattern matching for fewer changes on the id-configuration.

Also, add the indent offset of those new tree-sitter languages to the table as well


yyoncho:
> we should switch to using file name-based pattern matching for fewer changes on the id-configuration.

FWIW we will still need the mapping due to rendering the language for the hover functionality.

yyoncho (approved):

kiennq:
> > we should switch to using file name-based pattern matching for fewer changes on the id-configuration.
> 
> FWIW we will still need the mapping due to rendering the language for the hover functionality.

Hmm, perhaps we can do it in a round about way of converting the hover content to markdown by wrapping it in "``` language" tag? Not sure how feasible that approach is though

PR #3875, opened 2022-12-23
Add python-ts-mode to supported modes for pyls and pylsp

bastianbeischer:


jcs090218 (approved):

bastianbeischer:
Thanks @jcs090218. I received a comment from @yyoncho on https://github.com/emacs-lsp/emacs-ccls/pull/117 that it is better to use lsp-activate-on. So I went ahead and did this here as well.

PR #3883, opened 2022-12-30
Add elixir-ts-mode and heex-ts-mode

wkirschbaum:
This is for https://github.com/wkirschbaum/elixir-ts-mode

Both elixir-ts-mode and heex-ts-mode uses the same language server as elixir-mode.

### DIFF
--- a/clients/lsp-elixir.el
+++ b/clients/lsp-elixir.el
@@ -192,7 +192,6 @@ be available here: https://github.com/elixir-lsp/elixir-ls/releases/"
                                             "language_server.bat")
                                        ,@(cl-rest lsp-elixir-server-command))))
                   :activation-fn (lsp-activate-on "elixir")
-                  :major-modes '(elixir-mode elixir-ts-mode heex-ts-mode)
                   :priority -1
                   :server-id 'elixir-ls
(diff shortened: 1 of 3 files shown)


yyoncho (approved):

clients/lsp-elixir.el
```diff
@@ -192,6 +192,7 @@ be available here: https://github.com/elixir-lsp/elixir-ls/releases/"
                                             "language_server.bat")
                                        ,@(cl-rest lsp-elixir-server-command))))
                   :activation-fn (lsp-activate-on "elixir")
+                  :major-modes '(elixir-mode elixir-ts-mode heex-ts-mode)
```
kiennq, line 195: nit: this might not be needed as `lsp-elixir` will be activated when the language is `elixir` (the `cdr` part of the cons you added bellow) already.

kiennq (approved):

clients/lsp-elixir.el
wkirschbaum: thanks, makes sense. i will remove this now.

jcs090218 (approved):

wkirschbaum:
I updated the Changelog as well

Issue #3881, opened 2022-12-29
Semantic tokens mode issue in lua-mode


drcxd:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

![image](https://user-images.githubusercontent.com/24535087/209894631-929bb6cd-fe7a-4950-bc92-ade101d453d2.png)

As you can see in the picture, `m` and `test` are rendered using one color before the long comment, but after the long comment, the same tokens are rendered using different colors.

Using `describe-char` to check the properties of `test`, I have found that those before the long comment have text properties "property", while those after the long comment don't.

I first raise the issue under the language server I used, that is lua-language-server. (Link to the issue https://github.com/sumneko/lua-language-server/issues/1753). However, people using VSCode and NeoVim all report all the four `test` in my example code are labeled as "property". Thus, I doubt this might be a lsp-mode issue?

### Steps to reproduce

I am using Emacs 29.0.50 on Windows. 
lsp-mode version is 20221228.2109. 
lua-mode version is 20221218.605. 
lua-language-server version is 3.6.4.
I have used a clean configuration and reproduce the bug as the picture descripted.
The only configuration code I have is to set `package-archives`, `lsp-clients-lua-language-server-bin` and `lsp-clients-lua-language-server-main-location`.
No other packages are installed.
Just open a lua file and enable lsp-mode and lsp-semantic-tokens-mode, the bug could be reproduced.

### Expected behavior

Long comment should not affect the behavior of semantic tokens highlighting.

### Which Language Server did you use?

lua-language-server

### OS

Windows

yyoncho:
You have lsp-semantic-tokens-enable set to t, right?

drcxd:
@yyoncho No, it is nil. I  execute `lsp-semantic-tokens-mode` to enable it. Setting it to `t` does not change the behavior.

yyoncho:
@drcxd you should change that before running M-x lsp for the first time.

drcxd:
@yyoncho I have done what you said.
```
(setq lsp-clients-lua-language-server-bin "e:/lls/bin/lua-language-server.exe"
[...]

PR #3886, opened 2023-01-01
Report multiline token support, ...

sebastiansturm:
... truncate tokens to current line if multiline token support is disabled, as required by the lsp spec. Also, report overlappingTokenSupport.

This should fix issue #3881, though even if we report multilineTokenSupport, lua-language-server still reports out-of-bounds token lengths which I think is an error on the server side. With `lsp-semantic-tokens-enable-multiline-token-support` set to `nil`, however, lsp-mode will now truncate token ends to the end-of-line position and should thus produce the highlighting intended by the lua language server.


drcxd:
@sebastiansturm This PR is still not murged...

yyoncho (approved):

PR #3893, opened 2023-01-05
add :features to checkOnSave and use original descriptions

brotzeit:


jcs090218 (approved):

otavio (approved):

yyoncho (approved):

brotzeit:
rust-analyzer changes configuration options very frequently. In the past there was `:allTargets`. It was obviously removed.

ia0:
> It was obviously removed.

What does this mean? Maybe there's a confusion because `checkOnSave` was renamed to `check`, but `allTargets` [still exists](https://github.com/rust-lang/rust-analyzer/blob/f5401f620699b26ed9d47a1d2e838143a18dbe3b/crates/rust-analyzer/src/config.rs#L135) as far as I can tell.

I would suggest adding `allTargets` back.

(Ideally we should probably also rename `checkOnSave` to `check` to avoid relying on rust-analyzer doing the renaming for us.)

PR #3901, opened 2023-01-08
lsp-json: Multiple optional system dependency for json language server

psibi:


yyoncho (approved):

Manueljlin:
Doesn't quite work for me for some reason. I'm guessing it doesn't have to do with using jsonian-mode instead of json-mode since it worked before.
Screenshot_20230108_150331

Manueljlin:
oh nevermind, I can reproduce it with the master branch too. it does work with melpa for some reason.

psibi:
@Manueljlin Nice, thanks for the confirmation!

PR #3904, opened 2023-01-10
Add function for removing all folders in workspace

ishai1:
Add function for removing all folders in workspace.


yyoncho:
Can you use prefix argument so C-0 M-x lsp-workspace-folder-remove does this job?

ericdallo:
@yyoncho don't you think it makes harder for users find that if we use prefix argument?
I suppose the user would need to read the function doc of this function to know that with a prefix arg would remove all

yyoncho (approved):

PR #3909, opened 2023-01-15
Allow using the project's tsserver.js

shepmaster:
By using the project's tsserver.js, TypeScript language service plugins specified in the tsconfig.json will automatically be loaded.

VS Code has [a similar option][1], as mentioned by the typescript-plugin-css-modules plugin's [recommendation][2].

[1]: https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript
[2]: https://github.com/mrmckeb/typescript-plugin-css-modules#recommended-usage

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -118,6 +118,12 @@ exceed allowed memory usage."
   :group 'lsp-typescript
   :type 'string)
 
+(defcustom lsp-clients-typescript-prefer-use-project-ts-server nil
+  "When set, prefers using the tsserver.js from your project. This
+can allow loading plugins configured in your tsconfig.json."
+  :group 'lsp-typescript
+  :type 'boolean)
+
 (defcustom lsp-clients-typescript-plugins (vector)
   "The list of plugins to load.
 It should be a vector of plist with keys `:location' and `:name'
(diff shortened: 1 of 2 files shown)


shepmaster:
I'm marking this as draft as I haven't added the changelog entry. Once someone thinks this is a reasonable feature, I can do that.

clients/lsp-javascript.el
```diff
@@ -118,6 +118,12 @@ exceed allowed memory usage."
   :group 'lsp-typescript
   :type 'string)
 
+(defcustom lsp-clients-typescript-prefer-use-project-ts-server nil
```
yyoncho, line 121: What is the vscode default here? If it is true I prefer going with t as default.

clients/lsp-javascript.el
```diff
@@ -892,11 +898,23 @@ name (e.g. `data' variable passed as `data' parameter)."
     (when lsp-javascript-update-inlay-hints-on-scroll
       (setf window-scroll-functions (delete #'lsp-javascript-update-inlay-hints-scroll-function window-scroll-functions))))))
 
+(defun lsp-clients-typescript-project-ts-server-path ()
+  (f-join (lsp-workspace-root) "node_modules" "typescript" "lib" "tsserver.js"))
+
+(defun lsp-clients-typescript-server-path ()
```
yyoncho, line 904: I wish we somehow log this value.

yyoncho (commented):
Looks good to me.

clients/lsp-javascript.el
shepmaster: I'd be happy to add something more explicit, but it's indirectly already logged (see the `--tsserver-path` argument):

```
Command "/Users/shep/.emacs.d/.cache/lsp/npm/typescript-language-server/bin/typescript-language-server --tsserver-path /Users/shep/.emacs.d/.cache/lsp/npm/typescript/bin/tsserver --stdio" is present on the path.
```

clients/lsp-javascript.el
shepmaster: > VS Code ships with a recent stable version of the TypeScript language service and uses this by default to provide IntelliSense in your workspace. 
>
> - [Using newer TypeScript versions](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-newer-typescript-versions)

Based on that, I believe that the VS Code default corresponds to our `nil`. 

This makes sense to me from the "just get something that works" perspective. VS Code (and lsp-mode) can install a newer and ideally known-working version of typescript (and the language server) and use those. Users who need something more special can step outside that.

yyoncho: Sounds good. Feel free to convert it from draft to PR.

shepmaster:
Changelog entry added and marked as ready to review ❤️

yyoncho (approved):

yyoncho:
Thank you!

PR #3912, opened 2023-01-16
headerline: Speed up root detection

### LINKED ISSUE #3911: headerline updating is slow (with explanation and fix)
#### Description
headerline updating, which is called frequently via `lsp--on-idle-timer`, can be slow.

#### Reason
`lsp-headerline--build-string` usually calls `lsp-headerline--build-project-string` or `lsp-headerline--build-path-up-to-project-string`, which call `lsp-workspace-root`.
`lsp-workspace-root` is very slow when the lsp session contains many folders, because it iterates all folders.

#### Possible fix
Rely on field `root` of the workspace object associated with the buffer.

```elisp
(defun lsp-headerline--workspace-root ()
  (--if-let (car lsp--buffer-workspaces)
      (lsp--workspace-root it)))

;; Alternative: Handle the case of multiple workspaces per buffer
;; I'm not sure if this is required as I'm not familiar with lsp's architecture.
(defun lsp-headerline--workspace-root ()
  (->> lsp--buffer-workspaces
       (-map 'lsp--workspace-root)
       (--max-by (> (length it) (length other)))))
```

erikarvstedt:
Fixes #3911

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -250,7 +250,7 @@ PATH is the current folder to be checked."
 
 (defun lsp-headerline--build-project-string ()
   "Build the project-segment string for the breadcrumb."
-  (-if-let (root (lsp-workspace-root))
+  (-if-let (root (lsp-headerline--workspace-root))
       (propertize (lsp-headerline--directory-with-action
                    root
                    (f-filename root))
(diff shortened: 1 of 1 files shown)


lsp-headerline.el
```diff
@@ -413,6 +413,10 @@ PATH is the current folder to be checked."
   "Disable headerline breadcrumb mode."
   (lsp-headerline-breadcrumb-mode -1))
 
+(defun lsp-headerline--workspace-root ()
+  (--if-let (car lsp--buffer-workspaces)
```
yyoncho, line 417: can you change it to --when-let?

ericdallo (approved):

yyoncho:
Thank you!

PR #3913, opened 2023-01-16
headerline: Fix error for files with symlinks

erikarvstedt:
#### Copy of main commit msg

`lsp--workspace-root` (returned by `lsp-headerline--workspace-root`) always stores a fully resolved path with no symlinks.

When `(buffer-file-name)` includes a symlinked path to the workspace root, `root` is not a prefix of `(buffer-file-name)`.
In this case, the path parents traversing in `lsp-headerline--path-up-to-project-root` doesn't stop at root
and ultimately fails with a `(wrong-type-argument stringp nil)` error when "/" is reached.

Fix this by resolving symlinks in `(buffer-file-name)`.

Also simply use `file-name-directory` because `(buffer-file-name)` is always an absolute path to a file.

#### Discussion

The bug fixed here is a result of 16b13adf857f9cd2e7d48a1bfe3c0b1ed4107339.
Before 16b13adf857f9cd2e7d48a1bfe3c0b1ed4107339, the workspace root was never detected for buffers whose file name pointed to a symlinked root. In this case, `lsp-headerline--build-path-up-to-project-string` exited early.

You can reproduce the bug like so:
- Create a project in `/tmp/repro/myproject`
- Create a symlink `/tmp/repro/mylink` -> `/tmp/repro/myproject`
- Open a file in `/tmp/repro/mylink` with `lsp-mode`

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -310,7 +310,7 @@ PATH is the current folder to be checked."
                           (setq lsp-headerline--path-up-to-project-segments
                                 (list (lsp-headerline--path-up-to-project-root
                                        root
-                                       (lsp-f-parent (buffer-file-name))))))
+                                       (file-name-directory (file-truename (buffer-file-name)))))))
                         (car lsp-headerline--path-up-to-project-segments))))
         (mapconcat (lambda (next-dir)
                      (propertize next-dir
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -305,12 +305,13 @@ PATH is the current folder to be checked."
 (defun lsp-headerline--build-path-up-to-project-string ()
   "Build the path-up-to-project segment for the breadcrumb."
   (if-let ((root (lsp-headerline--workspace-root)))
-      (let ((segments (or
-                       lsp-headerline--path-up-to-project-segments
-                       (setq lsp-headerline--path-up-to-project-segments
-                             (lsp-headerline--path-up-to-project-root
-                              root
-                              (lsp-f-parent (buffer-file-name)))))))
+      (let ((segments (progn
+                        (unless lsp-headerline--path-up-to-project-segments
+                          (setq lsp-headerline--path-up-to-project-segments
+                                (list (lsp-headerline--path-up-to-project-root
+                                       root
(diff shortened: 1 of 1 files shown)


lsp-headerline.el
```diff
@@ -305,12 +305,13 @@ PATH is the current folder to be checked."
...
-      (let ((segments (or
-                       lsp-headerline--path-up-to-project-segments
-                       (setq lsp-headerline--path-up-to-project-segments
-                             (lsp-headerline--path-up-to-project-root
-                              root
-                              (lsp-f-parent (buffer-file-name)))))))
+      (let ((segments (progn
+                        (unless lsp-headerline--path-up-to-project-segments
+                          (setq lsp-headerline--path-up-to-project-segments
+                                (list (lsp-headerline--path-up-to-project-root
```
yyoncho, line 311: I don't understand why we introduce this `list` here. Seems like the list can go away so the `car` on 313.

erikarvstedt: Previously, we couldn't differentiate whether `lsp-headerline--path-up-to-project-root` returned no segments (`nil`) or if the caching variable was uninitialized (also `nil`).

yyoncho (approved):

gustav-b:
I'm able to reproduce this (or something similar) with lsp-mode `20230227.1358` following the steps in the description. This is with `GNU Emacs 28.2 (build 2, x86_64-apple-darwin22.1.0)`.

Here's `*Backtrace*`:

```
  f-exists\?(nil)
  lsp-f-same\?("/private/tmp/mylink/lib" nil)
  lsp-headerline--path-up-to-project-root("/private/tmp/mylink/lib" "/private/tmp/myproject/lib/api/")
  lsp-headerline--build-path-up-to-project-string()
  #<subr F616e6f6e796d6f75732d6c616d626461_anonymous_lambda_23>(path-up-to-project)
  lsp-headerline--build-string()
  lsp-headerline--check-breadcrumb()
  lsp--on-idle(#<buffer api.go>)
  apply(lsp--on-idle #<buffer api.go>)
  timer-event-handler([t 0 0 500000 nil lsp--on-idle (#<buffer api.go>) idle 0 nil])
```

Issue #3876, opened 2022-12-24
Emacs lsp client for python ruff language server


vikigenius:
So ruff is an extremely fast linter for python.

There already is a language server for python-ruff https://github.com/charliermarsh/ruff-lsp. This is similar in functionality to the eslint language server and would probably be used in conjunction with pylsp for completions similar to how eslint lsp is used alongside the ts language server.

Here is the relevant issue on the repository https://github.com/charliermarsh/ruff-lsp/issues/19

It would be great if we could have a client for ruff-lsp so that we can leverage the extremely fast linting.

rassie:
I would like to invoke Cunningham's law and post the configuration I'm using (which doesn't work too well, but it does nevertheless):
```
(defcustom lsp-ruff-executable "ruff-lsp"
  "Command to start the Ruff language server."
  :group 'lsp-python
  :risky t
  :type 'file)

(lsp-register-client
 (make-lsp-client
  :new-connection (lsp-stdio-connection (lambda () (list lsp-ruff-executable "")))
  :activation-fn (lsp-activate-on "python")
  :add-on? t
  :server-id 'ruff
  :initialization-options (lambda ()
                              (append (list :interpreter (list (concat poetry-project-venv "/bin/python3")))))))
```

I've tried to accomodate `poetry` support with the `:interpreter` initialization options, it's fairly obvious that it needs a proper fallback for non-`poetry` projects and probably quite a bunch more (global binaries, `venv` binaries etc.) But maybe someone who is more fluent in Elisp can create a proper configuration.

rassie:
Thanks to @yyoncho, current version looks like this:

```
(defcustom lsp-ruff-executable "ruff-lsp"
  "Command to start the Ruff language server."
  :group 'lsp-python
  :risky t
  :type 'file)

;; Register ruff-lsp with the LSP client.
(lsp-register-client
    (make-lsp-client
        :new-connection (lsp-stdio-connection (lambda () (list lsp-ruff-executable)))
        :activation-fn (lsp-activate-on "python")
        :add-on? t
        :server-id 'ruff
        :initialization-options (lambda ()
                                    (list :settings
                                        (cl-list*
                                          (when
                                            poetry-project-venv
                                                (list
                                                :interpreter (vector (f-join (f-long poetry-project-venv) "bin" "python3"))
                                                :workspace (f-long poetry-project-venv)
                                                :path (vector (f-join (f-long poetry-project-venv) "bin" "ruff")))
[...]

PR #3926, opened 2023-01-23
Added more LSP support for Python by adding ruff-lsp support.

frejanordsiek:
As it says on the tin, adds support for ruff-lsp for Python (https://github.com/charliermarsh/ruff-lsp) including adding it to the documentation (not completely sure I got it right).

ruff-lsp was brought up in Issue #3876 where others have been working on it and had some success in getting it working. 

This implementation provides customization options for every setting ruff-lsp has but one (showNotification since I am not sure what all values it supports). It is heavily based off of clients/lsp-pylsp.el, with some ideas from the other ones. Note, unlike the version from Nikolai Prokoschenko (@rassie) in Issue #3876, this does not have poetry support.

I set its priority to -2, putting it behind pylsp which most people seem to use and prefer (no surprises for users).


clients/lsp-ruff-lsp.el
```diff
@@ -0,0 +1,95 @@
...
+  :priority -2
+  :initialization-options
+  (lambda ()
+    (list :settings
+          (list :args lsp-ruff-lsp-ruff-args
+                :logLevel lsp-ruff-lsp-log-level
+                :path lsp-ruff-lsp-ruff-path
+                :interpreter (vector lsp-ruff-lsp-python-path)
+                :organizeImports (if lsp-ruff-lsp-advertize-organize-imports t :json-false)
+                :fixAll (if lsp-ruff-lsp-advertize-fix-all t :json-false))))))
```
yyoncho, line 90: nit: lsp-json-bool

clients/lsp-ruff-lsp.el
```diff
@@ -0,0 +1,95 @@
...
+  :type 'boolean
+  :group 'lsp-ruff-lsp)
+
+(defcustom lsp-ruff-lsp-advertize-fix-all t
+  "Whether to report ability to handle source.fixAll actions."
+  :type 'boolean
+  :group 'lsp-ruff-lsp)
+
+
+(lsp-register-client
```
yyoncho, line 75: shouldn't that be add-on? = t?

clients/lsp-ruff-lsp.el
frejanordsiek: Didn't realize that existed (not a surprise). Thank you. I have changed the code to use it.

clients/lsp-ruff-lsp.el
frejanordsiek: I don't know. Reading the FAQ, it seems like ``add-on?`` being ``t`` means load in parallel to other LSPs at the same priority? Is compatibility something to worry about? Only a few of the client modules use it:

* lsp-eslint.el
* lsp-emmet.el
* lsp-sorbet.el
* lsp-angular.el
* lsp-camel.el

None of the python ones do at present (though, of course, that doesn't mean that they shouldn't).

So basically, I don't know enough about the option to know.

rassie: Ruff doesn't provide anything you'd like a language server to have, e.g. code navigation, completion etc., it's just a linter. So running it standalone is usually not what you want, but pairing it with e.g. pylsp is extremely natural -- you'd have two servers running, both providing code actions and diagnostics at the same time (e.g. pylsp would provide mypy diagnostics, while ruff would be linting). It's exactly what we want and need (like I said, it's basically the same situation as with eslint). We haven't seen addon LSPs with Python yet, since pylint and pyflakes and flake8 and all the other linters have been typically used as plugins for pylsp, but ruff is mostly used standalone (but there is a plugin for pylsp too).

frejanordsiek: So that is why the other python LSPs have not done it. Makes sense. Speaking of, the option to enable ruff for pylsp does need to have its customization options added but that is for another day.

Thank you for the explanations.

frejanordsiek:
I rebased the PR, added the ``showNotifications`` option, added the missing values for the ``logLevel`` option, and added the missing ``importStrategy`` option (it is declared in the ruff-lsp code as well as its vscode plugin, but not in ruff-lsp's README).

frejanordsiek:
I do not know why the check on macos-latest with emacs 26.3 is failing

nasyxx:
Any update for this?  Will it merge to the main branch soon?

yyoncho:
@frejanordsiek thank you!

PR #3929, opened 2023-01-25
fix: Only detect CSS when .css is final extension

amake:
Prevents misidentifying e.g. *.css.ts, which is a convention used by for instance [vanilla-extract](https://vanilla-extract.style/).


yyoncho (approved):

yyoncho:
Thank you!

Issue #3918, opened 2023-01-20
ruby-ts-mode support


AndrewSwerlick:
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)?

No

Describe your feature here.

With emacs 29 coming, there are a number of new major modes that use the new treesitter parsers. I see that many of these new modes have been added the lsp, but not ruby-ts-mode. I've made a personal fork where I've got this working, and I'd be happy to submit a pull request if there's interest.

justinbarclay:
I'm running into this issue right now

yyoncho:
@AndrewSwerlick forgot the answer here - feel free to open a PR.

justinbarclay:
I think this can be closed now? At least, it works on my machine 🎉

PR #3941, opened 2023-02-02
Add ruby-ts-mode support

justinbarclay:
Update supported major mode for all ruby lsp clients to include `ruby-ts-mode`.

This is in response to #3918


yyoncho:
There is an alternative approach: to remove all `:major-modes` and replace them with `:activation-fn (lsp-activate-on "ruby")`. Let me know if you want to do that change.

justinbarclay:
Yeah I'd be happy to try that out

yyoncho (approved):

yyoncho:
Thank you for taking care of this!

justinbarclay:
I've gotten so much out of this your, and the communities work here, that it is the least I could do ♥️

PR #3946, opened 2023-02-04
Handle string inlay hints

### LINKED ISSUE #3945: More rust-analyzer inlay hints

### Bug description

With the change in to fix #3896, _most_ of the inlay hints now appear - but the ones that were appearing before are now gone.  This is because rust-analyzer only mostly produces inlay hints whose labels are vectors.  A few are still strings.

For example:

image

That second binding has an inlay hint whose value is just `": &str"`, not a vector.

### Steps to reproduce

Uncertain; I'm not sure of the conditions which cause rust-analyzer to do this.  I've seen cases (like the one above) where what looks like almost exactly the same kind of code produces a string label in one case and a vector label in another.  In the file that generated the image above, there are 54 inlay hints processed by `lsp-rust-analyzer-update-inlay-hints`, of which only 6 are bare strings.

### Expected behavior

For all the inlay hints to show up.

lsp-rust

### OS

Linux

### Error callstack

_No response_

### Anything else?

A fix for this is obvious; I'll be making a PR for it in a few seconds.

rjmac:
Also show a message if we get an inlay hint of an unexpected type.  Fixes #3945

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -1323,6 +1323,8 @@ meaning."
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "8.0.0"))
 
+(defvar lsp-rust-analyzer-already-warned-about-inlay-hint-type nil)
+
 (defun lsp-rust-analyzer-update-inlay-hints (buffer)
   (if (and (lsp-rust-analyzer-initialized?)
            (eq buffer (current-buffer)))
@@ -1340,12 +1342,21 @@ meaning."
                    (pos (lsp--position-to-point position))
                    (overlay (make-overlay pos pos nil 'front-advance 'end-advance)))
              (let ((concat-label
-                    (string-join
-                     (mapcar
-                      (lambda (label)
-                        (when (lsp-structure-p label)
-                          (gethash "value" label "")))
-                      label))))
+                    (cl-typecase label
+                      (vector
+                       (string-join
+                        (mapcar
+                         (lambda (label)
+                           (when (lsp-structure-p label)
+                             (gethash "value" label "")))
+                         label)))
+                      (string
+                       label)
+                      (t
+                       (unless lsp-rust-analyzer-already-warned-about-inlay-hint-type
+                         (message "Unexpected type for inlay hint: %s" (type-of label))
+                         (setq lsp-rust-analyzer-already-warned-about-inlay-hint-type t))
+                       ""))))
                (overlay-put overlay 'lsp-rust-analyzer-inlay-hint t)
                (overlay-put overlay 'before-string
                             (format "%s%s%s"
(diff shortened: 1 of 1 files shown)


clients/lsp-rust.el
```diff
@@ -1340,12 +1342,21 @@ meaning."
...
+                           (when (lsp-structure-p label)
+                             (gethash "value" label "")))
+                         label)))
+                      (string
+                       label)
+                      (t
+                       (unless lsp-rust-analyzer-already-warned-about-inlay-hint-type
+                         (message "Unexpected type for inlay hint: %s" (type-of label))
+                         (setq lsp-rust-analyzer-already-warned-about-inlay-hint-type t))
+                       ""))))
```
brotzeit, line 1359: Do we really need this ? Isn't it enough to just ignore this case ?

npajkovsky: I like the idea that it'll print out at least some message.

clients/lsp-rust.el
```diff
@@ -1340,12 +1342,21 @@ meaning."
...
+                      (vector
+                       (string-join
+                        (mapcar
+                         (lambda (label)
+                           (when (lsp-structure-p label)
+                             (gethash "value" label "")))
+                         label)))
+                      (string
+                       label)
+                      (t
```
yyoncho, line 1355: this `t` can be replaced with `(not lsp-rust-analyzer-already-war...)`

rjmac: No, because this `t` is a type, not an expression.  Replacing it with that will cause `Unknown type lsp-rust-analyzer-already-warned-about-inlay-hint-type` if neither of the first two branches match.

clients/lsp-rust.el
rjmac: Yeah, that was my thought too.  Silent failure is really annoying to figure out.

clients/lsp-rust.el
```diff
@@ -1340,12 +1342,21 @@ meaning."
...
-                        (when (lsp-structure-p label)
-                          (gethash "value" label "")))
-                      label))))
+                    (cl-typecase label
+                      (vector
+                       (string-join
+                        (mapcar
+                         (lambda (label)
+                           (when (lsp-structure-p label)
+                             (gethash "value" label "")))
```
yyoncho, line 1351: you should use `(lsp-get label :value)`

clients/lsp-rust.el
yyoncho: Yeah, you are right.

clients/lsp-rust.el
yyoncho: This is needed in order to work fine with plists and hash-tables

npajkovsky: yeah, that was my first lisp code :-/

brotzeit (approved):

PR #3953, opened 2023-02-09
Add bash-ts-mode support

SkydiveMike:


clients/lsp-bash.el
```diff
@@ -68,12 +68,14 @@ See instructions at https://marketplace.visualstudio.com/items?itemName=mads-har
   "Check whether `sh-shell' is sh or bash.
 
 This prevents the Bash server from being turned on in zsh files."
-  (and (eq major-mode 'sh-mode)
+  (and (or (eq major-mode 'sh-mode)
+           (eq major-mode 'bash-ts-mode))
```
jcs090218, line 72: I think we can use `memq` here?

SkydiveMike: yes, and the PR should be updated

jcs090218 (approved):
LGTM

PR #3958, opened 2023-02-14
Advertise utf-32 support

matklad:
LSP currently uses "emacs positions" to specify offsets. I don't fully understand what exactly are positions (and I was already mistaken once thinking they are utf8), but it seems like they are Unicode codepoints basically. So let's just tell the world that we are using unicode. Together with https://github.com/rust-lang/rust-analyzer/pull/14141, this should fix the issue for rust-analyzer at least.


matklad:
https://www.gnu.org/software/emacs/manual/html_node/elisp/Text-Representations.html -- that's the docs I am reading for Emacs. It says 

> Buffer positions are measured in character units.

yyoncho (approved):

yyoncho:
Thank you!

PR #3959, opened 2023-02-14
Fix customization parent group of lsp-volar

boskamp:
`M-x customize-group RET lsp-volar` previously showed itself as parent group. It should display **Language Server (lsp-mode)** instead, as other clients do. Below shows the customize buffer after fix:

grafik


yyoncho:
Thank you!

PR #3968, opened 2023-02-23
feat: Adding tilt lsp client

Konubinix:


clients/lsp-tilt.el
```diff
@@ -0,0 +1,37 @@
...
+
+;;; Commentary:
+
+;; Using tilt mode from https://github.com/Konubinix/tilt-mode
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(add-to-list 'lsp-language-id-configuration '(tiltfile-mode . "tiltfile"))
```
yyoncho, line 29: we typically do that in lsp-mode.el . Also add it in  `lsp-client-packages` to be loaded when lsp-mode starts.

yyoncho (commented):
Looks good. One minor nit.

clients/lsp-tilt.el
Konubinix: I did it. Is it ok that way?

Konubinix:
Is there anything more to be done before this can be approved?

Konubinix:
I rebased the code on top of master

yyoncho (approved):

PR #3974, opened 2023-02-25
Restore lsp-rust-analyzer-cargo-all-targets

ia0:
This partially reverts #3893.

The rust-analyzer.checkOnSave.allTargets configuration still exists, although it is internally renamed by rust-analyzer to rust-analyzer.check.allTargets which is the new version. Maybe the intent was to remove lsp-rust-all-targets which is indeed unused by lsp-rust.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -495,11 +495,11 @@ The command should include `--message=format=json` or similar option."
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "6.2.2"))
 
-(defcustom lsp-rust-analyzer-cargo-all-targets t
-  "Cargo watch all targets or not."
+(defcustom lsp-rust-analyzer-check-all-targets t
+  "Enables --all-targets for `cargo check`."
   :type 'boolean
   :group 'lsp-rust-analyzer
-  :package-version '(lsp-mode . "6.2.2"))
+  :package-version '(lsp-mode . "8.0.2"))
 
 (defcustom lsp-rust-analyzer-checkonsave-features []
   "List of features to activate.
@@ -753,7 +753,7 @@ or JSON objects in `rust-project.json` format."
     :checkOnSave (:enable ,(lsp-json-bool lsp-rust-analyzer-cargo-watch-enable)
                   :command ,lsp-rust-analyzer-cargo-watch-command
                   :extraArgs ,lsp-rust-analyzer-cargo-watch-args
-                  :allTargets ,(lsp-json-bool lsp-rust-analyzer-cargo-all-targets)
+                  :allTargets ,(lsp-json-bool lsp-rust-analyzer-check-all-targets)
                   :features ,lsp-rust-analyzer-checkonsave-features
                   :overrideCommand ,lsp-rust-analyzer-cargo-override-command)
     :files (:exclude ,lsp-rust-analyzer-exclude-globs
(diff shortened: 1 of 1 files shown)


clients/lsp-rust.el
```diff
@@ -495,6 +495,12 @@ The command should include `--message=format=json` or similar option."
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "6.2.2"))
 
+(defcustom lsp-rust-analyzer-cargo-all-targets t
+  "Cargo watch all targets or not."
+  :type 'boolean
+  :group 'lsp-rust-analyzer
+  :package-version '(lsp-mode . "6.2.2"))
```
brotzeit, line 502: The version should be 8.0.2.

ia0: Even if it's a revert? The flag only disappeared for a month or so.

If it counts as a new flag then I'd like to change the name to `lsp-rust-analyzer-check-all-targets` to match the new naming of rust-analyzer.

brotzeit: > If it counts as a new flag then I'd like to change the name to lsp-rust-analyzer-check-all-targets to match the new naming of rust-analyzer.

Probably better.

brotzeit:
> Maybe the intent was to remove lsp-rust-all-targets which is indeed unused by lsp-rust.

I think yes, but it's possible that they changed the path and I didn't realize it https://github.com/rust-lang/rust-analyzer/blame/87d57f51bcf07eb364fcf835f9be987006158961/editors/code/package.json#L705

And then they changed it again https://github.com/rust-lang/rust-analyzer/commit/d2bb62b6a81d26f1e41712e04d4ac760f860d3b3#diff-00784dde03b1be151004f435a4ca74754cb674a9e789840e12760d8f3ac1728fR564

I currently don't use rust so this is really annoying. Hopefully they stop changing the paths all the time.

clients/lsp-rust.el
```diff
@@ -747,6 +753,7 @@ or JSON objects in `rust-project.json` format."
     :checkOnSave (:enable ,(lsp-json-bool lsp-rust-analyzer-cargo-watch-enable)
```
brotzeit, line 753: > although it is internally renamed by rust-analyzer to rust-analyzer.check.allTargets which is the new version

So we also have to replace checkonsave with check right ?

brotzeit: But keep this line for rust-analyzer.checkOnSave

clients/lsp-rust.el
ia0: Sounds good. Done.

clients/lsp-rust.el
ia0: We don't need to rename `checkOnSave` (just yet), they do the renaming internally currently (for backward compatibility). However, eventually we probably want to go over all flags and try to follow the newest convention.

ia0:
> I currently don't use rust so this is really annoying. Hopefully they stop changing the paths all the time.

Yes, I didn't follow all the reasoning behind those renamings, but I know it generated quite some discussions on their side.

I might get some time starting from April to try to go over all flags in lsp-rust if that would be something useful to do.

clients/lsp-rust.el
brotzeit: I see, so it's probably the best solution to wait until they are happy and stop changing it =)

brotzeit:
> I might get some time starting from April to try to go over all flags in lsp-rust if that would be something useful to do.

Sure, thanks! rls was also deprecated a few months so it probably makes sense to remove the rls related code completely ?

yyoncho:
> > I might get some time starting from April to try to go over all flags in lsp-rust if that would be something useful to do.
> 
> Sure, thanks! rls was also deprecated a few months ago so it probably makes sense to remove the rls related code completely ?

I think we can move rls in lsp-rls.el and not load it by default.

PR #3976, opened 2023-02-27
terraform-ls: Prefill required fields and ability to validate on save

psibi:
Also add the corresponding docs and a gif to demonstrate the feature:

terraform-ls-prefill


yyoncho (approved):

clients/lsp-terraform.el
```diff
@@ -128,6 +146,11 @@ Defaults to side following treemacs default."
   (when lsp-terraform-ls-enable-show-reference
     '((experimental . ((showReferencesCommandId . "client.showReferences"))))))
 
+(defun lsp-terraform-ls--init-options ()
+  "Construct initialization options for the lanague server."
```
brotzeit, line 150: typo "lanague"

psibi: Ouch, missed it. Will open a follow up PR, thanks!

PR #3981, opened 2023-03-03
Add renamed rust-analyzer triggerParameterHints action.

jkl1337:
rust-analyzer PR #14147 removes the editor.action VSCode internal command for triggerParameterHints, so the new handler name needs to be added to lsp-rust..


yyoncho:
Then we dont need the method as well? As a side note, we may keep it for some time in order to work for folks using older RA version.

jkl1337:
Ok, well as it stands rust-analyzer is sending a rust-analyzer.triggerParameterHints action, when handled in `lsp--execute-command` it is not found and so it is sent back as a workspace/Command via `lsp-send-execute-command` which just ends up printing a verbose error to the minibuffer.

yyoncho:
I completely misread the diff. Thank you for contributing to `lsp-mode`.

PR #3990, opened 2023-03-11
Defer flymake updates

sebastiansturm:
I sometimes run into situations where language servers send overwhelming numbers of diagnostics (for instance, one way to trigger this is to introduce syntax errors somewhere high up in a TypeScript class checked by ts-ls). Eager insertion of flymake overlays may then render Emacs unusable for interactive editing.

This PR defers flymake updates using a timeout (0.5s by default), and retriggers overlay updates if they are interrupted via user input, using another timeout (of 4s by default)


sebastiansturm:
Just noticed I'm not protecting against buffer switches between timer setup and callback invocation, will do that later today

yyoncho:
Investigate whether you can use lsp-idle timer. it will fit better with the rest of the functionality

PR #3997, opened 2023-03-16
Use link on lsp-latex document page instead of writing description directly

ROCKTAKEY:
Hi, thanks for your maintenance!

I used to write document for lsp-mode for [lsp-latex](https://github.com/ROCKTAKEY/lsp-latex), which is available [here](https://emacs-lsp.github.io/lsp-mode/page/lsp-latex/), but it become very old.
I don't want to maintenance both document,  lsp-mode official document and [lsp-latex README](https://github.com/ROCKTAKEY/lsp-latex), so I remove old document and put link to lsp-latex repository instead.

### DIFF
--- a/docs/manual-language-docs/lsp-latex.md
+++ b/docs/manual-language-docs/lsp-latex.md
@@ -1,36 +1,5 @@
 ---
 root_file: docs/manual-language-docs/lsp-latex.md
 ---
-## Available functions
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:
Thank you!

PR #4004, opened 2023-03-26
fix lsp-javascript-rename-file

### LINKED ISSUE #4001: lsp-javascript-rename-file does mess the imports in files importing renamed file

### Bug description

When using lsp-javascript-rename-file,
every file that imports the renamed file
will have their imports updated badly.
Sometimes there is a missing "
Sometimes the directory is repeated twice : "/modules/modules/module"

From what I see it calls a tsserver function : applyRenameFile. Chances are this function is the cause of the bug, 
It happens on two of my stations.

### Steps to reproduce

Go in a typescript project, open a file, M-x lsp-javascript-rename-file, and check the files that import it. Try some various patterns : rename the file to "directory/directory/file" or "F", usually this is where errors happen.

### Expected behavior

Expected behavior is trivial : you rename a file, the files importing that file should update their import as intended or they should not update at all.

tsls

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

ArnaudValette:
resolve #4001 

_typescript.applyRenameFile is not part of the LSP spec, and is a custom function written by the people of typescript-language-server. [See this conversation about its possible deprecation](https://github.com/typescript-language-server/typescript-language-server/pull/685#pullrequestreview-1294721542)
Another problem that was the cause of the issue, is the fact that it wasn't used as it should : applyRenameFile must be called **after** the file is renamed.
Anyway, the commit uses the proper function [willRenameFile](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles) that must be called **before** the file is renamed.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 8.0.1
+  * Fix lsp-javascript-rename-file bug
   * Add support for clojure-ts-mode in clojure-lsp client
   * terraform-ls now supports prefill requied fields and the ability to validate on save.
   * Ignore =.terraform= and =.terragrunt-cache= directories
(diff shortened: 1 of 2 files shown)


ArnaudValette:
Needs more investigation

PR #4006, opened 2023-03-26
support fsautocomplete dll and remove obsolete code

augustfengd:
ionide has a similar feature, and emacs wants in on the fun. :)


elken (changes requested):
I don't understand why it's redundant, you're attempting to remove a variable that's used without even a deprecation notice.

augustfengd:
Hey @elken, I'm saying it's obsolete because the `lsp-fsharp-server-install-dir` looks like it was a location to install `fsautocomplete` and locate it it afterwards using an old installation method that's been removed. see the https://github.com/emacs-lsp/lsp-mode/commit/2d98a9565328a480f98813a7f0f4669d7cad3639 that I linked in one of the commit messages. This commit is a bit more than a year old, so I'm thinking most people have it's safe to remove this remaining "legacy" code now. 

At worst, I imagine users will just be prompted to reinstall the language server using the new method in the backend.

Issue #4003, opened 2023-03-26
Support crystalline for Crystal


restitux:
It seems `crystalline` has replaced `scry`. The `scry` repo `README` now links to `crystalline` (see https://github.com/elbywan/crystalline). 

It did some basic testing by renaming the `crystalline` binary to `scry`. Most basic features I tried to use seemed to work correctly.

yyoncho:
willing to provide a PR?

restitux:
Sure, would you suggest dropping support for `scry` or supporting both?

restitux:
PR assuming we want to drop `scry`: https://github.com/emacs-lsp/lsp-mode/pull/4007

xendk:
I'll second this.

The current workaround is `(setq lsp-clients-crystal-executable '("crystalline" "--stdio"))`, which works, but would be nice if it just worked out of the box.

PR #4007, opened 2023-03-27
Use crystalline instead of scry for Crystal (#4003)

restitux:
^^


jcs090218:
What's the intention of these changes? Is the scry server abandoned?

PR #4018, opened 2023-04-07
fix: Do not update the headerline of different windows

kassick:
When two windows (A and B) are opened in the same frame, changing the cursor position in winthe A would update the headerline of window B.

This behavior can be visually confusing, specially if windows A and B display different buffers.

This commit updates the headerline code to use window-parameters instead of frame-parameters. This ensures that updating the headerline while navigating window A won't trigger an update in other window's headerline.


yyoncho (approved):

yyoncho:
Thank you for taking care of this. There is a bug report for that, I will mark it as closed.

PR #4022, opened 2023-04-13
Changed to use lldb-vscode as the rust debugger did not start.

yalab:
Hi.

The debugger could not be started with lsp-rust, so it was changed so that it could be started with lldb-vscode.


yyoncho:
This should be configurable. I am marking the PR as closed. Please if you are still interested in this, reopen another one with settings being exposed as defcustom.

PR #4026, opened 2023-04-17
fix(glsl): Require lsp-glsl

jcs090218:
Follow up PR in #4025, forgot to require `lsp-glsl`.

I've organized the list so it fit nicer.


jcs090218:
There is an error with the Windows snapshot, please ignore that job's result.

yyoncho (approved):

PR #4037, opened 2023-05-01
Ignore bazel-* directories outside the root directory

GregOwen:
# Why is this change necessary?

[This previous PR](https://github.com/emacs-lsp/lsp-mode/pull/3768) added a default to ignore `bazel-*` directories that were children of the project root. However, the regex used does not ignore `bazel-*` directories elsewhere in the project repo. This can occur when a git submodule uses bazel.

# What does this change do?

Change the default value of `lsp-file-watch-ignored-directories` to ignore `bazel-*` directories wherever they are found in the repo, not just the root directory.

# How is this change tested?

I tested it locally by adding the new regex using the approach described [here](https://emacs-lsp.github.io/lsp-mode/page/file-watchers/) and I no longer see LSP warning that the number of file watches has exceeded its limit.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -374,7 +374,7 @@ the server has requested that."
     "[/\\\\]autom4te.cache\\'"
     "[/\\\\]\\.reference\\'"
     ;; Bazel
-    "bazel-[^/\\\\]+\\'"
+    "[/\\\\]bazel-[^/\\\\]+\\'"
     ;; CSharp
     "[/\\\\]\\.meta\\'"
     ;; Unity


jcs090218 (approved):

GregOwen:
@jcs090218 I'm not sure how to interpret the error I'm getting in the failing CI/test:
```
eask clean all
Can't read whole string
mingw32-make: *** [makefile:59: clean] Error 1
Error: Process completed with exit code 1.
```

I also don't have access to a Windows machine. Any suggestions on how to proceed?

jcs090218:
> I'm not sure how to interpret the error I'm getting in the failing CI/test:

Don't worry about that! There is an incompatible error in the upstream Emacs' Windows snapshot.

GregOwen:
What's expected from me at this point? Will I need to rebase once the upstream snapshot is fixed?

jcs090218:
> What's expected from me at this point? Will I need to rebase once the upstream snapshot is fixed?

Nope, you are all set! We regularly wait a day or two for other people to review it. :)

GregOwen:
Awesome, thanks for the fast response!

PR #4040, opened 2023-05-03
Introduce lsp-defcustom

yyoncho:
- we used `defcustom-lsp` due to formatting. Now the new version of elisp
formatter works fine

### DIFF
--- a/clients/lsp-ada.el
+++ b/clients/lsp-ada.el
@@ -32,7 +32,7 @@
   :tag "Language Server"
   :package-version '(lsp-mode . "6.2"))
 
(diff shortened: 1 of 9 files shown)


sr9992:
I just pulled this update, and I can't start LSP anymore:
```
Error running timer: (void-variable lsp-ada-project-file)
...
require: Symbol’s value as variable is void: lsp-ada-project-file
```
I don't use ada at all.  I run `(lsp-deferred)` in my C++ mode hook.  Any thoughts as to what I'm doing wrong?  Using Emacs 27.1.

sr9992:
Never mind, clearing out some `.elc` files seemed to get it working again.  Thanks.

PR #4049, opened 2023-05-13
Make lsp-fsharp directory exclusion more configurable

### LINKED ISSUE #4048: Make `lsp-fsharp` more configurable
The F# tooling for `lsp-mode` currently implements a workspace loading system in which this flow is implemented using custom `FsAutoComplete` RPCs: 

1. The client sends a "peek" request (`fsharp/workspacePeek`), which contains a list of directories to exclude from peeking.
2. The server replies with all the F# project files it can find. 
3. The client uses the list of discovered project files to issue a "load" request (`fsharp/workspaceLoad`). 
4. `FsAutoComplete` loads the specified project(s). 
5. Two success messages are issued: one when the projects _start_ loading, the second when all projects finish loading. 

Currently, the list of directories to exclude is hard-coded. This can be an issue depending on the rest of the development toolchain in use. For instance: some Nix tooling creates a directory in which symlinks to dependencies are stored. The `fsharp/workspacePeek` request erroneously searches this directory, which turns up _many_ F# project files; `lsp-fsharp` sends the first success message, but loading never completes. 

This is, admittedly, an edge-case, but it's an extremely simple PR to provide configs for, and serves as a fairly helpful debugging hint, given the way this edge-case causes false success reports.

Gastove:
This PR allows users of this code to add extra directories to be ignored during project loading, which can be extremely useful while also using toolsets like Nix (which may symlink other F# projects in to the repo, thus confounding project loading).

Closes #4048.

### DIFF
--- a/clients/lsp-eslint.el
+++ b/clients/lsp-eslint.el
@@ -42,7 +42,7 @@
   :group 'lsp-eslint
   :package-version '(lsp-mode . "8.0.0"))
 
-(defcustom lsp-eslint-download-url "https://github.com/emacs-lsp/lsp-server-binaries/blob/master/dbaeumer.vscode-eslint-2.2.2.vsix?raw=true"
+(defcustom lsp-eslint-download-url "https://github.com/emacs-lsp/lsp-server-binaries/blob/master/dbaeumer.vscode-eslint-3.0.10.vsix?raw=true"
   "ESLint language server download url."
   :type 'string
   :group 'lsp-eslint


clients/lsp-fsharp.el
```diff
@@ -176,6 +176,11 @@ available, else the globally installed tool."
   :type 'boolean
   :risky t)
 
+(defcustom lsp-fsharp-workspace-extra-exclude-dirs '()
+  "Additional directories to exclude from FsAutoComplete workspace loading / discovery."
+  :group 'lsp-fsharp
+  :type '(repeat string))
```
yyoncho, line 182: it should be lsp-string-vector. Why did you change the default?

Gastove: Oh jeeze, I completely lost track of this. Apologies!

I'll get this updated.

clients/lsp-fsharp.el
```diff
@@ -176,6 +176,11 @@ available, else the globally installed tool."
   :type 'boolean
   :risky t)
 
+(defcustom lsp-fsharp-workspace-extra-exclude-dirs '()
```
jcs090218, line 179: Probably better to just set it to `nil`.

PR #4053, opened 2023-05-14
Make it possible to configure lsp-ansible as `add-on? t`

mpereira:
So that it can work with yamlls.

### DIFF
--- a/clients/lsp-ansible.el
+++ b/clients/lsp-ansible.el
@@ -34,6 +34,15 @@
   :link '(url-link "https://github.com/ansible/ansible-language-server")
   :package-version '(lsp-mode . "8.0.1"))
 
+(defcustom lsp-ansible-add-on? t
+  "Make the client `add-on' so that it works with other language servers.
+`yamlls`is a common one.
+
+Enabled by default."
+  :type 'boolean
+  :group 'lsp-ansible
+  :package-version '(lsp-mode . "8.0.1"))
+
 (defcustom lsp-ansible-language-server-command
   '("ansible-language-server" "--stdio")
   "The command that starts the ansible language server."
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

clients/lsp-ansible.el
```diff
@@ -234,6 +241,7 @@ Pretty print the content of PARAMS."
                             (lsp-package-path 'ansible-language-server))
                        ,@(cl-rest lsp-ansible-language-server-command))))
   :priority 1
+  :add-on? lsp-ansible-add-on?
```
jcs090218, line 244: Maybe we should just set this to `t`? And remove custom variable `lsp-ansible-add-on?`.

mpereira: I'd be ok with that. I chose to make it configurable mostly for backwards compatibility.

jcs090218: Is the ansible server design to use with other language servers? If that's the case, we can just set it to `t`. If no, then the PR is ready to merge.

mpereira: @jcs090218 I force-pushed making the defcustom `t` by default, so that it can still be customized if necessary.

https://github.com/emacs-lsp/lsp-mode/pull/4053/commits/47d79d9c5bcce7d24488d667a3c0a08c16ef78a4

```diff
diff --git a/clients/lsp-ansible.el b/clients/lsp-ansible.el
index be2263bbe..931bf0d50 100644
--- a/clients/lsp-ansible.el
+++ b/clients/lsp-ansible.el
@@ -34,9 +34,11 @@
   :link '(url-link "https://github.com/ansible/ansible-language-server")
   :package-version '(lsp-mode . "8.0.1"))
 
-(defcustom lsp-ansible-add-on? nil
-  "Make the lsp client `add-on' so that it works with other language servers.
-E.g., yamlls."
+(defcustom lsp-ansible-add-on? t
+  "Make the client `add-on' so that it works with other language servers.
+`yamlls`is a common one.
+
+Enabled by default."
   :type 'boolean
   :group 'lsp-ansible
   :package-version '(lsp-mode . "8.0.1"))

```

jcs090218:
Should be good to merge. Thanks!

PR #4070, opened 2023-06-11
chore: Remove downstream deps list

jcs090218:
Do we still need the list? I don't see it being used elsewhere. 🤔

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1992,45 +1992,10 @@ regex in IGNORED-FILES."
 
 
 
-(defvar lsp-downstream-deps
-  '(;; external packages
-    ccls consult-lsp dap-mode helm-lsp lsp-dart lsp-docker lsp-focus lsp-grammarly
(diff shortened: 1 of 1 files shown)


ericdallo:
@yyoncho has context about that

yyoncho:
The idea of this section was to check if downstream deps can be compiled against lsp-mode.

jcs090218:
The list is recorded here now:

https://github.com/emacs-lsp/lsp-mode/blob/f61671be7a1afc66603abb93e594d41cede825da/test/downstream/Eask#L22-L43

 It's better to maintain the list in the Eask file since the downstream list isn't part of the core. 🤔 WDYT?

yyoncho:
> The list is recorded here now:
> 
> https://github.com/emacs-lsp/lsp-mode/blob/f61671be7a1afc66603abb93e594d41cede825da/test/downstream/Eask#L22-L43
> 
> It's better to maintain the list in the Eask file since the downstream list isn't part of the core. thinking WDYT?

As long as it works as expected.

jcs090218:
FYI, I cannot merge this since it stops at the "Enable auto-merge (squash)".

ericdallo (approved):

PR #4065, opened 2023-06-08
Filter lenses that are not visible without converting to Position

yyoncho:

PR #4072, opened 2023-06-14
Fix lsp-point-in-range? call: convert position to point

iitalics:
#4065 removes this call to `lsp--position-to-point` that was necessary. otherwise you get these errors:

```
Debugger entered--Lisp error: (wrong-type-argument hash-table-p (:line 37 :character 0))
  gethash("line" (:line 37 :character 0))
  lsp--position-compare((:line 37 :character 0) #<hash-table equal 2/2 0x15642abe71c5>)
```

### DIFF
--- a/lsp-lens.el
+++ b/lsp-lens.el
@@ -251,7 +251,7 @@ version."
   "Return t if LENS has to be loaded."
   (and (not command?)
        (not pending)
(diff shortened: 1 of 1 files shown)


iitalics:
nevermind, this doesn't fix it

PR #4075, opened 2023-06-18
Configurable nimlsp executable

actondev:
nim has 2 lsp implementations, `nimlangserver` and `nimlsp`, however only the first was configurable for its executable path. Also, the server id is renamed from `nimls` to `nimlsp`


jcs090218 (approved):
LGTM.

yyoncho (approved):

PR #3281, opened 2021-12-25
add ebuild-mode language id configuration to shellscript

epsilon-0:
Signed-off-by: Aisha Tammy <floss@bsd.ac>

PR #4084, opened 2023-06-30
Add ebuild-mode support to lsp-bash

titanofold:
PR #3281 initially added the ebuild-mode language id configuration. However, commit b77aecf added a major mode check that prevents bash-ls from loading up for ebuild-mode.

Ebuilds (Gentoo's package file "format") are explicitly executed within a Bash environment regardless of the user's or dev's preference.


yyoncho (approved):

yyoncho:

PR #4097, opened 2023-07-12
Improve documentation for `lsp-execute-code-action`

Hi-Angel:
`lsp-execute-code-action` is a user-facing function that can be executed e.g. when clangd says "fix available". But as it stands, the function description is too abstract, so unless you know details around lsp-protocol `codeAction`, you won't figure out what it's for. Fix that by clarifying what a user can use it for.

This came up after me trying to find some way to "apply the clangd fix" and not finding anything useful offhand, and then going to gitter and asking a question there.


ericdallo (approved):

ericdallo:
thank you!

PR #4098, opened 2023-07-13
fix(marksman): Update Download URL for Marksman Linux Binaries

ncaq:
The download URL for Marksman Linux binaries has been split by architecture, causing the previous URL to return a 404 error.
To resolve this issue,
the download URL has been updated to the new format. This ensures the correct architecture binary is downloaded properly.


yyoncho (approved):

yyoncho:
Thank you!

PR #4122, opened 2023-08-03
lsp-erlang: Add initial support for erlang-language-platform

alanz:
So it now still supports  [erlang-ls](https://github.com/erlang-ls/erlang_ls), but also [erlang-language-platform](https://github.com/WhatsApp/erlang-language-platform).

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 8.0.1
+  * Add [[https://github.com/WhatsApp/erlang-language-platform][erlang-language-platform]] support in lsp-erlang client.
   * Add [[https://github.com/elixir-tools/credo-language-server][credo-language-server]]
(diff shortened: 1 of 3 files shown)


alanz:
The windows ci job failing looks to be unrelated to this PR.

jcs090218:
Yeah, it's unrelated. There is a bug in that snapshot, so we can just ignore it for now.

yyoncho:
TBO I would prefer having each server into separate file based on the experience with rust-analyzer and rls. Willing to move it into separate one?

alanz:
> Willing to move it into separate one?

If it did split it would have to be into three, lsp-erlang, lsp-erlang-ls, lsp-erlang-elp.

Otherwise existing Erlang users don't know there is a new option, and we need a place for the common elements, such as switching servers.

And I am not sure if the effort warrants it at this stage.

yyoncho (approved):

PR #4126, opened 2023-08-06
fix(lsp-mode.el): Drop support for 26.3

jcs090218:
The CI is reporting errors due to one of our dependency `markdown-mode` has recently dropped support for 26.x. Another solution is to set `markdown-mode` as an optional dependency instead of the required one. 🤔 

cc @yyoncho @ericdallo


ericdallo (approved):
I'm ok dropping support for 26

yyoncho (approved):

PR #4132, opened 2023-08-12
Add client registration to lua-ts-mode and gdscript-ts-mode

lerouxrgd:


jcs090218 (approved):
LGTM! Thanks!

lerouxrgd:
Actually I have pushed force a different commit with a more lsp-mode idiomatic way of doing it I think.

lerouxrgd:
@jcs090218 I have improved this PR a bit since you have approved it, you might want to double check it again.

yyoncho (approved):

PR #4138, opened 2023-08-16
Add configuration to use new experimental ondisk cache for lsp-kotlin

themkat:
Recently a PR in kotlin-language-server for using an on-disk cache was merged: https://github.com/fwcd/kotlin-language-server/pull/337

While it is merged, it is still considered experimental. The PR I have created here adds configuration to lsp-kotlin to be able to use this new on-disk cache. If no configuration is done, it will just use the old in-memory approach to caching. If one uses the on-disk cache, it is probably wise to configure it for each of your projects (maybe using dir-locals, setq-local in hooks, or something else).

### DIFF
--- a/clients/lsp-kotlin.el
+++ b/clients/lsp-kotlin.el
@@ -107,8 +107,8 @@ to Kotlin."
   :package-version '(lsp-mode . "8.0.1"))
 
 ;; cache in this case is the dependency cache. Given as an initialization option.
-(defcustom lsp-kotlin-ondisk-cache-path ""
-  "Path to the ondisk cache if used."
+(defcustom lsp-kotlin-ondisk-cache-path nil
+  "Path to the ondisk cache if used. If lsp-kotlin-ondisk-cache-enabled is t, but path is nil, then the project root is used as a default."
   :type 'string
   :group 'lsp-kotlin)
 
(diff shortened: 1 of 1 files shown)


clients/lsp-kotlin.el
```diff
@@ -260,6 +271,10 @@ to Kotlin."
   :initialized-fn (lambda (workspace)
                     (with-lsp-workspace workspace
                       (lsp--set-configuration (lsp-configuration-section "kotlin"))))
+  :initialization-options (lambda ()
+                            (if lsp-kotlin-ondisk-cache-enabled
+                                (list :storagePath lsp-kotlin-ondisk-cache-path)
+                              (list)))
```
jcs090218, line 277: I think this can just be `nil`? Or:

```elisp
(when lsp-kotlin-ondisk-cache-enabled
  (list :storagePath lsp-kotlin-ondisk-cache-path))
```

themkat: You are right, off course 🙂 For some reason I was convinced that lsp-mode didn't like nil in places like these. Unsure where I got that from 😆

jcs090218 (approved):
LGTM! Other people might want to review this PR, so let's leave this open for a few days! ;)

themkat:
@jcs090218 . Found out after experimenting that the workspace root is a sensible default. It is also what is currently happening in the vscode-kotlin extension. Added this in my latest commit 🙂

yyoncho (approved):

PR #4139, opened 2023-08-17
[lsp-magik] Support JAVA_HOME with trailing slash

sebastiaanspeck:
If `$JAVA_HOME` ends with a `"/"`, which is the case when using OpenJDK e.a, the current code doesn't work. It would set it to `bin/java`. Using the new code, we add `bin/java` to the result of `$JAVA_HOME`. To work for a `$JAVA_HOME` that doesn't end with a `"/"`, we use `(expand-file-name...`, which adds a `"/"`, but only if needed.

### DIFF
--- a/clients/lsp-magik.el
+++ b/clients/lsp-magik.el
@@ -88,8 +88,11 @@
   :group `lsp-magik
   :package-version '(lsp-mode . "8.0.1"))
 
-(defcustom lsp-magik-java-path (cond ((eq system-type 'windows-nt) "$JAVA_HOME/bin/java")
-                                     (t "java"))
+(defcustom lsp-magik-java-path (lambda ()
+                                 (cond ((eq system-type 'windows-nt)
+                                        (or (lsp-resolve-value (executable-find (expand-file-name "bin/java" (getenv "JAVA_HOME"))))
+                                            (lsp-resolve-value (executable-find "java"))))
+                                       (t "java")))
   "Path of the java executable."
   :type 'string
   :group `lsp-magik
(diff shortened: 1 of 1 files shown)


sebastiaanspeck:
Can this PR please get a review?

clients/lsp-magik.el
```diff
@@ -88,7 +88,8 @@
   :group `lsp-magik
   :package-version '(lsp-mode . "8.0.1"))
 
-(defcustom lsp-magik-java-path (cond ((eq system-type 'windows-nt) "$JAVA_HOME/bin/java")
+(defcustom lsp-magik-java-path (cond ((eq system-type 'windows-nt) (or (executable-find (expand-file-name "bin/java" (getenv "JAVA_HOME")))
```
yyoncho, line 91: I dont like this to be executed on load time. Imagine every package to perform that on startup similar logic - it will kill lsp-mode's load time.

sebastiaanspeck: What would you suggest?

yyoncho: Use lambda and then lsp-resolve-value

sebastiaanspeck: Like this?:
```
(defcustom lsp-magik-java-path (cond ((eq system-type 'windows-nt)
                                      (lambda ()
                                        (or (lsp-resolve-value (executable-find (expand-file-name "bin/java" (getenv "JAVA_HOME"))))
                                            (lsp-resolve-value (executable-find "java")))))
                                     (t "java"))
  "Path of the java executable."
  :type 'string
  :group `lsp-magik
```

yyoncho: to me, it looks like it will be better to wrap the whole thing in a lambda

sebastiaanspeck: Please review the latest update

yyoncho (approved):

yyoncho:
Thank you!

PR #4154, opened 2023-08-29
Font lock JSON messages in the lsp-log buffer

dakra:
NOTE: I haven't tested this with servers that send very big
responses and if it has a big impact on performance.
For my use-case and the lsp-servers I use, it works great.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -3142,6 +3142,22 @@ TYPE can either be `incoming' or `outgoing'"
    :type type
    :body body))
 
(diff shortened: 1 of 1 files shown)


yyoncho (approved):

yyoncho:

asandroq:
What would happen if the user doesn't have any JSON modes installed?

dakra:
If the user has no mode for .json files my guess is it will just activate fundamental-mode.
I explicitly use `set-auto-mode` because I don't want an extra dependency on any specific json-mode.

PR #4164, opened 2023-09-09
Fix so that lsp-semgrep-languages will take effect immediately

gvol:
Updating `lsp-semgrep-languages` didn't take effect until a restart (or recompile of lsp-semgrep.el).  This is because it's `:activation-fn` is `(apply 'lsp-activate-on lsp-semgrep-languages)` which effectively embeds the versions rather than checking the value of the variable each time.

### DIFF
--- a/clients/lsp-semgrep.el
+++ b/clients/lsp-semgrep.el
@@ -204,7 +204,8 @@ If FULL is non-nil, scan all files in workspace, regardless of git status."
 (lsp-register-client
  (make-lsp-client
   :new-connection (lsp-stdio-connection (lambda () lsp-semgrep-server-command))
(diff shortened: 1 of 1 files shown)


gvol:
I'm not sure what would be wrong with the windows-latest, or why the 3 other checks haven't finished.  Is it something I should worry about, or a problem with the CI right now?

jcs090218:
> I'm not sure what would be wrong with the windows-latest, or why the 3 other checks haven't finished. Is it something I should worry about, or a problem with the CI right now?

There are errors from that snapshot, so no need to worry about it! The other three checks should be removed since we've already dropped support for Emacs 26.3.

yyoncho (approved):

yyoncho:

PR #4172, opened 2023-09-19
lsp-terraform: fix invalid keys

daedric:
Hi,
I'm a doomemacs user and I got errors in the terraform client. I'm not sure when/where those changes occurred.
Let me know whether this needs rework,
Cheers,

### DIFF
--- a/clients/lsp-terraform.el
+++ b/clients/lsp-terraform.el
@@ -302,7 +302,7 @@ This is a synchronous action."
                    :installed-version installed-version
                    :version-constraint (lsp-get provider :version_constraint)))
 
-(lsp-defun construct-tf-module ((&terraform-ls:Module :name :docs-link :version :source-type :dependent-modules))
+(lsp-defun construct-tf-module ((&terraform-ls:Module :name :docs_link :version :source_type :dependent_modules))
   "Construct `TF-MODULE' using MODULE."
   (make-tf-module :name name
                   :doc-link docs-link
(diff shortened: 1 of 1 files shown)


yyoncho:
CI does not like this change. Note that when lsp-interface member is using _ it is converted in - when destructoring.

cc @psibi

psibi:
@daedric Can you let me know the exact error message you are seeing ? Also, I remember someone previously facing same issue which got resolved by cleaning up the old `*.elc` and `*.eln` files.

daedric:
it started to work again without :/
Sorry for the noise

psibi:
@daedric No worries, thanks for the confirmation! I remember same thing happening previously to someone too: https://github.com/emacs-lsp/lsp-mode/pull/3757

jmcarcell:
I also get the same error when using lsp from master but not with 8.0.0

daedric:
Actually it is strange, for me:
- if I open a Go file with master, I'll get the error every time and need the patch;
- if I open a C++ file with master, it will work and the LSP starts.

I worked it around by opening a Go file before a C++ because it makes it work. I did not have the time to investigate properly yet.

yyoncho:
clean installation should fix the issue.

jmcarcell:
@yyoncho it doesn't do it for me fater deleting everything `lsp*` in my packages folder and also the built folders. I actually have no idea because I have two computers both with the same version of emacs from Arch Linux that share a configuration and for one of them this issue happens consistently and for the other it doesn't

PR #4178, opened 2023-09-23
Bump elixir-ls version

wkirschbaum:
Bump the the latest elixir-ls version and also fix the download url, as this changed for the last two versions. 

This allows Elixir developers to use lsp-mode for the latest Erlang version.


yyoncho (approved):

yyoncho:
Thank you!

PR #4179, opened 2023-09-25
lsp-php: Support `intelephense.environment.includePaths`

drzraf:
Don't merge it: I can't get it to work and but can't understand why.

While a symlink does load external framework's code, I can't get this directive to have the same effects. Whether it's being forwarded incorrectly or interpreted incorrectly by vscode-intelephense, I can't understand how to debug it.

I also wished https://phpactor.readthedocs.io/en/master/reference/configuration.html#indexer-index-path directive would be implemented too. The ability to load code from custom directories is a must-have for all backends.

### DIFF
--- a/clients/lsp-php.el
+++ b/clients/lsp-php.el
@@ -122,7 +122,7 @@ from all language server features."
 (lsp-defcustom lsp-intelephense-paths-include
   []
   "Configure additional paths outside workspace."
-  :type '(repeat string)
+  :type 'lsp-string-vector
   :group 'lsp-intelephense
   :package-version '(lsp-mode . "8.1")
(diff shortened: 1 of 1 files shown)


clients/lsp-php.el
```diff
@@ -119,6 +119,14 @@ from all language server features."
   :package-version '(lsp-mode . "6.1")
   :lsp-path "intelephense.files.exclude")
 
+(lsp-defcustom lsp-intelephense-paths-include
+  []
+  "Configure additional paths outside workspace."
+  :type '(repeat string)
```
yyoncho, line 125: it should be lsp-string-vector

yyoncho (approved):

yyoncho:
Thank you!

PR #4182, opened 2023-09-29
Make it possible to override lsp-cmake server command

### LINKED ISSUE #4180: cmakels server command option
It doesn't appear as if lsp-cmake allows setting the server command through a variable. Most other modules support something like `lsp-{name}-server-command`. It would be great to also have an `lsp-cmake-server-command` variable that could be set (defaulting to "cmake-language-server").

themkat:
Like the title says. 

Fixes #4180


rycee:
Thanks @themkat! I tried this patch out and it worked just fine, much appreciated 🙂

yyoncho (approved):

Issue #4158, opened 2023-09-02
adding remote lsp client for ess-r-mode does not work


beyondpie:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Following the instructions in lsp tramp mode, I set the following configuration for ess-r-mode
```
(defun remoteRlsp ()
  (interactive)
  (with-eval-after-load 'lsp-mode
    (lsp-register-client
     (make-lsp-client
      :new-connection (lsp-tramp-connection "/home/szu/mambaforge/envs/seurat/bin/R --slave -e languageserver::run()")
      :major-modes '(ess-r-mode)
      :remote? t
      :server-id 'lsp-r)))
  )
```
Then in a remote script, if I run `M-x lsp`
Lsp will report (in the *lsp-log*)

> Command "/home/szu/mambaforge/envs/seurat/bin/R --slave -e languageserver::run()" is not present on the path.

But using Eglot
I add the following codes
```
(defun remoteREglot()
  (interactive)
  (with-eval-after-load 'eglot
    (add-to-list 'eglot-server-programs
                 '( (R-mode ess-r-mode)
                    "/home/szu/mambaforge/envs/seurat/bin/R" "--slave" "-e" "languageserver::run()")))
  
  )
```
Then in a remote R script, I can run `M-x eglot`, which works as expected. 

### Steps to reproduce

1. def function below 
 (defun remoteRlsp ()
  (interactive)
  (with-eval-after-load 'lsp-mode
    (lsp-register-client
     (make-lsp-client

      ;; /home/szu/mambaforge/envs/seurat/bin/R should be replace to remote R 
      ;; the R should has language server package, which can be installed in remote R
      ;; install.packages("language server")

      :new-connection (lsp-tramp-connection "/home/szu/mambaforge/envs/seurat/bin/R --slave -e languageserver::run()")
      :major-modes '(ess-r-mode)
      :remote? t
      :server-id 'lsp-r)))
  )

2. load the above funtion
3. in a remote R script, run M-x lsp.

### Expected behavior

I can have lsp-mode in remote R script.

### Which Language Server did you use?

lsp-r

[...]

Issue #4150, opened 2023-08-23
Remote clangd JSON parse error


korteland:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

When using LSP mode with TRAMP, the remote clangd's status stays as "starting" indefinitely. On the other machine, a file `tmp/clangd-remote-1-stderr` appears with the following contents:

```
I[22:51:29.609] Ubuntu clangd version 13.0.1-2ubuntu2.2
I[22:51:29.609] Features: linux+grpc
I[22:51:29.609] PID: 2449947
I[22:51:29.609] Working directory: /home/ubuntu/dev/wt-11340/wiredtiger
I[22:51:29.609] argv[0]: clangd-13
I[22:51:29.609] Starting LSP over stdin/stdout
E[22:51:29.656] JSON parse error: [3:3085, byte=3087]: Expected , or } after object property
```

Setting `lsp-log-io` to `t` per the troubleshooting steps, the content of my `*lsp-log: clangd-remote:XXXX*` buffer contains the following:

```
[Trace - 08:53:52 AM] Sending request 'initialize - (1)'.
Params: {
  "processId": null,
  "rootPath": "/home/REDACTED",
  "clientInfo": {
    "name": "emacs",
    "version": "GNU Emacs 29.1 (build 1, aarch64-apple-darwin22.6.0, Carbon Version 169 AppKit 2299.7)\n of 2023-08-06"
  },
  "rootUri": "file:///home/REDACTED",
  "capabilities": {
    "general": {
      "positionEncodings": [
        "utf-32",
        "utf-16"
      ]
    },
    "workspace": {
      "workspaceEdit": {
        "documentChanges": true,
        "resourceOperations": [
          "create",
          "rename",
          "delete"
        ]
      },
      "applyEdit": true,
      "symbol": {
        "symbolKind": {
          "valueSet": [
            1,
            2,
            3,
            4,
            5,
            6,
            7,
            8,
            9,
            10,
            11,
            12,
            13,
            14,
            15,
            16,
            17,
            18,
            19,
            20,
            21,
            22,
            23,
            24,
            25,
            26
[...]

Issue #3841, opened 2022-12-06
update lsp-mode over TRAMP docs


saltsucker:
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? No.

Can you please update the LSP-mode over tramp tutorial? It is sorely unfulfilling for anyone that wants to try to make this work. There are so many broken semi-solutions all over the internet of people basically shooting in the dark and crossing their fingers. I would make a PR, but... I can't make it work no matter what I try, so I can't contribute.

Some examples of more helpful info:
What OS's are known to work?
What remote OS's are known to work?
What is the *exact* environment required locally and remotely? Beside just "install LSP on local and remote"?
Does it work from Windows Subsystem for Linux?
What if you're using pyenv? (or some other python virtual environment tool)

Basically having an actually fulfilling fleshed out example would just be way more helpful.

https://emacs-lsp.github.io/lsp-mode/page/remote/

yyoncho:
tramp support will be rewritten and will work OOTB with little/no configuration. Back then when I implemented tramp support it had tons of bizzare issues(e. g. for some os configurations it converts \n to \r).

yyoncho:
@saltsucker it will be soon. But IMHO the overall experience won't be good anyway if you want to use it for serious work it may not perform well(due to blocking IO). FWIW if you want to connect to docker container there are alternatives.

saltsucker:
> @saltsucker it will be soon. But IMHO the overall experience won't be good anyway if you want to use it for serious work it may not perform well(due to blocking IO). FWIW if you want to connect to docker container there are alternatives.

Ya... I had a feeling it was going to be a laggy experience (if that's what you mean). But VSCode can do it well (iirc, haven't used it in a while). Does that have non-blocking IO? 

I am not familiar with the docker way, so I am not sure what you mean. I want TRAMP so I can remote into my work servers. Navigate them, work on projects, etc. How does docker fit into that workflow? Spinning up a docker container on work server and trying to TRAMP into that from my PC doesn't seem better.

What other alternatives can I try? I think I tried sshfs and I don't think that worked well either.

cpoile:
@yyoncho Was there any success in getting this to work? (Appreciate your help!)

yyoncho:
@cpoile I am working on this right now.

Issue #3642, opened 2022-07-29
elixir-ls wont start over tramp


mogorman:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

When trying to use lsp mode with elxir locally it works without issue. trying to use it over tramp mode causes it to crash instantly on startup due to a json decoding bug, it looks like there is an off by one issue.

### Steps to reproduce

install elixir-ls
```
 (make-lsp-client :new-connection (lsp-tramp-connection "/path/to/elixir-ls")
                  :major-modes '(elixir-mode)
                  :remote? t
                  :server-id 'elixir-ls-remote))
```
open an elixir file over tramp

### Expected behavior

that I could use lsp-mode over tramp with elixir files.

### Which Language Server did you use?

elixir-ls

### OS

Linux

### Error callstack

```emacs-lisp
[Trace - 08:27:35 AM] Sending request 'initialize - (6)'.
Params: {
  "processId": null,
  "rootPath": "/home/mog/abcde",
  "clientInfo": {
    "name": "emacs",
    "version": "GNU Emacs 29.0.50 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.30, cairo version 1.16.0)"
  },
  "rootUri": "file:///home/mog/abcde",
  "capabilities": {
    "workspace": {
      "workspaceEdit": {
        "documentChanges": true,
        "resourceOperations": [
          "create",
          "rename",
          "delete"
        ]
      },
      "applyEdit": true,
      "symbol": {
        "symbolKind": {
          "valueSet": [
            1,
            2,
            3,
            4,
            5,
            6,
            7,
            8,
            9,
            10,
            11,
            12,
            13,
            14,
            15,
            16,
            17,
            18,
            19,
            20,
            21,
            22,
            23,
            24,
            25,
            26
          ]
        }
      },
      "executeCommand": {
        "dynamicRegistration": false
      },
      "didChangeWatchedFiles": {
        "dynamicRegistration": true
      },
      "workspaceFolders": true,
      "configuration": true,
      "codeLens": {
        "refreshSupport": true
      },
      "fileOperations": {
[...]

PR #4204, opened 2023-10-27
Fix lsp-mode's tramp support

yyoncho:
Fix lsp-mode's tramp support

- This fixes the implementation of `lsp-mode` tramp support. After this PR the
remote clients will be automatically registered and in most of the cases it will
work out of the box. The remote connection is managed to a way similar to what
eglot does.

Fixes https://github.com/emacs-lsp/lsp-mode/issues/4158
Fixes https://github.com/emacs-lsp/lsp-mode/issues/4150
Fixes https://github.com/emacs-lsp/lsp-mode/issues/4158
Fixes https://github.com/emacs-lsp/lsp-mode/issues/4150
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3841
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3642
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3579
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3530
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3491
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3490
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3391
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3369
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3364
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3020
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3018
Fixes https://github.com/emacs-lsp/lsp-mode/issues/3020

### DIFF
--- a/docs/page/remote.md
+++ b/docs/page/remote.md
@@ -5,33 +5,14 @@ root_file: docs/page/remote.md
 
 ## TRAMP
 
-LSP mode has support for tramp buffers with the following requirements:
+`lsp-mode` has support for tramp buffers with the following requirements:
 
 - The language server has to be present on the remote server.
 - Having multi folder language server (like [Eclipse JDT LS](https://github.com/eclipse/eclipse.jdt.ls)) cannot have local and remote workspace folders.
 
 ### How does it work?
 
-`lsp-mode` detects whether a particular file is located on remote machine and looks for a client which matches current file and it is marked as `:remote?` t. Then `lsp-mode` starts the client through tramp.
-
-### Sample configuration
-
-Here it is example how you can configure python language server to work when using `TRAMP`. Note that if you are trying to convert existing language server configuration you should copy all of it's properties(e. g. `:request-handlers`, `activation-fn`, etc). Also, when you are doing that you should make sure that none of the custom language server settings are not pointing to local path because those settings will be sent to the remote server.
-
-```elisp
-(lsp-register-client
-    (make-lsp-client :new-connection (lsp-tramp-connection "<binary name (e. g. pyls, rls)>")
-                     :major-modes '(python-mode)
-                     :remote? t
-                     :server-id 'pyls-remote))
-```
-
-_Note:_ when you do not have root privileges on the remote machine to put the language server on the path you may alter the remote path by changing `tramp-remote-path`.
-
-### Dealing with stderr
-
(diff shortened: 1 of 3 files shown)


docs/page/remote.md
```diff
@@ -5,33 +5,14 @@ root_file: docs/page/remote.md
...
-```
-
-_Note:_ when you do not have root privileges on the remote machine to put the language server on the path you may alter the remote path by changing `tramp-remote-path`.
-
-### Dealing with stderr
-
-With TRAMP, Emacs does not have an easy way to distinguish stdout and stderr, so when the underlying LSP process writes to stderr, it breaks the `lsp-mode` parser. As a workaround, `lsp-mode` is redirecting stderr to `/tmp/<process-name>-<id>~stderr`.
-
+`lsp-mode` detects whether a particular file is located on remote machine and looks for a client which matches current file and it is marked as `:remote?` t. Then `lsp-mode` starts the client through tramp. By default `lsp-mode` will copy the local client and mark it as `remote? t`. In most of the cases it is good enough but certain cases this may not work (e. g. if the server configuration contains references to local paths). In this case the user is supposed to create `.dir-local` configuration to override the references to local paths or open an issue on `lsp-mode` side to make the setting remote agnostic. To turn of automatic remote clients registration you can set `lsp-auto-register-remote-clients` to `nil`.
```
ericdallo, line 15: 💯

ericdallo (approved):

jcs090218 (approved):

docs/page/remote.md
sochotnicky: @yyoncho A small typo here "To turn off" (instead of "To turn of").

But as a side note I've applied this patch and it's been working nicely for my use cases.

yyoncho: @sochotnicky thank you, will be applied before merging.

sochotnicky:
While testing this out, I realized that if the local side has the LSP server installed but remote does not, this will fail with a bit of cryptic message (lsp server failed to start do you want to restart etc). Inspecting the logs it looks like the attempt was to start a server that didn't exist remotely.

Not sure if you'd want to tackle that one here or perhaps a separate PR.

yyoncho:
@sochotnicky pushed a fix for this issue. Please let me know if it works for you. Thank you for testing it out.

PR #4205, opened 2023-10-27
Make sure that only one error is sent for code actions

### LINKED ISSUE #3442: multiple errors on same line confuse code actions

### Bug description

I am using lsp-grammarly. With a sentence like:

Here is a santence with moltiple mistakees.

if I click on the second or third mistakes, the code actions offered (either via sideline or code-actions right-mouse-click menu item) always belong to the first error.  If I write the sentence instead as:

Here is a santence 
with moltiple 
mistakees.

the problem disappears.  Happy to provide additional information.

### Steps to reproduce

requires lsp-grammarly

### Expected behavior

clicking on a given error should show code actions for that error, not others

text file

### OS

MacOS

### Error callstack

```emacs-lisp
no error messages
```

### Anything else?

that's it

yyoncho:
Fixes #3442

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5065,6 +5065,25 @@ identifier and the position respectively."
                 (lsp--get-buffer-diagnostics))
                'vector)))
 
+(lsp-defun lsp-range-overlapping?((left &as &Range :start left-start :end left-end)
+                                  (right &as &Range :start right-start :end right-end))
+  (or (lsp-point-in-range? right-start left)
+      (lsp-point-in-range? right-end left)
+      (lsp-point-in-range? left-start right)
+      (lsp-point-in-range? left-end right)))
+
+(defun lsp-cur-possition-diagnostics ()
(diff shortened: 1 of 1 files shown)


ericdallo:
@yyoncho I think passing all line diagnostics is a good thing some times, at least for some servers like clojure-lsp, where to apply a code action you don't need to actually have your cursor exactly on the diagnostic but on the same line

ericdallo:
Curious if other clients are doing that

yyoncho:
I was wondering whether to make it configurable here. I havent checked it but I think that vscode is sending only the stuff under the cursor.

ericdallo:
Yeah, maybe I'm skewed since I use a lot this "feature" of applying code actions without the need of having the cursor under exactly the diagnostics, but only on the same line, maybe more people feel the same, maybe we can go with that and check if anyone complains?

yyoncho:
Ok, lets go this way and if it someone complains we will make it configurable.

PR #4089, opened 2023-07-04
Add AWK language server support

Hi-Angel:
Implements the following [wishlist item](https://github.com/emacs-lsp/lsp-mode/discussions/2892#discussioncomment-6333173).

PR #4216, opened 2023-11-03
Initial support for wgsl-analyzer

### LINKED ISSUE #3983: WGSL Analyzer support
Add support for [WGSL](https://gpuweb.github.io/gpuweb/wgsl/) using [wgsl-analyzer](/wgsl-analyzer/wgsl-analyzer).

themkat:
Fixes #3983

wgsl-analyzer does a lot of non-standard things, including how it handles the configuration requests. Others have also pointed out this in the issue tracker: https://github.com/wgsl-analyzer/wgsl-analyzer/issues/77 . Had to fight a bit with the language server to get it to read my configuration. Tried the standard way without success in several different ways. (standard way meaning using `lsp-register-custom-settings`). Have tried commenting the reasoning for the next person.

Implements the basic functionality to make the server usable, as well as configuration properties for it. Also implements the custom request to get the full source code (which includes preprocessor step being executed, as well as symbols fetched from dependencies etc.). Other custom requests like source tree I did not understand the response format of, so leaving that for a further improvement later.

Could not get the `lsp-wgsl-shaderdefs` settings variable to be serialized as a sequence in any way... Lists and vectors with contents give the famous `wrong argument consp, nil` error. Only thing working so far is an empty vector. Would very much like feedback on how this could be solved.

Feel free to suggest any improvements! 🙂 Felt a bit dirty with all the custom configuration handling 😛

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 8.0.1
+  * Add support for wgsl using [[https://github.com/wgsl-analyzer/wgsl-analyzer][wgsl-analyzer]]
   * Add [[https://github.com/wader/jq-lsp][jq-lsp]]
   * Add architecture triples for [[https://github.com/WhatsApp/erlang-language-platform][erlang-language-platform]] file downloads, to support macos on ARM and X86.
   * Add semantic token support for [[https://github.com/WhatsApp/erlang-language-platform][erlang-language-platform]] in lsp-erlang client.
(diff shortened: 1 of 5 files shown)


jcs090218:
You can use #4089 as a reference.

themkat:
> You can use #4089 as a reference.

No? I don't see any config handling or anything else there which is what I'm writing about in my description above. As far as I can see, the AWK PR works only as a template for creating new clients and documenting them (i.e, changelog etc.). Seems to me like you didn't read any of my text, but saw that an entry in `lsp-clients.json` was missing or something. That can be added once the issues I mention above is handled. I have made contributions where I make new clients before, so I know what is needed before merging.


Added the missing docs entry now, so no need to do it later.

Issue #4222, opened 2023-11-06
lsp broken by macro expansion failure and docs-link vs docs_link discrepancy


dod38fr:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Hi

On startup, emacs shows the following error message:
```
Beginning of buffer [3 times]
File mode specification error: (error Eager macro-expansion failure: (error "Unknown key: :docs-link.  Available keys: (:name :docs_link :version :source_type :dependent_modules)")) [2 times]
Mark set
```

This breaks lsp-mode for terraform-mode and go-mode. The latter is weird because lsp-terraform is the only mode dealing with `docs_link` (with underscore).

These modules are loaded with:
```
(use-package lsp-mode
  :ensure t
   ;; set prefix for lsp-command-keymap (few alternatives - "C-l", "C-c l")
  :init (setq lsp-keymap-prefix "C-c l")
  :hook (
         (terraform-mode . lsp)
         (go-mode . lsp)
         (python-mode . lsp) )
  :commands lsp)
```

### Steps to reproduce

This error happens when loading `lsp-mode`

### Expected behavior

No error when loading `lsp-mode`

### Which Language Server did you use?

- terraform-ls 0.30.1
- gopls

### OS

Linux

### Error callstack

```emacs-lisp
call stack:

Debugger entered--Lisp error: (error "Eager macro-expansion failure: (error \"Unknown key...")
  error("Eager macro-expansion failure: %S" (error "Unknown key: :docs-link.  Available keys: (:name :..."))
  internal-macroexpand-for-load((defalias 'construct-tf-module #'(lambda (input0) "Construct `TF-MODULE' using MODULE." (-let* (((&terraform-ls:Module :name :docs-link :version :source-type :dependent-modules) input0)) (make-tf-module :name name :doc-link docs-link :version version :source-type source-type :dependent-modules dependent-modules)))) t)
  load-with-code-conversion("/home/domi/.emacs.d/elpa/lsp-mode-20231103.1522/ls..." "/home/domi/.emacs.d/elpa/lsp-mode-20231103.1522/ls..." t t)
[...]

PR #4223, opened 2023-11-06
lsp-terraform: fix typo in make-tf-module call

dod38fr:
Before this change, I could not use`lsp-mode` in a terraform buffer. For more details in the issue, see #4222 

Closes: #4222


dod38fr:
My bad. The original issue comes from loading lsp-mode the wrong way. Closing.

PR #4228, opened 2023-11-11
Fix credo language server command issues

wkirschbaum:
The download did not work and command was incomplete. This fixes both. The minimum version also gets bumped to 3.0 as the new stable version.


yyoncho (approved):

wkirschbaum:
@yyoncho i made another commit, because we don't need the binary path if the store path is the same.

jcs090218 (approved):

jcs090218:
Thank you!

PR #4234, opened 2023-11-14
Make shaderdefs configurable for lsp-wgsl

themkat:
In my previous PR, #4216, I had some issues getting the shaderdefs (shader definitions) input to be configurable. I got weird error messages. Seems like it was only a brainfart from my end, and that I needed to make the type `(repeat string)`. Now it finally works 😄 Hopefully it will make the WGSL support slightly more usable.

### DIFF
--- a/clients/lsp-wgsl.el
+++ b/clients/lsp-wgsl.el
@@ -101,11 +101,11 @@
   :group 'lsp-wgsl
   :package-version '(lsp-mode . "8.0.1"))
 
-;; (defcustom lsp-wgsl-shaderdefs []
-;;   "Defines that should be valid for preprocessor operations like ifdef."
-;;   :type 'vector
-;;   :group 'lsp-wgsl
-;;   :package-version '(lsp-mode . "8.0.1"))
+(defcustom lsp-wgsl-shaderdefs []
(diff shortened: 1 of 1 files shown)


clients/lsp-wgsl.el
```diff
@@ -101,11 +101,11 @@
...
   :package-version '(lsp-mode . "8.0.1"))
 
-;; (defcustom lsp-wgsl-shaderdefs []
-;;   "Defines that should be valid for preprocessor operations like ifdef."
-;;   :type 'vector
-;;   :group 'lsp-wgsl
-;;   :package-version '(lsp-mode . "8.0.1"))
+(defcustom lsp-wgsl-shaderdefs []
+  "Defines that should be valid for preprocessor operations like ifdef, e.g, ['USE_TYPES', 'DEBUG']"
+  :type '(repeat string)
```
yyoncho, line 106: it should be lsp-string-vector

themkat: Pushed now 🙂 There are many other clients (e.g, xml, erlang) that uses `(repeat string)`, so I guess those should be changed in the future as well... (aka technical debt)?

yyoncho (approved):

yyoncho:

PR #4238, opened 2023-11-18
Add golangci-lint add-on client using golangci-lint-langserver

jimeh:
Operates as a add-on server alongside gopls when working with Go files. Only activates if golangci-lint-server executable is available: https://github.com/nametake/golangci-lint-langserver

---

I have been using this client as for about 6 months in my personal config. Before then I used a much shorter and hackier snippet that achieved the same thing for somewhere around a year.

Here's the client in the vendor directory of my Emacs config:
https://github.com/jimeh/.emacs.d/blob/v0.8.46/vendor/lsp-golangci-lint/lsp-golangci-lint.el

I'm more than happy to address any feedback :)


yyoncho:
Looks good. Can you update `/docs/lsp-clients.json` accordingly and also add it to  `lsp-client-packages` variable.

yyoncho (approved):

PR #4239, opened 2023-11-18
Add buf/protobuf support via bufls

jimeh:
While [`bufls`](https://github.com/bufbuild/buf-language-server) is pretty basic, it's still pretty nice to have the basic features it currently provides available when working on protobuf files. And hopefully it will only get better over time.

I have personally been using this client for around 2 months without issue, and generally find it quite useful.

Here's a link to it in the vendor directory of my Emacs config:
https://github.com/jimeh/.emacs.d/blob/v0.8.46/vendor/lsp-bufls/lsp-bufls.el

I'm more than happy to address any feedback :)

### DIFF
--- a/clients/lsp-bufls.el
+++ b/clients/lsp-bufls.el
@@ -1,0 +1,69 @@
+;;; lsp-bufls.el --- bufls-langserver Client settings -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2023  Jim Myhrberg
+
+;; Author: Jim Myhrberg
+;; Keywords: lsp, protobuf, buf, bufls
+
+;; This file is not part of GNU Emacs
+
+;;; License:
+;;
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+;;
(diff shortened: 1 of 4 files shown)


clients/lsp-bufls.el
```diff
@@ -0,0 +1,71 @@
...
+
+(defun lsp-bufls-server--stdio-command ()
+  "Return the command and args to start bufls-langserver."
+  (let ((args (list lsp-bufls-path "serve")))
+    (when (and (listp lsp-bufls-args)
+               (length> lsp-bufls-args 0))
+      (setq args (append args lsp-bufls-args)))
+    args))
+
+(add-to-list 'lsp-language-id-configuration '(protobuf-mode . "protobuf"))
```
yyoncho, line 58: no need to do it here, move it inline.

yyoncho (commented):
Update the same places I mentioned in: https://github.com/emacs-lsp/lsp-mode/pull/4238#issuecomment-1817512029

jimeh:
I assume this is something I need to address in the 27.2 builds?

>     lsp-bufls.el:70:1:Warning: the function ‘length>’ is not known to be defined.

jimeh:
#4240 also fixes the `length>` issue in the golangci-lint client which I recently added.

yyoncho (approved):

PR #4242, opened 2023-11-20
Add interactive function for viewing WGSL syntax tree

themkat:
When implementing the WGSL support the first time, I was not familiar with the syntax tree functionality and what it could be used for. Seems like other language servers like rust-analyzer has the same functionality and presents it in plain text to the user. Doing the same for wgsl. 

This implements support for the custom `wgsl-analyzer/syntaxTree` operation that wgsl-analyzer supports.


yyoncho (approved):

yyoncho:

PR #4243, opened 2023-11-20
Add option to install language servers using Cargo

themkat:
More and more software is developed with Rust, and it is not irregular these days that some language servers are distributed with Cargo. Adding the option to install automatically using Cargo makes a lot of sense in my view. Added a quick example for WGSL where we also install directly from a git repo. Some of the code is obviously inspired by the NPM package installation.

Also added a filter on `nil` values in  `lsp-async-start-process`. In my view, this makes the function more robust, and we can do handling like in the cargo logic (e.g, only sending git options when they are given).

Probably not perfect, so looking forward to hearing about possible improvements 🙂 Feel free to discard if this is not something that is wanted for lsp-mode. I understand if this is out of scope, though I find it really handy to have 🙂

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 8.0.1
+  * Add support for installing lsp dependencies via Cargo.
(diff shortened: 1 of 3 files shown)


yyoncho (approved):

yyoncho:

PR #4245, opened 2023-11-26
Add missing major-modes declaration for lsp-wgsl

themkat:
Seems like the major-mode declaration is missing for wgsl. First time installing lsp-wgsl from melpa instead of local evaluation (eval buffer etc.), so let's blame that for me not noticing sooner 😛 

Got the following error when trying it directly from lsp-mode in melpa:
```text
LSP :: There are no language servers supporting current mode `wgsl-mode' registered with `lsp-mode'.
2. The language server that you expect to run is not configured to run for major mode `wgsl-mode'. You may check that by checking the `:major-modes' that are passed to `lsp-register-client'.
3. `lsp-mode' doesn't have any integration for the language behind `wgsl-mode'. Refer to https://emacs-lsp.github.io/lsp-mode/page/languages and https://langserver.org/ .
4. You are over `tramp'. In this case follow https://emacs-lsp.github.io/lsp-mode/page/remote/.
5. You have disabled the `lsp-mode' clients for that file. (Check `lsp-enabled-clients' and `lsp-disabled-clients').
You can customize `lsp-warn-no-matched-clients' to disable this message.
```

### DIFF
--- a/clients/lsp-wgsl.el
+++ b/clients/lsp-wgsl.el
@@ -185,6 +185,7 @@
                                       ;; https://github.com/wgsl-analyzer/wgsl-analyzer/issues/77
                                       (lsp--set-configuration '())))
                   :request-handlers (lsp-ht ("wgsl-analyzer/requestConfiguration" #'lsp-wgsl--send-configuration))
+                  :major-modes '(wgsl-mode)
                   :activation-fn (lsp-activate-on "wgsl")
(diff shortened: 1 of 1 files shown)


themkat:
Is there anything missing here for this to be merged? 🙂 If you don't want it, I can also close it. It just fixed the minor issues I was having with wgsl, so thought it might be useful for someone else as well.

yyoncho (approved):

PR #4253, opened 2023-12-09
Add html-ts-mode to `lsp-language-id-configuration` alist

pawelkobojek:
html-ls is currently not automatically associated with treesit's `html-ts-mode`. This tiny PR just adds it to the `lsp-langauge-id-configuration` alist.


jcs090218 (approved):

jcs090218 (approved):
Thank you!

PR #4259, opened 2023-12-17
Fix: wrong name for lua-ts-mode indent variable

nverno:
`lsp--formatting-indent-alist` has wrong name for `lua-ts-indent-offset`


jcs090218 (approved):

jcs090218 (approved):
Confirmed this. Thanks!

PR #4262, opened 2023-12-17
yield thread while waiting for native json rpc

blahgeek:
sit-for may not yield if there's input pending

it seems to solve https://github.com/emacs-lsp/emacs/issues/12


wyuenho:
If you needs request that doesn't block, shouldn't the solution be https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-mode.el#L3452-L3488 ? Which RPC call is troubling you to request this change?

PR #4266, opened 2023-12-19
Move common test utilities into a separate file.

manphiz:
* This avoids loading a ERT test utility directly and causing ERT test redefined error.

* Also drop "provide" from test units to avoid loading an ERT test unit directly.

* More details
  - https://debbugs.gnu.org/cgi/bugreport.cgi?bug=66782
  - https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1052939

* It should be encouraged to move more common utility to the lsp-test-utils.el file.  It's not causing immediate issues so not doing this in this PR.

### DIFF
--- a/test/lsp-test-utils.el
+++ b/test/lsp-test-utils.el
@@ -1,0 +1,42 @@
+;;; lsp-test-utils.el --- unit test utilities -*- lexical-binding: t -*-
+
+;; Copyright (C) emacs-lsp maintainers
+
+;; This program is free software: you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; This file contains definitions required by other test units.  Reusable
+;; utilities like functions, macros, and customizables should prefer to be
+;; defined here to avoid having to load a unit test directly, because Emacs
+;; checks for reused ERT test names since 29.1, which may also be caused by
+;; loading the same test unit multiple times.  See also discussion in
+;; https://debbugs.gnu.org/cgi/bugreport.cgi?bug=66782
+
+;;; Code:
+
+(defun lsp-test--wait-for (form &optional d)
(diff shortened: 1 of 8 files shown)


test/lsp-test-utils.el
```diff
@@ -0,0 +1,42 @@
+;;; lsp-test-utils.el --- unit test utilities -*- lexical-binding: t -*-
+
+;; Copyright (C)  Xiyue Deng <manphiz@gmail.com>
```
manphiz, line 3: Please let me know if copyright should be assigned to an existing maintainer.

jcs090218 (approved):
I've no objection to this.

manphiz:
Thanks @jcs090218 !  

A few more questions:
- The test results from 26.3 are not showing up and no details are available.  How can I check their statuses? (Sorry not very familiar with the CI workflow here)
- Regarding the copyright question, do you have any concerns?  (I'm not keen on being part of the copyright owners and can list you instead if you are fine with it)

jcs090218:
> The test results from 26.3 are not showing up and no details are available. How can I check their statuses? (Sorry not very familiar with the CI workflow here)

Those are the old requirements after we dropped support for Emacs 26.x. We can just ignore that!

> Regarding the copyright question, do you have any concerns? (I'm not keen on being part of the copyright owners and can list you instead if you are fine with it)

You can use `emacs maintainers` for the copyright. Example:

https://github.com/emacs-lsp/lsp-mode/blob/5b2f3741bb797371bba2932653ca829167f9745a/clients/lsp-ansible.el#L3C23-L3C44

manphiz:
> > The test results from 26.3 are not showing up and no details are available. How can I check their statuses? (Sorry not very familiar with the CI workflow here)
> 
> Those are the old requirements after we dropped support for Emacs 26.x. We can just ignore that!
> 

Ack.

> > Regarding the copyright question, do you have any concerns? (I'm not keen on being part of the copyright owners and can list you instead if you are fine with it)
> 
> You can use `emacs maintainers` for the copyright. Example:
> 
> https://github.com/emacs-lsp/lsp-mode/blob/5b2f3741bb797371bba2932653ca829167f9745a/clients/lsp-ansible.el#L3C23-L3C44

Got it.  Updated and force pushed.  Will monitor the new CI test results.

test/lsp-test-utils.el
manphiz: Changed to `emacs-lsp maintainers` as per suggestion by @jcs090218

manphiz:
Hmm, the latest CI run failed on macos-latest 28.2[1] which didn't seem to relate to this PR.  Is there a way to trigger a rerun to see if it's flaky?

[1] https://github.com/emacs-lsp/lsp-mode/actions/runs/7257421192/job/19771204594?pr=4266

jcs090218:
The CI on macOS has recently become flaky. I have rerun it!

yyoncho (approved):

PR #4303, opened 2024-01-19
Add support for Nushell

fredericgiquel:
Add support for Nushell language (https://www.nushell.sh/).

No custom settings for this server, so the code is small. The main thing to note is the activation of completion when using plists : same problem as with Openscad (https://github.com/emacs-lsp/lsp-mode/pull/4120), so same solution.


jcs090218 (approved):

jcs090218 (approved):
Thank you for your contribution! :)

Issue #147, opened 2017-10-27
`$/cancelRequest` support


vibhavp:
Now that `lsp-mode` supports asynchronous requests, clients should also be able to cancel requests they've sent asynchronously to the server. 

Additionally, quitting (<kbd>Ctrl-G</kbd>) should cleanly cancel the current synchronously sent request too.

PR #4315, opened 2024-01-27
ci: bump: GitHub Actions setup-cmake and setup-python

ncaq:
What
===
I Bump GitHub Actions setup-cmake and setup-python.

Why
===
This may be due to the use of old Actions, or because an unsupported version of Node.js is being used. Numerous warnings are being generated.
Security support has expired, which may cause problems, even though it is only used for CI. In any case, it is best to remove the warnings.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,7 +41,7 @@ jobs:
 
     - name: Setup cmake
       if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
-      uses: jwlawson/actions-setup-cmake@v1.4
+      uses: jwlawson/actions-setup-cmake@v1.14
       with:
         cmake-version: '3.18.x'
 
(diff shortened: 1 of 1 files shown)


ncaq:
I'll give up for now since setup-eask doesn't seem to be a bit easy to improve.

jcs090218:
This PR looks fine to me. We usually ignore the snapshot test results since it's not stable enough.

> I'll give up for now since setup-eask doesn't seem to be a bit easy to improve.

I'm the Eask’s author. You can ask me any questions about it. :)

ncaq:
I see [Update to node20 (#147) · emacs-eask/setup-eask@04e2fac](https://github.com/emacs-eask/setup-eask/commit/04e2fac20eb34f25a6bf9ad320c1161e8dc64e3e).
If this commit was made, it doesn't seem to be a problem.
I'm not sure if this is the only solution, though, since I have to turn off various other warnings.

PR #4316, opened 2024-01-27
ci: The version of actions/labeler is fixed at 4

ncaq:
Until now, the main had been specified, so it was automatically raised to the latest version, 5. And as a link, the structure of the configuration file has changed significantly in v5. [Release v5.0.0 · actions/labeler](https://github.com/actions/labeler/releases/tag/v5.0.0) I think it would be preferable to update the file accordingly, but since I am not familiar with the previous structure and I don't think it is healthy for CI to always fail, I will fix it at 4 for the time being. We will consider moving to v5 later.

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -6,7 +6,7 @@ jobs:
   triage:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/labeler@main
+    - uses: actions/labeler@v4
       with:
         repo-token: "${{ secrets.GITHUB_TOKEN }}"
         sync-labels: true


ncaq:
I see that GitHub Actions only change behavior after merging, so I have no idea if this fixes the problem correctly or not.
I fixed it to use version 4 for now, but the main version is actually working.
I'm not sure if it will work properly until I actually merge it.
It's too much trouble to test the behavior here since it's related to the label, so what should I do?
I'll leave the draft off for now.

jcs090218 (approved):
Make sense! Thank you!

jcs090218:
> I see that GitHub Actions only change behavior after merging, so I have no idea if this fixes the problem correctly or not.
I fixed it to use version 4 for now, but the main version is actually working.
> I'm not sure if it will work properly until I actually merge it.
It's too much trouble to test the behavior here since it's related to the label, so what should I do?
I'll leave the draft off for now.

Ah, sorry, I didn't see this message because I was using the phone to review.

You can test it on your forked repo and paste the test result here!

ncaq:
I tried it and the version designation was v4, not 4.
This has been fixed.
I'm not sure if forking doesn't copy the label or not, but it doesn't seem to be an error without question for now.
[chore: fake commit · ncaq/lsp-mode@c5d7399](https://github.com/ncaq/lsp-mode/actions/runs/7712100169/job/21018866877?pr=1)

jcs090218 (approved):

ncaq:
Is there any reason why this isn't getting merged?
I think the CI is failing because this PR fixes the CI in the first place, and the GitHub Actions are supposed to run the previous CI until it is merged.

PR #4322, opened 2024-02-04
Prefer buffer's major mode for rendering strings.

brownts:
Currently, when multiple major modes exist and are available for a language (e.g., c-mode and c-ts-mode), the major mode used to render strings is determined by the ordering of these modes within `lsp-language-id-configuration'.  This ordering doesn't necessarily reflect the preference of the user.

This change alters this behavior by preferring to render strings, whose language matches that of the current buffer, with the same major mode as the buffer.  When the string's language is different from the buffer, the behavior falls back to selecting the first major mode matching the language from `lsp-language-id-configuration'.


yyoncho (approved):

yyoncho:
Thank you!

Issue #4254, opened 2023-12-09
ts-ls: Wrong lsp-clients-typescript-server-path


eginhard:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

https://github.com/emacs-lsp/lsp-mode/pull/4202 made some changes to handle a deprecated argument in the language server. This included changing `lsp-clients-typescript-server-path()` to return:
https://github.com/emacs-lsp/lsp-mode/blob/01bd566c0508993202b08071ed7037c9fc4e3329/clients/lsp-javascript.el#L803
For me this evaluates to `~/.emacs.d/var/lsp/server/npm/typescript/bin/node_modules/typescript/lib`, while `tsserver.js` is actually in `~/.emacs.d/var/lsp/server/npm/typescript/lib/node_modules/typescript/lib`. So I guess the line above should be:
```elisp
(f-join (f-parent (f-parent (lsp-package-path 'typescript))) "lib" "node_modules" "typescript" "lib"))))
```

### Steps to reproduce

1. Open a `.js` file with `ts-ls`, the following error appears: `LSP :: Error from the Language Server: Request initialize failed with message: Could not find a valid TypeScript installation. Please ensure that the "typescript" dependency is installed in the workspace or that a valid `tsserver.path` is specified. Exiting. (Internal Error)`

### Expected behavior

`ts-ls` starts normally

### Which Language Server did you use?

lsp-javascript/ts-ls

### OS

Linux

hamaker:
I'm having this issue as well. Reverting the change from #4202 to line 803 as mentioned by @eginhard resolved it.

yyoncho:
@kiennq - seems like it is a regression?

ncaq:
I am having a rather hard time getting lsp to work in a TypeScript project in a subdirectory due to this problem.
Also, the original PR does not work in environments that do not use node_modules such as yarn PnP, and I expect that some people will have trouble with that.

idhowardgj94:
> I'm having this issue as well. Reverting the change from #4202 to line 803 resolved it.

same issue here, and resolved by follow @hamaker's suggestion.

[...]

PR #4333, opened 2024-02-14
fix: remove direct references to the node_modules directory

ncaq:
Stop reading `node_modules` directly by the file system. The problem with this direct loading is that projects without `node_modules` in the top-level directory will not work at all. For example, when I develop with aws cdk,
I cut off the `cdk` or `infra` directory and create a new `package.json` and put `node_modules` there too, but in that case, the code before this modification will not work at all. I solved this problem by simply inserting an abstraction layer.

Relation.
* [lsp-javascript: supply correct path to tsserver for ts-ls by kiennq · Pull Request #4202 · emacs-lsp/lsp-mode](https://github.com/emacs-lsp/lsp-mode/pull/4202)
* [ts-ls: Wrong lsp-clients-typescript-server-path · Issue #4254 · emacs-lsp/lsp-mode](https://github.com/emacs-lsp/lsp-mode/issues/4254)

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -790,17 +790,28 @@ name (e.g. `data' variable passed as `data' parameter)."
   (when-let ((workspace (lsp-find-workspace 'ts-ls (buffer-file-name))))
     (eq 'initialized (lsp--workspace-status workspace))))
 
-(defun lsp-clients-typescript-project-ts-server-path ()
-  (f-join (lsp-workspace-root) "node_modules" "typescript" "lib" "tsserver.js"))
+(defun lsp-clients-typescript-server-path-by-node-require ()
+  "Get the location of the typescript library.
+Use Node.js.
+The node_modules directory structure is suspect
+and should be trusted as little as possible.
+If you call require in Node.js,
+it should take into account the various hooks."
+  (let ((output
+         (string-trim-right
+          (shell-command-to-string
+           "node -e \"console.log(require.resolve('typescript'))\" 2>/dev/null"))))
+    (if (string-empty-p output)
+        nil
+      (f-parent output))))
 
 (defun lsp-clients-typescript-server-path ()
-  (cond
-   ((and
-     lsp-clients-typescript-prefer-use-project-ts-server
-     (f-exists? (lsp-clients-typescript-project-ts-server-path)))
-    (lsp-clients-typescript-project-ts-server-path))
-   (t
-    (f-join (f-parent (lsp-package-path 'typescript)) "node_modules" "typescript" "lib"))))
+  (if lsp-clients-typescript-prefer-use-project-ts-server
+      (let ((server-path (lsp-clients-typescript-server-path-by-node-require)))
+        (if (f-exists? server-path)
+            server-path
+          (lsp-package-path 'typescript)))
+    (lsp-package-path 'typescript)))
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection (lambda ()
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -801,31 +801,26 @@ For example, yarn PnP.
 
 Optional argument DIR specifies the working directory
 to run the command in."
-  (let* ((default-directory (or dir default-directory))
-         (output
-          (string-trim-right
-           (shell-command-to-string
-            "node -e \"console.log(require.resolve('typescript'))\""))))
-    (if (string-empty-p output)
-        nil
-      (f-parent output))))
-
-(defun lsp-clients-typescript-package-path-direct ()
-  "`lsp-package-path' apply `'typescript' is modified.
-because the lsp server may not recognize the tsserver executable file,
-e.g., on Windows."
-  (if (memq system-type '(cygwin windows-nt ms-dos))
-      (lsp-clients-typescript-require-resolve (f-parent (lsp-package-path 'typescript)))
-    (lsp-package-path 'typescript)))
+  (when-let*
+      ((default-directory (or dir default-directory))
+       (output
+        (string-trim-right
+         (shell-command-to-string
+          "node -e \"console.log(require.resolve('typescript'))\"")))
+       (not-empty (not (string-empty-p output))))
+    (f-parent output)))
 
 (defun lsp-clients-typescript-server-path ()
-  "Return the TS sever path base on settings."
-  (if lsp-clients-typescript-prefer-use-project-ts-server
-      (let ((server-path (lsp-clients-typescript-require-resolve)))
-        (if (f-exists? server-path)
-            server-path
-          (lsp-clients-typescript-package-path-direct)))
-    (lsp-clients-typescript-package-path-direct)))
+  "Return the TS server path based on settings."
+  (if-let* ((use-project-ts lsp-clients-typescript-prefer-use-project-ts-server)
+            (server-path (lsp-clients-typescript-require-resolve))
+            (server-path-exist (f-exists? server-path)))
+      server-path
+    (if (memq system-type '(cygwin windows-nt ms-dos))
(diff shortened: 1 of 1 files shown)


ncaq:
As of the first modified PR, it was not working on Windows, so I modified it and confirmed that it works on Linux (WSL2) and Windows (Emacs installed with winget).

clients/lsp-javascript.el
```diff
@@ -790,20 +790,42 @@ name (e.g. `data' variable passed as `data' parameter)."
...
-   (t
-    (if (memq system-type '(cygwin windows-nt ms-dos))
-        (f-join (f-parent (lsp-package-path 'typescript)) "node_modules" "typescript" "lib")
-      (f-join (f-parent (f-parent (lsp-package-path 'typescript))) "lib" "node_modules" "typescript" "lib")))))
+  (if lsp-clients-typescript-prefer-use-project-ts-server
+      (let ((server-path (lsp-clients-typescript-require-resolve)))
+        (if (f-exists? server-path)
+            server-path
+          (lsp-clients-typescript-package-path-direct)))
+    (lsp-clients-typescript-package-path-direct)))
```
jcs090218, line 828: I think we can simplify this with `if-let*` block. 🤔 Try not to have `lsp-clients-typescript-package-path-direct` function in different places.

clients/lsp-javascript.el
```diff
@@ -790,20 +790,42 @@ name (e.g. `data' variable passed as `data' parameter)."
...
+Optional argument DIR specifies the working directory
+to run the command in."
+  (let* ((default-directory (or dir default-directory))
+         (output
+          (string-trim-right
+           (shell-command-to-string
+            "node -e \"console.log(require.resolve('typescript'))\""))))
+    (if (string-empty-p output)
+        nil
+      (f-parent output))))
```
jcs090218, line 811: This can also be simplified with `if-let*` or `when-let*`.

clients/lsp-javascript.el
```diff
@@ -790,20 +790,42 @@ name (e.g. `data' variable passed as `data' parameter)."
...
+it should take into account the various hooks.
+For example, yarn PnP.
+
+Optional argument DIR specifies the working directory
+to run the command in."
+  (let* ((default-directory (or dir default-directory))
+         (output
+          (string-trim-right
+           (shell-command-to-string
+            "node -e \"console.log(require.resolve('typescript'))\""))))
```
jcs090218, line 808: Normally, we should wrap this with `shell-quote-argument`. See https://github.com/emacs-grammarly/keytar/blob/69eb60440c21879ce65c7205f070d59574320bb4/keytar.el#L59-L71.

Not sure why we don't have something like this under the global `lsp-` namespace since it's very useful. 🤔

clients/lsp-javascript.el
```diff
@@ -790,20 +790,42 @@ name (e.g. `data' variable passed as `data' parameter)."
...
+        nil
+      (f-parent output))))
+
+(defun lsp-clients-typescript-package-path-direct ()
+  "`lsp-package-path' apply `'typescript' is modified.
+because the lsp server may not recognize the tsserver executable file,
+e.g., on Windows."
+  (if (memq system-type '(cygwin windows-nt ms-dos))
+      (lsp-clients-typescript-require-resolve (f-parent (lsp-package-path 'typescript)))
+    (lsp-package-path 'typescript)))
```
jcs090218, line 819: I need some clarification here. We were previously using `(f-join (f-parent (f-parent (lsp-package-path 'typescript))) "lib" "node_modules" "typescript" "lib")`, and now just `(lsp-package-path 'typescript)`? Am I missing something? Can you elaborate it? 🤔

clients/lsp-javascript.el
ncaq: I was certainly concerned about going through the shell as well and would have preferred not to if possible.
However, I could not immediately figure out how to execute the command directly and receive it as a string without creating a separate buffer or something. In this case, the argument is a fixed string constant, and since it is not passed through an argument, etc., I thought it would not be a security issue, so I passed it as it is.

clients/lsp-javascript.el
ncaq: Until the following PR, `(lsp-package-path 'typescript)` was just called.
[lsp-javascript: supply correct path to tsserver for ts-ls by kiennq · Pull Request #4202 · emacs-lsp/lsp-mode](https://github.com/emacs-lsp/lsp-mode/pull/4202)
Maybe that was a problem in the environment of the person who made this modification, so a direct directory deep dive was done to avoid that.
I honestly don't know why different environments have different directory path locations.
My suggestion this time is to just use `(lsp-package-path 'typescript)` by default, and only use Node.js calls that return a wide range of real PATHs in special environments.
If it is fundamentally buggy, we should fix `lsp-package-path` in the first place.
By this logic, the Windows-only branching on the caller side this time is also subtle, but I was puzzled as to why the TypeScirpt server did not read the PATH with tsserver.cmd at the top level and did not know how to fix it, so I used this method as a stopgap measure. I used this method as a stopgap measure.

clients/lsp-javascript.el
ncaq: I fixed.

clients/lsp-javascript.el
ncaq: I fixed.

clients/lsp-javascript.el
jcs090218: Make sense to me.

clients/lsp-javascript.el
```diff
@@ -790,20 +790,37 @@ name (e.g. `data' variable passed as `data' parameter)."
...
+
+Optional argument DIR specifies the working directory
+to run the command in."
+  (when-let*
+      ((default-directory (or dir default-directory))
+       (output
+        (string-trim-right
+         (shell-command-to-string
+          "node -e \"console.log(require.resolve('typescript'))\"")))
+       (not-empty (not (string-empty-p output))))
```
jcs090218, line 810: Variable `not-empty` is not used. May be we can leave it as blank?

clients/lsp-javascript.el
jcs090218: Interesting thing is `(lsp-package-path 'typescript)` gives me `tsserver` and not the full path. 🤔

clients/lsp-javascript.el
ncaq: I am aware that it is not used.
I was told that it would be better to use a form that would not proceed unless something other than nil was bound to it.
If you want to take advantage of this, I think the format before branching with `(string-empty-p output)` is correct.
I used `(string-empty-p output)` for the name, but it gave me a warning.

ncaq:
I am late in responding because my grandmother passed away and I caught a cold. My apologies.

clients/lsp-javascript.el
ncaq: I think cause by this.

```elisp
(lsp-dependency 'typescript
                '(:system "tsserver")
                '(:npm :package "typescript"
                       :path "tsserver"))
```

jcs090218:
> I am late in responding because my grandmother passed away and I caught a cold. My apologies.

Oh, I'm sorry for your loss. No worries, there is no reason to rush here! 

> Stop reading node_modules directly by the file system. The problem with this direct loading is that projects without node_modules in the top-level directory will not work at all.

At first, I thought you were right, but it was working on my side. Can you confirm this once again? 🤔 (Although I couldn't get the completion working)

ncaq:
@jcs090218 
Am I missing something?

jcs090218:
> Am I missing something?

I'm sorry, I'm a bit busy. I will take a look when I have time! 😓

idhowardgj94:
solved my issue, I think it's ok to merge.

jcs090218 (approved):

jcs090218:
Thank you!

PR #4334, opened 2024-02-15
Replace obsolete defadvice with add-advice

Hi-Angel:
Fixes a number of warnings:

    Warning: ‘defadvice’ is an obsolete macro (as of 30.1); use ‘advice-add’ or ‘define-advice’

-----------------

Side-note: I used a `'foo` syntax rather than `#'foo` because as of writing the words [byte-compiler does not correctly process functions defined inside a `defun` and `progn` and maybe other similar constructions](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=69120).

As the distinction only matters for byte-compiler warnings, but the byte-compiler does not process that correctly, I chose of course the suppressing syntax.

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -844,33 +844,37 @@ refresh in currently active buffer."
 
 This is a debugging tool, and may incur significant performance penalties."
   (setq lsp-semantic-tokens--log '())
-  (defadvice lsp-semantic-tokens--fontify (around advice-tokens-fontify activate)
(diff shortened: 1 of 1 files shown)


Hi-Angel:
Side-note: I used a `'foo` syntax rather than `#'foo` because as of writing the words [byte-compiler does not correctly process functions defined inside a `defun` and `progn` and maybe other similar constructions](https://debbugs.gnu.org/cgi/bugreport.cgi?bug=69120).

Hi-Angel:
Ping

PR #4338, opened 2024-02-18
treewide: replace `looking-at` with `looking-at-p` if match data unused

Hi-Angel:
`looking-at-p` does not change match data and so should be a bit faster.


Hi-Angel:
Ping

yyoncho (approved):

PR #4344, opened 2024-02-25
Add additional `%%%%` to fix doom-modeline presentation of percentage progress

codeasone:
Without these additional %%%%s I see no %-sign or trailing space before the message in progress-status

With the additional %%%%, I get e.g. <nn>% <message> as expected.

The issue appears confined to the `lsp--progress-status` function and not any presentation logic downstream of it, since it is the `progress-status` value that is faulty.


yyoncho:
It works fine with vanilla emacs without your fix, right? If yes, then the fix should go in the doom modeline.

codeasone:
True, it does appear to work fine with `doom-modeline` disabled in my config. Apologies.

PR #4349, opened 2024-02-27
Use variable watcher with lsp-defcustom to support local variables.

### LINKED ISSUE #4337: lsp-defcustom not compatible with setting local variables

### Bug description

LSP servers which need to be configured by a directory local variable (e.g., project file for Ada LS), aren't initialized correctly when that configuration is set using a normal directory local variable.  This happens even when delaying the initialization of `lsp` to the `hack-local-variables-hook` as specified in the FAQ.  The reason for this is that `lsp-defcustom` utilizes a custom set function which invokes `lsp-register-custom-settings`, and this setter is not invoked using the hack-local-variables initialization, which just creates a buffer local variable (via `make-local-variable`).  

I've been able to work around this by using `eval` and `setopt` in my .dir-locals.el to cause the setter to be invoked, like this:
```elisp
;;; Directory Local Variables            -*- no-byte-compile: t -*-
;;; For more information see (info "(emacs) Directory Variables")

((ada-ts-mode . ((eval . (setopt lsp-ada-project-file "src/gtkada.gpr")))))
```

However, I think it might be better for `lsp-mode` to look through the buffer local variables when it starts up and call the setter for any of them which correlate to an `lsp-defcustom`.  That would allow the normal setting of directory local variables to work without resorting to `eval` and `setopt` as demonstrated above.

### Steps to reproduce

Setting a directory local variable in the normal way for a lsp-defcustom (i.e., lsp-ada-project-file) should cause the issue to be observed.  Using this approach, the lsp-defcustom's default value (e.g., "default.gpr") is used when first connecting with the server.

### Expected behavior

I expect the value specified in the directory-local variable to be used when communicating with the server.

Ada Language Server

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

brownts:
When custom variables are set locally using `hack-local-variables`, a custom variable's registered setter is not invoked.  To address this, a variable watcher is used instead to cause the setter to be invoked when the variable is updated.

Fixes #4337.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8551,16 +8551,24 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
   "Defines `lsp-mode' server property."
   (declare (doc-string 3) (debug (name body))
            (indent defun))
-  (let ((path (plist-get args :lsp-path)))
+  (let ((path (plist-get args :lsp-path))
+        (setter (intern (concat (symbol-name symbol) "--set"))))
     (cl-remf args :lsp-path)
     `(progn
        (lsp-register-custom-settings
         (quote ((,path ,symbol ,(equal ''boolean (plist-get args :type))))))
 
-       (defcustom ,symbol ,standard ,doc
-         :set (lambda (sym val)
-                (lsp--set-custom-property sym val ,path))
-         ,@args))))
+       (defcustom ,symbol ,standard ,doc ,@args)
+
+       ;; Use a variable watcher instead of registering a `defcustom'
+       ;; setter since `hack-local-variables' is not aware of custom
+       ;; setters and won't invoke them.
+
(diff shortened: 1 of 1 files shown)


brownts:
Ping @yyoncho

brownts:
@yyoncho or @jcs090218, is there any reason this can't be merged?  

This causes problems for Ada when trying to specify the project file in a directory local variable as described in the linked issue.

yyoncho (approved):

brownts:
@jcs090218, can you merge this PR?  This was reviewed/approved a while ago by @yyoncho, but never merged.

jcs090218:
Sorry, I was a little busy. 😓 Glad Ivan has already merged this! 🚀

brownts:
Thanks!

PR #4352, opened 2024-03-01
lsp-xml: Update JAR download URL

okomestudio:
It appears that the download URL for the lemminx JAR has changed. The existing URL fetches a page with a 403 Forbidden error, and therefore the downloaded JAR file is corrupt.

This PR:

- Updates the download URL 
- Updates the JAR version to the up-to-date release (0.27.0)

### DIFF
--- a/clients/lsp-xml.el
+++ b/clients/lsp-xml.el
@@ -287,9 +287,9 @@ The value for `enabled' can be always, never or onValidSchema."
   ("xml.catalogs" lsp-xml-catalogs)
   ("xml.trace.server" lsp-xml-trace-server)))
 
-(defconst lsp-xml-jar-version "0.21.0")
+(defconst lsp-xml-jar-version "0.27.0")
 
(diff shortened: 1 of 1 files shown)


clients/lsp-xml.el
```diff
@@ -287,9 +287,9 @@ The value for `enabled' can be always, never or onValidSchema."
   ("xml.catalogs" lsp-xml-catalogs)
   ("xml.trace.server" lsp-xml-trace-server)))
 
-(defconst lsp-xml-jar-version "0.21.0")
+(defconst lsp-xml-jar-version "0.27.0")
 
-(defconst lsp-xml-jar-name (format "org.eclipse.lemminx-%s-uber.jar" lsp-xml-jar-version))
+(defconst lsp-xml-jar-name (format "org.eclipse.lemminx-uber.jar" lsp-xml-jar-version))
```
jcs090218, line 292: I think we can remove the entire `(format` part.

okomestudio: Applied a fix. Thanks for the review and suggestion.

jcs090218 (approved):

jcs090218:
Thank you!

Issue #4365, opened 2024-03-13
`csharp-ls` can not be found


Hi-Angel:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

I installed `csharp-ls` and it is in `$PATH`. However if I open a `.cs` file and execute `lsp`, I get a `Unable to find installed server supporting this file` error. It suggests to install `omnisharp` or `csharp-ls`, but the latter is already installed.

lsp-servers for C, python and bash work fine, so it is specific to just this one.

### Steps to reproduce

1. Install `csharp-ls`
2. Start emacs and execute `lsp-start-plain`
3. Open a file `test.cs` that has the following content:
    ```
    namespace HelloWorld
    {
        class Hello {
            static void Main(string[] args)
            {
                System.Console.WriteLine("Hello World!");
            }
        }
    }
    ```
4. Execute in Emacs `lsp`

### Expected behavior

`csharp-ls` gets started

### Which Language Server did you use?

`csharp-ls`

### OS

Linux

### Error callstack

```emacs-lisp
Command "semgrep lsp" is not present on the path.
```

Hi-Angel:
Ha! I couldn't figure out by code of lsp-mode, it's quite complicated. But I executed a [`bpftrace` to print new processes](https://unix.stackexchange.com/questions/260162/how-to-track-newly-created-processes-in-linux/606831#606831) while executing a `lsp` and found this:

```
pid: 40505, comm: emacs, args: /bin/zsh -c dotnet tool list -g
pid: 40505, comm: zsh, args: dotnet tool list -g
```

so lsp is starting this tool. Apparently it assumes the tool will print if `csharp-ls` exists, but that doesn't happen, its output is empty. Even though `csharp-ls` does exist 😊 Looking at `lsp-csharp.el` it seems like the test for just checking if `csharp-ls` binary is present is missing. So that needs to be fixed.

psibi:
What is the output of `which csharp-ls` ?

And can you check the variable `exec-path` and see the directory where `csharp-ls` is present exists in this variable.

psibi:
[...]

PR #4368, opened 2024-03-13
Revert "lsp-csharp: add a test to check if dotnet tool csharp-ls is actually installed"

Hi-Angel:
This commit broke detection of `csharp-ls`. I am not sure what was the purpose of the
change. The function it added says it checks whether `csharp-ls` is present in
the system, however the function `lsp-csharp--cls-make-launch-cmd` always did the
same and better. So why have this additional check is not at all clear.

The function was assuming that a `dotnet tool list -g` should return
if `csharp-ls` is installed, however for me its output is just empty:

    λ dotnet tool list -g
    Package Id      Version      Commands
    -------------------------------------

Let's revert this commit.

Fixes: https://github.com/emacs-lsp/lsp-mode/issues/4365
CC: @razzmatazz the original commit author


yyoncho (approved):

razzmatazz:
Thank you!
I do not remember why it was added in the first place–good riddance.
Can confirm the change works for me.

PR #4383, opened 2024-03-19
Fix client settings cannot be overridden

### LINKED ISSUE #4382: Client settings cannot be overridden

### Bug description

Calling `lsp-register-custom-settings` twice does not properly override the value because the hash table is created with the default `:test 'eql`. Therefore, when I try to override a setting using a string that is equal to the one used previously, it's added to the hash table as a new key. 

### Steps to reproduce

```
(clrhash lsp-client-settings)
(lsp-register-custom-settings '(("foo" "original value")))
(lsp-register-custom-settings '(("foo" "new value")))

(let ((hash-table-contents nil))
  (maphash (lambda (k v) (push (cons k v) hash-table-contents))
	   lsp-client-settings)
  hash-table-contents)
;; => (("foo" "new value") ("foo" "original value"))

(gethash "foo" lsp-client-settings)
;; => nil
```

### Expected behavior

```
(let ((hash-table-contents nil))
  (maphash (lambda (k v) (push (cons k v) hash-table-contents))
	   lsp-client-settings)
  hash-table-contents)
;; => (("foo" "new value"))

(gethash "foo" lsp-client-settings)
;; => ("new value")
```

pyright-langserver (but it shouldn't matter)

### OS

MacOS

### Error callstack

_No response_

### Anything else?

I'll submit a PR in a minute.

vljosa-vivodyne:
Create hash table with `:test 'equal` so that
`lsp-register-custom-settings` can be called again to override settings.

Resolves #4382

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8514,7 +8514,7 @@ session workspace folder configuration for the server."
         (funcall initialization-options-or-fn)
       initialization-options-or-fn)))
 
-(defvar lsp-client-settings (make-hash-table)
+(defvar lsp-client-settings (make-hash-table :test 'equal)
   "For internal use, any external users please use
   `lsp-register-custom-settings' function instead")
 
--- a/test/lsp-mode-test.el
+++ b/test/lsp-mode-test.el
@@ -175,4 +175,13 @@
                        "textDocument/hover")))
              2)))
 
+(ert-deftest lsp-test--register-custom-settings-override ()
+  "Test that custom settings can be overridden."
+  (clrhash lsp-client-settings)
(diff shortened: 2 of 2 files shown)


yyoncho (approved):

yyoncho:
Thank you!

alanz:
For rust analyzer, references will show up bold, and mutable items underlined.  So making a small change to RA code, I get

Screenshot 2024-03-20 at 15 33 08

ra_fixture is both bold and underlined.  If you remove the `mut` (as it normally is), the underline disappears.

alanz:
We chose bold and underline to keep things simple for themes

Issue #4355, opened 2024-03-03
Add support for YANG LSP


esmasth:
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? No.

TypeFox/yang-lsp has support for YANG language, and can be integrated into emacs via this project.

esmasth:
I have started working on this issue, and have draft ready soon.

esmasth:
@jcs090218 can this be assigned to me now?

jcs090218:
Completed in #4390.

PR #4390, opened 2024-03-23
Introduce YANG LSP support

esmasth:
This addresses emacs-lsp/lsp-mode#4355

### DIFF
--- a/clients/lsp-yang.el
+++ b/clients/lsp-yang.el
@@ -1,0 +1,122 @@
+;;; lsp-yang.el --- YANG Client settings         -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2024 Siddharth Sharma
+
+;; Author: Siddharth Sharma <siddharth.sharma@ericsson.com>
+;; Keywords: languages, yang, lsp
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+;;; Commentary:
+
+;; LSP support for YANG using using an external language server. Currently
+;; the supported server is:
+;;
+;; yang-lsp (yls).
+;; See https://github.com/TypeFox/yang-lsp/blob/master/docs/Settings.md
+;; for setting up the user/project/workspace files.
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(defgroup lsp-yang nil
+  "LSP support for the YANG data modeling language using yang-lsp server."
+  :group 'lsp-yang
+  :link '(url-link "https://github.com/TypeFox/yang-lsp"))
+
+(defcustom lsp-yang-yls-version "0.7.6"
+  "yang-lsp server version to download.
+
+It has to be set before `lsp-yang.el' is loaded and it has to
+be available here: https://github.com/TypeFox/yang-lsp/releases/"
+  :type 'string
+  :group 'lsp-yang
+  :package-version '(lsp-mode . "8.0.1"))
+
+(add-to-list 'auto-mode-alist '("^yang\\.settings$" . jsonc-mode))
+
+(defcustom lsp-yang-yls-settings-schema-url
+  (format "https://raw.githubusercontent.com/TypeFox/yang-lsp/v%s/schema/yang-lsp-settings-schema.json"
+          lsp-yang-yls-version)
+  "URL for yang-lsp server settings schema"
+  :type 'string
+  :group 'lsp-yang
+  :package-version '(lsp-mode . "8.0.1"))
+
+(defcustom lsp-yang-yls-executable "yang-language-server"
+  "The yang-lsp server executable to use.
+
+Leave as just the executable name to use the default behavior of finding the
+executable with variable `exec-path'."
+  :group 'lsp-yang
+  :type 'string)
+
+(defcustom lsp-yang-yls-download-url
+  (format "https://github.com/TypeFox/yang-lsp/releases/download/v%s/yang-language-server_%s.zip"
+          lsp-yang-yls-version
+          lsp-yang-yls-version)
+  "Automatic download url for yang-lsp server"
+  :type 'string
+  :group 'lsp-yang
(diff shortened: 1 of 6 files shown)


jcs090218:
There is a compile error:

```
In toplevel form:
clients/lsp-yang.el:56:2: Error: assignment to free variable ‘lsp-json-schemas’
..........................Test lsp-byte-compilation-test backtrace:


    (ert-test-failed
     ((should
       (byte-compile-file library))
      :form
      (byte-compile-file "/home/runner/work/lsp-mode/lsp-mode/clients/lsp-yang.el")
      :value nil))
```

Can you update our CHANGELOG.org too? :)

esmasth:
I'll fix both in the next patch, thanks.

esmasth:
@jcs090218 request your rereview here

clients/lsp-yang.el
```diff
@@ -0,0 +1,126 @@
...
+
+(defcustom lsp-yang-yls-settings-schema-url
+  (format "https://raw.githubusercontent.com/TypeFox/yang-lsp/v%s/schema/yang-lsp-settings-schema.json"
+          lsp-yang-yls-version)
+  "URL for yang-lsp server settings schema"
+  :type 'string
+  :group 'lsp-yang
+  :package-version '(lsp-mode . "8.0.1"))
+
+(setq lsp-json-schemas
```
jcs090218, line 59: Hmm... where does this variable come from? 🤔 This looks a bit odd to me. Usually, we don't change the variable at the root level.

lsp-mode.el
```diff
@@ -190,7 +190,8 @@ As defined by the Language Server Protocol 3.16."
      lsp-sonarlint lsp-tailwindcss lsp-tex lsp-terraform lsp-toml lsp-ttcn3
      lsp-typeprof lsp-v lsp-vala lsp-verilog lsp-vetur lsp-volar lsp-vhdl
      lsp-vimscript lsp-wgsl lsp-xml lsp-yaml lsp-ruby-lsp lsp-ruby-syntax-tree
-     lsp-solidity lsp-sqls lsp-svelte lsp-steep lsp-tilt lsp-trunk lsp-zig lsp-jq)
+     lsp-solidity lsp-sqls lsp-svelte lsp-steep lsp-tilt lsp-trunk lsp-zig lsp-jq
+     lsp-yang)
```
jcs090218, line 194: We tried to sort this in alphabetical order.

clients/lsp-yang.el
esmasth: This comes from lsp-json.el in this repo. Is it better to append to the list? How does one do that for `repeat alist`?

lsp-mode.el
esmasth: Will sort the whole list in next patch.

clients/lsp-yang.el
jcs090218: I guess my question is, why is the YANG lsp client required to modify JSON language server settings? 🤔 Most language server clients are isolated, so it's a bit odd to see something like this.

esmasth: The json schema association settings could be set in `init.el` or `.emacs` for user specified associations as per variable provided by `lsp-json.el`. json schema associations are not always possible to encode in the json itself due to variability in editor support and how strictly json schema is written.

The association, hence, is provided via json schema handling subsystem configuration, which is the `lsp-json.el` in lsp-mode based emacs world as I see. The validation et al under the hood happens via vscode json language server.

Now, it should be possible for "emacs extensions" which are json based config consumers, to provide schema and association (like it is possible via vscode extensions), as an addition to a list of known associations, else each user of `lsp-yang.el` for example will have to make the association themselves, or there would be a central database of association configuration, neither of which scale.

esmasth: I will remove this setting for this commit, since `setq` overrides any default settings. There should be a mechanism provided by `lsp-json.el` which allows for appending any default or further `setq`.

lsp-mode.el
```diff
@@ -174,23 +174,24 @@ As defined by the Language Server Protocol 3.16."
...
-     lsp-mint lsp-move lsp-nginx lsp-nim lsp-nix lsp-magik lsp-mojo lsp-metals
-     lsp-mssql lsp-nushell lsp-ocaml lsp-openscad lsp-pascal lsp-perl lsp-perlnavigator
-     lsp-pls lsp-php lsp-pwsh lsp-pyls lsp-pylsp lsp-pyright lsp-python-ms
-     lsp-purescript lsp-r lsp-racket lsp-remark lsp-ruff-lsp lsp-rf lsp-rubocop
-     lsp-rust lsp-semgrep lsp-shader lsp-solargraph lsp-sorbet lsp-sourcekit
-     lsp-sonarlint lsp-tailwindcss lsp-tex lsp-terraform lsp-toml lsp-ttcn3
-     lsp-typeprof lsp-v lsp-vala lsp-verilog lsp-vetur lsp-volar lsp-vhdl
-     lsp-vimscript lsp-wgsl lsp-xml lsp-yaml lsp-ruby-lsp lsp-ruby-syntax-tree
-     lsp-solidity lsp-sqls lsp-svelte lsp-steep lsp-tilt lsp-trunk lsp-zig lsp-jq)
+  '(ccls lsp-actionscript lsp-ada lsp-angular lsp-ansible lsp-asm lsp-astro
```
jcs090218, line 177: Missing a space here.

```
'( ccls ...
```

jcs090218 (approved):

esmasth:
Thanks @jcs090218. Could you please guide me on how to resolve the failing checks? I don't know where to start.

lsp-mode.el
```diff
@@ -795,6 +796,7 @@ Changes take effect only when a new session is started."
     ("^PKGBUILD$" . "shellscript")
     ("^go\\.mod\\'" . "go.mod")
     ("^settings.json$" . "jsonc")
+    ("^yang\\.settings$" . "jsonc")
```
esmasth, line 799: @jcs090218 should this be `"^yang\\.settings\\'"`? What's the meaning of `\\'` usage?

Also shouldn't `"^settings.json$"` be `"^settings\\.json$"`?

jcs090218: > should this be `"^yang\\.settings\\'"`? What's the meaning of `\\'` usage?

`\\'` matches the end of a string; `$` matches the end of line. So both work.

> Also shouldn't `"^settings.json$"` be `"^settings\\.json$"`?

Yes, you are right.

jcs090218:
> Thanks @jcs090218. Could you please guide me on how to resolve the failing checks? I don't know where to start.

Don't worry about the failing tests. It's irrelevant to this PR.

esmasth:
Thanks a lot @jcs090218. I have resolved the review comments hopefully, and also fixed the `settings.json` match pattern as a separate commit. OK to merge? I don't have permissions.

jcs090218 (approved):

jcs090218:
Thank you for your contribution!

Issue #4192, opened 2023-10-14
`lsp-completion--company-match` sometimes freeze my Emacs instance


sebastiencs:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

`lsp-completion--company-match` is sometimes called with cached data (from previous completion):

When I auto-complete something at line 1, the auto-completion is fine.
But when I go to a line way below, let's say 1000, and start auto-completing, it's using some data from line 1 (specifically it's using the `start-point` at line 1) so the prefix is every characters between line 1 and 1000.

I'm not sure why we are using the same cache at 2 different places, there might be a reason.
A fix would be to re-compute the correct `start-point` 

### Steps to reproduce

Auto-complete at 2 different places (very distant lines)

### Expected behavior

smooth completion

### Which Language Server did you use?

rust-analyzer

### OS

Linux

PR #4394, opened 2024-03-24
lsp-completion: support completion-lazy-hilit and quicker company-match

kiennq:
Currently `lsp-mode` doesn't highlight the candidates in the way `corfu` supports.
This PR adds:
- support for `completion-lazy-hilit` that can be used by `corfu` and alike completion UI. See https://github.com/minad/corfu/discussions/441
-  Also, in `lsp-completion--company-match` we can use `match-data` when possible, instead of going through matching again.
- Finally, in the work-around for #4192, we can just use `line-beginning-position` as a guard for cached start point. That would be better than having a take whole long substring and then dissect it for the first line.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -272,35 +272,34 @@ The MARKERS and PREFIX value will be attached to each candidate."
   (lsp--while-no-input
     (->>
      (if items
-         (-->
-             (let (queries fuz-queries)
-               (-keep (-lambda ((cand &as &plist :label :start-point :score))
-                        (let* ((query (or (plist-get queries start-point)
-                                          (let ((s (buffer-substring-no-properties
-                                                    start-point (point))))
-                                            (setq queries (plist-put queries start-point s))
-                                            s)))
-                               (fuz-query (or (plist-get fuz-queries start-point)
-                                              (let ((s (lsp-completion--regex-fuz query)))
-                                                (setq fuz-queries
-                                                      (plist-put fuz-queries start-point s))
-                                                s)))
-                               (label-len (length label)))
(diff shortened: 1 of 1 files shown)


sebastiencs:
> Finally, in the work-around for https://github.com/emacs-lsp/lsp-mode/issues/4192, we can just use line-beginning-position as a guard for cached start point. That would be better than having a take whole long substring and then dissect it for the first line.

That seems good 

> support for completion-lazy-hilit 

Cool, I didn't know that `completion-lazy-hilit`

sebastiencs (approved):

kiennq:
@yyoncho Do you think you can merge this change?

jcs090218:
I've tried this patch, but the completion is no longer working. 🤔

kiennq:
> I've tried this patch, but the completion is no longer working. 🤔

The completion via `company-mode`?

jcs090218:
> The completion via company-mode?

Yes.

kiennq:
I've tried with `emacs -Q` and it seems still working fine.
The company configuration looks like this

``` elisp
(setenv "LSP_USE_PLISTS" "1")
(setq company-idle-delay 0.05
      company-minimum-prefix-length 2
      company-selection-wrap-around t
      company-require-match 'company-explicit-action-p
      company-backends `(company-capf))
```

jcs090218:
> I've tried with emacs -Q and it seems still working fine.
> The company configuration looks like this

Hmm... It's working this time. I am unsure why it didn't work the first time I tested it. Sorry for the noise.

jcs090218 (approved):

PR #4396, opened 2024-03-27
lsp-xml: add supports for native binary language server

kiennq:
The xml language server that's created by Redhat doesn't require Java.
We can add that as a compliment to the current jar version.

Looking at the vscode, it does the switch between jar version and native version based on `java` is available or not, I used the same approach here instead of creating a new xml language server declaration.

### DIFF
--- a/clients/lsp-xml.el
+++ b/clients/lsp-xml.el
@@ -356,6 +356,7 @@ The value for `enabled' can be always, never or onValidSchema."
   (lambda () (or (and (lsp-xml-has-java?) lsp-xml-prefer-jar `("java" "-jar" ,lsp-xml-jar-file))
                  `(,lsp-xml-bin-file)))
   "Xml server command."
+  :type '(choice (repeat string) (function))
   :group 'lsp-xml
   :package-version '(lsp-mode . "6.1"))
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

kiennq:
@jcs090218, @yyoncho, the test for 26.x is not running due to it has been dropped from support.
Can you remove those rules for merge as well?

jcs090218:
I don't have the permission. @yyoncho and @ericdallo have it.

ericdallo:
@jcs090218 I gave you owner permission too, since you are quite active in lsp-mode for some time :)

jcs090218:
> @jcs090218 I gave you owner permission too, since you are quite active in lsp-mode for some time :)

That's very kind! Thank you! :D

jcs090218:
> Can you remove those rules for merge as well?

It's gone now. :)

clients/lsp-xml.el
```diff
@@ -287,13 +287,20 @@ The value for `enabled' can be always, never or onValidSchema."
   ("xml.catalogs" lsp-xml-catalogs)
   ("xml.trace.server" lsp-xml-trace-server)))
 
+(defconst lsp-xml-has-java? (executable-find "java"))
```
jcs090218, line 290: Can we move it elsewhere so it doesn't get executed at the root level? 🤔

kiennq: Good point.

jcs090218:
There are compile errors:

```
In toplevel form:
lsp-xml.el:355:1:Warning: defcustom for ‘lsp-xml-server-command’ fails to
    specify type
lsp-xml.el:355:1:Warning: defcustom for ‘lsp-xml-server-command’ fails to
    specify type
```

PR #4398, opened 2024-03-28
style(lsp-rust): Mnior clean up

jcs090218:
~~`~/` home directory under Windows is `c:/Users/USERPROFILE/AppData/Roaming`, but `.cargo/registry/src` and `.rustup/toolchains` is under `c:/Users/USERPROFILE/`. This patch fixes this issue.~~

Just minor clean up.


clients/lsp-rust.el
```diff
@@ -68,7 +68,12 @@
   :group 'lsp-rust-rls
   :package-version '(lsp-mode . "6.1"))
 
-(defcustom lsp-rust-library-directories '("~/.cargo/registry/src" "~/.rustup/toolchains")
+(defcustom lsp-rust-library-directories
+  (let ((home-dir (pcase system-type
+                    ('windows-nt (getenv "USERPROFILE"))
```
kiennq, line 73: The `~` should expand to `%USERPROFILE%` under windows, why do we need this change? Just to expand the path?

jcs090218: You are right. Sorry for the noisy. 😓

PR #4399, opened 2024-03-28
lsp-html - update to new version of custom data and add new configuration options

### LINKED ISSUE #4327: lsp-html: update settings
Microsoft's HTML language server configuration settings have evolved (see [here](https://github.com/sublimelsp/LSP-html/blob/master/LSP-html.sublime-settings) or [here](https://github.com/microsoft/vscode/blob/main/extensions/html-language-features/package.json)).

Most notably, `html.experimental.customData` has been renamed to `html.customData`, which means `lsp-html` currently can't be used in conjunction with web components.

themkat:
# Updating custom data
I could not get custom data to work on my machine with a simple option/property/setting. From reading the vscode html tools source code, it seems that this extension uses the property internally in the same way we do here (read diffs in this PR). In summary:
- Send the data paths to the language server in the initialization options field.
- Handle a request for the content of these files in the client code.

When the content of these files have been sent to the server, it seems to work perfectly. I can complete, view hover information and more. I used various projects, like [the example project from the vscode team](https://github.com/microsoft/vscode-custom-data/tree/main/samples/helloworld). I added an extra custom data definition with some bogus to make it more interesting:
```json
{
  "version": 1.1,
  "tags": [
    {
      "name": "dafuq",
      "description": "The super cool new tag for the cool kids",
      "attributes": [
        { "name": "isThisShit" },
        {
          "name": "test",
          "values": [
            {
              "name": "is-working"
            }
          ]
        }
      ]
    }
  ]
}

```

image
image
image

Completion of attributes also works perfectly. I just have to be sure to activate yas-minor-mode (or similar templating system). 

Fixes #4327
(probably)

# New configuration options
Added `lsp-html-hover-documentation` and `lsp-html-hover-references`, which are both on by default (even if not set from the language server side). I think it makes sense to be able to set these, especially on older machines where I try to get more performance out of turning off features.

### DIFF
--- a/clients/lsp-html.el
+++ b/clients/lsp-html.el
@@ -33,7 +33,8 @@
 
 (defcustom lsp-html-custom-data []
   "A list of JSON file paths that define custom tags, properties and other HTML
-syntax constructs. Only workspace folder setting will be read."
+syntax constructs. Only workspace folder setting will be read.
+All json file paths should be relative to your workspace folder."
   :type 'lsp-repeatable-vector
   :group 'lsp-html
   :package-version '(lsp-mode . "8.0.1"))
--- a/clients/lsp-html.el
+++ b/clients/lsp-html.el
@@ -23,7 +23,6 @@
 ;;; Code:
 
 (require 'lsp-mode)
-(require 'f)
 
 (defgroup lsp-html nil
   "LSP support for HTML, using vscode's built-in language server."
@@ -189,7 +188,7 @@ styles."
 
 ;; Caveat: uri seems to be sent as a single length vector.
 (defun lsp-html--get-content (_workspace files callback)
-  "Helper function for getting the content of a URI when the language server requests it."
+  "Helper function for getting the content of a URI/filename."
   (let* ((filename (aref files 0))
          (uri (f-join (lsp-workspace-root) filename))
          (file-content (f-read-text uri)))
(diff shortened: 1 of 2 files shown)


clients/lsp-html.el
```diff
@@ -182,9 +206,13 @@ styles."
                                      (cons (lsp-package-path 'html-language-server)
                                            lsp-html-server-command-args)))
                   :activation-fn (lsp-activate-on "html")
+                  :major-modes '(html-mode)
```
jcs090218, line 209: Why do we need this since we already have `:activation-fn`. 🤔

themkat: Had some issues with lsp-mode not starting for html files and giving me an error message on no servers available. Removed the line and will investigate the issues further. Might be some weirdness with my local setup.


Removed the major-mode line and rebased 🙂

jcs090218:
One small warning:

```
lsp-html.el:191:2: Warning: docstring wider than 80 characters
```

clients/lsp-html.el
```diff
@@ -23,19 +23,21 @@
 ;;; Code:
 
 (require 'lsp-mode)
+(require 'f)
```
jcs090218, line 26: I don't think require is necessary since `lsp-mode` already has it.

themkat: Removed it in latest commit

themkat:
> One small warning:
> 
> ```
> lsp-html.el:191:2: Warning: docstring wider than 80 characters
> ```

Fixed it in latest commit 🙂

jcs090218 (approved):

jcs090218:
Thank you!

PR #4401, opened 2024-03-28
lsp-help-mode: use Lisp macro "delay-mode-hooks" for safe

leuven65:
lsp-help-mode doesn't run the hook, the problem seems to be caused by delay-mode-hooks.
The problem can be solved by using the Lisp macro "delay-mode-hooks".


jcs090218 (approved):
LGTM.

PR #4408, opened 2024-03-29
lsp-pwsh: add supports for powershell-ts-mode

kiennq:


clients/lsp-pwsh.el
```diff
@@ -300,7 +300,7 @@ Must not nil.")
   :new-connection (lsp-stdio-connection #'lsp-pwsh--command
                                         (lambda ()
                                           (f-exists? lsp-pwsh-pses-script)))
-  :major-modes lsp-pwsh--major-modes
```
jcs090218, line 303: Should we remove the variable too?

jcs090218 (approved):

PR #4412, opened 2024-03-31
Base QML setup

### LINKED ISSUE #4131: Support Qt/QML language server
In Qt 6 there is LSP server for QML language,
executable is qmlls6.

It would be nice to add support of QML into lsp-mode.

themkat:
Fixes #4131. That issue wished for the some configuration for QML (Qt Modeling Language). I did not find any special configuration options for it, but just having a variable to configure makes life easier for the users (instead of duplicating lsp-register-clients in peoples private .emacs.d). I use qml-mode together with it, and it works quite well when I played around with it 🙂

### DIFF
--- a/clients/lsp-qml.el
+++ b/clients/lsp-qml.el
@@ -1,0 +1,43 @@
+;;; lsp-qml.el --- description -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2024 emacs-lsp maintainers
+
+;; Author: emacs-lsp maintainers
+;; Keywords: lsp, qt, qml, gui,
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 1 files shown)


jcs090218:
A few steps are missing; you can use this PR: https://github.com/emacs-lsp/lsp-mode/pull/4385/files as the reference.

themkat:
> A few steps are missing; you can use this PR: #4385 as the reference.

Sorry 🙁 Seems like I forget these every time I try to add a new client....

Added it now in newest commit 🙂 Unsure which server URL to use, as these are installed together with the rest of the Qt tooling when installing.

jcs090218:
> Unsure which server URL to use, as these are installed together with the rest of the Qt tooling when installing.

Normally, we want to point to documentation or instruction. LGTM, though.

themkat:
> > Unsure which server URL to use, as these are installed together with the rest of the Qt tooling when installing.
> 
> Normally, we want to point to documentation or instruction. LGTM, though.

Good thing I pointed to a documentation page then 😄 🎆

clients/lsp-qml.el
```diff
@@ -0,0 +1,43 @@
+;;; lsp-qml.el --- description -*- lexical-binding: t; -*-
```
jcs090218, line 1: Can you put a reasonable description? Thanks!

themkat: Sure 🙂 Added a short version of the one I put in "Commentary"

jcs090218 (approved):

jcs090218:
LGTM, thank you!

Issue #4420, opened 2024-04-08
Random regressions due to lsp overriding user settings via `lsp-register-custom-settings`


Hi-Angel:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Commit 782e1dc15a590ae9aa2c9b7e39c4cc8a65952e13 by @LaurenceWarne introduced a regression and broke mypy completely for me. That happened, because [I had set the same lsp-server settings](https://github.com/Hi-Angel/dotfiles/blob/5ca50352a0a298af5d223aded54964ce61eddc79/.emacs#L402-L405) that the commit modified.

Further research shown that there's nothing wrong with the commit per se, instead the problem is in how `lsp-mode` handles lsp-server settings. Whatever user sets with `lsp-register-custom-settings` is being overwritten by `lsp-mode`, leading to random problems like this one.

Instead `lsp-register-custom-settings` should not overwrite settings that a user has set.

### Steps to reproduce

1. Create `/tmp/.emacs` with the following content:
    ```lisp
    (package-initialize)
    (use-package lsp-mode
      :config
      (setq lsp-log-io t)
      (lsp-register-custom-settings '(("pylsp.plugins.pylsp_mypy.enabled" t t))))
    ```
2. Start emacs as `emacs -Q -l /tmp/.emacs test.py`
3. Start `lsp-mode` in the `test.py` buffer by evaluating <kbd>M</kbd>-<kbd>x</kbd> <kbd>lsp</kbd>
4. Switch to buffer with communication log by evaluating <kbd>M</kbd>-<kbd>x</kbd> <kbd>lsp-workspace-show-log</kbd>
5. Search for word `pylsp_mypy` and look at the value of `enabled` just below the `pylsp_mypy` line

### Expected behavior

It is `true`.

### Which Language Server did you use?

pylsp

### OS

Linux

Hi-Angel:
Should be fixed by https://github.com/emacs-lsp/lsp-mode/pull/4421

PR #4421, opened 2024-04-08
treewide: do not overwrite lsp-client-settings changed by user

Hi-Angel:
If a user has set a setting inside `lsp-client-settings`, changing its value will result in a silent bug in user configuration, that looks like a regression from `lsp-mode` mode update. Let's avoid that by adding a new function `lsp-register-new-settings` that avoids overwritting whatever was set by user and make use of it treewide.

While at it, prohibit `lsp-register-custom-settings` from internal use for all the same reasons.

Fixes: https://github.com/emacs-lsp/lsp-mode/issues/4420


Hi-Angel:
Hmm, apparently github workflows [doesn't work well with exclamation sign](https://github.com/actions/runner/issues/753). Had to quote per advice over at the link, hopefully it will work as expected.

Hi-Angel:
Yeah, it didn't fail with a "unknown command", so it is working.

yyoncho:
Thanks for the contribution but I have some concerns. `lsp-register-custom-settings` is supposed to be replaced by lsp-defcustom which does not have the option to use the `new` version because it will deviate from vanilla `defcustom`. This I would rather recommend using something like...

```
(with-eval-after-load 'lsp-python
   (lsp-register-custom-settings ...
```

... in your config.

Hi-Angel:
> Thanks for the contribution but I have some concerns. `lsp-register-custom-settings` is supposed to be replaced by lsp-defcustom which does not have the option to use the `new` version because it will deviate from vanilla `defcustom`.

Sorry, I'm a bit confused here… Are you talking about the PR changes or about an alternative way a user may configure lsp-server settings? In the first case I'm unclear what you want me to do. In the latter case, the documentation for `lsp-register-custom-settings` does not mention `lsp-defcusom`. I might add a commit with docs on that if you want, but that is a bit orthogonal to the problem the PR solves, because `lsp-register-custom-settings` has existed for years for user level configuration, so changes like 782e1dc15a5 will be happening and regressing random users.

> This I would rather recommend using something like...
> 
> ```
> (with-eval-after-load 'lsp-python
>    (lsp-register-custom-settings ...
> ```
> 
> ... in your config.

Thank you, nice to know there's a workaround.

yyoncho:
> so changes like [782e1dc](https://github.com/emacs-lsp/lsp-mode/commit/782e1dc15a590ae9aa2c9b7e39c4cc8a65952e13) will be happening and regressing random users.

The point is that when we migrate that file to lsp-defcustom you will still get the same issue unless you use the `with-eval-after-load`

Hi-Angel:
Ah, so you want me to fix `lsp-defcustom` as well, right?

yyoncho:
No. I mean that lsp-defcustom is supposed to work that way and you cannot fix it because it won't behave as expected. So the only option seems to be to fix your config otherwise we will introduce inconsistencies (lsp-defcustom wont work like defcustom).

Hi-Angel:
> I mean that lsp-defcustom is supposed to work that way

Why? You earlier mentioned that it's supposed to work similarly to `defcustom`, and the latter does not override whatever was set by user. You can test it yourself: evaluate in emacs `(setq foo nil)` and then `(defcustom foo t "test")`, and then check value of `foo`. It will be `nil`.

yyoncho:
> > I mean that lsp-defcustom is supposed to work that way
> 
> Why? You earlier mentioned that it's supposed to work similarly to `defcustom`, and the latter does not override whatever was set by user. You can test it yourself: evaluate in emacs `(setq foo nil)` and then `(defcustom foo t "test")`, and then check value of `foo`. It will be `nil`.

If you go to a defcustom form and eval only it - it will override the value.

Hi-Angel:
> > > I mean that lsp-defcustom is supposed to work that way
> >
> >
> > Why? You earlier mentioned that it's supposed to work similarly to `defcustom`, and the latter does not override whatever was set by user. You can test it yourself: evaluate in emacs `(setq foo nil)` and then `(defcustom foo t "test")`, and then check value of `foo`. It will be `nil`.
>
> If you go to a defcustom form and eval only it - it will override the value.

It doesn't, here's a testcase I just made with a separate file:

```haskell
λ cat test.el
(setq foo nil)
(defcustom foo nil "test")
λ emacs --batch -l test.el --eval "(print foo)"

nil
```

Hi-Angel:
Any news here?

Hi-Angel:
From what I read, questions raised seems resolved, so… I imagine it'd be useful to have some solution to the user regressions discussed. I kind of hoped to have this merged faster, before more users are gonna be hit by the Python commit, but another week passed, so the hope becomes lesser. However, we still can make the code future-proof.

Hi-Angel:
Ping

Hi-Angel:
Ping 2

yyoncho:
> > > > I mean that lsp-defcustom is supposed to work that way
> > > 
> > > 
> > > Why? You earlier mentioned that it's supposed to work similarly to `defcustom`, and the latter does not override whatever was set by user. You can test it yourself: evaluate in emacs `(setq foo nil)` and then `(defcustom foo t "test")`, and then check value of `foo`. It will be `nil`.
> > 
> > 
> > If you go to a defcustom form and eval only it - it will override the value.
> 
> It doesn't, here's a testcase I just made with a separate file:
> 
> ```haskell
> λ cat test.el
> (setq foo nil)
> (defcustom foo t "test")
> λ emacs --batch -l test.el --eval "(print foo)"
> 
> nil
> ```

It overrides it if you eval only the defcustom form interactively.

Hi-Angel:
> ```haskell
> (setq foo nil)
> (defcustom foo t "test")
> ```

What version are you testing with? I just did a `emacs -Q test.el` and manually evaluated first the `setq` then the `defcustom`, and `foo` afterwards is `nil`.

Also, I don't see how interactive behavior matters in this context.

Hi-Angel:
Oh, actually wait, I'm not sure why I had different result but now I am re-testing and indeed after iteractive re-evaluation it changes.

But anyway, why does interactive behavior matter here? It is frequent that interactive behavior differ from non-interactive one, and the bug we are facing here is for non-interactive coide.

wyuenho:
If you want to set some client prefs, just set the defcustom variables. Why do you have to file a 36 file PR for that?

Hi-Angel:
> If you want to set some client prefs, just set the defcustom variables. Why do you have to file a 36 file PR for that?

Sorry, this sounds a bit vague, can you please elaborate in terms of code for the regression this PR fixes?

Hi-Angel:
@wyuenho actually, so save you time: I think your question comes from misunderstanding the situation. I presume, you're assuming an lsp-mode "defcustom" exists for certain lsp server property, but this isn't always the case, and it wasn't the case in the regression. So people instead set such settings manually with `lsp-register-custom-settings`, which is purposed exactly for that as you can see by its name.

The problem then, is that `lsp-register-custom-settings` completely breaks when upstream uses `lsp-defcustom` for the same setting someone changed. IOW, whenever upstream adds a defcusom for a missing lsp-server setting, it potentially breaks someone's config out there.

Hi-Angel:
This is PR isn't that interesting in the context of **my** regression *(because started using the new variables, and the PR was hanging there far too long so every other people who also had the problem had to go through the problem as well)*, but it's more interesting in terms of ecosystem, because, let me emphasize, **every** new defcustom lsp-mode adds for any lsp-server is potentially breaking someone's config and takes out an hour or so debugging time from that person. This is certainly very bad situation.

PR #4428, opened 2024-04-15
lsp-completion-at-point: looking back a few more chars for trigger-char

### LINKED ISSUE #4419: Completions disappear after dot + symbol

### Bug description

`lsp-mode` treats `.` specially for many languages, as in, it provides completions immediately even though `company-minimum-prefix-length` isn't satisfied. Which makes sense of course. Unfortunately, `lsp-mode` loses completion after typing out the next symbol, instead of reducing them to ones that start with the symbol.

So e.g. `foo.` pops up completions, but `foo.c` kills the pop up even if there are candidates.

Also asked as a question here: https://emacs.stackexchange.com/questions/80977/prevent-lsp-company-from-losing-completion-after-dot-symbol

Side note: `elgot`, the competitor mode, handles that correctly.

### Steps to reproduce

1. Create `/tmp/.emacs` as follows:
    ```lisp
    (package-initialize)
    (use-package company)
    (use-package lsp)
    ```

2. Start emacs as `emacs -Q -l /tmp/.emacs test.py`
3. Execute <kbd>M</kbd>-<kbd>x</kbd> <kbd>lsp</kbd>
3. Type `"".`

    A completion should appear

4. Type `c` to reduce completions to ones starting with that letter

### Expected behavior

Completions get reduced to ones starting with `c`, i.e. `capitalize()`, `casefold()`, etc.

`lsp-python`, `lsp-go`, `clangd`.

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

kiennq:
Fix #4419

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -525,7 +525,15 @@ The MARKERS and PREFIX value will be attached to each candidate."
        :company-require-match 'never
        :company-prefix-length
        (save-excursion
-         (and (lsp-completion--looking-back-trigger-characterp trigger-chars) t))
+         (defvar company-minimum-prefix-length)
+         (let ((bounds-left (max (line-beginning-position) (- (point) company-minimum-prefix-length)))
+               triggered-by-char?)
+           (while (and (> (point) bounds-left)
+                       (not (equal (char-after) ?\s))
+                       (not triggered-by-char?))
+             (setq triggered-by-char? (lsp-completion--looking-back-trigger-characterp trigger-chars))
+             (goto-char (1- (point))))
+           (and triggered-by-char? t)))
        :company-match #'lsp-completion--company-match
        :company-doc-buffer (-compose #'lsp-doc-buffer
                                      #'lsp-completion--get-documentation)


lsp-completion.el
```diff
@@ -525,7 +525,15 @@ The MARKERS and PREFIX value will be attached to each candidate."
        :company-require-match 'never
        :company-prefix-length
        (save-excursion
-         (and (lsp-completion--looking-back-trigger-characterp trigger-chars) t))
+         (defvar company-minimum-prefix-length)
```
jcs090218, line 528: I rarely see something like this. Why not put it in the root level? 🤔

jcs090218:
We are using company-specific variables. Should we handle it for users who don't use company-mode?

kiennq:
> We are using company-specific variables. Should we handle it for users who don't use company-mode?

I guess no. This is company-specific property anyway. Using this with `corfu` has no issue.

lsp-completion.el
kiennq: It would be the same, since I only use the `company-minimum-prefix-length` here, I keep this in this functin for cleanliness. Functionality wise, this is just forward declaration and should be okay to put anywhere before its usage.

jcs090218 (approved):

guillaumebrunerie:
It does break with corfu for me, I get `(void-variable company-minimum-prefix-length)` at every character I type.

guillaumebrunerie:
Ok, I managed to fix locally it by replacing
```
(defvar company-minimum-prefix-length)
```
by
```
(defvar company-minimum-prefix-length 1)
```
(i.e. actually providing a value to that variable, not sure if that value matters)

But I think it is pretty weird to have a `defvar` here, as the variable created by it will be global anyway (if I understand correctly). Why not use `boundp` instead to check if the variable exists, and do something else instead?

Also I guess that if we load `company-mode` after `lsp-mode`, then `company-mode` won’t be able to provide its own default value and docstring to the defvar'd variable?

lsp-completion.el
```diff
@@ -525,7 +525,15 @@ The MARKERS and PREFIX value will be attached to each candidate."
...
+         (let ((bounds-left (max (line-beginning-position) (- (point) company-minimum-prefix-length)))
```
dgutov, line 529: I don't think you need `company-minimum-prefix-length` here at all.

Just see what `company-clang--prefix` does, for example: https://github.com/company-mode/company-mode/blob/master/company-clang.el#L336-L339

As a backend, you have two choices: either return a number (the real of "virtual" prefix length), or `t`, if the current prefix is preceded by a prefix character such as `.`. The framework itself will then do the prefix length check if needed (if you returned a number and not `t`).

jcs090218:
cc @kiennq

lsp-completion.el
kiennq: Yes, I do not need to use `company-minimum-prefix-length` and can just keep looking back until the begin of line or the first whitespace character, however using the `company-minimum-prefix-length` is better as it can be quite short and doesn't affect the perf much. I will use other mean to get the approximated length to check then.

kiennq:
Btw, to be clear, this issue only happens in `lsp-mode` and not other is because we're the only one who doesn't trust the completion boundary gotten by symbol at point of Emacs.
The boundary should be provided by the lsp server and we use a heuristic to determine that after gotten the data from the server. That also means the boundary/prefix for each candidate is different.

lsp-completion.el
dgutov: You could just limit the search by the beginning of the current line. Or the first whitespace char (like is already done). This is fast enough not to require additional performance tweaks, IMO.

PR #4429, opened 2024-04-16
Allow Flycheck to determine a diagnostic end position when needed.

brownts:
When a Language Server doesn't support an end position for a diagnostic (by specifying an end position the same as the start position) Flycheck should be asked to compute the end position using it's `flycheck-highlighting-mode` setting.  This is accomplished by passing nil as the end-column to `flycheck-error-new`.

This provides consistent behavior between Flymake and Flycheck as the Flymake integration already checks for this condition and uses it's similar functionality (i.e., `flymake-diag-region`) to compute an end position.


jcs090218 (approved):
Make sense! Thank you! :)

PR #4435, opened 2024-04-20
A couple minor fixes for Flymake diagnostics

brownts:


lsp-diagnostics.el
```diff
@@ -298,7 +307,7 @@ See https://github.com/emacs-lsp/lsp-mode."
                       (when (= start end)
                         (if-let ((region (flymake-diag-region (current-buffer)
                                                               (1+ start-line)
-                                                              character)))
+                                                              (1+ character))))
```
jcs090218, line 310: Can you elaborate why +1? 🤔  Thanks!

jcs090218 (approved):

lsp-diagnostics.el
brownts: `flymake-diag-region` expects the line and column values to start at 1, not start at 0 like LSP specifies.  Thus, both the line and column need incrementing when calling `flymake-diag-region`.  The line number was already incremented but the column was not.

Note that this logic is only executed when the LSP server doesn't support an end position (i.e., the start and end position are the same), so `flymake-diag-region` is used to compute it.

brownts:
I decided to take a different approach with respect to the default severity level.  Instead of adding an additional `defcustom` for Flymake (which seemed clunky to have two separate configurations for the same thing), I decided to make it more general and elevate it out of `lsp-diagnostics` to `lsp-mode`.  The reason for this is because there were multiple places that really need this value.  The `lsp-modeline`, `lsp-headerline`, `lsp-mode` and `lsp-diagnostics` all need this information.  I also noticed that in those places the absence of the severity provided by the server manifested differently.  Some didn't handle it at all (which caused an error to be raised), some ignored it completely, and some  hard-coded it as an "error", which might have been inconsistent to the existing Flymake `defcustom`.  Due to the needs outside of the `lsp-diagnostics` package, I made this more general.  Now, any place that needs the severity and it isn't provided, will utilize this `defcustom`.

On a side note, I have also seen `lsp-treemacs` cause an error as well, since it's not checking whether the server provided a severity before attempting to use it, but that's another fix in another repo, lol.

lsp-diagnostics.el
```diff
@@ -131,18 +127,19 @@ g. `error', `warning') and list of LSP TAGS."
...
-                          ((= tag lsp/diagnostic-tag-unnecessary) 'unnecessary)
-                          ((= tag lsp/diagnostic-tag-deprecated) 'deprecated)))
-                       tags)))
+  (let* ((severity (or severity
+                       (lsp-diagnostics-severity->numeric
+                        lsp-diagnostics-default-severity)))
+         (level (cond ((= severity lsp/diagnostic-severity-error)       'error)
+                      ((= severity lsp/diagnostic-severity-warning)     'warning)
+                      ((= severity lsp/diagnostic-severity-information) 'info)
+                      ((= severity lsp/diagnostic-severity-hint)        'info)))
```
jcs090218, line 136: Using `pcase` is probably cleaner.

And the one below, too.

jcs090218: I guess this isn't any cleaner. 🤔 

```elisp
(pcase severity
        ((pred (= lsp/diagnostic-severity-error))       'error)
        ((pred (= lsp/diagnostic-severity-warning))     'warning)
        ((pred (= lsp/diagnostic-severity-information)) 'info)
        ((pred (= lsp/diagnostic-severity-hint))        'info)
        (_ 'info))
```

It's okay to leave it as it is.

lsp-mode.el
```diff
@@ -634,6 +634,15 @@ diagnostics have changed."
...
   :group 'lsp-mode)
 
+(defcustom lsp-diagnostics-default-severity 'error
+  "Error level to use when the server does not report one."
+  :type '(choice (const :tag "Error"       error)
+                 (const :tag "Warning"     warning)
+                 (const :tag "Information" info)
+                 (const :tag "Hint"        hint))
+  :group 'lsp-mode
+  :package-version '(lsp-mode . "9.0.1"))
```
jcs090218, line 644: Should we keep this in `lsp-diagnostics.el`? 🤔

brownts: Hi @jcs090218, the reason I put it in lsp-mode.el is because there are multiple users, not just lsp-diagnostics.el.  lsp-modeline, lsp-headerline, lsp-mode and lsp-diagnostics all need this information, so it seemed more general and likely belongs in lsp-mode.el.  It answers the general question of what the user wants to do in the absence of LSP severity information.  lsp-diagnostics.el is just one such user of this information.

I placed it in lsp-mode.el near a couple other diagnostic-related user options (i.e., lsp-after-diagnostics-hook and lsp-diagnostics-updated-hook) to be consistent there.  It appears that lsp-diagnostics.el is strictly the diagnostics provider (i.e., Flymake/Flycheck) interfacing, but other LSP diagnostics related information is kept in lsp-mode.el.

PR #4439, opened 2024-04-25
fi: Attempt to fix CI

jcs090218:


jcs090218:
Seems like an upstream issue. 🤔 See https://github.com/actions/setup-python/issues/852.

jcs090218:
Remove test in `macos-latest 27.2 (true)` due to https://github.com/purcell/nix-emacs-ci/pull/268.

PR #4444, opened 2024-04-27
lsp-rust: Allow debugging with GDB 14.

daym:
Since GDB 14 now supports DAP natively, we can also just use GDB instead of vscode-cpptools. Try that in lsp-rust.

See also <https://github.com/emacs-lsp/dap-mode/issues/790>, which is required.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -1513,8 +1513,8 @@ and run a compilation"
 (defun lsp-rust-analyzer-debug (runnable)
   "Select and debug a RUNNABLE action."
   (interactive (list (lsp-rust-analyzer--select-runnable)))
-  (unless (featurep 'dap-cpptools)
-    (user-error "You must require `dap-cpptools'"))
+  (unless (or (featurep 'dap-cpptools) (featurep 'dap-gdb))
+    (user-error "You must require `dap-cpptools' or 'dap-gdb'"))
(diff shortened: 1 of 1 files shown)


clients/lsp-rust.el
```diff
@@ -1541,7 +1541,7 @@ and run a compilation"
               (`() (user-error "No compilation artifacts or obtaining the runnable artifacts failed"))
               (`(,spec) spec)
               (_ (user-error "Multiple compilation artifacts are not supported")))))
-         (list :type "cppdbg"
+         (list :type (if (featurep 'dap-gdb) "gdb" "cppdbg")
```
jcs090218, line 1544: Is `gdb` preferred over `cppdb`? 🤔

jcs090218 (approved):

yyoncho (approved):

clients/lsp-rust.el
daym: I wasn't sure how else to do this.

As far as I understand it: If the user doesn't do `(require 'dap-gdb)` in their emacs config file then dap-gdb.el won't be used anyway (even though the file is on disk), right?

So this way should be backward compatible with whatever users have configured now.

On the other hand, if someone adds `(require 'dap-gdb)` to their emacs config now then lsb-rust will switch over to gdb.

I didn't think of more use cases than that yet.

Issue #2997, opened 2021-07-14
Renaming from xref buffer doesn't work


tekhnus:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

`*xref*` buffers support `xref-query-replace-in-results` command, which renames the symbol in the list matches as requested.

This command seems to be broken in the `*xref*` buffers created specifically by `lsp-find-references`, at least with `lsp-clangd`, at least for me:)

When I use `xref-find-references` with the default (etags) backend instead of the backend provided by `lsp-mode`, `xref-query-replace-in-results` works well.

### Steps to reproduce

1. Open a C project
2. Place the cursor over any symbol
3. Execute either `lsp-find-references` or `xref-find-references` command
4. From `*xref*` buffer, execute `xref-query-replace-in-results` (bound to `r` key by default)
5. At the second prompt, type any new name
5. Get a "No suitable matches here" message

### Expected behavior

`xref-query-replace-in-results` should perform the renaming inside each listed match.

### Which Language Server did you use?

lsp-clangd

### OS

Linux

dgutov:
For this to work, the "find references" matches returned by lsp-mode should define the method `xref-match-length`.

PR #4453, opened 2024-05-19
fix: use xref-make-match instead of xref-make

JasZhe:
This allows us to use xref-query-replace-in-results, when exporting to an xref buffer via embark for instance, since it needs xref-match-length to be defined.

Fixes: https://github.com/emacs-lsp/lsp-mode/issues/2997


jcs090218 (approved):

jcs090218:
Thank you for your contribution! 🚀

PR #4463, opened 2024-05-27
Fix project root discovery when project features aren't loaded.

brownts:
It's possible for the corresponding project package (i.e., projectile or project) to not be loaded yet but there to exist autoloads for the functions.  In that situation, using `featurep` to check for the existence of the package will fail.  Therefore, `fboundp` is used instead to check for the existence (possibly an autoload) of the needed functions.


jcs090218 (approved):

jcs090218:
Make sense. Thank you!

PR #4465, opened 2024-05-29
rust-analyzer client: extract signature in hover content

blahgeek:
The full hover content from rust-analyzer usually looks like this:

Screenshot_20240529_232050
Screenshot_20240529_232031

By default, when `lsp-eldoc-render-all` is nil, only the first line (the comment) is show.

This PR implements the extraction method (like clangd) to return the first non-comment line.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -28,6 +28,7 @@
 (require 'ht)
 (require 'dash)
 (require 'lsp-semantic-tokens)
+(require 's)
 
 (defgroup lsp-rust nil
   "LSP support for Rust, using Rust Language Server or rust-analyzer."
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):
LGTM.

jcs090218:
Thank you!

scturtle:
Another solution is in https://github.com/emacs-lsp/lsp-mode/pull/1740 and it's [included in doom emacs now](https://github.com/doomemacs/doomemacs/pull/7754).

blahgeek:
Thanks for the link. I did not know that exists.

Based on my reading, these two PRs are not exactly the same though. https://github.com/emacs-lsp/lsp-mode/pull/1740 aims to extract a meaningful section for a type (for hovering on a type definition), the solution is rather heuristic and more format-specific, and arguably more opinionated. This PR merely strip comments from the result (usually for hovering on a variable).

So personally I can see why this PR is more suitable and accepted, or at least as a first step (I'm surely biased :) )  But of course I'm happy to see more improvement on this function by anyone

PR #4466, opened 2024-05-31
Added AUCTeX LaTeX-mode as a Latex mode

lima-limon-inc:
AUCTeX uses a different major mode for editing .tex files (LaTeX-mode).

Before, lsp-mode would not recognize LaTeX-mode as a major mode equivalent to latex-mode; and thus texlab would never be executed. 

I should point out that I added the LaTeX-mode to digestif list of modes; despite not using digestif. 

Would love to get some feedback. 
Thanks in advance!
 
PS: Should I update the change-log as stated in the guidelines?

Fixes: #4467

### DIFF
--- a/clients/lsp-tex.el
+++ b/clients/lsp-tex.el
@@ -46,7 +46,7 @@
 
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection lsp-clients-digestif-executable)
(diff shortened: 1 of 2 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

lima-limon-inc:
On the contrary, thank you!

Have a good week :blush:

Issue #3022, opened 2021-07-28
lsp-fsharp: failure when installing client


Tarmil:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

On a Windows system that has an `unzip` command in the PATH, the fsautocomplete client doesn't install properly.

### Steps to reproduce

On Windows:
* Put `unzip.exe` somewhere in the `PATH`.
* Try to install lsp-fsharp.

### Expected behavior

fsautocomplete installs properly.

### Which Language Server did you use?

lsp-fsharp

### OS

Windows

### Error callstack

```emacs-lisp
Contacting host: api.github.com:443
Contacting host: github.com:443
Wrote c:/Users/LDENUZ~1/AppData/Local/Temp/fsautocompleterj0Xsa.zip
The syntax of the command is incorrect.
Could not execute because the specified command or file was not found.
Possible reasons for this include:
  * You misspelled a built-in dotnet command.
  * You intended to execute a .NET program, but dotnet-c:/Users/ldenuziere/.emacs.d/.cache/lsp/fsautocomplete/fsautocomplete.dll does not exist.
  * You intended to run a global tool, but a dotnet-prefixed executable with this name could not be found on the PATH.
```

### Anything else?

The problem is that when `unzip` is present, the install function assumes that it's on a Unix system and runs a command intended for `sh`; but on Windows, it is misinterpreted by `cmd`. A possible fix would be to switch the conditions here to use `powershell` in priority if present: https://github.com/emacs-lsp/lsp-mode/blob/b4e51450188acf41f8273542db299936461087a4/clients/lsp-fsharp.el#L212-L214

razzmatazz:
On linux, at least, there is also an issue that the server installed had no permissions on the files (probably due to faulty .zip archive or unzip params?) -- I need to do `chmod +rwx -R *` on the extracted directory

s-kostyaev:
Fixed now

[...]

PR #4472, opened 2024-06-09
Switch tool detection for lsp-unzip (#3022)

Horrih:
It caused issues on windows if unzip was found on the PATH, because it fell back to the bash command + unzip.

Changing in order of detection fixes the issue.


yyoncho:
This will force powershell on linux which IIRC was causing trouble in the past. Can you add additional condition (if on windows and if powershell there - use it)

Horrih:
Just pushed a new version, is this what you had in mind ?
Just checked it on both my systems, it seems to work fine.

Btw, thanks a lot for your selfless work on lsp-mode.
Lsp-mode is the sole reason I came back to emacs in 2019 years after I gave up on the tags system.

yyoncho:
> ((executable-find "powershell") lsp-ext-pwsh-script

can you keep that too in the end?

Horrih:
Sure, is that better ?

I rebased onto master + squashed the previous commits

jcs090218 (approved):

jcs090218:
LGTM! Thank you for taking care of this!

PR #4484, opened 2024-06-26
Remove Alire support from lsp-mode Ada client.

brownts:
Alire support is now available directly in the Ada Language Server and handles the functionality currently implemented in lsp-mode.  In addition, when the project file is not specified by the LSP client, the language server will intelligently locate the project file using `alr` when an Alire configuration file exists.  Therefore, the default setting of the project file has been changed (from "default.gpr" to nil) to enable the server to automatically locate the project file in normal situations.  The server will also automatically find a project file for a non-Alire project if it uniquely exists in the root folder (i.e., no other project file exists in this folder).  The new project file default setting provides for a better out-of-the-box configuration to support common situations.

See discussion #4468 for additional details.

### DIFF
--- a/clients/lsp-ada.el
+++ b/clients/lsp-ada.el
@@ -33,18 +33,31 @@
   :tag "Language Server"
   :package-version '(lsp-mode . "6.2"))
 
-(lsp-defcustom lsp-ada-project-file "default.gpr"
-  "Set the project file full path to configure the language server with.
-  The ~ prefix (for the user home directory) is supported.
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

brownts:
Moved the safety predicate for the lsp-defcustom to an autoload, so that it's available during .dir-locals.el processing, which occurs prior to the lsp package being loaded in hack-local-variables-hook.

jcs090218 (approved):

jcs090218:
Thank you!

PR #4492, opened 2024-07-10
Add gnu/linux arm64 support in lsp--system-arch

chaos95:
Currently, only `darwin` (macos) includes aarch64/arm64 as a detected architecture when setting `lsp--system-arch`.

This change adds detection for arm64 on `gnu/linux` system-types, as well as adding an appropriate cond to the mapping in lsp-lua. The only other place this var is used is in lsp-ada, which already accounts for `gnu/linux` + `arm64`.


jcs090218 (approved):

jcs090218:
Thank you!

PR #4494, opened 2024-07-11
fix: lsp--find-workspaces-for workspace/executeCommand respects the target command

kassick:
Calling

    (lsp--find-workspaces-for '(:method "workspace/executeCommand"
                                :params (:command "arbitrary-command")))

would return *any* workspace that had the :executeCommandProvider capability registered.

This would result in a given command being forwarded for several active workspaces that wouldn't know how to handle that command. Some servers (such as ruff) would raise a KeyError for unknown commands -- causing the minibuffer to, sometimes, show an error.

This change introduces a new :check-message callback that can be used in lsp-method-requirements -- this callback receives the workspace and the message to be sent.

Ths entry for workspace/executeCommand has been updated to check if the target workspace can execute the required command.


kassick:
Do you folks think this is the right approach to avoid sending commands that a server will reject? Any suggestions?

kiennq:
> This would result in a given command being forwarded for several active workspaces that wouldn't know how to handle that command. Some servers (such as ruff) would raise a KeyError for unknown commands -- causing the minibuffer to, sometimes, show an error.

Can we just place the check of `lsp-can-execute-command?` before sending the `workspace/executeCommand` instead?

kassick:
> Can we just place the check of `lsp-can-execute-command?` before sending the `workspace/executeCommand` instead?

_We do just that -- in `lsp--find-workspace-for-method`., called in `lsp--send-request-async` to find a server that has the target capability (executeCommandProvider for "workspace/executeCommand") and which accepts the message, when `:check-message` is defined. **But I assume you mean if we could place the check earlier in the call stack, so I'll answer assuming that was the question** ;)_

Do you mean adding code in, say, `lsp-workspace-command-execute` to find the servers that support the command and then `(with-lsp-workspace w (lsp-request "workspace/executeCommand" ...))` ?

That would fix the issue I have with inline completions of copilot-ls [1] . Filtering the workspaces in `lsp-workspace-command-execute` before calling `lsp-request` would fix that specific issue and all places that use `lsp--execute-command`.

But *several other clients in lsp-mode send "workspace/executeCommand" requests directly via `lsp-request`* ([terraform](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-terraform.el#L242) and [sqls](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-sqls.el#L95) in this project). If those servers are used alongside some plugin server that responds errors to unknown commands, the issue would persist.

On the other hand, there may be misbehaving servers that do not correctly inform all the commands they support in `workspace/executeCommand` -- so maybe being able to bypass this check via explicitly sending a message is a good thing. But is they use `lsp--execute-command` counting on the broadcast behavior, they'd have to migrate to calling lsp-request explicitly.

(we could also provide a flag and a `(with-lsp-broadcasting-commands ...)` defun also)

What do you think would be the right approach here, @kiennq ?

-----

[1] copilot asks to call a `com.github....` command after accepting an inline completion and that ends up broadcast to all active servers. Ruff does not like to be spammed, so it responds with an error, which eventually pops up in the minibuffer. Not life threatning, but annoying.

kassick:
I'm currently running lsp-mode on ffc0f56074f129148cd79cb4ea26cd093454d7d6 with the following `:override` advice:

(defun kzk/lsp-workspace-command-execute (command &optional args)
  "Execute workspace COMMAND with ARGS."
  (condition-case-unless-debug err
      (let ((params (if args
                        (list :command command :arguments args)
                      (list :command command)))
            (ws (lsp--find-workspaces-for "workspace/execute-command")))
        (cl-loop for w in ws do
                 (with-lsp-workspace w
                   (when (lsp-can-execute-command? command)
                     (lsp-request "workspace/executeCommand" params)))))
    (error
     (error "`workspace/executeCommand' with `%s' failed.\n\n%S"
            command err))))```

PR #4508, opened 2024-07-27
Fix PHP Language Server detection

### LINKED ISSUE #4507: Test command for PHP Language Server returns true when lsp-clients-php-server-command is nil.

### Bug description

The test command for the PHP Language Server (php-ls) returns true when lsp-clients-php-server-command is nil. 

When invoking `lsp` for the first time, `lsp--filter-clients` will check all available clients for `lsp--supports-buffer?` and `server-binary-present?`. When it checks `php-ls`, `server-binary-present?` will rely on the output of the `test-command` passed to `lsp-stdio-connection`. I

Since the `test-command` runs before the `command`, `lsp-clients-php-server-command` is never set. This means that it will fail the first condition [here](https://github.com/emacs-lsp/lsp-mode/blob/master/clients/lsp-php.el#L65), and then will return `t`.

This results in `php-ls` appearing in the list of available servers despite not being installed. All the other servers (save for `phpactor`) have a higher priority. So `php-ls` will only be run if either no other server is installed, or if only `phpactor` is installed.

---
Subsequent invocations of `lsp` will behave as expected, since the `command` for `php-ls` sets `lsp-clients-php-server-command`.

### Steps to reproduce

1. Fresh emacs
2. Don't install Intelephense or Serenata
3. Don't install PHP Language Server
4. Install `phpactor`(Optional)
5. Open a php file

### Expected behavior

`php-ls` shouldn't appear in the available servers list if the PHP language server isn't installed.

`lsp-php`

### OS

MacOS

### Error callstack

```emacs-lisp
First Invocation:
Command "semgrep lsp" is not present on the path.                                                                                                                                                                                             
Command "/usr/local/bin/phpactor language-server" is present on the path.                                                                                                                                                                     
Command "serenata.phar -u 0" is not present on the path.                                                                                                                                                                                      
Command "semgrep lsp" is not present on the path.                                                                                                                                                                                             
Command "/usr/local/bin/phpactor language-server" is present on the path.                                                                                                                                                                     
Command "serenata.phar -u 0" is not present on the path.                                                                                                                                                                                      
Found the following clients for [php file]: (server-id phpactor, priority -4), (server-id php-ls, priority -3)                                                                                     
The following clients were selected based on priority: (server-id php-ls, priority -3)


Second Invocation:                                                                                                                                        
Command "semgrep lsp" is not present on the path.                                                                                                                                                                                             
Command "/usr/local/bin/phpactor language-server" is present on the path.                                                                                                                                                                     
Command "serenata.phar -u 0" is not present on the path.                                                                                                                                                                                      
[home dir]/.composer/vendor/felixfbecker/language-server/bin/php-language-server.php is not present.                                                                                                                                        
Command "semgrep lsp" is not present on the path.                                                                                                                                                                                             
Command "/usr/local/bin/phpactor language-server" is present on the path.                                                                                                                                                                     
Command "serenata.phar -u 0" is not present on the path.                                                                                                                                                                                      
[home dir]/.composer/vendor/felixfbecker/language-server/bin/php-language-server.php is not present.                                                                                                                                        
Found the following clients for [php file]: (server-id phpactor, priority -4)                                                                                                                      
The following clients were selected based on priority: (server-id phpactor, priority -4)                                                                                                                                                      
Creating watchers for following 25 folders:
```

### Anything else?

I believe I've identified the issue and will be submitting a PR shortly.

AlexanderSeto:
Currently, if `lsp-clients-php-server-command` is not set, the test-command for  the PHP Language Server (`php-ls`) will return as true. This results in `php-ls` appearing in the list of available servers even if not installed. This commit fixes the check to return false when not set.

Closes #4507


jcs090218 (approved):

jcs090218:
I'm a little confused. Since the function holds lambda, does that mean the result won't be accurate? 🤔

AlexanderSeto:
I apologize, I'm not entirely sure I follow what you mean. 

Currently, if `lsp-clients-php-server-command` is not initially set, the condition evaluates to `nil`, and so the lambda returns `t`.  This means `lsp` will present `php-ls` in the list of available servers despite it not existing.

cweiske:
I can confirm this patch fixes the bug on Debian 12 with Emacs 28.2 and lsp-mode 10.0.0

jcs090218 (approved):

jcs090218 (approved):

PR #4512, opened 2024-07-28
Add lsp-flush-delayed-changes-before-next-message option

necto:
Some LSP clients spam the server with message, such as "textDocument/codeAction" requests on every keystroke, so if we require synchronizing the delayed changes before the next LSP message, it looses its meaning: In most cases the next cursor movement comes right after a change, so `lsp--send-no-wait` forces the change notification out before it sends the codeAction request.

Let the users of the slow LSP servers that depend on this change notification throttling customize this behavior to choose their own tradeoff.


jcs090218 (approved):

jcs090218:
Thank you!

PR #4513, opened 2024-07-28
Fix #4501: Auto adjust stale diagnostic ranges on document changes

### LINKED ISSUE #4501: Auto-adjust diagnostic ranges on document modification while waiting for update from LSP server
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? No

I will repost from the discussion https://github.com/emacs-lsp/lsp-mode/discussions/4497 in order to connect this issue with coming PRs.

# The problem
I am using SonarLint CFamily via [lsp-sonalint](https://github.com/emacs-lsp/lsp-sonarlint). Unfortunately, the CFamily analyzer is pretty slow at the moment, and it takes several seconds or even minutes to analyze a file after it has been changed.
In the meantime, while I edit the contents, issue locations become stale and point to the wrong parts.
Here is an [example](https://asciinema.org/a/cBG08MXKe8lI32NmYYjjKyWQl):
[asciicast](https://asciinema.org/a/cBG08MXKe8lI32NmYYjjKyWQl)

The SonarLint diagnostic about a TODO comment starts off correctly on the TODO comment (on line 13).
When I delete a line, the diagnostic line stays the same, but the the TODO comment is now actually on line 12,
so I get a confusing issue location.
I then wait for a few seconds for SonarLint to publish the updated set of diagnostics, and the diagnostic location gets the correct line 12.

However, I often edit the files continuously and this problem arises on every line deletion/insertion, so the diagnostics constantly point to wrong locations.

I've noticed similar problem with clangd but to a lesser degree because of a shorter turn-around time.

I could set `lsp-diagnostic-clean-after-change` to remove the diagnostics altogether, but that means I'd need to restrain myself from changing a file for a few seconds or a minute to see the diagnostics.

# Solution
I propose to adjust the diagnostic locations on every change.
Particularly, I would go through the list of diagnostics and check which ones have a location that is after the change, and increment/decrement them to the corresponding amount.

# Implementation
My plan is the following:
- [ ] Add a few regression tests around the display of diagnostics using a mock LSP server.
- [ ] Modify the "position" plist from `(:line :character)` to `(:line :character :point)` (where `:point` is optional), and set `:point` to the
  absolute offset from the beginning of the buffer, i.e. the `(point)` value at the corresponding line and character. I'd need to clarify what to do with the "virtual buffers" created for org-mode code snippets.
- [ ] In the `lsp-on-change` add an alternative to cleaning up all diagnostics - adjusting the diagnostics:
  Iterate over the diagnostics and for each end of a diagnostic range
  - if its `:point` is before the `start` of the modified region, keep it as is
  - if its `:point` is after the `start + length` of the modified region, increase/decrease it to reflect the length shift
  - if its `:point` is between `start` and `start + length`, set it to `end`
  
  Then remove the diagnostics with empty ranges.
  
# Discussion
- Have you experienced similar issues?
- Is my plan sound? Do you think it can degrade `lsp-mode` performance? Any pitfalls I should watch out for?

necto:
To keep the diagnostic ranges actual after document edits and before LSP sends an update (can take multiple seconds or even minutes for SonarLint C++ analysis), allocate an overlay for each diagnostic and use it as the reference for its range.
This change works well together with #4512, which spares the LSP server from playing catch-up with the user modifications.

`lsp-diagnostics` and `lsp--get-buffer-diagnostics` now have an additional optional argument: `update-ranges?` that refreshes the diagnostic ranges.

This fixes #4501.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -8,6 +8,7 @@
   * Added a new optional ~:action-filter~ argument when defining LSP clients that allows code action requests to be modified before they are sent to the server. This is used by the Haskell language server client to work around an ~lsp-mode~ parsing quirk that incorrectly sends ~null~ values instead of ~false~ in code action requests.
   * Add support for C# via the [[https://github.com/dotnet/roslyn/tree/main/src/LanguageServer][Roslyn language server]].
   * Add basic support for [[https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics][pull diagnostics]] requests.
+  * Improve positions of stale diagnostics on a changed buffer via overlays.
 
 ** 9.0.0
   * Add language server config for QML (Qt Modeling Language) using qmlls.
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -154,7 +154,7 @@ CALLBACK is the status callback passed by Flycheck."
 
   (remove-hook 'lsp-on-idle-hook #'lsp-diagnostics--flycheck-buffer t)
 
-  (->> (lsp--get-buffer-diagnostics)
+  (->> (lsp--get-buffer-diagnostics t)
        (-map (-lambda ((&Diagnostic :message :severity? :tags? :code? :source?
                                     :range (&Range :start (start &as &Position
                                                                  :line      start-line
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2277,24 +2277,48 @@ Common usecase are:
 
 (defvar lsp-diagnostic-stats (ht))
 
-(defun lsp-diagnostics (&optional current-workspace?)
-  "Return the diagnostics from all workspaces."
-  (or (pcase (if current-workspace?
-                 (lsp-workspaces)
-               (lsp--session-workspaces (lsp-session)))
-        (`() ())
-        (`(,workspace) (lsp--workspace-diagnostics workspace))
-        (`,workspaces (let ((result (make-hash-table :test 'equal)))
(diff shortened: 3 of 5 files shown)


necto:
It is also compatible with the experimental support for code snippets inside org-mode documents, see
[`lsp-mock-server-org-flycheck-updates-diags-with-delay`](https://github.com/emacs-lsp/lsp-mode/blob/00954dc79b43003d22a4f05565f2bf73ab169883/test/lsp-mock-server-test.el#L1175)

jcs090218:
Can you resolve conflicts? Thank you!

necto:
> Can you resolve conflicts? Thank you!

done

msrd0:
I tried this PR and when I add/remove a line (regardless of whether that line is in front of or behind the diagnostic)  the diagnostics move for a split second but then completely disappear. I've checked `lsp-diagnostic-clean-after-change` is set to `nil`. Am I doing something wrong?

necto:
Hi @msrd0 
How can I reproduce this behavior?

msrd0:
@necto my emacs config is here: https://gitlab.com/msrd0/dotfiles/-/blob/master/HOME/.emacs.d/init.el?ref_type=heads#L63

Here's a video of the problem:

[2025_02_26_1517.webm](https://github.com/user-attachments/assets/34a5877b-3b33-4638-b3d1-99393242c488)

As you can see, after I made sure there's an error in the file, everytime I save the error appears, and after I add a newline, the error disappears a short time after I made that edit, and only reappears when I save (emacs tells you at the bottom of the screen when it saves the file).

Issue #4250, opened 2023-12-04
"error in process filter: Unable to find Content-Length header." over TRAMP after update to 2023.11.24 


JeremyWeed:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Seeing this in version 2023.11.24, rolling back to 2023.08.23 solves the issue.
```
error in process filter: Args out of range: "ntent-Length: 40^M
^M
{\"id\":114,\"jsonrpc\":\"2.0\",\"result\":null}Content-Length: 40^M
^M
{\"id\":115,\"jsonrpc\":\"2.0\",\"result\":null}Content-Length: 305^M
^M
{\"id\":116,\"jsonrpc\":\"2.0\",\"result\":{\"contents\":{\"kind\":\"markdown\",\"value\":\"### param `param_name`  \\n\\n---\\nType: `const char *`  \\n\\n---\\n```cpp\\n// In ClassName::FunctionName\\nconst char *param_name\\n```\"},\"range\":{\"end\":{\"character\":62,\"line\":75},\"start\":{\"character\":48,\"line\":75}}}}Content-Length: 262^M
^M
{\"id\":117,\"jsonrpc\":\"2.0\",\"result\":{\"contents\":{\"kind\":\"markdown\",\"value\":\"### type-alias `size_t`  \\n\\n---\\nType: `unsigned int`  \\n\\n---\\n```cpp\\ntypedef unsigned int size_t\\n```\"},\"range\":{\"end\":{\"character\":70,\"line\":75},\"start\":{\"character\":64,\"line\":75}}}}", 60, 16^M
error in process filter: Args out of range: "ntent-Length: 40^M
^M
{\"id\":114,\"jsonrpc\":\"2.0\",\"result\":null}Content-Length: 40^M
^M
{\"id\":115,\"jsonrpc\":\"2.0\",\"result\":null}Content-Length: 305^M
^M
[...]

PR #4519, opened 2024-08-01
Fix processing of multibyte messages

JeremyWeed:
* Too many characters are read from back-to-back multibyte messages due to a conversion from byte-length to character length

* Updated to immediately convert messages to unibyte when processing

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7041,8 +7041,8 @@ server. WORKSPACE is the active workspace."
         leftovers body-length body chunk)
     (lambda (_proc input)
       (setf chunk (if (s-blank? leftovers)
-                      (encode-coding-string input 'utf-8-unix)
-                    (concat leftovers (encode-coding-string input 'utf-8-unix))))
+                      (encode-coding-string input 'utf-8-unix t)
+                    (concat leftovers (encode-coding-string input 'utf-8-unix t))))
 
       (let (messages)
         (while (not (s-blank? chunk))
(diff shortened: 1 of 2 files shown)


JeremyWeed:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/4250

lsp-mode.el
```diff
@@ -7041,8 +7041,8 @@ server. WORKSPACE is the active workspace."
         leftovers body-length body chunk)
     (lambda (_proc input)
       (setf chunk (if (s-blank? leftovers)
-                      input
-                    (concat leftovers input)))
+                      (encode-coding-string input 'utf-8-unix)
```
kiennq, line 7044: nit: using the 3rd parameters of nocopy might be better due to we don't have to duplicate the string.
Can you check if it still working okay with that?

kiennq (approved):

nvimtor:
Any follow-ups? `lsp-mode` is unusable to me due to the `error in process filter: Unable to find Content-Length header.` error.

JeremyWeed:
Hi, sorry, lost track of this.  Can confirm that using the third param of `encode-coding-string` does appear to speed it up locally, and that I've been running with this fix locally with no detectable issues for the past three months.  Revised the pull request.

nvimtor:
@JeremyWeed same; I had forked your repository and rebased it with this one and seems like I have no issues anymore

PR #4521, opened 2024-08-02
Enable nixd in nix-ts-mode too

SKyletoft:
nixd was only set to run in nix-mode while nil and rnix were set to run in both nix-mode and nix-ts-mode. Also changed the id of nix-ts-mode from Nix to nix for consistency


jcs090218 (approved):

jcs090218:
Thank you!

PR #4522, opened 2024-08-04
feat: Add action-handlers for code-lens provided ruby-lsp-rails

eggc:
# Why

The language server `ruby-lsp` includes default add-on called `ruby-lsp-rails`.
`ruby-lsp-rails` provides additional commands like the following page:

https://shopify.github.io/ruby-lsp-rails/RubyLsp/Rails/CodeLens.html

I would like to use this feature, but it cannot be enabled in Emacs.

# What

I have implemented action-handlers to support additional commands.
Although I am not deeply familiar with the LSP specification and its details, I referred the following code:

https://github.com/emacs-lsp/lsp-mode/blob/c8d8bd0f5c40123821bf7c90afa5b6abb05074cb/clients/lsp-sqls.el#L174-L180

---

Please review this at your convenience. Thank you!

### DIFF
--- a/clients/lsp-ruby-lsp.el
+++ b/clients/lsp-ruby-lsp.el
@@ -43,11 +43,35 @@
    (if lsp-ruby-lsp-use-bundler '("bundle" "exec"))
    '("ruby-lsp")))
 
+(defun lsp-ruby-lsp--open-file (arg_hash)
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
LGTM. Thank you! :D

PR #4531, opened 2024-08-26
feat: Update TS config

jcs090218:
Discussed on Discord.

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -537,12 +537,23 @@ workspace."
           (const "non-relative"))
   :package-version '(lsp-mode . "6.1"))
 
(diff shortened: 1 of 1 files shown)


KaranAhlawat:
Doesn't seem to work if I just set `lsp-typescript-import-module-specifler-preference` to `non-relative` in my `.dir-locals.el` file

KaranAhlawat:
https://github.com/typescript-language-server/typescript-language-server/blob/master/docs/configuration.md

Looking at this, it looks like, `workspace/didChangeConfiguration` no longer supports the `importModuleSpecifierPreference`, and it has to be passed in the `initializationOptions` itself.

PR #4535, opened 2024-08-31
docs: fix `lsp-sqls-workspace-config-path` behavior

suzuki:
## About

The current `lsp-sql` documentation does not match the actual behavior.

> Alternatively, you can leave lsp-sqls-workspace-config-path to the default "workspace" value, and put a json file in <project>/.sqls/config.json containing

This is incorrect. The current behavior is to define "root" instead of "workspace."

## Codes

https://github.com/emacs-lsp/lsp-mode/blob/c36b95be6625dac5a37d3874a1a738e0c84ac39f/clients/lsp-sqls.el#L72-L79

## Notes

I am not a native English speaker. Please let me know if there are any unnatural phrases in this pull request.

### DIFF
--- a/docs/manual-language-docs/lsp-sqls.md
+++ b/docs/manual-language-docs/lsp-sqls.md
@@ -14,7 +14,10 @@ root_file: docs/manual-language-docs/lsp-sqls.md
 
 ```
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

Issue #4451, opened 2024-05-08
Support `ruff server`


rassie:
According to https://github.com/astral-sh/ruff/pull/10158, `ruff server` is a re-implementation of `ruff-lsp` in Rust and just by that property it is very probable that `ruff-lsp` will be deprecated in near future. Would be nice to have a `lsp-mode` client for `ruff server` by that point.

indigoviolet:
This is now available in beta: https://github.com/astral-sh/ruff/tree/main/crates/ruff_server#setup


Here's the blogpost about it: https://astral.sh/blog/ruff-v0.4.5

indigoviolet:
A very basic version that seems to be working for me:

```elisp
(with-eval-after-load 'lsp-mode
  (lsp-register-client
   (make-lsp-client :new-connection (lsp-stdio-connection '("ruff" "server" "--preview"))
     :major-modes '(python-mode)
     :priority 1
     :add-on? t
     :multi-root t
     :server-id 'ruff-server)
    )
  )
(setq! lsp-disabled-clients (python-mode . (ruff-lsp)))
```

PBorocz:
FWIW (as a Emacs elisp neophyte), this worked for me:
```
(with-eval-after-load 'eglot
  (add-to-list 'eglot-server-programs '(python-ts-mode . ("~/.local/bin/ruff-lsp"))))
```
went to:
```
(with-eval-after-load 'eglot
  (add-to-list 'eglot-server-programs '(python-ts-mode . ("~/.local/bin/ruff" "server"))))
```
at the end of my `use-package eglot` definition (against a ruff installed using pipx)

FrauH0lle:
I am not sure if `eglot` configuration will help `lsp-mode` users ...

What @indigoviolet wrote works, however even shorter is to modify `lsp-ruff-lsp-server-command`:

```el
(with-eval-after-load 'lsp-mode
  (setq lsp-ruff-lsp-server-command '("ruff" "server")))
```

jcs090218:
The server is now supported. Added in #4543.

PR #4546, opened 2024-09-16
Remove unsupported settings in `ruff server`

Daisuke897:
I removed unsupported settings in `ruff server`.
https://docs.astral.sh/ruff/editors/migration/#unsupported-settings

I also changed `lsp-ruff-ruff-args` to be used under `new-connection` function to remove the setting `args`.

related issue: #4451


jcs090218 (approved):

jcs090218:
Thank you!

PR #4548, opened 2024-09-17
feat: register library fn for ruby-lsp server

merrickluo:
- taken custom variable from lsp-solargraph
- add ~/.asdf to the default value


jcs090218 (approved):

jcs090218:
Thank you!

PR #4556, opened 2024-09-22
Add basic support for futhark mode

SKyletoft:
Added support for Futhark by [copying in the config from a gist by the creator of the language and lsp-server](https://gist.github.com/athas/c7b79f8f91ce31963bfdaa5e05d8f66a)

### DIFF
--- a/clients/lsp-futhark.el
+++ b/clients/lsp-futhark.el
@@ -25,8 +25,6 @@
 
 (require 'lsp-mode)
 
-(add-to-list 'lsp-language-id-configuration '(futhark-mode . "futhark"))
-
 (lsp-register-client
  (make-lsp-client :new-connection (lsp-stdio-connection '("futhark" "lsp"))
                   :activation-fn (lsp-activate-on "futhark")
--- a/clients/lsp-futhark.el
+++ b/clients/lsp-futhark.el
@@ -1,0 +1,38 @@
+;;; lsp-futhark.el --- lsp-mode futhark integration    -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2024 lsp-mode maintainers
+
+;; Keywords: languages
+
+;; This program is free software; you can redistribute it and/or modify
+;; it under the terms of the GNU General Public License as published by
+;; the Free Software Foundation, either version 3 of the License, or
+;; (at your option) any later version.
+
+;; This program is distributed in the hope that it will be useful,
+;; but WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+;; GNU General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


clients/lsp-futhark.el
```diff
@@ -0,0 +1,38 @@
...
+
+;;; Commentary:
+
+;; Client for the futhark language server.
+
+;;; Code:
+
+(require 'lsp-mode)
+
+(add-to-list 'lsp-language-id-configuration '(futhark-mode . "futhark"))
```
jcs090218, line 28: I don't think you need this anymore since you have already edit the `lsp-mode.el` file.

SKyletoft: Yeah, that makes sense.

SKyletoft: I added the require lsp-mode as that's what was in the config I was looking at (nix)

jcs090218:
You can use this PR as a reference, see https://github.com/emacs-lsp/lsp-mode/pull/4447/files.

SKyletoft:
I think that's it?

There is no configuration to support and the server is a part of the compiler so that seems out of scope.

jcs090218:
Can you add the `defgroup` and link it to the server URL?

I recommend adding that information. It's better to have some documentation rather than no documentation at all.

SKyletoft:
> Can you add the `defgroup` and link it to the server URL?
> 
> I recommend adding that information. It's better to have some documentation rather than no documentation at all.

Do I link the specific folder or just the repo? It's a part of the compiler and the lsp has no readme, just a blog annoucement

SKyletoft:
Added with a link to the lsp. And sorry for being a bother.

jcs090218:
> Added with a link to the lsp. And sorry for being a bother.

No worries. Thant make sense! :)

Can you change the `lsp-clients.json` and add the installation instruction. 🤔 See other settings for references.

docs/lsp-clients.json
```diff
@@ -334,6 +334,14 @@
     "installation": "pip install fortls",
     "debugger": "Yes"
   },
+  {
+    "name": "futhark",
+    "full-name": "Futhark",
+    "server-name": "futhark-lsp",
+    "server-url": "https://github.com/diku-dk/futhark"
+    "installation": "A part of the compiler since 0.21.9",
+    "debugger": "N/A"
```
jcs090218, line 343: Maybe change it to not available just for consistency.

SKyletoft: Both N/A and Not Available are in there

SKyletoft: Anyway, fixed.

But the file is already very inconsistent

jcs090218 (approved):

docs/lsp-clients.json
jcs090218: Yeah, you are right. I was using a mobile device, so I couldn't see the whole file.

Sorry, you have to deal with those legacy changes. I prefer some consistency rather than no consistency. 🤔

jcs090218:
Thank you for your hard work on this! 😋

PR #4559, opened 2024-09-23
Provide a unified (lsp-interface INTERFACE ...) pcase form

### LINKED ISSUE #4430: lsp-interface generates a lot of noise in pcase documentation
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? &mdash; No, I don't believe so.

### TL;DR

I'd like to consider if there's a way lsp-mode could be (or could evolve to be) more conscientious with its use of `pcase-defmacro`.

### Context

I recently opened the function documentation for `pcase` in my Emacs (to read about the `map` pattern). I noticed that, interspersed with the standard documentation, there was a lot of &ldquo;noise&rdquo; like:

> -- (WatchKind &rest PROPERTY-BINDINGS)
>
> Not documented.
>
> -- (CodeAction &rest PROPERTY-BINDINGS)
>
> Not documented.
>
> -- (JSONError &rest PROPERTY-BINDINGS)
>
> Not documented.
>
> &hellip;

To quantify, the function documentation for `pcase' in my Emacs is just under 1300 lines long, of which about 1200 are these auto-generated patterns.

I didn't know where these patterns came from at first, but a quick Google for &ldquo;emacs WatchKind&rdquo; lead me to [lsp-protocol.el][lsp-protocol:WatchKind], and from there I found [the `pcase-macro` in `lsp-interface`][lsp-protocol:lsp-interface] (added in #2481). `pcase-defmacro` includes each new pattern into `pcase`'s function documentation for discoverability.

[lsp-protocol:lsp-interface]: https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-protocol.el#L132
[lsp-protocol:WatchKind]: https://github.com/emacs-lsp/lsp-mode/blob/master/lsp-protocol.el#L782

As a user, this is not a great experience. It's important to me that the `pcase` function documentation be readable, because it's the primary in-editor way that I can discover what patterns are available (including via extensions). I do appreciate that these _are_ technically patterns that are available, but I get the sense they weren't really intended for general consumption, and anyway their sheer number overwhelms any usefulness in documenting them like this.

Also, since these patterns aren't namespaced, there's (a) no indication that they belong to this project, and (b) no guarantee they don't collide with patterns defined by a different package.

### My Ideal Solution

Forgetting for a moment that there is already code using the existing patterns, I would love if `lsp-interface` followed the approach used by `cl-struct` and defined a _single_ `pcase` pattern like `(lsp-interface INTERFACE &rest PROPERTY-BINDINGS)`. This `pcase-defmacro` could include a proper docstring&mdash;e.g., describing its use and the forms allowed as property bindings. It would also address the namespacing concern. And most importantly (for me), it would be a single entry in the `pcase` function documentation rather than hundreds.

I imagine that the current generated `pcase-defmacro` in `lsp-interface` could be repurposed as a generated `lsp-interface--pcase-pattern-INTERFACE` function, to which the `lsp-interface` pattern could delegate based on its *interface* parameter. (This would happen during macro compilation, so there wouldn't be a runtime cost for the indirection.)

So for example, code like

```lisp
(pcase expr
  ((JSONError :message :code) ...))
```

would instead be written

```lisp
(pcase expr
  ((lsp-interface JSONError :message :code) ...))
```

and expand to the exact same base `pcase` patterns.

### Something Practical?

Returning to the real world, I appreciate that my suggestion would be a breaking change. I suppose the first question is: Does my suggestion seem palatable? My primary goal is to keep the `pcase` function documentation readable, so I'm not married to it in particular&mdash;I'm just not sure there's a solution other than to reduce the number of patterns defined. (For example, I think it would be a non-starter to simply include a docstring in the existing `pcase-defmacro`, since that would add hundreds _more_ lines of repeated documentation to `pcase`'s documentation.)

Also, as a newcomer to lsp-mode's development, _are_ these interface patterns intended to be used outside of lsp-mode itself? I did a quick GitHub code search and only found a couple uses in lsp-mode itself. I also couldn't find the feature mentioned in the changelog or documentation. If this isn't a supported feature, maybe a breaking change wouldn't be too painful?

Alternatively, I imagine a new `lsp-interface` pattern could be added along side the existing patterns, which could then be deprecated. For example, the existing `pcase-defmacro` body could be turned into the generated `lsp-interface--pcase-pattern-INTERFACE` function, and the `pcase-defmacro` repurposed to return a new `lsp-interface` pattern.

Thanks for your attention and feedback. Whatever the path forward, I'll be happy to contribute&mdash;I don't mean this to be a rant or demand! I'll probably start poking at the deprecation approach when I get some free time.

chrisbouchard:
This pull request provides a new unified pcase form `(lsp-interface INTERFACE ...)` to replace the old per-interface `(INTERFACE ...)` forms&mdash;the latter have been removed. **This is a breaking change.**

I've turned the old `pcase-defmacro`s generated by `lsp-interface` into helper functions, which the new pcase form delegates to.

This change addresses a few issues, which are detailed in #4430. In short:

* The existing forms aren't namespaced.
* The lsp-mode package adds hundreds of forms, which each add several lines to pcase's generated docstring, adding up to over 1000 lines.
* Starting in Emacs 31, the number of forms added by lsp-mode causes a noticeable slowdown when loading the interactive help for pcase.

Fixes #4430

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -14,7 +14,9 @@
   * Change ~ruff-lsp~ to ~ruff~ for python lsp client. All ~ruff-lsp~ customizable variable change to ~ruff~. Lsp server command now is ~["ruff" "server"]~ instead of ~["ruff-lsp"]~.
   * Add futhark support
   * Optimize overlay creation by checking window visibility first
-
+  * Replace the per-interface ~(INTERFACE ...)~ pcase forms with a single,
+    unified ~(lsp-interface INTERFACE ...)~ form. The per-interface forms are no
+    longer generated. *This is a breaking change.* (See #4430.)
 
 ** 9.0.0
   * Add language server config for QML (Qt Modeling Language) using qmlls.
--- a/lsp-protocol.el
+++ b/lsp-protocol.el
@@ -225,6 +225,12 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
                                              output-bindings)
                                        (setf current-list (cddr current-list))))))
                                  output-bindings))))
+                    ;; These per-interface `pcase' forms are deprecated. Prefer
+                    ;; the new (lsp-interface INTERFACE ...) form defined below.
+                    ;; (See emacs-lsp/lsp-mode#4430.)
+                    ;;
+                    ;; TODO: Remove this `pcase-defmacro' in the next major
+                    ;; version.
                     `(pcase-defmacro ,interface (&rest property-bindings)
                        `(lsp-interface ,',interface ,@property-bindings))
                     (-mapcat (-lambda ((label . name))
--- a/lsp-protocol.el
+++ b/lsp-protocol.el
@@ -255,12 +255,16 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
          (cl-list* 'progn))))
 
 (pcase-defmacro lsp-interface (interface &rest property-bindings)
-  "If EXPVAL is an instance of the LSP interface INTERFACE, destructure its
-properties.
+  "If EXPVAL is an instance of INTERFACE, destructure it by matching its
+properties. EXPVAL should be a plist or hash table depending on the variable
+`lsp-use-plists'.
 
-Each :PROPERTY key may be followed by an optional PATTERN, which is a `pcase'
-pattern to apply to the property value. Otherwise, PROPERTY is bound to the
-property value.
+INTERFACE should be an LSP interface defined with `lsp-interface'. This form
+will not match if any of INTERFACE's required fields are missing in EXPVAL.
+
+Each :PROPERTY keyword matches a field in EXPVAL. The keyword may be followed by
+an optional PATTERN, which is a `pcase' pattern to apply to the field's value.
+Otherwise, PROPERTY is let-bound to the field's value.
 
 \(fn INTERFACE [:PROPERTY [PATTERN]]...)"
   (cl-check-type interface symbol)
--- a/lsp-protocol.el
+++ b/lsp-protocol.el
@@ -129,7 +129,7 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
                                               $$result))
                                    (-partition 2 plist))
                              $$result)))
-                    `(pcase-defmacro ,interface (&rest property-bindings)
+                    `(cl-defun ,(intern (format "lsp--pcase-macroexpander-%s" interface)) (&rest property-bindings)
                        ,(if lsp-use-plists
                             ``(and
                                (pred listp)
@@ -225,6 +225,8 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
                                              output-bindings)
                                        (setf current-list (cddr current-list))))))
                                  output-bindings))))
+                    `(pcase-defmacro ,interface (&rest property-bindings)
+                       `(lsp-interface ,',interface ,@property-bindings))
                     (-mapcat (-lambda ((label . name))
                                (list
                                 `(defun ,(intern (format "lsp:%s-%s"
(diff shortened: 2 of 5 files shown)


kiennq:
> * Starting in Emacs 1.31, the number of forms added by lsp-mode causes a noticeable slowdown when loading the interactive help for pcase.

Is that Emacs 31.1 ? :D

lsp-protocol.el
```diff
@@ -225,6 +225,8 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
                                              output-bindings)
                                        (setf current-list (cddr current-list))))))
                                  output-bindings))))
+                    `(pcase-defmacro ,interface (&rest property-bindings)
```
kiennq, line 228: Shouldn't this still add to the cases of `pcase` and cause the slowness issue?

dgutov:
> Starting in Emacs 1.31, the number of forms added by lsp-mode causes a noticeable slowdown when loading the interactive help for pcase

Could someone also report that slowdown to Emacs?

lsp-protocol.el
chrisbouchard: There's some ongoing discussion in #4430 about whether it's ok to drop the per-interface pcase right now, or whether they're part of lsp-mode's public interface and need to be deprecated for a bit first. 

If it is safe to drop these, I'll add another commit in this branch to remove this `pcase-defmacro` once I've rewritten all the `pcase`s in lsp-mode that use it.

Either way, I should add a comment to summarize all this in the code.

chrisbouchard:
> > * Starting in Emacs 1.31, the number of forms added by lsp-mode causes a noticeable slowdown when loading the interactive help for pcase.
> 
> Is that Emacs 31.1 ? :D

Whoops, I think just 31. Not sure where the “1.” snuck in, but I removed it. Thanks 👍

> > Starting in Emacs 1.31, the number of forms added by lsp-mode causes a noticeable slowdown when loading the interactive help for pcase
> 
> Could someone also report that slowdown to Emacs?

Yes, and that someone could be you! I'm not actually using Emacs 31 yet, so I can't speak to that issue personally. I'm mostly interested in cleaning up pcase's docstring.

That said, depending what new feature in Emacs 31 is triggering the slowdown, it *might* not actually be a bug. E.g., pcase's dynamic docstring function might have some new *O*(n²) feature that reasonably expects to handle a couple dozen pcase forms, not several hundred.

lsp-protocol.el
```diff
@@ -246,6 +254,25 @@ Allowed params: %s" interface (reverse (-map #'cl-first params)))
          (apply #'append)
          (cl-list* 'progn))))
 
+(pcase-defmacro lsp-interface (interface &rest property-bindings)
```
kiennq, line 257: For `lsp-interface` (the one in `defmacro`), can you add the declaration form for it? Something like `(declare (indent 1))`.
That would make the indent for `lsp-interface`, both outside and inside `pcase` looks much better

lsp-protocol.el
chrisbouchard: Looks like the consensus is to remove the per-interface forms and let it be a breaking change. In the interest of not breaking tests, I'll fix the existing pcases and then remove this `pcase-macrodef`. :+1:

lsp-protocol.el
chrisbouchard: I do agree that would look nicer. Do you know if there a bunch of existing `lsp-interface`s that would need to be re-indented, besides ones we'd already be touching for the `pcase` change? (Just trying to avoid blowing up the scope of this PR too much.)

chrisbouchard: Yuck. The problem with adding `(declare (indent 1))` or similar is that `lsp-interface` the macro actually takes multiple interface definitions at once. So, e.g., it winds up indented like

```lisp
(lsp-interface
    (eslint:StatusParams  (:state) nil)
  (eslint:OpenESLintDocParams (:url) nil)
  (eslint:ConfirmExecutionParams (:scope :file :libraryPath) nil))
```

Unless there's a strong opinion, I'd like to leave this alone for now&mdash;it's ugly, but I think the macro is used more frequently than the pcase form.

chrisbouchard: Also, the `lsp-make-INTERFACE` functions kind of have the same problem, so maybe there could be a separate PR focused on indentation.

lsp-protocol.el
chrisbouchard: These are gone now in 849cff06bc37bf4d7e80a293cd2ac0d83386b34b.

CHANGELOG.org
```diff
@@ -14,7 +14,9 @@
   * Change ~ruff-lsp~ to ~ruff~ for python lsp client. All ~ruff-lsp~ customizable variable change to ~ruff~. Lsp server command now is ~["ruff" "server"]~ instead of ~["ruff-lsp"]~.
   * Add futhark support
   * Optimize overlay creation by checking window visibility first
-
+  * Replace the per-interface ~(INTERFACE ...)~ pcase forms with a single,
+    unified ~(lsp-interface INTERFACE ...)~ form. The per-interface forms are no
+    longer generated. *This is a breaking change.* (See #4430.)
```
chrisbouchard, line 19: I'm definitely open to suggestions on the CHANGELOG entry.

chrisbouchard:
There are two test suites failing, but they're for Emacs snapshots. Something's making the byte compiler unhappy. It looks like these are also unhappy on master.

CHANGELOG.org
chrisbouchard: Also, does the unreleased version stay 9.0.1, or should that go to 10.0.0 now?

kiennq (approved):

CHANGELOG.org
jcs090218: 9.0.1 is good.

PR #4564, opened 2024-09-27
lsp-asm: if executable is found there's no need to install it

caiquefigueiredo:
This PR fixes the problem in which `lsp-asm` tries to install `asm-lsp` even when it's already installed through other means and available on path.

### DIFF
--- a/clients/lsp-asm.el
+++ b/clients/lsp-asm.el
@@ -81,7 +81,8 @@ Will update if UPDATE? is t."
  (make-lsp-client
   :new-connection (lsp-stdio-connection
                    #'lsp-asm--server-command
-                   (lambda () (f-exists? lsp-asm-store-path)))
+                   (lambda () (or (executable-find "asm-lsp")
+                                  (f-exists? lsp-asm-store-path))))
   :major-modes lsp-asm-active-modes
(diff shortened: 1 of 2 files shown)


caiquefigueiredo:
@jcs090218 would you be able to review this one? 🙏🏼

clients/lsp-asm.el
```diff
@@ -81,7 +81,8 @@ Will update if UPDATE? is t."
  (make-lsp-client
   :new-connection (lsp-stdio-connection
                    #'lsp-asm--server-command
-                   (lambda () (f-exists? lsp-asm-store-path)))
+                   (lambda () (or (executable-find "asm-lsp"))
+                             (f-exists? lsp-asm-store-path)))
```
jcs090218, line 85: Can you fix the indentation here? Thanks!

caiquefigueiredo: Damn, thanks for pointing this out. Apparently I didn't have parinfer on and the `or` clause was missing the `f-exists?`. My bad. Fixed @ latest

jcs090218 (approved):

jcs090218:
LGTM! Thank you!

PR #4567, opened 2024-09-30
Add missing configuration options for the Ada Language Server.

brownts:
Additionally, prefer "initialization-options" over "initialized-fn" as the Ada Language Server prefers to know the project file as part of initialization, and will have to create a default one if it's not specified as part of these options.  If a project file is not specified as part of the initialization options, the server might also display a diagnostic about no project being specified until a workspace/didChangeConfiguration is received specifying the project file.

Add support for Ada run-time library folders via the new user option `lsp-ada-library-folders`, defaulting to detecting the run-time directory with it's unique "adainclude" folder.

### DIFF
--- a/clients/lsp-ada.el
+++ b/clients/lsp-ada.el
@@ -290,6 +290,7 @@ must return a list of Ada library folders, or nil if none."
 (defvar lsp-ada--als-downloaded-executable
   (f-join lsp-server-install-dir
           "ada-ls"
+          "integration" "vscode" "ada"
           (symbol-name (lsp-resolve-value lsp--system-arch))
           (pcase system-type
             ('gnu/linux  "linux")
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

clients/lsp-ada.el
```diff
@@ -57,25 +60,191 @@ environment."
 ;;;###autoload(put 'lsp-ada-project-file 'safe-local-variable 'stringp)
 
 (lsp-defcustom lsp-ada-option-charset "UTF-8"
-  "The charset to use by the Ada Language server. Defaults to \\='UTF-8\\='."
+  "The charset to use by the Ada Language server.  Defaults to \\='UTF-8\\='."
   :type 'string
   :group 'lsp-ada
   :package-version '(lsp-mode . "6.2")
   :lsp-path "ada.defaultCharset")
+;;;###autoload(put 'lsp-ada-option-charset 'safe-local-variable 'stringp)
```
jcs090218, line 68: Do we need to autoload all these custom variables? 🤔

brownts: The reason I added them is that any of those could potentially be specified in a .dir-locals.el.  I find it annoying to have to manually approve the "potentially unsafe" settings in my .dir-locals.el.  If you think this is too much, I can cut these back to the most frequently used options.  I could of course add safety settings in my init.el, but since other people use this mode as well, it would require others to do the same in their init.el, and I was going for easy of use.

jcs090218: Let's cut back for now since it's unusual for only this file to do that. 🤔 We can open a new issue, discuss this, and see what others think.

brownts:
I've updated the commit to remove all of the additional autoloads.  I've also added a second commit which fixes the Ada Language Server installation, as the name, archive and path within the archive all changed recently.

brownts:
@jcs090218, did you have any further concerns?

jcs090218 (approved):

jcs090218:
Merged! Thank you for taking care of this! 😄

PR #4568, opened 2024-09-30
Fix typo in lsp-golangci-lint custom group

wyuenho:
This mistake current prevent the whole custom group from being loaded


jcs090218 (approved):

jcs090218:
Thanks!

PR #4575, opened 2024-10-02
lsp-graphql: add new defcustom `lsp-graphql-target-file-extensions`

suzuki:
## About

- Add new `lsp-graphql-target-file-extensions` customize variable
   - Add `.graphqls` extension to it

## Why

I want to use `lsp-graphql` on the file with `.graphqls` ext.

There was a way to write directly to a list in a function, but I thought it was better to isolate it as a customised variable.

### DIFF
--- a/clients/lsp-graphql.el
+++ b/clients/lsp-graphql.el
@@ -43,13 +43,18 @@
   :risky t
   :group 'lsp-graphql)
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

PR #4576, opened 2024-10-03
Remove trailing whitespace from `lsp-describe-thing-at-point'

### LINKED ISSUE #4572: lsp-describe-thing-at-point should trim whitespace from all lines

### Bug description

Using lsp-mode with clangd, and pointing anything at `lsp-describe-thing-at-point' I see trailing whitespace in the resulting buffer.
I notice because some of the faces being used use underline which extends past the end of the line.

Looking at the definition of `lsp-describe-thing-at-point' we see the actual buffer is inserted with:

``(insert (string-trim-right (lsp--render-on-hover-content contents t)))))``

which strips whitespace out of the last line only. Changing this to use split/trim/mapconcat works as expected, but I'm unsure looking at the code if the whitespace shouldn't be stripped by lsp--render-on-hover-content directly. If not, I would provide a PR.

### Steps to reproduce

Try to open/describe any C variable by using `lsp-describe-thing-at-point' with clangd.
A plain C file with:

```
float PI = 3.1415926f;
```

shows:

2024-10-02T151851

### Expected behavior

The description shouldn't contain trailing whitespace.

lsp-clangd

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

wavexx:
When calling `lsp-describe-thing-at-point' trailing whitespace can leak into the help buffer. Depending on the face settings, this whitespace can be visibile.

Instead of removing trailing whitespace at the end of the buffer, do so on a line-by-line basis.

Fixes #4572

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5405,7 +5405,10 @@ If EXCLUDE-DECLARATION is non-nil, request the server to include declarations."
             (delay-mode-hooks
               (lsp-help-mode)
               (with-help-window lsp-help-buf-name
-                (insert (string-trim-right (lsp--render-on-hover-content contents t)))))
+                (insert
+		 (mapconcat 'string-trim-right
+			    (split-string (lsp--render-on-hover-content contents t) "\n")
+			    "\n"))))
             (run-mode-hooks)))
       (lsp--info "No content at point."))))
 


jcs090218 (approved):
It sort of makes sense, but not sure if there would be any other side effects. 🤔

wavexx:
I was wondering this myself in the linked issue. As far as I can see, describe-at-point is not used elsewhere.

I'm fully open in not knowing where this trailing whitespace comes from (clangd, or some internal formatting?). For testing, if you on some whitespace-mode in lsp-help, there's some substantial but semi-random amount...

jcs090218:
There is an error:

```
lsp-mode.el:5410:30: Warning: the function ‘string-split’ is not known to be
```

🤔

wavexx:
Didn't realize string-split is a "new" alias for split-string... fixed.

jcs090218 (approved):

jcs090218:
Thank you!

PR #4577, opened 2024-10-03
Fix lsp-completion-default-behaviour const value

wyuenho:
The `:tag` keyword symbol must be the first keyword symbol in simple types, otherwise we are saying the value of `:insert` is `:tag`


jcs090218 (approved):

jcs090218:
Thank you!

PR #4588, opened 2024-10-19
Refactoring all-the-icons functions

wyuenho:
`lsp-modeline--code-actions-icon` isn't guarded by the proper checks for `all-the-icons-octicon` function, which will cause `all-the-icons` to load in the terminal, even though it is not supported. I've also taken the opportunity to refactor the 2 functions that use all-the-icons to use the new `lsp-icons-all-the-icons-icon` function.


wyuenho:
@jcs090218 Do you have any feedbacks for this PR?

jcs090218 (approved):

jcs090218:
LGTM. :)

wyuenho:
Can we get this merged?

Issue #3279, opened 2021-12-22
Support for related locations in diagnostics?


cpitclaudel:
Hi there,

LSPdiagnostics have a relatedInformation field that LSP-mode does not seem to be using yet.  How hard would it be to display these related locations?

This was briefly discussed in https://github.com/emacs-lsp/lsp-ui/issues/165, but since LSP discards the information it's not something that LSP-ui can do on it's own, I think.

I'm guessing there are two ways to go about this:

1. Use flycheck's `group` feature (this is how it was intended to be used, but LSP-mode uses it to store source information).  The code looks something like this:

```elisp
(defun lsp-diagnostics--flycheck-start (checker callback)
  "Start an LSP syntax check with CHECKER.

CALLBACK is the status callback passed by Flycheck."

  (remove-hook 'lsp-on-idle-hook #'lsp-diagnostics--flycheck-buffer t)

  (->> (lsp--get-buffer-diagnostics)
       (-mapcat
        (-lambda ((&Diagnostic :message :severity? :tags? :code? :source? :related-information?
                               :range (&Range :start (&Position :line      start-line
                                                                :character start-character)
                                              :end   (&Position :line      end-line
                                                                :character end-character))))
          (let ((group (gensym)))
            (cons (flycheck-error-new
                   :buffer (current-buffer)
                   :checker checker
                   :filename buffer-file-name
                   :message message
                   :level (lsp-diagnostics--flycheck-calculate-level severity? tags?)
                   :id code?
                   :group group
                   :line (lsp-translate-line (1+ start-line))
                   :column (1+ (lsp-translate-column start-character))
                   :end-line (lsp-translate-line (1+ end-line))
                   :end-column (1+ (lsp-translate-column end-character)))
                  (-mapcat
                   (-lambda ((&DiagnosticRelatedInformation
                              :message
                              :location
                              (&Location :range (&Range :start (&Position :line      start-line
                                                                          :character start-character)
                                                        :end   (&Position :line      end-line
                                                                          :character end-character))
                                         :uri)))
                     (and (equal (-> uri lsp--uri-to-path lsp--fix-path-casing)
                                 (-> buffer-file-name lsp--fix-path-casing))
                          `(,(flycheck-error-new
                              :buffer (current-buffer)
                              :checker checker
                              :filename buffer-file-name
                              :message message
                              :level (lsp-diagnostics--flycheck-calculate-level (1+ severity?) tags?)
                              :id code?
                              :group group
[...]

PR #4589, opened 2024-10-19
Add support for LSP diagnostic "additional information".

brownts:
This change provides support for additional information that can be provided with an LSP diagnostic, for both Flycheck and Flymake interfaces.  The main change treats each piece of "additional information" as a separate diagnostic with the same severity level as the main diagnostic.  Flycheck supports the concept of a group, so for Flycheck diagnostics the main diagnostic and all of it's "additional information" diagnostics are placed in the same Flycheck group.

"Additional information" diagnostics may refer to a file which is not the current buffer.  Flycheck handles this as part of it's normal diagnostic interface, however Flymake makes a distinction between diagnostics related to the current buffer and those external to it (i.e., foreign diagnostics).  For Flymake, the foreign diagnostics are managed in the `flymake-list-only-diagnostics` variable.  Flymake treats these as interim diagnostics until the file has been opened. Therefore, diagnostics placed in `flymake-list-only-diagnostics` are removed when that buffer reports its own diagnostics.  Additionally, with Flymake, foreign diagnostics are only visible when reporting diagnostics for the project (`flymake-show-project-diagnostics`), not when reporting diagnostics for the current
buffer (`flymake-show-buffer-diagnostics`).  Flycheck on the other hand, does not make this distinction and will report both types of diagnostics together (`flycheck-list-errors`).

Additional changes that drive conformity between Flycheck and Flymake were also applied.  When an LSP diagnostic provides a source, this is supplied to Flycheck as the "checker", otherwise the default "lsp" checker is used.  This is useful to properly identify the LSP backends generating the diagnostics when it is available.  Flymake doesn't provide a way to specify the checker (or a code), so the diagnostic message is appended with this information when provided.


brownts:
@jcs090218, any concerns with merging this?

brownts:
Ping @jcs090218, @yyoncho.

This PR has been sitting here for a while and I'm not sure what needs to be done in order to have it merged.  There are Language Servers which provide "additional information" which could benefit from this.  Additionally, it appears this was asked for in the past in #3279.

Any help in moving this along is appreciated.  Thanks.

PR #4593, opened 2024-10-24
Autocompletion options for Nix (nixd)

breitnw:
This PR adds more comprehensive options for the nixd language server, including those necessary for formatting and completion. As mentioned in #4592, these options were previously absent.

The options added include: 
- `lsp-nix-nixd-nixpkgs-expr`, which provides completion for nixpkgs
- `lsp-nix-nixd-nixos-options-expr`, which provides completion for NixOS options
- `lsp-nix-nixd-home-manager-options-expr`, which provides completion for Home Manager options
- `lsp-nix-nixd-formatting-command`, which sets the formatter used by nixd

An example configuration in Doom: 

```elisp
(after! lsp-mode
  (setq lsp-nix-nixd-formatting-command [ "nixfmt" ]
        lsp-nix-nixd-nixpkgs-expr "import <nixpkgs> { }"
        lsp-nix-nixd-nixos-options-expr "(builtins.getFlake \"/home/breitnw/Documents/code/nixos\").nixosConfigurations.mnd.options"
        lsp-nix-nixd-home-manager-options-expr "(builtins.getFlake \"/home/breitnw/Documents/code/nixos\").homeConfigurations.\"breitnw@mnd\".options"))

(add-hook! 'nix-mode-hook
           (setq company-idle-delay 0.1))
```

PR #4594, opened 2024-10-25
fix: nixd option groups and documentation page

breitnw:
My last PR (#4593) seemed to cause some documentation issues, since a few of the options were assigned to `'lsp-nix-nil` instead of `'lsp-nix-nixd`. Also, lsp-clients.json listed both nixd and rnix under a "nix" page, when they should be split into pages named "nix-nixd" and "nix-rnix", respectively.

### DIFF
--- a/clients/lsp-nix.el
+++ b/clients/lsp-nix.el
@@ -83,7 +83,7 @@
   Example:
   \"(builtins.getFlake \"/home/lyc/flakes\").nixosConfigurations.adrastea.options\""
   :type 'string
-  :group 'lsp-nix-nil
+  :group 'lsp-nix-nixd
   :lsp-path "nixd.options.nixos.expr"
   :package-version '(lsp-mode . "9.0.1"))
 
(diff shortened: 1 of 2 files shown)


breitnw:
I do think that might make things clearer, I felt there were quite a few moving parts in that file (that's why I got the groups confused, lol 😵‍💫) Thank you for your review!

jcs090218:
There are compile warnings in this PR:

```
In toplevel form:
lsp-nix.el:79:2: Warning: custom-declare-variable `lsp-nix-nixd-nixos-options-expr' docstring wider than 80 characters
lsp-nix.el:90:2: Warning: custom-declare-variable `lsp-nix-nixd-home-manager-options-expr' docstring wider than 80 characters
```

psibi:
Ah, I didn't notice it and can't still find the logs in github actions. CC: @breitnw Can you open a new PR fixing the docstring warnings ?

I wonder if we should fail the CI if those check. I think I have something like that enabled for one of my projects, but I think for a project this big we would have a bunch of warnings like that which needs to be fixed.

jcs090218:
> I wonder if we should fail the CI if those check. I think I have something like that enabled for one of my projects, but I think for a project this big we would have a bunch of warnings like that which needs to be fixed.

Agreed! We can set `byte-compile-error-on-warn` to `t` in the Eask-file. :)

PR #4596, opened 2024-10-25
fix: reduce nix docstring length

breitnw:
As noted in #4594, there were compile warnings in the last PR: 

```
In toplevel form:
lsp-nix.el:79:2: Warning: custom-declare-variable `lsp-nix-nixd-nixos-options-expr' docstring wider than 80 characters
lsp-nix.el:90:2: Warning: custom-declare-variable `lsp-nix-nixd-home-manager-options-expr' docstring wider than 80 characters
```

This PR should fix these warnings.

### DIFF
--- a/clients/lsp-nix.el
+++ b/clients/lsp-nix.el
@@ -81,7 +81,7 @@
   search path (`<nixpkgs>`) will be used.
 
   Example:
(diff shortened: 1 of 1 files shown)


jcs090218:
Thank you!

PR #4602, opened 2024-10-30
lsp-completion-passthrough-try-completion: make it more basic

kiennq:
Make lsp-completion-passthrough-try-completion more basic by just return the passed string and point. It will make the behavior closer to VsCode where the try completion will always be success.
Also respect the `completion-ignore-case` value now when showing the completion candidates.

@dgutov

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -284,7 +284,8 @@ The MARKERS and PREFIX value will be attached to each candidate."
                                                  (setq fuz-queries
                                                        (plist-put fuz-queries start-point s))
                                                  s)))
-                                (label-len (length label)))
+                                (label-len (length label))
+                                (case-fold-search completion-ignore-case))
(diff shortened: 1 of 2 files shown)


dgutov:
Just to note: both changes look reasonable, even though neither is my preference.

The second one moves the default behavior farther from VS Code (which matches case-insensitively right away). But it'd make it user-configurable, which is a plus.

dgutov:
Further: it would fix https://github.com/company-mode/company-mode/issues/1412, although that's done by removing a feature.

Also, https://github.com/company-mode/company-mode/pull/1499 should be relevant (without it, the default TAB binding becomes useless when lsp-mode is enabled).

kiennq:
Friendly ping @jcs090218

jcs090218 (approved):
Sounds good to me! :)

PR #4608, opened 2024-11-11
Touch up signature faces

wyuenho:
1. We should not use `font-lock-comment-face` as the signature posframe border color. There are actually proper faces that frame borders use for this purpose, so let's just use them so we can have a child frame that has borders that look like every other child frame. These are also faces `posframe-show` will set the border color on.
2. When propertizing the parameter hints highlight in the signature string, instead of using the `eldoc-highlight-function-argument` face, let's use a new LSP-specific face that by default inherits from `eldoc-highlight-function-argument` so users can theme it.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5855,6 +5855,11 @@ It will show up only if current point has signature help."
                                     (list lsp-signature-doc-lines))))
   (lsp-signature-activate))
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Make sense! Thank you!

PR #4620, opened 2024-11-24
Expose lsp-bash-allowed-shell and improve warning

### LINKED ISSUE #4201: Make `lsp-bash-check-sh-shell` configurable
`lsp-bash-check-sh-shell` prevents the Bash language server from being turned on, for example, in `zsh` files. However, it also does so for other interpreters that could use the Bash language server. One specific example is `nix-script-bash`.

It would be great if the list of interpreters suitable for the Bash language server were configurable, and not hard coded as `'(sh bash)`.

dcarley:
Fixes #4201

**lsp-bash: Remove major mode duplication**

These are already filtered by `:major-modes` so there's no need to
filter them again in the `:activation-fn`.

`:major-modes` was originally removed in:

- b77aecf

But then reintroduced in:

- 771342f

**lsp-bash: Expose lsp-bash-allowed-shells**

Make the `sh-shell` values that are filtered by `:activation-fn`
configurable without needing to override `lsp-bash-check-sh-shell`

This allows people to allow values like `bats` and also makes it
slightly clearer why those values aren't supported by default when the
major mode is, which took me a long time to debug.

**lsp-mode: Improve warning about :activation-fn**

To make it easier to debug situations where the `:major-mode` is
supported but `:activation-fn` prevents the buffer from being supported
for other reasons, such as `lsp-bash` rejecting an `sh-shell` value of
`bats`.


jcs090218 (approved):

jcs090218:
Sounds reasonable. Thank you!

Issue #4581, opened 2024-10-08
Support for inlineCompletionProvider


Konubinix:
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)?

Describe your feature here.

When getting completion, I try to use only the built-in capfs mechanism. The fact that lsp-mode integrates with it is great. Unfortunately, copilot.el does not and I have capfs and the ad-hoc copilot.el stuff so far.

I just realized that under the hood, copilot.el uses an lsp server : copilot-node-server, so I tried creating the lsp-copilot.el client.

But this server provides completion using the inlineCompletionProvider capability, not the completionProvider one.

I tried using https://github.com/kassick/lsp-inline-completions there are a few glitches, but it basically works. But it still does not use the capfs. So I'm back to square one.

So, I'm wondering if we could deal with completion using inlineCompletionProvider in lsp-mode, so that simply M-TAB would work seamlessly.

Is there already some discussion on the topic?

kiennq:
`inlineCompletions` is official in `3.18` lsp [spec](https://github.com/microsoft/language-server-protocol/blob/gh-pages/_specifications/lsp/3.18/language/inlineCompletion.md), we should support it if possible.
I've also created an issue in this kassick/lsp-inline-completions#1 to see if it's okay to integrate the `lsp-inline-completions` to the current `lsp-mode`.
@yyoncho @jcs090218

kassick:
> so I tried creating the lsp-copilot.el

@Konubinix I did the same here https://github.com/kassick/copilot-lsp ! Let me know if you want to use some of it or contribute -- it was created as a POC at the time, and then I switched to `copilot.el` because I did not have much time back then ...

kassick:
@kiennq as I mentioned in the `lsp-inline-completions` repo, if the maintainers think it makes sense to upstream the code, sure!

There is at least one glitch I'm aware of -- not sure if it is the same one @Konubinix ran into

kassick:
Regarding CAPF for inlineCompletions:

One aspect of inline completions is that *they can span multiple lines*.

With copilot, I did not see much of an issue -- I always get one-liners, not sure if it ever responds with longer text. On gitlab-lsp, though, some prompts resulted in multi-line code snippets (e.g. `# write a unit test for this file` would bring a test function) , and some would even result in changes *in other points of the text*.

[...]

PR #4623, opened 2024-11-25
Feature: inline completions

kassick:
As discussed in https://github.com/emacs-lsp/lsp-mode/issues/4581, this PR adds support for Inline Completions

Using some server with `inlineCompletionProvider` capability (I'm using the [copilot-node-server](https://www.npmjs.com/package/copilot-node-server) with [this implementation](https://github.com/kassick/copilot-lsp/)), the inline completion is triggered automatically if `lsp-inline-completion-enable` is set to `t`. The user can also call `lsp-inline-completion-display` explicitly to have the completion displayed.

Captura de tela de 2024-11-25 10-50-01

Kooha-2024-11-25-10-51-45

~Optionally, the  user can add `lsp-inline-completion-show-keys` to the `lsp-inline-completion-shown-hook` and the number of available completions and the keys used to interact with the tooltips are shown in the minibuffer.~
When the completion overlay is displayed, a message is shown in the minibuffer as part of the UI, so the user knows if there are more then one completion available.

image

When the inline completion tooltip is being displayed, pressing any key that is not mapped in `lsp-inline-completion-active-map` causes the tooltip to be hidden and the key is processed with the current active keymap. This way, if the completion is shown automatically, the user can continue typing to have the completion ignored.

### DIFF
--- a/lsp-inline-completion.el
+++ b/lsp-inline-completion.el
@@ -394,8 +394,10 @@ lsp-inline-completion-mode is active"
 (defvar-local lsp-inline-completion--idle-timer nil
   "The idle timer used by lsp-inline-completion-mode")
 
-(defun lsp-inline-completion--maybe-display ()
-  (unless (--any (funcall it) lsp-inline-completion-inhibit-predicates)
+(defun lsp-inline-completion--maybe-display (buffer)
+  (when (and (buffer-live-p buffer)
+             (eq (current-buffer) buffer)
+             (--none? (funcall it) lsp-inline-completion-inhibit-predicates))
     (setq last-command this-command)
     (setq this-command 'lsp-inline-completion-display)
     (lsp-inline-completion-display 'implicit)))
@@ -404,10 +406,12 @@ lsp-inline-completion-mode is active"
   (when (and lsp-inline-completion-mode lsp--buffer-workspaces)
     (when lsp-inline-completion--idle-timer
       (cancel-timer lsp-inline-completion--idle-timer))
-    (setq lsp-inline-completion--idle-timer
-          (run-with-timer lsp-inline-completion-idle-delay
-                          nil
-                          #'lsp-inline-completion--maybe-display))))
+    (let ((buffer (current-buffer)))
+      (setq lsp-inline-completion--idle-timer
+            (run-with-timer lsp-inline-completion-idle-delay
+                            nil
+                            #'lsp-inline-completion--maybe-display
+                            buffer)))))
 
 (define-minor-mode lsp-inline-completion-mode
   "Mode automatically displaying inline completions."
--- a/lsp-inline-completion.el
+++ b/lsp-inline-completion.el
@@ -66,4 +66,239 @@
     ;; Clean up
     (lsp--spinner-stop)))
 
+
+;;;;;; Default UI -- overlay
+
+(defvar lsp-inline-completion-active-map
+  (let ((map (make-sparse-keymap)))
+    (define-key map (kbd "<tab>") #'lsp-inline-completion-next)
+    (define-key map (kbd "C-n") #'lsp-inline-completion-next)
+    (define-key map (kbd "M-n") #'lsp-inline-completion-next)
+    (define-key map (kbd "C-p") #'lsp-inline-completion-prev)
+    (define-key map (kbd "M-p") #'lsp-inline-completion-prev)
+    (define-key map (kbd "<return>") #'lsp-inline-completion-accept)
+    (define-key map [down-mouse-1] #'ignore)
+    (define-key map [down-mouse-3] #'ignore)
+    (define-key map [up-mouse-1] #'ignore)
+    (define-key map [up-mouse-3] #'ignore)
+    (define-key map [mouse-1] #'lsp-inline-completion-accept)
+    (define-key map [mouse-3] #'lsp-inline-completion-accept)
+    (define-key map (kbd "C-l") #'recenter-top-bottom)
+    (define-key map (kbd "C-g") #'lsp-inline-completion-cancel)
+    (define-key map (kbd "<escape>") #'lsp-inline-completion-cancel)
+    (define-key map (kbd "C-c C-k") #'lsp-inline-completion-cancel)
+    (define-key map [mouse-movement] #'ignore)
+    (define-key map [t] #'lsp-inline-completion-cancel)
+    map)
+  "Keymap active when showing inline code suggestions")
+
+;;;###autoload
+(defface lsp-inline-completion-overlay-face
+  '((t :inherit shadow))
+  "Face for the inline code suggestions overlay.")
+
+;; Local Buffer State
+
+(defvar-local lsp-inline-completion--items nil "The completions provided by the server")
+(defvar-local lsp-inline-completion--current nil "The current suggestion to be displayed")
+(defvar-local lsp-inline-completion--overlay nil "The overlay displaying code suggestions")
+(defvar-local lsp-inline-completion--start-point nil "The point where the completion started")
+
+(defcustom lsp-before-inline-completion-hook nil
+  "Hooks run before starting code suggestions"
+  :type 'hook)
+
+(defcustom lsp-after-inline-completion-hook nil
+  "Hooks executed after asking for code suggestions."
+  :type 'hook)
+
+;; TODO: Add parameters to the hooks!
+(defcustom lsp-inline-completion-accepted-hook nil
+  "Hooks executed after accepting a code suggestion."
+  :type 'hook)
+
+(defcustom lsp-inline-completion-shown-hook nil
+  "Hooks executed after showing a suggestion."
+  :type 'hook)
+
+(defsubst lsp-inline-completion--overlay-visible ()
+  "Return whether the `overlay' is avaiable."
+  (and (overlayp lsp-inline-completion--overlay)
+       (overlay-buffer lsp-inline-completion--overlay)))
+
+(defun lsp-inline-completion--clear-overlay ()
+  "Hide the suggestion overlay"
+  (when (lsp-inline-completion--overlay-visible)
+    (delete-overlay lsp-inline-completion--overlay))
+  (setq lsp-inline-completion--overlay nil))
+
+
+(defun lsp-inline-completion--get-overlay (beg end)
+  "Build the suggestions overlay"
+  (when (overlayp lsp-inline-completion--overlay)
+    (lsp-inline-completion--clear-overlay))
+
+  (setq lsp-inline-completion--overlay (make-overlay beg end nil nil t))
+  (overlay-put lsp-inline-completion--overlay 'keymap lsp-inline-completion-active-map)
+  (overlay-put lsp-inline-completion--overlay 'priority 9000)
+
+  lsp-inline-completion--overlay)
+
+
+;;;###autoload
+(defun lsp-inline-completion-show-keys ()
+  "Shows active keymap hints in the minibuffer"
+
+  (unless (and lsp-inline-completion--items
+               (numberp lsp-inline-completion--current))
+    (error "No completions to show 1"))
+
+  (let ((message-log-max nil))
+    (message (concat "Completion "
+                     (propertize (format "%d" (1+ lsp-inline-completion--current)) 'face 'bold)
+                     "/"
+                     (propertize (format "%d" (length lsp-inline-completion--items)) 'face 'bold)
+
+                     (-when-let (keys (where-is-internal #'lsp-inline-completion-next lsp-inline-completion-active-map))
+                       (concat ". "
+                               (propertize " Next" 'face 'italic)
+                               (format ": [%s]"
+                                       (string-join (--map (propertize (key-description it) 'face 'help-key-binding)
+                                                           keys)
+                                                    "/"))))
+                     (-when-let (keys (where-is-internal #'lsp-inline-completion-accept lsp-inline-completion-active-map))
+                       (concat (propertize " Accept" 'face 'italic)
+                               (format ": [%s]"
+                                       (string-join (--map (propertize (key-description it) 'face 'help-key-binding)
+                                                           keys)
+                                                    "/"))))))))
+
+(defun lsp-inline-completion-show-overlay ()
+  "Makes the suggestion overlay visible"
+  (unless (and lsp-inline-completion--items
+               (numberp lsp-inline-completion--current))
+    (error "No completions to show 2"))
+
+  (lsp-inline-completion--clear-overlay)
+
+  (-let* ((suggestion
+           (elt lsp-inline-completion--items
+                lsp-inline-completion--current))
+          ((&InlineCompletionItem? :insert-text :filter-text? :range? :command?) suggestion)
+          ((&RangeToPoint :start :end) range?)
+          (start-point (or start (point)))
+          (showing-at-eol (save-excursion
+                            (goto-char start-point)
+                            (eolp)))
+          (beg (if showing-at-eol (1- start-point) start-point))
+          (end-point  (or end (1+ beg)))
+          (text (cond
+                 ((lsp-markup-content? insert-text) (lsp:markup-content-value insert-text))
+                 (t insert-text)))
+          (propertizedText (concat
+                            (buffer-substring beg start-point)
+                            (propertize text 'face 'lsp-inline-completion-overlay-face)))
+          (ov (lsp-inline-completion--get-overlay  beg end-point)))
+    (goto-char beg)
+
+    (put-text-property 0 1 'cursor t propertizedText)
+    (overlay-put ov 'display (substring propertizedText 0 1))
+    (overlay-put ov 'after-string (substring propertizedText 1))
+
+    (run-hooks 'lsp-inline-completion-shown-hook)))
+
+(defun lsp-inline-completion-accept ()
+  "Accepts the current suggestion"
+  (interactive)
+  (unless (lsp-inline-completion--overlay-visible)
+    (error "Not showing suggestions"))
+
+  (lsp-inline-completion--clear-overlay)
+  (-let* ((suggestion (elt lsp-inline-completion--items lsp-inline-completion--current))
+          ((&InlineCompletionItem? :insert-text :filter-text? :range? :command?) suggestion)
+          ((kind . text) (cond
+                          ((lsp-markup-content? insert-text)
+                           (cons 'snippet (lsp:markup-content-value insert-text) ))
+                          (t (cons 'text insert-text))))
+          ((start . end) (when range?
+                           (-let (((&RangeToPoint :start :end) range?)) (cons start end))))
+          (text-insert-start (or start lsp-inline-completion--start-point))
+          text-insert-end)
+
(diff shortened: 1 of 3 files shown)


kassick:
A note about company mode and the inline completion overlay: Since the completions are not provided by company, we may end up with the company popup displayed alongside (or over) the inline completion overlay (as [it happened already with copilot.el](https://github.com/copilot-emacs/copilot.el?tab=readme-ov-file#known-issues)).

To workaround this issue, I've added code to cancel company before showing the overlay and then restore it on cancel.

[Kooha-2024-11-26-14-40-17.webm](https://github.com/user-attachments/assets/f93198ba-fbe5-410c-aea3-f03e0c2e95d5)

Switching between the two completion UIs can be achieved by binding `company-manual-begin` and `lsp-inline-completion-display`.

Nonetheless, I've added a variable that can inhibit inline completion mode on certain conditions, such as company being active. For example, I have the following snippet in my user settings:

```elisp
  (defun lsp-inline-completion-inhibit-if-company-active ()
    (and (bound-and-true-p company-mode) (company--active-p)))

  (push 'lsp-inline-completion-inhibit-if-company-active lsp-inline-completion-inhibit-predicates)
```

lsp-protocol.el
```diff
@@ -624,6 +626,8 @@ See `-let' for a description of the destructuring mechanism."
  (CompletionItemKindCapabilities nil (:valueSet))
  (CompletionItemTagSupportCapabilities (:valueSet) nil)
  (CompletionOptions nil (:resolveProvider :triggerCharacters :allCommitCharacters))
+ (InlineCompletionItem (:insertText) (:filterText :range :command))
```
kiennq, line 629: Please sort this

lsp-inline-completion.el
```diff
@@ -0,0 +1,365 @@
...
+(require 'dash)
+(require 'cl-lib)
+
+(declare-function company--active-p "ext:company")
+(declare-function company-cancel "ext:company" (&optional result))
+(declare-function company-manual-begin "ext:company")
+
+(defun lsp-inline-completion--trigger-kind (implicit)
+  (plist-put (lsp--text-document-position-params)
+             :context (ht ("triggerKind"
```
kiennq, line 41: Please create `InlineCompletionContext` type in `lsp-protocol` and we can use `lsp-make-inline-completion-context` to create context for it. That would make it work for both `plist` and `ht` variant of `lsp-mode`

lsp-inline-completion.el
```diff
@@ -0,0 +1,365 @@
...
+                           (if implicit
+                               lsp/inline-completion-trigger-automatic
+                             lsp/inline-completion-trigger-invoked)))))
+
+;;;###autoload
```
kiennq, line 46: I don't think we need to make this function public and have `autoload`. Shouldn't it be used for internal function only?

lsp-inline-completion.el
```diff
@@ -0,0 +1,365 @@
...
+(defun lsp-inline-completion-parse-items (response)
+  "Calls textDocument_inlineCompletion and returns a list of InlineCompletionItem's"
+
+  (-some-->
+      response
+    ;; Kludge to workaround multiple backends responding -- it may
+    ;; come as a list (multiple servers) or as a single ht (single
+    ;; server). Promote it to list and move on
+    (if (ht-p it) (list it) it)
```
kiennq, line 55: The type can be `ht` or `plist` depend on `lsp-use-plists` so it would be better to just check the `lsp-inline-completion-list?` and then merge or extract and merge items.

lsp-inline-completion.el
```diff
@@ -0,0 +1,365 @@
...
+    (define-key map (kbd "C-l") #'recenter-top-bottom)
+    (define-key map (kbd "C-g") #'lsp-inline-completion-cancel)
+    (define-key map (kbd "<escape>") #'lsp-inline-completion-cancel)
+    (define-key map (kbd "C-c C-k") #'lsp-inline-completion-cancel)
+    (define-key map [mouse-movement] #'ignore)
+    (define-key map [t] #'lsp-inline-completion-cancel-with-input)
+    map)
+  "Keymap active when showing inline code suggestions")
+
+;;;###autoload
```
kiennq, line 94: nit: I think we don't need to make this autoload

lsp-inline-completion.el
```diff
@@ -0,0 +1,365 @@
...
+    (lsp-inline-completion--clear-overlay))
+
+  (setq lsp-inline-completion--overlay (make-overlay beg end nil nil t))
+  (overlay-put lsp-inline-completion--overlay 'keymap lsp-inline-completion-active-map)
+  (overlay-put lsp-inline-completion--overlay 'priority 9000)
+
+  lsp-inline-completion--overlay)
+
+
+;;;###autoload
```
kiennq, line 153: Same here. If this is not an interactive command, then we don't need to make it autoload as if it's called from another package, we can just `(require 'lsp-inline-completion` instead

lsp-inline-completion.el
```diff
@@ -0,0 +1,365 @@
...
+
+(defun lsp-inline-completion-show-overlay ()
+  "Makes the suggestion overlay visible"
+  (unless (and lsp-inline-completion--items
+               (numberp lsp-inline-completion--current))
+    (error "No completions to show 2"))
+
+  (lsp-inline-completion--clear-overlay)
+
+  (setq lsp-inline-completion--showing-company
```
kiennq, line 189: I wonder if we want to include `company-mode` integration here as we can have multiple alternative package provide the same functionality as `company-mode`, `corfu` for example.

What will happen if we don't cancel company mode here? The overlay from `company-mode` will be displayed over the inline completion? Can that be mitigate by making the overlay from inline completion topmost or bottommost?

Also, we can put it as a tip for the user (a simple function that the user can put to their init file) instead? Or even suggest the user to use a different company-mode's front end?
I've been using `copilot.el` with `company-mode` and haven't really notice issue so far though.

lsp-mode.el
```diff
@@ -9934,6 +9957,43 @@ string."
     (setf window-scroll-functions
           (delete #'lsp--update-inlay-hints-scroll-function window-scroll-functions)))))
 
+
+;; Inline Completions
```
kiennq, line 9961: nit: If possible we should move this to `lsp-inline-completion.el` instead.
An auto hook can be added to `lsp-inline-completion` to autoload this feature if it's on.
Example:
``` elisp
;;;###autoload
(add-hook 'lsp-configure-hook (lambda ()
                                (when (and lsp-auto-configure
                                           lsp-completion-enable)
                                  (lsp-completion--enable))))
```

lsp-mode.el
```diff
@@ -9934,6 +9957,43 @@ string."
...

+;; Inline Completions
+(defvar-local lsp-inline-completion--idle-timer nil
+  "The idle timer used by lsp-inline-completion-mode")
+
+(define-minor-mode lsp-inline-completion-mode
+  "Mode automatically displaying inline completions."
+  :lighter nil
+  (cond
+   ((and lsp-inline-completion-mode lsp--buffer-workspaces)
```
kiennq, line 9969: Can we use `lsp-on-idle-hook` instead? The `lsp-inlay-hints-mode` is using that and it has been working well so far. It also should fire less frequently compared to `on-change-hook` as when typing, we're likely don't want to request inline completion anyway.

kiennq:
Thank you @kassick for your contribution. I've left some comments in the PR. The feature looks good to me :)

lsp-protocol.el
kassick: I've added the relevante interfaces to the end of the list with the version of the protocol that supports them.

Let me know if you want to have the whole interface definition sorted.

lsp-inline-completion.el
kassick: Yeah, I was trying to make the test suite happy when I have the minor mode defined in `lsp-mode.el`; forgot to remove it later

lsp-inline-completion.el
kassick: Good call!
Though we must check if it is a completion-list or a completion-list-items 

> result: InlineCompletionItem[] | [InlineCompletionList](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#inlineCompletionList) | null defined as follows: 

I've refactored the parse function, should not depend on ht anymore

lsp-inline-completion.el
kassick: > I wonder if we want to include company-mode integration here as we can have multiple alternative package provide the same functionality as company-mode, corfu for example.

Hmmm I agree that having no way of plugging other completion frontends may be an issue -- corfu users would have to advise the inline completion functions to have a similar level of integration, so not very ideal.

> What will happen if we don't cancel company mode here?

It depends on what frontends will be active at the time we display the overlay. With the default company tooltip frontend active, I only managed to get one-liners such as in the video below. It does not obsure the company completion list, but pressing tab or whatever uses the inline completion overlay active map, so this would cause some confusion on the user side.

[Kooha-2024-11-27-15-58-38.webm](https://github.com/user-attachments/assets/df712989-5f55-428b-97df-3139b329381a)

What happens if company-preview frontend is active? Not sure, probably the inline completion would obscure the preview completely.

If the user uses company-box or company-posframe (my case), then overlay priority makes no difference because company displays completions in a child frame :/

>  a tip for the user (a simple function that the user can put to their init file) instead? 

Good one!

I've refactored the company integration into a minor mode. The user can then set up their code such as in the following snippet:

```elisp
  (require 'lsp-inline-completion)
  ;; Show the current/total plus keys in the 
  (add-hook 'lsp-inline-completion-shown-hook #'lsp-inline-completion-show-keys)

  ;; Make inline completions play nice with company mode
  (add-hook 'lsp-inline-completion-mode-hook #'lsp-inline-completion-company-integration-mode)
```

lsp-mode.el
kassick: Oh, that's a neat trick!

lsp-mode.el
kassick: Hmm, I don't think that's a good idea.

I've tested here, and triggering it via `lsp-idle-hook` means we get suggestions after moving the cursor around -- I don't think a user would expect that (I, for one, don't).

Regarding the change hook being more aggressive than idle: Since lsp's idle timer is set both on `post-command-hook` and on `after-change-hook`, I don't see how that could be.

In my implementation, what actually happens in `after-change` is a `cancel-timer` and `run-with-timer` (i.e. timer reschedule). We don't ping the server after each change -- we ping the server after `lsp-inline-completion-idle-delay` seconds have elapsed without any change in the buffer.

lsp-inline-completion.el
```diff
@@ -0,0 +1,487 @@
...
+    (setq this-command 'lsp-inline-completion-display)
+    (lsp-inline-completion-display 'implicit)))
+
+(defun lsp-inline-completion--after-change (&rest _)
+  (when (and lsp-inline-completion-mode lsp--buffer-workspaces)
+    (when lsp-inline-completion--idle-timer
+      (cancel-timer lsp-inline-completion--idle-timer))
+    (let ((buffer (current-buffer)))
+      (setq lsp-inline-completion--idle-timer
+            (run-with-timer lsp-inline-completion-idle-delay
```
kiennq, line 424: nit: `run-with-idle-timer`?

lsp-inline-completion.el
```diff
@@ -0,0 +1,487 @@
...
+  (when (and (buffer-live-p buffer)
+             (eq (current-buffer) buffer)
+             (--none? (funcall it) lsp-inline-completion-inhibit-predicates))
+    (setq last-command this-command)
+    (setq this-command 'lsp-inline-completion-display)
+    (lsp-inline-completion-display 'implicit)))
+
+(defun lsp-inline-completion--after-change (&rest _)
+  (when (and lsp-inline-completion-mode lsp--buffer-workspaces)
+    (when lsp-inline-completion--idle-timer
```
kiennq, line 420: perhaps we can move this outside of the `when` check to make sure `lsp-inline-completion--idle-timer` always be cancelled?

lsp-inline-completion.el
```diff
@@ -0,0 +1,487 @@
...
+
+(defcustom lsp-inline-completion-inhibit-predicates nil
+  "When a function of this list returns non nil, lsp-inline-completion-mode will not show the completion"
+  :type '(repeat function)
+  :group 'lsp-mode)
+
+(defvar-local lsp-inline-completion--idle-timer nil
+  "The idle timer used by lsp-inline-completion-mode")
+
+(define-minor-mode lsp-inline-completion-mode
```
kiennq, line 396: I think we need to mark this as `autoload` as well so the `add-hook` autoload below, when called will know how to trigger this.

lsp-mode.el
```diff
@@ -5429,9 +5430,9 @@ If EXCLUDE-DECLARATION is non-nil, request the server to include declarations."
               (lsp-help-mode)
               (with-help-window lsp-help-buf-name
                 (insert
-		 (mapconcat 'string-trim-right
-			    (split-string (lsp--render-on-hover-content contents t) "\n")
-			    "\n"))))
+         (mapconcat 'string-trim-right
```
kiennq, line 5433: Is the indentation here correct?

kiennq (approved):
I think this looks good now, just a few nit comments you might want to consider for a smooth load of this mode.

lsp-mode.el
kiennq: Ah okay. The Visual Studio will trigger inline completion on mouse move as well while the VS Code seems to only trigger it on buffer modification. Whichever as default is fine for me :)

lsp-inline-completion.el
```diff
@@ -0,0 +1,487 @@
...
+    (lsp-inline-completion--clear-overlay))
+
+  (setq lsp-inline-completion--overlay (make-overlay beg end nil nil t))
+  (overlay-put lsp-inline-completion--overlay 'keymap lsp-inline-completion-active-map)
+  (overlay-put lsp-inline-completion--overlay 'priority lsp-inline-completion-overlay-priority)
+
+  lsp-inline-completion--overlay)
+
+
+(defun lsp-inline-completion-show-keys ()
```
kiennq, line 168: nit: is this called from somewhere? Should we make this a command instead of just a function?

lsp-inline-completion.el
```diff
@@ -0,0 +1,487 @@
...
+                (items (lsp-inline-completion--parse-items resp)))
+
+          (progn
+            (lsp-inline-completion--clear-overlay)
+            (setq lsp-inline-completion--items items)
+            (setq lsp-inline-completion--current 0)
+            (setq lsp-inline-completion--start-point (point))
+            (lsp-inline-completion-show-overlay))
+        (unless implicit
+          (message "No Suggestions!")))
```
kiennq, line 369: nit: `lsp--info` or `lsp--warn`?

lsp-mode.el
kassick: > Visual Studio will trigger inline completion on mouse move as well

It does!?! THE HORRORS!!! :P 

> VS Code seems to only trigger it on buffer modification

We could add a custom variable to control which hook is used, but I would wait for a feature request to avoid implementing something that'd never be used ;) -- If you think you would prefer to have it activated on idle, I can consider this a first feature request and do it in this PR ;)

lsp-mode.el
kassick: Tabs vs Spaces :P 

Fixed.

lsp-inline-completion.el
kassick: Oh, I was calling this on the `shown-hook` , leaving up to the user to configure it. IMHO it does not make sense to show the completion overlay and not the `Completion 1/N` message, so I've updated the show-overlay function to call it.

kassick:
> I think this looks good now, just a few nit comments you might want to consider for a smooth load of this mode.

I've resolved the discussions that have been accepted and fixed an issue with state tracking on lsp-on-change-hook functions -- [quick diff for the commit range](https://github.com/emacs-lsp/lsp-mode/compare/f2b38e0671c4256b17e70e269ce0e17e1dfb21fd..a4243c2fc1da768d9e008f67f36b6eed35ee53ed)

I've left a couple of discussions opened as I'd like your input before merging @kiennq

lsp-inline-completion.el
kiennq: That sounds good to me. If this is called already, can we make this private function instead of public one?

lsp-inline-completion.el
kiennq: I like the Emacs binding. @kassick 
A question about `down-mouse ...` do we need to bind them to `#ignore`? Shouldn't that affect other minor-modes key binding?

kassick: > I like the Emacs binding. @kassick

:+1: 

> A question about down-mouse ... do we need to bind them to #ignore? 

We do need them -- down-mouse-1 is sent before mouse-1. Without the ignore, this event is then captured by the `[t]` in the keymap , which cancels the completion and then checks what was bound in the currently active keymap and executes it.

So we do need to ignore down-mouse-1 if we want mouse-1 to accept the completion. I found this to be the case -- click somewhere in the overlay, have it accepted.

(I did not manage to get the cursor positioned in the correct place afterwards, mouse events in emacs are sourcery to me)

> Shouldn't that affect other minor-modes key binding?

It doesn't affect any other mode because the keymap is only active when the cursor is inside of the overlay -- it's not a minor mode keymap that would override any other mapping.

kassick: I've updated the keymap accordingly.

kassick: RE: _(I did not manage to get the cursor positioned in the correct place afterwards, mouse events in emacs are sourcery to me)_

I've been successfully nerd-snipped by myself ...

Clicking in the completion now accepts it and places the cursor in the correct position.

lsp-inline-completion.el
kassick: done

lsp-inline-completion.el
```diff
@@ -0,0 +1,517 @@
...
+      (lsp--execute-command command?))
+
+    (if completion-is-substr
+        (goto-char lsp-inline-completion--start-point)
+      (goto-char text-insert-start))
+
+    ;; hooks
+    (run-hook-with-args-until-failure 'lsp-inline-completion-accepted-hook text text-insert-start text-insert-end)))
+
+(defun kzk/lsp-inline-completion-accept-on-click (event)
```
kiennq, line 307: nit: `kzk`?

kiennq (approved):

lsp-inline-completion.el
kassick: Fixed

kiennq:
Thanks for your contribution.

kiennq:
@kassick, do you mind provide `copilot-lsp` as a lsp client as well? Probably with the name of `lsp-github-copilot`?

kassick:
> @kassick, do you mind provide `copilot-lsp` as a lsp client as well? Probably with the name of `lsp-github-copilot`?

On it

kurnevsky:
For some reason this mode still collides with company for me, even with `lsp-inline-completion-company-integration-mode`. Company doesn't work even after I disable `lsp-inline-completion-mode`. Not sure why, I'm trying it with my own lsp server :)

PR #4627, opened 2024-11-27
[chore] Update comments with old `document/onHover` API

niontrix:
Why?:
- In the current form of the LSP specification `document/onHover` is replaced by `textDocument/hover`.

This change addresses the need by:
- Replace `document/onHover` with `textDocument/hover`


jcs090218 (approved):

jcs090218:
Thanks!

PR #4630, opened 2024-11-29
Fix ada language server download path

ivvil:
Tested for linux x64


ivvil:
Seems like it's downloading it as a *.tar instead of a *.tar.gz

jcs090218:
I'm not sure about this. Does this work on Windows? 🤔

jcs090218:
Why did you close this? I would like this to be merged. 🤔

ivvil:
Seems like what I was trying to fix got merged on #4567

PR #4632, opened 2024-12-01
    issue 4386: lsp-completion-at-point short-circuits the completion…

clarendonlabs:
… function chain

    lsp-completion-mode will make a buffer-local version of the `completion-at-point-functions' list and will add `lsp-completion-at-point' to the head of the list if that symbol is not already present.

    However, an `lsp-unconfigure-hook' is called which results in removal of `lsp-completion-at-point' from the buffer-local version. The next time `lsp-completion-mode' runs, `lsp-completion-at-point' is added to the head of the `completion-at-points-functions'.

    To fix, a flag variable is set whose use prevents `lsp-completion-at-point' from removed from `completion-at-point-functions' if it was present at the start.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -806,7 +806,9 @@ The return is nil or in range of (0, inf)."
         (after-completion-fn (lambda (result)
                                (when (stringp result)
                                  (lsp-completion--clear-cache))
-                               (setq-local lsp-inhibit-lsp-hooks nil))))
+                               (setq-local lsp-inhibit-lsp-hooks nil)))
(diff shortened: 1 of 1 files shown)


clarendonlabs:
Following an email with Kien Nguyen, I’ve generated a PR for a possible fix to the short-circuiting issue#4386 reported earler.

wyuenho:
If you read the docstring for `completion-at-point-functions`, the plist the `lsp-completion-at-point` can include a kvp `:exclusive 'no` to archive what you want.

PR #4643, opened 2024-12-16
Add SignatureInformation activeParameter support.

brownts:


wyuenho:
Does lsp-mode and friends currently have the ability to highlight the active parameter?

brownts:
> Does lsp-mode and friends currently have the ability to highlight the active parameter?

Yes, it does, although simply moving the cursor over a parameter won't activate it (like what happens in Elisp), unfortunately.  It seems you have to either trigger it to appear (based on the `lsp-signature-auto-activate` setting), or manually activate it (via `M-x lsp-signature-activate`), although even if you manually activate it, `lsp-mode` will automatically deactivate it again when you move off the line.

Oddly, Eglot behaves similar to Elisp in that you only have to move the cursor over the parameter for the signature to appear.  To be honest, I wish `lsp-mode` behaved like that, but it doesn't seem to be setup that way.

wyuenho (approved):

wyuenho:
> lsp-mode will automatically deactivate it again when you move off the line

That probably has to do with the various mouse-movement handlers lsp-mode registers.

> Oddly, Eglot behaves similar to Elisp in that you only have to move the cursor over the parameter for the signature to appear.

That's because lsp-mode's `eldoc-documentation-functions` is only setup to use `textDocument/hover`, whereas eglot uses both `textDocument/hover` and `textDocument/signatureHelp`. lsp-mode's signature help is only setup to trigger after self-insert or triggered from the server.

wyuenho:
@jcs090218 @kiennq this PR should be good to go

jcs090218 (approved):

jcs090218:
LGTM!

PR #4662, opened 2024-12-27
lsp-inline-completion, lsp-copilot: fix lints, package version and default behavior on inserting text

### LINKED ISSUE #4660: (lsp-package-version) error prevents lsp from initializing in buffer

### Bug description

`(lsp-version)` and recently introduced`(lsp-package-version)`  functions return error if the `lsp-mode` package is installed via non recommended methods, e.g. via `elpaca` or `straight.el`, and not with standard `package.el`. It was not a problem until recent introduction of lsp copilot feature (Commit 83da5b6). Now it seems the error prevents `lsp` from initializing (see backtrace).

### Steps to reproduce

1. Install latest version of the package via `elpaca` or `straight.el`.
2. Try to enable `lsp` in a buffer.

### Expected behavior

Successful connection to a language server

Pyright, fort-ls

### OS

Linux

### Error callstack

```emacs-lisp
Debugger entered--Lisp error: (wrong-type-argument package-desc nil)
  package-desc-version(nil)
  lsp-package-version()
  lsp-copilot--server-initialization-options()
  lsp--create-initialization-options(#s(lsp-session :folders ...))
  lsp--start-connection(#s(lsp-session :folders ...))
  #f(compiled-function (client) #<bytecode 0x6adf9810901972b>)(#s(lsp--client ...
  -map(#f(compiled-function (client) #<bytecode 0x6adf9810901972b>) (#s(lsp--client ...)
  lsp--ensure-lsp-servers(#s(lsp-session :folders ...))
  lsp--try-project-root-workspaces(nil nil)
  lsp(nil)
  funcall-interactively(lsp nil)
  command-execute(lsp record)
  execute-extended-command(nil "lsp" "lsp")
  funcall-interactively(execute-extended-command nil "lsp" "lsp")
```

### Anything else?

_No response_

kiennq:
Fix #4660 and #4656

### DIFF
--- a/lsp-inline-completion.el
+++ b/lsp-inline-completion.el
@@ -83,7 +83,7 @@ InlineCompletionItem objects"
     ;; Any event outside of the map, cancel and use it
     (define-key map [t] #'lsp-inline-completion-cancel-with-input)
     map)
-  "Keymap active when showing inline code suggestions")
+  "Keymap active when showing inline code suggestions.")
 
 (defface lsp-inline-completion-overlay-face
   '((t :inherit shadow))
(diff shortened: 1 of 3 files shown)


lsp-inline-completion.el
```diff
@@ -202,40 +202,36 @@ text range that was updated by the completion"
...
            (elt lsp-inline-completion--items
                 lsp-inline-completion--current))
-          ((&InlineCompletionItem? :insert-text :range?) suggestion)
-          ((&RangeToPoint :start :end) range?)
+          ((&InlineCompletionItem? :insert-text :range? (&RangeToPoint :start :end)) suggestion)
           (start-point (or start (point)))
           (showing-at-eol (save-excursion
                             (goto-char start-point)
-                            (eolp)))
+                            (and (not (bolp)) (eolp))))
```
kassick, line 209: If I recall correctly, I had to handle this specifically because the overlay did not display at the right place when EOL. In this case, I had display the overlay at the character before eol (I believe this was a hint I took from the company preview frontend).

I don't remember seeing the same issue on `bol` (and have not been able to get copilot to answer anything at bol, so I can't test it), so I'm curious -- why shift the overlay display point when bol?

kiennq: You can try with this elisp code where the `<|>` is cursor position.
``` elisp
;; fibonacci
<|>
```

Basically, in this case, the cursor will be moved back to the previous line since it's `eol`. However, we don't want that and will want to keep the cursor in the current line.

Here is before the fix
image

kassick: Good catch!

jcs090218 (approved):

jcs090218:
Consider rebasing this to check for any further errors in snapshots.

PR #4665, opened 2024-12-30
feat: add additional LSP support for YAML by adding helm-ls support

Aaronzinhoo:
Adds support for YAML [helm-ls](https://github.com/mrjosh/helm-ls) to provide more targeted features when editing Helm 
Charts and template files.

`helm-ls` starts a helm language server and conditionally starts a [yaml-language-server](https://github.com/redhat-developer/yaml-language-server) which means a lot of configuration for this client is duplicated with the `yaml-language-server` client in `lsp-yaml.el`. The implementation supports all of the helm specific configuration settings and the priority has been set to 0 for now since no major mode will activate the `helm-ls` server. This means users would need to make their own major mode and update the `lsp-language-id-configuration` settings accordingly to make it usable for now. 

I hope to make another PR here soon when I finish the minimal major mode for helm specific template files that I have been using to test with! Although `helm-ls` seems like a superset of the capabilities offered by `yaml-language-server` I would think its main purpose would editing kubernetes and helm / helmfile templated files specifically which would allow adding some tailoring and customizations that eventually diverge from the more generic yaml client! Open to ideas / thoughts here and any changes needed to get this through! 

Happy holidays everyone!

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -24,6 +24,7 @@
   * Add fennel support
   * Add support for [[https://github.com/nextflow-io/language-server][Nextflow]]
   * Add TypeSpec support
+  * Add [[https://github.com/mrjosh/helm-ls][helm-ls]] (Kubernetes Helm) support.
 
 ** 9.0.0
   * Add language server config for QML (Qt Modeling Language) using qmlls.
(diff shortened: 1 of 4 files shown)


jcs090218 (approved):

jcs090218:
LGTM. Thank you!

PR #4682, opened 2025-01-11
Return nil from passthrough-try-completion when table is nil

wyuenho:
Since #4602 the passthrough-try-completion function returns a cons cell even the list of completions is empty, so completion-in-region-functions that rely on the old behavior to call the exit function with an empty string `candidate` and nil `candidates`, which will make the exit func throw. This PR fixes this issue.

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -850,9 +850,10 @@ The return is nil or in range of (0, inf)."
   "Disable LSP completion support."
   (lsp-completion-mode -1))
 
-(defun lsp-completion-passthrough-try-completion (string _table _pred point)
+(defun lsp-completion-passthrough-try-completion (string table _pred point)
   "Passthrough try function, always return the passed STRING and POINT."
-  (cons string point))
+  (when table
+    (cons string point)))
 
 (defun lsp-completion-passthrough-all-completions (_string table pred _point)
   "Passthrough all completions from TABLE with PRED."


kiennq:
I wonder in which case candidate can be `nil` though. Have exit function called on nil seems to be bug of completion framework to me honestly.

wyuenho:
`candidate` won't be nil when given to the the exit func as a parameter, but it can be an empty string, such as the case of point sitting at the end of a line.

`candidate` in the let form can be nil [here](https://github.com/emacs-lsp/lsp-mode/blob/51022dc79486544730de6d7336e3f0a0afe2a5b3/lsp-completion.el#L687) when the CAPF returns nil as the list of `candidates`. In which case, everything that follows blows up.

kiennq:
> `candidate` in the let form can be nil [here](https://github.com/emacs-lsp/lsp-mode/blob/51022dc79486544730de6d7336e3f0a0afe2a5b3/lsp-completion.el#L687) when the CAPF returns nil as the list of `candidates`. In which case, everything that follows blows up.

When CAPF return `nil` as `candidates`, I don't think the exit-fn can be triggered in that case unless that's the new thing from CAPF (and I'm curious of why that even a thing to consider). Normally exit-fn is invoked when a candidate is selected, if there's no candidate, normally nothing can be selected and nothing should be run.
I purposely fallback to a candidate item in the `candidates` list if the exit-fn is invoked with a new string instead of the string that contains the necessary properties. The blow up can happen if the `candidate` and `candidates` list are out-of-sync. In that case, it would be better to allow the blow up to happen instead of silent fail so we can try to fix that, as the out-of-sync shouldn't happen at all.
Do you have a repro of when this happens so we can take a closer look to figure out why the out-of-sync happens?

wyuenho:
Ahhh, it appears I've opened a whole new can of worms...

Corfu's [completion-in-region-function](https://github.com/minad/corfu/blob/c7626a699b55c53a389916ef1816c59e1b853b9b/corfu.el#L927) calls `completion-try-completion`, which will call `lsp-completion-passthrough-try-completion`, which just returns `("" . nil)`, then corfu proceeds to [here](https://github.com/minad/corfu/blob/c7626a699b55c53a389916ef1816c59e1b853b9b/corfu.el#L957), which is when the exit func is called with an empty string and an empty candidates list.

I guess the correct fix should be that `lsp-completion-passthrough-try-completion` should return `nil` when `table` is nil.

wyuenho:
This appears to be a regression from #4602

kiennq:
> I guess the correct fix should be that `lsp-completion-passthrough-try-completion` should return `nil` when `table` is nil.

I agree, this should be a correct fix for the passthough function.

wyuenho:
Should be good to go now

kiennq (approved):

kiennq:
Thank you

PR #4683, opened 2025-01-13
fix: copilot respects the value of lsp-copilot-enabled [fixes #4679]

### LINKED ISSUE #4679: Copilot cannot be turned off

### Bug description

I am not allowed to use copilot at my company, so I have set `lsp-copilot-enabled` to `nil` in `customize-variable`.

The `lsp-copilot` wants to install `copilot-ls` despite the config.

If I let it install the `copilot-ls`, it will ask for login.

### Steps to reproduce

Set the `lsp-copilot-enabled` to `nil`
Open a file that triggers the copilot in `lsp`

### Expected behavior

My expectation was that the disabled `lsp-copilot` does not want to either install the binary or ask for logging.

lsp-copilot

### OS

MacOS

### Error callstack

_No response_

### Anything else?

_No response_

kassick:
`lsp-mode` sugests installing copilot-ls for any supported major-mode that has a installable server. This results in users installing the copilot server when they did not intend to use copilot.

Once installed, copilot-ls would try to log in every time the client started.

This PR updates the copilot client to, by default, activate only when `lsp-copilot-enabled` is non-nil.

It also adds a hint to the user in the login function, helping them find how to disable copilot.

### DIFF
--- a/clients/lsp-copilot.el
+++ b/clients/lsp-copilot.el
@@ -55,7 +55,7 @@ lsp-install-server to fetch an emacs-local version of the LSP."
   :type 'string
   :group 'lsp-copilot)
 
-(defcustom lsp-copilot-applicable-fn (-const t)
+(defcustom lsp-copilot-applicable-fn (lambda (&rest _) lsp-copilot-enabled)
   "A function which returns whether the copilot is applicable for the buffer.
 The input are the file name and the major mode of the buffer."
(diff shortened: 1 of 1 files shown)


kassick:
Fixes https://github.com/emacs-lsp/lsp-mode/issues/4679

kiennq (approved):

PR #4684, opened 2025-01-14
If table is a function, only passthrough if its result is non-nil

wyuenho:
Follow up for #4682

### DIFF
--- a/lsp-completion.el
+++ b/lsp-completion.el
@@ -850,9 +850,16 @@ The return is nil or in range of (0, inf)."
   "Disable LSP completion support."
   (lsp-completion-mode -1))
 
-(defun lsp-completion-passthrough-try-completion (string table _pred point)
-  "Passthrough try function, always return the passed STRING and POINT."
-  (when table
+(defun lsp-completion-passthrough-try-completion (string table pred point)
+  "Passthrough try function.
+
+If TABLE is a function, it is called with STRING, PRED and nil to get
+the candidates, otherwise it is treated as the candidates.
+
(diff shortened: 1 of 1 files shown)


lsp-completion.el
```diff
@@ -850,9 +850,16 @@ The return is nil or in range of (0, inf)."
...
-  "Passthrough try function, always return the passed STRING and POINT."
-  (when table
+(defun lsp-completion-passthrough-try-completion (string table pred point)
+  "Passthrough try function.
+
+If TABLE is a function, it is called with STRING, PRED and nil to get
+the candidates, otherwise it is treated as the candidates.
+
+If the candidates is non-empty, return the passed STRING and POINT."
+  (when (cond ((functionp table)
```
kiennq, line 860: The table can be a hash table as well, probably we should add that case for completeness.
https://www.gnu.org/software/emacs/manual/html_node/elisp/Basic-Completion.html
``` elisp
(pcase table
  ((pred functionp) (funcall table))
  ((pred hash-table-p) (not (ht-empty-p table)))
  (_ table))
```

kiennq (approved):

eval-exec:
Hello @wyuenho , thank you for the fix.

I'm new to lsp-mode and would like to know where `lsp:label-details-description?` is defined. Could you please guide me?

kiennq:
> Hello @wyuenho , thank you for the fix.
> 
> I'm new to lsp-mode and would like to know where `lsp:label-details-description?` is defined. Could you please guide me?

It's defined in `lsp-protocol`, an expansion of the macros over there.

PR #4685, opened 2025-01-15
Fix type and group of lsp-v-analyzer-path

bhrgunatha:
Custom setting lsp-v-analyzer-path has the wrong type and the wrong group.


jcs090218 (approved):

jcs090218:
Thank you!

PR #4686, opened 2025-01-16
lsp-graphql: fix regexp for ext check

suzuki:
## About

`lsp-graphql` should be enabled when matching specified file extensions, but it gets activated when the file path contains strings like "graphql"

### Ex.

Currently, `lsp-graphql` is enabled for the following filename:  
- `/path/to/graphql/test.go`


jcs090218 (approved):

jcs090218:
Thank you!

PR #4692, opened 2025-01-24
Add support for swift-ts-mode

SKyletoft:
[lsp-sourcekit.el already supports swift-ts-mode](https://github.com/emacs-lsp/lsp-sourcekit/blob/master/lsp-sourcekit.el#L84), but as base lsp-mode does not and lsp-sourcekit.el doesn't add it to the list itself lsp-mode does work properly


jcs090218 (approved):

jcs090218:
Thank you!

PR #4611, opened 2024-11-12
Add support for MATLAB language server

JohnC32:
This adds support for the MATLAB Language Server, https://github.com/mathworks/MATLAB-language-server

PR #4699, opened 2025-02-01
basic support for c3 language.

diasbruno:


jcs090218:
You can use #4611 as the reference. Please update the changelog, mkdocs.yml, etc.

diasbruno:
Ah ok. Thanks, @jcs090218.

jcs090218:
Still missing changes in lsp-mode.el and docs/lsp-clients.json.

docs/lsp-clients.json
```diff
@@ -162,6 +162,14 @@
     "lsp-install-server": "csharp-roslyn",
     "debugger": "Yes (netcoredbg)"
   },
+  {
+    "name": "lsp-c3",
```
jcs090218, line 166: The name looks a bit awkward. Maybe change to something else? 🤔

docs/lsp-clients.json
```diff
@@ -162,6 +162,14 @@
...
+    "full-name": "c23 language server",
```
diasbruno, line 167:
```suggestion
    "full-name": "c3 language server",
```

docs/lsp-clients.json
diasbruno: I was following the convention of `lsp-` + `language name`. but we can change...suggestions?

jcs090218: `lsp-` convention is for elisp file name. The `full-name:` should be the name of the language server, how about `c3-lsp` (from the repo name)?

diasbruno: Sure. Thanks for clarifying that. 

```suggestion
    "name": "c3-lsp",
```

jcs090218 (approved):

jcs090218:
Thank you!

diasbruno:
Thank you for helping me with the suggestions! 🎉

diasbruno:
@jcs090218, the page extracted for the doc has this name `lsp-c3-lsp`.

diasbruno:
It's looking for https://emacs-lsp.github.io/lsp-mode/page/lsp-c3.md.

PR #4709, opened 2025-02-12
fix: delay copilot login check and avoid raising error

kassick:
There appears to be some issues with being too eager to check for the login. The server may not have been properly initialized, which results in an error being raised by lsp-copilot--authenticated-as. That seems to wreck havoc in the call stack and leave lsp in a weird state -- there is a copilot server running, there seems to be a client that can be restarted... but there is no workspace registered ...

This change give some time for the server to spin up and avoids raising an error if the authentication request did not work.

### DIFF
--- a/clients/lsp-copilot.el
+++ b/clients/lsp-copilot.el
@@ -43,6 +43,13 @@
   :type 'boolean
   :group 'lsp-copilot)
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

PR #4711, opened 2025-02-14
fix(graphql): use `npm i-peers` instead of `npm-install-peers`

ncaq:
When I run `npm-install-peers` in the `graphql-language-service-cli` directory,
oddly enough, the `graphql-language-service` disappears from `node_modules` as a result of dependency resolution.
As a result, the graphql language server does not work at all.
From the PR I linked to, it seems that the mechanism to run `npm-install-peers` was originally put in for graphql.
[Pull npm peerDependencies as required by elken · Pull Request #3511 · emacs-lsp/lsp-mode](https://github.com/emacs-lsp/lsp-mode/pull/3511)
Maybe if it was a previous version, it would rather not work without installing it.
Replaced with a working package.
I honestly don't know why `npm-install-peers` removes existing packages when dependency conflicts occur.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -8395,7 +8395,7 @@ nil."
                                          (lsp-async-start-process callback
                                                                   error-callback
                                                                   (executable-find "npx")
-                                                                  "npm-install-peers")))))
+                                                                  "i-peers")))))
                                  error-callback
                                  npm-binary
                                  "-g"


jcs090218:
cc @elken

elken:
Seems to work fine using https://github.com/privatenumber/i-peers instead. Don't know why it suddenly stopped working.

In this case they've finally put graphql as a dev dependency, but this is still needed to handle other cases where deps are incorrectly placed in peerDeps and not in devDeps.

ncaq:
From looking at the commits, I thought the installation of peer was added for graphql, so I thought it was rather unnecessary in the current situation where graphql would not work.
I thought about modifying the dependency environment on the server side of graphql, but since the graphql side was doing monorepo and such with v1 of yarn and I didn't understand it well, I compromised and decided to change this to a normal installation.
If you say there are not enough dependencies, it's still easy to issue a PR, but it's hard to turn it off because there are too many.
It's difficult to show if it's not a problem, similar to proving the devil.
[graphiql/packages/graphql-language-service-cli at main · graphql/graphiql](https://github.com/graphql/graphiql/tree/main/packages/graphql-language-service-cli)

Do you know of any other servers that specifically depend on the PEER installation to run?
You should be able to look at that to determine how to deal with the problem.
Personally, I feel that if it doesn't work after npm install, that is clearly a bug in the package, and that is why you should issue a PR to add the dependency.
I think it's easier to be accepted as an obvious bug than to say it doesn't work after doing `npm-install-peers`.

elken:
As I said in my first message, I've already linked a package that works. There really doesn't need to be a massive debate here.

ncaq:
Indeed, if there is something that works, it is reasonable to use it in a straightforward manner.
I was thinking too much about principles and not enough about stable options.

ncaq:
I thought about creating a new PR, but I thought it would be too much trouble to follow the history of how it happened, so I rewrote it to use the package pointed out in this PR.

elken:
Much appreciated. Thank you for bringing this to our attention! 😄

elken (approved):

jcs090218 (approved):

jcs090218:
Thanks to everyone who participated!

PR #4713, opened 2025-02-15
add CI coverage for `aarch64-linux`

peterbecich:
https://github.com/actions/partner-runner-images/blob/main/images/arm-ubuntu-24-image.md

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,23 +41,29 @@ jobs:
       with:
         version: ${{ matrix.emacs-version }}
 
-    - uses: emacs-eask/setup-eask@master
+    - name: Setup Eask
+      uses: emacs-eask/setup-eask@master
       if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
       with:
         version: 'snapshot'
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -38,18 +41,32 @@ jobs:
       with:
         version: ${{ matrix.emacs-version }}
 
-    - uses: emacs-eask/setup-eask@master
+    - name: Setup Eask
+      uses: emacs-eask/setup-eask@master
+      if: matrix.os == 'ubuntu-latest' || matrix.os == 'macos-latest'
```
jcs090218, line 46: I think `windows-latest` is missing here? 🤔

peterbecich: you're right, good catch, thanks

jcs090218 (approved):

peterbecich:
There is something wrong with the `snapshot` jobs, but this is also on `master`.

jcs090218:
No worries! There is a compile error on the `master` with Emacs snapshots.

jcs090218:
Thanks for adding tests for the `arm` build! :D

PR #4721, opened 2025-03-01
Remove completion with no labels and filterText

wyuenho:
Some language servers such as typescript-language-server will occasionally return candidates with an empty string as their labels or filterTexts, this PR filters out these candidates from misbehaving servers, in addition to some slight refactoring to clean up some convoluted logic.


lsp-completion.el
```diff
@@ -334,62 +334,49 @@ Return `nil' when fails to guess prefix."
...
-                                                   &allow-other-keys)
-  "List all possible completions in cached ITEMS with their prefixes.
-We can pass LSP-ITEMS, which will be used when there's no cache.
-The MARKERS and PREFIX value will be attached to each candidate."
+  (-map (-lambda ((item &as &CompletionItem
+                        :label
+                        :filter-text?
+                        :_emacsStartPoint start-point
+                        :score?))
+          `( :label ,(or filter-text? label)
```
kiennq, line 342: There's no guarantee that the `filter-text?` needs to be `nil`, it can be just an empty string which is still evaluated to `t`
See this issue https://github.com/emacs-lsp/lsp-mode/pull/2762

lsp-completion.el
```diff
@@ -614,7 +611,12 @@ Returns resolved completion item details."
                                                              :_emacsStartPoint
                                                              (or (lsp-completion--guess-prefix item)
                                                                  bounds-start)))
-                                                  it))))
+                                                  it)
+                                            ;; remove items with no label or filterText
```
kiennq, line 615: with no label and filterText?

lsp-completion.el
```diff
@@ -614,7 +611,12 @@ Returns resolved completion item details."
...
+                                            (-remove
```
kiennq, line 616: I wonder why we need to filter them out from here? shouldn't the filter function does that already at a later time?

lsp-completion.el
```diff
@@ -625,17 +627,13 @@ Returns resolved completion item details."
...
                                                            ((not done?) :incomplete))
-                                                          :lsp-items nil
                                                           :markers markers
                                                           :prefix prefix)
-                              result (lsp-completion--filter-candidates
-                                      (cond (done?
-                                             (cl-second lsp-completion--cache))
-                                            (lsp-completion-filter-on-incomplete
-                                             (lsp-completion--to-internal items)))
-                                      :lsp-items items
```
kiennq, line 636: The reason why we use `lsp-items` here (probably I should name it as `foo-items` at that time) is because of a flickering issue in `company-mode`. Thus, when the candidates are not ready or empty, we faked it the previous result and do no such things like filtering to keep the computation code down 
See https://github.com/emacs-lsp/lsp-mode/pull/2148

lsp-completion.el
```diff
@@ -561,6 +548,16 @@ Returns resolved completion item details."
          (string-lessp label-left label-right)
        (string-lessp sort-text-left sort-text-right)))))
 
+(defun lsp-completion--make-items (items markers prefix)
```
kiennq, line 551: This function is misleading as we do the filtering here so it would be better to spell it out.
Or we can keep the previous `lsp-completion--filter-candidates` function. How about change `markers` and `prefix` to a combined value of `metadata` instead? Would it make the function clearer?

lsp-completion.el
wyuenho: ahhh this the result of putting down a branch and then coming back to it 2 weeks later forgetting what I've done haha

lsp-completion.el
wyuenho: It doesn't. That to-internal function was a mistake

lsp-completion.el
wyuenho: This is the result of putting down the branch and then coming back 2 weeks later forgetting what I've done

lsp-completion.el
wyuenho: should be fixed now

lsp-completion.el
wyuenho: deleted

jimmywongdialpad:
@kiennq how does this look now?

wyuenho:
@kiennq how does this look?

PR #4725, opened 2025-03-04
not push `company-capf` to `company-backends` when in grouped backends

EasternLi:


smile13241324:
@yyoncho could you have a look here? Without this fix the yasnippet integration in Spacemacs is broken for many layers.

miyabinomuratic:
Can someone review this PR? In case you need some details, let me clarify the issue and the new logic proposed by this PR.

The original code intends to add `company-capf` to `company-backends` if it doesn't exist in the list. To check the existence of `company-capf`, `equal` is used to check each element of `company-backends`:

```emacs-lisp
(setq-local company-backends (cl-adjoin 'company-capf company-backends :test #'equal)))
```

However, the use of `equal` is not appropriate here, because `company-backends` can contain a sublist as described in https://company-mode.github.io/manual/Backends.html#Grouped-Backends.

For example, if `company-capf` is included in a grouped backend in `company-backends` like this:

```emacs-lisp
(setq company-backends '((company-capf :with company-yasnippet)))
 ```

the original code fails to detect `company-capf` inside the sublist and the resulting `company-backends` has two `company-capf` like this:

```emacs-lisp
(company-capf (company-capf :with company-yasnippet))
```

To avoid this issue, the check needs to be done based on type of an element in `company-backends`: if the element is an atom, check if it is `company-capf`, if the element is a list, check if `campany-capf` is a member of the list.

The PR uses the correct logic and I've confirmed that the issue is resolved with it.

kiennq:
The work-around for this is to set `lsp-completion-provider` to `:none` and then your `company-backends` configuration will be kept intact.

miyabinomuratic:
Thank you for your suggestion, but I still would like to the PR to be merged because:
1. the original code's assumption that `company-backends` includes only atoms is incorrect
2. configuring `lsp-completion-provider` to `:none` causes an unwanted side effect: `company-mode` is not enabled via `lsp-mode`. When `lsp-completion-provider` is `:none`, `(company-mode 1)` at line 918 is not executed:

https://github.com/emacs-lsp/lsp-mode/blob/45ab3a83f6b498071635e031c1bfb7127800bf34/lsp-completion.el#L913-L919

miyabinomuratic:
@kiennq could you consider my points listed above and approve the change? Thanks.

PR #4728, opened 2025-03-05
feat(lsp-nix): Allow passing arguments to server

judemille:
Useful sometimes. See the [features page.](https://github.com/nix-community/nixd/blob/main/nixd/docs/features.md)

- [x] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)

- [x] ~~I updated documentation if applicable (`docs` folder)~~ Not applicable.

### DIFF
--- a/clients/lsp-nix.el
+++ b/clients/lsp-nix.el
@@ -54,6 +54,12 @@
   :type 'string
   :package-version '(lsp-mode . "8.0.0"))
 
+(defcustom lsp-nix-nixd-server-arguments '()
+  "Extra arguments for the server."
+  :group 'lsp-nix-nixd
+  :type '(repeat string)
+  :package-version '(lsp-mode . "9.0.1"))
+
 (lsp-defcustom lsp-nix-nixd-formatting-command nil
   "External formatter command with arguments.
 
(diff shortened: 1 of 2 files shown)


jcs090218 (approved):

clients/lsp-nix.el
```diff
@@ -54,6 +54,12 @@
   :type 'string
   :package-version '(lsp-mode . "8.0.0"))
 
+(defcustom lsp-nix-nixd-server-arguments '()
```
jcs090218, line 57: I think setting it to `nil` is recommended for a later version of Emacs. 🤔

judemille: > I think setting it to `nil` is recommended for a later version of Emacs. 🤔

Huh. Where could I find info on that?

jcs090218: It used to be a compile warning (either Emacs 30 or 31), but it's gone now. 🤔 I guess the Emacs Devel included this but reverted it after a few weeks.

jcs090218:
Thank you!

PR #4730, opened 2025-03-10
Add missing argument in call to cl-assert

cpitclaudel:
The current code says:

```
(cl-assert (fboundp lsp-pcase-macroexpander) "not a known LSP interface: %s" interface)
```

This is missing the SHOW-ARGS argument, so a failed assertion triggers a backtrace rather than an error message:

```
ELISP> (cl-assert nil "not a known LSP interface: %s" 'foo)
*** Eval error ***  Wrong type argument: stringp, foo
```


jcs090218 (approved):

jcs090218:
Thank you!

PR #4734, opened 2025-03-13
lsp-find-definition-mouse: Remember click position.

daym:
Remember click position so back-button can go back to it later.

See <https://github.com/rolandwalker/back-button>.


jcs090218 (approved):

jcs090218:
I’ve encountered a small issue: the mark stays active when I return to the original buffer. I’m hoping this isn’t intentional, as it can be a bit annoying.

PR #4735, opened 2025-03-15
fix(lsp-nix): use cons instead of push in :new-connection

mjoerg:
using push modifies lsp-nix-nixd-server-arguments
fix #4728


jcs090218 (approved):

jcs090218:
Thank you!

PR #4710, opened 2025-02-13
lsp-copilot: Replace LSP server with the official GitHub one

suzuki:
## About

Resolve #4708

Replace the LSP server used by the `lsp-copilot` client with the official GitHub one.
https://www.npmjs.com/package/@github/copilot-language-server

PR #4738, opened 2025-03-16
update names from `copilot-node-server` to `copilot-language-server`

peterbecich:
`copilot-node-server` has already been replaced with `copilot-language-server`: https://github.com/emacs-lsp/lsp-mode/pull/4710

### DIFF
--- a/clients/lsp-copilot.el
+++ b/clients/lsp-copilot.el
@@ -22,7 +22,8 @@
 
 ;; Commentary:
 
-;; LSP client for the copilot node server -- https://www.npmjs.com/package/copilot-node-server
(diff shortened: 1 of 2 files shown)


kassick:
We can probably bump the version as well
 
https://github.com/emacs-lsp/lsp-mode/blob/df049efbeb9c7ee6bea5c7bd984cde2ae9e5a1c8/clients/lsp-copilot.el#L82

peterbecich:
let's do that in a separate PR

jcs090218 (approved):

jcs090218:
Thank you!

PR #4739, opened 2025-03-16
fix(lsp-bufls): Resolve timeout issues

suzuki:
## About

- Fix timeout issue
  - The buf LSP server shuts down after exceeding the default timeout of 2 minutes.
  - So set the timeout to 0 to disable it.
- Modify log-format for more readable
  - The default log-format set to `color`
  - The `*buf:stderr*` buffer is difficult to read.

## Ref. `buf beta lsp` help

```
❯ buf beta lsp --help
Start the language server

Usage:
  buf beta lsp [flags]

Flags:
  -h, --help          help for lsp
      --pipe string   path to a UNIX socket to listen on; uses stdio if not specified

Global Flags:
      --debug               Turn on debug logging
      --log-format string   The log format [text,color,json] (default "color")
      --timeout duration    The duration until timing out, setting it to zero means no timeout (default 2m0s)
```

### log-format: `color`

```
[33mWARN[0m	notify returned	{"method": "$/progress"}
[33mWARN[0m	notify returned	{"method": "$/progress"}
```

### log-format: `json`

```json
{"level":"warn","time":"2025-03-14T15:30:36.890+0900","message":"notify returned","method":"$/progress"}
{"level":"warn","time":"2025-03-14T15:30:36.890+0900","message":"notify returned","method":"$/progress"}
```

### DIFF
--- a/clients/lsp-bufls.el
+++ b/clients/lsp-bufls.el
@@ -71,7 +71,7 @@
   :link '(url-lint "https://github.com/bufbuild/buf")
   :package-version '(lsp-mode . "9.0.0"))
 
-(defcustom lsp-buf-args `("beta" "lsp")
+(defcustom lsp-buf-args `("beta" "lsp" "--timeout" "0" "--log-format" "json")
   "Arguments to pass to buf CLI."
   :type '(repeat string)
   :group 'lsp-buf


suzuki:
IMO, the `buf` CLI is primarily a command for compiling and linting protocol buffers.The LSP server is implemented as one of the features of that command, so I think it is affected by the CLI's timeout.

The LSP server is still in beta, I think it would be better to control it using the `--timeout` option for now.

szykes:
"Official" explanation for the behavior of `buf beta lsp`.

> In order for `buf beta lsp` to be longrunning, it should be called with `--timeout 0`, which then sets no timeout for the command. Otherwise, all `buf` commands adhere to a `2m` timeout by default -- we are looking to address this for `buf beta lsp` since it is intended to be a longrunning command, however for the time-being, if you want a longrunning LSP, `--timeout 0` is required.

https://github.com/bufbuild/buf/issues/3700#issuecomment-2748952575

szykes:
I recommend to set `text` for logging. I think this would be fit better in `emacs` because human would read the log normally. And I think it is just easier to read a simple formatted log.

The `json` format is pretty good in an environment where non-human parties read the log.

Example for `text`:
```
WARN    notify returned {"method": "$/progress"}
WARN    notify returned {"method": "textDocument/publishDiagnostics"}
```

jcs090218 (approved):

PR #4742, opened 2025-03-21
Indicate active async requests via spinner in mode-line

joeyjyliu:
* provide a visual indication as lsp is waiting for async responses. It's especially helpful when a large file is being opened.
* add a new custom variable `lsp-display-pending-async-request-via-spinner`


joeyjyliu:
> Could you capture a screenshot or a GIF? I'd love to see how it looks! :)

Here you go! This is the :moon spinner. 
https://github.com/user-attachments/assets/fd42e4ff-720b-44e0-8221-328017dfcd0f

jcs090218:
I'm a little concerned that language servers making many small asynchronous requests could cause the spinner to flash constantly. 🤔 I wonder if this feature might be more distracting than helpful. 😅

joeyjyliu:
That's a valid concern. If I remember correctly spinner would delay the display of spinner for a little bit to minimize the flicker. Moreover there is a new config to turn on/off this feature (by default off).

jcs090218:
I'll leave it open for others to share their thoughts. ;)

ericdallo:
LGTM having it under a flag for now

Issue #3944, opened 2023-02-03
Add option to call `lsp-format-buffer` on save


michaelpj:
It would be nice to have an option to set up buffer formatting on save. This is something that users can definitely do for themselves (e.g. https://github.com/emacs-lsp/lsp-mode/blob/8c478ce4f4ffb7cd3f9c152966d5480f113d7b7e/docs/manual-language-docs/lsp-rust-rls.md#rustfmt), but it's a common IDE option and so it would be nice to provide a simple option for it from a "batteries included" perspective.

ericdallo:
isn't that available already via `lsp-enable-indentation`?

ericdallo:
ah, my bad, yeah, you are correct.
doom-emacs has custom logic to save after formatting buffer we could check how they do, but I agree a variable for that makes sense

yyoncho:
I think we should enforce a new rule. If ChatGPT can answer it, then we don't need an additional flag/setting.

![image](https://user-images.githubusercontent.com/13259670/217102712-e0f371a1-de51-4095-b6de-b8a687815857.png)

ericdallo:
that's crazy hahahahah

michaelpj:
I think it's reasonable to say that people should just know enough elisp to do this themselves. Not all of our users are as good at it as ChatGPT, though ;)

dschrempf:
I agree with @ericdallo that there should be an option for activating or deactivating "format on save" for all `lsp-mode` managed buffers. When searching for `lsp` `format` and `enable`, I get the following 10 (!) results:
```
lsp-html-format-enable
lsp-xml-format-enabled
lsp-yaml-format-enable
lsp-vetur-format-enable
lsp-javascript-format-enable
lsp-typescript-format-enable
lsp-enable-on-type-formatting
lsp-intelephense-format-enable
lsp-svelte-plugin-svelte-format-enable
lsp-rust-analyzer-rustfmt-rangeformatting-enable
```
That's a bit ridiculous. It is easy to configure on a "per-langauge-basis", but then, I think it would be better to be able to opt out than opt in.

dschrempf:
Suggestion:
```elisp
(defvar-local lsp-format-on-save t
  "Format `lsp-mode'-managed buffer before save.")
(defun lsp-format-on-save-not-apheleia ()
  "Format on save using LSP server, not `apheleia'."
  (if lsp-format-on-save
      (progn
[...]

PR #4753, opened 2025-04-03
Add format on save option (issue #3944)

Ndot:
Add format on save (issue #3944) with two option:

- `lsp-format-buffer-on-save` to turn on / off
- `lsp-format-buffer-on-save-list` to specify what major modes to save (nil saves all buffer)

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -575,6 +575,18 @@ before saving a document."
   :type 'boolean
   :group 'lsp-mode)
 
+(defcustom lsp-format-buffer-on-save nil
+  "If non-nil format buffer on save.
+To only format specific major-mode buffers see `lsp-format-buffer-on-save-list'."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-format-buffer-on-save-list '()
+  "If the list is empty format all buffer on save. Else only format buffers
+if their major-mode is in the list."
+  :type '(repeat symbol)
+  :group 'lsp-mode)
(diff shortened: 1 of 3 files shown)


lsp-mode.el
```diff
@@ -575,6 +575,18 @@ before saving a document."
...
+(defcustom lsp-format-buffer-on-save nil
+  "If non-nil format buffer on save.
+To only format specific major-mode buffers see `lsp-format-buffer-on-save-list'."
+  :type 'boolean
+  :group 'lsp-mode)
+
+(defcustom lsp-format-buffer-on-save-list '()
+  "If the list is empty format all buffer on save. Else only format buffers
+if their major-mode is in the list."
+  :type '(repeat string)
```
kiennq, line 587: I think we should store a list of symbols instead of string. Comparing symbol is always faster than compare string and cost less to store it. Plus, it's more natural to set a list of symbols than a list of string.

Ndot: I actually used symbols at first and then changed it to strings cause for some reason though that you be accessible, but I agree with you using symbols is better.

Updated the branch to use symbols.

ericdallo (approved):

jcs090218 (approved):

jcs090218 (approved):

jcs090218:
Thanks for adding this!

Ndot:
@jcs090218 Thanks. Maybe issue #3944 can be closed?

PR #4763, opened 2025-04-18
Fix #4757 only resolve at exit-fn when the candidate requires resolution

### LINKED ISSUE #4757: ts-ls slow to insert completion because of "completionItem/resolve"

### Bug description

I noticed that particular completion items take 1.5 - 2 secs to be inserted.
I traced it down to this:

```elisp
               ;; see #3498 typescript-language-server does not provide the
               ;; proper insertText without resolving.
               (if (lsp-completion--find-workspace 'ts-ls)
                   (lsp-completion--resolve candidate)
                 candidate))
```
Pretty much 99% of the 1.5 - 2 sec latency is from `lsp-completion--resolve`
I know that this was added to support text edits (e.g. auto-imports) etc, but I wonder, how does VSCode do it and not lag for 1 - 2 secs?

https://github.com/user-attachments/assets/dca6b2c2-0538-47d4-9787-1cc4c8258ddf

### Steps to reproduce

Try inserting completion items for a Typescript project

### Expected behavior

Completion should be almost instant

ts-ls

### OS

MacOS

### Error callstack

```emacs-lisp
N/A (performance issue)
```

### Anything else?

_No response_

wyuenho:
The latest typescript-language-server doesn't require resolution for every item, only for snippets and auto imports. This should hopefully speed things up.

Pending https://github.com/typescript-language-server/typescript-language-server/pull/953


lsp-completion.el
```diff
@@ -688,10 +688,15 @@ Others: CANDIDATES"
...
+              (item (plist-get (text-properties-at 0 candidate) 'lsp-completion-item))
               (candidate
                ;; see #3498 typescript-language-server does not provide the
                ;; proper insertText without resolving.
-               (if (lsp-completion--find-workspace 'ts-ls)
+               (if (or (and (lsp-completion--find-workspace 'ts-ls)
+                            lsp-enable-snippet
+                            (eql (lsp:completion-item-insert-text-format? item)
+                                 lsp/insert-text-format-snippet))
+                       (eq (plist-get (lsp:completion-item-data? item) :hasAction) t))
```
kiennq, line 699: what's the `:hasAction` here? Don't use `plist-get` directly as it will not work for hashtable case. The `lsp:completion-item-data` if has a schema in `lsp-protocol` should expose this `hasAction` property as `:has-action`

Also, why increase the cases where we resolve the completion in blocking fashion? Before it only includes `ts-ls`, now it expanded to a new case as well.
The command and action if existed, should be resolved in non-blocking (async) way, see below where we resolve the completion item and apply command or additional snippets in non-blocking fashion. Since the resolved completion item is cached, the early return will be used.

wyuenho: [CompletionItem.data](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionItem) in the protocol has an LSPAny type, it's arbitrary and has no schema. It's meant to be a catch all that every language server uses differently. The hasAction key is specific to ts-ls that I added in the PR I linked to.

> Don't use plist-get directly as it will not work for hashtable case

Ahhh, what else to use? Pretty much every function in this file pretends the hashtable case doesn't exist. I feel like `lsp-use-plists` should just be set to t at this point.

> Also, why increase the cases where we resolve the completion in blocking fashion? Before it only includes ts-ls, now it expanded to a new case as well.

brainfart, fixed.

kiennq: > Ahhh, what else to use?

There should be `lsp-get` or similar method in `lsp-protocol` that works for extracting property from both.

wyuenho: Ohh I see what you mean. Fixed.

wyuenho: > The command and action if existed, should be resolved in non-blocking (async) way, see below where we resolve the completion item and apply command or additional snippets in non-blocking fashion. Since the resolved completion item is cached, the early return will be used.

Ah thanks for pointing this out as well. I've deleted the hasAction condition

kiennq (approved):

jcs090218 (approved):

jcs090218:
Thank you!

PR #4774, opened 2025-04-29
Adds remove-unused-imports to lsp-mode.el

### LINKED ISSUE #4773: Support for source.organizeImports
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? No.

Describe your feature here.

An lsp command for running the source.organizeImports command.

phuhl:
Hi there,

great work on this project!

I was wondering, why organizing imports did not remove unused imports as tide-mode does and noticed, that `source.organizeImports` was not implemented.

I added one line to add support, it works on my machine. 

Cheers

Closes #4773

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6324,6 +6324,8 @@ execute a CODE-ACTION-KIND action."
 
 (lsp-make-interactive-code-action organize-imports "source.organizeImports")
 
+(lsp-make-interactive-code-action remove-unused-imports "source.removeUnusedImports")
+
 (defun lsp--make-document-range-formatting-params (start end)
   "Make DocumentRangeFormattingParams for selected region."
   (lsp:set-document-range-formatting-params-range (lsp--make-document-formatting-params)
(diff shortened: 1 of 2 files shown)


lsp-mode.el
```diff
@@ -6324,6 +6324,8 @@ execute a CODE-ACTION-KIND action."
 
 (lsp-make-interactive-code-action organize-imports "source.organizeImports")
 
+(lsp-make-interactive-code-action remove-unused-imports "source.removeUnusedImports")
```
ericdallo, line 6327: I dont see that source on https://microsoft.github.io/language-server-protocol/specifications/lsp/3.18/specification/#codeActionKind

phuhl: Uh, well... I assumed it was in the spec because typescript language server supports it: https://github.com/typescript-language-server/typescript-language-server?tab=readme-ov-file#features

How do we proceed? I mean, it is easy enough to make the function call I added in the PR outside of the module, but I thought, it would be helpful for others if this was part of lsp-mode...

ericdallo: If only a single language server is using, not sure it's a good idea add a function to whole lsp-mode, probably something specific for ts is best, in the lsp-typescript only.

phuhl: Got it, did some changes. Does this work for you?

ericdallo (approved):

jcs090218 (approved):

jcs090218:
I've opened the fix in https://github.com/emacs-lsp/lsp-mode/pull/4780.

Issue #4768, opened 2025-04-27
LSP :: invalid config value rust-analyzer


vikigenius:
### Thank you for the bug report

- [x] I am using the latest version of `lsp-mode` related packages.
- [x] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [x] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Latest version of rust-analyzer produces warnings

```
LSP :: invalid config value:
/cargo/extraEnv: invalid type: sequence, expected a map;
```

### Steps to reproduce

Install latest version of rust-analyzer and lsp-mode. Open a Rust project and see the warnings

### Expected behavior

No warning messages

### Which Language Server did you use?

rust-analyzer

### OS

Linux

### Anything else?

Possibly this type has changed on the server side

```
defcustom lsp-rust-analyzer-cargo-extra-env []
```

miyabinomuratic:
cargo.extraEnv has been a map since its introduction (rust-lang/rust-analyzer@c407cc554ee8657081ae5357a05025d3eaca1184). The definition in lsp-mode didn't match since then, but the mismatch was silently ignored. From rust-lang/rust-analyzer@46ce4746e67aa7e82a0520f8b52a99b9b3130906, rust-analyzer started showing a warning message on configuration error, which we are seeing.

With the diff below, the definition for cargo.extraEnv should match and the warning message is not displayed anymore, but I'm not sure if this is a proper fix. Please check if this is OK.

```
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -851,10 +851,10 @@ or JSON objects in `rust-project.json` format."
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "9.0.0"))
 
-(defcustom lsp-rust-analyzer-cargo-extra-env []
+(defcustom lsp-rust-analyzer-cargo-extra-env #s(hash-table)
   "Extra environment variables that will be set when running cargo, rustc or
 other commands within the workspace.  Useful for setting RUSTFLAGS."
-  :type 'lsp-string-vector
+  :type 'alist
```

miyabinomuratic:
Looks like the patch is OK? I'll open a PR if no one objects.

miyabinomuratic:
[...]

PR #4787, opened 2025-05-17
lsp-rust: fix definition of lsp-rust-analyzer-cargo-extra-env (#4768)

miyabinomuratic:
In rust-analyzer, cargo.extraEnv has been a map since its introduction (rust-lang/rust-analyzer@c407cc5). The definition in lsp-mode hasn't matched since then, but the mismatch has been silently ignored. From rust-lang/rust-analyzer@46ce474, rust-analyzer started showing a warning message on configuration error which we are seeing.

With this change, the definition for cargo.extraEnv should match and the warning message is not displayed.

### DIFF
--- a/clients/lsp-rust.el
+++ b/clients/lsp-rust.el
@@ -851,10 +851,10 @@ or JSON objects in `rust-project.json` format."
   :group 'lsp-rust-analyzer
   :package-version '(lsp-mode . "9.0.0"))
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218 (approved):

jcs090218:
Thank you!

PR #4788, opened 2025-05-17
Fix lsp--on-rename-file so that rename-file is always applied

djr7C4:
When `lsp-request` returns `nil` in `(when-let* ((edits (lsp-request "workspace/willRenameFiles" params)))`, the advised function `rename-file` is not run. This results in problems with LSP servers that do not support this request. I encountered this issue in particular with `jdtls`.


jcs090218:
Thank you!

PR #4796, opened 2025-05-26
Refactor lsp--create-filter-function for better performance

eval-exec:
- [x] I added a new entry to [CHANGELOG.md](https://github.com/clojure-lsp/clojure-lsp/blob/master/CHANGELOG.md)

- [x] I updated documentation if applicable (`docs` folder)

Refactored lsp--create-filter-function for lower latency and higher efficiency:

- Avoids string concatenation for message accumulation. Instead, uses a buffer and direct insertion, reducing memory copies and garbage.

- Parses headers and body in-place with minimal allocations, leveraging buffer regions and search instead of splitting strings.

- Dispatches messages as soon as a full message is available (header + body), eliminating message queue latency. Each message is parsed and sent to lsp--parser-on-message immediately upon arrival.

- Keeps parsing in a tight loop, so multiple messages in a single input are handled efficiently.

Result:
 - Lower latency, fewer memory allocations, and better performance when handling large or frequent LSP messages.


eval-exec:
> The CI is failing which indicates that there might be edge case where this change fails to work. Can you make the CI passes?

Thanks for pointing it out. I’ll take a look and fix it when I have time.

eval-exec:
@kiennq Hello, CI passed, request code review. Thank you.

lsp-mode.el
```diff
@@ -7151,69 +7151,44 @@ server. WORKSPACE is the active workspace."
...
+                            (header-lines (split-string headers "\r\n" t))
+                            (parsed-headers (mapcar #'lsp--parse-header header-lines))
+                            (body-length (lsp--get-body-length parsed-headers))
+                            (body-start header-end)
+                            (body-end (+ header-end body-length)))
+                      (when (<= body-end (point-max))
+                        ;; Copy and decode body
+                        (with-current-buffer json-body-buffer
+                          (erase-buffer)
+                          (insert-buffer-substring input-buffer body-start body-end)
```
kiennq, line 7180: Do we need this? Can't we do that in the `input-buffer`, i.e., move the cursor to header end and just call `lsp-json-read-buffer`, we save one copy to body buffer

PR #4803, opened 2025-05-30
Make lsp-headerline update breadcrumb when previewing with consult-xref

### LINKED ISSUE #4802: lsp-headerline doesn't update breadcrumb when previewing with consult-xref--preview

### Bug description

When previewing results with consult-xref with lsp-headerline-breadcrumb-mode enabled, the breadcrumb doesn't update to match the previewed buffer as you scroll through the results.

Consult provides 'consult-after-jump-hook for this. We already have similar handling for 'xref-after-jump-hook and can just copy the same logic for consult.

### Steps to reproduce

```
;; Ensure that consult is loaded.
(require 'consult)

;; Ensure that xref-show-xrefs-function is bound to consult-xref.
(setq xref-show-xrefs-function #'consult-xref)

;; Ensure that lsp-headerline-breadcrumb-mode is enabled
(lsp-headerline-breadcrumb-mode)

;; Place point over a symbol and invoke lsp-find-references
(lsp-find-references)
```

See gif showing current behavior.

Image

### Expected behavior

See gif showing expected behavior (using changes that I have made locally).

Image

Not applicable to this issue, but C++ with clangd.

### OS

Linux

### Error callstack

```emacs-lisp

```

### Anything else?

I plan to put up a PR with a fix soon.

dhashe:
lsp-mode doesn't depend on 'consult being loaded, but it is safe in emacs to call add-hook before loading the package that provides the hook, and that is the simplest approach here. If 'consult is not installed or is never loaded, then the hook will simply never be used.

Fixes #4802

after-fix

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -38,6 +38,7 @@
   * Add ~lsp-nix-nixd-server-arguments~ to allow passing arguments to the ~nixd~ LSP.
   * Improve the lsp-ocaml client (see [[https://github.com/emacs-lsp/lsp-mode/issues/4731][#4731]] for the follow-up issue. MRs: [[https://github.com/emacs-lsp/lsp-mode/pull/4741][#4741]], [[https://github.com/emacs-lsp/lsp-mode/pull/4732][#4732]])
   * Add support for ~source.removeUnusedImports~ for lsp-javascript
+  * Fix lsp-headerline to update breadcrumb when previewing with consult-xref.
 
 ** 9.0.0
   * Add language server config for QML (Qt Modeling Language) using qmlls.
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -441,6 +441,7 @@ PATH is the current folder to be checked."
     (add-to-list 'header-line-format '(t (:eval (window-parameter nil 'lsp-headerline--string) )))
 
     (add-hook 'xref-after-jump-hook #'lsp-headerline--check-breadcrumb nil t)
+    (add-hook 'consult-after-jump-hook #'lsp-headerline--check-breadcrumb nil t)
 
     (add-hook 'lsp-on-idle-hook #'lsp-headerline--check-breadcrumb nil t)
     (add-hook 'lsp-configure-hook #'lsp-headerline--enable-breadcrumb nil t)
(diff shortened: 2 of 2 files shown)


kiennq:
Shouldn't this be an issue of `consult-xref` instead? I.e., `consult-xref` should subscribe/invoke to `xref-after-jump-hook`. That seems to be more approriate

dhashe:
I discussed this with the consult author here: https://github.com/minad/consult/discussions/1229

I don't think that it makes sense for consult-xref to invoke xref-after-jump-hook during previews, because the preview action even on consult-xref doesn't use xref. xref is only called when doing the final jump after selecting a candidate.

It also wouldn't solve this problem in general if it did, because there are other consult commands besides consult-xref that can end up previewing buffers that have lsp-headerline-breadcrumb-mode enabled. For example, consult-mark, which selects between and previews the various marker locations in the marker list.

consult-mark does not use xref in any way, and xref is not necessarily even loaded when consult-mark is called. But it uses the same preview logic, and if we call it on a buffer with lsp-headerline-breadcrumb-mode enabled then the breadcrumb similarly does not update as you scroll through the candidates.

I think that it's better to frame the issue like this:

- lsp-headerline-breadcrumb-mode provides the "breadcrumb" feature, and wants the breadcrumb to be correct in common cases
- xref provides the "jump" feature, and lsp-headerline-breadcrumb-mode hooks into it using the provided hook to update the breadcrumb on xref jumps
- consult (if present) provides a related but distinct "preview-jump" feature, and lsp-headerline-breadcrumb-mode should hook into it using the provided hook to update the breadcrumb on consult previews

And with that framing I think that this is the best place to make the change, although I concede that it's a little weird to add to a hook for a project (consult) that is not a dependency of lsp-mode. I welcome other ideas for where it would make sense to put this.

dhashe:
Alternatively, this is solvable in user configuration (with the wrinkle of needing to call an internal function) with the following:

```
(add-hook 'lsp-headerline-breadcrumb-mode-hook
          (lambda ()
            (if lsp-headerline-breadcrumb-mode
                (add-hook 'consult-after-jump-hook #'lsp-headerline--check-breadcrumb nil t)
                (remove-hook 'consult-after-jump-hook #'lsp-headerline--check-breadcrumb t))))
```

Maybe that's better and we can close this and leave to downstream? Maybe it's worth exposing a public version of lsp-headerline--check-breadcrumb? What do you think?

kiennq:
> Maybe it's worth exposing a public version of lsp-headerline--check-breadcrumb? What do you think?

I agree, expose `lsp-headerline--check-breadcrumb` and update the wiki for this usage seems to be the most probable approach. Feel free to update your PR with that.

dhashe:
Let me know if `lsp-headerline-check-breadcrumb` should be autoload and/or interactive. I wasn't sure.

dhashe:
@kiennq polite bump on this. anything else needed for merge?

kiennq:
Thank you for your contribution.

PR #4808, opened 2025-06-05
lsp-system-path: return absolute path

kiennq:
Make sure to return the absolute path as documented


jcs090218:
Make sense! Thank you!

PR #4810, opened 2025-06-09
Fix compatibility with golangci-lint version 2+

Gleek:
From:
[golangci-lint-langserver README](https://github.com/nametake/golangci-lint-langserver/tree/main?tab=readme-ov-file#golangci-lint-version-compatibility
)
- For golangci-lint v2+: Use `--output.json.path stdout --show-stats=false` parameters
- For golangci-lint v1: Use `--out-format json` parameter


jcs090218 (approved):

jcs090218:
Thank you!

PR #4817, opened 2025-06-17
fix: extract major version from custom golangci-lint

hiroxy:
Allows the "v" prefix in the version string because a custom binary of golangci-lint built using the [module plugin system](https://golangci-lint.run/plugins/module-plugins/) has that prefix.

* original binary
`golangci-lint has version 2.1.6 built with go1.24.2 from eabc263 on 2025-05-04T15:36:41Z`

* custom binary
`golangci-lint has version v2.1.6-custom-gcl built with go1.24.3 from ? on 2025-05-18 10:07:06.838917 +0000 UTC`


jcs090218 (approved):

jcs090218:
Thank you!

PR #4818, opened 2025-06-21
feat: added odin language support through ols

samwdp:
Added odin language support though the [ols](https://github.com/DanielGavin/ols) server by Daniel Gavin 
This will automatically install the server based on the nightly release provided by the repository and should work for all platforms. I have only been able to test on Windows but I have included the binary names for mac os (x86 and arm) and also the linux binary. 

A user should install either [odin-mode](https://github.com/mattt-b/odin-mode) (this hasn't seen active development for the last 2 years and the maintainer seems to have abandoned it) or [odin-ts-mode](https://github.com/Sampie159/odin-ts-mode) (still in development)

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -41,6 +41,7 @@
   * Add Python(ty) support
   * Add support for [[https://github.com/tombi-toml/tombi][Tombi language server]] (TOML language).
   * Make lsp-headerline--check-breadcrumb public
+  * Added Odin langauge server support [[https://github.com/DanielGavin/ols][ols]]
 
 ** 9.0.0
   * Add language server config for QML (Qt Modeling Language) using qmlls.
--- a/clients/lsp-odin.el
+++ b/clients/lsp-odin.el
@@ -48,9 +48,7 @@
                 ((and (eq system-type 'gnu/linux)
                       (or (eq (string-match "^x86_64" system-configuration) 0)
                           (eq (string-match "^i[3-6]86" system-configuration) 0)))
-                 "ols-x86_64-unknown-linux-gnu.zip")
-
-                (t "omnisharp-mono.zip")))
+                 "ols-x86_64-unknown-linux-gnu.zip")))
   "Automatic download url for ols language server."
   :group 'lsp-odin-ols
   :type 'string)
--- a/clients/lsp-odin.el
+++ b/clients/lsp-odin.el
@@ -45,14 +45,13 @@
                 "ols-arm64-darwin.zip"
               "ols-x86_64-darwin.zip"))
            (_
-            "ols-x86_64-unknown-linux-gnu")))
-        (when suffix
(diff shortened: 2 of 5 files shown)


jcs090218:
For the record, there is a fork from original odin-mode at https://git.sr.ht/~mgmarlow/odin-mode.

jcs090218:
Hmmm... I am not sure why the CI fails. 🤔

samwdp:
> Hmmm... I am not sure why the CI fails. 🤔 

Same, I haven't touched anything in the clangd server. And the 2 previous commits passed all the checks. When I make compile locally it is all fine but as I stated I'm on Windows. I just checked using wsl and it is also fine. I don't have an arm machine 🤔

jcs090218:
The problem is the `f-join` doesn't handle `nil` for you.

```elisp
(defcustom lsp-odin-ols-binary-path
  (f-join lsp-odin-ols-server-install-dir "latest"
          ;; ERROR, this cond returns `nil'!
          (cond ((eq system-type 'windows-nt)
                 "ols-x86_64-pc-windows-msvc.exe")
                ((eq system-type 'darwin)
                 (if (string-match "aarch64-.*" system-configuration)
                     "ols-arm64-darwin"
                   "ols-x86_64-darwin"))
                ((and (eq system-type 'gnu/linux)
                      (or (eq (string-match "^x86_64" system-configuration) 0)
                          (eq (string-match "^i[3-6]86" system-configuration) 0)))
                 "ols-x86_64-unknown-linux-gnu")))
  "The path where ols binary after will be stored."
  :group 'lsp-odin-ols
  :type 'file)
```

The `cond` cannot be `nil`:

samwdp:
> The problem is the `f-join` doesn't handle `nil` for you.
> 
> ```emacs-lisp
> (defcustom lsp-odin-ols-binary-path
>   (f-join lsp-odin-ols-server-install-dir "latest"
>           ;; ERROR, this cond returns `nil'!
>           (cond ((eq system-type 'windows-nt)
>                  "ols-x86_64-pc-windows-msvc.exe")
>                 ((eq system-type 'darwin)
>                  (if (string-match "aarch64-.*" system-configuration)
>                      "ols-arm64-darwin"
>                    "ols-x86_64-darwin"))
>                 ((and (eq system-type 'gnu/linux)
>                       (or (eq (string-match "^x86_64" system-configuration) 0)
>                           (eq (string-match "^i[3-6]86" system-configuration) 0)))
>                  "ols-x86_64-unknown-linux-gnu")))
>   "The path where ols binary after will be stored."
>   :group 'lsp-odin-ols
>   :type 'file)
> ```
> 
> The `cond` cannot be `nil`:

right, got it. I switched over to a pcase statement defaulting to the linux versions

jcs090218 (approved):

jcs090218:
Awesome! Thank you for your cooperation! ;)

PR #4821, opened 2025-06-24
lsp-erlang: add new ELP semantic token mappings

alanz:
exported type,
deprecated function
dynamic type for variable

Note: exported type is not in the current ELP release, but is in the repo


jcs090218 (approved):

jcs090218:
LGTM! Thank you! :D

Issue #4454, opened 2024-05-20
Wrong server selected


rogorido:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

Every time I open a JS file lsp-mode selects vue-semantic-server. Even if I remove all emacs.d/ related files (.emacs, .init.el, etc.). There is some configure which is getting loaded somewhere. Where could it be?

### Steps to reproduce

1. Fresh emacs install 
2. M-x lsp-start-plain 
3. open a JS file.

### Expected behavior

Open the JS files with ts-ls and not with vue-semantic-server

### Which Language Server did you use?

lsp

### OS

Linux

### Error callstack

```emacs-lisp
In lsp-log I have: 


Command "vue-language-server --stdio" is present on the path.
Command "semgrep lsp" is not present on the path.
Command "deno lsp" is not present on the path.
Command "typescript-language-server --stdio" is present on the path.
Command "vue-language-server --stdio" is present on the path.
Command "semgrep lsp" is not present on the path.
Command "deno lsp" is not present on the path.
Command "typescript-language-server --stdio" is present on the path.
Found the following clients for /home/igor/geschichte/web/code/apis/apiatlantocracies/main.js: (server-id vue-semantic-server, priority 0), (server-id ts-ls, priority -2)
The following clients were selected based on priority: (server-id vue-semantic-server, priority 0)
```
```


### Anything else?

_No response_

erickisos:
I just came here to mention that I'm having a similar issue.

Given that I opened a graphql file instead of a TSX one when opening the project lsp mode suggested Graphql-lsp server (which is cool for the gql file), but whenever I open a tsx, it doesn't change the selected client, and I can't select it using `+lsp/switch-client` in doomemacs.

jcs090218:
Try configure the variable `lsp-enabled-clients`.

erickisos:
Currently, that variable is empty, however, after manually installing the corresponding clients it apparently worked!

alex-jara:
[...]

PR #4826, opened 2025-06-30
lsp-volar: enable hybrid mode by default & add add-on mode

ctjhoa:
`@vue/language-server` documentation state the following:

> Note: The "Take Over" mode has been discontinued. Instead, a new "Hybrid" mode has been introduced. In this mode, the Vue Language Server exclusively manages the CSS/HTML sections. As a result, you must run @vue/language-server in conjunction with a TypeScript server that employs @vue/typescript-plugin. Below is a streamlined configuration for Neovim's LSP, updated to accommodate the language server following the upgrade to version 2.0.0.

[source](https://github.com/vuejs/language-tools/tree/v2.2.10?tab=readme-ov-file#hybrid-mode-configuration-requires-vuelanguage-server-version-200)
 
In addition, lsp-mode state the following

> I have multiple language servers registered for language FOO. Which one will be used when opening a project?
>
> The highest number is highest priority. Note this is the opposite of [Unix priority (niceness)](https://en.wikipedia.org/wiki/Nice_(Unix)). Servers defined in `lsp-mode` tend to have lower priority than the external packages (priority 0 if unspecified). If a server is registered with `:add-on?` flag set to `t` it will be started in parallel to the other servers that are registered for the current mode. If the server that you want to use is not with the highest priority you may use `lsp-disabled-clients` to disable the server with higher `priority` or use `lsp-enabled-clients` to enable only the servers you want to use. In order to find the server ids you may check `*lsp-log*` buffer.

[source](https://github.com/emacs-lsp/lsp-mode/blob/8579c6c7771bc65564302e76296fb48855c558a4/docs/page/faq.md?plain=1#L20)

So this PR, enable hybrid mode by default as "Take Over" mode is discontinued and also create a new configuration to allow volar to be run aside other lsp (typescript for example).

Related issue: #4454

Code was inspired by
https://github.com/emacs-lsp/lsp-mode/blob/8579c6c7771bc65564302e76296fb48855c558a4/clients/lsp-sorbet.el#L41

### DIFF
--- a/clients/lsp-volar.el
+++ b/clients/lsp-volar.el
@@ -53,6 +53,12 @@
   :group 'lsp-volar
   :package-version '(lsp-mode . "9.0.1"))
 
+(defcustom lsp-volar-as-add-on nil
+  "Run volar LSP server alongside other LSP server(s)"
+  :type 'boolean
+  :group 'lsp-volar
+  :package-version '(lsp-mode . "9.0.1"))
+
 (defcustom lsp-volar-activate-file ".volarrc"
   "A file with a custom name placed in WORKSPACE-ROOT is used to force enable
  volar when there is no package.json in the WORKSPACE-ROOT."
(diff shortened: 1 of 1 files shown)


ctjhoa:
Note: Vue language server [no longer provide a configuration option to disable hybrid mode](https://github.com/vuejs/language-tools/pull/5248)

jcs090218:
Thank you!

PR #4827, opened 2025-07-01
lsp-format-region: Fix error "no region"

sunlin7:
The `lsp-format-region` will emit error message `The mark is not set now, so there is no region` at the very beginning.
It can be reproduced with follow steps:
Start the Emacs with lsp, and open a c++ file `/tmp/a.cpp`, and execute `lsp-format-region`, then got the error message.

To workaround the issue, just select one char/word and de-select the region, then execute the `lsp-format-region`. 

This PR will follow the document of `lsp-format-region` description to accept region or line. 

Please help review and approve this change, thanks

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6096,7 +6096,10 @@ Request codeAction/resolve for more info if server supports."
 
 (defun lsp-format-region (s e)
   "Ask the server to format the region, or if none is selected, the current line."
-  (interactive "r")
+  (interactive
+   (if (use-region-p)
+       (list (region-beginning) (region-end))
+     (list (point) (point))))
   (let ((edits (lsp-request
                 "textDocument/rangeFormatting"
                 (lsp--make-document-range-formatting-params s e))))


jcs090218:
Isn't that the expected behaviour? 🤔 `lsp-format-region` without selecting a region isn't allowed.

sunlin7:
@jcs090218 
No. The document of `lsp-format-region` says it should be current line when no region been selected:
`Ask the server to format the region, or if none is selected, the current line.`

lsp-mode.el
```diff
@@ -6096,7 +6096,10 @@ Request codeAction/resolve for more info if server supports."
 
 (defun lsp-format-region (s e)
   "Ask the server to format the region, or if none is selected, the current line."
-  (interactive "r")
+  (interactive
+   (if (use-region-p)
+       (list (region-beginning) (region-end))
+     (list (point) (point))))
```
jcs090218, line 6102: Could it be `(list (line-beginning-position) (line-end-position))`? 🤔

sunlin7: Current `(list (point) (point))` will be minimum changes for variant lsp backends; 
The `(list (line-beginning-position) (line-end-position))` maybe lead unexpected behavior on very long line for some lsp backends.

jcs090218:
Make sense! Thank you!

sunlin7:
Thank you!

wyuenho:
Why aren't `beginning-of-line` and `end-of-line` used for the current line?

sunlin7:
Ffor no selected region, using the '(beg-of-line end-of-line) will force the lsp-backend formatting current line. 
While `(list (point) (point))` will let the lsp-backend get chance to indent a line or a continued line. 
Then lsp-mode lisp part don't have to  know code syntax.

wyuenho:
> While `(list (point) (point))` will let the lsp-backend get chance to indent a line or a continued line. 

Which LSP servers have you tested that exhibit this behavior? Don't LSP servers handle this case using the `textDocument/onTypeFormatting` request?

sunlin7:
@wyuenho Please try lsp format region with clangd on follow lines which wrapped lines by `\`, the lsp format region still works on follow lines:
>int a \\
= 0;

PR #4829, opened 2025-07-04
fix: Correct postgres-ls system bin name

Briaoeuidhtns:
Fixes https://github.com/emacs-lsp/lsp-mode/discussions/4754#discussioncomment-12811041


jcs090218:
Thanks for fixing this!

PR #4833, opened 2025-07-12
fix(beancount): defer evaluation of initialization options

aisamu:
Per-project `journal_file` configuration was being ignored, as the backquote expansion is too eager (load time, not execution time)

Thanks for lsp-mode!


jcs090218 (approved):

jcs090218:
Thank you!

PR #4841, opened 2025-07-20
fix: Ols server download url getting malformed

samwdp:
f-join was collapsing `//` (doubleslash) into a single slash. This was causing the resulting URL to be incorrect when getting the binary from github. Using `format` as this will not modify the string in any unexpected way


jcs090218 (approved):

jcs090218:
Thank you!

PR #4842, opened 2025-07-21
fix: defer onTypeFormatting to after running other hooks

kassick:
If other functions on post-self-insert-hook (e.g. smart-parens, electric-indent) run after LSP's on-type-formatting, they may modify the buffer and therefore cancel the async reques and prevent on-type-formatting from working as expected.

This commit updates the hook to run in a timer (after 0 seconds), so it lets all other functions perform their changes before asking the server for a formatting action.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5036,6 +5036,8 @@ movements may have changed the position")
   :group 'lsp-mode
   :type 'boolean)
 
+(defun lsp--on-type-formatting-do-apply (data)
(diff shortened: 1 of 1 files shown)


kassick:
Another option would be to make the request synchronous (maybe have a flag, so users who want to use on-type-formatting with other hooks may opt-in to a potentially slower operation)

jcs090218:
LGTM. Thanks for the changes! :)

PR #4843, opened 2025-07-23
fix: cancel inline completion UI on non mapped commands

kassick:
On some situations, keybindings such as M-x are not caught by the default handler (lsp-inline-completion-cancel-with-input, via overriding-terminal-local-map). This results in the tooltip not being hidden when it should. The consequence is that the user has to explicitly call lsp-inline-completion-cancel to hide the completion overlay and continue working in the buffer

This change fixes that by placing a symbol property on the commands that should not cancel the UI (accept, next, prev, recenter-top-botton and ignore). On pre-command hook, we check for that property and cancel the completion if needed.


kassick:
(Done with pushing, should be good to review by now ;) )

kassick:
Should we continue with this PR ?

kassick:
Hey @jcs090218 , could I maybe have a yay-or-nay on this bugfix?

I had been running into weird situations where the completion overlay was not hidden when it should. I've followed the same strategy as company-mode (using pre-command-hook to check if the command is marked to either remain in the completion ui or not).

I have been using a fork of lsp-mode with these changes incorporated since Jun/25 and I've seen no issues.

kiennq (approved):

jcs090218:
Thank you for adding this!

PR #4851, opened 2025-08-05
Update  elixirLs for lsp-elixir and add new option introduced in v0.29.0

wkirschbaum:
reference: https://github.com/elixir-lsp/elixir-ls/releases/tag/v0.29.0

The download link also changed and will be updated with this commit.


jcs090218 (approved):

jcs090218:
Thank you!

PR #4852, opened 2025-08-06
Revert lsp-elixir-ls-download-url and update version to v0.29.1

wkirschbaum:
The previous url change was a mistake on the release for v0.29.0 and was corrected on v0.29.1.  The old format would be preserved for future minor releases, because other integrations like zed relies on the existing format.

Sorry for the back and forth on this one, I might have jumped the gun to quickly on the first PR for v0.29. 

Reference: https://github.com/elixir-lsp/elixir-ls/releases/tag/v0.29.1


jcs090218:
Thank you!

PR #4855, opened 2025-08-13
Fix: Don't watch Haskell IDE cache directories

walseb:
Ref: https://github.com/josephsumabat/static-ls/issues/175

Hello!

Some Haskell tooling (not haskell-language-server I believe) create folders to store some cache information about the project. These cause significant slowdown on relatively slow computers when lsp-mode installs filewatchers to them.

This adds them to the ignore list, which fixed the slowdowns I was having. 

You can read more about these tools in the static-ls readme: https://github.com/josephsumabat/static-ls

Thanks!


jcs090218:
Thank you! :)

PR #4859, opened 2025-08-19
Suppress obsolete warning on latest emacs

nelhage:
Emacs HEAD has obsoleted `revert-buffer-in-progress-p`: https://github.com/emacs-mirror/emacs/commit/
1fc4bb1fead205e425f69fa803a71f4e446d48a8

We can't use the new name because it doesn't exist on older emacs versions, so just ignore the warning for now.

### DIFF
--- a/Eask
+++ b/Eask
@@ -56,4 +56,5 @@
 (add-hook 'eask-before-compile-hook
            (lambda ()
              (setq byte-compile-error-on-warn t)
+             (setq byte-compile--suppressed-warnings '((obsolete revert-buffer-in-progress-p)))
              (setq byte-compile-docstring-max-column 1000)))


nelhage:
I don't love the random warning suppressions in two different places. The best other option I came up with was sprinkling `with-suppressed-warnings` around the `defun`s, which I also don't love.

Maybe conditionally doing a `defvaralias` of an `lsp--revert-buffer-in-progress-p` to either the old or the new name would be cleaner. Any opinions?

nelhage:
Okay, I like that better after all.

jcs090218:
The link doesn't seem to point to the relevant commit. 🤔 https://github.com/emacs-mirror/emacs/commit/

jcs090218 (approved):

nelhage:
> The link doesn't seem to point to the relevant commit.

Whoops, the link got line-broken somehow. Should work now -- https://github.com/emacs-mirror/emacs/commit/1fc4bb1fead205e425f69fa803a71f4e446d48a8

jcs090218 (approved):

jcs090218:
Thank you! :)

PR #4860, opened 2025-08-21
lsp-matlab: add support for newer matlab-ts-mode and matlab ls 1.3.3

JohnC32:
This adds support for tree-sitter based matlab mode, matlab-ts-mode. This new mode is near complete on the matlab-ts-mode branch of https://github.com/mathworks/Emacs-MATLAB-Mode.

Also, this updates lsp-matlab to work with the MATLAB Language Server (matlab ls) release 1.3.3.

I also improved the description of lsp-clients-matlab-server.

I also removed: (add-to-list 'lsp-language-id-configuration '(matlab-mode . "MATLAB"))
because it is not needed. The capitalization is wrong and this is already in lsp-mode.el

### DIFF
--- a/clients/lsp-matlab.el
+++ b/clients/lsp-matlab.el
@@ -20,22 +20,26 @@
   :type '(string))
 
 (defcustom lsp-clients-matlab-server "/usr/local/apps/matlabls/out/index.js"
-  "Path to the MATLAB language server.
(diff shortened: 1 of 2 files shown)


jcs090218 (approved):

jcs090218:
Thank you! :)

PR #4868, opened 2025-09-04
Add lsp-typos

thanhvg:
Add lsp-typos and minor fix for lsp-headerline

Screenshot from 2025-09-04 16-28-41

This client is disabled by default to avoid confusion as it will run
everywhere lsp mode is on if enabled. Users could use
https://www.gnu.org/software/emacs/manual/html_node/emacs/Directory-Variables.html
to have it run per project.

Consult https://github.com/tekumara/typos-lsp and
https://github.com/crate-ci/typos for more information.

### DIFF
--- a/lsp-headerline.el
+++ b/lsp-headerline.el
@@ -306,7 +306,8 @@ PATH is the current folder to be checked."
   "Get the severity level for RANGE."
   (let ((range-severity 10))
     (mapc (-lambda ((&Diagnostic :range (&Range :start) :severity?))
-            (when (lsp-point-in-range? start range)
+            (when (and severity?
+                       (lsp-point-in-range? start range))
               (setq range-severity (min range-severity severity?))))
           (lsp--get-buffer-diagnostics))
     range-severity))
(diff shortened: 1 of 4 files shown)


lsp-headerline.el
```diff
@@ -307,7 +307,7 @@ PATH is the current folder to be checked."
   (let ((range-severity 10))
     (mapc (-lambda ((&Diagnostic :range (&Range :start) :severity?))
             (when (lsp-point-in-range? start range)
-              (setq range-severity (min range-severity severity?))))
+              (setq range-severity (min range-severity (or severity? 10)))))
```
thanhvg, line 310: `severity?` can be nil which `min` fn will not accept

jcs090218 (approved):

lsp-headerline.el
jcs090218: How about adding the `(and severity?` in the `when` clause? 🤔 I'm trying to avoid the magic number and improve the readability. :)

thanhvg: yeah it's better your way, updated.

jcs090218:
Thank you! :)

PR #4872, opened 2025-09-07
lsp: Add support for external document show (showDocument param :external?)

yuansl:
This change enhances lsp--window-show-document to handle external URLs by:
- Adding `external?` parameter support from ShowDocumentParams
- Opening external URIs in browser via browse-url if external? is t
- Maintaining existing file-based document handling

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -2226,14 +2226,16 @@ PARAMS - the data sent from WORKSPACE."
         (completing-read (concat message " ") (seq-into choices 'list) nil t)
       (lsp-log message))))
 
-(lsp-defun lsp--window-show-document ((&ShowDocumentParams :uri :selection?))
-  "Show document URI in a buffer and go to SELECTION if any."
+(lsp-defun lsp--window-show-document ((&ShowDocumentParams :uri :selection? :external?))
+  "Show document URI in a buffer or in a external browser if EXTERNAL is t, and go to SELECTION if any."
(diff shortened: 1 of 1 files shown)


yuansl:
This is useful for gopls web-based features (https://go.dev/gopls/features/web)
1) Browse assembly for a function in browser
2) Browse document for a package in browser
3) "Split package" code action 
and so on

lsp-mode.el
```diff
@@ -2226,14 +2226,18 @@ PARAMS - the data sent from WORKSPACE."
...
-        t))))
+    (if external?
+        (progn
+          (browse-url uri)
+          t)
+      (when (f-exists? path)
+	(with-current-buffer (find-file path)
+          (when selection?
+            (goto-char (lsp--position-to-point (lsp:range-start selection?))))
+          t)))))
```
jcs090218, line 2240: It looks like the indentation is a bit off here. 🤔 Can you fix this for us? Thanks!

yuansl: Hi, thanks for your review, it seems github's issue mabye, it looks just ok in my emacs, and my git diff shows that the indentation is ok (I  also just  checkout this PR locally, it seems normal)

jcs090218 (approved):

jcs090218:
Thank you for adding this!

lsp-mode.el
jcs090218: No problem! :)

PR #4871, opened 2025-09-07
Refine vue volar ls to support v3

sdvcrx:
Fix #4838

Reference:
- https://github.com/vuejs/language-tools/discussions/5456
- https://github.com/vuejs/language-tools/wiki/Neovim#configuration

PR #4873, opened 2025-09-09
Fix lsp-dependency for volar

sir4ur0n:
Move `:system` before `:npm` as the order seems to matter when loading `lsp-dependency`. Now, system-wide `vue-language-server` should be correctly detected.

See conversation in https://github.com/emacs-lsp/lsp-mode/pull/4871

Thank you @sdvcrx for all the help!


sdvcrx:
Thanks for the fix :)

PR #4876, opened 2025-09-17
Don't return invalid region from lsp-semantic-tokens--fontify.

kurnevsky:
Currently it returns 0 . 0 which is outside of the buffer, (point-min) is 1. It's better to return nil instead because such invalid region might break other packages, i.e. indent-bars.

See https://github.com/jdtsmith/indent-bars/issues/117

### DIFF
--- a/lsp-semantic-tokens.el
+++ b/lsp-semantic-tokens.el
@@ -492,7 +492,7 @@ LOUDLY will be forwarded to OLD-FONTIFY-REGION as-is."
       (funcall old-fontify-region beg-orig end-orig loudly))
      ((not (= lsp--cur-version (plist-get lsp--semantic-tokens-cache :_documentVersion)))
       ;; delay fontification until we have fresh tokens
-      '(jit-lock-bounds 0 . 0))
+      nil)
      (t
       (setq old-bounds (funcall old-fontify-region beg-orig end-orig loudly))
       ;; this is to prevent flickering when semantic token highlighting


jdtsmith:
It also needs to call the downstream function no matter what.  These functions should only _expand_, not contract (or zero out) the indicated region.

kurnevsky:
> It also needs to call the downstream function no matter what. These functions should only _expand_, not contract (or zero out) the indicated region.

But then the downstream function will colorize that region, won't it? This will result in inconsistent syntax highlighting and flickering once lsp-mode gets semantic tokens from lsp-server. We don't really want to get colors from the major mode here.

jdtsmith:
> But then the downstream function will colorize that region

Yes, but otherwise you are "hogging" jit-lock all for yourself, and other packages which need access to jit-lock updates simply won't work.  The way eglot semantic and semel are solving this is to decorate text with semantic-based text-properties _as/when available_, then, in font-lock (just font-lock-keywords as Stefan is recommending) turn those into 'face.  So if there is no semantic info yet available, font-lock wins, but once available, lsp wins.  

Honestly, I don't think lsp should need to advise `font-lock-fontify-region-function` at all; it can just use keywords.  See this discussion for more: https://lists.gnu.org/archive/html/emacs-devel/2025-10/msg00024.html.  

Note that `indent-bars` _does_ need to wrap the `f-l-f-r-f`, because (with scope highlighting), fontification depends on _point_, which is unusual.  So rather than constantly redoing all of font-lock as you scroll point around in the buffer, we arrange to update _only_ the bars (unless the rest of font-lock is actually necessary).

sebastiansturm:
sorry, I just read this today; sounds like I was using the wrong approach but if I can get rid of having to interfere with `f-l-f-r-f` I'm all for it. @jdtsmith thanks for the pointer(s), I plan to have a look at that proper approach (the one taken by eglot, semantic and semel) around Sunday

sebastiansturm:
@kurnevsky I pushed a commit to my fork at sebastiansturm/lsp-mode that uses font-lock-keywords. Like @jdtsmith said, this seems to be much simpler and work at least as well. But I haven't tested this much and have only read some of the docs on font-lock-keywords so when I find the time next week I should probably do some more poking around before I open a PR with this. But if you can test this and check whether it solves your current issue, that'd be much appreciated

kurnevsky:
@sebastiansturm just tested your changes, it solves the problem with indent-bars :)

sebastiansturm:
thanks :) I'll open a PR then (not today but I think I should get to it over the weekend)

PR #4885, opened 2025-09-30
clients/lsp-trunk.el - prevent error when using lsp on non-file buffer

JohnC32:
Protect against error when (file-name-directory filename) is nil which occurs on non-file buffer. We were hitting this when using lsp-mode with the p4.el package which when diff'ing pulls files from Perforce into non-file buffers and then activates a major mode that runs lsp.  This ends up calling lsp-trunk-check-for-init.

### DIFF
--- a/clients/lsp-trunk.el
+++ b/clients/lsp-trunk.el
@@ -46,7 +46,8 @@
   "Check if the file exists in a workspace that has a .trunk/trunk.yaml"
   (let ((dir (file-name-directory filename))
         (trunk-file ".trunk/trunk.yaml"))
-    (locate-dominating-file dir trunk-file)))
+    (when dir
+      (locate-dominating-file dir trunk-file))))
 
 (defun lsp-trunk-check-disable (command)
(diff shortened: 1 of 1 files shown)


clients/lsp-trunk.el
```diff
@@ -46,7 +46,8 @@
   "Check if the file exists in a workspace that has a .trunk/trunk.yaml"
   (let ((dir (file-name-directory filename))
         (trunk-file ".trunk/trunk.yaml"))
-    (locate-dominating-file dir trunk-file)))
+    (when dir
+      (locate-dominating-file dir trunk-file))))
```
jcs090218, line 50: Maybe use `when-let*` instead? 🤔

jcs090218: `when-let` is an obsolete macro above Emacs 31.1 (next release). 🤔

JohnC32 (commented):
thanks

jcs090218 (approved):

jcs090218:
Thank you!

PR #4888, opened 2025-10-04
fix: Ignore persist when lsp-session-file is nil

veracioux:
Fixed this for myself and thought it might be useful for others as well.
This change ensures `lsp-mode` gracefully ignores persistence when `lsp-session-file` is `nil`.


jcs090218:
Thank you!

PR #4889, opened 2025-10-06
Fix `lsp-org` compatibility with Org 9.7

### LINKED ISSUE #4300: `lsp-org` breaks with `org 9.7-pre`

### Bug description

`lsp-org` breaks with `org 9.7-pre` (not released yet). 

The issue is that in `9.7` org stops exposing some properties from org elements (see [this comment in ORG-NEWS](https://github.com/emacs-mirror/emacs/blob/1434bc97dc9b4bee6b0463c969abe905a3b2ec1f/etc/ORG-NEWS#L1163-L1190)).

At least one of the issues is https://github.com/emacs-lsp/lsp-mode/blob/2da66b3dabce4925f4c03a1243fe702d8c2f7a93/lsp-mode.el#L9490

I can make it work by replacing the offending line with (apologies for the quick'n'dirty non-idiomatic code):

```elisp
              (elem (org-element-context))
              (begin (org-element-property :begin elem))
              (end (org-element-property :end elem))
              (post-blank (org-element-property :post-blank elem))
              (language (org-element-property :language elem))
```

As far as I tested, this seems enough.
The next line still works though changing it to use `org-element-property` might future-proof it as well.

### Steps to reproduce

`test.org`

```org
#+begin_src python :tangle "test.py"
print("hello")
#+end_src
```

With point in the source block, invoke `lsp-org` leads to `progn: Wrong type argument: integer-or-marker-p, nil`.

### Expected behavior

Initialization of lsp

`lsp-pyright`

### OS

MacOS

### Error callstack

_No response_

### Anything else?

_No response_

EDIT: Fixed link to relevant emacs doc with permalink since the ORG-NEWS file has changed a lot since issue was opened

aavanian:
With `org-mode` >= 9.7, the api to access some node data has changed and requires to use a new function `org-element-property`. This fixes #4300. 

The fix is a simple one-liner but, in order to preserve the functionality for `org-mode` < 9.7:

- we define a variable to indicate which api we need to use
- we initialize that variable when `org-element` is loaded by testing it
- we branch between the two ways of accessing the needed property depending on the variable when `lsp-org` is run

Thanks to @fanshi1028 for the more elegant one-liner.

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 9.0.1
+  * Fix =lsp-org= for org >= 9.7 (see #4300)
   * Add format on save support
   * Fix beancount journal file init option
   * Add support for [[https://github.com/glehmann/earthlyls][earthlyls]]
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -9713,6 +9713,24 @@ This avoids overloading the server with many files when starting Emacs."
 (declare-function org-src-get-lang-mode "ext:org-src")
 (declare-function org-element-context "ext:org-element")
 
+(defvar lsp--org-element-use-new-api nil
+  "Whether org-element supports the new property-based API.
+This is t for org-mode 9.7 and later, nil for earlier versions.
+Determined at load time to avoid runtime performance impact.")
+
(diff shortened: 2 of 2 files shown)


jcs090218:
LGTM, Thank you!

PR #4890, opened 2025-10-06
Make `multi-root` configurable for eslint

m4xxed:
Add variable `lsp-eslint-multi-root` so that a user can configure whether they want `eslint` to be `multi-root`.


m4xxed:
Please note that I have set the default of the variable to `nil`, which is the opposite of the value it had prior. I can change this, but I thought it is more intuitive that way.

jcs090218:
Thank you!

PR #4891, opened 2025-10-08
Name change wgsl_analyzer to wgsl-analyzer

Cmdv:
Fix wgsl-analyzer installation by using correct package name

The cargo package name is `wgsl-analyzer` (with hyphens), not `wgsl_analyzer` (with underscores). This was causing installation to fail with:

`error: could not find wgsl_analyzer in https://github.com/wgsl-analyzer/wgsl-analyzer with version *`

Updated the package name, binary path, and documentation to use hyphens.

### DIFF
--- a/clients/lsp-wgsl.el
+++ b/clients/lsp-wgsl.el
@@ -33,7 +33,7 @@
   :package-version '(lsp-mode . "9.0.0"))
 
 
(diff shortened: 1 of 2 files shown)


jcs090218 (approved):

jcs090218:
Thank you!

Issue #4731, opened 2025-03-12
Improve the OCaml LSP experience


mattiasdrp:
This is an issue to keep track of the things I want to add to the OCaml-lsp client:

- [ ] Implement [switchImplIntf](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/docs/ocamllsp/switchImplIntf-spec.md) 
- [ ] Implement [inferIntf](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md)
- [ ] Implement [typedHoles](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/docs/ocamllsp/typedHoles-spec.md)
- [ ] Implement [handleTypeEnclosing](https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/docs/ocamllsp/typeEnclosing-spec.md)
- [ ] Find occurrences
- [ ] Search for values
- [ ] Better construct/destruct

PR #4899, opened 2025-10-22
OCaml client: Add inferIntf and some fixes

mattiasdrp:
See
https://github.com/ocaml/ocaml-lsp/blob/master/ocaml-lsp-server/docs/ocamllsp/inferIntf-spec.md for more informations.

Following work on https://github.com/emacs-lsp/lsp-mode/issues/4731


jcs090218:
Thank you!

Issue #4893, opened 2025-10-15
typeprof fails


costa:
### Thank you for the bug report

- [x] I am using the latest version of `lsp-mode` related packages.
- [x] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [x] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

From https://github.com/syl20bnr/spacemacs/issues/17145#issuecomment-3401660173

```
---undefined method 'map' for nil (NoMethodError)
---
---      folders = @params[:workspaceFolders].map do |folder|
---                                          ^^^^
/opt/homebrew/Cellar/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/typeprof-0.30.1/lib/typeprof/lsp/messages.rb:58:in 'TypeProf::LSP::Message::Initialize#run': undefined method 'map' for nil (NoMethodError)

      folders = @params[:workspaceFolders].map do |folder|
                                          ^^^^
	from /opt/homebrew/Cellar/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/typeprof-0.30.1/lib/typeprof/lsp/server.rb:113:in 'block in TypeProf::LSP::Server#run'
	from /opt/homebrew/Cellar/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/typeprof-0.30.1/lib/typeprof/lsp/server.rb:181:in 'TypeProf::LSP::Reader#read'
	from /opt/homebrew/Cellar/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/typeprof-0.30.1/lib/typeprof/lsp/server.rb:106:in 'TypeProf::LSP::Server#run'
	from /opt/homebrew/Cellar/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/typeprof-0.30.1/lib/typeprof/lsp/server.rb:18:in 'TypeProf::LSP::Server.start_stdio'
	from /opt/homebrew/Cellar/ruby/3.4.6/lib/ruby/gems/3.4.0/gems/typeprof-0.30.1/lib/typeprof/cli/cli.rb:115:in 'TypeProf::CLI::CLI#run_lsp'
[...]

PR #4902, opened 2025-10-27
Fixed lsp-typeprof to work with typeprof 0.30.x

buzztaiki:
fixed https://github.com/emacs-lsp/lsp-mode/issues/4893

Since `workspaceFolders` became required from [typeprof 0.30.0](https://github.com/ruby/typeprof/releases/tag/v0.30.0), I made `multi-root` non-nil. Since more than half a year has passed since the release, I think it should be fine even if customization is not possible.

This change was worked fine with typeprof 0.30.1.
Screenshot From 2025-10-27 17-01-17

NOTE: It seems that the behavior was changed in 0.30.0 so that typeprof only works when `typeprof.conf.json[c]` exists in the workspace root ([https://github.com/ruby/typeprof/blob/v0.30.0/lib/typeprof/lsp/server.rb\#L66-L76](https://github.com/ruby/typeprof/blob/v0.30.0/lib/typeprof/lsp/server.rb#L66-L76)).

### DIFF
--- a/clients/lsp-typeprof.el
+++ b/clients/lsp-typeprof.el
@@ -52,7 +52,8 @@
                    #'lsp-typeprof--build-command)
   :priority -4
   :activation-fn (lsp-activate-on "ruby")
-  :server-id 'typeprof-ls))
+  :server-id 'typeprof-ls
+  :multi-root t))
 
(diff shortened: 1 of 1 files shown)


jcs090218:
Thank you!

PR #4903, opened 2025-10-28
fix `gopls` code lenses when using `lsp-use-plists`

vadimi:
this PR fixes displaying of code lenses in go and go.mod files when using plists. It looks like `{}` value of `codeLensProvider` resolves to `null` which disables code lenses.

image

this fix is similar to what's done in https://github.com/emacs-lsp/lsp-mode/pull/4882


jcs090218:
Thank you!

PR #4907, opened 2025-11-09
lsp-rust: Append --no-run flag for debugging

Rutherther:
For me the tests were always being ran when Clicking on "Debug" button on a test in a rust file. I have gone over the logic and I am pretty confident that the condition here is wrong as it seems to me it should append --no-run only if it's not already present. This is to prevent having two --no-run flags that lead to an error `error: the argument '--no-run' cannot be used multiple times`.

Change the logic of lsp-rust-analyzer-debug to append --no-run when it's not already in cargo-args, not the other way round. The result is used only for obtaining the compiler-artifact, no need to run the tests to obtain it.


jcs090218 (approved):

jcs090218:
The logic sounds correct to me. Thank you for fixing this! :)

PR #4912, opened 2025-11-16
Use font-lock-keywords for semantic tokens

sebastiansturm:
The current implementation of semantic tokens installs font-lock overrides that apparently cause conflicts with indent-bars (see PR #4876). As jdtsmith noted, there's no need to do so, as `font-lock-keywords` provides a much simpler and more robust mechanism to achieve the same result.


ericdallo (approved):
🚀

PR #4913, opened 2025-11-18
Add tsx-mode and tsx-ts-mode indent levels to the lsp--formatting-indent-alist

SGospodinov:
Add default values for `tsx-mode` and `tsx-ts-mode` so that the formatting is consistent across both TypeScript and TSX.


jcs090218:
Thank you!

PR #4914, opened 2025-11-19
lsp-volar: improve typescript plugin configuration

fredericgiquel:
Currently, typescript plugin configuration works well when the language server is installed by `lsp-install-server` but not with other installation methods (elsewhere with `npm`, `nix` package, ...) because the location used in plugin configuration is incorrect.

So this PR:
- allows the user to specify a manual value for the location
- improve automatic detection by adding a second method that works if vue-language-server is installed elsewhere by npm and made available in PATH by a symlink


jcs090218 (approved):

jcs090218:
Thank you!

PR #4920, opened 2025-11-22
bugfix: lsp-rust-analyzer-cargo-cfgs default value 

Cbeck527:
Resolves this error when working in a Rust project:

```
LSP :: invalid config value:
/cargo/cfgs: invalid type: map, expected a string;
```

The comma causes the value of `lsp-rust-analyzer-cargo-cfgs` to be `["debug_assertions" (\, "miri")]`


jcs090218:
Thank you!

PR #4926, opened 2025-11-29
fix: remove malformed and unnecessary Package-Requires

chosen-arisen:
Remove malformed `Package-Requires' from lsp-copilot, lsp-pls clients, to prevent parsing error while `lsp-mode' via use-package-vc. Besides being malformed, they are quite unnecessary too, so one such comment was removed from lsp-volar client too.


jcs090218:
Thank you!

PR #4929, opened 2025-12-03
lsp-javascript: add tsgo server

### LINKED ISSUE #4854: Add support for tsgo (typescript native preview)
Is your feature related or already mentioned on [the wishlist](https://github.com/emacs-lsp/lsp-mode/discussions/2892)? I don't see it there

See https://github.com/microsoft/typescript-go, https://marketplace.visualstudio.com/items?itemName=TypeScriptTeam.native-preview

This is a typescript reimplementation in go which is a lot faster.

So far here's what I've tried:

```el
(lsp-register-client (make-lsp-client
                      :new-connection (lsp-stdio-connection '("tsgo" "--lsp" "-stdio"))
                      :activation-fn 'lsp-typescript-javascript-tsx-jsx-activate-p
                      :server-id 'tsgo-ls))
```

It works right until I save the buffer. Then I get a couple of crashes. Not sure if it's a misconfiguration or if it's just that unstable in the current build.

kiennq:
Fix #4854

### DIFF
--- a/clients/lsp-javascript.el
+++ b/clients/lsp-javascript.el
@@ -72,6 +72,8 @@
                                     (warn (concat "The javascript-typescript-langserver (jsts-ls) is unmaintained; "
                                                   "it is recommended to use ts-ls or deno-ls instead.")))))
 
+
+
 (defgroup lsp-typescript nil
   "LSP support for TypeScript, using Theia/Typefox's TypeScript Language Server."
   :group 'lsp-mode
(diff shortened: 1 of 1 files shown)


clients/lsp-javascript.el
```diff
@@ -72,6 +72,8 @@
                                     (warn (concat "The javascript-typescript-langserver (jsts-ls) is unmaintained; "
                                                   "it is recommended to use ts-ls or deno-ls instead.")))))
 
+
+
```
jcs090218: What's wrong with all these extra newlines? 🤔

kiennq: They are page-break characters `^L`, not new line.
I've separated each server into a page

PR #4930, opened 2025-12-07
lsp-buf: change buf CLI args

suzuki:
## WHAT

Change args of `buf`

## WHY

- The lsp command in buf has graduated from beta
    - https://github.com/bufbuild/buf/releases/tag/v1.59.0
        > Promote buf beta lsp to buf lsp serve. Command buf beta lsp is now deprecated.
- No longer times out by default
    - https://github.com/bufbuild/buf/releases/tag/v1.60.0
        > Update default value of --timeout flag to 0, which results in no timeout by default.
- Use logformat default (text)
    - https://github.com/emacs-lsp/lsp-mode/pull/4739#issuecomment-2750505706

### DIFF
--- a/clients/lsp-bufls.el
+++ b/clients/lsp-bufls.el
@@ -71,7 +71,7 @@
   :link '(url-lint "https://github.com/bufbuild/buf")
   :package-version '(lsp-mode . "9.0.0"))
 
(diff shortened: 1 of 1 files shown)


jcs090218:
Thank you!

PR #4935, opened 2025-12-12
lsp-mode: Use cl-loop to stop allocating lambdas during hot file loops

kevinji:
In a large project with a ton of files, when I changed between branches that had a lot of changed files, `lsp--file-process-event` would freeze my editor for one full minute due to CPU and memory being pegged&mdash;I saw GC using 60% of CPU and the `string-match` inside that function using 40% of CPU.

After rewriting this function to avoid allocating the intermediate lambdas, the time my editor is spent frozen is now reduced to about two seconds.

I also updated `lsp--send-will-rename-files-p` to be rewritten in the same way given that it had a very similar `-any?` structure.


jcs090218 (approved):

jcs090218:
LGTM. Thank you!

PR #4938, opened 2025-12-15
add add L? to regex for links in ui-doc

andrew-christianson:
Some LSP servers report line-level links to files with github's convention: `#L$LINE,$COL`. which does not match the existing regex `#\d+[,#]\d+`. Instead we modify the rx expression to correspond to `#L?\d+[,#]\d+`


jcs090218 (approved):

jcs090218:
Sounds good to me! Thank you!

PR #4941, opened 2025-12-21
fix: handle empty capability options in plist mode

suzuki:
## Summary

Fix `lsp--capability` to correctly detect capabilities when `LSP_USE_PLISTS` is enabled and the server returns empty option objects like `"definitionProvider": {}`.

## Problem

When `LSP_USE_PLISTS=true`, Emacs's `json-parse-string` parses empty JSON objects `{}` as `nil`:

```elisp
(json-parse-string "{\"definitionProvider\": {}}" :object-type 'plist)
;; => (:definitionProvider nil)
```

This causes `lsp--capability` to incorrectly report that the capability is not supported, because it only checks if the value is truthy.

### Affected servers

Any LSP server that returns empty option objects, for example:

- **buf lsp** (`buf lsp serve`) returns:
  ```json
  {
    "definitionProvider": {},
    "typeDefinitionProvider": {},
    "referencesProvider": {}
  }
  ```

With the current implementation, "Go to Definition", "Find References", etc. are broken when using these servers with `LSP_USE_PLISTS=true`.

## Solution

Use `lsp-member?` to check if the capability key **exists** in the server capabilities, rather than just checking if the value is truthy:

```elisp
;; Before
(lsp-get (or capabilities (lsp--server-capabilities)) cap)

;; After
(let ((caps (or capabilities (lsp--server-capabilities))))
  (when (lsp-member? caps cap)
    (or (lsp-get caps cap) t)))
```

This correctly distinguishes between:
- Key exists with empty value (`{}` → `nil` in plist) → capability **supported** ✓
- Key does not exist → capability **not supported** ✗

## Test plan

- [x] Added unit tests for `lsp--capability` covering:
  - Empty object capabilities (plist mode)
  - Boolean capabilities
  - Structured capabilities (e.g., `completionProvider` with options)
  - Non-existent capabilities
  - String key conversion

### DIFF
--- a/test/lsp-mode-test.el
+++ b/test/lsp-mode-test.el
@@ -202,9 +202,9 @@ This tests the fix for empty capability objects like DefinitionOptions {}."
                           (list :definitionProvider nil
                                 :hoverProvider t
                                 :completionProvider (list :triggerCharacters ["." ":"]))
-                        ;; hash-table mode
+                        ;; hash-table mode: empty objects are empty hash-tables, null is nil
                         (let ((ht (make-hash-table :test 'equal)))
-                          (puthash "definitionProvider" nil ht)
+                          (puthash "definitionProvider" (make-hash-table :test 'equal) ht)
                           (puthash "hoverProvider" t ht)
                           (let ((completion-ht (make-hash-table :test 'equal)))
                             (puthash "triggerCharacters" ["." ":"] completion-ht)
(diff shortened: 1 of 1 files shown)


kiennq:
@suzuki, thanks for your changes, I hope it would resolve a large array of bugs we have regarding to deserialize json in plist.
Currently with plist deserialize we have option to set `:false-object` and `:null-object`, however empty list is not there.

Do you think it's possible to properly check for empty list using combination of `lsp-get` and `lsp-member`? So we can change the `lsp-empty?` and `lsp-get` to check both conditions?
Also, we should probably move all of the `:false-object` and `:null-object` to different values other than `nil` so we can properly distinguish between all of them

suzuki:
@kiennq Sorry for the late reply. I've introduced `lsp-null?` for null checking. Does this change align with your thoughts?

PR #4942, opened 2025-12-21
just-lsp: Add support for Just files

Dietr1ch:
Add support for [just](https://github.com/casey/just/) files through [`just-lsp`](https://github.com/terror/just-lsp).


jcs090218:
Thank you!

PR #4944, opened 2025-12-22
lsp-rust: introduce variables for configs needed for rustc development

dfireBird:
I added missing custom variables referenced in the eglot configuration in rustc repo (https://github.com/rust-lang/rust/blob/main/src/etc/rust_analyzer_eglot.el) for use in `.dir-locals.el`

I also changed the custom variable for build scripts override command since `lsp-rust-analyzer-cargo-override-command` and `lsp-rust-analyzer-cargo-build-scripts-override-command` (created by me) needs to have different values as per the file.


jcs090218:
Thank you!

PR #4956, opened 2026-01-05
feat: add source.fixAll code action and on-save auto-fix support

### LINKED ISSUE #1842: Support for source.fixAll code action

takeokunn:
## Summary

Add support for `source.fixAll` code action as requested in #1842.

This PR introduces:

- `lsp-fix-all`: Interactive command to execute `source.fixAll` code action for the entire buffer
- `lsp-fix-all-on-save`: Custom variable to enable auto-fix on save
- `lsp-fix-all-on-save-list`: Mode filter list (similar to `lsp-format-buffer-on-save-list`)
- `lsp-execute-code-action-by-kind-buffer-wide`: Helper function that passes entire buffer range and all diagnostics to the language server
- `lsp-code-actions-buffer-wide`: Retrieve code actions for the entire buffer

## Motivation

The `source.fixAll` code action is a standard LSP feature (since 3.15) that applies unambiguously safe fixes. Many language servers support this, including:

- gopls (Go)
- ESLint
- ruff (Python)
- typescript-language-server

Previously, `lsp-execute-code-action-by-kind` only considered diagnostics at the current position, which made `source.fixAll` ineffective for buffer-wide fixes.

## Changes

- Modified `lsp--text-document-code-action-params` to accept optional `buffer-wide` parameter
- Added new functions for buffer-wide code action execution
- Registered `lsp--fix-all-before-save` hook (runs before format hook)

Closes #1842

## Test Case

`M-x lsp-fix-all`

```go
package main

func main() {
    m := map[string]int{"a": 1}

    // gopls source.fixAll will simplify this:
    for _ = range m {  // → for range m
        println("test")
    }
}
```

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,6 +1,9 @@
 * Changelog
 ** Unreleased 9.0.1
   * Add ~lsp-fix-all~ command and ~lsp-fix-all-on-save~ option for ~source.fixAll~ code action support (see [[https://github.com/emacs-lsp/lsp-mode/issues/1842][#1842]])
+  * Add notification handlers for Deno language server (~deno/didRefreshDenoConfigurationTree~, ~deno/didChangeDenoConfiguration~, ~deno/didUpgradeCheck~).
+  * Fix "Unknown notification: semgrep/rulesRefreshed" warning in lsp-semgrep client.
+  * Add ~lsp-install-server~ support for Deno LSP via npm.
   * Add diagnostic and formatter custom settings for [[https://github.com/rcjsuen/dockerfile-language-server-nodejs][Dockerfile language server]].
   * Fix "Eager macro-expansion failure" errors with lsp-interface pcase patterns (see #4723)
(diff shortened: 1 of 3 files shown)


takeokunn:
I'd like to contribute to lsp-mode, so if possible, could I get collaborator permissions?

jcs090218:
I've added you as the collaborator! Thank you for contributing to lsp-mode! :D

takeokunn:
@jcs090218 
Thanks invitation!!!

jcs090218 (approved):

jcs090218:
Thank you for adding this!

PR #4962, opened 2026-01-08
Advice vc-git-rename-file with file renaming operations

remyferre:
`vc-rename-file` may call `rename-file`, but it does not when the underlying backend already renames the file by itself (this is the case for git). This means that workspace/willRenameFiles will not be sent because the advice of rename-file is never called. This PR makes the behavior consistent in git repo between `vc-rename-file` and `rename-file`  so that file operations are used.


jcs090218:
I think this make sense. Merged this now! Thank you!

PR #4966, opened 2026-01-12
docs: fix incorrect full name

g-plane:
Sorry I'm not clear about the `full-name` field in `lsp-clients.json` and it causes the docs page can't opened correctly. This PR fixes this.


jcs090218:
Thank you!

PR #4982, opened 2026-01-30
Remove outdated Ruff option and add new lint-related options

dzhu:
See https://docs.astral.sh/ruff/editors/migration/ for changes in the options between ruff-lsp and the new native server.

Sending empty lists for `select` and `extend-select` seems to leave the default rules enabled, so there's no need for logic to avoid sending the attributes in that case.


jcs090218 (approved):

jcs090218:
Thank you!

dzhu:
You're welcome, and thanks for responding so fast!

PR #4983, opened 2026-01-30
Remove url anchors in lsp--uri-to-path

### LINKED ISSUE #4976: `lsp--uri-to-path` does not handle uri anchors

### Bug description

I find that the function `lsp--uri-to-path` does not handle uris with anchors correctly, treating the anchor as part of the path rather than as a separate concept.

For instance, we have:
```
Eval: (lsp--uri-to-path "file:/my/file.html#anchor")
"/my/file.html#anchor"
```

This happens to me when using the [Haskell Language Server](https://haskell-language-server.readthedocs.io/en/stable/), which produces documentation popups with links in the form of 

```
[Documentation](file:///nix/store/.../Module.Main.html#v:main)
```

.. which are in turn handled by the function call sequence of `lsp--help-open-link -> lsp--document-link-handle-target -> (find-file (lsp--uri-to-path uri))`.

with the end result being emacs trying to open the file e.g. `/my/file.html#anchor` instead of `/my/file.html`.

### Steps to reproduce

```
Eval: (lsp--uri-to-path "file:/my/file.html#anchor")
"/my/file.html#anchor"
```

### Expected behavior

```
Eval: (lsp--uri-to-path "file:/my/file.html#anchor")
"/my/file.html"
```

[Haskell Language Server](https://haskell-language-server.readthedocs.io/en/stable/)

### OS

Linux

### Error callstack

```emacs-lisp

```

### Anything else?

_No response_

themkat:
Originally I thought about removing the "target" from the URL completely, but I found one test case:
```emacs-lisp
                 (should (equal (lsp--uri-to-path "file:///home/nim-%23devel")
                                "/home/nim-#devel"))
```

As I understand it, we want `#` to be in the resulting URL if is was encoded, but remove it if it was explicit. The big assumption is therefore that the uris we get will follow the rules of standard http paths. (I've never seen a address URL encode the anchor, only `#` symbols existing in text). I've added a test that illustrate my point and how I imagine language servers work with paths containing `#`. (though I have NEVER seen it IRL. Just assume from the tests and current code that it is wanted. Also read that `#` symbols in filenames are allowed, so we can probably assume that they are url encoded).

Fixes #4976

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1960,7 +1960,6 @@ On other systems, returns path without change."
   "Convert URI to a file path."
   (let* ((url (url-generic-parse-url uri))
          (type (url-type url))
-         (target (url-target url))
          (file
           (decode-coding-string (url-unhex-string (url-filename url))
                                         (or locale-coding-system 'utf-8)))


jcs090218:
Thank you for fixing this!

PR #4984, opened 2026-01-31
fix: Clear diagnostics after workspace edits to prevent stale line numbers

### LINKED ISSUE #3888: Flycheck assumes incorrect line numbers after goimports

### Bug description

After saving a Go file, the imports are very often automatically corrected by using a tool such as goimports.  This may change the line numbers for a large fraction of the source code file. Flycheck seems to mark syntax errors based on the _old_ line numbers, leading to a garbled display.

### Steps to reproduce

Use the following for `.emacs`:

```lisp
(setq flycheck-check-syntax-automatically '(save))
(require 'lsp-mode)
(add-hook 'go-mode-hook #'lsp-deferred)
(defun lsp-go-install-save-hooks ()
  (add-hook 'before-save-hook #'lsp-organize-imports t t))
(add-hook 'go-mode-hook #'lsp-go-install-save-hooks)
(setq lsp-idle-delay 10)
```
and then visit the following Go file:
```go
package main

import "fmt"

func main() {
	fmt.Println("Hello, World!")
}
```
Now, deliberately change the `fmt.Println` into `fsmt.Println` and save the file. goimports will then remove the `import "fmt"`, which changes the line numbers. For ten seconds, incorrect regions in the source code are marked as having syntax errors.

### Expected behavior

The appearance after 10 seconds, i.e. with the `fsmt` highlighted as being incorrect, should be the case immediately after the saving.

gopls

### OS

Linux

### Error callstack

_No response_

### Anything else?

_No response_

takeokunn:
## Summary

Clear diagnostics after workspace edits (like organize-imports, rename, code actions) to prevent stale diagnostics from appearing at wrong line numbers when these operations shift line positions.

Fixes #3888

## Changes

- Add `lsp--diagnostic-clean-after-edit` function hooked to `lsp-after-apply-edits-hook`
- Clears diagnostics for current buffer when workspace edits are applied
- Triggers `lsp-diagnostics-updated-hook` to update UI (Flycheck/Flymake)

## Design Decision

This is implemented unconditionally (no defcustom toggle) because:
1. Workspace edits are explicit user actions that definitely shift lines
2. There is no legitimate use case for keeping stale diagnostics
3. The server will immediately send correct diagnostics afterwards

## Test plan

- [x] Added `lsp-mock-server-clears-diags-after-workspace-edit` test
- [x] Manual testing with Go + gopls + organize-imports

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -5074,20 +5074,6 @@ movements may have changed the position")
                               buffer))
   (lsp--idle-reschedule buffer))
 
-(defun lsp--diagnostic-clean-after-edit (&rest _)
-  "Clear diagnostics for the current buffer after workspace edits.
-This prevents stale diagnostics from appearing at wrong line numbers
-when workspace edits shift line positions (see issue #3888)."
-  (when-let* ((file-path (buffer-file-name))
-              (workspaces (lsp-workspaces)))
-    (let ((path (lsp--fix-path-casing file-path)))
-      (dolist (workspace workspaces)
-        (-let [diagnostics (lsp--workspace-diagnostics workspace)]
-          (remhash path diagnostics))))
-    (run-hooks 'lsp-diagnostics-updated-hook)))
-
-(add-hook 'lsp-after-apply-edits-hook #'lsp--diagnostic-clean-after-edit)
-
 
 (defcustom lsp-trim-trailing-whitespace t
   "Trim trailing whitespace on a line."
--- a/lsp-diagnostics.el
+++ b/lsp-diagnostics.el
@@ -367,6 +367,20 @@ See https://github.com/emacs-lsp/lsp-mode."
                                 (when lsp-auto-configure
                                   (lsp-diagnostics--enable))))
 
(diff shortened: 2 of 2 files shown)


lsp-mode.el
```diff
@@ -5074,6 +5074,20 @@ movements may have changed the position")
...
+when workspace edits shift line positions (see issue #3888)."
+  (when-let* ((file-path (buffer-file-name))
+              (workspaces (lsp-workspaces)))
+    (let ((path (lsp--fix-path-casing file-path)))
+      (dolist (workspace workspaces)
+        (-let [diagnostics (lsp--workspace-diagnostics workspace)]
+          (remhash path diagnostics))))
+    (run-hooks 'lsp-diagnostics-updated-hook)))
+
+(add-hook 'lsp-after-apply-edits-hook #'lsp--diagnostic-clean-after-edit)
```
jcs090218: Maybe we should move these to `lsp-diagnostics.el` file instead? 🤔

jcs090218 (approved):

jcs090218:
Thank you for the fix!

PR #4986, opened 2026-02-05
Check for an empty ConfigurationItem list

cannero:
If no ConfigurationItems are passed in the ‘workspace/configuration’ request
by the server, send the settings for all clients instead.

This change is needed to use the grammar checker `Harper`.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7053,24 +7053,31 @@ If nil, and `lsp-debounce-full-sync-notifications' is non-nil,
 
 (lsp-defun lsp--build-workspace-configuration-response ((&ConfigurationParams :items))
   "Get section configuration.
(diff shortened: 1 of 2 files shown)


agzam:
Nice. Thank you for making this happen!

lsp-mode.el
```diff
@@ -8988,6 +8990,28 @@ TBL - a hash table, PATHS is the path to the nested VALUE."
              lsp-client-settings)
     ret))
 
+(defun lsp-configuration-total ()
```
jcs090218: I think the function name `lsp-configuration-total` is a bit vague. Maybe change it to something else? 🤔 

Make sure the function is private instead of public by adding one extra `-` after the package prefix `lsp-`.

jcs090218:
Thank you for adding this! :D

PR #4987, opened 2026-02-06
feat: Add support for fortitude, Fortran linter

ZedThree:
See https://github.com/PlasmaFAIR/fortitude

- feat: Add lsp-fortitude.el, based on lsp-ruff.el

- docs: Distinguish between Fortran clients

---

The support for LSP server in fortitude was added in the latest version (0.8.0). Is there something we need to do here to gate on/check the version of `fortitude`? Or should I just make a note of it in the docs for the `lsp-fortitude` group?

### DIFF
--- a/clients/lsp-fortitude.el
+++ b/clients/lsp-fortitude.el
@@ -1,0 +1,77 @@
+;;; lsp-fortitude.el --- fortitude lsp support             -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2025 Peter Hill
+;;
+;; Author: Peter Hill <peter.hill@york.ac.uk>
+;; Keywords: language tools
+;;
+;; This program is free software; you can redistribute it and/or modify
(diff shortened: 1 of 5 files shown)


clients/lsp-fortitude.el
```diff
@@ -0,0 +1,77 @@
+;;; lsp-fortitude.el --- fortitude lsp support             -*- lexical-binding: t; -*-
+
+;; Copyright (C) 2025 Peter Hill
```
jcs090218: Should be 2026! :D

jcs090218:
> The support for LSP server in fortitude was added in the latest version (0.8.0). Is there something we need to do here to gate on/check the version of fortitude? Or should I just make a note of it in the docs for the lsp-fortitude group?

If it works with the latest release, I think it should be fine for now! :)

jcs090218:
Merged! Thank you for adding this! :D

Issue #4445, opened 2024-04-29
Zig's ZLS download URLs have changed


nordlow:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

lsp-zig.el needs to be adjusted to download the now uncompressed binary from https://github.com/zigtools/zls/wiki/Installation#latest-nightly.

### Steps to reproduce

Open a .zig file and the error occurrs.

### Expected behavior

The output in Message become

LSP :: Download zls started.
LSP :: Starting to download https://github.com/zigtools/zls/releases/latest/download/zls-x86_64-linux.tar.gz to ~/.emacs.d/.cache/lsp/zig/temp.tar.gz...
Contacting host: github.com:443
Wrote ~/.emacs.d/.cache/lsp/zig/temp.tar.gz
LSP :: Finished downloading ~/.emacs.d/.cache/lsp/zig/temp.tar.gz...
LSP :: Decompressing ~/.emacs.d/.cache/lsp/zig/temp.tar.gz...
tar: This does not look like a tar archive

gzip: stdin: not in gzip format
tar: Child returned status 1
tar: Error is not recoverable: exiting now
LSP :: Decompressed ~/.emacs.d/.cache/lsp/zig/temp...
LSP :: Server zls downloaded, auto-starting in 1 buffers.
Quit [2 times]
Displaying messages...
Mark set

### Which Language Server did you use?

I used lsp-zig.

### OS

Linux

jcs090218:
Opened the issue in https://github.com/zigtools/zls/issues/1879.

Techatrix:
This issue has been unblocked. See https://github.com/zigtools/zls/issues/1879#issuecomment-2302889190

jcs090218:
Fixed in #4989.

PR #4989, opened 2026-02-10
fix: zls download support zigtools.org

maker2413:
This PR should resolve the issue: https://github.com/emacs-lsp/lsp-mode/issues/4445.

It adds a new function to get the latest version tag from the GitHub api for ZLS and then formats the download URL for ZLS to use the "new" zigtools.org download page. If any of these changes are improperly formatted or need to be done differently please let me know and I'd be glad to fix it!


jcs090218:
LGTM! Thank you for fixing this!

PR #4996, opened 2026-02-14
Remove mouse highlight in log buffer

guillaumebrunerie:
I find the mouse highlight in the log buffer particularly annoying as it blinks on and off while scrolling in the buffer. I also do not see the point of it, clicking on log messages doesn’t seem to do anything, unless I’m missing something. So I would suggest to remove it.
The `read-only` property seems useless as well as the whole buffer is read only anyway.


guillaumebrunerie:
@jcs090218  Gentle ping on this PR, I still find the mouse highlighting in the log buffer quite annoying, so I'd be happy to address any concerns with this change. Thanks!

Issue #4998, opened 2026-02-24
lsp-zig: path defcustoms default to "" instead of nil, breaking ZLS auto discovery


eval-and-prosper:
### Thank you for the bug report

- [x] I am using the latest version of `lsp-mode` related packages.
- [x] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [ ] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

All path defcustoms in lsp-zig.el default to "". Because "" is truthy in Elisp, lsp-mode's configuration builder sends these empty strings to ZLS via workspace/configuration responses. ZLS treats "" as an explicit invalid path rather than null, which prevents auto discovery.

The worst offender is global_cache_path, ZLS cannot locate the cache build runner output, so the entire module graph is missing. Named module imports (@import("module_name")) fail to resolve, while relative file imports (@import("file.zig")) still work.

These variables are all defaulted to "" and should be set to nil as ZLS expects null for auto discovery but interprets the "" strings as empty paths.

lsp-zig-builtin-path
lsp-zig-zig-lib-path
lsp-zig-zig-exe-path
lsp-zig-build-runner-path
lsp-zig-global-cache-path
lsp-zig-build-runner-global-cache-path

### Steps to reproduce

Open any Zig project that uses named modules defined in build.zig (e.g. @import("my_module")) with a default lsp-zig configuration — @import will show as unresolved because ZLS never runs the build runner. Setting the affected defcustoms to nil restores auto-discovery and resolves the imports.

### Expected behavior

Path settings left at their default should be omitted from workspace/configuration responses so ZLS auto discovers them.

### Which Language Server did you use?

ZLS 15.1

### OS

MacOS

### Error callstack

```emacs-lisp

```

maker2413:
Thank you for finding this! I was just trying to figure out how to fix this issue and your proposed change fixes it. I will get a PR out shortly.

jcs090218:
Should be fixed in #5000.

PR #5000, opened 2026-03-02
fix: changing lsp-zig path defcustoms default to nil

maker2413:
This PR should resolve the issue: https://github.com/emacs-lsp/lsp-mode/issues/4998

The linked issue already describes what needs to be fixed in that the following variables need to be defaulted to `nil` to satisfy the assumptions of ZLS:
- lsp-zig-builtin-path
- lsp-zig-zig-lib-path
- lsp-zig-zig-exe-path
- lsp-zig-build-runner-path
- lsp-zig-global-cache-path
- lsp-zig-build-runner-global-cache-path

The changes in this PR just default this defcustoms to nil and all the type to be either `nil` or `string` as to not break the existing logic of setting the types to `string`

### DIFF
--- a/clients/lsp-zig.el
+++ b/clients/lsp-zig.el
@@ -165,42 +165,42 @@ If no extra argument is given `record_session_path` is used as the default path.
   :group 'lsp-zig
   :type 'string)
 
-(defcustom lsp-zig-builtin-path ""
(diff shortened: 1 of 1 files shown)


jcs090218:
Thank you!

Issue #3006, opened 2021-07-16
Can't install LSP servers from raw.githubusercontent.com


liar666:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command
```bash
emacs -q -l lsp-start-plain.el
```
where `lsp-start-plain.el` can be downloaded [here](https://github.com/emacs-lsp/lsp-mode/blob/master/scripts/lsp-start-plain.el).

### Bug description

I can't install LSP servers from `raw.githubusercontent.com` (e.g. `lsp-java` & `lsp-metals`). I get error:

"LSP :: Server metals install process failed with the following error message: (error Attempt to accept output from process raw.githubusercontent.com locked to thread #<thread 0x9fe8a0>)."

### Steps to reproduce

In .emacs:
`(add-hook 'scala-mode-hook #'lsp)`

In Emacs (`scala-mode` + `lsp-metals` + `lsp` installed):
- M-x list-packages
- U
- x

In terminal:
- `rm -fr ~/.emacs.d/{.cache/*, auto-save-list/,semanticdb/*}`
- emacs  test.scala

In Emacs:
- M-x lsp-install-server
- metals

> LSP :: Server metals install process failed with the following error message: (error Attempt to accept output from process raw.githubusercontent.com locked to thread #<thread 0x9fe8a0>).

### Expected behavior

metals should install flawlessly.

### Which Language Server did you use?

lsp-metals / lsp-java

### OS

Linux

### Error callstack

```emacs-lisp
Command "zls" is not present on the path.
Command "bundle exec steep langserver --log-level warn" is not present on the path.
Command "sqls" is not present on the path.
Command "hdl_checker --lsp" is not present on the path.
Command "svlangserver" is not present on the path.
Command "vala-language-server" is not present on the path.
Command "vls" is not present on the path.
Command "terraform-lsp" is not present on the path.
Command "~/Softs/TeXLab/texlab" is present on the path.
Command "digestif" is not present on the path.
Command "srb typecheck --lsp --disable-watchman" is not present on the path.
Command "solargraph stdio" is not present on the path.
Command "rust-analyzer" is not present on the path.
Command "rls" is not present on the path.
[...]

PR #5022, opened 2026-03-23
refactor: fix macOS SIGABRT crash in lsp-download-install by replacing url-copy-file thread with make-process (curl)

roryhow:
## Problem

On macOS (Emacs 30+, ARM), installing a language server via `lsp-install-server` crashes Emacs with `EXC_CRASH (SIGABRT)`. The crash is a thread-safety violation in the NS/Cocoa backend.

`lsp-download-install` wraps the download in `make-thread`. Inside that thread, `url-copy-file` calls `url-retrieve-synchronously`, which pumps the macOS event loop via `ns_select_1`. That function contains an assertion that it only runs on the main thread — violated here, causing `emacs_abort`.

The crash stack:
```
lsp-download-install
  └─ make-thread
       └─ url-copy-file
            └─ url-retrieve-synchronously
                 └─ accept-process-output
                      └─ wait_reading_process_output
                           └─ ns_select_1        [src/nsterm.m]
                                └─ unblock_input  ← eassert (main_thread_p) → ABORT
```

This is also the root cause of the softer error seen on Linux in #3006 ("Attempt to accept output from process ... locked to thread"), where the NS assertion doesn't fire but the thread-safety violation still produces a Lisp error.

## Fix

Replace `url-copy-file` (and the `make-thread` wrapping it) with `make-process` running `curl`, plus a `:sentinel` callback. The sentinel fires on the main thread via the normal Emacs event loop — no thread-safety issues on any platform.

This is the same pattern used by established Emacs HTTP libraries:
- [`plz.el`](https://github.com/alphapapa/plz.el) — `make-process` + `:sentinel`
- [`request.el`](https://github.com/tkf/emacs-request) — `start-process` + `set-process-sentinel`

Decompression (`:gzip`, `:zip`, `:targz`, `:tarxz`) remains in a `make-thread` — those steps use `call-process` internally and do not touch the NS event loop, so they are thread-safe.

## Changes

- `lsp-mode.el`: add private `lsp--download-file` helper; refactor `lsp-download-install` to use it
- `test/lsp-mode-test.el`: add 5 ERT tests covering missing curl, correct invocation, success, failure, and all decompress variants
- `CHANGELOG.org`: entry in Unreleased

## Notes

`curl` is a hard requirement for the download path. It ships with macOS and is present on virtually all Linux systems. A clear error is raised if it is not found. Opinions or alternatives to this approach are welcome.


jcs090218 (approved):
I haven't done the manual test yet, but the code looks good to me!

Can someone give it a try on macOS? 🤔

roryhow:
Thanks for the review, @jcs090218! For what it's worth - I've personally been running this on my Mac since creating this PR (as it was a blocker for some LSP servers I wanted to download).

Let me know if there's something I can do in terms of getting this tested by someone else, though.

jcs090218 (approved):

Issue #4516, opened 2024-07-30
lsp-zig-enable-build-on-save is not working


iskyd:
### Thank you for the bug report

- [X] I am using the latest version of `lsp-mode` related packages.
- [X] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [X] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

using lsp-zig-enable-build-on-save dont report compile errors.

I'm using the following configuration
```
(use-package lsp-mode
  :ensure t
  :init
  (setq lsp-print-io t)
  (setq lsp-zig-enable-build-on-save t)
  (setq lsp-zig-build-on-save-step "install")
  :hook (;; replace XXX-mode with concrete major-mode(e. g. python-mode)
         (zig-mode . lsp-deferred)
:commands lsp-deferred)

(use-package zig-mode
  :ensure t
  (add-hook 'zig-mode-hook
            (lambda ()
              (setq compile-command "zig build"))
            )
  )
```

From lsp-log I see that the configuration sent seems ok, it sent enable_build_on_save and the correct build_on_save_step.

### Steps to reproduce

Setup a new zig project with zig init, go to src/main.zig and change anything in a way that the program will not compile.
Save the file, no error will be reported.

### Expected behavior

All compile error should be reported

### Which Language Server did you use?

lsp-zig

### OS

Linux

### Error callstack

```emacs-lisp
[Trace - 07:28:01 ] Received request 'workspace/configuration - (i_haz_configuration).
Params: {
  "items": [
    {
      "section": "zls.enable_snippets"
    },
    {
      "section": "zls.enable_argument_placeholders"
    },
    {
      "section": "zls.enable_build_on_save"
    },
    {
      "section": "zls.build_on_save_step"
    },
    {
      "section": "zls.enable_autofix"
    },
    {
      "section": "zls.semantic_tokens"
    },
    {
      "section": "zls.enable_inlay_hints"
    },
    {
      "section": "zls.inlay_hints_show_variable_type_hints"
    },
    {
      "section": "zls.inlay_hints_show_struct_literal_field_type"
    },
    {
      "section": "zls.inlay_hints_show_parameter_name"
    },
    {
      "section": "zls.inlay_hints_show_builtin"
    },
    {
[...]

PR #5028, opened 2026-03-26
ZLS build-on-save diagnostics fix

mwac-dev:
Previously workspaceFolders was only included when the server was registered as multi-root. Servers like ZLS depend on it to locate build files and enable features like build-on-save which silently failed without it.

Falls back to the project root when no explicit workspace folders are registered in the session. This matches behaviour similar to other editors which send workspaceFolders unconditionally, which is what allows ZLS to function as intended.

This addresses #4516


jcs090218 (approved):

jcs090218:
Thank you!

PR #5046, opened 2026-04-21
Add lsp-intelephense-phpdoc-use-fully-qualified-names

sergioalcaraz:
Add custom option for using fully qualified names (FQCN) in PHPDoc.

This allow to complete FQCN in PHPDoc `@var, @return, @param` declaration

 for example

When nil

`/** @var Person $person */`

When true

`/** @var \App\Model\Entity\Person $person */`

See also [phpdoc](https://docs.phpdoc.org/guide/references/phpdoc/types.html)

### DIFF
--- a/clients/lsp-php.el
+++ b/clients/lsp-php.el
@@ -264,6 +264,14 @@ language server."
   :group 'lsp-intelephense
   :package-version '(lsp-mode . "6.1"))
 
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218 (approved):

jcs090218:
Thank you!

PR #5048, opened 2026-04-24
fix(lsp): guard nil :location in symbols-hierarchy breadcrumb lookup (#5047)

### LINKED ISSUE #5047: lsp--on-idle: (wrong-type-argument hash-table-p nil) in breadcrumb when pylsp returns SymbolInformation entries without location field

### Bug description

`Error running timer 'lsp--on-idle': (wrong-type-argument hash-table-p nil)` fires repeatedly whenever the cursor is idle in a Python buffer with `lsp-headerline-breadcrumb-mode` enabled and the `symbols` segment active.

`lsp--symbols-informations->document-symbols-hierarchy` (lsp-mode.el ~line 6678) uses a `-keep` lambda that destructures `(&SymbolInformation :location (&Location :range))`. When pylsp returns `SymbolInformation` entries where the `location` field is absent/nil (which appears to happen with some duplicate symbol entries), the `(&Location :range)` destructuring attempts `gethash "range" nil`, crashing with `(wrong-type-argument hash-table-p nil)`.

### Steps to reproduce

1. Open a Python file with multiple symbols where pylsp returns some `SymbolInformation` entries without a `location` field
2. Have `lsp-headerline-breadcrumb-segments` include `symbols`
3. Observe the error firing on every idle tick

### Expected behavior

Not to see this error

I believe this issue is when using pylsp. I also have pyright running though I do not think that is the issue.

### OS

MacOS

### Error callstack

```emacs-lisp

```

### Anything else?

Investigation and patch from letting claude crank at it:

#### Investigation

Narrowing down via `condition-case` on each hook in `lsp-on-idle-hook` identified `lsp-headerline-check-breadcrumb` as the failing function. Further tracing showed that among 20 document symbols returned by pylsp, 5 had `nil` location (all were duplicates of symbols that appeared earlier with a valid location):

```
total: 20, with-location: 15, without-location: 5
no-location names: logger, _MESSAGE_ID_DOMAIN, _generate_message_id, dispatch_notification_email, send_email
```

#### Fix

Pre-filter out symbols with nil location before the `-keep` destructuring step:

```elisp
(defun lsp--symbols-informations->document-symbols-hierarchy (symbols-informations current-position)
  "Convert SYMBOLS-INFORMATIONS to symbols hierarchy on CURRENT-POSITION."
  (--> symbols-informations
    (-filter (lambda (sym) (gethash "location" sym)) it)  ; guard against nil location
    (-keep (-lambda ((symbol &as &SymbolInformation :location (&Location :range)))
             (when (lsp-point-in-range? current-position range)
               (lsp--symbol-information->document-symbol symbol)))
           it)
    (sort it (-lambda ((&DocumentSymbol :range (&Range :start a-start-position :end a-end-position))
                       (&DocumentSymbol :range (&Range :start b-start-position :end b-end-position)))
               (and (lsp--position-compare b-start-position a-start-position)
                    (lsp--position-compare a-end-position b-end-position))))))
```

Tested in-memory via `emacsclient --eval`: the error stops immediately after applying this redefinition.

#### Environment

- lsp-mode 20260408.702
- Emacs 30.2 (darwin)
- pylsp (python-lsp-server)
- Spacemacs

SAY-5:
## Summary

\`lsp--symbols-informations->document-symbols-hierarchy\` destructured

\`\`\`elisp
(-lambda ((symbol &as &SymbolInformation :location (&Location :range)))
\`\`\`

so a \`SymbolInformation\` entry without a \`:location\` field — pylsp occasionally emits duplicate entries this way — triggered an inner destructure of \`(&Location :range)\` against \`nil\`, which reduced to \`(gethash \"range\" nil)\` and raised

\`\`\`
(wrong-type-argument hash-table-p nil)
\`\`\`

on every \`lsp--on-idle\` tick while \`lsp-headerline-breadcrumb-mode\` had the \`symbols\` segment active.

## Fix

Switch to the optional \`:location?\` binding, then \`when-let*\` the location + range before checking \`lsp-point-in-range?\`. Entries without a location fall out silently — \`-keep\` drops \`nil\` returns already.

\`\`\`elisp
(-keep (-lambda ((symbol &as &SymbolInformation :location?))
         (when-let* ((loc location?)
                     (range (lsp:location-range loc)))
           (when (lsp-point-in-range? current-position range)
             (lsp--symbol-information->document-symbol symbol))))
       it)
\`\`\`

Fixes #5047

## Test plan

- Matches the existing optional-field convention (\`:container-name?\`, \`:deprecated?\`) already in use at \`lsp-mode.el:6664\`, \`7470\`.
- \`lsp:location-range\` is the same helper used elsewhere in the file (e.g. \`lsp-mode.el:5460\`) so no new API surface.
- Local byte-compile not run because \`emacs\` isn't on this workstation; the emacs-lsp CI byte-compile pipeline will validate.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6680,9 +6680,8 @@ and destructuring `(&Location :range)' off a nil `:location' blew
 up the breadcrumb idle timer with
 `(wrong-type-argument hash-table-p nil)' (issue #5047)."
   (--> symbols-informations
-    (-keep (-lambda ((symbol &as &SymbolInformation :location?))
-             (when-let* ((loc location?)
-                         (range (lsp:location-range loc)))
+    (-keep (-lambda ((symbol &as &SymbolInformation :location))
+             (when-let* ((range (and location (lsp:location-range location))))
                (when (lsp-point-in-range? current-position range)
                  (lsp--symbol-information->document-symbol symbol))))
            it)
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -6672,11 +6672,19 @@ perform the request synchronously."
                             :detail? container-name?))
 
 (defun lsp--symbols-informations->document-symbols-hierarchy (symbols-informations current-position)
-  "Convert SYMBOLS-INFORMATIONS to symbols hierarchy on CURRENT-POSITION."
+  "Convert SYMBOLS-INFORMATIONS to symbols hierarchy on CURRENT-POSITION.
+
+Skips SymbolInformation entries whose `:location' field is missing
+— pylsp occasionally emits duplicate entries without `:location',
+and destructuring `(&Location :range)' off a nil `:location' blew
+up the breadcrumb idle timer with
+`(wrong-type-argument hash-table-p nil)' (issue #5047)."
   (--> symbols-informations
-    (-keep (-lambda ((symbol &as &SymbolInformation :location (&Location :range)))
-             (when (lsp-point-in-range? current-position range)
-               (lsp--symbol-information->document-symbol symbol)))
+    (-keep (-lambda ((symbol &as &SymbolInformation :location?))
+             (when-let* ((loc location?)
+                         (range (lsp:location-range loc)))
+               (when (lsp-point-in-range? current-position range)
+                 (lsp--symbol-information->document-symbol symbol))))
            it)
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Sounds good to me! Thank you!

PR #5050, opened 2026-04-28
Fix clojure-lsp URLs

pikariop:
Hello! I was reading the Clojure LSP documentation and noticed that the Github repository URLs point to a fork. Thought I might as well send in a patch. Please enjoy.


jcs090218 (approved):

jcs090218:
Hmm... not sure why do we even point to the fork in the first place.

Anyways, thank you for updating the URL!

PR #5058, opened 2026-05-04
Rust-Analyzer fix resolve failure on imports.as_underscore value being null

xitep:
Client side fix for rust-lang/rust-analyzer#22284.


jcs090218 (approved):
LGTM. :)

PR #5061, opened 2026-05-10
feat: install lsp tailwindcss

merrickluo:
Migrate the https://github.com/merrickluo/lsp-tailwindcss client into lsp-mode's clients/ directory. This was previously maintained as a standalone package — moving it here reduces the setup burden for users since it will be bundled with lsp-mode directly.

- Add clients/lsp-tailwindcss.el with full support for the https://github.com/tailwindlabs/tailwindcss-intellisense language server (completions, hover, linting, code actions)
- Supports both Tailwind CSS v3 (config file detection) and v4+ (package.json version detection)
- Includes server auto-download from the VS Code extension, configurable lint rules, and experimental rustywind class sorting
- Add unit tests and changelog entry

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 10.0.1
+  * Add support for [[https://github.com/tailwindlabs/tailwindcss-intellisense][Tailwind CSS]] language server.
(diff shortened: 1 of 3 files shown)


jcs090218 (approved):

jcs090218:
LGTM. So the standalone version would be deprecated? 🤔

merrickluo:
thanks @jcs090218, yeah I'll deprecate the standalone repo.

PR #5063, opened 2026-05-17
Replace (not (null x)) with (and x t) in parser-on-message

alberti42:
This is a small follow-up to PR #5055 to implement a small cosmetic fix—after the recommendation
by `melpazoid` package-review tool.

The `has-method', `has-id', and `has-error' bindings in the `lsp--parser-on-message' patch were
originally `(not (null x))' to coerce a boolean instead of a truthy value.

The fix in this PR preserves the intent of an explicit boolean coercion, so each `has-*' binding holds
strictly t/nil to match its predicate-style name, but it avoids a confusing double-negation. It silences
the melpazoid double-negation warning.

Discovered using the MELPA review tool `melpazoid'.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -7394,9 +7394,9 @@ server. WORKSPACE is the active workspace."
         (let* ((client (lsp--workspace-client workspace))
                (method (json-get json-data "method"))
                (raw-id (json-get json-data "id"))
-               (has-method (not (null method)))
-               (has-id (not (null raw-id)))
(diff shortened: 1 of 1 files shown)


jcs090218 (approved):

jcs090218:
Can you rebase this? Thank you!

alberti42:
> Can you rebase this? Thank you!

Yes, I rebased it. There were no conflicts to resolve.

jcs090218 (approved):

jcs090218:
Thank you!

Issue #5060, opened 2026-05-08
Diganostics does not update with certain language server


drcxd:
### Thank you for the bug report

- [x] I am using the latest version of `lsp-mode` related packages.
- [x] I checked [FAQ](https://emacs-lsp.github.io/lsp-mode/page/faq/) and [Troubleshooting](https://emacs-lsp.github.io/lsp-mode/page/troubleshooting/) sections
- [x] You may also try reproduce the issue using clean environment using the following command: `M-x lsp-start-plain`

### Bug description

With [emmylua-analyzer-rust](https://github.com/EmmyLuaLs/emmylua-analyzer-rust) as lua language server, the diagnostics does not update after code changes. For example, if there is a diagnostic on line 5, after I insert an empty line at line 5, the diagnostic still points to line 5 rather than line 6.

### Steps to reproduce

Install emmylua-analyzer-rust, register it in lsp-mode, and open an empty Lua file. Write some code that would trigger diagnostics, such as unused function parameters. There would be no new diagnostics information added. If open the file again, the diagnostics would come up, but any edit would not cause updating.

### Expected behavior

The diagnostics should update after code changes.

### Which Language Server did you use?

Lua and [emmlua-analyzer-rust](https://github.com/EmmyLuaLs/emmylua-analyzer-rust).

### OS

Windows

### Error callstack

```emacs-lisp

```

### Anything else?

I have contacted the author of the language server, who claims the client should send `textDocument/diagnostic` after file changes to get updated diagnostic information. See this [issue](https://github.com/EmmyLuaLs/emmylua-analyzer-rust/issues/1064) for our correspondence.

Debugging the code, I find that lsp-mode only sends`textDocument/didChagne` in `lsp--flush-delayed-changes` for this particular language server. This behavior is determined by the code in `lsp-on-change`, with `(lsp--workspace-sync-method workspace)` returns 1 and `lsp-debounce-full-sync-notifications` set to `t` by default.

oscarfv:
Could you please that the problem persists after updating `lsp-mode`? I wonder if https://github.com/emacs-lsp/lsp-mode/pull/5059 fixed it.

drcxd:
[...]

PR #5067, opened 2026-05-19
[fix] Update diagnostics when flushing delayed changes

drcxd:
This should fix issue https://github.com/emacs-lsp/lsp-mode/issues/5060.

Simply requesting diagnostics when flushing delayed changes.


jcs090218 (approved):

jcs090218:
Thank you!

PR #5071, opened 2026-05-31
Implement support for the new lenses for Magik

sebastiaanspeck:
The "magik-utils" and "magik-session" come from https://github.com/roadrunner1776/magik.

Please let me know if this requirement when using the Magik LSP needs to be handled differently.


jcs090218 (approved):

jcs090218:
LGTM. Thank you!

PR #5078, opened 2026-06-09
Fix lsp-tailwindcss-add-on-mode default

claritystorm:
Enables lsp-tailwindcss-add-on-mode by default so that the Tailwind CSS language server runs alongside TypeScript language servers instead of blocking them in TypeScript buffers.


jcs090218 (approved):

jcs090218:
Thank you!

PR #5079, opened 2026-06-09
Fix Tailwind CSS server detection and download

claritystorm:
A prompt to automatically install the language server is never shown when node is found. lsp-mode checks for node instead of the server package. lsp-install-server also fails because curl does not decompress gzip from the Visual Studio Code Marketplace.

Checks for the server package instead of node. Uses --compressed with curl to decompress gzip on save.


jcs090218 (approved):

jcs090218:
LGTM! Thank you!

PR #5088, opened 2026-06-28
fix: cache LSP line position lookups

yuansl:
Cache line-start positions while translating LSP line/character pairs to buffer points, invalidating the cache whenever the buffer character tick changes.

This avoids repeatedly scanning from the beginning of the buffer for dense position conversions while preserving the existing widened conversion path.

### DIFF
--- a/lsp-mode.el
+++ b/lsp-mode.el
@@ -1881,14 +1881,31 @@ etc."
 ;; Forward-declare; actual `defvar-local' is in the "Position encoding" section.
 (defvar lsp--move-to-column-function)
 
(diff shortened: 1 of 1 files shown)


yuansl:
This fixes being stuck in lsp-go mode(emacs 32.0.50 (build 2, aarch64-apple-darwin23.2.0, NS
 appkit-2487.30 Version 14.2 (Build 23C64)), with native compilation enabled) on macOS 26.5.1.

yuansl:
hi, @jcs090218 , Could you please give this PR a code review when you have time? Thanks

PR #5089, opened 2026-06-28
Fix Magik server detection when JAR is missing

claritystorm:
The Magik client requires both a Java runtime and a downloaded JAR at `lsp-magik-ls-path`. The default installation check only verifies that Java is on PATH via `executable-find`, so the server appears as installed even when the JAR has never been downloaded.

This change adds a test-command to check both.


jcs090218 (approved):

jcs090218 (approved):

jcs090218:
Thank you!

PR #5097, opened 2026-07-15
fix(ruby-lsp): support plist deserialization in code lens handlers

FiloSpaTeam:
Replace direct gethash calls with lsp-get in the Ruby LSP action
handlers. This allows runTest, runTestInTerminal and openFile actions
to work when lsp-mode is compiled with LSP_USE_PLISTS=true.


jcs090218 (approved):

jcs090218:
Thank you!

PR #5103, opened 2026-08-02
add biome client

pradyuman:
## Summary

#5007 added `lsp-biome` to `lsp-client-packages` and the documentation navigation, but did not add a client implementation. Because client packages are loaded optionally, Biome support is otherwise silently skipped unless an external package provides the feature. This PR completes that in-tree integration.

The implementation follows existing in-tree client and dependency APIs and is based on Biome's documented LSP contract. It take a different approach from the external `cxa/lsp-biome` package, and instead aims to match default `biome lsp-proxy` behavior as closely as possible (including relying on `lsp-mode` code action facilities, exposing biome languge server configuration settings, supporting multi-root, etc).

## Testing

- `nix shell nixpkgs#eask-cli -c make compile EASK=eask`
- targeted `package-lint` for `clients/lsp-biome.el`
- targeted `checkdoc` for `clients/lsp-biome.el`
- generated `docs/page/lsp-biome.md` successfully
- initialized Biome 2.5.5 through `lsp-mode` and formatted a TypeScript buffer using the real LSP server

### DIFF
--- a/CHANGELOG.org
+++ b/CHANGELOG.org
@@ -1,5 +1,6 @@
 * Changelog
 ** Unreleased 10.0.1
+  * Add a built-in Biome client.
   * Add ~lsp-pylsp-plugins-ruff-format-enabled~ to expose python-lsp-ruff's ~formatEnabled~ setting.
   * Fix ~lsp-install-server~ download from the Visual Studio Code Marketplace.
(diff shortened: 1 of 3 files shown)


justinbarclay:
Hey, as the person who started the work with Biome and then forgot to add the Biome client, I approve this PR (depending on your stance on LLM-authored code). Pinging @jcs090218 too, as he reviewed the last PR.

(Sorry for taking so long to look into this too; shortly after that PR was approved, I had to move on to non-JS projects for a while, and I am just now coming back to them)

